PR #34, opened 2019-02-22
Movie financing app initial setup & connexion

Jackseed:
> initial app structure
> creation of Movie state
> test linked to issue #31
> import of Auth lib for connexion 
> financing homepage


apps/movie-financing/movie-financing-e2e/src/integration/app.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+});
+
+describe('story 1: Je peux me connecter et créer un Film', () => {
+  beforeEach(() => {
+    cy.clearCookies();
+    cy.visit('/');
+    cy.viewport('macbook-15');
+  });
+
+  it('', () => {
```
lsenta: `it.skip` to avoid pushing broken tests to master. We'll enable it when the test is supposed to be passing.

PR #94, opened 2019-03-25
Delivery views

ioaNikas:
Delivery views for : 
- Movie materials list
- Delivery materials list
- WIP : delivery list


apps/delivery/delivery/src/app/app-routing-module.ts
```diff
@@ -0,0 +1,42 @@
...
+
+// Guards
+import { MovieGuard } from '@blockframes/movie';
+
+export const routes: Routes = [
+  { path: '', redirectTo: 'delivery', pathMatch: 'full' },
+  //{path: 'auth', loadChildren: '@blockframes/auth#AuthModule'},
+
+  {
+    path: 'delivery',
```
GrandSchtroumpf: If the route of the application is `www.blockframes.com/delivery`, then this route will looks like : `www.blockframes.com/delivery/delivery`...

apps/delivery/delivery/src/app/delivery/home/home.component.ts
```diff
@@ -0,0 +1,16 @@
...
+  selector: 'delivery-home',
+  templateUrl: './home.component.html',
+  styleUrls: ['./home.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class HomeComponent implements OnInit {
+
+  constructor() { }
+
+  ngOnInit() {
```
GrandSchtroumpf: if `ngOnInit` is not needed don't use it.

apps/delivery/delivery/src/app/delivery/list/list.component.ts
```diff
@@ -0,0 +1,32 @@
+import { Component, OnInit } from '@angular/core';
+import { Observable } from 'rxjs';
+import { MovieQuery, Movie } from '@blockframes/movie';
+import { DeliveryService, Delivery } from '@blockframes/delivery';
+import { Location } from '@angular/common';
+
+@Component({
+  selector: 'delivery-list',
+  templateUrl: './list.component.html',
+  styleUrls: ['./list.component.scss']
```
GrandSchtroumpf: Is there a reason not to use `changeDectectionStrategy.OnPush` ?

apps/delivery/delivery/src/app/delivery/movie-materials/movie-materials.component.ts
```diff
@@ -0,0 +1,41 @@
...
+import { Movie, MovieQuery } from '@blockframes/movie';
+import { Observable } from 'rxjs';
+import { Location } from '@angular/common';
+import { Material } from '../../material/+state';
+import { DeliveryService } from '@blockframes/delivery';
+
+@Component({
+  selector: 'delivery-movie-materials',
+  templateUrl: './movie-materials.component.html',
+  styleUrls: ['./movie-materials.component.scss']
```
GrandSchtroumpf: Same as above `ChangeDetectionStrategy.OnPush`.

libs/delivery/src/lib/+state/delivery.service.ts
```diff
@@ -0,0 +1,148 @@
...
+    private deliveryQuery: DeliveryQuery,
+    private deliveryStore: DeliveryStore,
+    private materialStore: MaterialStore
+  ) {}
+
+  public deliveredToggle(material: Material, movieId: string) {
+    // Change material 'delivered' property value to true or false when triggered
+    return this.firestore
+      .collection<Movie>(`movies/${movieId}/materials/`)
+      .doc(material.id)
```
GrandSchtroumpf: you can do 
```typescript
this.firestore.doc<Material>(`movies/${movieId}/materials/${material.id}`)
```

libs/delivery/src/lib/+state/delivery.service.ts
```diff
@@ -0,0 +1,148 @@
...
+      .update({ delivered: !material.delivered });
+  }
+
+  public getDeliveryMaterialsByActiveMovie() {
```
GrandSchtroumpf: You are not using a parameter in the function, you can make it a getter (or readonly parameter) : 
```typescript
public get getDeliveryMaterialsByActiveMovie$() {
 return ...
}
```

libs/delivery/src/lib/+state/delivery.service.ts
```diff
@@ -0,0 +1,148 @@
...
+    // Get and sort the active movie's delivery's materials by category
+    return this.movieQuery.selectActive().pipe(
+      filter(movie => !!movie),
+      switchMap(movie =>
+        this.firestore.collection<Material>(`movies/${movie.id}/materials`).valueChanges()
+      ),
+      map(materials =>
```
GrandSchtroumpf: For better readibility make it an external function to the class : 
```typescript
function materialByCategory(materials: Material[]) {
  return materials.reduce(....)
}

class Query {
  ...
  map(materials => materialByCategory(materials))
}
```

PR #103, opened 2019-03-27
Feature/story 101

phpgeekfr:


libs/ui/src/lib/toolbar/header/header.component.html
```diff
@@ -1,5 +1,12 @@
 <mat-toolbar class="mat-elevation-z2" fxLayout="row-reverse" fxLayoutAlign="space-between center">
   <span>Blockframes</span>
+
+  <!-- Dynamic content -->
+  <section class="app-icon-container-toolbar">
+    <ng-content select=".top-toolbar-mat-icon"></ng-content>
```
GrandSchtroumpf: It might be better to use `ng-transclude` than classes because, you would look into the scss file if you see a class (I did that...).

phpgeekfr: Yes, this was just a test made in a hurry.

PR #131, opened 2019-04-02
Create delivery v2

ioaNikas:
updated layout for better navigation through the app.


apps/delivery/delivery/src/app/layout/layout.component.html
```diff
@@ -17,9 +17,16 @@
     <section class="top-toolbar-context-menu">
       <mat-nav-list>
         <mat-list-item>
-          <a [routerLink]="['/delivery/explorer']">Explorer</a>
+          <a [routerLink]="['/layout/home']">Explorer</a>
```
GrandSchtroumpf: You don't need `[]`

apps/delivery/delivery/src/app/layout/layout.component.html
```diff
@@ -17,9 +17,16 @@
...
+          <mat-divider [vertical]="true"></mat-divider>
```
GrandSchtroumpf: You can write `<mat-divider vertical></mat-divider>`

libs/material/src/lib/template/template.module.ts
```diff
@@ -21,13 +21,17 @@ import {
 import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatDialogModule } from '@angular/material/dialog';
 import { ReactiveFormsModule } from '@angular/forms';
-import { CategoryListComponent } from './category-list/category-list.component';
 import { MaterialModule } from '../material/material.module';
 
 
+export const templateRoutes: Routes = [
```
GrandSchtroumpf: Don't export it

PR #132, opened 2019-04-02
Issue#70 rename components template

Hugo-Hontans:


libs/material/src/lib/template/template-item/template-item.component.html
```diff
@@ -0,0 +1,24 @@
...
+  <mat-card-title fxFlex="80">
+    {{ template.name }}
+  </mat-card-title>
+
+  <button class="trigger" mat-icon-button [matMenuTriggerFor]="actions">
+    <mat-icon>more_vert</mat-icon>
+  </button>
+
+  <mat-menu #actions="matMenu">
+    <button mat-menu-item routerLink="../{{ template.id }}">
```
GrandSchtroumpf: `routerLink="../{{ template.id }}"` => `[routerLink]="['../', template.id]"`

PR #147, opened 2019-04-04
Movie teamwork tab - first implementation

Jackseed:


libs/movie/src/lib/stakeholder/list/list.component.ts
```diff
@@ -1,4 +1,8 @@
 import { Component, OnInit, ChangeDetectionStrategy } from '@angular/core';
+import { StakeholderService } from '../+state/stakeholder.service';
+import { Observable } from 'rxjs';
+import { Stakeholder } from '../+state/stakeholder.model';
```
GrandSchtroumpf: import directly from `../+state` for `Stakeholder` and `StakeholderService`.

PR #153, opened 2019-04-05
Feature: user can add stakeholders with authorizations in deliveries

ioaNikas:


libs/material/src/lib/delivery/delivery-list/delivery-list.component.ts
```diff
@@ -44,7 +46,8 @@ export class DeliveryListComponent implements OnInit {
   }
 
   public openDialog() {
-    this.dialog.open(TemplatePickerComponent, { width: '80%', height: '80%' });
+    this.store.setActive(this.db.createId());
```
GrandSchtroumpf: You cannot setActive an id before creating the entity in the store...

PR #161, opened 2019-04-08
Issue#158 materials copied into movie

ioaNikas:
Firebase function works no matters the number of stakeholders / 
Edit delivery now removes signatures from validated array

### DIFF
--- a/libs/material/src/lib/delivery/+state/delivery.service.ts
+++ b/libs/material/src/lib/delivery/+state/delivery.service.ts
@@ -51,14 +51,13 @@ export class DeliveryService {
    * @param templateId if templateId is present, the materials sub-collection is populated with materials from this template
    */
   public addDelivery(templateId?: string) {
-    const id = this.query.getActiveId();
-    const orgId = this.organizationQuery.getActiveId();
+    const id = this.db.createId();
     const stakeholderId = this.db.createId();
+    const orgId = this.organizationQuery.getActiveId();
     const delivery = createDelivery({ id, movieId: this.movieQuery.getActiveId() });
     const stakeholder = createStakeholder({ id: stakeholderId, orgId });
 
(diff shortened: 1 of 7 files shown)


libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -33,32 +36,31 @@ export class DeliveryService {
...
    *
    * @param templateId if templateId is present, the materials sub-collection is populated with materials from this template
    */
   public addDelivery(templateId?: string) {
-    const id = this.query.getActiveId();
-    const orgId = this.organizationQuery.getActiveId();
+    const id = this.db.createId();
     const stakeholderId = this.db.createId();
+    const orgId = this.organizationQuery.getActiveId();
     const delivery = createDelivery({ id, movieId: this.movieQuery.getActiveId() });
```
GrandSchtroumpf: you're doing `const id = this.db.createId();` and `const stakeholderId = this.db.createId();`, so you can do : 
```
const movieId = this.movieQuery.getActiveId()
```

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -95,9 +108,17 @@ export class DeliveryService {
...
+  /** Adds a stakeholder with specific authorization to the delivery */
   public addStakeholder(stakeholder: Stakeholder, authorization: string) {
     this.db
       .doc<Stakeholder>(`deliveries/${this.query.getActiveId()}/stakeholders/${stakeholder.id}`)
       .set({ ...stakeholder, authorizations: [authorization] });
   }
+
+  /** Returns true if number of signatures in validated equals number of stakeholders in delivery sub-collection */
+  public isDeliveryValidated() : boolean {
+    return this.query.getActive().validated.length === this.stakeholderQuery.getCount()
```
GrandSchtroumpf: You don't need the ternary condition

PR #182, opened 2019-04-11
Issue#180 fix relayer bug

pldespaigne:
- [x] fix relayer bug `HTTP 414` on contract deploy

### DIFF
--- a/functions/src/relayer.ts
+++ b/functions/src/relayer.ts
@@ -1,11 +1,8 @@
 import { Wallet, ContractFactory, Contract, utils, getDefaultProvider, providers } from 'ethers';
-import { functions } from './firebase';
 import * as ERC1077 from './contracts/ERC1077.json';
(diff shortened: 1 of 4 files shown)


functions/src/relayer.ts
```diff
@@ -68,7 +59,9 @@ export const relayerCreateLogic = async (
     hash,
     relayer.wallet.address
   );
+  console.log(`tx sent (register) : ${registerTx.hash}`); // display tx to firebase logging
   const erc1077 = await relayer.erc1077Factory.deploy(key);
+  console.log(`tx sent (deploy) : ${erc1077.deployTransaction.hash}`); // display tx to firebase logging
```
GrandSchtroumpf: Remove console.log.

pldespaigne: but those console.log are usefull to see the usage of the functions in the firebase console, it's kinda the only way to see outgoing tx, should I remove them anyway ?

PR #185, opened 2019-04-11
review and fix addDelivery()

Hugo-Hontans:


libs/material/src/lib/delivery/+state/delivery.query.ts
```diff
@@ -127,4 +130,12 @@ export class DeliveryQuery extends QueryEntity<DeliveryState, Delivery> {
       map(delivery => delivery.validated.includes(id))
     )
   }
+
+  /** Find the stakeholder from the movie and logged user organizations */
+  public findActiveStakeholder() {
+    const stakeholders = this.stakeholderQuery.getAll();
+    const orgIds = this.organizationQuery.getAll().map(org => org.id);
+    const activeStakeholder= stakeholders.find(({orgId}) => orgIds.includes(orgId));
+    return activeStakeholder;
```
GrandSchtroumpf: just return it directly : `return stakeholders.find(({orgId}) => orgIds.includes(orgId));`

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -81,20 +81,19 @@ export class DeliveryService {
...
-    const organization = this.organizationQuery.getActive();
     const delivery = createDelivery({ id, movieId });
-    const stakeholder = createStakeholder({
-      id: stakeholderId,
-      orgId: organization.id,
+    const deliveryStakeholder = createStakeholder({
+      id: stakeholder.id,
+      orgId: stakeholder.orgId,
       authorizations: ['canValidateDelivery']
     });
```
GrandSchtroumpf:
```typescript
const { id, orgId } = stakeholder;
const authorizations = ['canValidateDelivery'];
const deliveryStakeholder = createStakeholder({ id, orgId, authorizations });
```

Hugo-Hontans: id is already used by :
```
const id = this.db.createId();
const delivery = createDelivery({ id, movieId });
```

PR #200, opened 2019-04-15
Issue#195 add contract address to profile

pldespaigne:
- [x] set erc1077 contract address in the firestore user document
- [x] display the contract address in the user profile

### DIFF
--- a/functions/src/relayer.ts
+++ b/functions/src/relayer.ts
@@ -33,8 +33,10 @@ export const initRelayer = (config: any): Relayer => {
   };
 };
 
+interface UserInfos { uid:string, username: string; key: string };
+
 export const relayerCreateLogic = async (
-  { uid, username, key }: { uid:string, username: string; key: string },
+  { uid, username, key }: UserInfos,
   config: any
 ) => {
   const relayer: Relayer = initRelayer(config);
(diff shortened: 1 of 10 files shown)


functions/src/relayer.ts
```diff
@@ -33,7 +34,7 @@ export const initRelayer = (config: any): Relayer => {
 };
 
 export const relayerCreateLogic = async (
-  { username, key }: { username: string; key: string },
+  { uid, username, key }: { uid:string, username: string; key: string },
```
GrandSchtroumpf: Create an interface : 
```
interface RelayerInfos {
  uid: string
  username: string
  key: string
}
```

functions/src/relayer.ts
```diff
@@ -66,6 +67,8 @@ export const relayerCreateLogic = async (
   const waitForDeploy: Promise<Contract> = erc1077.deployed();
   const [deployedErc1077] = await Promise.all([waitForDeploy, waitForRegister]);
 
+  db.collection('users').doc(uid).update({identity: deployedErc1077.address}).catch(err => console.error(err)); // store contract address in firestore
```
GrandSchtroumpf: use try / catch

libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -33,7 +33,7 @@ export class AuthService {
     await this.create(user.user);
```
GrandSchtroumpf: Change `user.user` to `cred.user` or tsomething like that

libs/ethers/src/lib/relayer-wallet/relayer-wallet.spec.ts
```diff
@@ -73,7 +73,7 @@ describe('RelayerWallet', () => {
     });
     test('signup', async () => {
       const relayerWalletK = new RelayerWallet(vault, relayer, provider);
-      await relayerWalletK.signup(name, password);
+      await relayerWalletK.signup('', name, password);
```
GrandSchtroumpf: `await relayerWalletK.signup('UID', name, password);`

Issue #194, opened 2019-04-15
Add e2e tests to the wallet/relayer


pldespaigne:
Add e2e tests to the wallet/relayer part at the user signup

PR #205, opened 2019-04-16
add e2e for signup

pldespaigne:
⚠️This PR refer to issue #194 and not 192

- [x] add e2e tests for user signup & erc1077 address


apps/delivery/delivery-e2e/src/integration/app.spec.ts
```diff
@@ -84,3 +88,21 @@ describe('I m a user and i can save a delivery as template', () => {
     p.clickMenu(TEMPLATE_NAME_1);
   });
 });
+
+describe('I\'m a paranoid user, I would like to create an account and verify that I own a contract on the blockchain', () => {
+  it('should type values in signup form', () => {
+    // Connexion
+    let p: any = new LandingPage();
+    p.fillSignupEmail(generateRandomEmail());
+    p.fillSignupPassword('azerty123');
```
ioaNikas: You should use const instead of raw strings

apps/delivery/delivery-e2e/src/integration/app.spec.ts
```diff
@@ -40,8 +44,8 @@ describe('I m a user and i can save a delivery as template', () => {
   it('', () => {
     // Connexion
     let p: any = new LandingPage();
-    p.fillEmail('delivery-test-e2e@cascade8.com');
-    p.fillPassword('blockframes4tw');
+    p.fillSigninEmail('delivery-test-e2e@cascade8.com');
+    p.fillSigninPassword('blockframes4tw');
```
ioaNikas: You should use const instead of raw strings (I didn't.)

apps/delivery/delivery-e2e/src/integration/app.spec.ts
pldespaigne: - [x] done in 50787d3

apps/delivery/delivery-e2e/src/integration/app.spec.ts
pldespaigne: - [x] done in 50787d3

PR #206, opened 2019-04-16
style material-item and template-form

Hugo-Hontans:


libs/material/src/lib/template/+state/template.query.ts
```diff
@@ -29,6 +29,8 @@ export class TemplateQuery extends QueryEntity<TemplateState, Template> {
 
   public form$ = this.select(state => state.form);
 
+  public templateActive$ = this.selectActive(template => template)
```
ioaNikas: It is the same as ```this.templateQuery.selectActive();```

PR #219, opened 2019-04-18
Anchor template form

Hugo-Hontans:


apps/delivery/delivery/src/app/app-routing-module.ts
```diff
@@ -51,6 +51,7 @@ export const routes: Routes = [
 @NgModule({
   imports: [RouterModule.forRoot(routes,{
     anchorScrolling: 'enabled',
+    onSameUrlNavigation: "reload",
```
ioaNikas: Use simple 'quotes' as you did on 'enabled'.

libs/material/src/lib/delivery/delivery-form/delivery-form.component.scss
```diff
@@ -24,13 +24,22 @@ section {
   margin: 0 10%;
 }
 
+.sign {
+  margin-top: 7%;
+  margin-bottom: 7%;
+  font-weight: bold;
+  font-size: 13px;
+  color: #6100ed;
```
ioaNikas: Could you use ```color="primary"``` in HTML since it is on a button ?

PR #220, opened 2019-04-18
Make blockframes prettier

ioaNikas:


libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -175,7 +177,9 @@ export class DeliveryService {
         ? deliveryStakeholder.authorizations
         : [...deliveryStakeholder.authorizations, authorization];
       this.db
-        .doc<Stakeholder>(`deliveries/${this.query.getActiveId()}/stakeholders/${deliveryStakeholder.id}`)
+        .doc<Stakeholder>(
+          `deliveries/${this.query.getActiveId()}/stakeholders/${deliveryStakeholder.id}`
+        )
```
Hugo-Hontans: indent correctly

libs/material/src/lib/delivery/delivery-list/delivery-list.component.ts
```diff
@@ -18,12 +18,11 @@ import { takeWhile } from 'rxjs/operators';
 })
 export class DeliveryListComponent implements OnInit, OnDestroy {
   public movie$: Observable<Movie>;
-  public deliveries$: Observable<Delivery[]>;
+  public deliveries$: any;
```
Hugo-Hontans: type observable

libs/material/src/lib/delivery/+state/delivery.query.ts
```diff
@@ -30,6 +31,7 @@ export class DeliveryQuery extends QueryEntity<DeliveryState, Delivery> {
     private stakeholderQuery: StakeholderQuery,
     private organizationQuery: OrganizationQuery,
     private db: AngularFirestore,
+    private fireQuery: FireQuery
```
GrandSchtroumpf: Don't put it inside the Query.

libs/material/src/lib/delivery/delivery-list/delivery-list.component.html
```diff
@@ -1,6 +1,6 @@
 <main class="container">
   <section *ngIf="(movie$ | async) as movie" class="delivery-list">
-    <article fxLayoutAlign="space-between">
+    <article class="sticky" fxLayoutAlign="space-between">
```
GrandSchtroumpf: Sticky class is commented. You don't use it

PR #222, opened 2019-04-18
akita bug fix & merge new movie list

pldespaigne:
- [x] refactor the org query to properly get movies inside org obj
- [x] refactor movie list component
- [x] properly solve merge conflict from new list component


libs/movie/src/lib/movie/list/list.component.ts
```diff
@@ -43,4 +40,7 @@ export class ListComponent implements OnInit, OnDestroy {
     this.service.remove(id);
   }
 
+  public trackOrgs(index: number, org: Organization) {
+    return org ? org.id : undefined;
```
GrandSchtroumpf: `return org.id`

libs/organization/src/lib/+state/organization.query.ts
```diff
@@ -1,20 +1,39 @@
...
     super(store);
   }
 
   public getOrgId(name: string) {
-    return this.getAll().find(org => org.name === name).id;
+    const orgs = this.getAll({
+      filterBy: org => org.name === name,
+      limitTo: 0
+    });
+    return orgs[0].id;
```
GrandSchtroumpf: Why ?

libs/movie/src/lib/movie/list/list.component.ts
pldespaigne: - [x] done

libs/organization/src/lib/+state/organization.query.ts
pldespaigne: - [x] done

PR #224, opened 2019-04-18
Refactoring some Akita x Firebase interaction

pldespaigne:


libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -41,33 +41,18 @@ export class MovieService {
     this.collection.doc(id).update(movie);
   }
 
-  public remove(id: string) {
-    // TODO: make a firebase function
+  public async remove(id: string) {
```
GrandSchtroumpf: Remove async

PR #226, opened 2019-04-19
Prettier deliveries

ioaNikas:


libs/material/src/lib/delivery/movie-materials/movie-materials.component.html
```diff
@@ -39,23 +63,172 @@ <h5>Global progression</h5>
         </mat-card>
       </ng-template>
     </article>
-    <article class="materials" fxLayout="column">
-      <h2>{{ movie.title.original }}</h2>
-      <div *ngFor="let category of (materials$ | async | keyvalue)">
+    <article class="materials">
```
Hugo-Hontans: create a new component

PR #232, opened 2019-04-23
templates fireQuery in guard

pldespaigne:
First try at normalizing firestore x akita
- [x] query the tempalte firestore from template route guard


libs/material/src/lib/template/guards/template-active.guard.ts
```diff
@@ -0,0 +1,51 @@
...
+    private router: Router,
+    private fireQuery: FireQuery
+    ) {
+      super()
+  }
+
+  startListeningOnActive(): void {
+    this.listenOnActive = true;
+    this.query.selectActive().pipe(
+      distinctUntilChanged((a, b) => JSON.stringify(a) === JSON.stringify(b)), // this prevent infinite loop
```
GrandSchtroumpf: Use `distinctUntilChanged((a, b) => a.id === b.id)`.
The distinctUntilChanged will only apply on the `selectActive()` so it shouldn't impact the switchMap.

libs/material/src/lib/template/guards/template-active.guard.ts
```diff
@@ -0,0 +1,51 @@
...
+      switchMap(templ => this.fireQuery.fromQuery(templateActiveQuery(templ.orgId, templ.id))),
+      takeWhile(_ => !!this.listenOnActive),
+    )
+    .subscribe(activeTemplate => {
+      this.store.update(activeTemplate.id, activeTemplate);
+    });
```
GrandSchtroumpf: Do `subscribe(template => this.store.update(template.id, template)` for readibility

libs/material/src/lib/template/guards/template-active.guard.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+  
+  canActivate(route: ActivatedRouteSnapshot): boolean | UrlTree {
+    // TODO what happen if user copy/paste url and Template state wasn't set previously ? Find a falback mechanism
+    this.store.setActive(route.params.templateId);
+    this.startListeningOnActive();
+    return true;
```
GrandSchtroumpf: Do : 
```
const id = route.params.templateId;
if (!id) return this.router.parseUrl(route.url);
if (!this.query.hasEntity(id)) return this.router.parseUrl(route.url);
this.store.setActive(id)
this.startListeningOnActive();
return true;
```

PR #234, opened 2019-04-23
Fancier movie materials

ioaNikas:
New style for delivery view

### DIFF
--- a/libs/material/src/lib/delivery/delivery-view/delivery-view.component.html
+++ b/libs/material/src/lib/delivery/delivery-view/delivery-view.component.html
@@ -1,60 +1,58 @@
-<main class="container">
-  <section fxLayout="row" fxLayoutGap="5%">
-    <article class="progress-spinner-block" fxLayoutAlign="center start">
-      <mat-card
-        *ngIf="(progressionValue$ | async) as value; else empty"
-        fxLayout="column"
-        fxLayoutAlign="center center"
-        class="mat-elevation-z6"
-      >
-        <mat-progress-spinner
-          class="spinner"
-          color="primary"
-          mode="determinate"
-          [value]="value"
-          diameter="200"
-          strokeWidth="13"
+<div *ngIf="(movie$ | async) as movie" class="container">
+    <section fxLayout="row" fxLayoutGap="15px">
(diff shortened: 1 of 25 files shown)


libs/material/src/lib/delivery/delivery-view/delivery-view.component.html
```diff
@@ -1,60 +1,58 @@
...
-          </p>
+          <div class="progress-value">{{ value }} %</div>
+          <p class="global-progression">Global progression</p>
+          <p class="spinner-text"><i>Lorem ipsum dolor sit amet, consectetur adipisicing elit.</i></p>
         </mat-card>
-      </div>
-    </article>
-  </section>
-</main>
+        <ng-template #empty>
```
Jackseed: It would be better to integrate the 0 case inside the calculation of progressionValue

libs/material/src/lib/delivery/movie-material-item/movie-material-item.component.scss
```diff
@@ -0,0 +1,80 @@
...
+  width: 50%;
+}
+
+h5 {
+  color: rgba(0, 0, 0, 0.418);
+  margin: 5px 0;
+}
+
+p {
+  font-size: 90%;
```
Jackseed: would be better to make an absolute font-size like the others

libs/material/src/lib/delivery/movie-materials/movie-materials.component.scss
```diff
@@ -48,11 +43,6 @@ h4 {
   font-size: 14px;
 }
 
-h5 {
-  color: rgba(0, 0, 0, 0.418);
-  margin: 5px 0;
-}
-
 p {
   font-size: 90%;
```
Jackseed: same remark here, it would be better to put an absolute font-size like the others

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -55,4 +60,22 @@ export class MovieService {
       tap((movies: Movie[]) => this.store.set(movies))
     );
   }
+
+  public get activeMovieStakeholders() {
```
Jackseed: This method should be in the stakeholder state

PR #235, opened 2019-04-23
Refactory template form

Hugo-Hontans:


libs/material/src/lib/material/material-delivery-item/material-delivery-item.component.html
```diff
@@ -0,0 +1,24 @@
+<mat-card class="mat-elevation-z2" fxLayout="row">
+    <mat-icon>drag_indicator</mat-icon>
+    <mat-card-content fxFlex="60" fxFlexOffset="10" fxLayout="column">
```
GrandSchtroumpf: Can't you use `width` and `margin` or `padding` instead?

PR #247, opened 2019-04-26
Add movie financing explorer

lsenta:


apps/movie-financing/movie-financing/src/app/app-routing.module.ts
```diff
@@ -27,6 +30,21 @@ export const routes: Routes = [
...
+  {
+    path: 'explorer/search',
+    component: FinancingExplorerSearchComponent,
+    canActivate: [AuthGuard]
+  },
+  {
+    path: 'explorer/movie/:id',
+    component: FinancingExplorerDetailsComponent,
+    canActivate: [AuthGuard]
+  },
```
GrandSchtroumpf: You could use nested architecture to avoid triggering `AuthGuard` each time : 
```typescript
{
  path: 'explorer',
  component: FinancingExplorerComponent,
  canActivate: [AuthGuard],
  children: [
    { path: '', component: FinancingExplorerHomeComponent },
    { path: 'search', component: FinancingExplorerSearchComponent },
    { path: ':movieId', component: FinancingExplorerDetailsComponent }
  ]
}
```

apps/movie-financing/movie-financing/src/app/explorer/details/details.component.html
```diff
@@ -0,0 +1,200 @@
+<financing-explorer-header></financing-explorer-header>
+
+<div class="content">
```
GrandSchtroumpf: Could you use main semantic (<main>, <section>, <article>) as much as possible.
Also we have `<main fxLayout="row" fxLayoutAlign="center stretch">` directives for inline flexbox.

apps/movie-financing/movie-financing/src/app/explorer/home/home.component.html
```diff
@@ -0,0 +1,18 @@
...
+    <div class="financing-movie-list">
+      <financing-movie-card movieID="42"></financing-movie-card>
+      <financing-movie-card movieID="42"></financing-movie-card>
+      <financing-movie-card movieID="42"></financing-movie-card>
+      <financing-movie-card movieID="42"></financing-movie-card>
+    </div>
+  </div>
+</div>
+
+<financing-explorer-footer></financing-explorer-footer>
```
GrandSchtroumpf: Becaus you're use `financing-explorer-header` and `financing-explorer-footer` in home, search and details, you can create a `FinancingExplorerComponent` as a parent of the three (see comment about routes) :
```html
<financing-explorer-header></financing-explorer-header>
<router-outlet></router-outlet>
<financing-explorer-footer></financing-explorer-footer>
```
And you don't event need financing-explorer-header and financing-explorer-footer because you can add them directly inside your component.

lsenta: I'm worried that'll complexify the code and make it harder to reach the demo:
we have 3 pages, 1 header, 3 different heroes, 2 different footers => I can't factor out the footer, and we're one design change away to have the same thing for the header.

GrandSchtroumpf: In you're code the footer is the same for the 3 pages right now. Or at least it doesn't depend on the page. I don't see the problem for that.
What are "3 different heroes" ?

apps/movie-financing/movie-financing/src/app/explorer/details/details.component.html
lsenta: Nice, thanks! (using only `fxLayoutAlign="center"`, `stretch` sets a `max-height: 100%` that breaks the flow).

I use `main`, not sure which components are `section` and which are `article`, we need a ui kit!

apps/movie-financing/movie-financing/src/app/explorer/home/home.component.html
lsenta: (resolved on slack: the target design is mostly non reusable between pages)

PR #257, opened 2019-04-26
sanitize username

pldespaigne:
Sanitize username to prevent relayer crash due to ethers checking if names follows UTS46 normalization


libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -5,6 +5,7 @@ import { AuthStore, User, createUser } from './auth.store';
 import { switchMap, takeWhile } from 'rxjs/operators';
 import { RelayerWallet } from '@blockframes/ethers';
 import { MatSnackBar } from '@angular/material';
+import * as punycode from 'punycode';
```
GrandSchtroumpf: do `import { encode } from 'punycode'` if possible.

PR #258, opened 2019-04-26
Backup & Restore functions for the firestore

lsenta:


functions/src/backup.ts
```diff
@@ -0,0 +1,195 @@
...
+type QueryDocumentSnapshot = admin.firestore.QueryDocumentSnapshot;
+type DocumentReference = admin.firestore.DocumentReference;
+
+interface StoredDocument {
+  docPath: string;
+  content: any;
+}
+
+class Queue {
+  content: string[];
```
GrandSchtroumpf: set the initial value here : `content: string[] = []`

functions/src/users.ts
```diff
@@ -0,0 +1,98 @@
...
+  id: string,
+  email: string,
+}
+
+interface OrgProposal {
+  id: string,
+  name: string,
+}
+
+const onUserCreate = (user: UserRecord) => {
```
GrandSchtroumpf: We do this client side right now. Any reason why to do it inside a function ?

functions/src/users.ts
```diff
@@ -0,0 +1,98 @@
...
+  const incLast: string = String.fromCharCode(prefix.slice(-1).charCodeAt(0) + 1);
+  const prefixEnd: string = prefix.slice(0, -1) + incLast;
+
+  return db.collection('users')
+    .where('email', '>=', prefix)
+    .where('email', '<', prefixEnd)
+    .get()
+    .then(q => {
+      // leave if there are too many results.
+      if (q.size > 10) {
```
GrandSchtroumpf: Why ?

functions/src/users.ts
lsenta: More reliable to do this server side, that's legacy btw, I just structure the code to clean the index.ts

functions/src/users.ts
lsenta: Avoid showing the whole database to a user

GrandSchtroumpf: Then why returning an empty array instead of the 10 first ?

PR #264, opened 2019-04-29
Issue#248 add steps

Hugo-Hontans:


libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -225,12 +248,22 @@ export class DeliveryService {
   ////////////////////////
 
   public suscribeOnDeliveriesByActiveMovie() {
+    function mungeFirestoreDelivery(delivery: Delivery) {
```
GrandSchtroumpf: When reading the name of the function I cannot understand what it does...

libs/material/src/lib/delivery/delivery-settings-form/delivery-settings-form.component.scss
```diff
@@ -0,0 +1,3 @@
+mat-card {
+  margin-bottom: 4%;
```
GrandSchtroumpf: Try to avoid % on margin, it has super random outcomes depending on the screen. Use px instead.

libs/material/src/lib/delivery/delivery-settings-form/delivery-settings-form.component.ts
```diff
@@ -0,0 +1,51 @@
...
+    this.cancelForm.emit();
+  }
+
+  public updateStep() {
+    // If step exists we update step, else we add a new step
+    if (this.step) {
+      this.service.updateStep(this.step, this.form.value);
+    } else {
+      this.service.addStep(this.form.value);
+    }
```
GrandSchtroumpf: use ternary condition here, and call the fonction `upsert` (update or create) : 
```typescript
!!this.step
  ? this.service.updateStep(this.stp, this.form.value)
  : this.service.addStep(this.form.value);
```

libs/material/src/lib/delivery/delivery-settings-form/delivery-settings-form.component.ts
```diff
@@ -0,0 +1,51 @@
...
+import { DeliveryService, Step } from '../+state';
+
+@Component({
+  selector: 'delivery-settings-form',
+  templateUrl: './delivery-settings-form.component.html',
+  styleUrls: ['./delivery-settings-form.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class DeliverySettingsFormComponent implements OnInit {
+  @Output() cancelForm = new EventEmitter();
```
GrandSchtroumpf: I personally use "-ed" version for output event : `cancel` -> `cancelled`. It's easier to find a name.

libs/material/src/lib/delivery/delivery-settings-item/delivery-settings-item.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  styleUrls: ['./delivery-settings-item.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class DeliverySettingsItemComponent implements OnInit {
+  @Input() step: Step;
+  @Output() update = new EventEmitter();
+
+  constructor(private service: DeliveryService) { }
+
+  ngOnInit() {
```
GrandSchtroumpf: Don't put `ngOnInit` if you don't use it.

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -225,12 +248,22 @@ export class DeliveryService {
...
   public suscribeOnDeliveriesByActiveMovie() {
+    function modifyTimestampToDate(delivery: Delivery) {
+      return {
+        ...delivery,
+        steps: delivery.steps.map(step => ({
+          ...step,
+          date: step.date.toDate()
+        }))
+      }
+    }
```
GrandSchtroumpf: You can use mutable operations here because you're using it inside a map : 
```typescript
function modifyTimestampToDate(delivery: Delivery) {
  delivery.steps.forEach(step => step.date = step.date.toDate())
  return delivery
}
```

lsenta: Note, I implemented this with Hugo, this is generally bad practice to modify the object you've been passed, the code gets harder to reason about since you have to look into the function to understand if the data you passed is going to get mutated or not.

purity > side effect.

GrandSchtroumpf: One can argue :  readibily > *.
In this case we have a closure, so it cannot be use outside of the function. In a scope model language, purity is a reference to the outside world (out of the scope). We could put those two lines of code inside the `map` below to be sure you don't have to look into the function to see if it's mutating anything.

PR #271, opened 2019-04-29
Epic/31 movies bis

phpgeekfr:
[WIP] missing functionality to be able to reuse form in another component and be able to add new inputs.


libs/movie/src/lib/movie/form/form.component.scss
```diff
@@ -50,8 +50,8 @@
   position: absolute;
   right: 0px;
   left: 0px;
-  bottom: 0px;
   top: 66px;
+  flex: 1 1 auto;
```
GrandSchtroumpf: Be careful with `flex: 1 1 auto`. Most of the time you want to do `width / height: 100%` instead.

libs/movie/src/lib/movie/form/form.component.scss
```diff
@@ -69,13 +69,50 @@
...
 }
 
+.movie-card {
+  .mat-chip.mat-standard-chip {
+    background-color: #6200EE;
+    color: #FFF;
+    font-weight: bold;
+  }
+
+  ::ng-deep .mat-chip-list-wrapper {
```
GrandSchtroumpf: Super dangerous. This can have pretty bad side-effect if not used correctly (like applying this style on other element that are not in the component). Please check again if you cannot do it another way. Then use with caution.

libs/movie/src/lib/movie/staticModels.ts
```diff
@@ -1161,4 +1161,8 @@ const models = {
   ]
 };
 
+export const getLabelBySlug = function (scope: string, slug: string) {
```
GrandSchtroumpf: We usually do `function getLabelBySlug(scope: string, slug: string)` for normal functions.
Or `const getLabelBySlug = (...) => ` for function that need to have access to the parent "this".

libs/movie/src/lib/movie/form/form.component.scss
phpgeekfr: Is it dangerous even if ::ng-deep in wrapped inside another class ? In this case, `::ng-deep .mat-chip-list-wrapper` is wrapped inside `.movie-card`

libs/movie/src/lib/movie/form/form.component.scss
phpgeekfr: ok, corrected

libs/movie/src/lib/movie/form/form.component.scss
GrandSchtroumpf: Good question. I don't know. Can you make a quick test (put a mat chip list outside of the movie-card class) ?

phpgeekfr: I made a test, if it is wrapped there is no side effects.

libs/movie/src/lib/movie/staticModels.ts
phpgeekfr: updated

PR #276, opened 2019-04-30
Issue#252 notifications materials

ioaNikas:
- [x] Added notifications on material state changes
- [x] New primary color to match wireframes
- [x] Added Lodash library to project


functions/package.json
```diff
@@ -22,13 +22,13 @@
     "firebase-admin": "^7.0.0",
     "firebase-functions": "^2.1.0",
     "line-reader": "^0.4.0",
-    "lodash.sortby": "^4.7.0",
+    "lodash": "^4.17.11",
```
GrandSchtroumpf: Why did you decided to switch to the whole version of lodash ?

functions/src/delivery.ts
```diff
@@ -2,9 +2,8 @@ import { db, functions } from './firebase';
 import { triggerNotifications, prepareNotification } from './notify';
 import { getOrgsOfDelivery, Organization } from './stakeholder';
 
-export const APP_DELIVERY = 'delivery'; // This string refers to svg icon name
+export const APP_DELIVERY = 'delivery'; /** This string refers to svg icon name */
```
GrandSchtroumpf: comment with `/** */` should be above the variable.

functions/src/delivery.ts
```diff
@@ -69,7 +68,7 @@ export const onDeliveryUpdate = async (
 
   const orgs = await getOrgsOfDelivery(delivery.id);
 
-  // Notifications are triggered when a new id is pushed into delivery.validated, which is considered as a signature
+  /** Notifications are triggered when a new id is pushed into delivery.validated, which is considered as a signature */
```
GrandSchtroumpf: Use `/** */` comments for classes, variables, methods and functions only.
- `/** */` comments are for documentation
- `//` comments are for explaining your code.
In this case this should be `//`

functions/src/material.ts
```diff
@@ -0,0 +1,79 @@
...
+import { triggerNotifications, prepareNotification } from './notify';
+
+interface Material {
+  id: string;
+  value: string;
+  deliveriesIds: string[];
+  state: string;
+}
+
+export const onMaterialUpdate = async (
```
GrandSchtroumpf: You don't need to have access to the `this` of the parent scope. Personally I like better to use a normal function instead of an arrow function in this case : 
```typescript
async function onMaterialUpdate(...) { ... }
```

functions/src/material.ts
```diff
@@ -0,0 +1,79 @@
...
+  change: functions.Change<FirebaseFirestore.DocumentSnapshot>,
+  context: functions.EventContext
+) => {
+  if (!change.after || !change.before) {
+    return true;
+  }
+
+  const movie = await getDocument(`movies/${context.params.movieID}`);
+  const material: Material = change.after.data() as Material;
```
GrandSchtroumpf: Either : 
` const material= change.after.data() as Material;`
Or 
`const material: Material = change.after.data()`

libs/material/src/lib/delivery/+state/delivery.query.ts
```diff
@@ -83,8 +83,8 @@ export class DeliveryQuery extends QueryEntity<DeliveryState, Delivery> {
       map(materials => {
         const id = this.getActiveId();
         const totalMaterials = materials.filter(material => material.deliveriesIds.includes(id));
-        const deliveredMaterials = totalMaterials.filter(material => !!material.delivered);
-        return Math.round((deliveredMaterials.length / (totalMaterials.length / 100)) * 10) / 10;
+        const acceptedMaterials = totalMaterials.filter(material => material.state === 'accepted');
+        return Math.round((acceptedMaterials.length / (totalMaterials.length / 100)) * 10) / 10;
```
GrandSchtroumpf: Maybe you can create an intermediary variable to explain what `(acceptedMaterials.length / (totalMaterials.length / 100)) * 10` is.

PR #282, opened 2019-04-30
Delivery form validators

Hugo-Hontans:


libs/material/src/lib/delivery/+state/delivery.model.ts
```diff
@@ -8,6 +8,7 @@ export interface Delivery {
   delivered: boolean;
   stakeholders?: Stakeholder[];
   steps: Step[];
+  dueDate: Date | any;
```
GrandSchtroumpf: why `any` ?

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -35,12 +35,12 @@ export class MaterialService {
...
-            map(materials =>
-              materials.map(material => ({
+            map(materials => {
+              return materials.map(material => ({
                 ...material,
-                step: delivery.steps.find(step => step.id === material.stepId)
-              }))
-            )
+                step: delivery.steps.find(step => step.id === material.stepId) ?  delivery.steps.find(step => step.id === material.stepId) : null
+              }));
```
GrandSchtroumpf: Make it more readable : 
```typescript
map(materials => {
  return materials.map(material => {
    const step = delivery.steps.find(step => step.id === material.stepId) || null;
    return { ...material, step }
  })
})
```

GrandSchtroumpf: But I'm curious why you want to force `null` over `undefined`...

libs/material/src/lib/material/material-delivery-form/material-delivery-add-form.component.ts
```diff
@@ -16,29 +24,55 @@ export class MaterialDeliveryAddFormComponent implements OnInit, OnDestroy {
...
 
     this.query
       .select(state => state.materialDeliveryForm)
       .pipe(
         takeWhile(() => this.isAlive),
         filter(materialDeliveryForm => !!materialDeliveryForm)
       )
-      .subscribe(materialDeliveryForm => this.form.setValue(materialDeliveryForm));
+      .subscribe(materialDeliveryForm => {
+        const form = {...materialDeliveryForm};
```
GrandSchtroumpf: You don't need to do that. RXJS provide immutable outputs. You can write: 
```javascript
.subscribe(form => {
  if (!this.hasStep) delete form.stepId;
  return this.form.setValue(form);
})
```

libs/material/src/lib/material/material-delivery-form/material-delivery-add-form.component.ts
```diff
@@ -16,29 +24,55 @@ export class MaterialDeliveryAddFormComponent implements OnInit, OnDestroy {
...
+        // If the delivery doesn't have steps : don't set stepId to form
+        if (!this.hasStep) delete form.stepId;
+        return this.form.setValue(form);
+      });
+  }
+
+  private createFormWithStep() {
```
GrandSchtroumpf: Hmmm I'm pretty sure you can do it without duplication (making `stepId` optional for example).

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.ts
```diff
@@ -19,34 +19,79 @@ import { Step, DeliveryQuery } from '../../delivery/+state';
...
   ngOnInit() {
     this.steps$ = this.deliveryQuery.steps$;
-    this.form.setValue({
-      value: this.material.value,
-      description: this.material.description,
-      category: this.material.category,
-      stepId: this.material.step.id
+
+    if (this.deliveryQuery.getActive().steps.length > 0) {
+      this.form = this.createFormWithStep();
```
GrandSchtroumpf: I don't understand why you have to duplicate this code with the one above. This should raise an alarm if you're copy pasting code like that : 🚨

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.ts
```diff
@@ -19,34 +19,79 @@ import { Step, DeliveryQuery } from '../../delivery/+state';
...
+          value: this.material.value,
+          description: this.material.description,
+          category: this.material.category,
+          stepId: ''
+        });
+      }
+    }
+  }
+
+  private createFormWithStep() {
```
GrandSchtroumpf: 🚨

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.ts
```diff
@@ -19,34 +19,79 @@ import { Step, DeliveryQuery } from '../../delivery/+state';
...
+      this.hasStep = true;
+      this.setForm(this.material);
+    } else {
+      this.form = this.createFormWithoutStep();
+      this.hasStep = false;
+      this.setForm();
+    }
+  }
+
+  private setForm(material?: Material) {
```
GrandSchtroumpf: This should be done inside the HTML : https://angular.io/guide/reactive-forms

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.ts
```diff
@@ -19,34 +19,79 @@ import { Step, DeliveryQuery } from '../../delivery/+state';
...
+  private createFormWithStep() {
+    return new FormGroup({
+      value: new FormControl(''),
+      description: new FormControl(''),
+      category: new FormControl(''),
+      stepId: new FormControl('')
+    });
+  }
+
+  private createFormWithoutStep() {
```
GrandSchtroumpf: 🚨

libs/material/src/lib/delivery/+state/delivery.model.ts
Hugo-Hontans: I need it for the subscribe : 
```
  public suscribeOnDeliveriesByActiveMovie() {
    function modifyTimestampToDate(delivery: Delivery) {
      delivery.steps.forEach(step => step.date = step.date.toDate());
      if (delivery.dueDate) delivery.dueDate = delivery.dueDate.toDate();
      return delivery
    }
```
I can't apply toDate() to dueDate if its type is Date

libs/material/src/lib/material/+state/material.service.ts
Hugo-Hontans: I don't need finaly

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.ts
Hugo-Hontans: I can't really do that because I use the same HTML for material-delivery-form and material-delivery-add-form or I must do 2 differents html page

libs/material/src/lib/delivery/+state/delivery.model.ts
GrandSchtroumpf: Is it not `Date | string` then ?

libs/material/src/lib/delivery/+state/delivery.query.ts
```diff
@@ -36,6 +36,10 @@ export class DeliveryQuery extends QueryEntity<DeliveryState, Delivery> {
     super(store);
   }
 
+  get hasStep(): boolean {
+    return this.getActive().steps.length > 0 ? true : false;
```
GrandSchtroumpf: You don't need `? true : false`. `this.getActive().steps.length > 0` is already a boolean value.

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.ts
```diff
@@ -19,34 +19,39 @@ import { Step, DeliveryQuery } from '../../delivery/+state';
...
-  constructor(private deliveryQuery: DeliveryQuery,) {}
+  constructor(private deliveryQuery: DeliveryQuery) {}
 
   ngOnInit() {
     this.steps$ = this.deliveryQuery.steps$;
+    this.hasStep = this.deliveryQuery.hasStep;
+    this.setForm();
+  }
+
+  private setForm() {
```
GrandSchtroumpf: You're not using that elsewhere so do everything at the same time : 
```typescript
this.form = new FormGroup({
    value: new FormControl(this.material.value || ''),
    description: new FormControl(this.material.description|| ''),
    category: new FormControl(....),
    stepId: new FormControl(....)
})
```

libs/material/src/lib/delivery/+state/delivery.model.ts
Hugo-Hontans: No it's a Timestamp (back from firebase)

Hugo-Hontans: I must create 2 types for Delivery and Step => 1 for back from firebase and 1 for the application

PR #284, opened 2019-04-30
Fix add stakeholder

lsenta:


libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
```diff
@@ -15,16 +15,17 @@ export class StakeholderService {
...
     private store: StakeholderStore,
     private movieQuery: MovieQuery,
-    private fireQuery: FireQuery,
-  ) {}
+    private fireQuery: FireQuery
+  ) {
+  }
 
-  public get stakeholdersByActiveMovie$(){
+  public get stakeholdersByActiveMovie$() {
```
GrandSchtroumpf: With your function below, this should be removed.

libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
```diff
@@ -66,43 +83,60 @@ export class StakeholderService {
...
-  public get activeMovieStakeholders() {
-    return this.movieQuery
-      .selectActiveId()
-      .pipe(
-        switchMap(id => this.fireQuery.fromQuery(this.getStakeholderList(id))),
-        tap((stakeholders: any) => this.store.set(stakeholders))
-      );
-  }
-
   private getStakeholderList(movieId: string): Query<Stakeholder> {
```
GrandSchtroumpf: This should be of type `Query<Stakeholder[]>` for you not to hardcode `(stakeholders: any)` [here](https://github.com/blockframes/blockframes/pull/284/files#diff-e59883ca999f0ad1242461e434d76b02R42)

libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
```diff
@@ -33,20 +34,28 @@ export class StakeholderService {
     );
   }
 
+  public get activeMovieStakeholders() {
+    return this.movieQuery
+      .selectActiveId()
+      .pipe(
+        switchMap(id => this.fireQuery.fromQuery(this.getStakeholderList(id))),
+        tap((stakeholders: any) => this.store.set(stakeholders))
```
GrandSchtroumpf: See comment below about the type of `this.getStakeholderList(id)`

libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
```diff
@@ -33,20 +34,28 @@ export class StakeholderService {
...
+      );
+  }
+
   public getAllStakeholdersWithOrg(stakeholders: Stakeholder[]) {
```
GrandSchtroumpf: We don't need this function anymore because we have `getStakeholderList` that take care of that

libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
```diff
@@ -58,6 +67,14 @@ export class StakeholderService {
 
       // Update the org
       const movieIds = org.data().movieIds || [];
+
+      // DEMO: Prevent double add
+      // TODO: allow to add the org multiple time with different orgs.
+      // BEWARE: update the delete method accordingly when you do.
+      if (movieIds.indexOf(movieId) >= 0) {
```
GrandSchtroumpf: Use `moviesIds.includes(movieId)` if possible instead of `indexOf`. It's more readible for junior devs.

libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
```diff
@@ -66,43 +83,60 @@ export class StakeholderService {
 
       return Promise.all([p1, p2]);
     })
-    .then(() => console.log('Transaction successfully committed!'))
-    .catch((error) => console.log('Transaction failed: ', error));
+      .then(() => console.log('Transaction successfully committed!'))
+      .catch((error) => console.log('Transaction failed: ', error));
 
+    // TODO: make this function truly async and wait for the tx to complete before returning the id.
```
GrandSchtroumpf: Adding `await` before `this.firestore.firestore.runTransaction(async (tx) => {...})` shoud do the trick no ?

libs/movie/src/lib/stakeholder/list/list.component.html
```diff
@@ -8,19 +8,17 @@ <h1>Add a new Organization</h1>
         <input type="text" matInput placeholder="Company Name" formControlName="org" [matAutocomplete]="autoName"/>
       </mat-form-field>
       <mat-autocomplete #autoName="matAutocomplete" [displayWith]="displayFn">
-        <mat-option *ngFor="let option of orgOptions" [value]="option" (onSelectionChange)="submit(option)">
+        <mat-option *ngFor="let option of orgOptions" [value]="option" (click)="submit(option)">
```
GrandSchtroumpf: You can also use `(change)` instead of `(click)` if you prefer

libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
lsenta: Yea I left this as a todo since this is outside the scope of the bugix. The way we handle async code is broken (we don't await them nor send back feedback to the user)

libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
lsenta: other than `any` breaks, will leave as is for now.

libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
lsenta: I cleaned some of the code and left todos for parts that requires more attention.

libs/movie/src/lib/stakeholder/list/list.component.html
lsenta: will leave as is for now, if it ain't broken...

libs/movie/src/lib/stakeholder/+state/stakeholder.service.ts
GrandSchtroumpf: It's still not working event if getStakeholderList returns a `Query<Stakeholder[]>` instead of a `Query<Stakeholder>` ? o.O

PR #290, opened 2019-05-02
Demos todos

ioaNikas:
- [x] Added a "Select all materials" button and functions into delivery view
- [x] Laboratory can now validate materials and define "approved" property of a material
- [x] Fixed a bug into firebase function "onDeliveryUpdateEvent"

### DIFF
--- a/functions/src/delivery.ts
+++ b/functions/src/delivery.ts
@@ -3,7 +3,7 @@ import { triggerNotifications, prepareNotification } from './notify';
 import { getOrgsOfDelivery, Organization } from './stakeholder';
 
 // This string refers to svg icon name
-export const APP_DELIVERY = 'delivery';
+export const APP_DELIVERY_ICON = 'delivery';
 
 export async function getCollection(path: string) {
   return db
@@ -31,7 +31,7 @@ async function notifyOnNewSignee(delivery: any, orgs: Organization[]): Promise<v
     .reduce((ids: string[], { userIds }) => [...ids, ...userIds], [])
     .map((userId: string) =>
       prepareNotification({
-        app: APP_DELIVERY,
+        app: APP_DELIVERY_ICON,
         message: `Stakeholder ${newStakeholder!.id}
         from ${newStakeholderOrg!.name}
         signed delivery ${delivery.id}`,
@@ -77,14 +77,16 @@ export const onDeliveryUpdate = async (
    *  Note: It doesn't trigger if this is the last signature, as another notification will be sent to notify
    *  that all stakeholders approved the delivery.
    */
-  if (delivery.validated.length === deliveryBefore.validated.length + 1 && delivery.validated.length !== stakeholderCount) {
+
+  const isNewSignature = delivery.validated.length === deliveryBefore.validated.length;
+  const isFullSignatures = delivery.validated.length === stakeholderCount;
+  const isBeforeStateClean = deliveryBefore.validated.length === stakeholderCount - 1;
+
+  if (isNewSignature && !isFullSignatures) {
     await notifyOnNewSignee(delivery, orgs);
   }
 
-  if (
-    deliveryBefore.validated.length !== stakeholderCount - 1 ||
-    delivery.validated.length !== stakeholderCount
-  ) {
+  if (!isBeforeStateClean || !isFullSignatures) {
     return true;
   }
 
@@ -124,7 +126,7 @@ export const onDeliveryUpdate = async (
       .reduce((ids: string[], { userIds }) => [...ids, ...userIds], [])
       .map((userId: string) =>
         prepareNotification({
-          app: APP_DELIVERY,
+          app: APP_DELIVERY_ICON,
           message: `Delivery with id ${delivery.id} has been approved by all stakeholders.`,
           userId,
           path: `/layout/${delivery.movieId}/view/${delivery.id}`
--- a/functions/src/delivery.ts
+++ b/functions/src/delivery.ts
@@ -2,7 +2,8 @@ import { db, functions } from './firebase';
 import { triggerNotifications, prepareNotification } from './notify';
 import { getOrgsOfDelivery, Organization } from './stakeholder';
 
-export const APP_DELIVERY = 'delivery'; /** This string refers to svg icon name */
+// This string refers to svg icon name
+export const APP_DELIVERY = 'delivery';
 
 export async function getCollection(path: string) {
   return db
@@ -67,17 +68,19 @@ export const onDeliveryUpdate = async (
   const processedId = deliveryDoc.data()!.processedId;
 
   const orgs = await getOrgsOfDelivery(delivery.id);
-
-  /** Notifications are triggered when a new id is pushed into delivery.validated, which is considered as a signature */
-  if (delivery.validated.length === deliveryBefore.validated.length + 1) {
-    return notifyOnNewSignee(delivery, orgs);
-  }
-
   const stakeholderCount = await db
     .collection(`deliveries/${delivery.id}/stakeholders`)
     .get()
     .then(snap => snap.size);
-  /** When validated.length reaches stakeholderCount.length */
+
+  /** Notifications are triggered when a new id is pushed into delivery.validated, which is considered as a signature
+   *  Note: It doesn't trigger if this is the last signature, as another notification will be sent to notify
+   *  that all stakeholders approved the delivery.
+   */
+  if (delivery.validated.length === deliveryBefore.validated.length + 1 && delivery.validated.length !== stakeholderCount) {
+    await notifyOnNewSignee(delivery, orgs);
+  }
+
   if (
     deliveryBefore.validated.length !== stakeholderCount - 1 ||
     delivery.validated.length !== stakeholderCount
@@ -89,6 +92,7 @@ export const onDeliveryUpdate = async (
     return true;
   }
 
+  // When validated.length reaches stakeholderCount.length
   try {
     await db.doc(`deliveries/${delivery.id}`).update({ processedId: context.eventId });
     const [materialsMovie, materialsDelivery] = await Promise.all([
@@ -114,7 +118,7 @@ export const onDeliveryUpdate = async (
       return db.doc(`movies/${delivery.movieId}/materials/${materialDelivery.id}`).set(material);
     });
 
-    /** When delivery is signed, we create notifications for each stakeholder */
+    // When delivery is signed, we create notifications for each stakeholder
     const notifications = orgs
       .filter(org => !!org && !!org.userIds)
       .reduce((ids: string[], { userIds }) => [...ids, ...userIds], [])
--- a/libs/material/src/lib/delivery/+state/delivery.query.ts
+++ b/libs/material/src/lib/delivery/+state/delivery.query.ts
@@ -8,8 +8,9 @@ import { MaterialStore } from '../../material/+state/material.store';
 import { Material } from '../../material/+state/material.model';
 import { switchMap, map, tap, filter } from 'rxjs/operators';
 import { materialsByCategory } from '../../material/+state/material.query';
-import { combineLatest } from 'rxjs';
+import { combineLatest, Observable } from 'rxjs';
 import { OrganizationQuery } from '@blockframes/organization';
+import { TemplateView } from '../../template/+state';
 
 @Injectable({
   providedIn: 'root'
@@ -31,7 +32,7 @@ export class DeliveryQuery extends QueryEntity<DeliveryState, Delivery> {
     private materialStore: MaterialStore,
     private stakeholderQuery: StakeholderQuery,
     private organizationQuery: OrganizationQuery,
-    private db: AngularFirestore,
+    private db: AngularFirestore
   ) {
     super(store);
   }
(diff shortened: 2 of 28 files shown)


functions/src/delivery.ts
```diff
@@ -67,17 +68,19 @@ export const onDeliveryUpdate = async (
...
     .collection(`deliveries/${delivery.id}/stakeholders`)
     .get()
     .then(snap => snap.size);
-  /** When validated.length reaches stakeholderCount.length */
+
+  /** Notifications are triggered when a new id is pushed into delivery.validated, which is considered as a signature
+   *  Note: It doesn't trigger if this is the last signature, as another notification will be sent to notify
+   *  that all stakeholders approved the delivery.
+   */
+  if (delivery.validated.length === deliveryBefore.validated.length + 1 && delivery.validated.length !== stakeholderCount) {
```
GrandSchtroumpf: Give intermediary names to your boolean values instead on chaining them : 
```typescript
const is**1 = delivery.validated.length === deliveryBefore.validated.length + 1;
const is**2 = delivery.validated.length !== stakeholderCount
if (is**1 && is**2) {...}
```

libs/material/src/lib/delivery/delivery-view-item/delivery-view-item.component.html
```diff
@@ -52,9 +52,13 @@ <h5>Step :</h5>
       <h5>State :</h5>
       <mat-icon class="icons" matTooltip="{{ material.state.toUpperCase() }}" svgIcon="{{ material.state }}"></mat-icon>
     </span>
-    <span *ngIf="!disableMaterial" class="time">
-      <h5>Delivery time :</h5>
-      <p class="time-text">{{ randomNumberPicker(30) }} days</p>
+    <span *ngIf="!disableMaterial && material.state === 'accepted'" class="time">
```
GrandSchtroumpf: Maybe you can have an upper div element that check the `disableMaterial`.
does `state` have other value that accepted and refused ? If not prefer a `isAccepted` boolean value.

libs/material/src/lib/delivery/delivery-view-item/delivery-view-item.component.ts
```diff
@@ -52,4 +51,8 @@ export class DeliveryViewItemComponent implements OnInit {
       ? (this.disableButtonLabel = 'ENABLE')
       : (this.disableButtonLabel = 'DISABLE');
   }
+
+  ngOnChanges(){
```
GrandSchtroumpf: Use a setter in the Input instead. `ngOnChanges` will be trigger for any kind of change. 
But in this case you don't need a `isChecked` parameter, you already have your input availble. Then in the `selectMaterial(isChecked)` pass the value you want.

libs/material/src/lib/delivery/delivery-view/delivery-view.component.scss
```diff
@@ -27,6 +27,22 @@ mat-progress-spinner {
...
+  font-weight: bold;
+  color: var(--disabled);
+}
+
+.select-button {
+  font-weight: bolder;
+  font-size: 13px;
+  height: 45px;
+  color: var(--accent);
+  border: solid 0.09rem var(--accent);
```
GrandSchtroumpf: Use px instead on 0.09rem.

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -11,44 +11,73 @@ import { Router } from '@angular/router';
...
 
   public selectMaterial(isChecked: any, id: string) {
-    isChecked
-      ? this.materialStore.addActive(id)
-      : this.materialStore.removeActive(id)
-    console.log(this.materialQuery.getActive())
+    isChecked ? this.materialStore.addActive(id) : this.materialStore.removeActive(id);
+  }
+
+  public log(bool) {
```
GrandSchtroumpf: Remove it.

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -11,44 +11,73 @@ import { Router } from '@angular/router';
...
 
   ngOnInit() {
+    this.delivery = this.query.getActive();
     this.movie = this.movieQuery.getActive();
     this.materials$ = this.query.materialsByActiveDelivery$;
     this.progressionValue$ = this.query.deliveryProgression$;
+    this.allChecked = false;
   }
 
   public selectMaterial(isChecked: any, id: string) {
```
GrandSchtroumpf: isChecked should be a boolean type

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -11,44 +11,73 @@ import { Router } from '@angular/router';
...
+    isChecked ? this.materialStore.addActive(id) : this.materialStore.removeActive(id);
+  }
+
+  public log(bool) {
+    console.log(bool)
+  }
+
+  public selectAllMaterials() {
+    this.allChecked = !this.allChecked;
+    if (this.allChecked === true) {
```
GrandSchtroumpf: prefer `if (this.allChecked) {....} else { ... }`

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -11,44 +11,73 @@ import { Router } from '@angular/router';
...
+      this.buttonLabel = 'Unselect materials';
+      this.materials$.subscribe(categories =>
```
GrandSchtroumpf: where are you unsubscribing ?

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -11,44 +11,73 @@ import { Router } from '@angular/router';
...
+      this.buttonLabel = 'Select all materials';
+      this.materials$.subscribe(categories =>
+        Object.keys(categories).forEach(categoryKey =>
+          categories[categoryKey].map(material => this.materialStore.removeActive(material.id))
+        )
+      );
+    }
   }
 
   public changeState(state: string) {
```
GrandSchtroumpf: What can be the values of `state` ? use litterals types :  `public changeState(state: 'state1' | 'state2' | 'state3')`

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -11,44 +11,73 @@ import { Router } from '@angular/router';
...
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class DeliveryViewComponent implements OnInit {
+export class DeliveryViewComponent implements OnInit, OnDestroy {
   public movie: Movie;
+  public delivery: Delivery;
   public materials$: Observable<Object>;
   public progressionValue$: Observable<number>;
+  public allChecked: boolean;
+  public buttonLabel = 'Select all materials';
```
GrandSchtroumpf: You should not have a `buttonLabel` property, this should be manage by the HTML with `*ngIfElse`

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
GrandSchtroumpf: Anyway, I don't understand why you need a subscription here. It feels like you want to use a snapshot instead.

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -11,44 +11,73 @@ import { Router } from '@angular/router';
...
-  public randomNumberPicker(scale: number) {
-    return Math.floor(Math.random() * scale) + 1;
+  public editDelivery() {
+    this.router.navigate([`layout/${this.movie.id}/form/${this.query.getActiveId()}`]);
   }
 
-  public editDelivery() {
-    this.router.navigate([`layout/${this.movie.id}/form/${this.query.getActiveId()}`])
+  ngOnDestroy() {
+    this.isAlive = false;
```
GrandSchtroumpf: You're not using this value.

libs/material/src/lib/delivery/movie-material-item/movie-material-item.component.ts
```diff
@@ -33,9 +33,9 @@ export class MovieMaterialItemComponent implements OnInit {
     ).name;
   }
 
-  public deliveredToggle(material: Material) {
+  public approvedToggle(material: Material) {
     this.service
-      .deliveredToggle(material, this.movieQuery.getActiveId())
+      .approvedToggle(material, this.movieQuery.getActiveId())
       .catch(err => console.log(err));
     //TODO: Ask why we put .catch ? And why we put it without "try" before
```
GrandSchtroumpf: So there are two way of handling errors : 
1 - Synchronously with `try { ... } catch (err) { ... }`
2 - Asynchronously with Promise : `promise.then(...).catch(...)`
In this case `approvedToggle` seems to return a promise, so you can chain it with `.then()` and `catch()`. As you don't want to do anything after it, you don't need the `then()` and go directly for the `catch()`

libs/ui/src/lib/notification/notification-list/notification-list.component.html
```diff
@@ -2,11 +2,15 @@
   *ngFor="
     let notification of (notifications$ | async | sort: 'date');
     let isLast = last;
-    let i = index">
+    let i = index
+  "
+>
```
GrandSchtroumpf: Reformat like it was before.

libs/material/src/lib/delivery/delivery-view-item/delivery-view-item.component.html
ioaNikas: As ```state``` and ```payment``` still appears when the material is disabled, I can't encapsulate all field into an upper div with only one boolean check. 

Property ```state``` got more than 2 potential values (```pending | available | delivered | accepted | refused```), I can't use a boolean here.

GrandSchtroumpf: Well you can do : 
```html
....
  <h5>State :</h5>
  <mat-icon class="icons" matTooltip="{{ material.state.toUpperCase() }}" svgIcon="{{ material.state }}"></mat-icon>
</span>
<ng-container *ngIf="!disableMaterial">
  <span *ngIf="material.state === 'accepted'" class="time"> ... </span>
  <span *ngIf="material.state === 'refused'" class="time">...</span>
</ng-container>
<span>
  <h5>Payement: </h5>
...
If you ends up using more than 2 values think about *ngSwitch.

File: libs/material/src/lib/delivery/+state/delivery.query.ts [hunk 14]
```diff
@@ -50,21 +51,27 @@ export class DeliveryQuery extends QueryEntity<DeliveryState, Delivery> {
...
-      map(materials => materialsByCategory(materials))
+      tap(materials => this.materialStore.set(materials))
     );
   }
 
+  /** Returns the active delivery materials sorted by category */
+  public get materialsByCategoryForActiveDelivery() {
+    return this.materialsByActiveDelivery().pipe(
+      map(materials => materialsByCategory(materials))
+    ) as Observable<TemplateView>;
```
GrandSchtroumpf: I think the name is waaaaay too long. And type should be like that: 
```typescript
  public get currentTemplateView(): Observable<TemplateView> {
    return this.materialsByActiveDelivery().pipe(
      map(materials => materialsByCategory(materials))
    );
}
```

GrandSchtroumpf: Also this doesn't need to be a getter (because it's an observable so it always have the last value). Make it a property instead.

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -262,6 +262,26 @@ export class DeliveryService {
...
   }
 
+  public get materialList() {
+    return this.movieQuery.selectActiveId().pipe(
+      switchMap(id => this.fireQuery.fromQuery(this.getDeliveryListWithStakeholders(id))),
+      tap(deliveries => this.store.set(deliveries))
+    );
+  }
+
+  private getDeliveryMovieMaterials(movieId: string): Query<Delivery[]> {
```
GrandSchtroumpf: You don't need this method in the class (no `this` inside). Maybe you can put it as an external function above the service : 
```typescript
function getDeliveryMovieMaterials(movieId: string): Query<Delivery[]> {
  ....
}

@Injectable(...)
```

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -32,52 +32,31 @@ export class DeliveryViewComponent implements OnInit, OnDestroy {
...
-    }
-    if (this.allChecked === false) {
-      this.buttonLabel = 'Select all materials';
-      this.materials$.subscribe(categories =>
-        Object.keys(categories).forEach(categoryKey =>
-          categories[categoryKey].map(material => this.materialStore.removeActive(material.id))
-        )
-      );
-    }
+    const op = this.allChecked
```
GrandSchtroumpf: Maybe give a better name ^^

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -32,52 +32,31 @@ export class DeliveryViewComponent implements OnInit, OnDestroy {
...
+      ? material => this.materialStore.addActive(material.id)
+      : material => this.materialStore.removeActive(material.id);
+    this.materialQuery.getAll().forEach(op)
```
GrandSchtroumpf: Do that inside a transaction (from Akita) : 
```typescript
applyTransaction(() => this.materialQuery.getAll().forEach(op))
```

libs/ui/src/lib/notification/notification-list/notification-list.component.html
```diff
@@ -1,16 +1,13 @@
-<div
-  *ngFor="
-    let notification of (notifications$ | async | sort: 'date');
-    let isLast = last;
-    let i = index
-  "
->
+<div *ngFor=
+    "let notification of (notifications$ | async | sort: 'date');
+     let isLast = last;
```
GrandSchtroumpf: You're not using `isLast` remove it.

libs/material/src/lib/delivery/+state/delivery.service.ts
ioaNikas: My bad, this method isn't used at all, I just forgot to delete it before pushing again.

PR #296, opened 2019-05-03
First version for movie-financing, explorer homepage

lsenta:
- Adds the homepage with a naive & hardcoded movie render components,
- Pretty full page scrolling (https://alvarotrigo.com/fullPage/),
- Split the navbar and header for reusability.


apps/movie-financing/movie-financing/src/app/explorer/home/home.component.scss
```diff
@@ -1,91 +1,26 @@
...
-.hero {
-  min-height: 55vh;
-  display: flex;
-  justify-content: flex-end;
-  //position: relative;
+financing-explorer-navbar {
+  position: fixed;
+  top: 0;
+  left: 0;
+  z-index: 10000;
```
GrandSchtroumpf: Use real value like "1" (or 2 if needed, but this should never happen)

apps/movie-financing/movie-financing/src/app/explorer/home/home.component.ts
```diff
@@ -7,7 +7,19 @@ import { ChangeDetectionStrategy, Component, OnInit } from '@angular/core';
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class FinancingExplorerHomeComponent implements OnInit {
+  config: any;
+  fullpage_api: any;
+
   constructor() {
+    this.config = {
```
GrandSchtroumpf: Put this into `ngOnInit`. It doesn't need to in this case, but at least you're sure not to make any mistake with the lifecycle.

apps/movie-financing/movie-financing/src/app/explorer/movie-home/movie-home.component.html
```diff
@@ -0,0 +1,40 @@
+<div class="movie-block">
```
GrandSchtroumpf: Use html semantic : main > section > article

apps/movie-financing/movie-financing/src/app/explorer/movie-home/movie-home.component.html
```diff
@@ -0,0 +1,40 @@
...
+    pulvinar blandit nulla sagittis mattis.
+  </p>
+
+  <div class="progress">
+    <mat-progress-bar mode="determinate" value="40"></mat-progress-bar>
+    <p>64% Completed</p>
+  </div>
+
+  <div class="cta">
+    <a [routerLink]="['/explorer/movie/42']">
```
GrandSchtroumpf: In this case you don't need `[]` because this is only a string. But if you intend to add dynamic data later inside just keep it like that.

apps/movie-financing/movie-financing/src/app/explorer/movie-home/movie-home.component.scss
```diff
@@ -0,0 +1,108 @@
...
+
+.key {
+  color: white;
+}
+
+.value {
+  color: white;
+}
+
+.left-decoration {
```
GrandSchtroumpf: If this is supposed to be changed maybe add a comment flag for that. It'll be easier later to update it.

apps/movie-financing/movie-financing/src/app/explorer/movie-home/movie-home.component.html
```diff
@@ -0,0 +1,40 @@
+<div class="movie-block">
+  <div class="status-block">
+    Almost Closed
+  </div>
+
+  <h3>Nom Réalisateur</h3>
+  <h1>Movie Name Sooooooooooo Long</h1>
+  <table class="data">
```
GrandSchtroumpf: don't use table if this is not meant to be a real table, use flexbox. Else use Angular material table.

apps/movie-financing/movie-financing/src/app/explorer/movie-home/movie-home.component.scss
```diff
@@ -0,0 +1,108 @@
...
+
+  a {
+    padding: 1rem;
+    display: inline-block;
+    width: 100%;
+    height: 100%;
+    color: white;
+    text-decoration: none;
+
+    &:after {
```
GrandSchtroumpf: Why not using icon ?

apps/movie-financing/movie-financing/src/app/explorer/movie-home/movie-home.component.html
lsenta: It's meant to be a table, it's data displayed and arranged in a tabular way:

Screenshot 2019-05-03 at 17 12 44

apps/movie-financing/movie-financing/src/app/explorer/movie-home/movie-home.component.scss
lsenta: It's straightforward and does 90% of the job until we have a ui kit that we can reuse.

apps/movie-financing/movie-financing/src/app/explorer/movie-home/movie-home.component.scss
lsenta: Should be as fixed as the rest of the design, won't change or go dynamic in the short term.

apps/movie-financing/movie-financing/src/app/explorer/movie-home/movie-home.component.html
lsenta: Yup, next step after designing the pages is to use dynamic content!

PR #299, opened 2019-05-06
Demos todos #2

ioaNikas:
- [x] Duplicated "Select All" checkboxes into delivery form (from delivery view) to delete or change step on multiple materials.
- [x] New stakeholder on delivery receive a custom notification to accept or decline to join the teamwork.
- [x] Deliveries now have state and payment properties to match the demonstration walkthrough.
- [x] Template picker now displays user's templates even if we didnt initilizate template-list component.
- [x] Refactored and renamed some stuff for a better readability of the code.

### DIFF
--- a/functions/src/stakeholder.ts
+++ b/functions/src/stakeholder.ts
@@ -52,29 +52,20 @@ export const onDeliveryStakeholderCreate = async (
         .filter(org => !!org && !!org.userIds)
         .reduce((ids: string[], { userIds }) => [...ids, ...userIds], [])
         .map((userId: string) => {
-          if (newStakeholderOrg.userIds.includes(userId) && stakeholderCount > 1) {
-            return prepareNotification({
-              app: APP_DELIVERY_ICON,
-              message: `You have been invited to delivery : ${
-                delivery.id
-              }. Do you wish to work on it ?`,
-              userId,
-              path: `/layout/${delivery.movieId}/form/${delivery.id}/teamwork`,
-              deliveryId: delivery.id,
-              stakeholderId: newStakeholder.id
-            });
-          } else {
-            return prepareNotification({
-              app: APP_DELIVERY_ICON,
-              message: `Stakeholder ${newStakeholder.id} from ${
-                newStakeholderOrg.name
-              } has been added to delivery ${delivery.id}`,
-              userId,
-              path: `/layout/${delivery.movieId}/form/${delivery.id}/teamwork`,
-              deliveryId: delivery.id,
-              stakeholderId: newStakeholder.id
-            });
-          }
+          const message =
+            newStakeholderOrg.userIds.includes(userId) && stakeholderCount > 1
+              ? `You have been invited to delivery : ${delivery.id}. Do you wish to work on it ?`
+              : `Stakeholder ${newStakeholder.id} from ${
+                  newStakeholderOrg.name
+                } has been added to delivery ${delivery.id}`;
+          return prepareNotification({
+            app: APP_DELIVERY_ICON,
+            message,
+            userId,
+            path: `/layout/${delivery.movieId}/form/${delivery.id}/teamwork`,
+            deliveryId: delivery.id,
+            stakeholderId: newStakeholder.id
+          });
         });
 
       await triggerNotifications(notifications);
--- a/functions/src/stakeholder.ts
+++ b/functions/src/stakeholder.ts
@@ -43,21 +43,39 @@ export const onDeliveryStakeholderCreate = async (
     }
 
     try {
+      const stakeholderCount = await db
+        .collection(`deliveries/${delivery.id}/stakeholders`)
+        .get()
+        .then(col => col.size);
       const orgs = await getOrgsOfDelivery(delivery.id);
       const notifications = orgs
         .filter(org => !!org && !!org.userIds)
         .reduce((ids: string[], { userIds }) => [...ids, ...userIds], [])
-        .map((userId: string) =>
-          prepareNotification({
-            app: APP_DELIVERY_ICON,
-            message: `Stakeholder ${newStakeholder.id} from ${
-              newStakeholderOrg.name
-            } has been added to delivery ${delivery.id}`,
-            userId,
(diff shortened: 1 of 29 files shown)


functions/src/stakeholder.ts
```diff
@@ -43,21 +43,39 @@ export const onDeliveryStakeholderCreate = async (
...
-            userId,
-            path: `/layout/${delivery.movieId}/form/${delivery.id}/teamwork`
-          })
-        );
-        // TODO: Make a custom notification for the stakeholder added to this delivery
+        .map((userId: string) => {
+          if (newStakeholderOrg.userIds.includes(userId) && stakeholderCount > 1) {
+            return prepareNotification({
+              app: APP_DELIVERY_ICON,
+              message: `You have been invited to delivery : ${
```
GrandSchtroumpf: If only the message change you can do : 
```typescript
const message = newStakeholderOrg.userIds.includes(userId) && stakeholderCount > 1
 ? ....
  : ...
return prepareNotification({ .... })
```

libs/material/src/lib/delivery/delivery-form/delivery-form.component.html
```diff
@@ -55,54 +51,101 @@
   <!-- Content -->
   <mat-sidenav-content class="delivery-form">
     <section>
-      <h3 *ngIf="(movie$ | async) as movie">{{ movie.title.original }}</h3>
+      <div fxLayout="row" fxLayoutAlign="space-between center">
```
GrandSchtroumpf: You can use an `<article>` instead of `div`

libs/material/src/lib/delivery/delivery-form/delivery-form.component.html
```diff
@@ -55,54 +51,101 @@
...
+            [matMenuTriggerFor]="actions"
+            [disabled]="isDeliveryValidated$ | async">
+            Actions <mat-icon class="action-icon">expand_more</mat-icon>
+          </button>
+          <mat-menu #actions="matMenu" xPosition="before">
+            <button mat-menu-item [matMenuTriggerFor]="steps">
+              <span>Change step</span>
+            </button>
+            <mat-menu #steps="matMenu" xPosition="after">
+              <button
```
GrandSchtroumpf: The best would be to have : 
```html
<ng-container *ngIf="...; else">

</ng-container>
<ng-template>

</ng-template>
```

libs/material/src/lib/delivery/delivery-form/delivery-form.component.html
```diff
@@ -55,54 +51,101 @@
...
+                  (click)="changeStep(step.id)"
+                  mat-menu-item>
+                  {{ step.name }}
+                </button>
+              </ng-template>
+            </mat-menu>
+            <mat-divider></mat-divider>
+            <button mat-menu-item>
+              <mat-icon color="warn">delete</mat-icon>
+              <span (click)="deleteSelectedMaterials()">Delete</span>
```
GrandSchtroumpf: click should be on button.

libs/material/src/lib/delivery/delivery-form/delivery-form.component.html
```diff
@@ -55,54 +51,101 @@
...
+          </mat-menu>
+        </span>
+      </div>
       <div *ngIf="(form$ | async) as form">
         <material-delivery-add-form
           [isDeliveryValidated]="isDeliveryValidated$ | async"
           *ngIf="form.category === ''"
-          (material)="addMaterial($event)"
-        ></material-delivery-add-form>
+          (material)="addMaterial($event)">
```
GrandSchtroumpf: `material` is not a good name for an event.

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -60,4 +77,33 @@ export class MaterialService {
...
+    const batch = this.db.firestore.batch();
+    materials.forEach(material => {
+      const materialRef = this.db.doc<Material>(
+        `deliveries/${this.deliveryQuery.getActiveId()}/materials/${material.id}`
+      ).ref;
+      return batch.delete(materialRef);
+    });
+    materials.forEach(material => {
+      const materialRef = this.db.doc<Material>(
+        `movies/${this.movieQuery.getActiveId()}/materials/${material.id}`
```
GrandSchtroumpf: make it more readible with `const movieId = this.movieQuery.getActiveId()`.

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -60,4 +77,33 @@ export class MaterialService {
...
+      return batch.delete(materialRef);
+    });
+    batch.commit();
+  }
+
+  public changeStep(materials: Material[], stepId: string) {
+    const batch = this.db.firestore.batch();
+    materials.forEach(material => {
+      const materialRef = this.db.doc<Material>(
+        `deliveries/${this.deliveryQuery.getActiveId()}/materials/${material.id}`
```
GrandSchtroumpf: same

libs/ui/src/lib/notification/notification-item/notification-item.component.ts
```diff
@@ -6,19 +6,27 @@ import { NotificationService } from '../+state';
...
+    private service: NotificationService,
+  ) {}
 
   public goToPath() {
     this.router.navigate([this.notification.path]);
     this.service.readNotification(this.notification.id);
   }
 
+  public goToTeamwork() {
+    this.service.joinTeamwork(this.notification.stakeholderId, this.notification.deliveryId);
```
GrandSchtroumpf: It may need a try / catch here in case of errors.

libs/material/src/lib/delivery/delivery-form/delivery-form.component.html
ioaNikas: As we have an "update" event, I'll rename it to "add".

GrandSchtroumpf: For me "add" / "update" are action. "updated", "added" are events.

Issue #266, opened 2019-04-29
Handle signin without encrypted key in the localStorage


pldespaigne:
Currently signin require to have a keystore in the localStorage, if the locaStorage has been cleaned, the signin will crash

pldespaigne:
fixed by #301

PR #301, opened 2019-05-06
refactoring and bugfixe

pldespaigne:
Clean sigin/signup
- [x] auth service refactoring
- [x] relayer wallet refactoring
---
- [x] slight modification of account-view component to display/refresh balance

### DIFF
--- a/libs/account/src/lib/account-view/account-view.component.html
+++ b/libs/account/src/lib/account-view/account-view.component.html
@@ -66,22 +66,34 @@
         </mat-expansion-panel-header>
         <mat-list>
           <mat-list-item>
-            <mat-card-header>
-              <eth-blockie mat-card-avatar [round]="false" [height]="40" width="40" [address]="user.identity.address"></eth-blockie>
-              <mat-card-title>{{user.identity.domain}}</mat-card-title>
-              <mat-card-subtitle>
-                <a *ngIf="network === 'homestead'; else testNet" [href]="'https://etherscan.io/address/'+user.identity.address" target="_blank" rel="noopener">{{user.identity.address}}</a>
-                <ng-template #testNet>
-                  <a [href]="'https://'+network+'.etherscan.io/address/'+user.identity.address" target="_blank" rel="noopener">{{user.identity.address}}</a>
+            <ng-container *ngIf="user.identity; else walletLoading">
+              <mat-card-header>
+                <eth-blockie mat-card-avatar [round]="false" [height]="40" width="40" [address]="user.identity.address"></eth-blockie>
+                <mat-card-title>{{user.identity.domain}}</mat-card-title>
+                <mat-card-subtitle>
+                  <a *ngIf="network === 'homestead'; else testNet" [href]="'https://etherscan.io/address/'+user.identity.address" target="_blank" rel="noopener">{{user.identity.address}}</a>
+                  <ng-template #testNet>
+                    <a [href]="'https://'+network+'.etherscan.io/address/'+user.identity.address" target="_blank" rel="noopener">{{user.identity.address}}</a>
+                  </ng-template>
+                </mat-card-subtitle>
+              </mat-card-header>
+              <mat-card-content >
+                <button *ngIf="(notLoading$ | async); else isLoading" mat-button matTooltip="Click to refresh" (click)="refreshBalance()">
+                  {{user.balance || '0.0'}} ETH
+                </button>
+                <ng-template #isLoading>
+                  <button mat-button [disabled]="true">
+                    <mat-spinner [diameter]="24" [strokeWidth]="2"></mat-spinner>
+                  </button>
                 </ng-template>
-              </mat-card-subtitle>
-            </mat-card-header>
-            <mat-card-content >
-              {{user.balance || '0.0'}} ETH
-            </mat-card-content>
-          </mat-list-item>
-          <mat-list-item>
-            <mat-card-header></mat-card-header>
+              </mat-card-content>
+            </ng-container>
+            <ng-template #walletLoading>
+              <div class="wallet-loading">
+                <p>Your Wallet is loading...</p>
+                <mat-progress-bar mode="indeterminate"></mat-progress-bar>
+              </div>
+            </ng-template>
           </mat-list-item>
         </mat-list>
       </mat-expansion-panel>
--- a/libs/account/src/lib/account-view/account-view.component.html
+++ b/libs/account/src/lib/account-view/account-view.component.html
@@ -78,9 +78,11 @@
                 </mat-card-subtitle>
               </mat-card-header>
               <mat-card-content >
-                <button *ngIf="(notLoading$ | async); else isLoading" mat-button matTooltip="Click to refresh" (click)="refreshBalance()">
-                  {{user.balance || '0.0'}} ETH
-                </button>
+                <ng-container *ngIf="(notLoading$ | async); else isLoading">
+                  <button mat-button matTooltip="Click to refresh" (click)="refreshBalance()">
+                    {{user.balance || '0.0'}} ETH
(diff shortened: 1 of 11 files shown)


pldespaigne:
this PR also fix #266

libs/account/src/lib/account-view/account-view.component.html
```diff
@@ -66,22 +66,34 @@
...
-                <ng-template #testNet>
-                  <a [href]="'https://'+network+'.etherscan.io/address/'+user.identity.address" target="_blank" rel="noopener">{{user.identity.address}}</a>
+            <ng-container *ngIf="user.identity; else walletLoading">
+              <mat-card-header>
+                <eth-blockie mat-card-avatar [round]="false" [height]="40" width="40" [address]="user.identity.address"></eth-blockie>
+                <mat-card-title>{{user.identity.domain}}</mat-card-title>
+                <mat-card-subtitle>
+                  <a *ngIf="network === 'homestead'; else testNet" [href]="'https://etherscan.io/address/'+user.identity.address" target="_blank" rel="noopener">{{user.identity.address}}</a>
+                  <ng-template #testNet>
+                    <a [href]="'https://'+network+'.etherscan.io/address/'+user.identity.address" target="_blank" rel="noopener">{{user.identity.address}}</a>
```
GrandSchtroumpf: You should have a getter on the component side instead of a ngIfElse : 
```typescript
get etherscanAddress() {
  return this.network === 'homestead'
    ? `https://etherscan.io/address/${this.user.identity.address}`
    :  `https://${network}.etherscan.io/address/${this.user.identity.address}`
}
```

libs/account/src/lib/account-view/account-view.component.html
```diff
@@ -66,22 +66,34 @@
...
+              </mat-card-header>
+              <mat-card-content >
+                <button *ngIf="(notLoading$ | async); else isLoading" mat-button matTooltip="Click to refresh" (click)="refreshBalance()">
+                  {{user.balance || '0.0'}} ETH
+                </button>
+                <ng-template #isLoading>
+                  <button mat-button [disabled]="true">
+                    <mat-spinner [diameter]="24" [strokeWidth]="2"></mat-spinner>
+                  </button>
                 </ng-template>
```
GrandSchtroumpf: For consistency I'd recommand doing
```typescript
<ng-container *ngIf="...; else **">
  ...
</ng-container>
<ng-template #**>

</ng-template>
```

libs/account/src/lib/account-view/account-view.component.ts
```diff
@@ -1,34 +1,30 @@
...
-      map(user => makeBlockie.default(user.identity.address))
-    );
     this.network = network ;
   }
 
-  ngOnDestroy() {
+  refreshBalance() {
+    this.notLoading$.next(false);
+    this.authService.refreshBalance();
+    setTimeout(() => this.notLoading$.next(true), 600);
```
GrandSchtroumpf: Why are you using a setTimeout ? Is it not better to `await this.authService.refreshBalance()` ?

libs/account/src/lib/account-view/account-view.component.ts
```diff
@@ -1,34 +1,30 @@
...
   styleUrls: ['./account-view.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class AccountViewComponent implements OnInit, OnDestroy {
+export class AccountViewComponent implements OnInit {
   public user$: Observable<User>;
-  public asciiUsername$: Observable<string>;
-  public blockie$: Observable<string>;
   public network: string;
+  public notLoading$ = new BehaviorSubject<boolean>(true);
```
GrandSchtroumpf: So basically the BehaviorSubject work like that : 
```typescript
private _notLoading$ = new BehaviorSubject<boolean>(true);
public notLoading$ = this._notLoading$.asObservable();
```
But this shuld definitely be managed by Akita in the account store.

libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -21,35 +22,43 @@ export class AuthService {
...
-    this.store.update({isEncrypting: true});
-    const userCredentials = await this.afAuth.auth.createUserWithEmailAndPassword(mail, password);
-    await this.create(userCredentials.user);
+  public async signin(email: string, password: string) {
+    await this.afAuth.auth.signInWithEmailAndPassword(email, password);
     this.subscribeOnUser();
-    const username = mail.split('@')[0];
+    this.initWallet(email);
+    
+    this.wallet.loadKey('web', password);  // no await -> do the job in background
```
GrandSchtroumpf: Would it make sense to use a webworker here ?

libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -21,35 +22,43 @@ export class AuthService {
...
-    this.wallet.signup(userCredentials.user.uid, this._sanitizeUSername(username), password).then(async () => {  // no await -> do the job in background
-      const balance = await this.wallet.getBalance();
-      this.store.updateUser({balance});
+
+    const userCredentials = await this.afAuth.auth.createUserWithEmailAndPassword(email, password);
+    await this.create(userCredentials.user);
+    
+    this.wallet.createLocalKey('web', password, email)  // no await -> do the job in background
+    .then(() => this.wallet.createERC1077(userCredentials.user.uid))
+    .then((_) => {
```
GrandSchtroumpf: Don't put an _ in the (). It's either `(_ => ...)` or `(() => )`

GrandSchtroumpf: Also same question. Could we use webworker ?

libs/account/src/lib/account-view/account-view.component.html
pldespaigne: Did this at first, but the linter wouldn't let me because since network is "hardcoded" in env file the linter say that condition is always true and doesn't compile

libs/account/src/lib/account-view/account-view.component.html
pldespaigne: ✔️

libs/auth/src/lib/+state/auth.service.ts
pldespaigne: ✔️

libs/auth/src/lib/+state/auth.service.ts
pldespaigne: Webworker will add a layer of complexity whereas async/await already handle the encrypting/decrypting. This can maybe reduce the encrypting/decrypting time but this is not really a big issue for now. Webworker could be a good optimization in the future, but I would say Keep It Stupid Simple. see #305

libs/account/src/lib/account-view/account-view.component.ts
pldespaigne: ✔️

libs/account/src/lib/account-view/account-view.component.ts
pldespaigne: Did you mean that I need to add a `balanceLoading` flag to the user store ?
Because if I use the global akita loading flag I fear that there will be side effects.

GrandSchtroumpf: Yes `balanceLoading` is ok.

pldespaigne: ✔️

libs/account/src/lib/account-view/account-view.component.html
GrandSchtroumpf: try `return this.network === 'homestead' as string`

pldespaigne: ✔️

PR #307, opened 2019-05-07
Movie materials as delivery

ioaNikas:
- [x] User can now create a delivery using already existing movie materials as a template
- [x] Materials with no step property are now rendered correctly

### DIFF
--- a/functions/src/notify.ts
+++ b/functions/src/notify.ts
@@ -33,7 +33,6 @@ export async function triggerNotifications(notifications: Notification[]): Promi
 
 export function prepareNotification(notif: PartialNotification): Notification {
   return {
-   ...notif,
     id: db.collection('notifications').doc().id,
     isRead: false,
     date: serverTimestamp(),
--- a/functions/src/notify.ts
+++ b/functions/src/notify.ts
@@ -33,6 +33,7 @@ export async function triggerNotifications(notifications: Notification[]): Promi
 
 export function prepareNotification(notif: PartialNotification): Notification {
   return {
+   ...notif,
     id: db.collection('notifications').doc().id,
     isRead: false,
     date: serverTimestamp(),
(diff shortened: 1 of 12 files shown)


functions/src/notify.ts
```diff
@@ -26,15 +26,10 @@ export async function triggerNotifications(notifications: Notification[]): Promi
   return notificationBatch.commit();
 }
 
-export function prepareNotification({ app, message, userId, path, deliveryId, stakeholderId }: PartialNotification): Notification {
+export function prepareNotification(notif: PartialNotification): Notification {
   return {
+   ...notif,
```
GrandSchtroumpf: This should be in the end for default not to override the input.

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -120,6 +122,43 @@ export class DeliveryService {
...
+      ['canValidateDelivery'],
+      true
+    );
+
+    this.db.doc<Delivery>(`deliveries/${id}`).set(delivery);
+    this.db
+      .doc<Stakeholder>(`deliveries/${id}/stakeholders/${deliveryStakeholder.id}`)
+      .set(deliveryStakeholder);
+    this.store.setActive(id);
+    this.router.navigate([`layout/${this.movieQuery.getActiveId()}/form/${id}/settings`]);
```
GrandSchtroumpf: this should be done by the component when promise has resolved.

libs/material/src/lib/delivery/delivery-view/delivery-view.component.ts
```diff
@@ -1,36 +1,43 @@
...
     private movieQuery: MovieQuery,
     private service: DeliveryService,
+    private materialService: MaterialService,
     private materialStore: MaterialStore,
     private materialQuery: MaterialQuery,
     private router: Router
   ) {}
 
   ngOnInit() {
+    this.materialService
```
GrandSchtroumpf: Add todo : remove after cannes in favor of routes.

PR #312, opened 2019-05-08
Financing tweaks

lsenta:
- [x] fix typo
- [x] fix header
- [x] fix navbar font and add profile summary in navbar


apps/movie-financing/movie-financing/src/app/explorer/navbar/navbar.component.scss
```diff
@@ -1,4 +1,4 @@
-@import url('https://fonts.googleapis.com/css?family=Roboto');
+@import url('https://fonts.googleapis.com/css?family=Montserrat:300,500,800');
```
GrandSchtroumpf: Why do you need to import it here? It's not in the index.html ? Or the global the style.scss ?

lsenta: True, we could import the whole montserrat family in the index.html, will patch!

PR #313, opened 2019-05-08
Stakeholder movie invitation

ioaNikas:
- [x] Added a custom notification for new stakeholders invited on a movie.
- [x] Added a download link for materials with "available" state in delivery view item.
- [x] Added a download link for materials with "approved = true" in movie material item.

### DIFF
--- a/functions/src/delivery.ts
+++ b/functions/src/delivery.ts
@@ -37,7 +37,7 @@ async function notifyOnNewSignee(delivery: any, orgs: Organization[]): Promise<v
         signed delivery ${delivery.id}`,
         userId,
         path: `/layout/${delivery.movieId}/form/${delivery.id}`,
-        deliveryId: delivery.id
+        docId: delivery.id
       })
     );
 
@@ -131,7 +131,7 @@ export const onDeliveryUpdate = async (
           message: `Delivery with id ${delivery.id} has been approved by all stakeholders.`,
           userId,
           path: `/layout/${delivery.movieId}/view/${delivery.id}`,
-          deliveryId: delivery.id
+          docId: delivery.id
         })
       );
 
--- a/functions/src/delivery.ts
+++ b/functions/src/delivery.ts
@@ -37,7 +37,7 @@ async function notifyOnNewSignee(delivery: any, orgs: Organization[]): Promise<v
         signed delivery ${delivery.id}`,
         userId,
         path: `/layout/${delivery.movieId}/form/${delivery.id}`,
-        doc: {id: delivery.id, type: 'delivery'}
+        docID: {id: delivery.id, type: 'delivery'}
       })
     );
 
@@ -131,7 +131,7 @@ export const onDeliveryUpdate = async (
           message: `Delivery with id ${delivery.id} has been approved by all stakeholders.`,
           userId,
           path: `/layout/${delivery.movieId}/view/${delivery.id}`,
-          doc: {id: delivery.id, type: 'delivery'}
+          docID: {id: delivery.id, type: 'delivery'}
         })
       );
 
--- a/functions/src/delivery.ts
+++ b/functions/src/delivery.ts
@@ -37,7 +37,7 @@ async function notifyOnNewSignee(delivery: any, orgs: Organization[]): Promise<v
         signed delivery ${delivery.id}`,
         userId,
         path: `/layout/${delivery.movieId}/form/${delivery.id}`,
-        docId: delivery.id
+        doc: {id: delivery.id, type: 'delivery'}
       })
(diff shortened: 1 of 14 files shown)


functions/src/stakeholder.ts
```diff
@@ -150,19 +150,30 @@ export const onMovieStakeholderCreate = async (
...
+        .get()
+        .then(col => col.size);
       const orgs = await getOrgsOfMovie(movie.id);
       const notifications = orgs
         .filter(org => !!org && !!org.userIds)
         .reduce((ids: string[], { userIds }) => [...ids, ...userIds], [])
-        .map((userId: string) =>
-          prepareNotification({
+        .map((userId: string) => {
+          const message =
```
GrandSchtroumpf: Create an function outside of the map to do that. It's difficult to read. Especially with prettier formatting

libs/material/src/lib/delivery/movie-material-item/movie-material-item.component.scss
```diff
@@ -48,3 +48,14 @@ p {
   color: var(--primary);
   font-weight: bolder;
 }
+
+.labo-approval {
+  position: relative;
```
GrandSchtroumpf: Why is it relative ? Do you have something with position absolute inside ?

libs/material/src/lib/delivery/movie-material-item/movie-material-item.component.scss
```diff
@@ -48,3 +48,14 @@ p {
...
+  bottom: 70px;
+  left: 950px;
+}
+
+.done-icon {
+  position: relative;
```
GrandSchtroumpf: Same here ?

functions/src/delivery.ts
```diff
@@ -37,7 +37,7 @@ async function notifyOnNewSignee(delivery: any, orgs: Organization[]): Promise<v
         signed delivery ${delivery.id}`,
         userId,
         path: `/layout/${delivery.movieId}/form/${delivery.id}`,
-        deliveryId: delivery.id
+        docId: delivery.id
```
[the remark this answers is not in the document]
ioaNikas: As notifications often refers to a delivery or a movie, I use a single variable for both (this variable is used to access the document and do some operations like updating the stakeholder status on a delivery or a movie)

functions/src/notify.ts
```diff
@@ -5,7 +5,7 @@ interface PartialNotification {
   message: string;
   userId: string;
   path?: string;
-  docId?: string;
+  doc?: { id: string, type: 'movie' | 'delivery' } ,
```
GrandSchtroumpf: Create a type for that : 
```typescript
interface DocId {
  id: string;
  type: 'movie' | 'delivery'
}
```

functions/src/stakeholder.ts
```diff
@@ -22,10 +29,20 @@ export async function getOrgsOfMovie(movieId: string): Promise<Organization[]> {
   return orgs.map(doc => doc.data() as Organization);
 }
 
-export const onDeliveryStakeholderCreate = async (
+function customMessage(userId: string, snapInformations: snapObject, deliveryId?: string, movieId?: string) {
+  const documentType = (!!deliveryId) ? 'delivery' : 'movie';
+  const documentId = (!!deliveryId) ? deliveryId : movieId;
```
GrandSchtroumpf: You'll have a problem here is movieId is not defined.

functions/src/stakeholder.ts
```diff
@@ -43,27 +60,26 @@ export const onDeliveryStakeholderCreate = async (
     }
 
     try {
-      const stakeholderCount = await db
+      const snapInformations = {
+        stakeholderId: newStakeholder.id,
+        orgName: newStakeholderOrg.name,
+        count: await db
```
GrandSchtroumpf: create a `const count = await db.collection(...).get().then(col => col.size)` above for better readibility.

functions/src/stakeholder.ts
```diff
@@ -150,27 +165,26 @@ export const onMovieStakeholderCreate = async (
...
```
GrandSchtroumpf: Same here. And create a common function :
```typescript
function getCount(collection: string) {
  return db.collection(collection).get().then(col => col.size)
}
```

functions/src/notify.ts
```diff
@@ -1,11 +1,16 @@
 import { db, serverTimestamp } from './firebase';
 
+export interface Doc {
```
GrandSchtroumpf: Call it DocID instead. Because it only about the ID

PR #324, opened 2019-05-10
Style: Styling delivery app with Mathilde

ioaNikas:
- [x] New style for the header
- [x] Prettier delivery list
- [x] Prettier delivery form
- [x] Defined custom css variables and font-size/font-weight into style.scss


libs/material/src/lib/delivery/stakeholder-item/stakeholder-item.component.html
```diff
@@ -1,14 +1,14 @@
...
   <mat-card-content>
-    <h2>Signee:</h2>
+    <h4>Signee</h4>
     <p>{{ stakeholder.organization.name }}</p>
   </mat-card-content>
-  <mat-card-footer *ngIf="!(hasStakeholderSigned$ | async)" fxLayout="row" fxLayoutAlign="space-evenly center" class="footerLocked">
-    <mat-icon>lock_open</mat-icon>
-    <span>The signature will appear here</span>
+  <mat-card-footer *ngIf="!(hasStakeholderSigned$ | async)" fxLayout="row" fxLayoutAlign="center center" class="footerLocked">
+      <span><mat-icon class="lock-icon">lock_open</mat-icon></span>
```
GrandSchtroumpf: You don't need a span around a mat-icon.

libs/material/src/lib/delivery/stakeholder-item/stakeholder-item.component.html
```diff
@@ -1,14 +1,14 @@
...
+    <span>Your signature will appear here</span>
+  </mat-card-footer>
+  <mat-card-footer *ngIf="hasStakeholderSigned$ | async" fxLayout="row" fxLayoutAlign="center center" class="footerSigned">
+    <span><mat-icon svgIcon="signed" class="sign-icon"></mat-icon></span>
```
GrandSchtroumpf: You don't need a span around the mat-icon.

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.html
```diff
@@ -1,30 +1,30 @@
 <mat-card class="container mat-elevation-z6">
-  <mat-card-title>Edit your material</mat-card-title>
+  <mat-card-title><h4>Edit your material</h4></mat-card-title>
```
GrandSchtroumpf: You should not have to add h4 for mat-card-title. Why do you do that ?

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.html
```diff
@@ -1,30 +1,30 @@
...
+      <mat-form-field appearance="outline" class="inputs">
         <mat-label>Category</mat-label>
         <input class="category" formControlName="category" matInput placeholder="Category" />
       </mat-form-field>
     </div>
 
-    <div class="step" fxLayout="column" fxLayoutAlign="space-around center">
-      <mat-form-field *ngIf="hasStep" appearance="outline">
+    <div class="step" fxLayout="column" fxLayoutAlign="space-between end">
+      <mat-form-field *ngIf="hasStep else noStep" appearance="outline" class="mat-select-step">
```
GrandSchtroumpf: It misses a ";" before the "else" I think.

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.scss
```diff
@@ -1,11 +1,26 @@
 mat-card {
-  margin-bottom: 4%;
+  margin-bottom: 20px;
+  padding: 30px 50px 50px 50px;
```
GrandSchtroumpf: Best would be to do : `padding: 50px; padding-top: 30px`. I find it more readible.

libs/material/src/lib/material/material-delivery-item/material-delivery-item.component.html
```diff
@@ -21,12 +21,19 @@
...
-      {{ material.step.name }}
-    </mat-card-title>
+  <mat-card-content
+    *ngIf="material.step"
+    class="content-step"
+    fxLayout="column"
+    fxLayoutAlign="start end"
+  >
+    <div>
+      <mat-card-subtitle class="step">
```
GrandSchtroumpf: I think this is a missuse of `mat-card-subtitle` and title. Just use normal html sementic here (you're inside the <mat-card-content>

libs/material/src/lib/material/material-delivery-item/material-delivery-item.component.scss
```diff
@@ -1,24 +1,41 @@
...
+.description-text {
+  font-size: 11px;
+  font-weight: 400;
+  font-style: italic;
+  margin-bottom: 0;
+}
+
+mat-card-actions {
+  visibility: hidden;
+  margin: 0 0 0 -6px;
```
GrandSchtroumpf: Divide in two : `margin: 0; margin-left: -6px`. And you're still using a LOT of negative values...

libs/material/src/lib/template/template-picker/template-picker.component.scss
```diff
@@ -36,19 +48,18 @@ mat-card {
...
 }
 
 p {
-  color: rgba(0, 0, 0, 0.514);
-}
-
-h2 {
-  font-weight: bold;
+  font-size: 14px;
+  margin-top: -18px;
```
GrandSchtroumpf: negative.

libs/movie/src/lib/movie/view/view.component.ts
```diff
@@ -33,8 +33,8 @@ export class ViewComponent implements OnInit, OnDestroy {
   public getStaticBySlug (scope: string, slug: string) {
     return getLabelBySlug (scope, slug) as string;
   }
-  
+
   ngOnDestroy() {
-    this.store.setActive(null);
+    // this.store.setActive(null);
```
GrandSchtroumpf: Remove it if you don't need it.

libs/ui/src/lib/toolbar/header/header.component.html
```diff
@@ -11,35 +9,53 @@
...
-    </a>
-  </mat-menu>
+  <form>
+    <mat-icon color="primary" class="search-icon">search</mat-icon>
+    <mat-form-field class="searchInput">
+      <input type="text" matInput placeholder="Search any film" />
+    </mat-form-field>
+  </form>
+
+  <ng-container>
```
GrandSchtroumpf: Why do you need a ng-container here ????

libs/ui/src/lib/toolbar/header/header.component.html
```diff
@@ -11,35 +9,53 @@
...
+    <button mat-icon-button [matMenuTriggerFor]="notifications" class="notification-button">
+      <mat-icon
+        *ngIf="(notificationsLength$ | async) as length; else empty"
+        svgIcon="notifications"
+        matBadge="{{ length }}"
+        matBadgeColor="primary"
+        matBadgeSize="medium"
+        class="notification-icon"
+      ></mat-icon>
       <ng-template #empty>
```
GrandSchtroumpf: ng-template should be outside of the container, or it won't work....

libs/ui/src/lib/toolbar/header/header.component.scss
```diff
@@ -37,3 +42,34 @@
...
+
+.notification-icon {
+  height: 25px;
+  width: 25px;
+}
+
+.account-icon {
+  height: 50px;
+  width: 50px;
+  margin-top: -5px;
```
GrandSchtroumpf: Negative

styles/style.scss
```diff
@@ -70,7 +82,12 @@ mat-card {
 
 button {
   border-radius: 0px !important;
-  font-family: 'Montserrat', 'Helvetica Neue', sans-serif !important;
+  font-family: 'Montserrat', 'Helvetica Neue', sans-serif !important; // Only for menu in movie card, WHY?
+}
+
+button:disabled {
+  color: var(--disabled)!important;
```
GrandSchtroumpf: You should not need this !important. You should customize the mat-button component instead.

libs/material/src/lib/delivery/delivery-form/delivery-form.component.scss
```diff
@@ -1,73 +1,96 @@
...
   box-shadow: 0px 2px 9px rgba(0, 0, 0, 0.2);
+  padding-bottom: 40px;
 }
 
 mat-sidenav-content {
   height: 100%;
   overflow: auto;
-  background-color: white;
-  padding-bottom: 5%;
+  background-color: #FFFFFF;
```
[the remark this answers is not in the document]
ioaNikas: Should I create a css variable for white color to respect our logic ?
We override defaut background-color with white one, as default is `#FAFAFA`.

libs/material/src/lib/delivery/delivery-item/delivery-item.component.html
```diff
@@ -4,27 +4,39 @@
     </span>
     <span class="second-dot mat-elevation-z4" [ngStyle]="{ 'background-color': getRandomColor() }">
     </span>
-    <span *ngFor="let stakeholder of delivery.stakeholders; last as isLast" class="names" fxLayout="row">
-      <h4>{{stakeholder.organization.name}}</h4>
+    <span
```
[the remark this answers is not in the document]
ioaNikas: 👌👌👌

libs/material/src/lib/delivery/delivery-item/delivery-item.component.ts
```diff
@@ -21,11 +21,13 @@ export class DeliveryItemComponent implements OnInit {
 
   public progression$: Observable<number>;
   public paymentIcon: string;
+  public stakeholders: string;
 
   constructor() {}
 
   ngOnInit() {
     this.paymentIcon = this.delivery.isPaid ? 'payed' : 'not_payed';
+    this.stakeholders = this.delivery.stakeholders.map(({ organization }) => organization.name).join(' | ');
```
GrandSchtroumpf: It won't update over time ? (if it does, use an observable)

libs/material/src/lib/delivery/stakeholder-list/stakeholder-list.component.html
```diff
@@ -1,15 +1,15 @@
...
       *ngIf="stakeholder.isAccepted"
       [stakeholder]="stakeholder"
     ></delivery-stakeholder-item>
   </div>
-  <button mat-icon-button (click)="navigateToTeamwork()">
-    <mat-icon>add_circle_outline</mat-icon>
-    Add a new signee
-  </button>
+  <div (click)="navigateToTeamwork()" class="add-signee mat-elevation-z6" fxLayout="row" fxLayoutAlign="center center">
+    <mat-icon svgIcon="add"></mat-icon>
```
GrandSchtroumpf: Why not using `<mat-icon>add</mat-icon>` instead ?

libs/material/src/lib/material/material-delivery-form/material-delivery-form.component.html
```diff
@@ -35,6 +35,17 @@
           </mat-option>
         </mat-select>
       </mat-form-field>
+      <ng-template #noStep>
+        <mat-form-field appearance="outline" class="mat-select-step">
+          <mat-label>Choose step</mat-label>
+          <mat-select formControlName="stepId">
+            <mat-option value="">
```
GrandSchtroumpf: If the value is empty, don't add it.

libs/material/src/lib/material/material-delivery-item/material-delivery-item.component.scss
```diff
@@ -37,28 +51,29 @@ mat-card-content {
...
 
 .border-left {
-  border-left: 1px solid #ECECEC;
+  border-left: 1px solid #ececec;
 }
 
 mat-icon {
-  color:#C1C1C1;
+  color: #c1c1c1;
   float: left;
```
GrandSchtroumpf: Never use float. This was already an old deprecated property when I start programing. Use flexbox instead.

libs/material/src/lib/material/material-delivery-item/material-delivery-item.component.scss
```diff
@@ -37,28 +51,29 @@ mat-card-content {
...
 }
 
 .border-left {
-  border-left: 1px solid #ECECEC;
+  border-left: 1px solid #ececec;
 }
 
 mat-icon {
-  color:#C1C1C1;
+  color: #c1c1c1;
```
GrandSchtroumpf: Is it changing the color of the icon ?

libs/material/src/lib/delivery/delivery-item/delivery-item.component.ts
ioaNikas: As delivery is an @Input() from parent already using and observable of deliveries, behavior seems ok. I just checked, new stakeholder is showing immediatly and without needing to reload the page when we add it from another tab/browser.

libs/material/src/lib/delivery/stakeholder-list/stakeholder-list.component.html
ioaNikas: This add icon was supposed to be a two-tone icon from ours wireframes. As `add_circle_outline` basic mat-icon is exactly the same, I'll switch to this one until we got the two-tone icon.

libs/material/src/lib/material/material-delivery-item/material-delivery-item.component.scss
ioaNikas: Got it. Anyway their is no mat-icon into the html so this is useless css.
Another float is used in notifications-item.scss, but fxFlex seems to behave differently in a mat-list (that has a fixed maximum width). I'll work to change it after Cannes if it's ok.

PR #328, opened 2019-05-10
Filter movie by ether

pldespaigne:
Really simple implem of filtering movie by tokens, this is mostly hardcoded values for the cannes demo !
It will need a big refactoring after cannes !

### DIFF
--- a/apps/movie-financing/movie-financing/src/app/explorer/movie-card-horizontal/movie-card-horizontal.component.html
+++ b/apps/movie-financing/movie-financing/src/app/explorer/movie-card-horizontal/movie-card-horizontal.component.html
@@ -1,48 +1,52 @@
-<div [class.swapVisual]="swapVisual">
-  <mat-card class="mat-elevation-z0">
-    <mat-card-header class="poster-container mat-elevation-z1">
-      <img mat-card-lg-image src="https://fakeimg.pl/657x251/" alt="Movie Poster">
-      <mat-card-title-group>
-        <mat-card-subtitle class="subtitle">Prestige Directors</mat-card-subtitle>
-        <div class="title">
-          <mat-card-subtitle class="stroke">Prestige Directors</mat-card-subtitle>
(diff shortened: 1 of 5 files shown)


apps/movie-financing/movie-financing/src/app/explorer/movie-card-horizontal/movie-card-horizontal.component.html
```diff
@@ -1,48 +1,52 @@
...
-          <mat-card-subtitle class="stroke">Prestige Directors</mat-card-subtitle>
-          <mat-card-subtitle class="director-name">Director Name</mat-card-subtitle>
-          <mat-card-title class="movie-title">Movie Name soooooo Long</mat-card-title>
-        </div>
-      </mat-card-title-group>
-    </mat-card-header>
-    <mat-card-content>
+<ng-container *ngIf="(user$ | async) as user">
+  <ng-container *ngIf="user.balance >= movie.minTokensRequired">
+    <div [class.swapVisual]="swapVisual">
```
GrandSchtroumpf: You don't need a div here, there is only one element inside. Just set the class to the mat-card.

libs/ethers/src/lib/relayer-wallet/relayer-wallet.ts
```diff
@@ -176,7 +176,7 @@ export class RelayerWallet implements ethers.Signer {
   public async getBalance(): Promise<string> {
     this._requireUsername();
     const balance = await this.provider.getBalance(`${this.username}.${baseEnsDomain}`).catch(error => {
-      console.error('Ooops something bad happened when querying user balance :', error);
+      // console.warn('Ooops the app try to get the balance of a (yet) none existing ENS domain, don`t worry everything is fine');
```
GrandSchtroumpf: Remove it if you don't need it.

pldespaigne: ✔️

PR #331, opened 2019-05-10
Issue#295 delivery send tx

pldespaigne:
- [x] sending a simple transaction from the relayer to the user's contract with the hash of the delivry id in the data field
- [x] update the stakeholder firestore with the tx hash
- [x] display a link to etherscan on the signed delivery component

### DIFF
--- a/functions/src/relayer.ts
+++ b/functions/src/relayer.ts
@@ -177,3 +177,24 @@ export const relayerRequestTokensLogic = async (
   console.log(`tx sent (request tokens) : ${tx.hash}`); // display tx to firebase logging
   return tx;
 };
+
+
+export const relayerSignDeliveryLogic = async (
+  { username, deliveryId, stakeholderId }: { username: string; deliveryId: string, stakeholderId: string },
+  config: any
+) => {
+  if (!username || !deliveryId || !stakeholderId) {
+    throw new Error('"username", "deliveryId" and "stakeholderId" are mandatory parameters !');
+  }
+  const relayer: Relayer = initRelayer(config);
+  // compute needed values
+  const fullName = `${username}.${config.relayer.basedomain}`;
+  const hash = utils.keccak256(utils.toUtf8Bytes(deliveryId));
+  const tx = await relayer.wallet.sendTransaction({to: fullName, data: hash});
+
+  await db.collection('deliveries').doc(deliveryId).collection('stakeholders').doc(stakeholderId).update({tx: tx.hash});
+
+  console.log(`tx sent (sign delivery) : ${tx.hash}`); // display tx to firebase logging
+  return tx;
+  
+}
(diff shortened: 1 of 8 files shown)


functions/src/relayer.ts
```diff
@@ -176,4 +176,25 @@ export const relayerRequestTokensLogic = async (
   const tx = await relayer.wallet.sendTransaction({to: fullName, value: weiAmount});
   console.log(`tx sent (request tokens) : ${tx.hash}`); // display tx to firebase logging
   return tx;
-};
\ No newline at end of file
+};
+
+
+export const relayerSignDeliveryLogic = async (
+  { username, deliveryId, stakeholderId }: { username: string; deliveryId: string, stakeholderId: string },
```
GrandSchtroumpf: Create a type for that.

functions/src/relayer.ts
```diff
@@ -176,4 +176,25 @@ export const relayerRequestTokensLogic = async (
...
+  config: any
```
GrandSchtroumpf: Don't use any. Create a type if needed.

functions/src/relayer.ts
```diff
@@ -176,4 +176,25 @@ export const relayerRequestTokensLogic = async (
...
+  if (!username || !deliveryId || !stakeholderId) {
+    throw new Error('"username", "deliveryId" and "stakeholderId" are mandatory parameters !');
+  }
+  const relayer: Relayer = initRelayer(config);
+  // compute needed values
+  const fullName = `${username}.${config.relayer.basedomain}`;
+  const hash = utils.keccak256(utils.toUtf8Bytes(deliveryId));
+  const tx = await relayer.wallet.sendTransaction({to: fullName, data: hash});
+
+  await db.collection('deliveries').doc(deliveryId).collection('stakeholders').doc(stakeholderId).update({tx: tx.hash});
```
GrandSchtroumpf: use 
```typescript
db.doc(`deliveries/${deliveryId}/stakholders/${stakeholderId}`).update({tx: tx.hash})
```

libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -114,6 +114,14 @@ export class AuthService {
     }
   }
 
+  public async signDelivery(deliveryId: string, stakeholderId: string) {
+    try {
+      return this.wallet.signDelivery(deliveryId, stakeholderId);
+    } catch (error) {
+      console.error('The relayer has failed to forward the signature to the blockchain');
```
GrandSchtroumpf: This error should be in the `signDelivery` of the relayer.
Why don't you inject direcly the wallet in the component ?

functions/src/relayer.ts
pldespaigne: ✔️

functions/src/relayer.ts
pldespaigne: ✔️

functions/src/relayer.ts
pldespaigne: ✔️

libs/auth/src/lib/+state/auth.service.ts
pldespaigne: ✔️

PR #332, opened 2019-05-10
create delayed-wrapper component for load movie home

Hugo-Hontans:


libs/movie/src/lib/movie/delayed-wrapper/delayed-wrapper.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  templateUrl: './delayed-wrapper.component.html',
+  styleUrls: ['./delayed-wrapper.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class DelayedWrapperComponent {
+
+  public isLoading: Observable<boolean>;
+
+  constructor() {
+    this.isLoading = new Observable(subscriber => {
```
GrandSchtroumpf: Use $ to specify that it's an observable.
What about : 
```
this.isLoading$ = of(true, false).pipe(debounceTime(1000));
```

PR #335, opened 2019-05-13
Prettier blockframes#2

ioaNikas:
- [x] Select all materials is now included in "Actions" button
- [x] Modified box shadow on cards
- [x] Signature blocks are now responsive for a better look on tablet format.
- [x] New style for teamwork (WIP)

### DIFF
--- a/libs/material/src/lib/delivery/movie-materials/actions/actions.component.ts
+++ b/libs/material/src/lib/delivery/movie-materials/actions/actions.component.ts
@@ -1,4 +1,4 @@
-import { Component, ChangeDetectionStrategy, Output, EventEmitter, Input, OnInit  } from '@angular/core';
+import { Component, ChangeDetectionStrategy, Output, EventEmitter, Input  } from '@angular/core';
 import { Step } from '../../+state';
 
 @Component({
@@ -7,19 +7,13 @@ import { Step } from '../../+state';
   styleUrls: ['./actions.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class ActionsComponent implements OnInit {
+export class ActionsComponent {
   @Output() newState : EventEmitter<string> = new EventEmitter();
   @Output() selectAllMaterials: EventEmitter<void> = new EventEmitter();
   @Input() allChecked: boolean;
   @Input() steps: Step[];
 
-  deliverySteps: Step[];
-
   constructor() { }
-
-  ngOnInit() {
-    this.deliverySteps = this.steps;
-  }
 
   public selectAll() {
     this.selectAllMaterials.emit();
--- a/libs/material/src/lib/delivery/movie-materials/actions/actions.component.ts
+++ b/libs/material/src/lib/delivery/movie-materials/actions/actions.component.ts
@@ -1,4 +1,5 @@
-import { Component, ChangeDetectionStrategy, Output, EventEmitter } from '@angular/core';
+import { Component, ChangeDetectionStrategy, Output, EventEmitter, Input, OnInit  } from '@angular/core';
+import { Step } from '../../+state';
 
 @Component({
   selector: 'material-actions',
@@ -6,12 +7,25 @@ import { Component, ChangeDetectionStrategy, Output, EventEmitter } from '@angul
   styleUrls: ['./actions.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class ActionsComponent {
+export class ActionsComponent implements OnInit {
   @Output() newState : EventEmitter<string> = new EventEmitter();
+  @Output() selectAllMaterials: EventEmitter<void> = new EventEmitter();
+  @Input() allChecked: boolean;
+  @Input() steps: Step[];
+
+  deliverySteps: Step[];
 
   constructor() { }
 
+  ngOnInit() {
+    this.deliverySteps = this.steps;
+  }
+
+  public selectAll() {
+    this.selectAllMaterials.emit();
+  }
+
   public changeState(state: string) {
-    this.newState.emit(state)
+    this.newState.emit(state);
   }
 }
--- a/libs/material/src/lib/delivery/movie-materials/actions/actions.component.ts
+++ b/libs/material/src/lib/delivery/movie-materials/actions/actions.component.ts
@@ -1,4 +1,4 @@
-import { Component, ChangeDetectionStrategy, Output, EventEmitter, Input  } from '@angular/core';
+import { Component, ChangeDetectionStrategy, Output, EventEmitter, Input } from '@angular/core';
 import { Step } from '../../+state';
 
 @Component({
@@ -7,11 +7,11 @@ import { Step } from '../../+state';
   styleUrls: ['./actions.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class ActionsComponent {
+export class ActionsComponent{
   @Output() newState : EventEmitter<string> = new EventEmitter();
   @Output() selectAllMaterials: EventEmitter<void> = new EventEmitter();
   @Input() allChecked: boolean;
-  @Input() steps: Step[];
+  @Input() deliverySteps: Step[];
 
   constructor() { }
 
@@ -22,4 +22,5 @@ export class ActionsComponent {
   public changeState(state: string) {
     this.newState.emit(state);
   }
+
 }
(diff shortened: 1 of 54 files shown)


libs/material/src/lib/material/material-template-form/material-template-form.component.html
```diff
@@ -1,44 +1,47 @@
...
-        class="add-button mat-elevation-z0"
-        (click)="updateMaterial()"
-      >
-        Save
-      </button>
     </div>
-  </mat-card-content>
+
+    <div class="step" fxLayout="column" fxLayoutAlign="end end">
+      <div fxLayoutAlign="center">
```
GrandSchtroumpf: This should have no impact because it's not a flexbox. I think you have one layer more than you should : 
```html
<form [formGroup]="form" (ngSubmit)="updateMaterial()" class="container" fxLayout="column">
  <mat-form-field>...</mat-form-field>
  <mat-form-field>...</mat-form-field>
  <div fxLayout="row" fxLayoutAlign="end center">
    <button></button>
    <button></button>
  </div>
</form>
```
This should work fine.

libs/material/src/lib/material/material-template-item/material-template-item.component.scss
```diff
@@ -1,45 +1,63 @@
...
-  margin-left: 5%;
+mat-card-actions {
+  visibility: hidden;
+  margin-right: 6px;
 }
 
-mat-card-content {
-  font-size: 12px;
-  color: #3B3B3B;
+.content {
```
GrandSchtroumpf: Try to avoid this names for classes. It's like variables in Javascript, you don't want to have `const x = ...`.

libs/material/src/lib/material/material-template-item/material-template-item.component.scss
```diff
@@ -1,45 +1,63 @@
...
+  width: 50%;
+  margin-left: 20px;
 }
 
 button {
-  color: #9C9C9C;
-  background-color: Transparent;
+  color: var(--disable);
```
GrandSchtroumpf: I don't understand why you use a "disable" color for a button that is not disable. Use the right semantic at the right time.

libs/material/src/lib/material/material-template-item/material-template-item.component.scss
```diff
@@ -1,45 +1,63 @@
...
-  cursor:pointer;
+  color: var(--primary);
+  cursor: pointer;
+  text-decoration: none;
 }
 
-button:focus{
-  color: #6200EE;
+button:focus {
+  color: var(--primary);
```
GrandSchtroumpf: You should not override mat-button like that. There is a default behavior, keep it. If we don't like it, we change it in a global customization, not in each file.

libs/material/src/lib/material/material-template-item/material-template-item.component.scss
```diff
@@ -1,45 +1,63 @@
...
 
 .border-left {
-  border-left: 1px solid #ECECEC;
+  border-left: 1px solid #ececec;
 }
 
 mat-icon {
-  color:#C1C1C1;
-  float: left;
+  color: var(--disabled);
```
GrandSchtroumpf: Same here. Is this icon disabled ?

libs/material/src/lib/template/template-item/template-item.component.scss
```diff
@@ -1,31 +1,30 @@
+mat-card {
+  margin: 5% 5%;
```
GrandSchtroumpf: Are you sure you want to use % ? You can write : `margin: 5%;`

libs/material/src/lib/template/template-item/template-item.component.scss
```diff
@@ -1,31 +1,30 @@
...
+  min-height: 102px;
+}
+
 mat-card-title {
   font-weight: 900;
-  font-size: 20px;
-  line-height: 24px;
-  letter-spacing: 0.25px;
+  font-size: 15px;
```
GrandSchtroumpf: For font-size is usually good practice to use rem value (1rem = standard size of text in the current browser).

libs/material/src/lib/template/template-item/template-item.component.scss
```diff
@@ -1,31 +1,30 @@
...
-  line-height: 20px;
-  letter-spacing: 0.196429px;
-  color: rgba(0, 0, 0, 0.87);
+  margin: 0;
+  margin-top: 10px;
+  font-weight: 400;
+  font-size: 10px;
+  line-height: 10px;
+  letter-spacing: 0.05px;
+  color: var(--disabled);
```
GrandSchtroumpf: Is it disabled for real or just the style ?

libs/material/src/lib/template/template-list/template-list.component.html
```diff
@@ -7,10 +7,10 @@
   >
     <mat-expansion-panel-header>
       <mat-panel-title fxLayout="row wrap" fxLayoutAlign="space-between center">
-        <h2>{{ org.name }}</h2>
+        {{ org.name }}
       <div class="divider"></div>
-      <button mat-mini-fab color="basic" class="view-button mat-elevation-z0"><mat-icon class="view-icon" color="primary">view_quilt</mat-icon></button>
-      <button mat-mini-fab color="basic" class="view-button mat-elevation-z0"><mat-icon class="view-icon" color="primary">view_module</mat-icon></button>
+      <button class="view-button mat-elevation-z0"><mat-icon class="view-icon" color="primary">view_quilt</mat-icon></button>
```
GrandSchtroumpf: Why not using mat-button ?

libs/material/src/lib/template/template-list/template-list.component.scss
```diff
@@ -52,13 +50,16 @@ mat-panel-title {
...
   width: 15px;
+  background-color: transparent;
   border: solid 1px var(--primary);
 }
 
 .view-icon {
-  margin-top: -18px;
-  margin-left: -6.5px;
   font-size: 13px;
+  position: relative;
```
GrandSchtroumpf: Why are you using relative ?

libs/movie/src/lib/movie/list/list.component.scss
```diff
@@ -49,15 +54,18 @@
...
-  border: solid 0.09rem var(--primary);
+  background-color: transparent;
+  border: solid 1px var(--primary);
 }
 
 .view-icon {
-  margin-top: -18px;
-  margin-left: -6.5px;
   font-size: 13px;
+  position: relative;
```
GrandSchtroumpf: Why are you using relative ?

libs/material/src/lib/delivery/movie-materials/actions/actions.component.ts
```diff
@@ -1,17 +1,31 @@
...
+  @Output() selectAllMaterials: EventEmitter<void> = new EventEmitter();
+  @Input() allChecked: boolean;
+  @Input() steps: Step[];
+
+  deliverySteps: Step[];
 
   constructor() { }
 
+  ngOnInit() {
+    this.deliverySteps = this.steps;
```
GrandSchtroumpf: You don't need this variable. You can access directly this.steps.

libs/material/src/lib/template/template-list/template-list.component.html
ioaNikas: mat-button seems to have a min-width set to 64px and I'll need to :ng-deep if I want to override this button in particular. As it is not a button we will use for now, and if you're ok, I'll switch this to a simple div for the moment and re-style it when we will have to use it for real.

libs/material/src/lib/material/material-template-form/material-template-form.component.html
```diff
@@ -1,47 +1,38 @@
 <mat-card class="container mat-elevation-z6">
   <mat-card-title>Edit your material</mat-card-title>
-  <form [formGroup]="form" (ngSubmit)="updateMaterial()" class="container" fxLayout="row">
-    <div class="inputs" fxLayout="column" fxLayoutAlign="center">
-      <mat-form-field appearance="outline" class="inputs">
-        <mat-label>Value</mat-label>
-        <input class="value" formControlName="value" matInput placeholder="Value" />
-      </mat-form-field>
+  <form [formGroup]="form" (ngSubmit)="updateMaterial()" class="container" fxLayout="column">
```
GrandSchtroumpf: Don't use the class "container". You have only one form in the component, use the selector `form` direcly.

libs/movie/src/lib/movie/list/list.component.scss
```diff
@@ -62,10 +63,8 @@ mat-panel-title {
 }
 
 .view-icon {
+  position: absolute;
```
GrandSchtroumpf: You should use "top / left / bottom / right" with absolute, or it means that you don't need it.

PR #346, opened 2019-05-14
functions: skip triggers when restoring

lsenta:


functions/src/backup.ts
```diff
@@ -174,6 +189,41 @@ function reEncodeObject(x: any): any {
...
 
+/**
+ * Set the restore timestamp to now.
+ */
+const setRestoreFlag = async () => {
+  return db.collection('_restore').doc('_DOC').set({ restoredAt: admin.firestore.FieldValue.serverTimestamp() });
+};
+
+const isRestoring = async () => {
+  const x = await db.collection('_restore').doc('_DOC').get();
```
GrandSchtroumpf: Pas sûr de comprendre ce qu'est "x" en lisant le code. Peut-être trouver un meilleur nom de variable.

functions/src/backup.ts
```diff
@@ -174,6 +189,41 @@ function reEncodeObject(x: any): any {
...
+
+    const now = admin.firestore.Timestamp.now();
+
+    // If we started the restoration less than twenty seconds ago, we are still in the restore process,
+    return restoredAt.toMillis() + THIRTY_SECONDS_IN_MS > now.toMillis();
+  }
+  return false;
+};
+
+export const skipWhenRestoring = (f: any) => {
```
GrandSchtroumpf: Maybe take the time to type it when you have some.

functions/src/backup.ts
lsenta: FIXED

functions/src/backup.ts
lsenta: Added as a TODO! (I'll look into replacing this solution by something more reliable after Cannes)

PR #351, opened 2019-05-14
implement mock up

pldespaigne:
compare


apps/movie-financing/movie-financing/src/app/explorer/compare/compare.component.scss
```diff
@@ -0,0 +1,35 @@
...
+  margin-left: 20%;
+  margin-right: 20%;
+  section {
+    h1 {
+      letter-spacing: 8px;
+      text-transform: uppercase;
+      color: #323232;
+      font-weight: 200;
+    }
+    mat-toolbar {
```
phpgeekfr: overriding material style without `::ng-deep` works ?

pldespaigne: yes it works  perfectly !

phpgeekfr: OK thanks ;)

PR #365, opened 2019-05-17
Movie guards

ioaNikas:
- [x] New guards to set the movie store.
- [x] Deleted unused functions and deprecated code.


apps/delivery/delivery/src/app/app-routing-module.ts
```diff
@@ -7,8 +7,8 @@ import { LayoutComponent } from './layout/layout.component';
 
 // Guards
 import { AuthGuard } from '@blockframes/auth';
-import { MovieGuard } from '@blockframes/movie';
-import { OrganizationListGuard } from '@blockframes/organization';
+import { MovieListGuard, MovieActiveGuard } from '@blockframes/movie';
+import { OrganizationListGuard } from 'libs/organization/src/lib/guards/organization-list.guard';
```
GrandSchtroumpf: Use @blockframes/organization

libs/movie/src/lib/movie/guards/movie-active.guard.ts
```diff
@@ -0,0 +1,22 @@
...
+export class MovieActiveGuard extends StateActiveGuard<Movie> {
+  readonly params = ['movieId'];
+  readonly urlFallback: 'layout';
+
+  constructor(private fireQuery: FireQuery, store: MovieStore, router: Router) {
+    super(store, router);
+  }
+
+  query({ movieId }) {
+    return this.fireQuery.fromQuery<Movie>(movieActiveQuery(movieId));
```
GrandSchtroumpf: You could just do : `this.fireQuery.fromQuery<Movie>(`moves/${movieId}`)` if you want.
Else for consistency you can do : 
```typescript
const query = movieActiveQuery(movieId);
return this.fireQuery.fromQuery(query);
```
You only need to type de fromQuery if you're using a string query.

libs/utils/src/lib/abstract-state-guard.ts
```diff
@@ -14,10 +14,10 @@ export abstract class StateListGuard<T> implements CanActivate, CanDeactivate<an
   canActivate(): Promise<boolean | UrlTree> {
     return new Promise((res, rej) => {
       this.subscription = this.query.pipe(
-        tap(entities => this.store.set(entities)),
+        tap(entities => this.store.set(entities))
       ).subscribe({
         next: result => res(!!result),
-        error: err => res(this.router.parseUrl(this.urlFallback))
+        error: err => {res(this.router.parseUrl(this.urlFallback))}
```
GrandSchtroumpf: Why ? you're making it less readable.

ioaNikas: Sorry, this is some remains of a console.log

PR #373, opened 2019-05-21
Template guards

ioaNikas:
- [x] Templates are now a collection, with materials as sub-collection
- [x] Template guards now use our new abstract guard
- [x] Template-picker dialog is no more, we now use delivery-template-picker component (with route and guard).


libs/material/src/lib/template/+state/template.service.ts
```diff
@@ -1,143 +1,96 @@
...
+      });
+      this.db.doc<Template>(`templates/${template.id}`).set(template);
 
       const batch = this.db.firestore.batch();
       materials.forEach(material => {
-        const materialWithoutStep = {...material, step: null}
+        const materialWithoutStep = { ...material, step: null };
         delete materialWithoutStep.step;
-        const materialRef = this.db.doc<Material>(`orgs/${orgId}/materials/${material.id}`).ref;
+        const materialRef = this.db.doc<Material>(
```
GrandSchtroumpf: Remove the ".ref" in the end, and do : `batch.set(materialDoc.ref, materialWithoutStep);`;
Like that it stays on one line.

libs/material/src/lib/template/guards/template-active.guard.ts
```diff
@@ -1,58 +1,26 @@
...
 });
 
-@Injectable({
-  providedIn: 'root'
-})
-export class TemplateActiveGuard extends StateActiveGuard implements CanActivate, CanDeactivate<any> {
+@Injectable({ providedIn: 'root' })
+export class TemplateActiveGuard extends StateActiveGuard<Template> {
+  readonly params = ['templateId'];
+  readonly urlFallback: 'layout';
```
GrandSchtroumpf: If urlFallback is "layout" most of the time, maybe we can make it the default value in the StateActiveGuard. Like that you specify it only when you want something different.

libs/material/src/lib/template/template-list/template-list.component.html
```diff
@@ -35,6 +35,6 @@
         <template-item class="mat-cards" [template]="template"></template-item>
       </article>
     </div>
+    <template-empty *ngIf="!(hasTemplates$ |async)"></template-empty>
```
GrandSchtroumpf: What does template-empty ?

ioaNikas: template-empty displays this message : 

This organization has no template yet. To continue, please create a template by clicking on Create a template.

I moved it into the mat-panel so each user organization show this message if it doesn't have any template yet.

Maybe we should display this message only if user got 0 template in the state ?

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -106,12 +106,10 @@ export class DeliveryService {
       .doc<Stakeholder>(`deliveries/${id}/stakeholders/${deliveryStakeholder.id}`)
       .set(deliveryStakeholder);
     if (!!templateId) {
-
-      const filterByMaterialId = (material: Material) => {
-        return this.templateQuery.getActive().materialsId.includes(material.id);
-      }
-
-      const materials = this.materialQuery.getAll({ filterBy: filterByMaterialId });
+      const snapshot = await this.db
```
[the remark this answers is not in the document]
ioaNikas: I got a problem here : it doesn't want to use .docs() on a variable that can be either a QuerySnapshot or a DocumentSnapshot.

Do you think we can go with this instead  ?

```
async function getCollectionOrDocument<T>(db: AngularFirestore, path: string) {
  if (path.split('/').length % 2 === 0) {
    const snapshot = await db.collection(path).ref.get();
    return snapshot.docs.map(doc => doc.data()) as T[];
  } else {
    const snapshot = await db.doc(path).ref.get();
    return snapshot.data() as T;
  }
}
```

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -120,13 +107,8 @@ export class DeliveryService {
...
   }
 
-
   public async addMovieMaterialsDelivery() {
-    const snapshot = await this.db
-      .collection<Material>(`movies/${this.movieQuery.getActiveId()}/materials/`)
-      .ref.get();
-    const movieMaterials = snapshot.docs.map(doc => doc.data()) as Material[];
-
+    const movieMaterials = await this.db.snapshot<Material[]>(`movies/${this.movieQuery.getActiveId()}/materials/`);
```
GrandSchtroumpf: User `const movieId = this.movieQuery.getActiveId();`

PR #374, opened 2019-05-22
refactoring as requested

pldespaigne:


libs/account/src/lib/profile/profile-view/profile-view.component.scss
```diff
@@ -9,10 +9,19 @@ main {
       color: rgba(0, 0, 0, 0.3);
       margin-right: 20px;
       min-width: 100px;
+      display: inline-block;
     }
     .value {
       font-size: 24px;
     }
+    .bio {
+      vertical-align: top;
```
GrandSchtroumpf: manage vertical and horizontal alignment with flexbox

PR #378, opened 2019-05-22
Wallet State

pldespaigne:


libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -34,7 +34,7 @@ export class AuthService {
   }
 
   public async signup(email: string, password: string) {
-    this.store.update({isEncrypting: true});
+    // this.store.update({isEncrypting: true});
     this.snackBar.open('We are curently encrypting your key pair, DO NOT CLOSE THIS PAGE BEFORE THE ENCRYPTION HAS ENDED !', 'OK', {
```
GrandSchtroumpf: Remove all capital letters.

libs/ethers/src/lib/wallet/+state/wallet.service.ts
```diff
@@ -0,0 +1,61 @@
+import { Injectable } from '@angular/core';
+import { WalletStore } from './wallet.store';
+
+@Injectable({ providedIn: 'root' })
+export class AuthService {
```
GrandSchtroumpf: Why is it called AuthService ?

libs/ethers/src/lib/wallet/+state/wallet.store.ts
```diff
@@ -0,0 +1,28 @@
+import { Injectable } from '@angular/core';
+import { Store, StoreConfig } from '@datorama/akita';
+
+export interface Wallet {
+  exist: boolean,
```
GrandSchtroumpf: You should not need this property

libs/ethers/src/lib/wallet/+state/wallet.store.ts
```diff
@@ -0,0 +1,28 @@
...
+  domain: string,
+  address: string
+  balance: string;
+}
```
GrandSchtroumpf: What about the "key" value ("default", "web", "mobile", ...)

libs/ethers/src/lib/wallet/+state/wallet.store.ts
```diff
@@ -0,0 +1,28 @@
+import { Injectable } from '@angular/core';
+import { Store, StoreConfig } from '@datorama/akita';
+
+export interface Wallet {
+  exist: boolean,
+  domain: string,
```
GrandSchtroumpf: call it 'ens' or "ensName", or "ensDomain", or something more explicit.

libs/ethers/src/lib/wallet/+state/wallet.store.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function createWallet(wallet: Partial<Wallet>) {
+  return { ...wallet } as Wallet;
+}
+
+const initialState: Wallet = {
+  exist: false,
+  domain: null,
+  address: null,
+  balance: '0.0',
```
GrandSchtroumpf: If there is no wallet, balance is 0

libs/ethers/src/lib/wallet/guards/wallet-active.guard.ts
```diff
@@ -0,0 +1,58 @@
...
+import { applyTransaction } from '@datorama/akita';
+import { User } from '@blockframes/auth';
+import { AngularFireAuth } from '@angular/fire/auth';
+
+export const UserQuery = (uid: string): Query<User> => ({
+  path: `users/${uid}`
+});
+
+@Injectable({ providedIn: 'root' })
+export class WalletActiveGuard implements CanActivate, CanDeactivate<Wallet> {
```
GrandSchtroumpf: Maybe you can create an abstract `StateGuard` for non-entity stores.

libs/ethers/src/lib/wallet/+state/wallet.store.ts
```diff
@@ -0,0 +1,28 @@
...
+  domain: string,
+  address: string
+  balance: string;
+}
+
+export function createWallet(wallet: Partial<Wallet>) {
+  return { ...wallet } as Wallet;
+}
+
+const initialState: Wallet = {
```
GrandSchtroumpf: Not initialize it if it doesn't exist. You can set to undefined or null.

libs/ethers/src/lib/wallet/guards/wallet-active.guard.ts
GrandSchtroumpf: And change the name to "WalletGuard".

libs/ethers/src/lib/wallet/+state/wallet.store.ts
pldespaigne: We are not currently creating new erc1077 at signup this is here for now only to distinguish old account (with erc1077) from new (without)
But as we discuss it will be remove during the Wallet refactoring

libs/ethers/src/lib/wallet/+state/wallet.store.ts
pldespaigne: will be fix by issue #379

libs/ethers/src/lib/wallet/guards/wallet-active.guard.ts
pldespaigne: This will change completely after #379 , did I need to change it any way ?

libs/ethers/src/lib/wallet/wallet-view/wallet-view.component.html
```diff
@@ -1,20 +1,20 @@
 <main>
-  <ng-container *ngIf="(user$ | async) as user; else userLoading">
-    <ng-container *ngIf="user.identity; else noWallet">
+  <ng-container *ngIf="(wallet$ | async) as wallet; else userLoading">
+    <ng-container *ngIf="wallet.exist; else noWallet">
```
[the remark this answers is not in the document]
pldespaigne: For now wallet = deployed erc1077, and we do not deploy new contract at signup, but this will change with #379

PR #384, opened 2019-05-27
backend(templates): remove materials when template is deleted

ioaNikas:
- [x] Template materials sub-collection is now properly deleted along with template. #380

### DIFF
--- a/apps/backend-functions/src/delete.ts
+++ b/apps/backend-functions/src/delete.ts
@@ -8,9 +8,9 @@ export async function deleteFirestoreMovie (
   context: functions.EventContext
 ) {
   const movie = snap.data();
+
   if (!movie) {
-    console.error(`This movie doesn\'t exist !`);
-    return null;
+    throw new Error(`This movie doesn\'t exist !`);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


apps/backend-functions/src/delete.ts
```diff
@@ -103,6 +103,25 @@ export async function deleteFirestoreDelivery (
...
 
+export async function deleteFirestoreTemplate (
+  snap: FirebaseFirestore.DocumentSnapshot,
+  context: functions.EventContext
+) {
+  const template = snap.data();
+
+  if (!template) {
+    console.error(`This template doesn't exist !`);
+    return null;
```
GrandSchtroumpf: Why not throwing an error ?

apps/backend-functions/src/main.ts
```diff
@@ -158,3 +163,7 @@ export const deleteDelivery = functions.firestore
 export const deleteMaterial = functions.firestore
   .document('deliveries/{deliveryId}/materials/{materialId}')
   .onDelete(backup.skipWhenRestoring(deleteFirestoreMaterial));
+
+export const deleteTemplate = functions.firestore
+  .document('templates/{templateId}')
+  .onDelete(backup.skipWhenRestoring(deleteFirestoreTemplate))
```
GrandSchtroumpf: You're doing that several time, maybe you want to create an reusable function for that

PR #403, opened 2019-05-28
Issue#386 check deliveries ids

ioaNikas:
- [x] onDeliveryUpdate is updated to check material.deliveriesIds and push or not delivery.id.
- [x] deleteFirestoreMaterial is updated and no longer checks material.id as movie material and delivery material don't always share the same id. Instead, it checks value, description and category fields.
- [x] Cleaned some unused code in material lib.

### DIFF
--- a/apps/backend-functions/src/delivery.ts
+++ b/apps/backend-functions/src/delivery.ts
@@ -121,6 +121,7 @@ export async function onDeliveryUpdate(
       );
 
       if (!!materialExist) {
+        // Check if delivery.id is already in material.deliveriesIds before pushing it in.
         if (!materialExist.deliveriesIds.includes(delivery.id)) {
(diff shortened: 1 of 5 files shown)


apps/backend-functions/src/delivery.ts
```diff
@@ -119,8 +119,12 @@ export async function onDeliveryUpdate(
           materialDelivery.category === materialMovie.category &&
           materialDelivery.description === materialMovie.description
       );
+
       if (!!materialExist) {
-        materialExist.deliveriesIds.push(delivery.id); // TODO: Check if delivery.id already exists before pushing
+        if (!materialExist.deliveriesIds.includes(delivery.id)) {
```
GrandSchtroumpf: Add a comment here, not straightforward to understand.

PR #404, opened 2019-05-28
[FIX] initial state for orgs and persistform plugin

phpgeekfr:


libs/organization/src/lib/+state/organization.store.ts
```diff
@@ -5,10 +5,19 @@ import { Organization } from './organization.model';
 export interface OrganizationState extends EntityState<Organization>, ActiveState<string>{
 }
 
+const initialState: OrganizationState  = {
+  form: {
```
GrandSchtroumpf: You don't need to provide the values of the form. Especially if there're null. You can just do  `form: {}`

libs/organization/src/lib/org-form/org-form.component.ts
```diff
@@ -59,7 +58,7 @@ export class OrgFormComponent implements OnInit, OnDestroy {
   public async submit() {
     if (!this.form.valid) {
```
GrandSchtroumpf: Check with PL to put this function into utils. It's going to be use everywhere.

phpgeekfr: Can you explain us (@pldespaigne  and me) a bit more on this particular point ? You want to put submit functions  into utils when it is possible (ie when form === firebase model )? For this organization form, this is not a modal form like PL is doing.
However, PL created #408 and assigned it to me to factorize some passwords validators.

GrandSchtroumpf: No, not all the submit fonction, just the if statement. Something like : 
```
function checkValidation(form: AbstractFormControl) {
  if (!form.valid) { ... }
}
```
Or we could create a based Form just on top of FormGoup, that other Form could inherit from : 
`class BaseFormGroup extends FormGroup`, and `class MaterialForm extends BaseFormGroup`.

PR #409, opened 2019-05-29
sort folder styles

Hugo-Hontans:


styles/variables.scss
```diff
@@ -1,9 +1,14 @@
...
-$blockframes-theme-primary: mat-palette($mat-bf-blue);
-$blockframes-theme-accent: mat-palette($mat-bf-lightblue);
-$blockframes-theme-warn: mat-palette($mat-bf-crimson);
-
 $primary: mat-color($blockframes-theme-primary);
 $accent: mat-color($blockframes-theme-accent);
 $warn: mat-color($blockframes-theme-warn);
+
+// CSS color variables
+:root {
```
GrandSchtroumpf: This should be remove at some point no ?

Hugo-Hontans: We can use directly variables like $primary in css but it requires to import variables.scss in each css. And like this we can add others colors

Hugo-Hontans: And change color easily for a dark-theme

GrandSchtroumpf: I think it would be better to use scss variables if possible, because most of documentation will do that.

PR #416, opened 2019-05-31
Issue#412 export keys

pldespaigne:
- [x] display private key of active key
- [x] display mnemonic (with path `0`) of active key
- [x] fix import key by mnemonic : automatically set path to `0`

### DIFF
--- a/libs/ethers/src/lib/key-manager/export-dialog/export-dialog.component.ts
+++ b/libs/ethers/src/lib/key-manager/export-dialog/export-dialog.component.ts
@@ -1,0 +1,37 @@
+import { Component, ChangeDetectionStrategy, OnInit, Inject } from '@angular/core';
+import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material';
+import { Wallet as EthersWallet, utils } from 'ethers';
+
+export interface ExportData {
+  wallet: EthersWallet;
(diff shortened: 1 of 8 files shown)


libs/ethers/src/lib/key-manager/export-dialog/export-dialog.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  wallet: EthersWallet;
+}
+
+@Component({
+  selector: 'key-manager-export',
+  templateUrl: './export-dialog.component.html',
+  styleUrls: ['./export-dialog.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ExportComponent implements OnInit {
```
GrandSchtroumpf: Call it "KeyExportComponent" or something more descriptif than "ExportComponent"

libs/ethers/src/lib/key-manager/export-dialog/export-dialog.component.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  constructor(
```
GrandSchtroumpf: Add `TODO: Add copy to clipboard for private key`. Maybe not for mnemonic...

PR #418, opened 2019-06-03
review services

Hugo-Hontans:
- [x] Add comments and review services

### DIFF
--- a/libs/utils/src/lib/firequery/firequery.ts
+++ b/libs/utils/src/lib/firequery/firequery.ts
@@ -13,10 +13,12 @@ export class FireQuery extends AngularFirestore {
   //////////////
 
   /** Return the current value of the path from Firestore */
-  async snapshot<T>(path: string): Promise<T> {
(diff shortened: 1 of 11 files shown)


libs/utils/src/lib/firequery/firequery.ts
```diff
@@ -13,8 +13,8 @@ export class FireQuery extends AngularFirestore {
   //////////////
 
   /** Return the current value of the path from Firestore */
-  async snapshot<T>(path: string): Promise<T> {
-    if (path.split('/').length % 2 === 0) {
+  public async snapshot<T>(path: string): Promise<T> {
+    if (path.split('/').length % 2 !== 0) {
```
pldespaigne: Are you sure about that ? Have you tested it ? I don't understand this line could you add a comment ?

Hugo-Hontans: Sure, i added comments for explain

PR #419, opened 2019-06-03
Refactor functions in stakeholder.ts

ioaNikas:
- [x] Created a common function to generate notifications on different type of stakeholders collection changes (create and delete events).

### DIFF
--- a/apps/backend-functions/src/stakeholder.ts
+++ b/apps/backend-functions/src/stakeholder.ts
@@ -1,8 +1,50 @@
 import { functions, db } from './firebase';
 import { prepareNotification, triggerNotifications, customMessage } from './notify';
-import { getDocument, getCount, getOrgsOfDelivery, getOrgsOfMovie, Delivery, DocID, Organization, Movie } from './utils';
+import {
+  getDocument,
+  getCount,
+  Delivery,
+  DocID,
+  Organization,
+  Movie,
+  SnapObject,
+  getOrgsOfDocument
+} from './utils';
 
-export async function onDeliveryStakeholderCreate (
+export async function onDeliveryStakeholderCreate(
+  snap: FirebaseFirestore.DocumentSnapshot,
+  context: functions.EventContext
+) {
+  stakeholdersCollectionEvent(snap, context);
+}
+
+export async function onMovieStakeholderCreate(
+  snap: FirebaseFirestore.DocumentSnapshot,
+  context: functions.EventContext
+) {
+  stakeholdersCollectionEvent(snap, context);
+}
+
+export async function onDeliveryStakeholderDelete(
+  snap: FirebaseFirestore.DocumentSnapshot,
+  context: functions.EventContext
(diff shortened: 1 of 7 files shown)


apps/backend-functions/src/stakeholder.ts
```diff
@@ -1,243 +1,130 @@
...
   if (!newStakeholder) {
-    throw new Error(`New stakeholder not found !`)
+    throw new Error(`New stakeholder not found !`);
   }
 
-  const [movie, newStakeholderOrg] = await Promise.all([
-    getDocument<Movie>(`movies/${context.params.movieID}`),
-    getDocument<Organization>(`orgs/${newStakeholder.orgId}`)
-  ]);
+  const collection = !!context.params.movieID ? 'movies' : 'deliveries';
```
GrandSchtroumpf: If you need more than two options you'll need a switch at some point.

apps/backend-functions/src/stakeholder.ts
```diff
@@ -1,243 +1,130 @@
...
-        getOrgsOfMovie(movie.id),
-        getCount(`movies/${movie.id}/stakeholders`)
+        getOrgsOfDocument(document.id, collection),
+        getCount(`${collection}/${document.id}/stakeholders`)
       ]);
-      const docID = {id: movie.id, type: 'movie'} as DocID;
-      const snapInformations = {
-        movieTitle: movie.title.original,
+      const snapInformations: SnapObject = {
+        movie: !!context.params.movieID
```
GrandSchtroumpf: Make a variable before the map.

apps/backend-functions/src/stakeholder.ts
```diff
@@ -1,243 +1,130 @@
...
-    const movieDoc = await db.doc(`movies/${movie.id}`).get();
-    const processedId = movieDoc.data()!.processedId;
-
-    if (processedId === context.eventId) {
-      return true;
-    }
+        eventType: context.eventType,
+        newStakeholderId: newStakeholder.id,
+        count: stakeholderCount,
+        delivery: !!context.params.deliveryID
```
GrandSchtroumpf: Make a variable before the map.

apps/backend-functions/src/stakeholder.ts
```diff
@@ -1,243 +1,130 @@
...
+/**
+ * Takes a list of Organization and a SnapObject to generate one notification for each users
+ * working on the document, with custom path and message.
+ */
+function createStakeholderNotifications(orgs: Organization[], snap: SnapObject) {
+  return orgs
+    .filter(org => !!org && !!org.userIds)
+    .reduce((ids: string[], { userIds }) => [...ids, ...userIds], [])
+    .map((userId: string) => {
+      const path = !!snap.delivery
```
GrandSchtroumpf: This should be before "return" it doesn't require `userId`. It would be easier to read.

Issue #423, opened 2019-06-04
Refactor File Uploader


pldespaigne:
Currently file uploader component is in `ui` but it import external module and is not generic enough.

- [x] create an abstract file uploader
- [x] create a firebase implementation
- [x] create a in browser implementation
- [x] make the component generic

PR #424, opened 2019-06-04
Issue417 save keystore

pldespaigne:
- [x] download encrypted key as a json file
- [x] import json file


libs/ethers/src/lib/key-manager/recover/recover.component.ts
```diff
@@ -40,32 +34,37 @@ export class RecoverComponent implements OnInit {
...
+      confirm: new FormControl('', [Validators.required])
+    }, { validators: [samePassword] });
   }
 
   cancel() {
     this.dialog.close(false);
   }
 
-  async recover() {
+  async recover(importType: string) {
```
GrandSchtroumpf: You should create a literal type : `type ImportType = 'json' | 'privatekey' | 'mnemonic';`

libs/ethers/src/lib/key-manager/recover/recover.component.ts
```diff
@@ -40,32 +34,37 @@ export class RecoverComponent implements OnInit {
...
       this.dialog.close(false);
       return;
     }
-    const { mnemonic, privateKey, password } = this.form.value;
-    if (!!mnemonic) {
-      this.service.importFromMnemonic(this.data.ensDomain, mnemonic, password);
-    } else if (!!privateKey) {
-      this.service.importFromPrivateKey(this.data.ensDomain, privateKey, password);
+    const { importValue, password } = this.form.value;
+    if (importType === 'mnemonic') {
```
GrandSchtroumpf: Use switch/default instead of if/else

libs/movie/src/lib/movie/form/form.promotional.component.html
```diff
@@ -12,6 +12,7 @@ <h3> Images </h3>
             accept="['.jpg', '.png']"
             path='test'
             [types]="['image/jpeg', 'image/png']"
+            [uploadOnFirestore]="true"
```
GrandSchtroumpf: Later this logic should be out of the component.

pldespaigne: issue #423

libs/ethers/src/lib/key-manager/+state/key-manager.service.ts
```diff
@@ -52,15 +52,20 @@ export class KeyManagerService {
   }
   /** create / encrypt / store / from mnemonic */
   importFromMnemonic(ensDomain: string, mnemonic: string, encryptionPassword: string) {
-    const wallet = EthersWallet.fromMnemonic(mnemonic, '0'); // path '0'
-    this._encryptAndStore(wallet, ensDomain, encryptionPassword);
+    const privateKey = utils.HDNode.mnemonicToEntropy(mnemonic); // mnemonic is a 24 word phrase corresponding to private key !== BIP32/39 seed phrase
```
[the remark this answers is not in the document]
pldespaigne: It seems that seeds are 32bit long (12 word) and private key are 64 bit long (24 word), I don't think it's possible to recover private key with 12 word mnemonic without storing HD wallet/seed/path info

libs/ethers/src/lib/key-manager/+state/key-manager.service.ts
```diff
@@ -52,15 +52,20 @@ export class KeyManagerService {
...
   }
   /** create / encrypt / store / from private key */
   importFromPrivateKey(ensDomain: string, privateKey: string, encryptionPassword: string) {
     const wallet = new EthersWallet(privateKey);
     this._encryptAndStore(wallet, ensDomain, encryptionPassword);
   }
 
+  importFromJsonFile(jsonString: string) {
+    const {address, ensDomain, keyStore} = JSON.parse(jsonString);
+    this.store.add({address, ensDomain, keyStore});
```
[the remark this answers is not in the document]
pldespaigne: if the key is already stored nothing will happen because key are stored by address, so in this case the key will be overwritten by itself

PR #427, opened 2019-06-05
Angular v8

GrandSchtroumpf:
Update : 
- @angular/core
- @angular/cli
- @angular/material
- @nrwl/schematics

@angular/fire should be update when it's ready.


tsconfig.json
```diff
@@ -24,8 +24,10 @@
       "@blockframes/organization": ["libs/organization/src/index.ts"],
       "@blockframes/account": ["libs/account/src/index.ts"],
       "@blockframes/material": ["libs/material/src/index.ts"],
-      "@blockframes/template": ["libs/material/src/index.ts"], // TODO: remove this in favor of dynamic imports when Angular 8 is live
-    }
+      "@blockframes/template": ["libs/material/src/index.ts"]
```
Hugo-Hontans: Do we still need that ? I saw the TODO

GrandSchtroumpf: I don't think so. Could you create an issue to remove it (I would like to keep this PR only for migration)

PR #441, opened 2019-06-11
rename ui-form into form and moved  files to utils

phpgeekfr:


libs/utils/src/lib/form/entity.ts
```diff
@@ -0,0 +1,26 @@
+import { FormControl, FormGroup, FormArray } from "@angular/forms";
+import { rules } from "./validators";
```
GrandSchtroumpf: Should be inside a "validators" folder. At the same level of "form"

libs/utils/src/lib/form/entity.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// Generic Control 
+export type EntityControl<T> = {
+  [key in keyof T]: FormControl | FormArray
```
GrandSchtroumpf: Could also be "| FormGroup".

libs/utils/src/lib/form/entity.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// Generic Entities
+export class EntityForm<T> extends FormGroup {
+  constructor(controls: EntityControl<T>, validators?: any) {
+    super(controls, validators)
+  }
+}
+
+export class EntityRulesForm<T> extends EntityForm<T> {
+  public rules = rules;
```
GrandSchtroumpf: Maybe call that "validators"

libs/account/src/lib/profile/profile-edit/profile-edit.form.ts
```diff
@@ -1,35 +1,31 @@
...
-  protected form : AbstractFormControls;
-
-  constructor(data?: any, controls? : any, validators?: any ) {
-    const f = new ProfileFormControls(data);
-    super(
-      controls !== undefined ? controls : f.controls,
-      validators !== undefined ? validators : f.validators
-    );
-    this.form = f;
+function createControls(entity: Profile): EntityControl<Profile> {
```
GrandSchtroumpf: Shouldn't it be `createProfileControls`?
Maybe you can have a `createProfile` method that will take care of filling all empty params : 
```
function createProfile(params?: Partial<Profile>): Profile {
  return {
    email: '',
    ...,
    ...(params || {})
  } as Profile
}
```
like that you can do :
```
function createProfileControls(entity: Partial<Profile>): EntityControl<Profile> {
  const profile = createProfile(entity);
  return {
    email: new EmailControl(profile.email, true)
  }
}
```

phpgeekfr:
closed by mistake

PR #444, opened 2019-06-12
build: use test-setup to load angular & jest presets

lsenta:
PR: https://github.com/blockframes/blockframes/pull/441
was merged despite breaking the build. It introduced links between module that broke the tests.

The origin commit: https://github.com/blockframes/blockframes/commit/f415c170b68e218a1241e77393c936454a5b4ab3

This patch fixes the issue by mocking DragEvent and make the `test-setup.ts` config consisten across libs and apps.


angular.json
```diff
@@ -240,7 +240,8 @@
           "builder": "@nrwl/jest:jest",
           "options": {
             "jestConfig": "libs/ui/jest.config.js",
-            "tsConfig": "libs/ui/tsconfig.spec.json"
+            "tsConfig": "libs/ui/tsconfig.spec.json",
+            "setupFile": "libs/ui/src/test-setup.ts"
```
GrandSchtroumpf: This file is compiled by the `tsconfig.json` of each app / lib.

lsenta: This config RUNS it at the beginning of the tests, since `setupFilesAfterEnv: '<rootDir>/jest.setup.js'` is broken for now,
https://github.com/nrwl/nx/issues/1343

Note that the PR doesn't create these, it create & config the missing ones so that all apps and libs are consistent.

GrandSchtroumpf: Ok, I see, thanks.

PR #445, opened 2019-06-13
frontend(rights): write rights on DB with documents via TX

ioaNikas:
- [x] Movies, Templates and Deliveries are now added to the DB with a transaction.
- [x] Documents rights are added to the DB within the same transaction.


libs/utils/src/lib/doc-transaction/doc-transaction.ts
```diff
@@ -0,0 +1,41 @@
...
+import { Template, Delivery, Material } from '@blockframes/material';
+import { FireQuery } from '../firequery/firequery';
+import { initializeOwnDocRights, initializeSharedDocRights } from './types';
+
+@Injectable({ providedIn: 'root' })
+export class DocTransaction {
+
+  constructor(private db: FireQuery) {}
+
+  public async createTransaction<T>(document: Movie | Template | Delivery, orgId: string, templateId?: string) {
```
GrandSchtroumpf: Why it's not inside FireQuery ?

GrandSchtroumpf: Need more documentation

libs/utils/src/lib/doc-transaction/doc-transaction.ts
```diff
@@ -0,0 +1,41 @@
...
+      const orgDocRightsRef = this.db.doc<any>(`rights/${orgId}/docs/${document.id}`).ref;
+      promises.push(tx.set(orgDocRightsRef, orgDocRights));
+
+      const sharedDocRightsRef = this.db.doc<any>(`rights/${orgId}/apps/${document.collectionName}Rights/docRights/${document.id}`).ref;
+      promises.push(tx.set(sharedDocRightsRef, sharedDocRights))
+
+      const documentRef = this.db.doc<T>(`${document.collectionName}/${document.id}`).ref;
+      promises.push(tx.set(documentRef, document));
+
+      if (!!templateId) {
```
GrandSchtroumpf: This looks super specific to templates and materials. Why is it inside the utils lib ?

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -95,19 +93,16 @@ export class DeliveryService {
       true
     );
 
-    this.db.doc<Delivery>(`deliveries/${id}`).set(delivery);
-    this.db
-      .doc<Stakeholder>(`deliveries/${id}/stakeholders/${deliveryStakeholder.id}`)
-      .set(deliveryStakeholder);
+    await this.db.createAndSetRights(delivery, stakeholder.orgId);
```
GrandSchtroumpf: Why `createAndSet`, for me `create` is the same thing as `set` ?

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -95,19 +93,16 @@ export class DeliveryService {
...
+
     if (!!templateId) {
-      const materials = await this.db.snapshot<Material[]>(`templates/${templateId}/materials`);
-      await Promise.all(
-        materials.map(material =>
-          this.db.doc<Material>(`deliveries/${id}/materials/${material.id}`).set(material)
-        )
-      );
-      return id;
+      await this.copyTemplate(delivery, templateId);
```
GrandSchtroumpf: Is copyTemplate dependant from createAndSetRights ? If not you might want to use a Promise.all.

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -209,12 +204,25 @@ export class DeliveryService {
...
     if (!validated.includes(stakeholderSignee.id)) {
       const updatedValidated = [...validated, stakeholderSignee.id];
-      this.db
-        .doc<Delivery>(`deliveries/${delivery.id}`)
-        .update({ validated: updatedValidated });
+      this.db.doc<Delivery>(`deliveries/${delivery.id}`).update({ validated: updatedValidated });
     }
   }
 
+  public async copyTemplate(delivery: Delivery, templateId: string) {
```
GrandSchtroumpf: Add a comment.

libs/material/src/lib/template/+state/template.model.ts
```diff
@@ -10,6 +11,7 @@ export interface BaseTemplate {
 export interface Template extends BaseTemplate {
   materials?: Material[];
   created: firestore.Timestamp;
+  _type: BFDocType;
```
GrandSchtroumpf: Here you should have : `_type: 'templates'`. You don't want a Template to be something else.

libs/movie/src/lib/movie/form/container.component.html
```diff
@@ -8,7 +8,7 @@
         <mat-icon>arrow_left</mat-icon>
       </ng-template>
     </button>
-    <movie-financing-view [form$]="form$" mode="preview"></movie-financing-view>
+    <movie-view [form$]="form$" mode="preview"></movie-view>
```
GrandSchtroumpf: You should pass an Observable as input. Add an issue on that.

libs/utils/src/lib/firequery/firequery.ts
```diff
@@ -96,27 +104,59 @@ export class FireQuery extends AngularFirestore {
...
     });
 
     return combineLatest(subQueries$).pipe(
       // Add the content to the object
-      map((subEntities) => subEntities.reduce((acc, {key, subentity}) => ({
-        ...acc,
-        [key]: subentity
-      }), {})),
-      map(subEntityObj => ({ ...entity, ...subEntityObj })),
+      map(subEntities =>
```
GrandSchtroumpf: It feels less readable like that.

libs/utils/src/lib/firequery/firequery.ts
```diff
@@ -96,27 +104,59 @@ export class FireQuery extends AngularFirestore {
...
+  public async createAndSetRights<T>(
+    document: BFDoc,
+    orgId: string
+  ) {
+    const promises = [];
+    const orgDocRights = initializeOwnerDocRights(document.id);
+    const sharedDocRights = initializeSharedDocRights(document.id);
+
+    return this.firestore.runTransaction(async (tx: firebase.firestore.Transaction) => {
+      const orgDocRightsRef = this.doc<any>(`rights/${orgId}/docs/${document.id}`).ref;
```
GrandSchtroumpf: Why put `<any>` and not `<T>` ?

libs/utils/src/lib/firequery/firequery.ts
```diff
@@ -96,27 +104,59 @@ export class FireQuery extends AngularFirestore {
...
+      promises.push(tx.set(orgDocRightsRef, orgDocRights));
+
+      const sharedDocRightsRef = this.doc<any>(
```
GrandSchtroumpf: Same here.

libs/utils/src/lib/firequery/firequery.ts
```diff
@@ -96,27 +104,59 @@ export class FireQuery extends AngularFirestore {
...
     return subQuery.length > 0;
   }
+
+  //////////////////////
+  // DOC TRANSACTIONS //
+  //////////////////////
+
+  /** Create a transaction for the document and add document rights (organization document rights and shared document rights) at the same time */
+  public async createAndSetRights<T>(
+    document: BFDoc,
```
GrandSchtroumpf: We could have some nice typescript tricks for BFDoc using generic like `BFDoc<T>` but this is not very important here. Just to tell we can go a little bit further into the type checking.

libs/utils/src/lib/firequery/types.ts
```diff
@@ -26,3 +26,54 @@ export function createQuery<T>(path: string): Query<T> {
...
+export interface SharedDocRights {
+  id: string;
+  canCreate: [];
+  canRead: [];
+  canUpdate: [];
+  canDelete: [];
+  admins: [];
+}
+
+export function initializeOwnerDocRights (id: string) {
```
GrandSchtroumpf: don't use space between function and parenthesis. Also you can add a small comment here.

PR #446, opened 2019-06-17
review movie-routing

Hugo-Hontans:


libs/movie/src/lib/movie/movie-routing.module.ts
```diff
@@ -23,30 +23,26 @@ export const routes: Routes = [
...
-      { path: 'promo',component: FormPromotionalComponent },
+      { path: '', redirectTo: 'view', pathMatch: 'full' },
+      { path: 'view', component: MovieViewComponent },
+      { path: 'teamwork', component: StakeholderViewComponent },
+      {
+        path: 'edit',
+        component: ContainerComponent,
+        children: [
+          { path: '', redirectTo: 'main', pathMatch: 'full' },
+          { path: 'main',component: FormMainComponent },
```
GrandSchtroumpf: If you ask me "What is `FormMainComponent`", I cannot answer. User `MovieMainFormComponent`, `MovieStoryFormComponent`, ... Or `MainMovieFormComponent` and `StoryMovieFormComponent`

Hugo-Hontans: Yes, I agree but I want continue to rename movie components in other PR

PR #449, opened 2019-06-18
Feat 448 lazy router

phpgeekfr:


apps/delivery/delivery/src/app/app-routing-module.ts
```diff
@@ -61,6 +61,7 @@ export const routes: Routes = [
   imports: [RouterModule.forRoot(routes,{
     anchorScrolling: 'enabled',
     onSameUrlNavigation: 'reload',
+    paramsInheritanceStrategy: 'always'
```
GrandSchtroumpf: Not sure in which case we want that. This will improve the amount of data to keep in memory.

phpgeekfr: I needed this because on movie form I use a sub router outlet for each parts of form. Without this param, movieId is not recognized as param in context-menu service and context-menu in toolbar is not displayed

PR #454, opened 2019-06-18
sub formsArrays needs to be cleared on form reset

phpgeekfr:
since form reset only set value to null, forms arrays needs to be cleared or else data will be duplicated (with null values) when re-populating

### DIFF
--- a/libs/movie/src/lib/movie/components/movie-form/movie.form.ts
+++ b/libs/movie/src/lib/movie/components/movie-form/movie.form.ts
@@ -205,9 +205,7 @@ export class MovieForm extends EntityForm<Movie> {
 
   /*
   * Clear controls that are FormArrays because form.reset() only set values to null.
-  * ie: If I have a control that is a FormArray with 3 items ctrl=[toto,tata,tutu],
-  * form reset will set data to null: names=[null,null,null] 
-  * but expected behavior in our case should be names=[] ( and this is what clear does)
+  * For more information @see: https://github.com/angular/angular/issues/31110
   */
   protected clearFormArrays() {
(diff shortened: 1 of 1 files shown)


libs/movie/src/lib/movie/components/movie-form/movie.form.ts
```diff
@@ -194,4 +194,18 @@ export class MovieForm extends EntityForm<Movie> {
       input.value = '';
     }
   }
+
+  reset(value?: any, options?: Object): void {
+    super.reset(value, options);
+    this.clearFormArrays();
+  }
+
+  protected clearFormArrays() {
```
GrandSchtroumpf: Add a comment. If I understand well, it clears all the keys that are arrays ?
The reset method doesn't do that already ?

phpgeekfr: Will add a comment. 
If I have a control that is a formArray with 3 items names=[toto,tata,tutu], form `reset` will set data to null: names=[null,null,null] but expected behavior in our case should lead to names=[] ( and this is what `clear` does)

GrandSchtroumpf: Ok, I think you can add an issue on Angular github also.

GrandSchtroumpf: Comment is a little bit long XD. But that's fine ;)

phpgeekfr: Comment added and issue opened : https://github.com/angular/angular/issues/31110

PR #457, opened 2019-06-19
frontend(rights): Initialize org with rights and check superAdmin

ioaNikas:
-[x] Added a .json with initial informations on organizations rights.
-[x] User has now only one organization (referenced as orgId in user document).
-[x] Some content only showed to the org superAdmin (no back rules yet).


libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -1,36 +1,43 @@
...
 
-      return Promise.all([orgTransaction, userOrgRightTransaction]);
-    }).then(() => {
-      console.log('Transaction successfully committed!');
-    }).catch((error) => {
-      console.log('Transaction failed: ', error);
-    });
+        return Promise.all([orgTransaction, updateUserTransaction]);
+      })
+      .then(() => {
```
GrandSchtroumpf: Remove your {} if you don't do anything else. then console.log.

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -1,36 +1,43 @@
...
+        console.log('Transaction successfully committed!');
```
GrandSchtroumpf: You shouldn't console.log in the frontend. You can return the tx you display something to the end user, if you think the user should see something. In any case don't log in the console in case of success.

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -1,36 +1,43 @@
...
 
 @Injectable({ providedIn: 'root' })
 export class OrganizationService {
   constructor(
     private store: OrganizationStore,
+    private query: OrganizationQuery,
     private db: FireQuery
   ) {}
 
   public async addMember(orgId: string, member: OrgMember): Promise<string> {
```
GrandSchtroumpf: Here your method is async, but there is no `await`. Do you want to wait for the transaction to be fullfilled ?

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -40,17 +47,30 @@ export class OrganizationService {
...
-        transaction.set(orgDoc.ref, newOrg),
-        // @todo admin slug comes from json
-        transaction.set(orgRightsDoc.ref, { orgId, rightNameSlug: [ROLES.ADMIN] })
-      ]);
-    }).catch((error) => {
-      console.log('Transaction failed: ', error);
-    });
+    this.db.firestore
+      .runTransaction(transaction => {
+        return Promise.all([
```
GrandSchtroumpf: It's really not readable. I don't understand what's happening here. Use a promises array, and create variables with understandable names.

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -62,4 +82,10 @@ export class OrganizationService {
   public remove(id: ID | ID[]) {
     this.store.remove(id);
   }
+
+  public getOrgRoles() {
+    return this.query
+      .selectActiveId()
+      .pipe(switchMap(id => this.db.collection(`rights/${id}/roles`).valueChanges()));
```
GrandSchtroumpf: By default I would say to not use `db` inside the query.
Does it make sense to have a store of the roles ? If yes, then I would create a `Rights` store, and use it's query here instead of firestore.

libs/organization/src/lib/org-initial-rights.json
```diff
@@ -0,0 +1,129 @@
+{
+  "orgPermissions": {
```
GrandSchtroumpf: Is it suppose to stay static ? Or is it a mock ?

libs/organization/src/lib/org-members-show/org-members-show.component.ts
```diff
@@ -21,23 +21,27 @@ export class OrgMembersShowComponent implements OnInit {
   public org$: Observable<Organization>;
   public addMemberForm: FormGroup;
   public mailsOptions: User[];
-  public rolesOptions: string[] = Object.values(ROLES);
+  public $rolesOptions: Observable<any[]>;
```
GrandSchtroumpf: $ should be in the end of the name of the variable. Please type your Observable

libs/organization/src/lib/org-members-show/org-members-show.component.ts
```diff
@@ -21,23 +21,27 @@ export class OrgMembersShowComponent implements OnInit {
...
     private orgQuery: OrganizationQuery,
     private snackBar: MatSnackBar,
     private builder: FormBuilder
   ) {
   }
 
   ngOnInit() {
-    this.mailsOptions = [];
-    this.org$ = this.orgQuery.selectActive();
+    this.enabled = from(this.service.isSuperAdmin());
```
GrandSchtroumpf: Why do you use `from` ? It doesn't do anything here. Either you want a snapshot on init, and you use `this.service.isSuperAdmin` as it. Or you [here](https://github.com/blockframes/blockframes/pull/457/files#diff-6235eca447f327e73e2e6e43caddfd31R55) you `selectActiveId` instead of `getActiveId`.

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -62,4 +82,10 @@ export class OrganizationService {
   public remove(id: ID | ID[]) {
     this.store.remove(id);
   }
+
+  public getOrgRoles() {
```
GrandSchtroumpf: Call it `selectRoles`. We are in the organisation service, and it returns an observable.

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -40,17 +47,30 @@ export class OrganizationService {
...
-    }).catch((error) => {
-      console.log('Transaction failed: ', error);
-    });
+    this.db.firestore
+      .runTransaction(transaction => {
+        return Promise.all([
+          transaction.set(orgDoc.ref, newOrg),
+          // @todo admin slug comes from json
+          transaction.set(orgRightsDoc.ref, { orgId, superAdmin: userId, ...INITIALRIGHTS.orgPermissions }),
+          INITIALRIGHTS.roles.forEach(role => {
```
lsenta: won't work, forEach returns `undefined`, you'll lose the promises produced by `transaction.set`

libs/organization/src/lib/org-members-show/org-members-show.component.html
```diff
@@ -24,18 +24,19 @@
       </mat-card>
     </mat-list>
 
-    <mat-card>
+    <mat-card *ngIf="!!(enabled | async)">
```
lsenta: This is the code that you're going to reuse throughout the whole application, make it more explicit?

Something like:
```
*ngIf="(role.isAdmin | async)
```

As you add more role-based template you'll end up with a nice reusable API, something like:

```
- role.isAdmin
- role.canEdit(deliveryID)
- role.etc
```

libs/auth/src/lib/+state/auth.query.ts
```diff
@@ -24,4 +24,8 @@ export class AuthQuery extends Query<AuthState> {
     return this.getValue().requestedRoute;
   }
 
+  get userId() {
+    return this.user.uid;
+  }
+
```
lsenta: autoformat before saving

libs/organization/src/lib/+state/organization.service.ts
lsenta: +1 also, you're just catching the error and logging it => it's better to throw it explicitly => await without the console

libs/organization/src/lib/+state/organization.service.ts
GrandSchtroumpf: Why not using a map then ?

lsenta: factor out this code, you'll reuse "assign the hardcoded role X to a user" often

libs/organization/src/lib/org-initial-rights.json
```diff
@@ -0,0 +1,129 @@
...
+    "admins": [],
+    "canCreate": [],
+    "canRead": [],
+    "canUpdate": [],
+    "canDelete": []
+  },
+  "appsPermissions": [
+    {
+      "name": "MediaDelivering",
```
lsenta: Use a regular typescript object and use typing to avoid typo

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -1,36 +1,43 @@
 import { Injectable } from '@angular/core';
 import { ID } from '@datorama/akita';
-import { createOrganization, Organization, OrgMember, ROLES } from './organization.model';
+import { createOrganization, Organization, OrgMember } from './organization.model';
 import { OrganizationStore } from './organization.store';
 import { FireQuery } from '@blockframes/utils';
+import * as INITIALRIGHTS from './../org-initial-rights.json';
```
lsenta: INITIAL_RIGHTS

libs/organization/src/lib/org-members-show/org-members-show.component.ts
lsenta: Does it makes sense to move the observable to the service?

GrandSchtroumpf: It should be inside the service.

libs/organization/src/lib/+state/organization.service.ts
ioaNikas: We'll leave all roles logic for another sprint session and focus on basic authorizations in the first time. I deleted everything related to the roles.

GrandSchtroumpf: Ok, but still call it `select***` as this is an observable.

libs/organization/src/lib/+state/organization.service.ts
ioaNikas: Indeed, I don't need to wait for the transaction.

libs/organization/src/lib/+state/organization.service.ts
ioaNikas: Sure, I get the concept :+1:

libs/organization/src/lib/+state/organization.service.ts
ioaNikas: I'll remove this function until we come back to work on roles again. A Rights store could be a good option to have an easy access to org and user authorizations without querying firestore each time.

libs/organization/src/lib/org-initial-rights.json
ioaNikas: It is supposed to be static. Maybe we could created a function into à rights.model to initialize Organizations rights (with types advantage) and get rid of that json.

GrandSchtroumpf: If it's static, then I would recommand to create a .ts file with variables : 
```typescript
/** Comment here */
export const orgPermissions = {
 ...
}
```
Like that you can add comments, improve typing, and you don't have to load everything all the time : 
```
import { orgPermissions } from '...' // Load only one
import * as PERMISSIONS from '...' // Load everything
```

libs/organization/src/lib/org-initial-rights.json
ioaNikas: As we decided together, we'll switch to a store.

libs/rights/src/lib/+state/rights.service.ts
```diff
@@ -1,9 +1,39 @@
...
+  ) {
+    const promises = [];
+    const orgDocRights = initializeOrgDocRights(document.id, orgId);
+    const userDocRights = initializeUserDocRights(document.id);
+
+    return this.db.firestore.runTransaction(async (tx: firebase.firestore.Transaction) => {
+      const orgDocRightsRef = this.db.doc<T>(`rights/${orgId}/orgDocsRights/${document.id}`).ref;
+      promises.push(tx.set(orgDocRightsRef, orgDocRights));
+
+      const userDocRightsRef = this.db.doc<T>(
```
GrandSchtroumpf: do : `const { ref } = this.db.doc<T>(....)`

libs/rights/src/lib/guards/rights.active.guard.ts
```diff
@@ -1,44 +1,68 @@
...
     return this.afAuth.authState.pipe(
-      switchMap(user => {
-        const userDoc = this.fireQuery.doc(`users/${user.uid}`)
-        const userOrgId = userDoc.orgId;
-        const query = rightsActiveQuery(userOrgId);
-        return this.fireQuery.fromQuery<OrganizationRights>(query);
-      })
-    );
+      switchMap(fbUser => {
+        return this.fireQuery.doc<User>(`users/${fbUser.uid}`).valueChanges()
```
GrandSchtroumpf: You shouldn't do that. The `AuthQuery` should be the only source of thruth for the user.
```
constructor(... private authQuery: AuthQuery) {}

query() {
    switchMap(user => this.fireQuery.fromQuery<OrganizationRights>(rightsActiveQuery(user.orgId)))
  )
}
```

libs/utils/src/lib/abstract-state-guard.ts
```diff
@@ -32,7 +32,7 @@ export abstract class StateActiveGuard<T> implements CanActivate, CanDeactivate<
   /** Specify the names of the params to check,
    *  should be same name as the route params
    */
-  abstract params: string[];
+  abstract params?: string[];
```
GrandSchtroumpf: Discard this change because you're not using it anymore.

libs/auth/src/lib/+state/auth.store.ts
```diff
@@ -36,7 +36,10 @@ export interface AuthState {
 }
 
 export function createUser(user: Partial<User>) {
```
GrandSchtroumpf: do `createUser(user: Partial<User> = {})`. Like that you can do `createUser()`

libs/organization/src/lib/+state/org-members.service.ts
```diff
@@ -61,4 +63,12 @@ export class OrgMembersService {
   private collection(orgID: string): AngularFirestoreDocument<Organization> {
     return this.db.collection('orgs').doc(orgID);
   }
+
+  /** Checks if the connected user is the Super Admin of his organization */
+  public isSuperAdmin(): boolean {
```
GrandSchtroumpf: Would this make sense to  move this to `RightsQuery` ? Else in `OrgQuery`

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -44,8 +44,10 @@ export class OrganizationService {
     const newOrg: Organization = createOrganization({ ...org, id: orgId, userIds: [userId] });
 
     const orgDoc = this.db.doc(`orgs/${orgId}`);
+    const orgRights = initializeOrgRights({orgId, superAdmin: userId});
     const orgRightsDoc = this.db.doc(`rights/${orgId}`);
     const userDoc = this.db.doc(`users/${userId}`);
+    const apps : App[] = [App.mediaDelivering, App.mediaFinanciers, App.storiesAndMore ]
```
GrandSchtroumpf: It looks like mock value

libs/organization/src/lib/+state/org-members.service.ts
GrandSchtroumpf: Shouldn't it be an observable ?

libs/rights/src/lib/guards/rights.active.guard.ts
```diff
@@ -30,35 +30,54 @@ export class RightsActiveGuard {
...
+        return this.fireQuery
+          .doc<User>(`users/${fbUser.uid}`)
+          .valueChanges()
           .pipe(
-            switchMap(user => this.fireQuery.fromQuery<OrganizationRights>(rightsActiveQuery(user.orgId)) )
+            switchMap(user => {
+              // If the user is new to blockframes and doesn't have an orgId,
+              // he's still able to sign up or sign in.
+              if (!user.orgId) {
+                return from([null]);
```
GrandSchtroumpf: Use `of(null)` instead of `from([null])`

libs/rights/src/lib/guards/rights.active.guard.ts
```diff
@@ -30,35 +30,54 @@ export class RightsActiveGuard {
...
+    return new Promise(res => {
+      this.subscription = this.query()
+        .pipe(
+          tap(entity => {
+            // If there's no entity, return and don't apply transaction as the user
+            // doesn't belong to an organization yet.
+            if (!entity) {
+              return;
+            }
+            applyTransaction(() => {
```
GrandSchtroumpf: I don't understand what you're doing here.

libs/organization/src/lib/+state/org-members.service.ts
ioaNikas: How can I get a Observable<boolean> ? Even if I compare 2 Observable<string> , I get a boolean snapshot in return.

libs/rights/src/lib/guards/rights.active.guard.ts
ioaNikas: Also so much easier to read, thanks !

libs/organization/src/lib/+state/org-members.service.ts
GrandSchtroumpf:
```typescript
return this.rightsQuery.selectActive().pipe(
  map(({ superAdmin }) => superAdmin === this.auth.userId)
)
```

libs/rights/src/lib/guards/rights.active.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+    private router: Router
+  ) {}
+
+  query() {
+    return this.auth.user$.pipe(
+      switchMap(user => {
+        // If the user is new to blockframes and doesn't have an orgId,
+        // he's still able to sign up or sign in.
+        if (!user.orgId) {
+          return of(null);
```
GrandSchtroumpf: You shouldn't have to do that. The guard is used after authentication. So if user has no orgId you should just redirect to the select organisation page....

libs/rights/src/lib/guards/rights.active.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+        return this.fireQuery.fromQuery<OrganizationRights>(rightsActiveQuery(user.orgId));
+      })
+    );
+  }
+
+  canActivate(): Promise<boolean | UrlTree> | UrlTree {
+    return new Promise(res => {
+      this.subscription = this.query()
+        .pipe(
+          tap(entity => {
```
GrandSchtroumpf: `entity` is not a good name in this case. Only use this kind of generic names for generic classes. In this case this is a specific class ("Rights")

libs/rights/src/lib/guards/rights.active.guard.ts
```diff
@@ -0,0 +1,79 @@
...
+            // If the entity is null, don't we don't set the store and
+            // use our fallback path to connect the user.
+            if (!entity) {
+              return this.router.parseUrl(this.urlFallback);
```
GrandSchtroumpf: I don't understand. If you want to return the urlFallback, then just throw, you'll go into the `error` of subscribe.

libs/rights/src/lib/rights.module.ts
```diff
@@ -0,0 +1,7 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+
+@NgModule({
+  imports: [CommonModule]
```
GrandSchtroumpf: I don't think you need `CommonModule` as you don't have components. If you plan to have components, then keep it.

libs/ui/src/lib/form/inputs/password/password.component.html
```diff
@@ -1,13 +1,13 @@
 <mat-form-field [formGroup]="form">
   <input matInput type="password" [placeholder]="placeholder" [formControlName]="name"  required [errorStateMatcher]="matcher ? matcher : undefined"/>
-  <mat-hint *ngIf="showHint">Between {{ minLength.requiredLength }} and {{ maxLength.requiredLength }} characters.</mat-hint>
+  <mat-hint *ngIf="showHint && !!minLength && !!maxLength">Between {{ minLength.requiredLength }} and {{ maxLength.requiredLength }} characters.</mat-hint>
```
GrandSchtroumpf: Maybe add ` && !!minLength && !!maxLength` inside the `showHint` getter.

libs/auth/src/lib/+state/auth.store.ts
```diff
@@ -34,8 +35,11 @@ export interface AuthState {
   isBalanceLoading: boolean;
 }
 
-export function createUser(user: Partial<User>) {
-  return { ...user } as User;
+export function createUser(user: Partial<User> = {}) {
+  return {
+    orgId: '',
```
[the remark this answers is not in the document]
ioaNikas: Indeed, forgot to delete it

libs/organization/src/lib/+state/org-members.service.ts
```diff
@@ -47,7 +49,15 @@ export class OrgMembersService {
...
   }
 
+  public selectOrgRights(): Observable<OrganizationRights> {
+    // TODO: select user.orgId instead of active organization
+    return this.orgQuery.selectActiveId().pipe(
+      switchMap(id => this.db.doc<OrganizationRights>(`rights/${id}`).valueChanges)
+    )
+  }
+
   private collection(orgID: string): AngularFirestoreDocument<Organization> {
```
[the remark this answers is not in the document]
ioaNikas: Deleted most of the code here as it seems to be unused (smells like v1 of orgRights).

libs/organization/src/lib/+state/org-members.service.ts
```diff
@@ -47,7 +49,15 @@ export class OrgMembersService {
       );
   }
 
+  public selectOrgRights(): Observable<OrganizationRights> {
+    // TODO: select user.orgId instead of active organization
```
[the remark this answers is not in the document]
ioaNikas: Also removed this as the store is now setted with organization rights with guard, sorry for the loss of time.

libs/organization/src/lib/+state/org-members.service.ts
```diff
@@ -47,7 +49,15 @@ export class OrgMembersService {
       );
   }
 
+  public selectOrgRights(): Observable<OrganizationRights> {
```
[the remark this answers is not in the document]
ioaNikas: Removed this in profit of guard

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -11,46 +12,64 @@ export class OrganizationService {
...
 
-      return Promise.all([orgTransaction, userOrgRightTransaction]);
-    }).then(() => {
-      console.log('Transaction successfully committed!');
-    }).catch((error) => {
-      console.log('Transaction failed: ', error);
-    });
+        return Promise.all([orgTransaction, updateUserTransaction]);
+      })
+      .catch(error => {throw Error(error)});
```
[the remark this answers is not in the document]
ioaNikas: `.catch(error => throw Error(error));`
Got an error with '{' expected. Writing only Error(error) show no error but seems really strange. Is there another way to write this ?

libs/organization/src/lib/org-members-show/org-members-show.component.html
```diff
@@ -24,36 +24,25 @@
       </mat-card>
     </mat-list>
 
-    <mat-card>
+    <mat-card *ngIf="!!(isSuperAdmin | async)">
       <mat-card-header>
         <mat-card-title>Add Member</mat-card-title>
+        <mat-card-subtitle *ngIf="!!disabled">(Only Super Admin can add a member)</mat-card-subtitle>
```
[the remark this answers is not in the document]
ioaNikas: Once more, its an oversight, used it for testing and didn't removed, sorry about that.

libs/organization/src/lib/+state/organization.service.ts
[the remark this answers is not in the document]
GrandSchtroumpf: Well I don't really see the need of throwing anything here because you're not going to catch the error anywhere else. So for this time I thing you can use `console.error(error)`

libs/organization/src/lib/org-members-show/org-members-show.component.ts
```diff
@@ -21,23 +22,24 @@ export class OrgMembersShowComponent implements OnInit {
...
   }
 
   ngOnInit() {
-    this.mailsOptions = [];
-    this.org$ = this.orgQuery.selectActive();
+    this.isSuperAdmin = this.rightsQuery.isSuperAdmin;
     this.addMemberForm = this.builder.group({
-      user: null,
-      role: ''
+      user: null
```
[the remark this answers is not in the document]
ioaNikas: Ok I get it. If you're ok, I'll put an empty `role` again as we will need to use this formGroup again on roles implementation.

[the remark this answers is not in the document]
GrandSchtroumpf: Sure, you know better.

libs/rights/src/lib/guards/rights.active.guard.ts
ioaNikas: When I throw here, it loop endlessly and never get into `applyTransaction`, it just go back into the `tap` and I can't figure out why.

I found 2 "solutions" here, either redirect with `urlFallBack` or simply `return` but none of them seems to be the correct way to do.

libs/auth/src/lib/guard/auth.guard.ts
```diff
@@ -17,19 +23,27 @@ export class AuthGuard implements CanActivate {
     private router: Router
   ) {}
 
-  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): boolean | Observable<boolean | UrlTree> {
-    this.store.update({ requestedRoute: null })
+  canActivate(
+    route: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot
+  ): boolean | Observable<boolean | UrlTree> {
+    this.store.update({ requestedRoute: null });
```
GrandSchtroumpf: Is it supposed to stay ?

libs/auth/src/lib/guard/auth.guard.ts
```diff
@@ -17,19 +23,27 @@ export class AuthGuard implements CanActivate {
...
     // Connected on the app
     if (!!this.query.user) return true;
     // Wait for the server to give first answer
     return this.afAuth.authState.pipe(
-      map(auth => {
-        if (auth) { // Connected on Firebase
+      switchMap(auth => {
+        if (auth) {
+          // Connected on Firebase
           this.service.subscribeOnUser();
```
GrandSchtroumpf: What is this doing ? When do you unsubscribe ?

libs/auth/src/lib/guard/auth.guard.ts
```diff
@@ -17,19 +23,27 @@ export class AuthGuard implements CanActivate {
...
-          return true;
-        } else {    // Not connected
-          this.store.update({ requestedRoute: state.url })
-          return this.router.parseUrl('/auth')
+          return this.query.user$.pipe(
+            map(user => !!user),
+            filter(hasUser => hasUser)
```
GrandSchtroumpf: Why do you filter ? If there is no user, the guad will never complete and the app will freeze...

libs/rights/src/lib/guards/rights.guard.ts
```diff
@@ -32,30 +32,34 @@ export class RightsGuard {
   ) {}
 
   isUrlTree(result: OrganizationRights | UrlTree) {
+    console.log("result: ", result)
```
GrandSchtroumpf: remove console.log

libs/auth/src/lib/guard/auth.guard.ts
ioaNikas: Yes, as it is supposed to save the url state in case someone sent you a link to a document but you're not signed in already. After signin in, it is supposed to redirect you on this route.

libs/auth/src/lib/guard/auth.guard.ts
ioaNikas: This function is subscribing when they're is a user in AngularFireAuth.authState as you can see here :

```
public subscribeOnUser() {
    this.afAuth.authState.pipe(
      takeWhile(user => !!user),
      switchMap(({ uid }) => this.db.doc<User>(`users/${uid}`).valueChanges()),
    ).subscribe(user => {
      this.store.update({ user });
    });
```
It also set/update the akita store with the user and watch on changes to update it when needed (tell my if I'm wrong).

Maybe we should use our subscription logic with canActivate and canDeactivate  for auth guard too ?

libs/auth/src/lib/guard/auth.guard.ts
ioaNikas: The guard won't wait for the observable to return his value in this case. Laurent told me it is a race condition issue and we can solve this by waiting for our user to be loaded. If there is no user, won't filter return false ?

GrandSchtroumpf: No filter emit only if true. It means that nothing will be emitted until it's true. It means that in your case canActivate never returns false.

lsenta: Remember the marble diagram we drew @ioaNikas? filter will skip the false values.

@GrandSchtroumpf this is related to the question I had around Guards & assumptions:
Until then the guard would call the `service. subscribeOnUser()` that starts the user retrieval and return true immediately.
=> The guard would complete if there are no user data and every component downstream would have to check again if the user profile is actually loaded or not.

This is a race condition since it depends on how fast firestore syncs the profile, on low latency network the profile is loaded before components are rendered. On slower networks, you'll get cases where some components breaks because they forgot to check `user has been loaded`.

With this change we returns an observable that waits for the user profile.
=> Component "within" the AuthGuard can assume that the profile is correctly setup when they are invoked.

Since we already know the user is logged in and the subscribe is live when we enter this observable, the case "no user" falls between very unlikely to impossible.

PR #458, opened 2019-06-19
Generic forms

GrandSchtroumpf:
Create Generic Forms for Entities, Field and List.


libs/utils/src/lib/form/forms/entity.form.ts
```diff
@@ -0,0 +1,36 @@
+import { FormControl, FormGroup, FormArray } from '@angular/forms';
+import { Validator, AsyncValidator } from './types';
+
+/** Generic EntityControl */
+export type EntityControl<T> = { [key in keyof T]: FormControl | FormGroup | FormArray };
+
+/** Generic FormGroup for Entity */
+export class FormEntity<E, Control extends EntityControl<E>> extends FormGroup {
```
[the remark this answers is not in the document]
GrandSchtroumpf: The `Validators.min` provide already this kind of information. See documentation [here](https://angular.io/api/forms/Validators#min-usage-notes). I'll update the password.component.html.

libs/ethers/src/lib/key-manager/forms/password.form.ts
```diff
@@ -1,25 +1,25 @@
...
-    password: '',
-    ...(params || {})
-  } as Password
-}
+// function createPassword(params?: Partial<Password>): Password {
+//   return {
+//     password: '',
+//     ...(params || {})
+//   } as Password
+// }
```
Hugo-Hontans: Why not delete the file if you comment it completely ?

GrandSchtroumpf: Yes you're right.

PR #461, opened 2019-06-20
Issue#459 update to create2 clean

pldespaigne:


contracts/sol/ERC1077.sol
```diff
@@ -8,13 +10,22 @@ import "../../node_modules/openzeppelin-solidity/contracts/math/SafeMath.sol";
...
-    constructor(address _key) KeyHolder(_key) public {
+    /* solium-disable-next-line no-empty-blocks */
+    constructor() KeyHolder(address(0x0)) public {} // this should never be called and is here only to satisfy the solc compiler
+
+    // this function play the role of the "constructor",
+    // it can be called only once, and should be called before any other (non pure/view) functions
+    function init() public requireNotInit() {
+        Initable.init();
+        this.addKey(address(0xdeAD00000000000000000000000000000000dEAd), 1); // ! WARNING CHANGE HARDCODED ADDRESS WITH USER KEY AT DEPLOY
+        recoverAddress = 0xdeAD00000000000000000000000000000000dEAd; // ! WARNING CHANGE HARDCODED ADDRESS AT DEPLOY
```
GrandSchtroumpf: Create a Static storage value instead of specifying it inside the init function twice.

contracts/sol/ERC1077.sol
```diff
@@ -8,13 +10,22 @@ import "../../node_modules/openzeppelin-solidity/contracts/math/SafeMath.sol";
...
     }
 
     function canExecute(
```
GrandSchtroumpf: Need comments

contracts/sol/ERC1077.sol
```diff
@@ -99,7 +110,7 @@ contract ERC1077 is KeyHolder, IERC1077 {
         address gasToken,
         uint gasLimit,
         OperationType operationType,
-        bytes memory signatures) public returns (bytes32)
+        bytes memory signatures) public requireInit() returns (bytes32)
```
GrandSchtroumpf: Format : close the parenthesis next line.

contracts/sol/ERC1077.sol
```diff
@@ -99,7 +110,7 @@ contract ERC1077 is KeyHolder, IERC1077 {
         address gasToken,
         uint gasLimit,
         OperationType operationType,
```
GrandSchtroumpf: If OperationType is dynamic (struct with at least one dynamic field), it should be flag as memory.

contracts/sol/ERC1077.sol
```diff
@@ -118,7 +129,7 @@ contract ERC1077 is KeyHolder, IERC1077 {
         return messageHash;
     }
 
-    function refund(uint256 gasUsed, uint gasPrice, address gasToken) private {
+    function refund(uint256 gasUsed, uint gasPrice, address gasToken) private requireInit() {
```
GrandSchtroumpf: Need comments

contracts/sol/ERC1077.sol
```diff
@@ -153,4 +164,9 @@ contract ERC1077 is KeyHolder, IERC1077 {
         }
         return true;
     }
+
+    function destroy() public requireInit() {
```
GrandSchtroumpf: Comments

contracts/sol/ERC1077.sol
```diff
@@ -8,13 +10,22 @@ import "../../node_modules/openzeppelin-solidity/contracts/math/SafeMath.sol";
 import "../../node_modules/openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
 
 
-contract ERC1077 is KeyHolder, IERC1077 {
+contract ERC1077 is KeyHolder, IERC1077, Initable {
     using ECDSA for bytes32;
     using SafeMath for uint;
 
+    address payable public recoverAddress;
```
GrandSchtroumpf: Flag it as private and create a getter. Don't assume reader know that this is going to generate a getter.

contracts/sol/Factory2.sol
```diff
@@ -0,0 +1,51 @@
+pragma solidity ^0.5.2;
```
GrandSchtroumpf: Use latest version.

contracts/sol/ERC1077.sol
```diff
@@ -8,13 +10,22 @@ import "../../node_modules/openzeppelin-solidity/contracts/math/SafeMath.sol";
...
 
-    constructor(address _key) KeyHolder(_key) public {
+    /* solium-disable-next-line no-empty-blocks */
+    constructor() KeyHolder(address(0x0)) public {} // this should never be called and is here only to satisfy the solc compiler
+
+    // this function play the role of the "constructor",
+    // it can be called only once, and should be called before any other (non pure/view) functions
+    function init() public requireNotInit() {
+        Initable.init();
+        this.addKey(address(0xdeAD00000000000000000000000000000000dEAd), 1); // ! WARNING CHANGE HARDCODED ADDRESS WITH USER KEY AT DEPLOY
```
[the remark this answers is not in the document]
pldespaigne: tested on Remix : 
* `KeyHolder.addKey` throw
* `addKey()` (without `this`) throw

the other solution is to copy 100 lines of KeyHolder at the top of ERC1077, what do you think ?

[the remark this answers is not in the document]
GrandSchtroumpf: Ok, and `super.addKey()` ? I thing it might make event more sense. And it's easier to understand that the function is defined inside KeyHolder.

contracts/sol/ERC1077.sol
pldespaigne: `enum OperationType {CALL, DELEGATECALL, CREATE}` should enum also be flagged as memory ?
Anyway operation type is definitely important and we have to keep it.

contracts/sol/ERC1077.sol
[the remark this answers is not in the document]
pldespaigne: doesn't work either :/

contracts/sol/ERC1077.sol
```diff
@@ -12,124 +12,135 @@ import "../../node_modules/openzeppelin-solidity/contracts/token/ERC20/ERC20.sol
...
-    constructor() KeyHolder(address(0x0)) public {} // this should never be called and is here only to satisfy the solc compiler
-
-    // this function play the role of the "constructor",
-    // it can be called only once, and should be called before any other (non pure/view) functions
+    using SafeMath for uint256;
+
+    address payable private recoverAddress;
+    uint256 private lastNonce;
+
+    /**
```
GrandSchtroumpf: use natspec : 
```
/// @dev blablabla
/// blablabla
```

PR #487, opened 2019-06-26
Issue#484 simple org contract

pldespaigne:
Only af9a67a needs review, the other commits are contracts build & a merge


contracts/sol/Organization.sol
```diff
@@ -0,0 +1,333 @@
...
+    }
+    function isWhitelisted(address user, uint256 operationId) public view returns(bool) {
+        return adminList[user] || operations[operationId].whitelist[user];
+    }
+
+    // Admin ----------------------
+    function isAdmin(address user) public view returns (bool){
+        return adminList[user];
+    }
+    function getAdminCount() public view returns(uint256) {
```
GrandSchtroumpf: Make it external.

contracts/sol/Organization.sol
```diff
@@ -0,0 +1,333 @@
...
+          operations[operationId].quorum,
+          operations[operationId].active
+        );
+    }
+    function isWhitelisted(address user, uint256 operationId) public view returns(bool) {
+        return adminList[user] || operations[operationId].whitelist[user];
+    }
+
+    // Admin ----------------------
+    function isAdmin(address user) public view returns (bool){
```
GrandSchtroumpf: Make it external

contracts/sol/Organization.sol
```diff
@@ -0,0 +1,333 @@
...
+    function hasApprovedAction(address user, bytes32 hash) public view returns(bool) {
+        return actions[hash].approvals[user];
+    }
+
+    // Operations -----------------
+    function operationExists(uint256 operationId) public view returns(bool){
+        if (bytes(operations[operationId].name).length == 0) return false;
+        return true;
+    }
+    function getOperation(uint256 operationId) public view returns(string memory name, uint256 whitelistLength, uint256 quorum, bool active) {
```
GrandSchtroumpf: Why not returning the struct ? Add `experimental ABIEncoder v2` at the top of your contract.

contracts/sol/Organization.sol
```diff
@@ -0,0 +1,333 @@
...
+        return true;
+    }
+    function getAction(bytes32 hash) public view returns(
+      uint256 operationId,
+      uint256 approvalsCount,
+      bool active,
+      bool exectuted,
+      address to, uint256 value, bytes memory data
+    ) {
+        return (
```
GrandSchtroumpf: Why not return ing a struct ?

contracts/sol/Organization.sol
```diff
@@ -0,0 +1,333 @@
...
+    /// and if the current approval let the action reach its quorum, then the action tx will be triggered
+    function approve(
+        bytes32 hash, uint256 operationId, address payable to, uint256 value, bytes memory data
+    ) public operationShouldExists(operationId) operationIsActive(operationId) onlyAuthorizedUser(operationId, hash) {
+
+        require(to != address(0x0), "You cannot use the 0x0 address as a destination"); // because we check actionExists() on the 0x0 address
+
+        // if action doesn't exists create it
+        if(!actionExists(hash)) {
+            actions[hash].operationId = operationId;
```
GrandSchtroumpf: Use a cast instead. It's more readable. 
```
actions[hash].operationId = Action({
...
})
```

contracts/sol/Organization.sol
pldespaigne: even with `experimental ABIEncoder v2` we cannot return dynamic types, and here the struct contains a `mapping`

pldespaigne: using `ABIEncoderV2` with returns struct will throw `TypeError: Only libraries are allowed to use the mapping type in public or external functions.`

contracts/sol/Organization.sol
pldespaigne: same has [here](https://github.com/blockframes/blockframes/pull/487/files/fd99b9f746163fea6e696d66d98501d8bcabbc4e#r298083299)

PR #493, opened 2019-06-28
prevent double ens register

pldespaigne:


apps/backend-functions/src/relayer.ts
```diff
@@ -114,7 +114,7 @@ export const relayerCreateLogic = async (
 
     const ensWorkFlow = async () => {
       const retreivedAddress = await relayer.wallet.provider.resolveName(fullName);
-      if (retreivedAddress !== getAddress(erc1077address)) { // if name is already link to the good address : skip ensWorkflow
+      if (retreivedAddress !== '0x00000000000000000000000000000000000000') { // if name is already link to a non-zero address : skip ensWorkflow
```
GrandSchtroumpf: Create a const for that `const ADDRESS_ZERO = 0x00000000000000000000000000000000000000`

PR #538, opened 2019-07-02
add dark-theme

Hugo-Hontans:
add dark-theme with global variables


styles/variables.scss
```diff
@@ -1,14 +1,29 @@
 @import '~@angular/material/theming';
+@import './theme.scss';
 
+// Main theme
 $primary: mat-color($blockframes-theme-primary);
 $accent: mat-color($blockframes-theme-accent);
 $warn: mat-color($blockframes-theme-warn);
 
+// Dark-theme
+$dark-primary: mat-color($blockframes-dark-theme-primary);
```
GrandSchtroumpf: Couldn't we have : 
```css
.dark-theme {
$primary: mat-color($blockframes-dark-theme-primary);
$accent: mat-color($blockframes-dark-theme-accent);
$warn: mat-color($blockframes-dark-theme-warn);
}
```

Hugo-Hontans: No, we need to re-define global colors CSS when we change theme with : 
```
.dark-theme {
  @extend :root;
  --primary: #{$dark-primary};
  --warn: #{$dark-warn};
  --accent: #{$dark-accent};
}
```

PR #545, opened 2019-07-03
Issue#540 refactor wallet page

pldespaigne:


libs/ethers/src/lib/key-manager/key-manager-list/key-manager-list.component.scss
```diff
@@ -1,4 +1,9 @@
-
-section {
-  margin-top: 30px;
+h2 {
+  margin-top: 48px;
+  font-weight: 600;
```
GrandSchtroumpf: Can't you use an existing typography rule ? https://material.angular.io/guide/typography
Ask Mathilde about her typography rule (based on doc above) so she can flag the title on need.

libs/ethers/src/lib/wallet/wallet-view/wallet-view.component.html
```diff
@@ -1,46 +1,36 @@
...
+        </div>
+      </header>
+      <div class="container">
+        <h2>My Profile</h2>
         <div class="wallet-profile">
-          <eth-blockie mat-card-avatar size="40" [address]="wallet.address">
-          </eth-blockie>
+
+          <!-- TODO replace by Max's component -->
+          <eth-blockie mat-card-avatar size="40" [address]="wallet.address"></eth-blockie>
```
GrandSchtroumpf: Here `size` will be evaluated has string I think.

libs/ethers/src/lib/wallet/wallet-view/wallet-view.component.scss
```diff
@@ -1,13 +1,43 @@
...
 main {
-  margin-left: 20%;
-  margin-right: 20%;
-  margin-top: 10%;
-  mat-card {
-    margin-top: 30px;
-    margin-bottom: 30px;
-    p {
-      max-width: 70%;
+  background-color: #fafafa;
```
GrandSchtroumpf: Use variable color : `var(--something)`. Ask Hugo about that.

libs/ethers/src/lib/wallet/wallet-view/wallet-view.component.scss
```diff
@@ -1,13 +1,43 @@
...
+  margin-left: 200px;
+  margin-right: 200px;
+  margin-top: 5%;
```
GrandSchtroumpf: Be very carful about % in margin. It breaks 80% of the time with responsive.

libs/ethers/src/lib/wallet/wallet-view/wallet-view.component.scss
```diff
@@ -1,13 +1,43 @@
...
+  header {
+    h1 {
```
GrandSchtroumpf: check material typography.

libs/ethers/src/lib/wallet/wallet-view/wallet-view.component.scss
```diff
@@ -1,13 +1,43 @@
...
+      font-size: 30px;
+      font-weight: 800;
+      letter-spacing: 0.37px;
+    }
+    .spacer {
+      border-bottom: solid #e4e4e4 1px;
+      margin: 0 30px 0 30px;
+    }
+    button {
+      border-radius: 20px !important; // sorry for the !important but there is already an !important on button somewhere else
```
GrandSchtroumpf: Don't change border-radius of button. Use the one provided in the theme by Hugo.

libs/ethers/src/lib/wallet/wallet-view/wallet-view.component.scss
```diff
@@ -1,13 +1,43 @@
...
+      background-color: rgba(26,72,245,0.05);
```
GrandSchtroumpf: Use variable button

libs/ethers/src/lib/wallet/wallet-view/wallet-view.component.scss
```diff
@@ -1,13 +1,43 @@
...
+      margin-left: 16px;
+      font-size: 12px;
+      font-weight: 600;
+      mat-icon {
+        font-size: 18px;
+        font-weight: 600;
+      }
+    }
+  }
+  .container {
```
GrandSchtroumpf: Never use a class names container.

libs/ethers/src/lib/wallet/wallet-view/wallet-view.component.html
pldespaigne: this component will be replaced by Max's anyway

styles/variables.scss
```diff
@@ -18,6 +18,9 @@ $dark-warn: mat-color($blockframes-dark-theme-warn);
   --accent: #{$accent};
   --lightGrey: #4D5664;
   --label: #B5BDCE;
+  --whiteBackground: #FAFAFA;
+  --borderGrey: #E4E4E4;
+  --flatButtonGrey: rgba(26,72,245,0.05);
```
GrandSchtroumpf: Please don't create variable for everything. Keep it as small as possible.
Also, when you add variables you need to set the counter part for darktheme.

pldespaigne: this color are widely used across all the designs that's why we have added them here, also
we will put them in the darktheme as soon as we know the value of their counter part
Or do you want us to put them in the darktheme as is, just to remember to update them later ?

GrandSchtroumpf: Add them with placeholder colors. We don't have any dark theme scheme now, but it'll help us to know how to build it if we have at least the keys.

PR #546, opened 2019-07-03
Issue#529 login signin

Hugo-Hontans:
Implement new design for login-view


libs/auth/src/lib/components/signin-form/signin-form.component.scss
```diff
@@ -0,0 +1,11 @@
+section {
+  margin-top: 50%;
```
GrandSchtroumpf: Be careful with % in margin. Check responsive.

libs/auth/src/lib/components/signin-form/signin-form.component.ts
```diff
@@ -0,0 +1,20 @@
...
+})
+
+export class SigninFormComponent implements OnInit {
+  @Output() opened = new EventEmitter();
+  @Output() submited = new EventEmitter();
+
+  public signinForm: SigninForm;
+
+  ngOnInit() {
+    this.signinForm = new SigninForm();
```
GrandSchtroumpf: I think you can set it directly line 15.

libs/auth/src/lib/components/signup-form/signup-form.component.html
```diff
@@ -0,0 +1,11 @@
+<section testId="signup" fxLayout="column" fxLayoutAlign="start center">
+  <h2>Sign Up</h2>
+  <form [formGroup]="signupForm" (ngSubmit)="submited.emit(signupForm)" fxLayout="column">
+    <input-email [form]="signupForm"></input-email>
```
GrandSchtroumpf: You shouldn't have to pass the form. But only the controlName....

libs/auth/src/lib/components/signup-form/signup-form.component.html
```diff
@@ -0,0 +1,11 @@
...
+    <form-group-password-confirm [form]="signupForm"></form-group-password-confirm>
```
GrandSchtroumpf: I don't really like this component. It's overengineered and not really explicit... I would recommend to have two password input, and use the confirm validator in the SigninForm to display errors.

libs/auth/src/lib/components/signup-form/signup-form.component.scss
```diff
@@ -0,0 +1,11 @@
+section {
+  margin-top: 50%;
```
GrandSchtroumpf: Same here about margin

libs/auth/src/lib/components/signup-form/signup-form.component.ts
```diff
@@ -0,0 +1,20 @@
...
+})
+
+export class SignupFormComponent implements OnInit {
+  @Output() opened = new EventEmitter();
+  @Output() submited = new EventEmitter();
+
+  public signupForm: SignupForm;
+
+  ngOnInit() {
+    this.signupForm = new SignupForm();
```
GrandSchtroumpf: Maybe you can set it line 15

libs/auth/src/lib/pages/login-view/login-view.component.ts
```diff
@@ -45,13 +43,13 @@ export class AuthRootComponent implements OnInit {
     }
```
GrandSchtroumpf: You need to catch if you try.

libs/auth/src/lib/pages/login-view/login-view.component.ts
```diff
@@ -60,4 +58,20 @@ export class AuthRootComponent implements OnInit {
       this.snackBar.open(err.message, 'close', { duration: this.snackbarDuration });
     }
```
GrandSchtroumpf: Same

libs/auth/src/lib/pages/login-view/login-view.component.html
```diff
@@ -0,0 +1,16 @@
+<mat-sidenav-container>
+  <mat-sidenav-content [fxLayoutAlign]="align">
+    <button class="mat-elevation-z0" mat-icon-button mat-fab *ngIf="isSignin" (click)="toSignup()">
+      <mat-icon class="rotate" svgIcon="arrow"></mat-icon>
+    </button>
+    <button class="mat-elevation-z0" mat-icon-button mat-fab *ngIf="!isSignin" (click)="toSignin()">
+      <mat-icon svgIcon="arrow"></mat-icon>
+    </button>
+  </mat-sidenav-content>
+  <mat-sidenav #signinSidenav [opened]="isSignin" mode="side">
```
GrandSchtroumpf: Put this one above the `mat-sidenav-content` for better readibility

libs/auth/src/lib/pages/login-view/login-view.component.ts
```diff
@@ -60,4 +58,20 @@ export class AuthRootComponent implements OnInit {
...
   }
+
+  public toSignup() {
+    this.signupSidenav.open();
```
GrandSchtroumpf: What about having the sidenav `opened` value react to the `isSignin` boolean ?  Like that you don't need to right ".open()" and ".close()".
```
<mat-sidenav [opened]="isSignin">...</..>
<mat-sidenav [opened]="!isSignin">...</..>
```

libs/auth/src/lib/pages/login-view/login-view.component.ts
Hugo-Hontans: I don't understand, the catch is line 40-43

libs/auth/src/lib/pages/login-view/login-view.component.ts
Hugo-Hontans: Same, the catch is line 56-58

libs/auth/src/lib/components/signup-form/signup-form.component.html
Hugo-Hontans: I need to finish other PR for that, I will work on that

libs/auth/src/lib/components/signup-form/signup-form.component.html
Hugo-Hontans: I don't think it's over-engineered because we will use this component in auth, wallet and delivery...
It can simplify the implementation to call this component directly

Hugo-Hontans: We will change this component in another PR to manage it as a single control

libs/auth/src/lib/components/signup-form/signup-form.component.html
Hugo-Hontans: We agree to remove these custom form-field ?

libs/auth/src/lib/components/signin-form/signin-form.component.html
```diff
@@ -0,0 +1,11 @@
+<section testId="signin" fxLayout="column" fxLayoutAlign="center center">
+  <h2>Sign in</h2>
+  <form [formGroup]="signinForm" (ngSubmit)="submited.emit(signinForm)" fxLayout="column">
+    <email-form-field [form]="signinForm"></email-form-field>
```
GrandSchtroumpf: So if I understood well, this will be removed later, right ?

libs/auth/src/lib/pages/login-view/login-view.component.scss
```diff
@@ -0,0 +1,63 @@
...
+}
+
+.rotate {
+  transform: rotate(180deg);
+}
+
+.mat-fab {
+  height: 120px;
+  width: 120px;
+  background-color: rgba(0,0,0,0.3);
```
GrandSchtroumpf: Please use `var(--...)` here.

libs/auth/src/lib/pages/login-view/login-view.component.ts
GrandSchtroumpf: Ok, sorry, I didn't see them.

libs/auth/src/lib/pages/login-view/login-view.component.ts
GrandSchtroumpf: ok sorry

libs/auth/src/lib/pages/login-view/login-view.component.ts
```diff
@@ -60,4 +58,16 @@ export class AuthRootComponent implements OnInit {
       this.snackBar.open(err.message, 'close', { duration: this.snackbarDuration });
     }
   }
+
+  public toSignup() {
+    this.isSignin = false;
```
GrandSchtroumpf: This could be implemented directly on the html now, because this is only one line change.

libs/auth/src/lib/components/signin-form/signin-form.component.html
Hugo-Hontans: Yes, maybe we can talk about that to Bruce, because if we delete these elements we must manage mat-errors, mat-form-field directly in the form

libs/auth/src/lib/pages/login-view/login-view.component.ts
Hugo-Hontans: Yes I did that

PR #548, opened 2019-07-03
User story#503

ioaNikas:
- [x] Add first firestore rules
- [x] New org members are added as Admins
- [x] An organization can now have multiple SuperAdmins


firestore.rules
```diff
@@ -1,7 +1,129 @@
...
+    	return isSignedIn() && currentUser().uid in
+    		get(/databases/$(database)/documents/rights/$(orgId)).data.admins
+    }
+
+    function getOrgsDocumentRights(docId) {
+    	return get(/databases/$(database)/documents/rights/$(getUserOrgId())/orgDocsRights/$(docId)).data
     }
   }
 }
+
```
lsenta: Non

firestore.rules
```diff
@@ -1,7 +1,129 @@
 service cloud.firestore {
   match /databases/{database}/documents {
-    match /{document=**} {
-      allow read, write: if request.auth.uid != null;
+
+    /// USERS RULES ///
+
+    match /users/{userId} {
+    	allow read: if isSignedIn();
```
lsenta: Any logged user can retrieve all emails, profiles, and orgs. Please add a TODO and make a note about this somewhere if it's not easy to solve (cross-organization user profiles might be tricky)

firestore.rules
```diff
@@ -1,7 +1,129 @@
...
+      allow update: if isSuperAdmin(getUserOrgId()); // Necessary to add users to an org. TODO : remove
```
lsenta: Can I update my own profile?
Make sure the orgid and other important, immutable field are not modified in before / after request.

firestore.rules
```diff
@@ -1,7 +1,129 @@
...
+    	allow write: if currentUser().uid == incomingData().uid
+        || currentUser().uid == existingData().uid
```
lsenta: do not  use write & update, they overlap and your rules become hard to think about (there's probably a bug),
https://firebase.google.com/docs/firestore/security/rules-structure#granular_operations

`write` = `create`, `update`, `delete`.

firestore.rules
lsenta: See my comment below, superadmin can update a user but not create or delete it in firestore, is that ok? it doesn't match the comment.

firestore.rules
lsenta: I can change my orgId, email, etc stored in the firestore document, risky

firestore.rules
```diff
@@ -1,7 +1,129 @@
...
+    /// NOTIFICATIONS RULES ///
+
+    match /notifications/{notificationId} {
+    	allow read: if currentUser().uid == existingData().userId
+    }
+
+    /// ORGANIZATION RULES ///
+
+    match /orgs/{orgId} {
+    	allow create, read: if isSignedIn()
```
lsenta: combined with the previous rule on user docs, looks like I can list all blockframes customers company and find all their admin's email.

firestore.rules
```diff
@@ -1,7 +1,129 @@
...
+    	allow write: if isSuperAdmin(orgId)
```
lsenta: confusing, `write = create, update, delete`

firestore.rules
ioaNikas: I don't get it ?

libs/organization/src/lib/rights/+state/rights.model.ts
```diff
@@ -48,7 +48,7 @@ export enum App  {
 
 export function initializeOrgRights(params: Partial<OrganizationRights>) {
```
GrandSchtroumpf: Call this function `createRights(params: Partial<OrganizationRights> = {})`. Like that you can create a right by doing : `createRights()` without parameters.

LICENSE
```diff
@@ -4,7 +4,7 @@ Copyright (c) 2019 blockframes
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
+in the Software without restriction, including without limitation the permissions
```
lsenta: Don't touch the LICENSE

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -3,36 +3,44 @@ import { createOrganization, Organization, OrgMember } from './organization.mode
...
 
   /** Add a new user to the organization */
   public addMember(member: OrgMember) {
     const orgId = this.query.getValue().org.id;
+    const permissions = this.permissionsQuery.getValue();
     const orgDoc = this.db.doc(`orgs/${orgId}`);
+    const permissionsDoc = this.db.doc(`permissions/${orgId}`);
     const userDoc = this.db.doc(`users/${member.id}`);
 
     this.db.firestore
```
lsenta: use an async function and await here, this'll prepare the code for the same concurrency issues you've had yesterday

lsenta: or return the promise created by runTransaction if you don't need to run code after the transaction.

libs/organization/src/lib/permissions/+state/permissions.model.ts
```diff
@@ -46,40 +46,40 @@ export enum App  {
...
   return {
-    superAdmin: params.superAdmin,
+    superAdmins: [params.superAdmins],
     canCreate: [],
     canRead: [],
     canUpdate: [],
     canDelete: [],
     ...params
-  } as OrganizationRights;
+  } as Permissions;
```
lsenta: When you use `as` you lose all the type safety tools that typescript provides you. Provide default values and remove the `as` if possible.
Same for every other `createX` functions.

nx.json
```diff
@@ -62,7 +62,7 @@
     "backend-ops": {
       "tags": []
     },
-    "libs-rights": {
+    "libs-permissions": {
```
lsenta: no change to `angular.json` or `tsconfig.json`?

firestore.rules
```diff
@@ -1,7 +1,131 @@
...
-      allow read, write: if request.auth.uid != null;
+
+    /// USERS RULES ///
+
+    match /users/{userId} {
+    	allow read, update: if isSignedIn(); // Secure it as soon as we got Inivitations collection
+    	allow create: if currentUser().uid == existingData().uid
+				|| currentUser().uid == incomingData().uid
+      allow update, delete: if currentUser().uid == existingData().uid
+      	&& existingData() == incomingData()
```
lsenta: Ambiguous, if you `delete` these rules will always be false. just disallow delete.

What's firestore's behavior when you have overlapping rules in the same `match`?

For multiple match: if any rule is `true` then the whole access is true. If that's the case within a match you have a security issue: the first `allow update` will always be true if I'm signed in and I can change my uid.

firestore.rules
```diff
@@ -1,7 +1,131 @@
...
+    /// NOTIFICATIONS RULES ///
+
+    match /notifications/{notificationId} {
+    	allow read: if currentUser().uid == existingData().userId
+    }
+
+    /// ORGANIZATION RULES ///
+
+    match /orgs/{orgId} {
+    	allow create: if isSignedIn()
```
lsenta: `and currentUser.orgID is empty`?

firestore.rules
```diff
@@ -1,7 +1,131 @@
...
+
+    /// PERMISSIONS RULES ///
+
+    match /permissions/{orgId} {
+    	allow create: if isSignedIn()
+    	allow read, update, delete: if isSuperAdmin(orgId)
+        || isOrgAdmin(orgId)
+
+      match /{documents=**} {
+      	allow read, create: if isSignedIn()
```
lsenta: Any user can create permission for any orgs.

firestore.rules
```diff
@@ -1,7 +1,131 @@
...
+      }
+    }
+
+    /// MOVIES RULES ///
+
+    match /movies/{movieId} {
+    	allow create: if isSuperAdmin(getUserOrgId())
+        || isOrgAdmin(getUserOrgId())
+    	allow read, update, delete: if isSignedIn()
+        && getOrgsDocumentPermissions(movieId).owner == getUserOrgId()
```
lsenta: cross organization would require another check in the permission collection of the org.

firestore.rules
```diff
@@ -1,7 +1,131 @@
...
+    match /deliveries/{deliveryId} {
+    	allow create, list: if isSuperAdmin(getUserOrgId())
+        || isOrgAdmin(getUserOrgId())
+    	allow read, update, delete: if isSignedIn()
+        && getOrgsDocumentPermissions(deliveryId).owner == getUserOrgId()
+
+      match /{documents=**} {
+        allow create: if isSuperAdmin(getUserOrgId())
+          || isOrgAdmin(getUserOrgId())
+        allow read, update, delete: if isSignedIn()
```
lsenta: Any user can edit deliveries.

nx.json
ioaNikas: Nope, permissions are not a library anymore, we changed it to an organization subfolder as they are closely related to an org

firestore.rules
```diff
@@ -1,7 +1,131 @@
...
+
+    match /notifications/{notificationId} {
+    	allow read: if currentUser().uid == existingData().userId
+    }
+
+    /// ORGANIZATION RULES ///
+
+    match /orgs/{orgId} {
+    	allow create: if isSignedIn()
+      allow get: if isSuperAdmin(orgId) || isOrgAdmin(orgId)
```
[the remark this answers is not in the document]
ioaNikas: For that first part of the user story, we decided that each new members of an org are Admins. We plan to use more roles later (when I'll be more confortable with our data model and firestore rules)

firestore.rules
ioaNikas: As we saw together, we let permissions free for now and maybe only allow firebase functions to write on it ?

firestore.rules
```diff
@@ -1,7 +1,163 @@
...
   match /databases/{database}/documents {
-    match /{document=**} {
-      allow read, write: if request.auth.uid != null;
+
+
+    /// USERS RULES ///
+
+    match /users/{userId} {
+    	allow read, update: if isSignedIn(); // Secure it as soon as we got Inivitations collection
+    	allow create: if currentUser().uid == existingData().uid
```
lsenta: Do we create the user in the frontend? As far as I know, the backend function is solely responsible for that -> its in admin mode -> we can disallow create and simplify this code. (you check before / after data, that looks broken)

firestore.rules
```diff
@@ -1,7 +1,163 @@
...
 service cloud.firestore {
   match /databases/{database}/documents {
-    match /{document=**} {
-      allow read, write: if request.auth.uid != null;
+
+
+    /// USERS RULES ///
+
+    match /users/{userId} {
+    	allow read, update: if isSignedIn(); // Secure it as soon as we got Inivitations collection
```
lsenta: `Invitations`

firestore.rules
```diff
@@ -1,7 +1,163 @@
...
+
+    match /notifications/{notificationId} {
+    	allow read: if currentUser().uid == existingData().userId;
+    }
+
+    /// ORGANIZATION RULES ///
+
+    match /orgs/{orgId} {
+    	allow create: if isSignedIn() && userOrgId() == null;
+      allow read, update, delete: if isSuperAdmin(userId(), orgId)
```
lsenta: simplify if the precedence fix worked.

firestore.rules
```diff
@@ -1,7 +1,163 @@
...
+  	}
+
+  	function isSuperAdmin(userId, orgId) {
+    	return userId in getOrgPermissions(orgId).superAdmins;
+  	}
+
+		// TODO: Figure out why this always return false
+    // function isOrgAdmin(userId, orgId) {
+    // 	return userId in getOrgPermissions(orgId).admins || isSuperAdmin(userId, orgId);
+    // }
```
lsenta: Remove

firestore.rules
```diff
@@ -1,7 +1,163 @@
...
+    function userCan(x, userId, orgId, docId) {
+      return isOrgAdmin(userId, orgId)
+      	|| isSuperAdmin(userId, orgId)
+        || userId in getOrgPermissions(orgId)[x]
+        || userId in getUserDocumentPermissions(orgId, docId).admins
+        || userId in getUserDocumentPermissions(orgId, docId)[x];
     }
   }
 }
+
```
lsenta: Non

firestore.rules
```diff
@@ -1,7 +1,163 @@
...
+/////////////////////
+// DEPLOYING RULES //
+/////////////////////
+//
+// firebase init firestore
+// firebase deploy --only firestore:rules
+//
+// OR
+//
+// Copy and paste these rules into the rules tab of your project database
```
lsenta: Keep deploy guidelines in the `README.md` not in the code

firestore.rules
```diff
@@ -1,7 +1,163 @@
...
+      return get(/databases/$(database)/documents/permissions/$(orgId)/userDocsPermissions/$(docId)).data;
+    }
+
+    function orgCan(x, orgId, docId) {
+      return getOrgDocumentPermissions(orgId, docId).owner == orgId
+        || getOrgDocumentPermissions(orgId, docId).admin == true
+        || getOrgDocumentPermissions(orgId, docId)[x] == true;
+    }
+
+    function userCan(x, userId, orgId, docId) {
```
lsenta: rename `x` & comment the expected value

firestore.rules
GrandSchtroumpf: Why not creating user on the frontend ? Why would you disallow create ?

lsenta: As far as I know the backend is solely responsible for that, we use a trigger on authentication to create the corresponding user object. We can disallow create to make the code simpler, this check already appears to be broken.

GrandSchtroumpf: Well th backend is not solely responsible for that right now. Here is the [code that create user](https://github.com/blockframes/blockframes/blob/master/libs/auth/src/lib/%2Bstate/auth.service.ts#L72) on frontend. I'm fine for listen on authentication event if you can, and keep this backend only though. But we need to remove the frontend functions then.

ioaNikas: I'll just keep `allow create: if userId() == incomingData().uid;` otherwise I get blocked when signing up.

firestore.rules
```diff
@@ -1,7 +1,149 @@
...
   match /databases/{database}/documents {
-    match /{document=**} {
-      allow read, write: if request.auth.uid != null;
+
+
+    /// USERS RULES ///
+
+    match /users/{userId} {
+    	allow read, update: if isSignedIn(); // TODO: Secure it as soon as we got Inivitations collection
+    	allow create: if userId() == incomingData().uid;
```
lsenta: Check `userId`

lsenta:
LGTM, please add a note regarding the validation of `/some-document/ID` and the content of the document: `{id: ID_SHOULD_BE_EQUAL_TO_PATH}`

PR #550, opened 2019-07-04
Issue#529 welcome page

Hugo-Hontans:
Create the welcome page


libs/auth/src/lib/pages/welcome-view/welcome-view.component.html
```diff
@@ -0,0 +1,19 @@
...
+    fxLayoutAlign.lt-md="start center"
+    fxLayoutAlign="start start"
+  >
+    <h1>Welcome to Blockframes</h1>
+    <p>
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec luctus dictum metus quis
+      sagittis. Curabitur scelerisque neque erat, ac bibendum nisi aliquet et. Duis a auctor nulla,
+      a finibus mauris. Nullam eu maximus lorem.
+    </p>
+    <a routerLink="/auth/connexion" routerLinkActive="router-link-active">
```
GrandSchtroumpf: Where are you using : router-link-active class ?

libs/auth/src/lib/pages/welcome-view/welcome-view.component.scss
```diff
@@ -0,0 +1,56 @@
+section {
+  height: 100%;
+  background-color: var(--whiteBackground);
+}
+
+h1 {
+  width: 308px;
```
GrandSchtroumpf: Why 308px?

libs/auth/src/lib/pages/welcome-view/welcome-view.component.html
Hugo-Hontans: It was my auto-complete....

libs/auth/src/lib/pages/welcome-view/welcome-view.component.scss
Hugo-Hontans: to wrap the title to the line, maybe it's not a good practice

GrandSchtroumpf: Yes, try to keep things simple and not "pixel perfect" for normal CSS. If it's inside the UI Kit, then you can have very specific values. But for that, if you do that it'll break on responsive.

PR #556, opened 2019-07-05
refactor & clean

pldespaigne:


libs/ethers/src/lib/key-manager/+state/key-manager.service.ts
```diff
@@ -44,14 +41,7 @@ export class KeyManagerService {
...
   /**  create / encrypt / store / from random */
-  async createFromRandom(ensDomain: string, password?: string) { // at signup password is already provided and we don't want to ask again
-    if (!password) {
-      const ref = this.dialog.open(CreatePasswordComponent, { width: '320px' });
-      password = await ref.afterClosed().toPromise();
-      if (!password) {
-        throw new Error('No password provided');
-      }
-    }
+  async createFromRandom(ensDomain: string, password: string) { // at signup password is already provided and we don't want to ask again
```
GrandSchtroumpf: Add comment on top of the function for documentation. Put it under the name of the function for explanation. Try to keep as few explanation comments as possible. Your code should be self-explained.

libs/ethers/src/lib/key-manager/+state/key-manager.service.ts
```diff
@@ -103,20 +90,14 @@ export class KeyManagerService {
...
   async deleteKey(key: Key) {
-    const ref = this.dialog.open(AskPasswordComponent, { width: '320px' });
-    const password = await ref.afterClosed().toPromise();
-    if (!password) {
-      throw new Error('No password provided');
-    }
     this.store.remove(key.address);
   }
 
   /** export key */
```
GrandSchtroumpf: What does it means to export a key ? From where, to where, in which format ????

libs/ethers/src/lib/key-manager/+state/key-manager.service.ts
```diff
@@ -103,20 +90,14 @@ export class KeyManagerService {
 
   /** delete key */
```
GrandSchtroumpf: From where ?

PR #558, opened 2019-07-08
ops: auto deploy with configuration functions

lsenta:


firebase.json
```diff
@@ -27,8 +27,10 @@
   },
   "functions": {
     "predeploy": [
+      "cp ./package.json dist/apps/backend-functions/package.json",
```
GrandSchtroumpf: Will it always runs inside a Unix environment ?

PR #560, opened 2019-07-08
Issue#507

ioaNikas:
- [x] Better naming/sorting of organization folders and files to fit project conventions
- [x] SuperAdmins can change member roles to SuperAdmin or Admin


apps/backend-functions/src/users.ts
```diff
@@ -44,7 +44,7 @@ const findUserByMail = async (data: any, context: CallableContext): Promise<User
         return [];
       }
 
-      return q.docs.map(d => ({ id: d.id, email: d.data().email }));
+      return q.docs.map(d => ({ uid: d.id, email: d.data().email }));
```
GrandSchtroumpf: What is "q", and what is "d" ?

libs/organization/src/lib/components/member-form/member-form.component.html
```diff
@@ -0,0 +1,22 @@
...
+          type="email" placeholder="email" aria-label="email"
+          formControlName="user"
+          [matAutocomplete]="autoEmail"/>
+      </mat-form-field>
+      <mat-autocomplete #autoEmail="matAutocomplete" [displayWith]="displayFn">
+        <mat-option *ngFor="let option of mailsOptions" [value]="option">
+          {{ option.email }}
+        </mat-option>
+      </mat-autocomplete>
+      <button testId="addMember" mat-raised-button color="primary" (click)="submit()">Add</button>
```
GrandSchtroumpf: use `(submit)` on the form instead of `(click)`

libs/organization/src/lib/components/member-form/member-form.component.html
```diff
@@ -0,0 +1,22 @@
+<mat-card fxLayout="column">
```
GrandSchtroumpf: What is the module of this component ? There should be a module `OrganizationMemberModule`.

libs/organization/src/lib/components/member-form/member-form.component.ts
```diff
@@ -82,11 +71,10 @@ export class OrgMembersShowComponent implements OnInit {
   private async onChange() {
     this.addMemberForm.valueChanges.subscribe(x => {
       // TODO: debounce
-      this.listUserByMail(x.user)
-        .then(xs => {
-          // TODO: use an observable
-          this.mailsOptions = xs;
-        });
+      this.listUserByMail(x.user).then(xs => {
```
GrandSchtroumpf: What is "x" and "xs"

libs/organization/src/lib/pages/member-list/member-list.component.html
```diff
@@ -0,0 +1,7 @@
+<section>
+  <div *ngIf="(org$ | async).members as members">
```
GrandSchtroumpf: use `ng-container`

firestore.rules
```diff
@@ -115,6 +115,11 @@ service cloud.firestore {
     	return userId in getOrgPermissions(orgId).superAdmins;
   	}
 
+		// TODO: Figure out why this always return false
```
GrandSchtroumpf: Add an issue for that, and reference the id of the issue in the TODO comment

libs/organization/src/lib/pages/member-view/member-view.component.html
```diff
@@ -0,0 +1,19 @@
...
+  <mat-card-header>
+    <mat-card-title> {{ member.email }} </mat-card-title>
+    <mat-card-subtitle> {{ member.uid }} </mat-card-subtitle>
+  </mat-card-header>
+  <mat-card-actions
+    *ngIf="!!(isSuperAdmin$ | async)"
+    fxLayout="column"
+    fxLayoutAlign="center center"
+  >
+    <a *ngIf="!!(isUserSuperAdmin$ | async)" mat-icon-button color="primary" (click)="switchRoles()"
```
GrandSchtroumpf: Format those lines

libs/organization/src/lib/pages/org-view/org-view.component.html
```diff
@@ -0,0 +1,12 @@
+<section>
+  <div *ngIf="(org$ | async) as org">
```
GrandSchtroumpf: use `ng-container`

apps/backend-functions/src/users.ts
ioaNikas: Renamed it into something more explicit (query and doc)

libs/organization/src/lib/components/member-form/member-form.component.html
ioaNikas: Do we really need it as organizations members are part of the organization and are also displayed via the organization state ?

Do you mean we need it for further implementations like members profile pages etc ?

ioaNikas: After some brainstorming with Hugo, we found that we will never have a use of an `OrganizationMemberModule` without calling/importing the `OrganizationModule` aswell.

As `OrganizationModule` size is still pretty reasonable, I prefer to merge all the organizations members logic in it for now but still open to a module for members if we tend to make more organization members features such as complete profile pages.

firestore.rules
ioaNikas: Sorry I already figured out why this wasn't working.

I take note of the good practice on TODO / Issues for next times, thanks !

libs/organization/src/lib/components/member-form/member-form.component.ts
```diff
@@ -65,15 +65,15 @@ export class MemberFormComponent implements OnInit {
 
   private async listUserByMail(prefix: string): Promise<User[]> {
     const f = firebase.functions().httpsCallable('findUserByMail');
-    return f({ prefix }).then(x => x.data);
+    return f({ prefix }).then(usersProposal => usersProposal.data);
   }
 
   private async onChange() {
-    this.addMemberForm.valueChanges.subscribe(x => {
+    this.addMemberForm.valueChanges.subscribe(userObject => {
```
GrandSchtroumpf: when do you unsubscribe ?

libs/organization/src/lib/components/org-form/org-form.component.html
```diff
@@ -1,33 +1,33 @@
 <section>
   <mat-card>
-    <mat-card-header>
-      <mat-card-title>Principal Informations</mat-card-title>
-    </mat-card-header>
+    <mat-card-header> <mat-card-title>Principal Informations</mat-card-title> </mat-card-header>
```
GrandSchtroumpf: Previous format was prefered.

PR #567, opened 2019-07-09
fix mnemonic import

pldespaigne:
- [x] fix the bug (the component was emitting an empty string instead of the mnemonic)
- [x] refactoring to use proper form structure


libs/ethers/src/lib/wallet/wallet-import-key-form/wallet-import-key-form.component.html
```diff
@@ -13,17 +13,23 @@ <h2>Import your Key</h2>
     </div>
     <ng-container *ngIf="!isMnemonic; else mnemonicImport">
       <file-upload [accept]="['.json']" [types]="['application/json']" (uploaded)="importedJson($event)"></file-upload>
+      <img src="/assets/images/import_key_pic.png">
+      <div>
```
GrandSchtroumpf: why do you need a div here ?

libs/ethers/src/lib/wallet/wallet-import-key-form/wallet-import-key-form.component.ts
```diff
@@ -11,11 +11,14 @@ import { FormControl, FormGroup } from "@angular/forms";
...
 export class WalletImportKeyFormComponent {
 
+  fileUploaded = false;
   isMnemonic = false;
   keyObject: Key;
-  mnemonic = '';
-  mnemonicField = new FormControl(''); // TODO use Max's custom mnemonic control
   
+  mnemonicForm = new FormGroup({
+    mnemonicField: new FormControl('', [Validators.required]), // TODO use Max's custom mnemonic control
```
GrandSchtroumpf: If you have only one field, you should use the `mnemonic = new FormControl('', [Validators.required])`

libs/ethers/src/lib/wallet/wallet-import-key-form/wallet-import-key-form.component.html
```diff
@@ -13,17 +13,23 @@ <h2>Import your Key</h2>
...
 
-      <mat-form-field>
-        <textarea matInput placeholder="Enter or paste your mnemonic" [formControl]="mnemonicField"></textarea>
-      </mat-form-field>
+      <form [formGroup]="mnemonicForm" (ngSubmit)="test()">
+        <mat-form-field>
+          <textarea matInput placeholder="Enter or paste your mnemonic" formControlName="mnemonicField"></textarea>
+        </mat-form-field>
+        <img src="/assets/images/import_key_pic.png">
+        <div>
```
GrandSchtroumpf: Same question here.

libs/ethers/src/lib/wallet/wallet-import-key-form/wallet-import-key-form.component.html
```diff
@@ -15,15 +15,16 @@ <h2>Import your Key</h2>
...
       <mat-form-field>
         <textarea matInput placeholder="Enter or paste your mnemonic" [formControl]="mnemonicField"></textarea>
       </mat-form-field>
-
     </ng-template>
+    <img src="/assets/images/import_key_pic.png">
+    <button
+      mat-raised-button
+      color="primary"
+      [disabled]="(!isMnemonic && !fileUploaded) || (isMnemonic && !mnemonicField.valid)"
```
GrandSchtroumpf: Create a getter in the component instead of doing that in the html

libs/ethers/src/lib/wallet/wallet-import-key-form/wallet-import-key-form.component.ts
```diff
@@ -11,11 +11,13 @@ import { FormControl, FormGroup } from "@angular/forms";
 })
 export class WalletImportKeyFormComponent {
 
+  fileUploaded = false;
   isMnemonic = false;
   keyObject: Key;
-  mnemonic = '';
-  mnemonicField = new FormControl(''); // TODO use Max's custom mnemonic control
   
+  mnemonicField = new FormControl('', [Validators.required]); // TODO use Max's custom mnemonic control
```
GrandSchtroumpf: Let's call it mnemonic.

PR #610, opened 2019-07-10
Add xlsx package & excel utils

phpgeekfr:


libs/utils/src/index.ts
```diff
@@ -8,3 +8,4 @@ export * from './lib/form/controls';
 export * from './lib/form/control-list';
 export * from './lib/form/validators/validators';
 export * from './lib/form/matchers';
+export * from './lib/excel.utils';
```
GrandSchtroumpf: Create a folder `spreadsheet` instead of file excel.utils.

libs/utils/src/lib/excel.utils.ts
```diff
@@ -0,0 +1,71 @@
...
+type AOA = any[][];
+
+export interface SheetTab {
+  name: string;
+  index: number;
+  headers: any[];
+  rows: any[][];
+}
+
+export class ExcelUtils {
```
GrandSchtroumpf: Don't use `Excel` in the code. Use Spreadsheet

libs/utils/src/lib/excel.utils.ts
```diff
@@ -0,0 +1,71 @@
+import * as XLSX from 'xlsx';
```
GrandSchtroumpf: Import only what you need : `import { WorkBook, WorkSheet, ... } from 'xlsx'` to take advantage of treeshaking

libs/utils/src/lib/excel.utils.ts
```diff
@@ -0,0 +1,71 @@
...
+export class ExcelUtils {
+  private tabs : SheetTab[] = [];
+  private fileName = 'SheetJS.xlsx';
+
+  public import (bytes: Uint8Array) : void {
+    this.tabs = [];
+
+    // convert Uint8Array to binary String
+    let bstr = "";
+    for (var i = 0; i < bytes.byteLength; i++) {
```
GrandSchtroumpf: Use let, instead of var.

libs/utils/src/lib/excel.utils.ts
```diff
@@ -0,0 +1,71 @@
...
+      bstr += String.fromCharCode(bytes[i]);
+    }
+
+    const wb: XLSX.WorkBook = XLSX.read(bstr, { type: 'binary' });
+
+    // For each tab
+    wb.SheetNames.forEach( (name, index)  => {
```
GrandSchtroumpf: I would recommend to use map instead: 
```typescript
const workBook =  read(bstr, { type: 'binary' });
const rows = utils.sheet_to_json(ws, { header: XXX }) as AOA;
this.tabs = workBook.SheetNames.map((name, index)  => ({
        name,
        index,
        headers: wb.Sheets[name],
        rows: rows[index]
}));
```

libs/utils/src/lib/excel.utils.ts
```diff
@@ -0,0 +1,71 @@
...
+  public import (bytes: Uint8Array) : void {
+    this.tabs = [];
+
+    // convert Uint8Array to binary String
+    let bstr = "";
+    for (var i = 0; i < bytes.byteLength; i++) {
+      bstr += String.fromCharCode(bytes[i]);
+    }
+
+    const wb: XLSX.WorkBook = XLSX.read(bstr, { type: 'binary' });
```
GrandSchtroumpf: call it "workBook" for better readibility

PR #613, opened 2019-07-10
Spreadsheet import movie bis

phpgeekfr:


libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -2,7 +2,19 @@ import { Organization } from "@blockframes/organization";
 import { Material } from "@blockframes/material";
 import { Stakeholder } from "../../stakeholder/+state";
 
+export interface MovieAvailability {
```
GrandSchtroumpf: shouldn't this be a subfield of a movie ?

libs/movie/src/lib/movie/+state/movie.query.ts
```diff
@@ -13,4 +13,23 @@ export class MovieQuery extends QueryEntity<MovieState, Movie> {
...
 
+
+  /**
+   * @dev this method will be changed since we now have an internal movie reference 
+   * Checks if a movie with given title, productionYear and directorName alreadyExists
+   * @param originalTitle 
+   * @param productionYear 
+   * @param directorName 
+   */
+  public movieExists(originalTitle : string, productionYear : number, directorName : string) : string {
```
GrandSchtroumpf: Wouldn't it make more sense to have a generic search based on an object ?

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -30,12 +30,13 @@ export class MovieService {
...
 
-  public update(id: string, movie: any) {
+  public update(id: string, movie: any) : Promise<boolean>{
     // we don't want to keep orgId in our Movie object
     if (movie.org) delete movie.org;
     if (movie.stakeholders) delete movie.stakeholders;
+    if (movie.errors) delete movie.errors;
 
-    this.db.doc<Movie>(`movies/${id}`).update(movie);
+    return this.db.doc<Movie>(`movies/${id}`).update(movie).then(_ => true);
```
GrandSchtroumpf: I don't understand why you need a boolean. Just return `this.db.doc<Movie>(`movies/${id}`).update(movie)`. Then if it doesn't throw, it means that it worked.

libs/movie/src/lib/movie/movie.shared.module.ts
```diff
@@ -0,0 +1,25 @@
...
+  ],
+  imports: [
+    CommonModule,
+    MatCardModule,
+    MatDividerModule,
+    MatGridListModule,
+  ],
+  exports: [MovieViewComponent],
+})
+export class MovieSharedModule {}
```
GrandSchtroumpf: I don't like the name of the module. But I don't have a better idea XD

libs/movie/src/lib/movie/+state/movie.query.ts
phpgeekfr: As said in comments ` @dev this method will be changed since we now have an internal movie reference ` This method will disappear

libs/movie/src/lib/movie/+state/movie.model.ts
phpgeekfr: It is. Check checked with Vincent, MovieAvailability will be MovieSales and fields are going to change (we will have to check together for the database model)

libs/movie/src/lib/movie/+state/movie.service.ts
phpgeekfr: OK no problem, my main need was that this method must return a promise (void or boolean I don't care)

phpgeekfr: updated in commit:465cf99eb6e846723daf6167fe1a4596163b887f

PR #614, opened 2019-07-10
Issue#600: Introduce sentry for error tracking

### LINKED ISSUE #600: Sentry
Add Sentry to the project

- [x] Add simple SentryModule
- [x] Configure with env & user details
- [ ] Figure out where to store it: maybe in an `analytics` library
- [x] Use in every app + add app info in the event info
- [x] add sentry to firebase functions

lsenta:
- [x] Add simple SentryModule
- [x] Configure with env & user details
- [x] Figure out where to store it: maybe in an `analytics` library -> stored in utils to make it part of every app easily.
- [x] Use in every app + add app info in the event info
- [x] add sentry to firebase functions

Fixes #600


libs/utils/src/lib/analytics.module.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  handleError(error) {
+    const eventId = Sentry.captureException(error.originalError || error);
+    const { uid, email } = this.authQuery.user || { uid: undefined, email: undefined };
+    Sentry.showReportDialog({ eventId, user: { email } });
+  }
+}
+
+// Init and add the Sentry ErrorHandler if and only if sentry is defined in env.
+if (sentryDsn) {
```
GrandSchtroumpf: Better not import the module at all than verify than push a method with no provider.
In `app.module`, do : 
```typescript
import: [ sentryDsn ? AnalyticsModule : [] ]
```

libs/utils/src/lib/analytics.module.ts
```diff
@@ -0,0 +1,45 @@
+import { ErrorHandler, Injectable, NgModule } from '@angular/core';
+import { sentryDsn } from '@env';
+import * as Sentry from '@sentry/browser';
+import { AuthQuery } from '@blockframes/auth/+state/auth.query';
+
+// Analytics
+const providers: any[] = [];
+
+@Injectable()
+export class SentryErrorHandler implements ErrorHandler {
```
GrandSchtroumpf: Don't listen to every changes of the user 
```typescript
export class SentryErrorHandler implements ErrorHandler  {
  constructor(authQuery: AuthQuery) {
    authQuery.user$.pipe(
      filter(user => !!user), // If you don't want to catch logout
      // distinctUntilChanged((prev, curr) => /* Check if email, uid, name or username changed */)
    ).subscribe(user => {
      
    })
  }
}
```

libs/utils/src/lib/analytics.module.ts
```diff
@@ -0,0 +1,45 @@
...
+  Sentry.init({
+    dsn: sentryDsn
+  });
+  providers.push({ provide: ErrorHandler, useClass: SentryErrorHandler });
+}
+
+@NgModule({
+  providers
+})
+export class AnalyticsModule {}
```
GrandSchtroumpf: Call it `SentryErrorModule` or something more specific than `AnalyticsModule`.
Expect if we want to have one module to manage all `Analytics` tools, but then let's do that in another PR and use this one only for sentry in-between.

libs/utils/src/lib/sentry.module.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  handleError(error) {
+    const eventId = Sentry.captureException(error.originalError || error);
+    const { uid, email } = this.authQuery.user || { uid: undefined, email: undefined };
+    Sentry.showReportDialog({ eventId, user: { email } });
+  }
+}
+
+// Init and add the Sentry ErrorHandler if and only if sentry is defined in env.
+if (sentryDsn) {
```
GrandSchtroumpf: Is it not a duplicate of doing : `sentryDsn ? SentryModule : []` in the app modules ?

Mussuranae: It is, I removed it

PR #625, opened 2019-07-12
Movies with delivery ids

ioaNikas:
- [x] Delivery List guard now load deliveries from movie.deliveryIds to match the other guards logic.
- [x] Updated deleteDelivery backend function to remove delivery id ref in movie.
- [x] Updated guard so "Nothing found at path..." doesn't redirect us on our fallbackUrl.

### DIFF
--- a/apps/backend-functions/src/delete.ts
+++ b/apps/backend-functions/src/delete.ts
@@ -2,7 +2,7 @@ import { db, functions } from './firebase';
 import { prepareNotification, triggerNotifications } from './notify';
 import { isTheSame } from './utils';
 import { getDocument, getCollection, getOrgsOfDocument } from './data/internals';
-import { Delivery, Material } from './data/types';
+import { Delivery, Material, Movie } from './data/types';
 
 export async function deleteFirestoreMovie (
   snap: FirebaseFirestore.DocumentSnapshot,
@@ -39,7 +39,7 @@ export async function deleteFirestoreMovie (
 
   const deliveries = await db
     .collection(`deliveries`)
-    .where('movieId', '==', movie.id)
+    .where(movie.deliveryIds, 'array-contains', 'id')
     .get();
   deliveries.forEach(doc => {
     console.log(`delivery ${doc.id} deleted`);
@@ -49,9 +49,9 @@ export async function deleteFirestoreMovie (
   await batch.commit();
 
   return true;
-};
+}
 
-export async function deleteFirestoreDelivery (
+export async function deleteFirestoreDelivery(
   snap: FirebaseFirestore.DocumentSnapshot,
   context: functions.EventContext
 ) {
@@ -83,6 +83,15 @@ export async function deleteFirestoreDelivery (
       }
     }
   });
+  const movieDoc = await db.doc(`movies/${delivery.movieId}`).get();
+  const movie = await getDocument<Movie>(`movies/${delivery.movieId}`);
+  if (movie.deliveryIds.includes(delivery.id)) {
+    console.log(`delete delivery id reference in movie ${movie.id}`);
+    const newDeliveryIds: string[] = movie.deliveryIds.filter(
+      (deliveryId: string) => deliveryId !== delivery.id
+    );
+    batch.update(movieDoc.ref, { movieIds: newDeliveryIds });
+  }
 
   await batch.commit();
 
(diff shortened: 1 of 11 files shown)


apps/backend-functions/src/delete.ts
```diff
@@ -84,6 +85,17 @@ export async function deleteFirestoreDelivery (
     }
   });
 
+  const movieDoc = await db.doc(`movies/${delivery.movieId}`).get();
+  const movie = await getDocument<Movie>(`movies/${delivery.movieId}`);
+
+  if (movie.deliveryIds.includes(delivery.id)) {
```
GrandSchtroumpf: It would be easier here to use the `indexOf` method : 
```typescript
const index = movie.deliveryIds.indexOf(delivery.id);
if (index !== -1) {
  const deliveryIds = [ ...movie.deliveryIds.slice(0, index), ...movie.deliveryIds.slice(index + 1) ]
  batch.update(movieDoc.ref, { deliveryIds });
}
```

apps/backend-functions/src/delete.ts
```diff
@@ -141,22 +152,22 @@ export async function deleteFirestoreMaterial (
...
     if (movieMaterial.deliveriesIds.length === 1) {
-      db.doc(`movies/${delivery.movieId}/materials/${movieMaterial.id}`).delete()
-    }
-    else {
-      const newdeliveriesIds = movieMaterial.deliveriesIds.filter((id: string) => id !== delivery.id);
-      db.doc(`movies/${delivery.movieId}/materials/${movieMaterial.id}`)
-        .update({ deliveriesIds: newdeliveriesIds });
+      db.doc(`movies/${delivery.movieId}/materials/${movieMaterial.id}`).delete();
+    } else {
+      const newdeliveriesIds = movieMaterial.deliveriesIds.filter(
```
GrandSchtroumpf: Format in one line. You shouldn't have to type the `(id: string)`.

libs/material/src/lib/delivery/guards/delivery-list.guard.ts
```diff
@@ -31,13 +33,24 @@ export class DeliveryListGuard extends StateListGuard<Delivery> {
...
-        return this.fireQuery.fromQuery<DeliveryDB[]>(query);
-      }),
-      map(deliveries => deliveries.map(delivery =>
-        modifyTimestampToDate(delivery)))
-    );
+    return this.movieQuery
+      .selectActive(movie => movie.deliveryIds)
+      .pipe(
+        switchMap(ids => {
+          if (!ids || ids.length === 0) throw new Error('No Delivery yet');
```
GrandSchtroumpf: If ids is [] return `of([])`, like that you can return the url in the second pipe : 
```typescript
switchMap(ids => {
  if (!ids || ids.length === 0) return of([]);
  ....
}),
map(deliveries => {
  if (deliveries.length === 0) return this.urlFallback;
})
```

libs/material/src/lib/delivery/guards/delivery-list.guard.ts
```diff
@@ -31,13 +33,24 @@ export class DeliveryListGuard extends StateListGuard<Delivery> {
...
+          const queries = ids.map(id => {
+            return this.fireQuery.fromQuery<DeliveryDB>(deliveryQuery(id)).pipe(
```
GrandSchtroumpf: What are you doing here... ? Why do you have to catchError in the `fromQuery` ?

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -53,22 +50,15 @@ export class MovieService {
...
-    // Delete the movie in movies collection
-    batch.delete(movieDoc.ref);
-    // Delete the movie id in organization.movieIds
-    batch.update(organizationDoc.ref, { movieIds });
-    // Remove the movie from the movies store
-    // TODO: Wait for firestore response before removing the movie. => ISSUE#611
-    this.store.remove(movieId);
-
-    return batch.commit();
+    await this.db.firestore.runTransaction(async(tx: firebase.firestore.Transaction) => {
```
GrandSchtroumpf: Not sure to understant why you need a transaction here ...

libs/material/src/lib/delivery/guards/delivery-list.guard.ts
ioaNikas: I don't want to throw an error here otherwise I'll be redirected to the urlFallBack.
At the end, I only want to catch NotFoundErrors and return undefined in that particular case.
We can talk about a better way to do it as you told me earlier.

libs/movie/src/lib/movie/+state/movie.service.ts
ioaNikas: I send my `tx` object to `createDocAndPermissions` and `addStakeholder` to create every documents in the same transaction. This way, it won't create anything if there is a problem (invalid data, disconnect etc) and nobody can write on those docs (like updating organization.movieIds) during the process.

PR #628, opened 2019-07-12
Issue#624 wallet message view

Maxstgt:
Implementing the wallet message view


libs/ethers/src/lib/wallet/wallet-message-view/wallet-message-view.component.html
```diff
@@ -0,0 +1,19 @@
...
+  <p>{{ message.subline }}</p>
+  <img
+    src="/assets/images/ppl_celebrating.png"
+    alt="People celebrating"
+  />
+  <a mat-button (click)="redirectUser.emit()">Okay</a>
+</section>
+<!-- Section to display the inputted error message -->
+<ng-template #isError>
+  <h1>Something went wrong. Please go back. 404 or something like that</h1>
```
GrandSchtroumpf: " 404 or something like that" don't leave that here. Ask the team what to write here.

libs/ethers/src/lib/wallet/wallet-message-view/wallet-message-view.component.ts
```diff
@@ -0,0 +1,25 @@
+import { Component, Input, EventEmitter, Output } from '@angular/core';
+import { ChangeDetectionStrategy } from '@angular/core';
+
+interface WalletCreationMsg {
```
pldespaigne: Please use another name here because this message is not displayed only for the creation

libs/ethers/src/lib/wallet/wallet-message-view/wallet-message-view.component.ts
```diff
@@ -0,0 +1,25 @@
...
+})
+export class WalletMessageViewComponent {
+  /** The message which should be displayed when this component is called. Could also be an error message */
+  @Input() public message: WalletCreationMsg;
+
+  /** 
+   * An event that notify the parent component that the user want to back
+   * to his wallet or depending on the routerparams where he came from
+   */
+  @Output() public redirectUser: EventEmitter<void> = new EventEmitter();
```
GrandSchtroumpf: Why not using the [`Location`  service](https://angular.io/api/common/Location) to go back ?

libs/ethers/src/lib/wallet/wallet-message-view/wallet-message-view.component.html
pldespaigne: I think we should display the input message in both case,
the `isError` boolean can be used to change the picture and/or the css

libs/ethers/src/lib/wallet/wallet-message-view/wallet-message-view.component.ts
pldespaigne: because the parent component handle the redirection not this component, moreover the redirection could be elsewhere (not only previous location)

PR #629, opened 2019-07-12
Issue 578: Add email verification flow to the delivery app

lsenta:
Add email verify message + icon + flow to the delivery app.

### DIFF
--- a/libs/auth/src/lib/+state/auth.store.ts
+++ b/libs/auth/src/lib/+state/auth.store.ts
@@ -1,3 +1,4 @@
+import * as firebase from 'firebase';
 import { Injectable } from '@angular/core';
 import { Store, StoreConfig } from '@datorama/akita';
 
@@ -31,6 +32,7 @@ export interface AccountForm {
 
 export interface AuthState {
   user: User;
+  auth?: { emailVerified: boolean };
   form: UserForm;
   accountForm: AccountForm;
   requestedRoute?: string;
--- a/libs/auth/src/lib/+state/auth.store.ts
+++ b/libs/auth/src/lib/+state/auth.store.ts
@@ -1,4 +1,3 @@
-import * as firebase from 'firebase';
 import { Injectable } from '@angular/core';
 import { Store, StoreConfig } from '@datorama/akita';
 
--- a/libs/auth/src/lib/components/email-verify/email-verify.component.html
+++ b/libs/auth/src/lib/components/email-verify/email-verify.component.html
@@ -1,4 +1,4 @@
-<p *ngIf="!(hasVerifiedEmail$ | async)">
+<article *ngIf="!(hasVerifiedEmail$ | async)">
   <button
     mat-icon-button
     matTooltip="Your email is not verified, click here to send another verification mail."
@@ -8,4 +8,4 @@
   >
     <mat-icon matBadge="!" matBadgePosition="before" matBadgeColor="accent">email</mat-icon>
   </button>
-</p>
+</article>
--- a/libs/auth/src/lib/components/email-verify/email-verify.component.html
+++ b/libs/auth/src/lib/components/email-verify/email-verify.component.html
@@ -1,0 +1,11 @@
+<p *ngIf="!(hasVerifiedEmail$ | async)">
+  <button
+    mat-icon-button
+    matTooltip="Your email is not verified, click here to send another verification mail."
+    matTooltipClass="tooltip-red"
+    [disabled]="triggeredOnce"
+    (click)="sendEmail()"
+  >
+    <mat-icon matBadge="!" matBadgePosition="before" matBadgeColor="accent">email</mat-icon>
+  </button>
+</p>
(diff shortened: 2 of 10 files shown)


libs/auth/src/lib/+state/auth.store.ts
```diff
@@ -1,3 +1,4 @@
+import * as firebase from 'firebase';
```
GrandSchtroumpf: Don't import everything from firebase. Only the methods you need.

libs/auth/src/lib/components/email-verify/email-verify.component.html
```diff
@@ -0,0 +1,11 @@
+<p *ngIf="!(hasVerifiedEmail$ | async)">
```
GrandSchtroumpf: Don't put a button inside a paragraphe. Use `ng-container` or `article` instead.

libs/auth/src/lib/components/email-verify/email-verify.component.ts
```diff
@@ -0,0 +1,27 @@
...
+
+@Component({
+  selector: 'auth-email-verify',
+  templateUrl: './email-verify.component.html',
+  styleUrls: ['./email-verify.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class EmailVerifyComponent implements OnInit {
+  public hasVerifiedEmail$: Observable<boolean>;
+  public triggeredOnce: boolean;
```
GrandSchtroumpf: where do you use that ?

libs/auth/src/lib/guard/auth.guard.ts
```diff
@@ -28,9 +28,12 @@ export class AuthGuard implements CanActivate {
     return new Promise((res, rej) => {
       this.subscription = this.afAuth.authState
         .pipe(
-          switchMap(user => {
-            if (!user) throw new Error('Not connected');
-            return this.db.doc<User>(`users/${user.uid}`).valueChanges();
+          tap(userAuth => {
```
GrandSchtroumpf: It should be after the switchMap. Else if `userAuth` doesn't exist you'll get an error.

libs/auth/src/lib/components/email-verify/email-verify.component.ts
lsenta: In the header component:
https://github.com/blockframes/blockframes/pull/629/files#diff-79a39819f78f9789a2d5b4c49c766166R9

https://github.com/blockframes/blockframes/pull/629/files#diff-b077ca89ded5a70d8e24c5592bbf70b4R21

libs/auth/src/lib/guard/auth.guard.ts
lsenta: I'll lose the user auth object, I've fixed the tap to ignore the null case (the rest is handled by the switchMap)

GrandSchtroumpf: I would recommand to put this code in the switchMap to avoid duplicate, and make it more readable : 
```typescript
switchMap(userAuth => {
  if (!userAuth) throw new Error('Not connected');
  this.store.update({ auth: { emailVerified: userAuth.emailVerified } });
  return this.db.doc<User>(`users/${userAuth.uid}`).valueChanges();
})
```

lsenta: SGTM, PR updated

libs/auth/src/lib/components/email-verify/email-verify.component.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  constructor(private query: AuthQuery, private service: AuthService) {}
+
+  ngOnInit() {
+    this.hasVerifiedEmail$ = this.query.hasVerifiedEmail$;
+    this.triggeredOnce = false;
```
GrandSchtroumpf: If this is something that you want to keep even if the component is destroyed, you should put it in the state. Else, keep it here.

libs/auth/src/lib/components/email-verify/email-verify.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  public async sendEmail() {
+    await this.service.sendVerifyEmail();
```
GrandSchtroumpf: Maybe you want to try/catch this.

libs/auth/src/lib/components/email-verify/email-verify.component.ts
lsenta: I'll keep it here for now!

libs/auth/src/lib/components/email-verify/email-verify.component.ts
lsenta: I'd rather throw explicitly for now, we can catch later in the component.
It'll trigger the sentry logger if something goes wrong (low probability since this is a google API).

PR #630, opened 2019-07-12
Issue 578: verify by default the user and generate password

lsenta:


apps/backend-functions/src/users.ts
```diff
@@ -1,5 +1,6 @@
 import * as admin from 'firebase-admin';
 import * as functions from 'firebase-functions';
+import * as generator from 'generate-password';
```
GrandSchtroumpf: Only import what you need: `import { generate } from 'generate-password'`

PR #631, opened 2019-07-15
Issue#529 identity

Hugo-Hontans:
Create 2 pages to redirect user who don't have name and surname.


apps/delivery/delivery/src/app/app-routing-module.ts
```diff
@@ -4,13 +4,13 @@ import { RouterModule, Routes } from '@angular/router';
 
 // Components
 import { LayoutComponent } from './layout/layout.component';
+import { MovieEmptyComponent } from '@blockframes/movie/movie/components/movie-empty/movie-empty.component';
```
GrandSchtroumpf: You should just do '@blockframes/movie'. You need to export the component. If you're using it here.

libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -50,8 +50,8 @@ export class AuthService {
   }
 
   /** Upate a user */
-  public update(uid: string, user: Partial<User>) {
-    return this.db.doc<User>(`users/${uid}`).update(user);
+  public async update(uid: string, user: Partial<User>) {
+    await this.db.doc<User>(`users/${uid}`).update(user);
```
GrandSchtroumpf: Why ? If `update` returns a Promise, it's better as it was.

libs/auth/src/lib/+state/auth.store.ts
```diff
@@ -9,8 +9,11 @@ export interface User {
   };
   balance: string;
   email: string;
+  //TODO: delete lastName and firstName
```
GrandSchtroumpf: I think you can do that in this PR. Or add an issue, and set the issue id in this comment.

libs/auth/src/lib/auth.module.ts
```diff
@@ -20,11 +20,15 @@ import { UiFormModule } from '@blockframes/ui';
 import { SignupFormComponent } from './components/signup-form/signup-form.component';
 import { SigninFormComponent } from './components/signin-form/signin-form.component';
 import { WelcomeViewComponent } from './pages/welcome-view/welcome-view.component';
+import { InvitationComponent } from './pages/invitation/invitation.component';
```
GrandSchtroumpf: I though invitation and congratulation where in the notification module. Can you check with Yohann ?

libs/auth/src/lib/guard/auth.guard.ts
```diff
@@ -33,11 +33,17 @@ export class AuthGuard implements CanActivate {
             return this.db.doc<User>(`users/${user.uid}`).valueChanges();
           }),
           tap(user => this.store.update({ user })),
+          tap(user => {
+            if (!(user.name && user.surname)) throw new Error('User don\'t have name and surname');
```
GrandSchtroumpf: If this is a problem, you should force it in the form. A user should not being able to create an account without a name/surname. Else, you shouldn't check that in the guard.

libs/auth/src/lib/guard/auth.guard.ts
```diff
@@ -33,11 +33,17 @@ export class AuthGuard implements CanActivate {
...
+          }),
           map(user => !!user)
         )
         .subscribe({
           next: (response: boolean) => res(response),
-          error: () => res(this.router.parseUrl('auth/connexion'))
+          error: (error) => {
+            if (error.message === 'User don\'t have name and surname') res(this.router.parseUrl('auth/invitation'))
```
GrandSchtroumpf: I really don't think this is a good idea to have different output depending on the error. It should always be the same url in case of error.

libs/auth/src/lib/guard/auth.guard.ts
```diff
@@ -33,11 +33,17 @@ export class AuthGuard implements CanActivate {
             return this.db.doc<User>(`users/${user.uid}`).valueChanges();
           }),
           tap(user => this.store.update({ user })),
+          tap(user => {
+            if (!(user.name && user.surname)) throw new Error('User don\'t have name and surname');
+          }),
           map(user => !!user)
```
GrandSchtroumpf: You could just do : 
```
map(user => !(user.name && user.surname) ? this.router.parseUrl('auth/invitation') : true)
```

libs/auth/src/lib/pages/congratulation/congratulation.component.scss
```diff
@@ -0,0 +1,29 @@
...
+  background-size: contain;
+  background-repeat: no-repeat;
+}
+
+button {
+  margin-top: 80px;
+}
+
+// extra small
+@media screen and (max-width: 599px) {
```
GrandSchtroumpf: Create a var(--small) or use one that already exist (in FlexLayout for exemple ?)

libs/auth/src/lib/pages/invitation/invitation.component.scss
```diff
@@ -0,0 +1,60 @@
...
+
+mat-form-field {
+  width: 100%;
+}
+
+.image {
+  height: 400px;
+  width: 500px;
+  background-image: url('libs/shared-assets/images/rocket.png');
+  background-size: contain;
```
GrandSchtroumpf: Why not cover ?

libs/auth/src/lib/pages/invitation/invitation.component.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  constructor(
+    private service: AuthService,
+    private query: AuthQuery,
+    private snackBar: MatSnackBar,
+    private router: Router
+  ) {}
+
+  public async update() {
```
GrandSchtroumpf: Maybe it would be more readable with : 
```
if (!this.form.valid) {
  return this.snackBar.open('Please enter valid name and surname', 'close', { duration: 2000 });
}
try {
  const uid = this.query.userId;
  await this.service.update(uid, this.form.value);
  this.router.navigate(['auth/congratulation']);
} catch (err) {
  this.snackBar.open(error.message, 'close', { duration: 5000 });
}
```

libs/auth/src/lib/guard/auth.guard.ts
Hugo-Hontans: It is for a user invited by an organization, it must return name and surname at the first connexion.
Otherwise the user must create his account directly with his name and surname

GrandSchtroumpf: Ok, then use the solution with the map described below

libs/auth/src/lib/guard/auth.guard.ts
Hugo-Hontans: I tried something like that but it does not work

libs/auth/src/lib/auth.module.ts
Hugo-Hontans: I rename InvitationComponent to IdentityComponent

libs/auth/src/lib/guard/auth.guard.ts
GrandSchtroumpf: It should work ;). 
Also add 'UrlTree' in the `next: (response: boolean | UrlTree) => res(response)`

libs/auth/src/lib/guard/auth.guard.ts
```diff
@@ -33,11 +33,11 @@ export class AuthGuard implements CanActivate {
             return this.db.doc<User>(`users/${user.uid}`).valueChanges();
           }),
           tap(user => this.store.update({ user })),
-          map(user => !!user)
+          map(user => user && user.name && user.surname ? true : this.router.parseUrl('auth/identity'))
```
GrandSchtroumpf: add parenthesis for better readibility. Or create a function : 
```typescript
// function is ouside of the class, at the beginning of the file.
function isIdentitied(user) {
  return user && user.name && user.surname;
}
....
map(user => isIdentified(user) ? true : this.router.parseUrl('auth/identity'))
```

PR #647, opened 2019-07-17
fix responsive and css of editable-sidenav, fix css layout of delivery

Hugo-Hontans:


libs/ui/src/lib/editable-sidenav/editable-sidenav.component.html
```diff
@@ -1,8 +1,8 @@
-<mat-sidenav-container [class.padding-right]="!opened" fxLayout="row" fxLayoutAlign="center">
+<mat-sidenav-container fxLayout="row" fxLayoutAlign="center">
   <mat-sidenav-content>
     <ng-content select="[content]"></ng-content>
   </mat-sidenav-content>
-  <mat-sidenav #sidenav [opened]="opened" mode="side" position="end">
+  <mat-sidenav #sidenav fxLayout="row" fxLayoutAlign="center center" [opened]="opened" mode="side" position="end">
```
GrandSchtroumpf: So you don't use the properties from angular material ?

libs/ui/src/lib/editable-sidenav/editable-sidenav.component.scss
```diff
@@ -3,25 +3,19 @@
...
   padding-top: 50px;
-  padding-left : 200px;
-}
-
-.padding-right {
-  padding-right: 200px;
 }
 
 mat-sidenav-content {
   width: 55%;
```
GrandSchtroumpf: Maybe switch to 60% ? and add some padding if needed

libs/ui/src/lib/editable-sidenav/editable-sidenav.component.scss
```diff
@@ -38,48 +32,38 @@ mat-card {
...
+    padding-top: 25px;
   }
-  .padding-right {
-    padding-right: 50px;
-  }
-}
+ }
 
 // small
 @include tablet {
```
GrandSchtroumpf: Is it needed, as this is the same as above ?

Hugo-Hontans: not exactly the same, I reduce the padding 50px to 25px

libs/ui/src/lib/editable-sidenav/editable-sidenav.component.scss
Hugo-Hontans: You're right like that we manage perfectly the size

libs/ui/src/lib/editable-sidenav/editable-sidenav.component.html
Hugo-Hontans: I prefer like that because the property ``fixedInViewport`` overloads the top position to 0

GrandSchtroumpf: You can change to with `fixedTopGap="66"`

PR #649, opened 2019-07-18
Issue#633 signup

Hugo-Hontans:


apps/backend-functions/src/users.ts
```diff
@@ -20,10 +20,19 @@ interface OrgProposal {
...
-    .doc(user.uid)
-    .set({ email, uid });
+
+  const userDocRef = db.collection('users').doc(user.uid);
+
+  // transaction to UPSERT the user doc
+  return db.runTransaction(async tx => {
+    const userDoc = await tx.get(userDocRef);
+
+    if (userDoc.exists) {
```
GrandSchtroumpf: I prefere ternary for this kind of operation. But if / else if good too.

libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -44,9 +53,17 @@ export class AuthService {
...
   /** Create a user based on firebase user */
-  public create({ email, uid }: firebase.User) {
-    const user = createUser({ email, uid })
-    return this.db.doc<User>(`users/${uid}`).set(user);
+  public create(user: User) {
+    const userDocRef = this.db.firestore.collection('users').doc(user.uid);
+    // transaction to UPSERT the user doc
+    return this.db.firestore.runTransaction(async tx => {
+      const userDoc = await tx.get(userDocRef);
+      if (userDoc.exists) {
```
GrandSchtroumpf: Same here.

apps/backend-functions/src/users.ts
Hugo-Hontans: it's ok for me like that

PR #651, opened 2019-07-18
Issue#542 wallet send tunnel

pldespaigne:


apps/backend-functions/src/relayer.ts
```diff
@@ -26,9 +24,29 @@ export interface RelayerConfig {
   factoryContract: string;
 }
 
+/** Regular Ethereum transaction */
+export interface Tx {
```
GrandSchtroumpf: Maybe call it differencly, because it's not complete: `BaseTx` ?

apps/backend-functions/src/relayer.ts
```diff
@@ -52,30 +69,32 @@ export const initRelayer = (config: RelayerConfig): Relayer => {
...
+interface RegisterParams {
+  name: string;
+  address: string;
+}
+
+//---------------------------------------------------
+//                   DEPLOY
+//---------------------------------------------------
+
+export const relayerDeployLogic = async (
```
GrandSchtroumpf: Create a real function for that as you don't need the property of arrow function : 
```
export async function relayerDeployLogic(...) { .... }
```

apps/backend-functions/src/relayer.ts
```diff
@@ -93,106 +112,136 @@ export const relayerCreateLogic = async (
...
 
+    const result: {[key: string]: TxReceipt | undefined } = {};
+ 
+    const codeAtAddress = await relayer.wallet.provider.getCode(erc1077address);
+    if (codeAtAddress === '0x') { // if there is already some code at this address : skip deploy
+      const deployTx: TxResponse = await relayer.contractFactory.deploy(
+        hash,
+        `0x${ERC1077.bytecode}`,
+        key.toLocaleLowerCase(),
+        '0x4D7e2f3ab055FC5d484d15aD744310dE98dD5Bc3'.toLocaleLowerCase(), // TODO this will be the org address in the future
```
GrandSchtroumpf: create a variable for that above. And create an issue for your TODO.

apps/backend-functions/src/relayer.ts
```diff
@@ -93,106 +112,136 @@ export const relayerCreateLogic = async (
...
+    console.error(error);
+    throw new Error(error);
+  }
+};
+
+//---------------------------------------------------
+//                   REGISTER
+//---------------------------------------------------
+
+export const relayerRegisterENSLogic = async (
```
GrandSchtroumpf: Noram functoin as explain above

apps/backend-functions/src/relayer.ts
```diff
@@ -93,106 +112,136 @@ export const relayerCreateLogic = async (
...
-      return result;
-    };
-    const [ ensResult, deployResult ] = await Promise.all([ensWorkFlow(), deployWorkFlow()]);
-    return { ens: ensResult, deploy: deployResult };
+    
+    const retreivedAddress = await relayer.wallet.provider.resolveName(fullName);
+    if (!!retreivedAddress && retreivedAddress !== ZERO_ADDRESS) { // if name is already link to a non-zero address : skip
+      throw new Error(`${fullName} already linked to an address (${retreivedAddress})`);
+    }
+    const result: {[key: string]: TxReceipt | undefined } = {};
```
GrandSchtroumpf: You can use the build-in `Recorder<string, TxReceipt | undefined>` type for that. And I think you should remove the `| undefined`.

apps/backend-functions/src/relayer.ts
```diff
@@ -93,106 +112,136 @@ export const relayerCreateLogic = async (
...
   }
 };
 
+
+
+//---------------------------------------------------
+//                   SEND
+//---------------------------------------------------
+
 export const relayerSendLogic = async (
```
GrandSchtroumpf: Same as above

libs/ethers/src/lib/types.ts
```diff
@@ -10,4 +10,25 @@ export interface Addresses {
   kovan?: string;
   rinkeby?: string;
   local?: string;
+}
+
+/** Regular Ethereum transaction */
+export interface Tx {
```
GrandSchtroumpf: You shouldn't reimplement that. The best is to create it here, and import it inside the backend.

libs/ethers/src/lib/wallet/+state/wallet.service.ts
```diff
@@ -91,14 +91,107 @@ export class WalletService {
...
-      key = this.keyManagerQuery.getMainKeyOfUser(ensDomain).address;
-    } catch(error) {
-      throw new Error('You are trying to deploy an ERC1077 without any key');
+  public async deployERC1077(ensDomain: string, pubKey: string) {
+    if (this.query.getValue().hasERC1077) {
+      throw new Error('Your smart-wallet is already deployed');
+    }
+    this.store.setLoading(true);
+    try {
+      const name = ensDomain.split('.')[0]; // `alice.blockframes.eth` -> `alice`
```
GrandSchtroumpf: You should create a pure function for that.

libs/ethers/src/lib/wallet/+state/wallet.service.ts
```diff
@@ -91,14 +91,107 @@ export class WalletService {
...
+      this.store.setLoading(false);
+      return result;
+    } catch(err) {
+      this.store.setLoading(false);
+      console.error(err);
+      throw new Error('Deploy seems to have failed, but firebase function is maybe still runing');
+    }
+  }
+
+  private getUsersERC1077(ensDomainOrAddress: string) {
```
GrandSchtroumpf: Add some comments on your methods.

libs/ethers/src/lib/wallet/+state/wallet.service.ts
```diff
@@ -91,14 +91,107 @@ export class WalletService {
...
+    this._requireProvider();
+    return new Contract(ensDomainOrAddress, ERC1077.abi, this.provider);
+  }
+
+  public async setDeleteKeyTx(pubKey: string) {
+    this.setTx(getMockTx()); // TODO use getDeleteKeyTx
+    // TODO call key manager delete key to delete localy after the tx has been mined
```
GrandSchtroumpf: Create issues and reference them in your comments

libs/ethers/src/lib/wallet/+state/wallet.service.ts
```diff
@@ -91,14 +91,107 @@ export class WalletService {
...
+  }
+
+  public setTx(tx: Tx) {
+    this.store.update({tx});
+  }
+
+  private hashMetaTx(from: string, metaTx: MetaTx) {
+    return this.getUsersERC1077(from).calculateMessageHash(
```
GrandSchtroumpf: Why calcultatMessageHash doesn't takes an object instead of all those arguments ?

libs/ethers/src/lib/wallet/wallet-send-tx/wallet-send-tx.component.html
```diff
@@ -0,0 +1,89 @@
+<main [ngSwitch]="step">
```
GrandSchtroumpf: I would recommend to use a `<ng-container>` here too.
Also start with the ` *ngIf="(wallet$ | async)`, then do the `ngSwitch`.

libs/ethers/src/lib/wallet/+state/wallet.service.ts
pldespaigne: because the standard ERC1077 contract that we use has this instead of a struct as params, we could probably change that with `ABIv2` but it's not currently a priority

libs/ethers/src/lib/wallet/wallet-send-tx/wallet-send-tx.component.html
```diff
@@ -1,89 +1,78 @@
-<main [ngSwitch]="step">
+<main>
```
GrandSchtroumpf: Remove this "main", it's useless here.

PR #659, opened 2019-07-19
Issue#506 organization editable

ioaNikas:
- [x] Organization is editable (only address at the moment, but can be extended to other fields later) using the new editable side-nav component.
- [x] Rename and sort components and pages for easier reading.
- [x] Invitations are triggered once, and queried by organizationId. (issue#645)


libs/notification/invitation/+state/invitation.service.ts
```diff
@@ -19,16 +19,16 @@ export class InvitationService {
   public get userInvitations$() {
     return this.authQuery.user$.pipe(
       filter(user => !!user),
-      switchMap(user => this.db.fromQuery(this.getInvitationByUserId(user.uid))),
-      tap((invitations: any) => this.store.set(invitations)) // TODO : Find a way to cast invitations as Invitation[];
+      switchMap(user => this.db.fromQuery(this.getInvitationsByOrgId(user.orgId))),
+      tap((invitations: any) => this.store.set(invitations))
```
GrandSchtroumpf: You should type "invitiations"

libs/organization/src/lib/components/organization-display/organization-display.component.html
```diff
@@ -0,0 +1,43 @@
+<ng-container *ngIf="organization; else: loading">
+  <mat-card class="mat-elevation-z1">
```
GrandSchtroumpf: Don't specify the elevation everywhere. If there is a "standard" elevation level it should be settled in the theme.

GrandSchtroumpf: Also, I think the `fxLayout="row"` should be on the mat-card, no ? 
```html
<mat-card fxLayout="row">
  <img />
  <mat-card-content></>
</mat-card>
```

libs/organization/src/lib/components/organization-display/organization-display.component.html
```diff
@@ -0,0 +1,43 @@
...
+      <!-- TODO: issue#579 import picture -->
+      <img
+        mat-card-avatar
+        src="https://material.angular.io/assets/img/examples/shiba2.jpg"
+        alt="Profile picture"
+      />
+      <article class="information" fxLayout="column" fxLayoutAlign="start start">
+        <div>
+          <h4 class="name">
+            <b>{{ organization.name }}</b>
```
GrandSchtroumpf: Remove b tag, it's already a title. Use material typpography

libs/organization/src/lib/components/organization-form/organization-form.component.ts
```diff
@@ -0,0 +1,16 @@
+import { Component, ChangeDetectionStrategy } from '@angular/core';
+import { ControlContainer } from '@angular/forms';
+
+@Component({
+  selector: 'organization-form',
```
GrandSchtroumpf: Add `[formGroupName] organization-form, [formGroup] organization-form` to force the selector to have `formGroupName` or `formGroup` attribute.

libs/organization/src/lib/components/organization-form/organization-form.component.ts
```diff
@@ -0,0 +1,16 @@
...
+  templateUrl: './organization-form.component.html',
+  styleUrls: ['./organization-form.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class OrganizationFormComponent {
+  constructor(public controlContainer: ControlContainer) {}
+
+  public get control() {
```
GrandSchtroumpf: if there is an OrganizationForm you can do : `public get control(): OrganizationForm {}`

libs/organization/src/lib/organization.module.ts
```diff
@@ -21,35 +21,39 @@ import { MatMenuModule } from '@angular/material/menu';
...
-    component: OrgFormComponent
+    path: 'create',
+    component: OrganizationCreateComponent
   },
   {
     path: ':orgId',
     children: [
       { path: '', redirectTo: 'view', pathMatch: 'full' },
-      { path: 'view', component: OrgViewComponent },
+      { path: 'view', component: OrganizationEditableComponent },
```
GrandSchtroumpf: It's suppose to be "edit" the route now.

libs/organization/src/lib/pages/organization-editable/organization-editable.component.ts
```diff
@@ -0,0 +1,52 @@
...
+    private query: OrganizationQuery,
+    private permissionsQuery: PermissionsQuery,
+    private service: OrganizationService,
+    private snackBar: MatSnackBar
+    ) {}
+
+  ngOnInit() {
+    this.organization$ = this.query.select('org');
+    this.isSuperAdmin$ = this.permissionsQuery.isSuperAdmin$;
+    this.organizationProfileForm = new OrganizationProfileForm(this.query.getValue().org);
```
GrandSchtroumpf: You need to keep listening of the update of the server I think : 
```typescript
this.organizationProfileForm = new OrganizationProfileForm();
this.organization$ = this.query.select('org').pipe(tap(org => this.organizationProfileForm.patchValue(org));
this.isSuperAdmin$ = this.permissionsQuery.isSuperAdmin$;
```

GrandSchtroumpf: Maybe the best would be to have : 
```typescript
this.organization$ = this.query.select('org').pipe(
  tap(org => this.organizationProfileForm.patchValue(org)),
  switchMap(_ => this.organizationProfileForm.valueChanges)
);
```
Like that your form is always in sync with the DB, but still you have local changes available. But the effect might be weird if someone is updating the DB in the same time as you're writing...

libs/organization/src/lib/pages/organization-editable/organization-editable.component.ts
```diff
@@ -0,0 +1,52 @@
...
+  public get organizationInformations$() {
+    return this.organizationProfileForm.valueChanges.pipe(startWith(this.organizationProfileForm.value));
+  }
+
+  public openSidenav() {
+    this.opened = true;
+  }
+
+  public update() {
+    try {
```
GrandSchtroumpf: All this should be a generic method that should be reuse all the time. Let me check of to do that.

libs/organization/src/lib/pages/organization-editable/organization-editable.component.ts
ioaNikas: I really don't get what you're trying to explain me. Maybe we can have a short call about this to make it clear for me ?

PR #662, opened 2019-07-19
Issue 577: Add backend part for orgs full-text search

lsenta:


apps/backend-functions/src/algolia.ts
```diff
@@ -0,0 +1,25 @@
...
+import { algoliaAdminKey, algoliaId } from './environments/environment';
+
+const indexOrganizationsBuilder = () => {
+  const client = algoliasearch(algoliaId, algoliaAdminKey);
+  const INDEX_NAME_ORGANIZATIONS = 'orgs';
+  return client.initIndex(INDEX_NAME_ORGANIZATIONS);
+};
+
+export function storeSearchableOrg(orgId: string, name: string): Promise<any> {
+  if (algoliaId === "") {
```
GrandSchtroumpf: You should check of `!algoliaId`, like that you can check for undefined and null also.

apps/backend-functions/src/environments/environment.prod.ts
```diff
@@ -10,3 +10,7 @@ export { factoryContract, backupBucket, relayer, appUrl } from '@env';
 
 export const sendgridAPIKey = functions.config().sendgrid.apiKey;
 export const mnemonic = functions.config().relayer.mnemonic;
+
+export const algoliaId = functions.config().algolia.app_id;
```
GrandSchtroumpf: if functions.config() require ressources, I would rather store the return value and exports the keys : 
```
export const {
 app_id: algoliaId,
 api_key: algoliaAdminKey,
 search_key: algoliaSearchKey
} = functions.config().algolia;
```

apps/backend-functions/src/algolia.ts
lsenta: That's only for dev mode when the rest of the team work with the app without an algolia account, the tighter the check the safer.

apps/backend-functions/src/environments/environment.prod.ts
lsenta: Its OK, its designed to be used like a process.env

PR #664, opened 2019-07-19
Issue#508 firestore rules

ioaNikas:
- [x] New set of firestore rules

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -2,41 +2,53 @@ rules_version = '2';
 service cloud.firestore {
   match /databases/{database}/documents {
 
-    // Note: Add a verification of each document ID => {id: ID_SHOULD_BE_EQUAL_TO_PATH}
-
     /// USERS RULES ///
 
     match /users/{userId} {
-    	allow read, update: if isSignedIn(); // TODO: Secure it as soon as we got Inivitations collection
-    	allow create: if userId() == incomingData().uid;
+    	allow read, update: if isSignedIn();
+    	allow create: if (userId() == incomingData().uid && userId() == userId)
+      	|| isOrgAdmin(userId(), userOrgId());
     }
 
     /// NOTIFICATIONS RULES ///
 
     match /notifications/{notificationId} {
-    	allow read: if currentUser().uid == existingData().userId;
+    	allow read: if (currentUser().uid == existingData().userId)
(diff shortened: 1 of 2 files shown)


firestore.rules
```diff
@@ -2,41 +2,53 @@ rules_version = '2';
...
   match /databases/{database}/documents {
 
-    // Note: Add a verification of each document ID => {id: ID_SHOULD_BE_EQUAL_TO_PATH}
-
     /// USERS RULES ///
 
     match /users/{userId} {
-    	allow read, update: if isSignedIn(); // TODO: Secure it as soon as we got Inivitations collection
-    	allow create: if userId() == incomingData().uid;
+    	allow read, update: if isSignedIn();
```
lsenta: I'm pretty sure we still have a security issue, (I can update any user if I'm signed in) keep the TODO

firestore.rules
```diff
@@ -2,41 +2,53 @@ rules_version = '2';
...
+    	allow create: if (userId() == incomingData().uid && userId() == userId)
+      	|| isOrgAdmin(userId(), userOrgId());
     }
 
     /// NOTIFICATIONS RULES ///
 
     match /notifications/{notificationId} {
-    	allow read: if currentUser().uid == existingData().userId;
+    	allow read: if (currentUser().uid == existingData().userId)
```
lsenta: `currentUser().uid ` is `userId()`

firestore.rules
```diff
@@ -2,41 +2,53 @@ rules_version = '2';
...
+        || (userOrgId() == existingData().organizationId)
+      	&& (existingData().state != incomingData().state);
     }
 
     /// ORGANIZATION RULES ///
 
     match /orgs/{orgId} {
-    	allow create: if isSignedIn();
-      allow read, update, delete: if isOrgAdmin(userId(), orgId);
+    	allow create, read: if isSignedIn();
```
lsenta: any user can read any org

firestore.rules
ioaNikas: Indeed. We'll remove the TODO and secure it when the invitations will trigger writing of `organizationId` into `user` document.

firestore.rules
ioaNikas: Shouldn't we let user read organizations if they want to find / join one ?

ioaNikas:
Duplicate with #1342

PR #665, opened 2019-07-19
fix guard

pldespaigne:


libs/ethers/src/lib/wallet/guards/wallet-key.guard.ts
```diff
@@ -2,22 +2,42 @@ import { Injectable } from '@angular/core';
...
-  canActivate(): boolean | UrlTree {
-    const { ensDomain } = this.walletQuery.getValue();
-    const count = this.keyQuery.getKeyCountOfUser(ensDomain);
-    if (count === 0) {
-      return this.router.parseUrl('layout/o/account/wallet/add');
-    }
+  canActivate(): Promise<boolean | UrlTree> | UrlTree {
+    return new Promise(async (res, rej) => {
+      this.subscription = this.keyQuery.selectLoading().pipe(
+        filter(loading => !loading), // ensure keys are retreived from local storage
```
GrandSchtroumpf: What happend if there is nothing on the localstorage ? Should you stay on the page ? Should you be redirected ... ?

libs/ethers/src/lib/wallet/guards/wallet-key.guard.ts
```diff
@@ -2,22 +2,42 @@ import { Injectable } from '@angular/core';
...
+        switchMap(() => this.walletQuery.select()),
+        filter(wallet => !!wallet.ensDomain), // ensure that wallet has ensDomain value
```
GrandSchtroumpf: The problem here is that if you don't have a ensDomain you won't be redirected. You'll just stay on your page.
I would recommand to return the page you want to redirect to : 
```
this.keyQuery.selectLoading().pipe(
  ...
  filter(wallet => wallet.ensDomain ? wallet : 'redirectedpage')
).subscribe({
  next: (res: Wallet | string) => typeof res === "string" ? res(this.router.parseUrl(res) : res(true)
})
```

libs/ethers/src/lib/wallet/guards/wallet-key.guard.ts
GrandSchtroumpf: Update comment to be `wait until we retrieve something from localstorage`.

PR #667, opened 2019-07-19
Issue#572: send verification mail on account creation

lsenta:


libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -81,28 +82,33 @@ export class AuthService {
     await this.db.doc<User>(`users/${uid}`).delete();
   }
 
+  public async sendVerifyEmail() {
+    return this.afAuth.auth.currentUser.sendEmailVerification();
+  }
+
   /** Deletes user subCollections */
-  private async _deleteSubCollections (uid) {
+  private async _deleteSubCollections(uid) {
```
GrandSchtroumpf: Please remove "_". private flag is here for that. also type your parameter.

libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -81,28 +82,33 @@ export class AuthService {
...
     // @todo check if user is the only member of org (and the only ADMIN)
     // @todo remove uid from org.userIds
     const permissions = await this._getUserSubcollectionItems(uid, 'permissions');
-    return await permissions.map(o => this.db.doc<User>(`users/${uid}`)
-      .collection('permissions')
-      .doc(o.id)
-      .delete()
+    return Promise.all(
+      permissions.map(o =>
+        this.db
```
GrandSchtroumpf: For consistency: 
```typescript
permissions.map(({id}) => this.db.doc(`users/${uid}/permissions/${id}`).delete())
```

libs/auth/src/lib/+state/auth.service.ts
lsenta: C'est bien parcque c'est toi je suis passé dessus pour virer les bugs et fixer le style ;)

libs/auth/src/lib/+state/auth.service.ts
lsenta: Idem ;)

libs/auth/src/lib/+state/auth.service.ts
GrandSchtroumpf: I know. But we need to clean the code bit by bit. I think it's kind of house keeping in a flatshare ;).
I'm glad you went there to fixe the style so I can see that ! 👍

PR #668, opened 2019-07-19
Issue 576: org invite a user (backend part mostly)

lsenta:
- [x] the org create the user if it does not exists
- [x] the org writes to the user invitations
- [x] a function is triggered to generate a notification 
- [x] and send a mail
- [x] on accept, a function is triggered that updates the user
- [x] also updates the org
- [x] also updates the permissions

Future PR will take care of the frontend:
- [ ] the invitation is displayed as a "pending invites"
- [ ] the user can accept the invitation,

### DIFF
--- a/apps/backend-functions/src/assets/mailTemplates.ts
+++ b/apps/backend-functions/src/assets/mailTemplates.ts
@@ -1,8 +1,9 @@
 import { appUrl } from '../environments/environment';
+import { EmailRequest } from '../internals/email';
 
 const USER_WELCOME_PATH = '/auth/welcome';
 
-export const userInviteTemplate = ({ email, password }: { email: string; password: string }) =>
+const userInviteTemplate = ({ email, password }: { email: string; password: string }) =>
   `
   You've been invited to a project on the Blockframes Platform!\n
   \n
@@ -12,3 +13,19 @@ export const userInviteTemplate = ({ email, password }: { email: string; passwor
   Click on the following link to login to your account and join the project:\n
   ${appUrl}${USER_WELCOME_PATH}
   `;
+
+export function userInvite(email: string, password: string): EmailRequest {
+  return {
+    to: email,
+    subject: 'Your Blockframes account is waiting for you',
+    text: userInviteTemplate({ email, password })
+  };
+}
+
+export function userInviteToOrg(email: string): EmailRequest {
+  return {
+    to: email,
+    subject: 'You have been invited to an organization',
+    text: 'TODO'
+  };
+}
--- a/apps/backend-functions/src/assets/mailTemplates.ts
+++ b/apps/backend-functions/src/assets/mailTemplates.ts
@@ -1,3 +1,6 @@
+/**
+ * Templates for transactional emails we send to user and to cascade8 admins.
+ */
 import { appUrl } from '../environments/environment';
 import { EmailRequest } from '../internals/email';
 
@@ -14,6 +17,7 @@ const userInviteTemplate = ({ email, password }: { email: string; password: stri
   ${appUrl}${USER_WELCOME_PATH}
   `;
 
+/** Generates a transactional email request for user invited to the application. */
 export function userInvite(email: string, password: string): EmailRequest {
   return {
     to: email,
@@ -22,6 +26,7 @@ export function userInvite(email: string, password: string): EmailRequest {
   };
 }
 
+/** Generates a transactional email request for user invited to an organization. */
 export function userInviteToOrg(email: string): EmailRequest {
   return {
     to: email,
--- a/apps/backend-functions/src/data/types.ts
+++ b/apps/backend-functions/src/data/types.ts
@@ -135,6 +135,9 @@ interface RawInvitation {
   processedId?: string;
 }
 
+/**
+ * Invite a stakeholder to work on a document.
+ */
 export interface InvitationStakeholder extends RawInvitation {
   type: InvitationType.stakeholder;
   docId: string;
@@ -142,12 +145,19 @@ export interface InvitationStakeholder extends RawInvitation {
   organizationId: string;
 }
 
+/**
+ * Invite a user to an organization.
+ */
 export interface InvitationToOrganization extends RawInvitation {
   type: InvitationType.toOrganization;
   userId: string;
   organizationId: string;
 }
 
+/**
+ * This is the generic type for invitation,
+ * use the type field to figure out which kind of invitation you are working with.
+ */
 export type Invitation = InvitationStakeholder | InvitationToOrganization;
 export type InvitationOrUndefined = Invitation | undefined;
 
--- a/apps/backend-functions/src/data/types.ts
+++ b/apps/backend-functions/src/data/types.ts
@@ -78,6 +78,9 @@ export interface Notification extends BaseNotification {
   appIcon: AppIcon;
 }
 
+export type InvitationState = 'accepted' | 'declined' | 'pending';
+export type InvitationType = 'orgInvitation';
+
 export interface BaseInvitation {
   message: string;
   userInformations?: {
@@ -86,6 +89,8 @@ export interface BaseInvitation {
     surname?: string;
     email: string
   };
+  userId: string;
+  type?: InvitationType;
   docInformations: DocInformations;
   organizationId: string;
   path?: string;
@@ -93,7 +98,7 @@ export interface BaseInvitation {
 
 export interface Invitation extends BaseInvitation {
   id: string;
-  state: 'accepted' | 'declined' | 'pending';
+  state: InvitationState;
   date: FirebaseFirestore.FieldValue;
   appIcon: AppIcon;
   processedId?: string;
--- a/apps/backend-functions/src/data/types.ts
+++ b/apps/backend-functions/src/data/types.ts
@@ -147,3 +147,5 @@ export interface UserDocPermissions {
 export function asIDMap<T extends DocWithID>(items: T[]): IDMap<T> {
   return items.reduce((result, item) => ({ ...result, [item.id]: item }), {});
 }
+
+export type InvitationOrUndefined = Invitation | undefined;
--- a/apps/backend-functions/src/data/types.ts
+++ b/apps/backend-functions/src/data/types.ts
@@ -97,27 +97,27 @@ export interface UserDocPermissions {
 // Internal Interaction Types
 // ==========================
 
-export enum App {
+export const enum App {
   main = 'main',
   mediaDelivering = 'media_delivering',
   mediaFinanciers = 'media_financiers'
 }
 
 // Legacy for compat between Notifications & Invitations
-// TODO: use App everywhere and let the frontend / concrete code deal with
-//       the app specifics (icons, message, etc).
+// TODO(issue#684): use App everywhere and let the frontend / concrete
+//  code deal with the app specifics (icons, message, etc).
 export type AppIcon = App;
 
 // Invitations
 // -----------
 
-export enum InvitationState {
+export const enum InvitationState {
   accepted = 'accepted',
   declined = 'declined',
   pending = 'pending'
 }
 
-export enum InvitationType {
+export const enum InvitationType {
   stakeholder = 'stakeholder',
   toOrganization = 'toOrganization'
 }
@@ -135,9 +135,7 @@ interface RawInvitation {
   processedId?: string;
 }
 
-/**
- * Invite a stakeholder to work on a document.
- */
+/** Invite a stakeholder to work on a document. */
 export interface InvitationStakeholder extends RawInvitation {
   type: InvitationType.stakeholder;
   docId: string;
@@ -145,9 +143,7 @@ export interface InvitationStakeholder extends RawInvitation {
   organizationId: string;
 }
 
-/**
- * Invite a user to an organization.
- */
+/** Invite a user to an organization. */
 export interface InvitationToOrganization extends RawInvitation {
   type: InvitationType.toOrganization;
   userId: string;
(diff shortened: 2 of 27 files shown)


apps/backend-functions/src/assets/mailTemplates.ts
```diff
@@ -12,3 +13,19 @@ export const userInviteTemplate = ({ email, password }: { email: string; passwor
   Click on the following link to login to your account and join the project:\n
   ${appUrl}${USER_WELCOME_PATH}
   `;
+
+export function userInvite(email: string, password: string): EmailRequest {
```
GrandSchtroumpf: Add comments on your functions

apps/backend-functions/src/data/types.ts
```diff
@@ -5,17 +5,30 @@
...
+// Low Level Types
+// ===============
+
 export type IDMap<T> = Record<string, T>;
 
 interface DocWithID {
   id: string;
 }
 
+export enum DocType {
```
GrandSchtroumpf: Use `export const enum DocType` instead. Compiled result is lighter.

apps/backend-functions/src/data/types.ts
```diff
@@ -63,45 +76,113 @@ export interface Material {
...
+  canCreate: string[];
+  canDelete: string[];
+  canRead: string[];
+  canUpdate: string[];
+}
+
+// Internal Interaction Types
+// ==========================
+
+export enum App {
```
GrandSchtroumpf: Same here, and below

apps/backend-functions/src/data/types.ts
```diff
@@ -63,45 +76,113 @@ export interface Material {
...
+  main = 'main',
+  mediaDelivering = 'media_delivering',
+  mediaFinanciers = 'media_financiers'
+}
+
+// Legacy for compat between Notifications & Invitations
+// TODO: use App everywhere and let the frontend / concrete code deal with
```
GrandSchtroumpf: Add an issue for that.

apps/backend-functions/src/data/types.ts
```diff
@@ -63,45 +76,113 @@ export interface Material {
...
+  userId: string;
+  organizationId: string;
+}
+
+/**
+ * This is the generic type for invitation,
+ * use the type field to figure out which kind of invitation you are working with.
+ */
+export type Invitation = InvitationStakeholder | InvitationToOrganization;
+export type InvitationOrUndefined = Invitation | undefined;
```
GrandSchtroumpf: You should not have type what can be undefined. This should be used directly at the variable/parameter/output level.

apps/backend-functions/src/invitation.ts
```diff
@@ -1,105 +1,246 @@
...
+    return;
+  }
+
+  const userRef = db.collection('users').doc(userId);
+  const organizationRef = db.collection('orgs').doc(organizationId);
+  const permissionsRef = db.collection('permissions').doc(organizationId);
+  const invitationRef = db.collection('invitations').doc(invitation.id);
+
+  return db.runTransaction(async tx => {
+    const [user, org, perm] = await Promise.all([
```
ioaNikas: For easier reading, it's better to use the full name of `organization`. I would say the same for `permissions`.

apps/backend-functions/src/invitation.ts
```diff
@@ -1,105 +1,246 @@
...
+      tx.get(userRef),
+      tx.get(organizationRef),
+      tx.get(permissionsRef)
+    ]);
+
+    const userData = user.data();
+    const orgData = org.data();
```
ioaNikas: Same here (and the nexts)

apps/backend-functions/src/invitation.ts
```diff
@@ -1,105 +1,246 @@
...
+      tx.delete(invitationRef)
+    ]);
+  });
+}
+
+async function onInvitationToOrgCreate({ userId }: InvitationToOrganization) {
+  const userMail = await getUserMail(userId);
+
+  if (!userMail) {
+    console.error('No user email provided for userId:', userId);
```
ioaNikas: Space is missing before closing quote, sweet revenge ! :smile:

apps/backend-functions/src/stakeholder.ts
```diff
@@ -97,17 +113,17 @@ async function stakeholdersCollectionEvent(
       };
 
       const notifications = createNotifications(organizations, snapInformations);
-      triggerNotifications(notifications);
+      const invitation = createInvitation(organizations, snapInformations);
 
-      const invitations = createInvitations(organizations, snapInformations);
       // TODO: trigger only one invitation shared among all the organization members (loaded with user.orgId in the frontend) => ISSUE#645
```
ioaNikas: This issue has been solved, forgot to delete the TODO, sorry about that.

apps/backend-functions/src/stakeholder.ts
```diff
@@ -120,39 +136,56 @@ function createNotifications(organizations: Organization[], snap: SnapObject) {
   const path = !!snap.delivery
     ? `/layout/o/${snap.delivery.movieId}/${snap.delivery.id}/teamwork`
     : `/layout/o/home/${snap.movie.id}/teamwork`;
+
+  const isNotTheInvitedMember = (organization: Organization): boolean => {
```
ioaNikas: For clearer wording, should we call it `isNotTheInvitedStakeholder` as those notifications are send to all users of an organization/stakeholder ?

apps/backend-functions/src/invitation.ts
```diff
@@ -1,105 +1,246 @@
...
+  Organization
+} from './data/types';
 import { prepareNotification, triggerNotifications } from './notify';
+import { sendMail } from './internals/email';
+import { userInviteToOrg } from './assets/mailTemplates';
+
+/**
+ * @param before
+ * @param after
+ * @returns whether the invitation just got accepted.
```
GrandSchtroumpf: Better to put the description on the first line. The @returns is used mainly by javascript to give the output type. Or in Typescript if return value has everal meanings not explicit in the description : 
```
/**
 * Check if invitation has been accepted.
 * @param before The previous state of the invitation
 * @param after The current state of the invitation
 */
```

apps/backend-functions/src/internals/firebase.ts
```diff
@@ -11,3 +11,8 @@ export const serverTimestamp = admin.firestore.FieldValue.serverTimestamp;
 export const getBackupBucketName = (): string => backupBucket; // TODO: secure bucket
 
 export { admin, functions };
+
+export async function getUserMail(userId: string): Promise<string | undefined> {
```
GrandSchtroumpf: I think this is the perfect example where I would put a `@returns` comment that explains what it means when the return value is undefined.

apps/backend-functions/src/invitation.ts
```diff
@@ -1,105 +1,246 @@
...
+/**
+ * @param before
+ * @param after
+ * @returns whether the invitation just got created
+ */
+function wasCreated(before: InvitationOrUndefined, after: Invitation) {
+  return !before && !!after;
+}
+
+async function onInvitationToOrgAccept(invitation: InvitationToOrganization) {
```
GrandSchtroumpf: Comment would be nice here.

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -24,43 +25,27 @@ export class OrganizationService {
...
-        const organizationTransaction = tx.update(organizationDoc.ref, { userIds: nextUserIds });
-        // Update the permissions and add the new member as an org admin
-        const nextAdminsIds = [...permissions.admins, member.uid];
-        const permissionsTransaction = tx.update(permissionsDoc.ref, { admins: nextAdminsIds });
-        // Update the user
-        // TODO: Move this to the user side as we shouldn't be authorized to write in user document if we're not the concerned user
-        const updateUserTransaction = tx.update(userDoc.ref, { orgId });
+    // TODO: use a definitive data type
+    // TODO: compare with backend-functions
+    const invitation = { userId: uid, orgId, type: 'orgInvitation', state: 'pending' };
```
ioaNikas: Frontend and backend types should match. Would be better to have an enum for `state` in this invitation.model too ?

apps/backend-functions/src/invitation.ts
```diff
@@ -1,105 +1,246 @@
...
-        ]);
-      });
-      return true;
-    } catch (error) {
-      await db.doc(`invitations/${invitation.id}`).update({ processedId: null });
-      throw error;
+  try {
+    switch (invitation.type) {
+      case InvitationType.stakeholder:
+        return await onStakeholderInvitationUpdate(invitationDocBefore, invitationDoc, invitation);
```
GrandSchtroumpf: I don't think you need await because you're not doing anything after that.

apps/backend-functions/src/stakeholder.ts
```diff
@@ -1,41 +1,53 @@
...
+  DocInformations,
+  DocType,
+  InvitationStakeholder,
+  Movie,
+  Organization,
+  SnapObject,
+  Stakeholder
+} from './data/types';
+
+const COLLECTION_DELIVERIES = 'deliveries';
```
GrandSchtroumpf: Maybe a enum with the collection names ?

apps/backend-functions/src/stakeholder.ts
```diff
@@ -45,26 +57,25 @@ export async function onMovieStakeholderDelete(
...
-
-  const newStakeholder = snap.data();
+  const newStakeholder = snap.data() as Stakeholder | undefined;
 
   if (!newStakeholder) {
     throw new Error(`New stakeholder not found !`);
   }
 
-  const collection = !!context.params.movieID ? 'movies' : 'deliveries'; // TODO : Change to switch for 3 or more options.
+  // TODO: extract semi-generic part, too many if then else, won't work with more types.
```
GrandSchtroumpf: Add an issue for this todo

apps/backend-functions/src/stakeholder.ts
```diff
@@ -97,17 +113,17 @@ async function stakeholdersCollectionEvent(
...
 
       const notifications = createNotifications(organizations, snapInformations);
-      triggerNotifications(notifications);
+      const invitation = createInvitation(organizations, snapInformations);
 
-      const invitations = createInvitations(organizations, snapInformations);
       // TODO: trigger only one invitation shared among all the organization members (loaded with user.orgId in the frontend) => ISSUE#645
-      triggerInvitations(invitations);
-
+      return await Promise.all([
```
GrandSchtroumpf: You don't need await here.

apps/backend-functions/src/stakeholder.ts
```diff
@@ -120,39 +136,56 @@ function createNotifications(organizations: Organization[], snap: SnapObject) {
...
+        organizationId: snap.newStakeholderId,
         docInformations: snap.docInformations
       });
     });
 }
 
+/**
+ * Temporary function, extract an app from a snap object,
+ *
+ * TODO: refactor out the SnapObject and fix typing to keep application
```
GrandSchtroumpf: Add issue for TODO

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -24,43 +25,27 @@ export class OrganizationService {
...
+    private authService: AuthService,
     private db: FireQuery
   ) {}
 
-  /** Add a new user to the organization */
-  public async addMember(member: OrganizationMember) {
+  /**
+   * Add a new user to the organization
+   *
+   * @param member
```
GrandSchtroumpf: If you don't give a desciption for the param, it's better to keep comment on one line.

apps/backend-functions/src/data/types.ts
lsenta: DataOrUndefined is an expected type to move around when you're working with firebase events.
I can replace all usage by `(Invitation | undefined)` but readability is going to be terrible.

apps/backend-functions/src/internals/firebase.ts
lsenta: Can't happen, the function will throw, I'll add documentation.

apps/backend-functions/src/invitation.ts
lsenta:
```
console.log('space', 'are', 'added', 'automatically', '😅')
> space are added automatically 😅
```

apps/backend-functions/src/invitation.ts
lsenta: We need to keep exceptions within the scope of the current `try/catch` to make sure we can log and update the processed id (l.243). => await is required, else the promise will bubble up and be handled at the firebase functions runtime.

apps/backend-functions/src/stakeholder.ts
lsenta: Later, with #686 686

apps/backend-functions/src/stakeholder.ts
lsenta: Same as before, usually if you protect your code with a `try/catch` and manage errors, you'll await any promises produced within the try catch. Else the promise will exit the `try/catch` before the exception is triggered.

apps/backend-functions/src/invitation.ts
ioaNikas: :angry:

libs/organization/src/lib/+state/organization.service.ts
lsenta: Later when I get to implement the frontend part and test it.

apps/backend-functions/src/invitation.ts
lsenta: nice try ;)

PR #669, opened 2019-07-22
add wallet import tunnel & refactoring

pldespaigne:


libs/ethers/src/lib/wallet/+state/wallet.store.ts
```diff
@@ -2,6 +2,12 @@ import { Injectable } from '@angular/core';
 import { Store, StoreConfig } from '@datorama/akita';
 import { Tx } from '../../types';
 
+export interface InputMessage {
```
GrandSchtroumpf: Find a better name. I don't understand what it does when reading "InputMessage". What kind of message ?

libs/ethers/src/lib/wallet/wallet-add-key/wallet-add-key.component.html
```diff
@@ -1,11 +1,10 @@
-<main [ngSwitch]="step">
-  <h1>Create a new Key</h1>
+<main [ngSwitch]="step" fxLayout="column" fxLayoutAlign="start center">
   <section *ngSwitchCase="steps.password">
     <ng-container *ngIf="!(loading$ | async) as loading; else: isLoading">
```
GrandSchtroumpf: Change "loading" to "encrypting" in your case.

libs/ethers/src/lib/wallet/wallet-add-key/wallet-add-key.component.ts
```diff
@@ -26,6 +26,11 @@ export class WalletAddKeyTunnelComponent implements OnInit {
   loading$ = new Observable<boolean>();
   redirectRoute: string;
   @ViewChild('downloadLink', {static: false}) downloadLink: ElementRef<HTMLAnchorElement>;
```
GrandSchtroumpf: Put ViewChild *after* all properties

libs/ethers/src/lib/wallet/wallet-add-key/wallet-add-key.component.ts
```diff
@@ -26,6 +26,11 @@ export class WalletAddKeyTunnelComponent implements OnInit {
...
+  message = <InputMessage> {
```
GrandSchtroumpf: Why do you hardcode that ?
Do `message: InputMessage = { ... }`

libs/ethers/src/lib/wallet/wallet-create-password-form/wallet-create-password-form.component.html
```diff
@@ -1,13 +1,11 @@
 <section fxLayout="row" fxLayoutAlign="center" fxLayoutGap="80px">
   <img src="/assets/images/key_lock_laptop.png" alt="Laptop which is protected by a lock" />
-  <article fxLayout="column" fxLayoutAlign="center center">
-    <h1>Confirm your key</h1>
+  <article fxFlex="20%" fxLayout="column" fxLayoutAlign="center center">
```
GrandSchtroumpf: Avoid using fxFlex. Keep sizing in the CSS and layouting in the HTML. Like that if we want to find it it's easier.
Also, most of the time `fxFlex` is missused, as it's relative to the siblings and not the parent. Prefer width/height most of the time.

libs/ethers/src/lib/wallet/wallet-create-password-form/wallet-create-password-form.component.html
```diff
@@ -1,13 +1,11 @@
...
+    <h2>Choose a key Password</h2>
     <p>
       Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
       labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco
-      laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in
-      voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat
-      non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+      laboris nisi ut aliquip ex ea commodo consequat.
```
GrandSchtroumpf: Why do you make it smaller ? Does it means that the view doesn't work when there is too much text ?
I would recommend to set the real text as soon as possible.

libs/ethers/src/lib/wallet/wallet-create-password-form/wallet-create-password-form.component.scss
```diff
@@ -1,4 +1,11 @@
-img {
-  width: 100%;
-  height: auto;
+section {
+  margin-top: 15%;
```
GrandSchtroumpf: Avoid % in margin. This will break in responsive most of the time.

libs/ethers/src/lib/wallet/wallet-create-password-form/wallet-create-password-form.component.scss
```diff
@@ -1,4 +1,11 @@
...
+  img {
+    width: 410px;
+    height: auto;
```
GrandSchtroumpf: This is the default you don't need it.

libs/ethers/src/lib/wallet/wallet-import-key/wallet-import-key.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  templateUrl: './wallet-import-key.component.html',
+  styleUrls: ['./wallet-import-key.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class WalletImportKeyComponent implements OnInit {
+
+  steps = steps;
+  step = this.steps.import;
+  mnemonic: string;
+  message = <InputMessage> {
```
GrandSchtroumpf: Same as above.

libs/ethers/src/lib/wallet/wallet-import-key/wallet-import-key.component.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  handleImportMnemonic(mnemonic: string) {
+    this.mnemonic = mnemonic;
+    this.step = this.steps.password;
+  }
+
+  async handlePassword(encryptionPassword: string) {
+    const { ensDomain } = this.walletQuery.getValue();
+    const keyName = this.service.getDefaultKeyName(ensDomain);
+    this.service.importFromMnemonic(keyName, ensDomain, this.mnemonic, encryptionPassword)
```
GrandSchtroumpf: Use await.

libs/ethers/src/lib/wallet/wallet-message-view/wallet-message-view.component.html
```diff
@@ -2,6 +2,6 @@
   <h1>{{ message.headline }}</h1>
   <p>{{ message.subline }}</p>
   <img src="/assets/images/ppl_celebrating.png" alt="People celebrating" />
-  <a mat-button (click)="redirectUser.emit()">Okay</a>
+  <button mat-raised-button color="accent" (click)="redirectUser.emit()">Okay</button>
```
GrandSchtroumpf: Use "a" because it'll change the path.

libs/ethers/src/lib/wallet/wallet-add-key/wallet-add-key.component.ts
pldespaigne: When you send a tx to the blockchain this message is stored in the wallet state (under the tx object), but here we are just creating/importing locally, so the message depend only on witch tunnel is the parent component, moreover because create/import works only with the key-manager state and the tx.message is in the wallet state it will require to import wallet from key-manager causing a circular dependency

libs/ethers/src/lib/wallet/+state/wallet-known-tx.ts
```diff
@@ -1,27 +1,48 @@
...
+    isError: false,
+  };
+  
+  return <LocalTx> {...tx, message, callback};
 }
 
-export function getDeleteKeyTx(erc1077Address:string, pubKey: string): Tx {
-  return {
+export function getDeleteKeyTx(erc1077Address:string, pubKey: string, callback: (...args) => void): LocalTx {
+  const tx = <Tx> {
```
GrandSchtroumpf: do `const tx: Tx = ...`

libs/ethers/src/lib/wallet/+state/wallet-known-tx.ts
```diff
@@ -1,27 +1,48 @@
-import { Tx } from "../../types";
+import { LocalTx, Tx, InformationMessage } from "../../types";
 
-export function getMockTx(): Tx {
-  return {
+export function getMockTx(callback: (...args) => void): LocalTx {
```
GrandSchtroumpf: Add comments. Is it supposed to be removed ? When ? Which issue ?
If not, don't use the word "mock"

libs/ethers/src/lib/wallet/+state/wallet-known-tx.ts
```diff
@@ -1,27 +1,48 @@
...
+    subline: 'Your transaction was successfully sent to the Blockchain !',
+    isError: false,
+  };
+  
+  return <LocalTx> {...tx, message, callback};
 }
 
-export function getDeleteKeyTx(erc1077Address:string, pubKey: string): Tx {
-  return {
+export function getDeleteKeyTx(erc1077Address:string, pubKey: string, callback: (...args) => void): LocalTx {
```
GrandSchtroumpf: Add some comments

libs/ethers/src/lib/wallet/+state/wallet-known-tx.ts
```diff
@@ -1,27 +1,48 @@
...
+  const tx = <Tx> {
     to: erc1077Address,
     value: '0x0',
     data: `0x81dd7b8e000000000000000000000000${pubKey.substr(2)}0000000000000000000000000000000000000000000000000000000000000001`
```
GrandSchtroumpf: What is this data. Impossible to understand for any developer who looks at the code.

libs/ethers/src/lib/wallet/wallet-import-key/wallet-import-key.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  templateUrl: './wallet-import-key.component.html',
+  styleUrls: ['./wallet-import-key.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class WalletImportKeyComponent implements OnInit {
+
+  steps = steps;
+  step = this.steps.import;
+  mnemonic: string;
+  isEncrypting$ = new Observable<boolean>();
```
GrandSchtroumpf: Hmmm, I think you wanted to do : `isEncrypting$: Observable<boolean>;`

libs/ethers/src/lib/wallet/wallet-import-key/wallet-import-key.component.ts
```diff
@@ -0,0 +1,67 @@
...
+    const key = await this.service.importFromMnemonic(keyName, ensDomain, this.mnemonic, encryptionPassword);
+    this.service.storeKey(key);
+    this.step = this.steps.end;
+  }
+
+  handleRedirection() {
+    this.router.navigateByUrl('/layout/o/account/wallet'); // TODO handle dynamic redirect from state : issue#617
+  }
+
+  get message() {
```
GrandSchtroumpf: `get message(): InformationMessage {...}`

PR #685, opened 2019-07-25
service functions & mock data

pldespaigne:


libs/organization/src/lib/+state/organization.model.ts
```diff
@@ -1,9 +1,27 @@
 export interface OrgMember {
```
GrandSchtroumpf: Call it OrganizationMember

libs/organization/src/lib/+state/organization.model.ts
```diff
@@ -1,9 +1,27 @@
...
+  avatar?: string;
   uid: string;
+  name?: string;
   email: string;
   roles: string[];
 }
 
+export interface OrgOperation {
```
GrandSchtroumpf: OrganizationOperation

libs/organization/src/lib/+state/organization.model.ts
```diff
@@ -1,9 +1,27 @@
...
+  id: string;
+  name: string;
+	quorum: number;
+	members: OrgMember[]
+}
+
+export interface OrgAction {
```
GrandSchtroumpf: OrganizationAction

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -103,4 +103,122 @@ export class OrganizationService {
...
+    newOperations.push(newOperation); // we could not use `actions.push(newAction)` direclty otherwise the action will have been duplicated
+
+    try {
+      // send tx to the org smart-contract and wait for result // TODO replace with the real implemntation : issue 676
+
+      // update the store
+      this.store.update(state => {
+        return {
+          ...state, // keep everything of the state
+          org: { // update only the org
```
GrandSchtroumpf: Too many comments ^^.
```typescript
this.store.update(state => ({
  ...state, // keep everything of the state
 org: { ...state.org, operations: newOperations }  // update the actions array
}))
```

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -103,4 +103,122 @@ export class OrganizationService {
...
   }
+
+  //-------------------------------------------------
+  //            BLOCKCHAIN PART OF ORGS
+  //-------------------------------------------------
+
+  /** create a newOperation, or update it if it already exists */
+  private async upsertOperation(newOperation: OrgOperation) {
+    const { operations } = this.query.getValue().org // get every actions
+    const newOperations = operations.filter(currentOperation => currentOperation.id !== newOperation.id); // get all actions except the one we want to upsert
```
GrandSchtroumpf: Avoid mutation when possible : 
```typescript
const newOperations = [
  ...operations.filter(currentOperation => currentOperation.id !== newOperation.id), all actions except the one we want to upsert
  newOperation // updated action
]
```

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -103,4 +103,122 @@ export class OrganizationService {
...
+      });
+    } catch(err) {
+      console.error('The transaction has failed :', err); // TODO better error handling : issue 671
+    }
+  }
+
+  updateOperationQuorum(id: string, newQuorum: number) {
+    const operation = this.query.getOperationById(id);
+    if(!operation) throw new Error('This operation doesn\'t exists');
+    const newOperation = {
```
GrandSchtroumpf: I think here the intermediary variable is too much

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -103,4 +103,122 @@ export class OrganizationService {
...
+      quorum: newQuorum,
+    };
+    this.upsertOperation(newOperation);
+  }
+
+  addOperationMember(id: string, newMember: OrgMember) {
+    const operation = this.query.getOperationById(id);
+    if(!operation) throw new Error('This operation doesn\'t exists');
+    
+    const [memberExists] = operation.members.filter(member => member.uid === newMember.uid);
```
GrandSchtroumpf: Use find when you're look for one: 
```typescript
const memberExists = operation.members.find(member => member.uid === newMember.uid);
```
BUT in your case you want to check if the array `includes` the member
```typescript
const memberExists = operation.members.includes(member => member.uid === newMember.uid); // boolean
```

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -103,4 +103,122 @@ export class OrganizationService {
...
+    if (!!memberExists) throw new Error('This member is already a signer of this operation');
+
+    const newOperation = {
```
GrandSchtroumpf: same as above

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -103,4 +103,122 @@ export class OrganizationService {
...
+    };
+    
+    this.upsertOperation(newOperation);
+  }
+
+  removeOperationMember(id: string, memberToRemove: OrgMember) {
+    const operation = this.query.getOperationById(id);
+    if(!operation) throw new Error('This operation doesn\'t exists');
+    
+    const newMembersList = operation.members.filter(member => member.uid !== memberToRemove.uid);
```
GrandSchtroumpf: call it members : 
```typescript
const members = operation.members.filter(member => member.uid !== memberToRemove.uid);
this.upsertOperation({...operation, members });
```

libs/organization/src/lib/+state/organization.service.ts
pldespaigne: `includes()` cannot take an arrow function as a parameters, it accept only a value, and perform a value equality test, witch will not work on objects.
BUT in my case I think I can use `some() => boolean`
```typescript
const memberExists = operation.members.some(member => member.uid === newMember.uid);
```

What do you think ?

GrandSchtroumpf: Yes `some` is good.

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -103,4 +103,115 @@ export class OrganizationService {
...
+    if(!operation) throw new Error('This operation doesn\'t exists');
+    
+    const members = operation.members.filter(member => member.uid !== memberToRemove.uid);
+    const newOperation = { ...operation, members };
+    
+    this.upsertOperation(newOperation);
+  }
+
+  // TODO REMOVE THIS ASAP : issue 676
+  public instantiateMockData() {
```
GrandSchtroumpf: Put the mocked value inside a .mock.ts file inside +state folder.

Issue #571, opened 2019-07-10
3. Dapp installation


Jackseed:
- [ ] When an org opt in for a dapp, C8 receives an email
- [ ] When an org acquires a new dapp (accepted by C8), admins of the org receive an email

lsenta:
- [x] add state pending during organization creation
- [x] send a mail to cascade8 admins with a link to activate the organization
  - [x] configure admin mail in functions
  - [x] trigger on org creation
  - [x] send the mail
- [x] raw page to activate the organization
  - [x] create the function per org with a page on GET
  - [x] create the function per org with a reaction on POST
  - [x] send a mail to org's admin when the org is accepted

We'll refactor the express app with the next admin feature (issue #571).

In a future PR we'll deal with the frontend:
- [ ] update the frontend to display the org in pending

In a future PR we'll deal with the rules:
- [ ] update the rules to block organization actions until state is accepted
- [ ] update the rules to allow organization creation if and only if its state is pending

lsenta:
Current PR with the MVP: https://github.com/blockframes/blockframes/pull/698

After a first review:

- [ ] secure the request apps collection (only allow org admins to create requests) #697
- [ ] secure the admin pages #700
- [ ] Implement a nicer design
- [ ] Refactor the data types so they can be shared between the back and front
- [ ] show "request to app" only to admins (OrganizationGuard)
- [ ] guard each app if and only if the org has access to the app

PR #690, opened 2019-07-26
Issue 573: Add Organization's pending state and implement admin flow

lsenta:
- [x] add state pending during organization creation
- [x] send a mail to cascade8 admins with a link to activate the organization
  - [x] configure admin mail in functions
  - [x] trigger on org creation
  - [x] send the mail
- [x] raw page to activate the organization
  - [x] create the function per org with a page on GET
  - [x] create the function per org with a reaction on POST
  - [x] send a mail to org's admin when the org is accepted

We'll refactor the express app with the next admin feature (issue #571).

In a future PR we'll deal with the frontend:
- [ ] update the frontend to display the org in pending

In a future PR we'll deal with the rules:
- [ ] update the rules to block organization actions until state is accepted
- [ ] update the rules to allow organization creation if and only if its state is pending


apps/backend-functions/src/assets/adminTemplates.ts
```diff
@@ -0,0 +1,28 @@
+/**
```
GrandSchtroumpf: for the naming of the files we use kebab-case format. Could you change to `admin-templates.ts` ? Same for the others.

apps/backend-functions/src/assets/mailTemplates.ts
```diff
@@ -17,6 +18,13 @@ const userInviteTemplate = ({ email, password }: { email: string; password: stri
   ${appUrl}${USER_WELCOME_PATH}
   `;
 
+const organizationCreateTemplate = (orgId: string): string =>
+  `
+  A new organization was created on the blockframes project,\n
+  \n
```
GrandSchtroumpf: I think `\n` is redundant with backquotes. Have you tried ? Also when using backquotes you'll have the format kept (tabs included).

apps/backend-functions/src/assets/mailTemplates.ts
```diff
@@ -34,3 +42,21 @@ export function userInviteToOrg(email: string): EmailRequest {
     text: 'TODO'
   };
 }
+
+/** Generates a transactional email request to let cascade8 admin know that a new org is waiting for approval. */
+export function organizationCreate(orgId: string): EmailRequest {
```
GrandSchtroumpf: I think you mean `organizationCreated` with a 'd' in the end.

apps/backend-functions/src/orgs.ts
```diff
@@ -50,3 +59,52 @@ export function onOrganizationDelete(
...
+    return sendMail(organizationWasAccepted(email));
+  });
+}
+
+// We serve an express app in the function
+// this let us deal easily with get / post, url params, etc.
+export const onAcceptNewOrg = express();
+
+// When an admin access the page, they'll see the "accept org" form.
+onAcceptNewOrg.get(
```
GrandSchtroumpf: I think it would be more readable with the route at the same line for `get(`. But it's personal point of view here.

apps/backend-functions/src/orgs.ts
```diff
@@ -50,3 +59,52 @@ export function onOrganizationDelete(
...
+// Organization Administration: Accept new orgs
+// ============================================
+
+/** Update an organization when it has been accepted by admins. */
+function acceptOrganization(organizationRef: DocumentReference): Promise<any> {
+  return organizationRef.update({ status: OrganizationStatus.accepted });
+}
+
+/** Send an email to organization admins when it has been accepted. */
+async function mailOrganizationAdminOnAccept(organizationRef: DocumentReference): Promise<any> {
```
GrandSchtroumpf: type at least the return value as an array `Promise<any[]>`. Best would be `Promise<(undefined | ***)[]>`

firestore.rules
```diff
@@ -20,8 +20,10 @@ service cloud.firestore {
     /// ORGANIZATION RULES ///
 
     match /orgs/{orgId} {
+        // TODO: allow create only if state is pending.
```
GrandSchtroumpf: add an issue Id for your TODO

libs/organization/src/lib/+state/organization.model.ts
```diff
@@ -1,3 +1,8 @@
+export enum OrganizationStatus {
```
GrandSchtroumpf: Use `const enum`, smaller bundle size.

apps/backend-functions/src/orgs.ts
lsenta: Prettier

apps/backend-functions/src/orgs.ts
lsenta: We don't use the return value it's 100% side effecting code (we only care about errors), I'd rather not overtype this.

GrandSchtroumpf: I would recommend to use `forEach` + not returning anything then. Like that we know when reading the code that it's side effect stuff.

apps/backend-functions/src/assets/mailTemplates.ts
lsenta: SGTM, I'll remove the useless `\n`,
for the tabs: we'll rewrite the content anyway when we test & update on the copywriting. Favouring readability for ow.

apps/backend-functions/src/orgs.ts
lsenta: Bad practice, we'll lose the exceptions and the function might get killed before the promises are finished. Always await promises.

I'll update the code though, their should be a `Promise.all` here!

Issue #697, opened 2019-07-29
Finalize permissions for the release


lsenta:

lsenta:
block orgs with pending status (https://app.zenhub.com/workspaces/blockframes-5c6a8ada8de1b31d5443abd7/issues/blockframes/blockframes/573)

Jackseed:
Update with #1097

PR #698, opened 2019-07-29
Issue 571: request and accept access to apps (minimal version)

lsenta:
- [x] add basic UI to show app access for your organization
- [x] send c8 admin's an email on request
- [x] implements a basic admin page to allow access to app for an org
- [x] move app request code to organization store
- [x] update the data document with `app-requests` right before merging

Following PRs will:
- secure the request apps collection (only allow org admins to create requests) #697
- secure the admin pages #700
- Implement a nicer design
- Refactor the data types so they can be shared between the back and front
- show "request to app" only to admins (permissions related)
- guard & secure each app if and only if the org has access to the app


apps/main/main/src/app/app-grid/app-grid.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  private orgIdSubscription: Subscription;
+
+  constructor(
+    private organizationQuery: OrganizationQuery,
+    private organizationService: OrganizationService
+  ) {}
+
+  ngOnInit() {
+    this.apps$ = this.organizationQuery.appsDetails$;
+    this.orgIdSubscription = this.organizationQuery.orgId$.subscribe(orgId => (this.orgId = orgId));
```
GrandSchtroumpf: If you're using this value only in the ts you can use a snapshot (so you don't need a subscription). See comment below

apps/main/main/src/app/app-grid/app-grid.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  ngOnDestroy(): void {
+    this.orgIdSubscription.unsubscribe();
+  }
+
+  public requestAccess(appId: string) {
+    return this.organizationService.requestAccessToApp(this.orgId, appId);
```
GrandSchtroumpf: You can do : 
```typescript
const orgId = this.query.getValue().org.id; // snapshot of the current value in the store
return this.organizationService.requestAccessToApp(orgId, appId);
```

libs/organization/src/lib/+state/organization.model.ts
```diff
@@ -1,3 +1,24 @@
+/** Gives information about an application */
+import { App } from '../permissions/+state';
+
+export interface AppDetails {
+  name: string;
+  logoSrc: string;
```
GrandSchtroumpf: just "logo" should be enough to understand I think.

libs/organization/src/lib/+state/organization.model.ts
```diff
@@ -1,3 +1,24 @@
...
+}
+
+export const enum AppStatus {
+  none = 'none', // no request nor accept.
+  requested = 'requested',
+  accepted = 'accepted'
+}
+
+/** An application details with the organization authorizations */
+export interface AppDetailsWithStatus extends AppDetails {
```
GrandSchtroumpf: Why do you need a different interface for that ?

libs/organization/src/lib/+state/organization.query.ts
```diff
@@ -1,18 +1,69 @@
...
+  get orgId$(): Observable<string> {
+    return this.select(state => state.org.id);
+  }
+
+  /**
+   * Returns an Observable that describe the list
+   * of application that are accessible to the current
+   * organization.
+   */
+  get appsDetails$(): Observable<AppDetailsWithStatus[]> {
```
GrandSchtroumpf: As discussed, because observables are lazy, I would recommand to use variables for them : 
```typescript
appsDetails$ = this.orgId$.pipe(...)
```

libs/organization/src/lib/+state/organization.query.ts
```diff
@@ -1,18 +1,69 @@
...
+    return this.orgId$.pipe(
+      map(orgId => this.db.collection('app-requests').doc(orgId)),
+      switchMap(docRef => docRef.valueChanges()),
+      map(appRequest => appRequest || {}), // default to an empty document if no requests happened done so fare
+      map(appRequest =>
```
GrandSchtroumpf: I think you can do `map((appRequest = {}) => ...)` to avoid to map twice

libs/organization/src/lib/+state/organization.model.ts
lsenta: `AppDetails` are fixed value that describe an application, AppDetailsWithStatus is the data enhanced with stuff from the database.

PR #702, opened 2019-07-29
Issue 573: pending orgs in the frontend

lsenta:


libs/organization/src/lib/guard/organization.guard.ts
```diff
@@ -24,6 +29,18 @@ export const orgQuery = (orgId: string): Query<Organization> => ({
   // })
 });
 
+/** Returns an observable over organization, to be reused when you need orgs without guards */
+export function orgObservable(auth: AuthQuery, db: FireQuery) {
```
GrandSchtroumpf: Call it `organization$` for consistency.

GrandSchtroumpf: This should be in the service. This is the file linked to the backend

libs/organization/src/lib/pages/organization-feedback/organization-feedback.component.ts
```diff
@@ -1,14 +1,32 @@
...
-  constructor(private router: Router) {}
+export class OrganizationFeedbackComponent implements OnInit {
+  /** The organization status should be `accepted' before we can move on to the app */
+  public canMoveOn: Observable<boolean>;
+
+  constructor(private router: Router, private auth: AuthQuery, private db: FireQuery) {
+  }
+
+  ngOnInit(): void {
+    this.canMoveOn = orgObservable(this.auth, this.db).pipe(
```
GrandSchtroumpf: The organization should be in the state. It's hard to know where the data come from, or are, if we don't keep everything in the state. If you don't want to store it inside `state.org`, you can create another field.
Here you should have something like : 
```typescript
this.organizationService.selectUserOrganization(this.auth.getValue().user);
this.canMoveOn = this.organizationQuery.isAccepted$; // create this query in the organization query.
```

libs/organization/src/lib/guard/organization.guard.ts
lsenta: That's what I expected but the query is defined in the guard, thanks for the confirmation, I'll move both to the service.

GrandSchtroumpf: We put the query object inside the guard because it was used only there at that time I guess. Your function should be a method of service. Maybe : 
```typescript
sync(): Observable<Organization> {
  return this.authQuery.user$.pipe(....);
}
```

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -18,8 +17,28 @@ import { OrganizationStore } from './organization.store';
...
+  //     return action.activeMembers.map(id => ({
+  //       path: `users/${id}`
+  //     }))
+  //   }
+  // })
+});
+
 @Injectable({ providedIn: 'root' })
 export class OrganizationService {
+  private orgObservable$: Observable<Organization>;
```
GrandSchtroumpf: Call it `organization$`. $ already means "observable"

PR #663, opened 2019-07-19
Add main app

lsenta:

PR #709, opened 2019-07-30
Issue 569: create main app LIGHT VERSION

lsenta:
This PR adds the minimal we need to start building the main version.
It creates a new app called `main` whose only job is to:
- [x] replace the old index.html with a page that has the angular machinery
- [x] provide the basic of auth & org creation, pretty much like any other app.

Later we'll turn this into a modular app that can load sub applications easily,
see: https://github.com/blockframes/blockframes/pull/663


apps/main/main/src/app/home/home.component.html
```diff
@@ -0,0 +1,21 @@
+<section>
+  <h2 fxLayout="row" fxLayoutAlign="start center">
+    <span>Welcome to Blockframes</span>
+    <mat-divider></mat-divider>
+  </h2>
+  <article>
+    <h3>My DApps</h3>
+    <article fxLayout="row" fxLayoutAlign="space-between center">
+      <mat-card *ngFor="let dapp of dapps">
+        <a routerLink="{{dapp.routerLink}}">
```
GrandSchtroumpf: Do `[routerLink]="[dapp.routerLink]"`

apps/main/main/src/app/home/home.component.html
```diff
@@ -0,0 +1,21 @@
...
+          <mat-card-content><img src="{{dapp.logoSrc}}" alt="DApp logo"/></mat-card-content>
```
GrandSchtroumpf: Do `[src]="dapp.logoSrc"`

apps/main/main/src/app/app-routing-module.ts
```diff
@@ -0,0 +1,80 @@
...
+            pathMatch: 'full'
+          },
+          {
+            path: 'organization',
+            loadChildren: () => import('@blockframes/organization').then(m => m.OrganizationModule)
+          },
+          {
+            path: 'account',
+            loadChildren: () => import('@blockframes/account').then(m => m.AccountModule)
+          },
```
GrandSchtroumpf: I think you're missing the routes for the other apps : 
```
{ path: 'delivery', loadChildren: () => import('@blockframes/delivery').then(m => m.DeliverySubAppModule') }
```
Be careful to use the DeliverySubAppModule and not the normal module or it won't work.

lsenta: (logging the slack discussion: we use href to link to other app, main is an independent app for now, delivery is not part of main)

apps/main/main/src/app/home/home.component.html
```diff
@@ -7,8 +7,8 @@ <h2 fxLayout="row" fxLayoutAlign="start center">
     <h3>My DApps</h3>
     <article fxLayout="row" fxLayoutAlign="space-between center">
       <mat-card *ngFor="let dapp of dapps">
-        <a href="{{dapp.href}}">
-          <mat-card-content><img src="{{dapp.logoSrc}}" alt="DApp logo"/></mat-card-content>
+        <a [href]="[dapp.href]">
```
GrandSchtroumpf: The "[]" were only when using routerlink. Here you need : `<a [href]="dapp.href"` and `[src]="dapp.logoSrc"`

PR #718, opened 2019-08-01
worked on e2e for laurent

Maxstgt:
the example is found in the auth lib


libs/auth/cypress.json
```diff
@@ -0,0 +1,11 @@
...
+  "fileServerFolder": "../../../dist/out-tsc/apps/delivery/delivery-e2e",
+  "fixturesFolder": "../../../dist/out-tsc/apps/delivery/delivery-e2e/src/fixtures",
+  "integrationFolder": "../../../dist/out-tsc/apps/delivery/delivery-e2e/src/integration",
+  "pluginsFile": "../../../dist/out-tsc/apps/delivery/delivery-e2e/src/plugins/index.js",
+  "supportFile": false,
+  "video": true,
+  "videosFolder": "../../../dist/out-tsc/apps/delivery/delivery-e2e/videos",
+  "screenshotsFolder": "../../../dist/out-tsc/apps/delivery/delivery-e2e/screenshots",
+  "chromeWebSecurity": false
+}
```
lsenta: That's configuration for the runner, that should not be here (what happens when you import 2 libs?).

PR #723, opened 2019-08-01
Issue 590: new delivery flow, phase 1

lsenta:
- [x] /delivery/add/1-find-movie
- [x] /delivery/add/:movieId/2-choose-starter
- [x] /delivery/add/:movieId/3-pick-template
- [x] /delivery/add/:movieId/:templateId/4-settings

This implements 1/4th of the "create a delivery" flow (pick a template). Next the other 3/4th, then design and the actual creation + updated app.

Introduce 3 new components to prepare for design & reusability:
- [x] actions-list: a list of actions
- [x] actions-picker: pick an action from a list
- [x] actions-picker-list: pick multiple actions from a list

The last two are simple wrappers around materials components for now and will be redesigned to follow our guidelines.

### DIFF
--- a/libs/material/src/lib/delivery/pages/delivery-add-choose-starter/delivery-add-choose-starter.component.ts
+++ b/libs/material/src/lib/delivery/pages/delivery-add-choose-starter/delivery-add-choose-starter.component.ts
@@ -1,5 +1,6 @@
 import { ChangeDetectionStrategy, Component } from '@angular/core';
 import { Router } from '@angular/router';
+import { ActionItem } from '@blockframes/ui';
 
 /**
  * Page for the flow: "create a delivery"
@@ -12,5 +13,17 @@ import { Router } from '@angular/router';
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class DeliveryAddChooseStarterComponent {
+  items: ActionItem[] = [
+    { icon: 'order', title: 'Choose a template', description: 'Lorem ipsum', routerLink: 'a' },
+    { icon: 'order', title: 'Import material list', description: 'Lorem ipsum', routerLink: 'a' },
+    {
+      icon: 'order',
+      title: 'Import specific delivery list',
+      description: 'Lorem ipsum',
+      routerLink: 'a'
+    },
+    { icon: 'order', title: 'Blank', description: 'Lorem ipsum', routerLink: 'a' }
+  ];
+
   constructor(private router: Router) {}
 }
--- a/libs/material/src/lib/delivery/pages/delivery-add-choose-starter/delivery-add-choose-starter.component.ts
+++ b/libs/material/src/lib/delivery/pages/delivery-add-choose-starter/delivery-add-choose-starter.component.ts
@@ -1,0 +1,16 @@
+import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { Router } from '@angular/router';
+
+/**
+ * Page for the flow: "create a delivery"
+ * second step, choose a starter template.
+ */
+@Component({
+  selector: 'delivery-add-choose-starter',
+  templateUrl: './delivery-add-choose-starter.component.html',
+  styleUrls: ['./delivery-add-choose-starter.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class DeliveryAddChooseStarterComponent {
+  constructor(private router: Router) {}
+}
--- a/libs/material/src/lib/delivery/pages/delivery-add-choose-starter/delivery-add-choose-starter.component.ts
+++ b/libs/material/src/lib/delivery/pages/delivery-add-choose-starter/delivery-add-choose-starter.component.ts
@@ -1,5 +1,4 @@
 import { ChangeDetectionStrategy, Component } from '@angular/core';
-import { Router } from '@angular/router';
 import { ActionItem } from '@blockframes/ui';
 
 /**
@@ -14,16 +13,19 @@ import { ActionItem } from '@blockframes/ui';
 })
 export class DeliveryAddChooseStarterComponent {
   items: ActionItem[] = [
-    { icon: 'order', title: 'Choose a template', description: 'Lorem ipsum', routerLink: 'a' },
-    { icon: 'order', title: 'Import material list', description: 'Lorem ipsum', routerLink: 'a' },
+    {
+      icon: 'order',
+      title: 'Choose a template',
+      description: 'Lorem ipsum',
+      routerLink: '../3-pick-template'
+    },
+    { icon: 'order', title: 'Import material list', description: 'Lorem ipsum', routerLink: '#' },
     {
       icon: 'order',
       title: 'Import specific delivery list',
       description: 'Lorem ipsum',
-      routerLink: 'a'
+      routerLink: 'import-delivery-list'
     },
-    { icon: 'order', title: 'Blank', description: 'Lorem ipsum', routerLink: 'a' }
+    { icon: 'order', title: 'Blank', description: 'Lorem ipsum', routerLink: '#' }
   ];
-
-  constructor(private router: Router) {}
 }
--- a/libs/material/src/lib/delivery/pages/delivery-add-settings/delivery-add-settings.component.html
+++ b/libs/material/src/lib/delivery/pages/delivery-add-settings/delivery-add-settings.component.html
@@ -1,0 +1,11 @@
+<section fxLayout="row" fxLayoutAlign="center center">
+  <article fxLayout="column" fxLayoutAlign="center center">
+    <h2>Settings</h2>
+    <p>Lorem ipsum dolor sit amet</p>
+
+    <actions-picker-list [items]="items" (picked)="this.picked($event)"></actions-picker-list>
+    <button (click)="onCompleteFlow()" mat-button color="primary" [disabled]="this.options.length === 0">
+      Continue
+    </button>
+  </article>
+</section>
--- a/libs/material/src/lib/delivery/pages/delivery-add-settings/delivery-add-settings.component.html
+++ b/libs/material/src/lib/delivery/pages/delivery-add-settings/delivery-add-settings.component.html
@@ -3,8 +3,8 @@
     <h2>Settings</h2>
     <p>Lorem ipsum dolor sit amet</p>
 
-    <actions-picker-list [items]="items" (picked)="this.picked($event)"></actions-picker-list>
-    <button (click)="onCompleteFlow()" mat-button color="primary" [disabled]="this.options.length === 0">
+    <actions-picker-list [items]="items" (picked)="picked($event)"></actions-picker-list>
+    <button (click)="onCompleteFlow()" mat-button color="primary" [disabled]="options.length === 0">
       Continue
     </button>
   </article>
--- a/libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.html
+++ b/libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.html
@@ -8,7 +8,15 @@
       <actions-picker [items]="items$ | async" (picked)="selectTemplate($event)"></actions-picker>
     </ng-template>
 
-    <a (click)="onContinue()" mat-button color="primary" [disabled]="!currentTemplate">Continue</a>
-    <a [routerLink]="['../2-choose-starter']"> Cancel </a>
+    <ng-container *ngIf="currentTemplate; else: disabled">
+      <a mat-button color="primary" [routerLink]="[continueURL]" [disabled]="!currentTemplate">
+        Continue
+      </a>
+    </ng-container>
+    <ng-template #disabled>
+      <span>Select a template</span>
+    </ng-template>
+
+    <a [routerLink]="['../2-choose-starter']">Cancel</a>
   </article>
 </section>
--- a/libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.html
+++ b/libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.html
@@ -3,9 +3,9 @@
     <h2>Choose a template</h2>
     <p>Lorem ipsum dolor sit amet</p>
 
-    <ng-container *ngIf="(loading$ | async); else: loaded">Loading</ng-container>
+    <ng-container *ngIf="(isLoading$ | async); else: loaded">Loading</ng-container>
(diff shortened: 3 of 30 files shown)


libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  constructor(
+    private templateQuery: TemplateQuery,
+    private router: Router,
+    private movieQuery: MovieQuery
+  ) {}
+
+  ngOnInit(): void {
+    this.loading$ = this.templateQuery.selectLoading();
+    this.items$ = this.templateQuery.selectAll().pipe(
```
Hugo-Hontans: It's better to create a getter in  templateQuery for do that

libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts
```diff
@@ -0,0 +1,51 @@
...
+      map(templates =>
+        templates.map((template: Template) => ({
+          title: template.name,
+          payload: template
+        }))
+      )
+    );
+  }
+
+  public picked(template: Template) {
```
Hugo-Hontans: Prefer a name representing the action like selectTemplate()

libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts
```diff
@@ -0,0 +1,51 @@
...
+ * third step, choose a template.
+ */
+@Component({
+  selector: 'delivery-add-template-picker',
+  templateUrl: './delivery-add-template-picker.component.html',
+  styleUrls: ['./delivery-add-template-picker.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class DeliveryAddTemplatePickerComponent implements OnInit {
+  public loading$: Observable<boolean>;
```
Hugo-Hontans: Rename with isLoading$

libs/movie/src/lib/movie/components/movie-picker/movie-picker.component.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@Component({
+  selector: 'movie-picker',
+  templateUrl: './movie-picker.component.html',
+  styleUrls: ['./movie-picker.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class MoviePickerComponent implements OnInit {
+  public movies$: Observable<Movie[]>;
+  public loading$: Observable<boolean>;
```
Hugo-Hontans: Same isLoading$

libs/ui/src/lib/actions-list/actions-list.component.html
```diff
@@ -0,0 +1,13 @@
+<mat-nav-list>
+  <a *ngFor="let item of items; last as last" [routerLink]="[item.routerLink]">
```
Hugo-Hontans: You don't need [ ] here

libs/ui/src/lib/actions-picker-list/actions-picker-list.component.ts
```diff
@@ -0,0 +1,24 @@
...
+  selector: 'actions-picker-list',
+  templateUrl: './actions-picker-list.component.html',
+  styleUrls: ['./actions-picker-list.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ActionsPickerListComponent<T> {
+  @Input() public items: ActionPickerListItem<T>[];
+  @Output() public picked = new EventEmitter<T[]>();
+
+  public forwardToParent(event: MatSelectionListChange) {
```
Hugo-Hontans: Maybe just name it selectItem()

libs/ui/src/lib/actions-picker-list/actions-picker-list.component.html
```diff
@@ -0,0 +1,5 @@
+<mat-selection-list (selectionChange)="forwardToParent($event)">
+  <mat-list-option *ngFor="let item of items" [value]="item">
```
Hugo-Hontans: Set the value with :
```[value]="item.payload"``` if you just need payload

File: libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts (hunk 1 above)
Comment (lsenta) [reply libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts]: This transformation is really specific to the `<action-picker>` we are using in the template

File: libs/ui/src/lib/actions-picker-list/actions-picker-list.component.html (hunk 7 above)
Comment (lsenta) [reply libs/ui/src/lib/actions-picker-list/actions-picker-list.component.html]: Nice catch, thanks!

File: libs/material/src/lib/delivery/pages/delivery-add-choose-starter/delivery-add-choose-starter.component.ts [hunk 8]
```diff
@@ -0,0 +1,31 @@
+import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { ActionItem } from '@blockframes/ui';
+
+/**
+ * Page for the flow: "create a delivery"
+ * second step, choose a starter template.
+ */
+@Component({
+  selector: 'delivery-add-choose-starter',
```
GrandSchtroumpf: Not sure to understand why "add". I think you can stick to `delivery-choose-starter`. or `delivery-create-starter`...

libs/material/src/lib/delivery/pages/delivery-add-settings/delivery-add-settings.component.html
```diff
@@ -0,0 +1,11 @@
+<section fxLayout="row" fxLayoutAlign="center center">
+  <article fxLayout="column" fxLayoutAlign="center center">
+    <h2>Settings</h2>
+    <p>Lorem ipsum dolor sit amet</p>
+
+    <actions-picker-list [items]="items" (picked)="picked($event)"></actions-picker-list>
+    <button (click)="onCompleteFlow()" mat-button color="primary" [disabled]="options.length === 0">
```
GrandSchtroumpf: Use `a` tag, as it's going to change to another page. At least a blind person will now it's a link.

libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts
```diff
@@ -0,0 +1,51 @@
...
+    private templateQuery: TemplateQuery,
+    private router: Router,
+    private movieQuery: MovieQuery
+  ) {}
+
+  ngOnInit(): void {
+    this.isLoading$ = this.templateQuery.selectLoading();
+    this.items$ = this.templateQuery.selectAll().pipe(
+      map(templates =>
+        templates.map((template: Template) => ({
```
GrandSchtroumpf: I would have create a pure function at the top of the file `createAction(template: Template): ActionPickerItem<Template>` that would have deal with type and all. Like that you can do : 
```typescript
this.templateQuery.selectAll().pipe(
  map(templates=> createAction(templates))
);
```
I think it's more readable and avoid formating issues from prettier.

libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts
```diff
@@ -0,0 +1,51 @@
...
+          title: template.name,
+          payload: template
+        }))
+      )
+    );
+  }
+
+  public selectTemplate(template: Template) {
+    this.currentTemplate = template;
```
GrandSchtroumpf: If this is related only to this component you can keep it like that. But just in case, have you consider `this.templateStore.setActive(template.id)` ?

libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.html
```diff
@@ -0,0 +1,14 @@
...
+  <article fxLayout="column" fxLayoutAlign="center center">
+    <h2>Choose a template</h2>
+    <p>Lorem ipsum dolor sit amet</p>
+
+    <ng-container *ngIf="(isLoading$ | async); else: loaded">Loading</ng-container>
+    <ng-template #loaded>
+      <actions-picker [items]="items$ | async" (picked)="selectTemplate($event)"></actions-picker>
+    </ng-template>
+
+    <button (click)="onTemplatePicked()" mat-button color="primary" [disabled]="!currentTemplate">Continue</button>
```
GrandSchtroumpf: Use `a` also as you're ultimately moving to a new page.

libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public onTemplatePicked() {
```
GrandSchtroumpf: Can you change the name of the method ? As the output of `action-picker` is "picked", it was a little confusing for me.

libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts
```diff
@@ -0,0 +1,51 @@
...
+    const templateId = this.currentTemplate.id;
+    const movieId = this.movieQuery.getActiveId();
+    return this.router.navigate([`/layout/o/delivery/add/${movieId}/${templateId}/4-settings`]);
```
GrandSchtroumpf: I think I should have created a getter with the route to take advantage of the relative route of the directive `routerLink` + using tag "a" : 
```typescript
get movieId() {
  if (!this.currentTemplate) return;
  const templateId = this.currentTemplate.id;
  const movieId = this.movieQuery.getActiveId();
  return `${movieId}/${templateId}/4-settings`
}
```
In the html : 
```html
<ng-container *ngIf="currentTemplate; else disabled">
  <a ... [routerLink]="[movieId, currentTemplate.id, '4-settings']">Continue</a>
  <a ...>Cancel</a>
</ng-container>
<ng-template #disabled>
  <span>Select a template</span> // Or a disabled button
  <a ...>Cancel</a>
</ng-template>
```

libs/ui/src/lib/actions-list/actions-list.component.scss
```diff
@@ -0,0 +1,25 @@
...
+  margin-bottom: 4px;
+}
+
+.left-icon {
+  margin-right: 16px;
+}
+
+.right-icon {
+  margin-left: 16px;
+  color: var(--dynamicIcon);
```
GrandSchtroumpf: dyanmic Icon will be removed. It should be used for clickable icon. If you want a clickable icon you should put it inside a mat-icon-button tag

libs/ui/src/lib/actions-picker-list/actions-picker-list.component.html
```diff
@@ -0,0 +1,5 @@
+<mat-selection-list (selectionChange)="selectItem($event)">
```
GrandSchtroumpf: I don't really the point a creating a 5 lines component...

libs/ui/src/lib/actions-picker/actions-picker.component.html
```diff
@@ -0,0 +1,5 @@
+<mat-radio-group (change)="picked.emit($event.value.payload)">
```
GrandSchtroumpf: Same as above.

libs/ui/src/lib/actions-picker/actions-picker.component.scss
```diff
@@ -0,0 +1,9 @@
+mat-radio-group {
+  display: flex;
+  flex-direction: column;
```
GrandSchtroumpf: why not putting the in the HTML

libs/material/src/lib/delivery/pages/delivery-add-choose-starter/delivery-add-choose-starter.component.ts
lsenta: Because it's the flow to add a new delivery, I prefixed all the pages with `delivery-add`, then each page

libs/material/src/lib/delivery/pages/delivery-add-settings/delivery-add-settings.component.html
lsenta: Are you sure about that? It's not only navigation, this will contain the action that creates the delivery object then redirect you to this object, it's similar to submitting a form to create a document rather than navigating to a URL.

GrandSchtroumpf: for this one, I'm not sure because there were no code in the function yet.
But for the next one (comment below) I'm sure.

libs/material/src/lib/delivery/pages/delivery-add-template-picker/delivery-add-template-picker.component.ts
lsenta: looks much better with a getter and a link, thanks for the tip!
I'll keep the full continue URL in the component though, since this is highly dependant to other parts of the code (routing, etc).

libs/ui/src/lib/actions-picker-list/actions-picker-list.component.html
lsenta: This is what it looks like now (material):
Screenshot 2019-08-07 at 15 29 31

this is what it will look like soon:
Screenshot 2019-08-07 at 15 29 43

It's used in at least 2 locations in the code and when we add all the additional html + sass required for styling, this'll be a large reusable component. The other option is to have 2 piece of duplicated code that'll become 2 large piece of duplicated code.

libs/ui/src/lib/actions-picker/actions-picker.component.html
lsenta: Same answer!

libs/ui/src/lib/actions-picker/actions-picker.component.scss
lsenta: I extracted duplicated code and turned it into a component, I don't know the initial reasons to have this in sass rather than HTML, do you want me to put it in the HTML?

GrandSchtroumpf: Yes, put it in the HTML if there is no specific reason to live it in the css.

PR #731, opened 2019-08-06
update of imported movies

phpgeekfr:


libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
```diff
@@ -53,7 +53,7 @@ export class TableExtractedMoviesComponent implements OnInit {
   }
 
   createMovie(movie: Movie): Promise<boolean> {
-    return this.addMovie(movie)
+    return this.doCreateMovie(movie)
```
GrandSchtroumpf: This is not readable to have a `createMovie` and `doCreateMovie` in the same class. Why not only one method ? What's the difference ?

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
```diff
@@ -74,20 +74,38 @@ export class TableExtractedMoviesComponent implements OnInit {
     });
   }
 
-  private addMovie(movie: Movie): Promise<void> {
+  private doCreateMovie(movie: Movie): Promise<void> {
```
GrandSchtroumpf: use async / await : 
```typescript
const { id } = await this.movieService.addMovie(movie.main.title.original);
const updated = { ...movie, id };
return this.movieService.update(id, updated);
```
The process is super weird though. I expect `service.addMovie` to create the whole movie in firestore. You shouldn't have to update it later. Please take a look at that.

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
```diff
@@ -74,20 +74,38 @@ export class TableExtractedMoviesComponent implements OnInit {
...
         movie.id = id;
         return this.movieService.update(id, JSON.parse(JSON.stringify(movie))); //@todo remove #483
       });
   }
 
-  updateSelectedMovies() { // : Promise<boolean>
-    // @todo
+  updateSelectedMovies() : Promise<boolean> {    
+    const moviesToUpdate = [];
+    this.selection.selected.forEach((data: MovieImportState) => {
```
GrandSchtroumpf: Use filter, create a pure function `hasImportErrors` outside of the class, give a better name than "data", give a better name that "doUpdateMovie"
```typescript
try {
  const updates = this.selection.selected
    .filter(data => data.movie.id && !hasImportErrors(data))
    .map(data => prepareMovieUpdate(data.movie));
  await Promise.all(updates);
  this.snackBar.open(`${moviesToUpdate.length} movies updated!`, 'close', { duration: 3000 });
} catch(err) {
  this.snackBar.open(`could not update movies`, 'close', { duration: 3000 });
}
```

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
```diff
@@ -74,20 +74,38 @@ export class TableExtractedMoviesComponent implements OnInit {
...
+      }
+    })
+
+    return Promise.all(moviesToUpdate).then(() => {
+      this.snackBar.open(`${moviesToUpdate.length} movies updated!`, 'close', { duration: 3000 });
+      return true;
+    });
   }
 
   updateMovie(movie: Movie) {
```
GrandSchtroumpf: I think you don't want to display a snackBar for 3sec each time a movie is updated. This is super invasif. We should have something like "unsaved"/"saved" icon or text instead.

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -140,26 +132,30 @@ export class ViewExtractedElementsComponent {
...
-          importErrors.errors.push({
-            type: 'error',
-            field: 'salesInfo.scoring',
-            name: "Scoring",
-            reason: 'Required field could not be parsed',
-            hint: 'Edit corresponding sheet field.'
-          } as SpreadsheetImportError);
+        if(spreadSheetRow[SpreadSheetMovie.scoring]) {
+          const scoring = getCodeIfExists('SCORING', spreadSheetRow[SpreadSheetMovie.scoring]);
+          if (scoring !== false) {
```
GrandSchtroumpf: That's weird. why scoring would be false... ?

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
phpgeekfr: Because I have two methods one for creating one movie and one for creating several movies. The two methods use `doCreateMovie` to prevent code duplication.

Same thing for update. Do you have a better idea ? code duplication or better naming ?

GrandSchtroumpf: I think you've code duplication here.

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
phpgeekfr: No, it display only one message saying how many films have been updated.

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
phpgeekfr: Because cell content in spreadsheet could not be matching with expected value. getCodeIfExists returns false if nothing found.

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
GrandSchtroumpf: this funtion is per movie right ? (only one movie as parameter). So each time it's call, you display a snackBar. I think you have a duplication of code 'updateMovie', 'doUpdateMovie', 'updateSelectedMovies'.

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
GrandSchtroumpf: `getCodeIfExists` should return null or undefined if nothing exist. Also you should check the type here : `if (typeof scroring === 'number')` to be sure.

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
phpgeekfr: ok

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
phpgeekfr: If multi update or create => one snack bar saying how many update/creates
if single update or create => one snack bar

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
phpgeekfr: OK getCodeIfExists now returns null.
scoring is not a number, its a string

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
```diff
@@ -52,48 +56,58 @@ export class TableExtractedMoviesComponent implements OnInit {
...
+    return true;
   }
 
-  updateMovie(movie: Movie) {
-    console.log(`todo ${movie.id}`);
+  async updateSelectedMovies(): Promise<boolean> {
+    try {
+      const updates = this.selection.selected
+        .filter((importState: MovieImportState) => importState.movie.id && !hasImportErrors(importState))
+        .map((importState: MovieImportState) => this.movieService.update(
```
GrandSchtroumpf: You don't need the type here and in the filter. It should be given bay the selected.
Also for redability I would have written that : 
```typescript
.map(importState => {
  const movieToSave = this.prepareMovieSave(importState.movie)
  return this.movieService.update(importState.movie.id, movieToSave);
})
```

libs/movie/src/lib/movie/+state/movie.query.ts
```diff
@@ -18,9 +18,9 @@ export class MovieQuery extends QueryEntity<MovieState, Movie> {
   /**
    * @param internalRef 
    */
-  public movieExists(internalRef: string) : string {
+  public existingMovie(internalRef: string) : Movie {
     const movie = this.getAll().find(entity =>  entity.main.internalRef === internalRef );
```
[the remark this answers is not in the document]
phpgeekfr: lol, sorry for that..

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -96,40 +109,19 @@ export class ViewExtractedElementsComponent {
 
     sheetTab.rows.forEach(async spreadSheetRow => {
       if (spreadSheetRow[SpreadSheetMovie.originalTitle]) {
+        const existingMovie = this.movieQuery.existingMovie(spreadSheetRow[SpreadSheetMovie.internalRef]);
         const movie = {
```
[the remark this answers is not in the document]
phpgeekfr: I don't understand, this function is not about creating movie but fetching an existing movie into state. Can you explain more ?

PR #735, opened 2019-08-08
created the two catalog apps

Maxstgt:
I worked on the angular json. So I guess the error was somewhere there. Not quite sure Nrwl is working properly...? I also renamed the home component from the catalog marketplace cause we got a mostly same comoponent on the dashboard app


nx.json
```diff
@@ -73,6 +73,18 @@
     },
     "backend-ops": {
       "tags": []
+    },
```
GrandSchtroumpf: You'll need to remove "catalog" and "catalog-e2e"

PR #747, opened 2019-08-13
checks if a sale already exists

phpgeekfr:


libs/movie/src/lib/movie/+state/movie.query.ts
```diff
@@ -22,4 +22,30 @@ export class MovieQuery extends QueryEntity<MovieState, Movie> {
     return this.getAll().find(entity =>  entity.main.internalRef === internalRef );
   }
 
+  /**
+   * Checks if a sale is already existing for a given movie and returns it
+   * @param internalRef
```
GrandSchtroumpf: Explain what it is this parameter.

phpgeekfr: added comment

PR #754, opened 2019-08-14
Issue 751 752

phpgeekfr:


libs/movie/src/lib/movie/display/main/main.component.html
```diff
@@ -39,11 +39,11 @@
         </span>
       </mat-card-subtitle>
     </ng-container>
-    <ng-container *ngIf="data.originCountry">
-      <mat-card-title>Origin Country</mat-card-title>
+    <ng-container *ngIf="data.originCountries.length > 0">
```
GrandSchtroumpf: You can do `*ngIf="data.originCountries.length"` as 0 is evaluated to "false" (which can be annoying some time)

PR #755, opened 2019-08-14
backend(ops): add auto remove users for tests & setup

lsenta:


lsenta:
replaced by https://github.com/blockframes/blockframes/pull/867

PR #765, opened 2019-08-20
Issue#758 delivery dates

Hugo-Hontans:
Update delivery dates


libs/material/src/lib/delivery/+state/delivery.model.ts
```diff
@@ -17,6 +17,8 @@ interface AbstractDelivery {
 
 export interface Delivery extends AbstractDelivery {
   dueDate?: Date;
+  acceptationPeriod?: Number,
```
GrandSchtroumpf: "Number" is not a type, it's a class. "number" is the type.
Maybe add some comment on top of the acceptationPeriod and reWorkingPeriod to explain what it is.

Hugo-Hontans: Obviously !
I added comments

PR #767, opened 2019-08-21
Integrate calendar & datepicker

VicoDraye:


libs/ui/src/lib/datepicker/datepicker.component.html
```diff
@@ -0,0 +1,25 @@
+  <mat-form-field appearance="outline" class="">
```
GrandSchtroumpf: You have an empty class

libs/ui/src/lib/datepicker/datepicker.component.ts
```diff
@@ -0,0 +1,83 @@
...
+import {MatSnackBar} from '@angular/material/snack-bar';
+import { CalendarRange, isBetween } from '@blockframes/utils';
+
+@Component({
+  selector: 'datepicker-range',
+  templateUrl: './datepicker.component.html',
+  styleUrls: ['./datepicker.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DatepickerComponent {
```
GrandSchtroumpf: Let's call it DatepickerRangeComponent.

libs/ui/src/lib/datepicker/datepicker.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  
+  @Input() selectedRange: CalendarRange;
+  @Input() disabledDates: CalendarRange[];
+
+  @Output() wantedRange = new EventEmitter<CalendarRange>();
+  
+  @ViewChild('picker', {static: false}) picker: MatDatepicker<Date>;
+
+  datepickerRange: CalendarRange = {
```
GrandSchtroumpf: There is a DateRange type in the application (in movie.model.ts I think). You should use it.

libs/ui/src/lib/datepicker/datepicker.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DatepickerComponent {
+  
+  @Input() selectedRange: CalendarRange;
+  @Input() disabledDates: CalendarRange[];
+
+  @Output() wantedRange = new EventEmitter<CalendarRange>();
+  
+  @ViewChild('picker', {static: false}) picker: MatDatepicker<Date>;
```
GrandSchtroumpf: Does it has a good reason to be false ?

libs/ui/src/lib/datepicker/datepicker.component.ts
GrandSchtroumpf: Also, just call this property "range" we already know you're in a datepicker ;)

libs/ui/src/lib/datepicker/datepicker.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  initDatePicker(){
+    this.picker.open();
+  }    
+
+  disableDates = (date: Date): boolean => {
+    return !this.disabledDates.some(range => isBetween(date, range.begin, range.end));
+  }
+
+  datepickerRangeChange(range: CalendarRange) {
+    if(range){
```
GrandSchtroumpf: Do : 
```typescript
if (range) {
  const { end, begin } = range;
  this.range = (end < begin) ? { begin: end, end; begin } : range;
  this.validateDatepickerInput(this.datepickerRange);
}

```

libs/ui/src/lib/datepicker/datepicker.component.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+    this.isValidRange = !this.disabledDates.some(disabledDate => isInvalid(disabledDate));
+    if (!this.isValidRange){
+      this.openSnackBar();
+    }
+    return this.isValidRange;
+  }
+
+  openSnackBar(){
+    let snackBarRef = this._snackBar.open("Your selection contain a reserved date range.", "Change it", {
```
GrandSchtroumpf: use const

libs/ui/src/lib/datepicker/datepicker.module.ts
```diff
@@ -0,0 +1,32 @@
+import { NgModule } from '@angular/core';
+import {
+    //MatNativeDateModule,
```
GrandSchtroumpf: remove if you don't use

libs/ui/src/lib/month-calendar/month-calendar.component.html
```diff
@@ -0,0 +1,39 @@
+<div class="month-calendar">
```
GrandSchtroumpf: You don't need this tag. Remove it. If you need to access the root element of a component in CSS use `:host { ... }`

libs/ui/src/lib/month-calendar/month-calendar.component.html
```diff
@@ -0,0 +1,39 @@
...
+    [ngClass]="tile.state"
+    [colspan]="tile.cols"
+    [rowspan]="tile.rows"
+    (click)="getMonthRange(i)"
+    (mouseenter)="displayDetails[i] = true"
+    (mouseleave)="displayDetails[i] = false"
+  >
+  <ng-container *ngIf="displayDetails[i]; else displayMonth">
+    <ng-container select="details[tile.date]">
+        {{tile.state}}
```
GrandSchtroumpf: Use `<span>` around the text.

libs/ui/src/lib/month-calendar/month-calendar.component.ts
```diff
@@ -0,0 +1,124 @@
...
+  monthTiles: Tile[];
+  headerTile: Tile;  
+  displayedDates: Date[];
+  begin: Date;
+  end: Date;
+  isClicked = [];
+  displayDetails: boolean[];
+
+  ngOnInit(): void {
+    this.headerTile = {date: this.currentDate = new Date(), cols: 4, rows: 1, state:''};
```
GrandSchtroumpf: Don't do 2 operations on the same variable :
```
this.currentData = new Date();
this.headerTile = { date: this.currentDate, cols: 4, rows: 1, state: '' };
```

libs/ui/src/lib/month-calendar/month-calendar.component.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  getRange(month: number): CalendarRange{
+    this.begin = new Date(this.currentDate.getFullYear(), month, 1);
+    this.end = new Date(this.currentDate.getFullYear(), month + 1, 0);
+    return {begin: this.begin, end: this.end};
+  }
+
+  /** generate the calendar grid */
+  generateCalendar() {
+    this.months = this.fillDates();
```
GrandSchtroumpf: Don't do mutation in Javascript : 
```
this.months = this.fillDates();
this.monthTiles = this.months.map(month => ({ date: month, cols: 1, rows: 1, state: 'empty' });
```

libs/ui/src/lib/month-calendar/month-calendar.component.ts
```diff
@@ -0,0 +1,124 @@
+import { Component, OnInit, Output, EventEmitter, ChangeDetectionStrategy, Input} from '@angular/core';
+import { CalendarRange, isBetween } from '@blockframes/utils';
+
+interface Tile {
+  cols: number;
+  rows: number;
+  date: Date;
+  state: string;
```
GrandSchtroumpf: Do : `state?: 'empty', 'partial', 'full'`.
Best would be to use an enum

libs/ui/src/lib/datepicker/datepicker.component.ts
VicoDraye: So i need to import that ? 

```
export interface DateRange {
  from: Date;
  to: Date;
}
```

libs/ui/src/lib/datepicker/datepicker.component.ts
VicoDraye: PL said me to always add that

pldespaigne: Yeah, I think you can completely remove `,{static: false}`

libs/ui/src/lib/datepicker/datepicker.component.ts
pldespaigne: @GrandSchtroumpf maybe we should move this interface outside of movie.model.ts, because ui components should not import external thing no ?

libs/ui/src/lib/datepicker/datepicker.component.ts
GrandSchtroumpf: No you cannot since version 8 of Angular. But check if it's not "true" instead. Look at the doc:
https://angular.io/api/core/ViewChild

libs/ui/src/lib/datepicker/datepicker.component.ts
GrandSchtroumpf: I agree. It was introduced but Bruce very recently. You can do that.

libs/ui/src/lib/datepicker/datepicker.component.ts
pldespaigne: the doc says : true means query before change detection runs, false after, true is useful for getting the element even inside of a not displayed *ngIf, so for this purpose I don't think we need to set it to true, false should be ok

libs/ui/src/lib/datepicker/datepicker.component.ts
```diff
@@ -0,0 +1,78 @@
+import { Component, Input, ChangeDetectionStrategy, ViewChild, Output, EventEmitter } from '@angular/core';
+import { MatDatepicker} from '@angular/material';
+
+import { FormControl, Validators } from '@angular/forms';
+import { MatSnackBar } from '@angular/material/snack-bar';
+import { isBetween } from '@blockframes/utils';
+import { DateRange } from 'libs/movie/src/lib/movie/+state/movie.model';
```
pldespaigne: Go to `libs/movie/src/lib/movie/+state/movie.model.ts` and move the `DateRange` interface to `libs/utils/src/lib/helpers.ts`,
then update the import everywhere in the project until there is no more error (should not be many).

But be careful, the project contains several different app (delivery, catalog, etc...) so even if there is no error in your `ng serve` that doesn't mean there will be no error in the ci (because `ng serve` check only one app), to be sure you have fix every error you could run `npm run build:all` (but that can take some time)

libs/ui/src/lib/month-calendar/month-calendar.component.html
```diff
@@ -0,0 +1,39 @@
+<div>
```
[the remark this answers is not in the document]
pldespaigne: Angular does this automatically for you, so if you put a `<div>` around your component it will be translated in the html into something like
```html
<div>
  <div>
    the component html
  </div>
</div>
```

libs/ui/src/lib/month-calendar/month-calendar.module.ts
```diff
@@ -0,0 +1,17 @@
+import { NgModule } from '@angular/core';
+import { 
+    MatGridListModule, MatIconModule, MatButtonModule, 
+  } from "@angular/material";
+import { CalendarComponent } from './month-calendar.component';
+import { CommonModule } from '@angular/common';
+
+@NgModule({
+  declarations: [CalendarComponent],
+  imports: [MatGridListModule,
```
pldespaigne: put `MatGridListModule` to the next line

libs/ui/src/lib/datepicker/datepicker.component.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  range: DateRange = {
+    from: new Date(),
+    to: new Date(),
+  };
+  
+  isValidRange= false;
+  rangeForm = new FormControl('', [Validators.required]);
+
+  disableDates = (date: Date) => {
```
pldespaigne: https://github.com/blockframes/blockframes/pull/767#discussion_r316586847

PR #768, opened 2019-08-21
With Minimum Guarantee

lsenta:


lsenta:
Replaced by: https://github.com/blockframes/blockframes/pull/780

PR #769, opened 2019-08-21
Issue#759 delivery steps

Hugo-Hontans:
Editable steps of delivery


libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -25,7 +25,7 @@ interface AddDeliveryOptions {
 export function modifyTimestampToDate(delivery: DeliveryDB): Delivery {
   return {
     ...delivery,
-    dueDate: delivery.dueDate ? delivery.dueDate.toDate() : undefined,
+    dueDate: delivery.dueDate ? delivery.dueDate.toDate() : null,
```
GrandSchtroumpf: Why ?

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -188,48 +188,64 @@ export class DeliveryService {
...
-    const steps = [...delivery.steps];
-    const index = steps.indexOf(step);
-    steps.splice(index, 1);
-    this.db.doc<Delivery>(`deliveries/${delivery.id}`).update({ steps });
+    // Find steps that need to be removed
+    const deletedSteps = oldSteps.filter(
+      oldStep => !stepsWithId.some(newStep => newStep.id === oldStep.id)
+    );
+    // Remove stepId from the materials according to this array
+    this.removeMaterialsStepId(deletedSteps, batch);
```
GrandSchtroumpf: I reallly don't understand the process here. Why do you need to remove the steps ?

libs/material/src/lib/delivery/components/delivery-informations-steps-form/delivery-informations-steps-form.component.html
```diff
@@ -0,0 +1,40 @@
+<div>
```
GrandSchtroumpf: don't use a div like that. Maybe header, article, ...

libs/material/src/lib/delivery/components/delivery-informations-steps-form/delivery-informations-steps-form.component.html
```diff
@@ -0,0 +1,40 @@
...
+  <h4>Steps</h4>
+  <button mat-flat-button (click)="addStep()"><mat-icon>add</mat-icon>Add a step</button>
```
GrandSchtroumpf: put it on 2lines

libs/material/src/lib/delivery/components/delivery-informations-steps-form/delivery-informations-steps-form.component.html
```diff
@@ -0,0 +1,40 @@
...
+</div>
+<form [formGroup]="control">
+  <div formArrayName="deliverySteps">
+    <div *ngFor="let step of deliverySteps.controls; let i = index">
```
GrandSchtroumpf: use `ng-container`

libs/material/src/lib/delivery/+state/delivery.service.ts
Hugo-Hontans: This function is used by the delivery active guard for convert TimeStamp to type Date and I can't set my form with an undefined

libs/material/src/lib/delivery/pages/delivery-informations-editable/delivery-informations-editable.component.ts
```diff
@@ -26,7 +27,10 @@ export class DeliveryInformationsEditableComponent implements OnInit {
 
   ngOnInit() {
     this.delivery$ = this.query.selectActive().pipe(
-      tap(delivery => this.deliveryDates.patchValue(delivery)),
+      tap(delivery => {
+        this.deliveryDates.patchValue(delivery);
+        this.deliverySteps.patchValue(delivery.steps);
```
GrandSchtroumpf: Why do you need two forms ? It couldn't be in one centralized form ?

libs/material/src/lib/delivery/pages/delivery-informations-editable/delivery-informations-editable.component.ts
```diff
@@ -35,17 +39,24 @@ export class DeliveryInformationsEditableComponent implements OnInit {
...
 
-  public updateInformations() {
-    // TODO: update steps informations: issue#759
+  public async updateInformations() {
     // TODO: update guaranteed minimum payment deadline informations: issue#764
     try {
-      if (this.informationsFormGroup.invalid)
-        throw new Error('Delivery informations are not valid');
-      this.service.updateDates(this.deliveryDates.value);
+      if (this.informationsFormGroup.invalid) throw new Error('Delivery informations are not valid');
```
GrandSchtroumpf:
```typescript
if (this.informationsFormGroup.invalid) {
  throw new Error('Delivery informations are not valid');
}
```

libs/material/src/lib/delivery/+state/delivery.service.ts
Hugo-Hontans: I update an array of steps but steps that existed before may have been removed. In this case, I need to delete each stepId of each connected materials.

libs/material/src/lib/delivery/pages/delivery-informations-editable/delivery-informations-editable.component.ts
Hugo-Hontans: I rework it for have one centralized form looks better than the model and simplifies the code.

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -188,48 +188,64 @@ export class DeliveryService {
...
+  }
+
+  /** Remove stepId of materials of delivery for an array of steps */
+  private removeMaterialsStepId(steps: Step[], batch: firebase.firestore.WriteBatch) {
+    const deliveryId = this.query.getActiveId();
+    // We also set the concerned materials stepId to an empty string
+    steps.forEach(step => {
+      const materials = this.materialQuery.getAll().filter(material => material.stepId === step.id);
+      materials.forEach(material => {
+        const doc = this.db.doc(`deliveries/${deliveryId}/materials/${material.id}`);
```
lsenta: Advice: use `docRef = this.(...).ref;` when you're not using the doc, it makes the code much easier to understand since the reader KNOWS you're not accessing the data
(I'm going through this atm: refactoring the code and this would make my life MUCH easier)

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -188,48 +188,64 @@ export class DeliveryService {
...
-      return batch.update(doc.ref, { step: '' });
+    return batch.update(deliveryDocRef, { steps: stepsWithId });
+  }
+
+  /** Remove stepId of materials of delivery for an array of steps */
+  private removeMaterialsStepId(steps: Step[], batch: firebase.firestore.WriteBatch) {
+    const deliveryId = this.query.getActiveId();
+    // We also set the concerned materials stepId to an empty string
+    steps.forEach(step => {
+      const materials = this.materialQuery.getAll().filter(material => material.stepId === step.id);
```
lsenta: Note: you're relying on data that is in the db with `materialQuery.getAll`, this code should work from inside a transaction so that you:
1. start the transaction = freeze the db state (more or less)
2. read the data
3. update the data
4. close the transaction.

Here you:
1. read the material data
2. start the batch
3. update the data
4. end the batch
(simplified)

There are race conditions case where something can be written to materials between step 1 and 2.

libs/material/src/lib/delivery/pages/delivery-informations-editable/delivery-informations-editable.component.ts
```diff
@@ -26,26 +27,25 @@ export class DeliveryInformationsEditableComponent implements OnInit {
 
   ngOnInit() {
     this.delivery$ = this.query.selectActive().pipe(
-      tap(delivery => this.deliveryDates.patchValue(delivery)),
-      switchMap(delivery => this.deliveryDates.valueChanges.pipe(startWith(delivery)))
+      tap(delivery => {
+        this.informationsFormGroup.patchValue(delivery);
+      }),
+      switchMap(delivery => this.informationsFormGroup.valueChanges.pipe(startWith(delivery)))
```
lsenta: Could you write something like:
```ts
    this.delivery$ = this.query.selectActive().pipe(
      tap(delivery => {
        this.informationsFormGroup.patchValue(delivery);
      }),
      switchMap(delivery => this.informationsFormGroup.valueChanges),
      startWith(delivery)
  )
```
?

A few debugging sessions taught me that nested pipes makes the code rather hard to comprehend.

libs/material/src/lib/delivery/+state/delivery.service.ts
lsenta: Have you changed the type in Delivery?
It should throw a type error, if not, we're probably coding without a parachute (types).

libs/material/src/lib/delivery/+state/delivery.service.ts
lsenta: Risk is low, you may add a todo with a reference to #773

Hugo-Hontans: I added a comment with a TODO

libs/material/src/lib/delivery/pages/delivery-informations-editable/delivery-informations-editable.component.ts
Hugo-Hontans: No, we need to use startWith inside the pipe of valueChanges

libs/material/src/lib/delivery/+state/delivery.service.ts
Hugo-Hontans: Yes we will fix this in my next PR #764

PR #771, opened 2019-08-21
Issue#592 search

Maxstgt:


apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.component.ts
```diff
@@ -0,0 +1,152 @@
...
+@Component({
+  selector: 'catalog-marketplace-search',
+  templateUrl: './marketplace-search.component.html',
+  styleUrls: ['./marketplace-search.component.scss'],
+  changeDetection: ChangeDetectionStrategy.Default
+})
+export class MarketplaceSearchComponent implements OnInit {
+  public movieSearchResults$: Observable<Movie[]>;
+  public filterForm = new CatalogSearchForm();
+  // TODO#748: split up into compoennts
```
ioaNikas: Yes, it would be a good idea to have a component for each filter.

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.component.ts
```diff
@@ -0,0 +1,152 @@
...
+    const filterValue = territory.toLowerCase();
+    return this.movieTerritories.filter(movieTerritory => {
+      return movieTerritory.toLowerCase().includes(filterValue);
+    });
+  }
+
+  private _languageFilter(value: string): string[] {
+    const filterValue = value.toLowerCase();
+    return this.movieLanguages.filter(language => language.toLowerCase().includes(filterValue));
+  }
```
ioaNikas: Same as earlier, maybe this is more readable, what do you think ?
```
  private _languageFilter(filterValue: string): string[] {
    return this.movieLanguages.filter(language =>
      language.toLowerCase().includes(filterValue.toLowerCase())
    );
  }
```

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.module.ts
```diff
@@ -0,0 +1,63 @@
...
+    MatCheckboxModule,
+    MatAutocompleteModule,
+    MatListModule,
+    MatChipsModule,
+    FormsModule,
+    HttpClientModule,
+    MatNativeDateModule,
+    ReactiveFormsModule,
+    MatCardModule,
+    MatDatepickerModule,
```
ioaNikas: You should sort the imports (some here are not Material imports)

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.module.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    RouterModule.forChild([
+      {
+        path: '',
+        component: MarketplaceSearchComponent
```
ioaNikas: As we saw earlier, try to loading components on empty paths

libs/movie/src/lib/movie/components/display-card-item/display-card-item.component.ts
```diff
@@ -0,0 +1,25 @@
...
+          <mat-card-title>{{ movie.title.original }}</mat-card-title>
+          <mat-card-subtitle *ngFor="let director of movie.directors">
+          {{ director.firstName }} {{ director.lastName }}
+          </mat-card-subtitle>
+        </mat-card-title-group>
+
+        <span>Production Year</span><span>{{ movie.productionYear }}</span>
+      </mat-card-content>
+    </mat-card>
+  `,
```
ioaNikas: Maybe html here is a bit too large to use the template. Ask François to be sure, but I think it would be better to have html and scss files aswell.

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-home/marketplace-home.module.ts
```diff
@@ -21,9 +21,8 @@ import { CatalogMarketplaceHomeComponent } from './marketplace-home.component';
     FlexLayoutModule,
     MatCarouselModule,
     RouterModule.forChild([
-      { path: '', redirectTo: 'home', pathMatch: 'full' },
       {
-        path: 'home',
+        path: '',
         component: CatalogMarketplaceHomeComponent
```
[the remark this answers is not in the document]
GrandSchtroumpf: `Don't load components on empty path.` indead this is often a bad idea.
**But** when you work with a One component module like here, this is the best option. Your solution is good **if** we had "marketplace" & "dashboard" living on the same application, which is not our case right now. 
-> good review but in this case I would leave it like that.

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/filter.util.ts
```diff
@@ -0,0 +1,123 @@
...
+    })
+  );
+}
+
+function territories(movie: Movie, territory: string): boolean {
+  if (!territory) {
+    return true;
+  }
+  const territoryToLowerCase = territory.toLowerCase();
+  return movie.salesAgentDeal.territories.includes(territoryToLowerCase);
```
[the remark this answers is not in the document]
GrandSchtroumpf: @ioaNikas It's usally better to have a const to make it more readable.
@Maxstgt But here as you're litterally giving the "name + method" in the name of your variable I think Yohann is right.

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.component.ts
GrandSchtroumpf: @ioaNikas don't put a one line arrow function into a 3 lines without return. See code below :
```typescript
  private _languageFilter(filterValue: string): string[] {
    return this.movieLanguages.filter(language => {
      return language.toLowerCase().includes(filterValue.toLowerCase())
    });
  }
```

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.module.ts
GrandSchtroumpf: See comment above. Good practice to not load component with empty path. But for one component module, this is the best solution.

libs/movie/src/lib/movie/components/display-card-item/display-card-item.component.ts
GrandSchtroumpf: I think it's ok. But might be confusing for other in the team. I would recommand to have 3 files.

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.component.ts
Maxstgt: this is a copy of the material doc example. But sure, it is more readable without the const

PR #776, opened 2019-08-23
Issue#676 blockchain org

pldespaigne:


contracts/sol/Organization.sol
```diff
@@ -143,6 +153,17 @@ contract Organization {
...
+
+        adminCount++; // ! NO RELEASE DEBUG PURPOSE ONLY
+        adminList[0x4D7e2f3ab055FC5d484d15aD744310dE98dD5Bc3] = true; // ! NO RELEASE DEBUG PURPOSE ONLY
+
+        // hardcoded operation(s)
+        operations[1].name = 'Signing Delivery';
+        operations[1].quorum = 2;
+        operations[1].active = true;
+        // operations[1].whitelist[admin] = true;
+        operations[1].whitelistLength = 0;
```
GrandSchtroumpf: Not sure to understand the purpose of all of this. If it's for having an initial state you should create a script for that, and not hardcode this here (even for debug)

libs/ethers/src/lib/wallet/+state/wallet.service.ts
```diff
@@ -58,27 +45,10 @@ export class WalletService {
         return address;
       }
     }
-    return await this.precomputeAddress(ensDomain);
+    return await precomputeAddress(ensDomain, this.provider);
```
GrandSchtroumpf: you don't need await when you return. Remove it.

libs/ethers/src/lib/wallet/+state/wallet.service.ts
```diff
@@ -198,7 +164,7 @@ export class WalletService {
   }
 
   public async sendSignedMetaTx(ensDomain: string, signedMetaTx: SignedMetaTx, ...args: any[]) {
-    const txReceipt = await this.relayer.send(await this.retreiveAddress(this.getNameFromENS(ensDomain)), signedMetaTx);
+    const txReceipt = await this.relayer.send(await this.retreiveAddress(getNameFromENS(ensDomain)), signedMetaTx);
```
GrandSchtroumpf: No readable, create a variable : 
```typescript
const address = await this.retreiveAddress(getNameFromENS(ensDomain));
const txReceipt = await this.relayer.send(address, signedMetaTx);
```

libs/organization/src/lib/components/organization-signer-form/organization-signer-form.component.html
```diff
@@ -1,10 +1,11 @@
 <form [formGroup]="control">
   <article>
     <h2>{{member.name}}</h2>
+    {{ isAdmin ? '(Admin)' : '' }}
```
GrandSchtroumpf: Use *ngIf instead : `<span *ngIf="isAdmin">(Admin)</span>`

libs/organization/src/lib/components/organization-signer-repertory/organization-signer-repertory.component.ts
```diff
@@ -57,13 +58,18 @@ export class OrganizationSignerRepertoryComponent {
...
   private joinMemberAndOperation() {
 
     const getOperationIds = (member: OrganizationMember) => this._operations
     .filter(operation => operation.members.some(operationMember => member.uid === operationMember.uid))
     .map(operation => operation.id);
 
-    const operationMembers: OperationMember[] = this._members.map(member => ({...member, operationIds: getOperationIds(member)}));
+    const operationMembers: OperationMember[] = this._members.map(member => {
+      const isAdmin = this.permissionQuery.isUserSuperAdmin(member.uid);
+      return { ...member, operationIds: getOperationIds(member), role: isAdmin ? UserRole.admin : UserRole.member }
```
GrandSchtroumpf: Create variables, make it readable : 
```typescript
const operationIds = getOperationIds(member);
const role = isAdmin ?  UserRole.admin : UserRole.member;
return { ...member, operationIds, role, isAdmin };
```

libs/organization/src/lib/guard/organization.guard.ts
```diff
@@ -19,6 +19,7 @@ export class OrganizationGuard {
           if (organization.status === OrganizationStatus.pending) {
             return res(this.router.parseUrl('layout/organization/congratulation'));
           }
+          this.orgService.retreiveDataAndAddListeners();
```
GrandSchtroumpf: What does this do ? Why it's not a subscription ?

libs/organization/src/lib/pages/organization-admin-view/organization-admin-view.component.ts
```diff
@@ -42,6 +42,7 @@ export class OrganizationAdminViewComponent implements OnInit {
     /** Return the operationFormGroup linked to the selected operation.id */
     this.operations$ = this.query.select('org').pipe(
       tap(organization => this.organizationName = organization.name),
+      filter(organization => !!organization.operations),
```
GrandSchtroumpf: is it only when operations is null/undefined, or also when operations is empty ?

libs/utils/src/lib/helpers.ts
```diff
@@ -1,11 +1,23 @@
...
   end: string;
 }
 
+export interface Key {
+  name: string,
+  ensDomain: string,
+  keyStore: string,
+  address: string,
+  isMainKey: boolean,
+  isLinked: boolean, // does the key also exists inside the ERC1077
```
GrandSchtroumpf: If comment gives information about what the key is about, use TSDoc : /** Comment above the key */

libs/utils/src/lib/helpers.ts
```diff
@@ -21,3 +33,39 @@ export interface DateRange {
...
+export async function precomputeAddress(ensDomain: string, provider: providers.Provider) { // !!!! there is a copy of this function in 'apps/backend-functions/src/relayer.ts'
+
+  const factoryAddress = await provider.resolveName(factoryContract);
+
+  // CREATE2 address
+  let payload = '0xff';
+  payload += factoryAddress.substr(2);
+  payload += utils.keccak256(utils.toUtf8Bytes(getNameFromENS(ensDomain))).substr(2); // salt
+  payload += utils.keccak256(`0x${ERC1077.bytecode}`).substr(2);
+  return `0x${utils.keccak256(payload).slice(-40)}`;
```
GrandSchtroumpf: I would prefer to have several variables that explain what happen here : 
```typescript
const address = factoryAddress.substr(2);
const salt = utils.keccak256(utils.toUtf8Bytes(getNameFromENS(ensDomain))).substr(2);
const hash = utils.keccak256(`0x${ERC1077.bytecode}`).substr(2);
....
return `0x${}${}`
```

libs/organization/src/lib/guard/organization.guard.ts
pldespaigne: this function has to be called to get data from the org smart-contract and to create blockchain event listeners, I put it in the guard so I'm sure it has been called when we are in an org pages, otherwise I will need to put it in every `ngOnInit` of needed pages

libs/organization/src/lib/pages/organization-admin-view/organization-admin-view.component.ts
pldespaigne: it's for null/undefined because `operations` is an optional key, if it is empty, there is no problem

PR #789, opened 2019-08-23
Deploy toronto

lsenta:
- [x] fix a typo around catalog in angular.json
- [x] ⚠️ Disable firestore security rules so that the demo works
- [x] fix blockchain network for staging
- [x] add catalog to the main app
- [x] fix app requests name mismatches

### DIFF
--- a/apps/backend-functions/src/admin.ts
+++ b/apps/backend-functions/src/admin.ts
@@ -23,7 +23,7 @@ import {
 import { getSuperAdmins } from './data/internals';
 
 // TODO(#714): Synchronize data types with the frontend
-const APPS = ['MediaDelivering', 'MediaFinanciers', 'StoriesAndMore'];
+const APPS = ['MediaDelivering', 'MediaFinanciers', 'StoriesAndMore', 'catalog'];
 
 /**
  * Handles firestore update on request to application access,
@@ -42,7 +42,7 @@ export async function onRequestAccessToAppWrite(
   }
 
   const requestedApps = APPS.filter(appId => {
-    return after && after[appId] === 'requested' && (!before || !before[appId]);
+    return after && after[appId] === 'requested' && (!before || before[appId] !== 'requested');
   });
 
   return Promise.all(
--- a/apps/backend-functions/src/admin.ts
+++ b/apps/backend-functions/src/admin.ts
@@ -23,7 +23,7 @@ import {
 import { getSuperAdmins } from './data/internals';
 
 // TODO(#714): Synchronize data types with the frontend
-const APPS = ['MediaDelivering', 'MediaFinanciers', 'StoriesAndMore', 'catalog'];
+const APPS = ['delivery', 'movie-financing', 'StoriesAndMore', 'catalog'];
 
 /**
  * Handles firestore update on request to application access,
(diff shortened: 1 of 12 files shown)


apps/backend-functions/src/admin.ts
```diff
@@ -23,7 +23,7 @@ import {
 import { getSuperAdmins } from './data/internals';
 
 // TODO(#714): Synchronize data types with the frontend
-const APPS = ['MediaDelivering', 'MediaFinanciers', 'StoriesAndMore'];
+const APPS = ['delivery', 'movie-financing', 'StoriesAndMore', 'catalog'];
```
pldespaigne: every app name are in kebab-case except for `StoriesAndMore` is it intended ?

apps/backend-functions/src/orgs.ts
```diff
@@ -54,21 +54,27 @@ export async function onOrganizationUpdate(
...
     const admin = await db.collection('users').doc(userIds[0]).get().then(adminSnapShot => adminSnapShot.data()!); // TODO use laurent's code after the merge of PR #698
 
     const adminENS = emailToEnsDomain(admin.email, RELAYER_CONFIG.baseEnsDomain);
     const orgENS = emailToEnsDomain(before.name.replace(' ', '-'), RELAYER_CONFIG.baseEnsDomain);
-    const isOrgRegistred = await isENSNameRegistred(adminENS, RELAYER_CONFIG);
-    if (isOrgRegistred) {
-      throw new Error(`This organization has already an ENS name : ${adminENS}`);
+
+    // TODO: use orgENS
+    const isOrgRegistered = await isENSNameRegistred(adminENS, RELAYER_CONFIG);
```
pldespaigne: I just checked, there is no reason to use `adminENS` instead of `orgENS`.
Could change it, so I don't need to create a PR just for 1 line of code ?

apps/backend-functions/src/relayer.ts
```diff
@@ -65,26 +65,33 @@ export interface SignDeliveryParams {
...
   const provider = getDefaultProvider(config.network);
+
   wallet = wallet.connect(provider);
+
   const contractFactory = new Contract(config.factoryContract, CREATE2_FACTORY.abi, wallet);
   const namehash = utils.namehash(config.baseEnsDomain);
   const registry = new Contract(config.registryAddress, ENS_REGISTRY.abi, wallet);
   const resolver = new Contract(config.resolverAddress, ENS_RESOLVER.abi, wallet);
 
+  // await Promise.all([
```
pldespaigne: Can you delete those comments, or add a TODO comment to know what to do with this code ?

libs/organization/src/lib/permissions/+state/permissions.model.ts
```diff
@@ -39,6 +39,9 @@ export interface OrgDocPermissions {
   canDelete: boolean;
 }
 
+// TODO(#714): Synchronize data types with the frontend
+// Do not change these values without upgrading the backend too!
+// You'll find relevant spots by searching for the issue number.
 export const enum App {
   mediaDelivering = 'delivery',
   mediaFinanciers = 'movie-financing',
```
pldespaigne: same thing here for `StoriesAndMore`, the naming case is not consistent

PR #817, opened 2019-08-26
Part1 for issue 603 distribution rights

Maxstgt:


apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.service.ts
```diff
@@ -1,8 +1,16 @@
...
 
 @Injectable({ providedIn: 'root' })
 export class BasketService {
 
-  constructor() {}
+  constructor(private db: FireQuery, private organizationQuery: OrganizationQuery) {}
+
+  public createDistributionRight(distributionRight: DistributionRight){
+    const organizationId = this.organizationQuery.getValue();
+    this.db.doc<DistributionRight>(`orgs/${organizationId}/distributionRight/`).set(distributionRight);
```
ioaNikas: " / " not necessary at the end of your path

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.component.ts
```diff
@@ -8,8 +9,14 @@ import { DistributionRightForm } from './create.form';
 })
```
ioaNikas: Don't forget to add `changeDetection: ChangeDetectionStrategy.OnPush`

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.component.ts
```diff
@@ -8,8 +9,14 @@ import { DistributionRightForm } from './create.form';
...
 export class DistributionRightCreateComponent implements OnInit {
   public form = new DistributionRightForm();
-
+  public movieTerritories: string[] = staticModels['TERRITORIES'].map(key => key.label);
+  public movieMedias: string[] = staticModels['MEDIAS'].map(key => key.label);
   constructor() {}
```
ioaNikas: Remove if you don't need it.

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.form.ts
```diff
@@ -18,4 +21,22 @@ export class DistributionRightForm extends FormEntity<DistributionRight> {
...
+    const territories = this.get('territories').value;
+    if (!territories.includes(territory)) {
+      this.get('territories').push(new FormControl(territory));
+    }
+    // Else do nothing as it's already in the list
+  }
+
+  removeTerritory(index: number) {
+    this.get('territories').removeAt(index);
+  } */
```
ioaNikas: Is this dead code ? If so, you should remove it.
If there is a reason to keep it this way, add a comment to tell why :)

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.module.ts
```diff
@@ -18,6 +20,8 @@ import { MatFormFieldModule } from '@angular/material/form-field';
     MatSidenavModule,
     MatInputModule,
     MatFormFieldModule,
+    MatExpansionModule,
+    MatCheckboxModule,
     RouterModule.forChild([{ path: '', component: DistributionRightCreateComponent }])
```
ioaNikas: If possible, avoid loading component on empty path.
If it's not possible, ignore this comment.

apps/catalog-marketplace/catalog-marketplace/src/app/layout/context-menu.ts
```diff
@@ -1,15 +1,22 @@
...
-      { name: 'home', path: '/layout/o/catalog/home' },
-      { name: 'search', path: '/layout/o/catalog/search' }
+      {
+        name: 'view',
+        path: '/layout/o/catalog/:movieId/view'
+      },
+      {
+        name: 'create',
+        path: '/layout/o/catalog/:movieId/create'
+      }
```
ioaNikas: You either put `name` and `path` in one or multiple lines, but choose only one way to display them.

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.html
```diff
@@ -1 +1,203 @@
...
+                {{ subtitle }}{{isLast ? '' : ', '}}
+              </li>
+            </ng-container>
+          </ul>
+        </article>
+      </mat-card>
+    </section>
+  </section>
+  <a mat-raised-button color="accent" (click)="goToCreateDistribution(movie.id)">Create distribution right</a>
+</main>
```
ioaNikas: That's one big ass html ! You should consider separate it sub-components in a near future.

More important : it seems that this view will be needed in other applications. You should consider moving it to the movie library at this precise path : `libs/movie/src/lib/movie/pages/movie-view`.

I'm not 100% sure it has to be a reusable components through the apps, we should ask François about that.

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -1,15 +1,58 @@
...
+  ) {}
 
   ngOnInit() {
+    this.getMovie();
   }
 
+  private async getMovie() {
+    const movieId = this.activatedRoute.snapshot.params['movieId'];
+    this.movie$ = of(await this.fireQuery.snapshot<Movie>(`movies/${movieId}`));
+    this.movie$.subscribe(data => this.endRights = (data.salesAgentDeal.rightsEnd.to as any).toDate());
```
ioaNikas: I suggest you to use our MovieActiveGuard (or a custom version of it if movie-view will not be shared across applications).

Here you never unsubscribe from your movie observable. If you need some help to understand how guards works on the project, just give me a call :).

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -1,15 +1,58 @@
...
+  }
+
+  public goToCreateDistribution(movieId: string) {
+    this.router.navigateByUrl(`layout/o/catalog/${movieId}/create`);
+  }
```
ioaNikas: You don't have to use the router when you have a variable to pass in the url.
You can use something like this instead in the html

`[routerLink]="['/layout/o/catalog', movie.id, '/create']"`

or

`[routerLink]="['../', movie.id, '/create']"` (not 100% about this one, but try it)

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.module.ts
```diff
@@ -1,13 +1,28 @@
...
+// Component
 import { MovieViewComponent } from './view.component';
 
+// Material
+import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
+import { MatChipsModule } from '@angular/material/chips';
+import { MatCardModule } from '@angular/material/card';
+import { MatDividerModule } from '@angular/material';
+import { MatButtonModule } from '@angular/material/button';
+import { FlexLayoutModule } from '@angular/flex-layout';
```
ioaNikas: /!\ Not a material module, he belongs somewhere else

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.component.ts
```diff
@@ -61,7 +62,7 @@ export class MarketplaceSearchComponent implements OnInit {
   public movieTerritories: MovieTerritories[];
   public territoryControl: FormControl = new FormControl();
 
-  constructor(private fireQuery: FireQuery) {}
+  constructor(private fireQuery: FireQuery, private router: Router) {}
```
ioaNikas: You should simply name it `db : FireQuery` as you did in basket.service

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.component.ts
```diff
@@ -148,4 +149,8 @@ export class MarketplaceSearchComponent implements OnInit {
     this.filterForm.addTerritory(territory.option.viewValue as MovieTerritories);
     this.territoryInput.nativeElement.value = '';
   }
+
+  public goToMovieInformation(movieId: string){
+    this.router.navigateByUrl(`/layout/o/catalog/${movieId}`);
+  }
```
ioaNikas: Same as above, I think you can use [routerLink] with concatenation in html to navigate without any other operations

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -2,6 +2,7 @@ import { Organization } from "@blockframes/organization";
 import { Material } from "@blockframes/material";
 import { Stakeholder } from "../../stakeholder/+state";
 import { DateRange } from "@blockframes/utils";
+import { DistributionRight } from "apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.model";
```
ioaNikas: Try to avoid absolute path.

You can set custom paths to an app or a lib using `tsconfig.json`. Something like :

```
"@blockframes/catalog-marketplace": ["apps/catalog-marketplace/catalog-marketplace/src/index.ts"]
```

Here index.ts doesn't exist, so you can create it and import `distribution-right/+state` in it, and then just use :

```
import { DistributionRight } from '@blockframes/catalog-marketplace'
```

But you'll get circular dependencies with that king of imports as the catalog app will grow. Using libs is a good way to avoid it.

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -93,7 +94,7 @@ export interface MovieFestivalPrizes {
 }
 
 export interface MovieSalesAgentDeal {
-  rightsEnd: Date,
+  rightsEnd: any,
```
ioaNikas: Why not using Date ?

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -106,6 +107,7 @@ export interface Movie {
   materials?: Material[];
   stakeholders?: Stakeholder[];
   sales: MovieSale[], //@todo 581 => move to subcollection
+  distributionRight: DistributionRight[],
```
ioaNikas: `distributionRights` as it is an array of DistributionRight

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.component.ts
Maxstgt: I will need it later in the second PR

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.form.ts
Maxstgt: It is for the second PR

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.component.ts
GrandSchtroumpf: @ioaNikas gives the code : `<a ... [routerLink]="['layout/o/catalog', movieId]" ...>`

libs/movie/src/lib/movie/+state/movie.model.ts
GrandSchtroumpf: Come back to Date. If this is because of mock data, change the mock data.

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.model.ts
```diff
@@ -3,7 +3,7 @@ import { staticModels } from '@blockframes/movie';
 type Languages = ((typeof staticModels)['LANGUAGES'])[number]['slug'];
 type Currency = ((typeof staticModels)['MOVIE_CURRENCIES'])[number]['slug'];
 type Media = ((typeof staticModels)['MEDIAS'])[number]['slug'];
-type Territories = ((typeof staticModels)['TERRITORIES'])[number]['slug'];
+export type Territories = ((typeof staticModels)['TERRITORIES'])[number]['slug'];
```
GrandSchtroumpf: Hugo has created an issue to move these types into a `static-models` folder inside movie. Add a comment here that reference this issue.

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.service.ts
```diff
@@ -1,8 +1,16 @@
...
+import { DistributionRight } from './basket.model';
+import { OrganizationQuery } from '@blockframes/organization';
 
 @Injectable({ providedIn: 'root' })
 export class BasketService {
 
-  constructor() {}
+  constructor(private db: FireQuery, private organizationQuery: OrganizationQuery) {}
+
+  public createDistributionRight(distributionRight: DistributionRight){
```
GrandSchtroumpf: You can just call it "add" or "create", as you're already in the service that deal with that. Also the name and type of the parameter are enough to understand what it's going to add ^^.

libs/movie/src/lib/movie/+state/movie.model.ts
Maxstgt: This will cause trouble with the data we got from the database. And when I change it now, the rest of the monorepo screams about that

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.component.html
```diff
@@ -1,9 +1,40 @@
...
+          <mat-panel-title>
+            Dates
+          </mat-panel-title>
+        </mat-expansion-panel-header>
+        <span formGroupName="duration">
+          <input matInput type="text" formControlName="from" placeholder="From">
+          <input matInput type="text" formControlName="to" placeholder="To">
+        </span>
+      </mat-expansion-panel>
+      <mat-expansion-panel>
```
GrandSchtroumpf: Add a comment between each panel to make it easy to read :

libs/movie/src/lib/movie/+state/movie.model.ts
Maxstgt: No it is an interface which needs to be cast into an array

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.component.ts
```diff
@@ -8,8 +9,14 @@ import { DistributionRightForm } from './create.form';
...
 
-  ngOnInit() {}
+  ngOnInit() {
+  }
+
+  public hasTerritory(territory: string) {
+    this.form.get('territories');
```
GrandSchtroumpf: With the name, I expect "hasTerritory" to return a boolean, but it returns a undefined. And the parameter is not used. Please update.

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.module.ts
GrandSchtroumpf: You are in a One component/module. So it's ok.

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.html
```diff
@@ -1 +1,203 @@
-<p>movie-view works!</p>
+<ng-template #loading>
+  <mat-spinner color="accent" #loading></mat-spinner>
+</ng-template>
+<main *ngIf="(movie$ | async) as movie; else loading">
+  <section class="header">
+    <img class="header-img" [src]="movie.main.poster" />
+    <h1 class="mat-display-3">{{movie.main.title.original}}</h1>
+    <span class="mat-body-1" *ngFor="let director of movie.main.directors">a film by {{director.firstName}}
+      {{director.lastName}}</span>
```
GrandSchtroumpf: Format + iteration not at the right place
```html
<span class="mat-body-1">
  a film by <ng-container  *ngFor="let director of movie.main.directors">{{director.firstName}} {{director.lastName}} <ng-container>
</span>
```

libs/movie/src/lib/movie/+state/movie.model.ts
ioaNikas: If it is data from database, then consider using Timestamp type.

```
import { firestore } from 'firebase/app';

export interface MovieSalesAgentDeal {
  rightsEnd: Timestamp,
...
```

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.html
GrandSchtroumpf: The problem is that the UI and the data don't match. So to spare us some useless effort you did a one component.

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.scss
```diff
@@ -0,0 +1,48 @@
...
+
+mat-divider {
+  margin-left: 20px;
+  width: 100%;
+}
+h2 {
+  white-space: nowrap;
+}
+
+:host {
```
GrandSchtroumpf: duplicate from above

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.scss
```diff
@@ -0,0 +1,48 @@
...
+  display: flex;
+}
+
+nav {
```
GrandSchtroumpf: I didn't see a nav component

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -1,15 +1,58 @@
...
+  styleUrls: ['./view.component.scss'],
+  changeDetection: ChangeDetectionStrategy.Default
 })
 export class MovieViewComponent implements OnInit {
+  public movie$: Observable<Movie>;
+  public endRights;
 
-  constructor() { }
+  constructor(
+    private fireQuery: FireQuery,
```
GrandSchtroumpf: name it "db"

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -1,15 +1,58 @@
...
+    private activatedRoute: ActivatedRoute,
+    private router: Router
+  ) {}
 
   ngOnInit() {
+    this.getMovie();
   }
 
+  private async getMovie() {
+    const movieId = this.activatedRoute.snapshot.params['movieId'];
```
GrandSchtroumpf: Never use snapshot on a component, or it won't update when you change to another movie. See comment from Yohann to best practice

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -1,15 +1,58 @@
...
+    let name, year;
+    this.movie$.subscribe(data => {
+      name = data.salesInfo.internationalPremiere.name;
+      year = data.salesInfo.internationalPremiere.year;
+    });
+    return name !== '' ? `${name}, ${year}` : null;
+  }
+
+  get color() {
+    let color;
```
GrandSchtroumpf: You should store the movie inside the a MovieStore, and inject MovieQuery (you can take them from `@blockframes/movie`).
Then use `this.movieQuery.getActive().salesInfo.color`. Same for below.

libs/movie/src/lib/movie/+state/movie.model.ts
GrandSchtroumpf: I think he means that you need to add a "s" at the end the key "distributionRightS"

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.model.ts
Maxstgt: @Hugo-Hontans can you tell mich which number this issue has?

libs/movie/src/lib/movie/+state/movie.model.ts
Maxstgt: Like I said, this will cause errors from all over the app. Like rightsEnd doesn't have that function etc.

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.model.ts
Hugo-Hontans: The issue is #818

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.model.ts
```diff
@@ -23,7 +24,7 @@ export interface DistributionLanguage {
   subtitle: boolean;
 }
 
-export interface DistributionRight {
+export interface DistributionRights {
```
GrandSchtroumpf: It's one distribution right. No "s" on the interface

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.component.ts
GrandSchtroumpf: For this PR it's fine, but as a rule of thumb, don't put half finished code in a PR. You should wait for the next PR to create the method.

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/create/create.form.ts
GrandSchtroumpf: Same comment as above

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -1,15 +1,53 @@
...
+    this.getMovie();
   }
 
+  private async getMovie() {
+    this.loading$ = this.query.selectLoading();
+    this.movieId = this.query.getActiveId();
+    this.movie$ = this.query.selectActive();
+    this.movie$.subscribe(
+      data => (this.parsedRightEnds = (data.salesAgentDeal.rightsEnd as any).to.toDate())
+    );
```
GrandSchtroumpf: Do : 
```typescript
this.movie$ = this.query.selectActive().pipe(
  tap(({ salesAgentDeal }) => this.parseRightEnds = (salesAgentDeal.rightsEnd as any).to.toDate())
);
```
Like that you don't have to worry about the subscriptions

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -1,15 +1,53 @@
...
+  }
+
+  get internationalPremiere() {
+    let name, year;
+    this.movie$.subscribe(data => {
```
GrandSchtroumpf: Never do that. You're subscribing multiple times (each call to this function) on the same value without unsubscribing.
You should do : 
```
internationalPremiere(movie: Movie) {
  const name = ...
  const year = ...
  return ....
}
```
And pass the argument in the function

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -1,15 +1,53 @@
...
+      name = data.salesInfo.internationalPremiere.name;
+      year = data.salesInfo.internationalPremiere.year;
+    });
+    return name !== '' ? `${name}, ${year}` : null;
+  }
+
+  get color() {
+    let color;
+    this.movie$.subscribe(data => (color = data.salesInfo.color));
+    return color === 'c' ? 'color' : 'black & white';
```
GrandSchtroumpf: Same as above.

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -1,15 +1,53 @@
...
+  }
+
+  get europeanQualification() {
+    let europeanQualification;
+    this.movie$.subscribe(data => (europeanQualification = data.salesInfo.europeanQualification));
+    return europeanQualification ? 'Yes' : 'No';
```
GrandSchtroumpf: Same as above

apps/catalog-marketplace/catalog-marketplace/src/app/pages/marketplace-search/marketplace-search.component.ts
GrandSchtroumpf: @Maxstgt You didn't update that.

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -106,6 +110,7 @@ export interface Movie {
   materials?: Material[];
   stakeholders?: Stakeholder[];
   sales: MovieSale[], //@todo 581 => move to subcollection
+  distributionRight: DistributionRights[],
```
GrandSchtroumpf: the "key" "distributionRights" should have a "s" as this is a list of something.
The "interface" "DistributionRight" shouldn't have a "s" as this is one entity.

apps/catalog-marketplace/catalog-marketplace/src/app/movie/view/view.component.ts
```diff
@@ -24,30 +25,29 @@ export class MovieViewComponent implements OnInit {
...
+      );
   }
 
-  get internationalPremiere() {
-    let name, year;
-    this.movie$.subscribe(data => {
-      name = data.salesInfo.internationalPremiere.name;
-      year = data.salesInfo.internationalPremiere.year;
-    });
+  public internationalPremiere(movie: Movie) {
```
GrandSchtroumpf: If you are doing getActive in the other function, do that here too. Either use the getter or a function with parameter. But don't mix them in the same component ;)

PR #821, opened 2019-08-27
Requeted changes without step 1

VicoDraye:


libs/material/src/lib/delivery/pages/delivery-add-complete/delivery-add-complete.component.scss
```diff
@@ -0,0 +1,3 @@
+a {
```
pldespaigne: Have you try this view with different screen size ? Is it responsive ?

libs/material/src/lib/delivery/pages/delivery-add-settings/delivery-add-settings.component.html
```diff
@@ -4,7 +4,7 @@ <h2>Settings</h2>
     <p>Lorem ipsum dolor sit amet</p>
 
     <actions-picker-list [items]="items" (picked)="picked($event)"></actions-picker-list>
-    <button (click)="onCompleteFlow()" mat-button color="primary" [disabled]="options.length === 0">
+    <button (click)="onCompleteFlow()" mat-raised-button color="warn" [disabled]="options.length === 0">
```
pldespaigne: color should be `accent` I think

libs/ui/src/lib/actions-picker/actions-picker.component.scss
```diff
@@ -1,7 +1,9 @@
-mat-radio-group {
-  margin: 15px 0;
-}
+@import 'variables';
 
 mat-radio-button {
   margin: 5px;
 }
+
+:host{
```
pldespaigne: Usually `:host` is the first css rule, move it above `mat-radio-button`
Also since you don't use things like `var(--someVariable)` you don't need to `@import 'variables';`

styles/components/mat-list.scss
```diff
@@ -2,3 +2,13 @@
...
+
+.mat-list {
+  border-radius: 10px;
+  border: 1px solid var(--borderGrey);
+}
+
+.mat-selection-list{
+  border-radius: 10px;
+  border: 1px solid var(--borderGrey);
+}
```
pldespaigne: This is ok, but usually it's @Hugo-Hontans who is in charge of that. Is he aware of those change ?

VicoDraye: Yes, Hugo and Vincent give me the autorisation

PR #826, opened 2019-08-27
Issue#786 template editable

Hugo-Hontans:
New version for template-editable that uses editable-sidenav


libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -164,22 +164,35 @@ export class MaterialService {
   // CRUD MATERIAL (TEMPLATE) //
   //////////////////////////////
 
+  /** Remove material of template */
   public deleteTemplateMaterial(id: string) {
     const templateId = this.templateQuery.getActiveId();
-    this.db.doc<Material>(`templates/${templateId}/materials/${id}`).delete();
+    return this.db.doc<Material>(`templates/${templateId}/materials/${id}`).delete();
```
phpgeekfr: Why do you return something whereas you do not use the return of the function in `deleteMaterial` ? If return is usefull, do `public deleteTemplateMaterial(id: string) : Promise<void>`

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -164,22 +164,35 @@ export class MaterialService {
...
-  public saveTemplateMaterial(material: Material) {
-    const templateId = this.templateQuery.getActiveId();
-    const materialId = this.db.createId();
-    this.db
-      .doc<Material>(`templates/${templateId}/materials/${materialId}`)
-      .set({ ...material, id: materialId });
+  /** Returns a material of template to be pushed in a formGroup */
+  public addTemplateMaterial(): Material {
+    const id = this.db.createId();
+    const newMaterial = createTemplateMaterial({ id });
```
phpgeekfr: return createTemplateMaterial({ id }); directly

libs/material/src/lib/material/+state/material.model.ts
```diff
@@ -52,6 +52,18 @@ export function createMaterial(material: Partial<Material>): Material {
   };
 }
 
+export function createTemplateMaterial(material: Partial<Material>): Material {
+  return {
+    id: material.id,
+    category: '',
```
phpgeekfr: do you really need to initialize string or number ? 
try: 
``` 
return {
    id: material.id,
    ...material
  } as material;
```

libs/material/src/lib/template/forms/material.form.ts
```diff
@@ -0,0 +1,18 @@
+import { FormList } from '@blockframes/utils';
+import { FormControl, FormGroup } from '@angular/forms';
+import { Material } from '@blockframes/material';
+
+export function createMaterialFormGroup(material: Material) {
+  return new FormGroup({
+    id: new FormControl(material.id),
+    value: new FormControl(material.value),
+    description: new FormControl(material.description),
+    price: new FormControl(material.price),
```
phpgeekfr: you can use `FormField<number>` or `FormField<string>` instead of `FormControl`

libs/material/src/lib/template/pages/template-editable/template-editable.component.ts
```diff
@@ -19,77 +16,67 @@ import { MaterialService } from '../../../material/+state';
...
   ) {}
 
   ngOnInit() {
-    this.templateActive$ = this.query.selectActive();
-    this.template$ = this.query.materialsByTemplate$;
-    this.form$ = this.materialQuery.templateForm$;
-  }
+    this.template$ = this.query.selectActive();
+    this.materials$ = this.query.selectActive().pipe(
+      filter(template => !!template.materials),
```
phpgeekfr: Maybe add a comment to explain why you need this (or why you should not if there was'nt a bug)

libs/material/src/lib/material/+state/material.service.ts
Hugo-Hontans: Yes sure, I did not wait my promise in ``deleteMaterial``

libs/material/src/lib/template/pages/template-editable/template-editable.component.ts
Hugo-Hontans: I added a comment :
```
We need to filter materials because when we go into the template list, the guard does not load materials in templates
```

libs/material/src/lib/template/forms/material.form.ts
Hugo-Hontans: Yes but we do not do it in the other forms of material like the delivery. So I prefer to keep the same consistency. Maybe we can make a pass later

libs/material/src/lib/material/+state/material.model.ts
Hugo-Hontans: We should not do that because we could use the function with an argument that did not create a real material and we then force the type. It's better to initialize default values

PR #829, opened 2019-08-28
Feat 794

phpgeekfr:


apps/movie-financing/movie-financing/src/app/app.module.ts
```diff
@@ -20,6 +20,7 @@ import { MatToolbarModule } from '@angular/material/toolbar';
 import { MatIconModule } from '@angular/material/icon';
 import { MatListModule } from '@angular/material/list';
 import { MatButtonModule } from '@angular/material/button';
+import { MatSidenavModule } from '@angular/material/sidenav';
 // import { MatButtonModule } from '@angular/material/button';
```
ioaNikas: Dead code (also duplicate from above import).

apps/movie-financing/movie-financing/src/app/layout/layout.component.scss
```diff
@@ -3,16 +3,28 @@ header {
...
 }
+
+account-profile-menu {
+  margin-left: 20px;
+}
+
+mat-sidenav {
+  position: fixed;
+  width: 360px;
+  border-radius: 0 5px 5px 0;
```
ioaNikas: We don't want to redifine css for material components. But if "vanilla" style for this component is too ugly, maybe keep it this way.

PR #835, opened 2019-08-28
Dark theme

GrandSchtroumpf:
Add dark theme to calalog-marketplace


apps/catalog-marketplace/catalog-marketplace/src/app/movie/home/home.component.html
```diff
@@ -1,38 +1,41 @@
...
-    <article fxLayout="column" fxLayoutAlign="center center" class="header-img">
-        <h1 class="mat-display-3">Welcome to Bigger Boat</h1>
-        <p class="mat-body-1">
-            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt
-            ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris
-            nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehcentererit in voluptate velit
-            esse cillum dolore eu fugiat nulla pariatur.</p>
-        <button mat-button color="accent" test-id="discover">Discover our films</button>
-    </article>
+<div test-id="content" page-id="catalog-homepage">
```
Hugo-Hontans: Maybe use a section instead of a div

apps/catalog-marketplace/catalog-marketplace/src/app/movie/home/home.component.ts
```diff
@@ -51,4 +52,12 @@ export class CatalogMarketplaceHomeComponent implements OnInit {
       })
     );
   }
+
+  layout(index: number) {
```
Hugo-Hontans: Do a ```public``` function

apps/catalog-marketplace/catalog-marketplace/src/app/movie/home/home.component.ts
```diff
@@ -51,4 +52,12 @@ export class CatalogMarketplaceHomeComponent implements OnInit {
...
+    return index%2 === 0? 'row': 'row-reverse';
+  }
+
+  alignment(index: number) {
```
Hugo-Hontans: The same

apps/catalog-marketplace/catalog-marketplace/src/app/movie/search/search.component.html
```diff
@@ -0,0 +1,187 @@
+<div test-id="content" page-id="catalog-search">
```
Hugo-Hontans: Use a section instead of a div

apps/catalog-marketplace/catalog-marketplace/src/app/movie/search/search.component.html
```diff
@@ -0,0 +1,187 @@
...
+
+            <!-- Languages -->
+            <mat-expansion-panel>
+              <mat-expansion-panel-header>
+                <mat-panel-title>
+                  Languages
+                </mat-panel-title>
+              </mat-expansion-panel-header>
+              <form formGroupName="languages" fxLayout="column">
+                <span fxLayout="row" fxLayoutAlign="start start">
```
Hugo-Hontans: I think it's not a good practice of using a span to group different mat-form-field

apps/catalog-marketplace/catalog-marketplace/src/app/movie/search/search.component.html
```diff
@@ -0,0 +1,187 @@
...
+              </div>
+            </mat-expansion-panel>
+          </mat-accordion>
+        </mat-card-content>
+      </mat-card>
+    </aside>
+
+    <!-- List of result -->
+    <article>
+      <h3 class="movies" fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="10px">
```
Hugo-Hontans: You are using a tag h3 in an other h3, use a header would be better

apps/catalog-marketplace/catalog-marketplace/src/app/movie/search/search.component.html
```diff
@@ -0,0 +1,187 @@
...
+        <h3>Search</h3>
+        <mat-divider></mat-divider>
+      </h3>
+      <div *ngIf="movieSearchResults$ | async as searchResults" fxLayout="row wrap" fxLayoutGap="30px">
```
Hugo-Hontans: I think use ng-container can be better

styles/components/mat-card.scss
```diff
@@ -1,3 +1,7 @@
+.dark-theme .mat-card {
+  background-color: rgba(249,248,255,0.05);
```
Hugo-Hontans: I think it's better to use a global css variable, like this you can change the color only in one place

styles/style.scss
```diff
@@ -24,7 +24,7 @@ html, body {
 
 body {
   margin: 0;
-  background-color: var(--whiteBackground);
+  // background-color: var(--whiteBackground);
```
Hugo-Hontans: Delete comment if you don't need it

styles/style.scss
```diff
@@ -48,7 +48,7 @@ h4 {
 
 p {
   font-size: 16px;
-  color: var(--textGrey);
+  // color: var(--textGrey);
```
Hugo-Hontans: The same

styles/components/mat-expansion-panel.scss
```diff
@@ -1,7 +1,28 @@
+.dark-theme .mat-accordion .mat-expansion-panel,
+.light-theme .mat-accordion .mat-expansion-panel {
+  box-shadow: none;
+  border-radius: 0;
+  background-color: transparent;
+}
+
+.dark-theme .mat-accordion .mat-expansion-panel {
```
Hugo-Hontans: You can use global css variable and do this : 
```
.mat-expansion-panel {
  border-bottom: solid 1px var(--borderColor);
}
```
and in variables.sccs : 
```
.dark-theme {
  @extend :root;
  --borderColor: mycolor;
}
```

styles/components/mat-expansion-panel.scss
```diff
@@ -1,7 +1,28 @@
...
+
+.light-theme .mat-accordion .mat-expansion-panel {
+  border-bottom: solid 1px #D8D8D8;
+}
+
+.mat-accordion .mat-expansion-panel:last-of-type {
+  border-bottom: none;
+}
+
+/*
```
Hugo-Hontans: Delete comments if we don't need it

styles/components/mat-list.scss
```diff
@@ -1,14 +1,14 @@
 .mat-nav-list {
   border-radius: 10px;
-  border: 1px solid var(--borderGrey);
+  // border: 1px solid var(--borderGrey);
```
Hugo-Hontans: same

styles/components/mat-list.scss
```diff
@@ -1,14 +1,14 @@
...
 }
 
 .mat-list {
   border-radius: 10px;
-  border: 1px solid var(--borderGrey);
+  // border: 1px solid var(--borderGrey);
```
Hugo-Hontans: same

styles/components/mat-list.scss
```diff
@@ -1,14 +1,14 @@
...
 }
 
 .mat-selection-list{
   border-radius: 10px;
-  border: 1px solid var(--borderGrey);
+  // border: 1px solid var(--borderGrey);
```
Hugo-Hontans: same

apps/catalog-marketplace/catalog-marketplace/src/app/movie/home/home.component.html
GrandSchtroumpf: This element is going to be remove later. That's why I switch to div

apps/catalog-marketplace/catalog-marketplace/src/app/movie/search/search.component.html
GrandSchtroumpf: Same as above

styles/components/mat-card.scss
GrandSchtroumpf: No. We should not create global variable for everything. We should only use variables from Angular material to keep things standard and coherent across themes. In this case, Mathilde wants ALL mat-card in dark-theme to behave like that, so it make sense to define it here.

styles/components/mat-expansion-panel.scss
GrandSchtroumpf: borderColor is not a variable exposed by angular material.
The only reason I would create a variable is if Mathilde says "I want this color everywhere". But then I would first look for the variables provided by material :
https://github.com/angular/components/blob/master/src/material/core/theming/_palette.scss#L674

styles/components/mat-expansion-panel.scss
GrandSchtroumpf: I kept comments in case it breaks the design in other apps. I didn't have the time to check.

apps/catalog-marketplace/catalog-marketplace/src/app/movie/search/search.component.html
GrandSchtroumpf: No, because you wouldn't be able to use `fxLayout="row wrap"`

apps/catalog-marketplace/catalog-marketplace/src/app/movie/search/search.component.html
```diff
@@ -0,0 +1,188 @@
...
+        <mat-card-content>
+          <mat-accordion multi="true">
+            <!-- Production Year -->
+            <mat-expansion-panel>
+              <mat-expansion-panel-header>
+                <mat-panel-title>
+                  Production Year
+                </mat-panel-title>
+              </mat-expansion-panel-header>
+              <span fxLayout="row" fxLayoutAlign="start start" fxLayoutGap="10px" formGroupName="productionYear">
```
Hugo-Hontans: You also use a span to contain mat-form-field here

apps/catalog-marketplace/catalog-marketplace/src/app/movie/search/search.component.html
```diff
@@ -0,0 +1,188 @@
...
+            </mat-expansion-panel>
+
+            <!-- Availabilities -->
+            <mat-expansion-panel>
+              <mat-expansion-panel-header>
+                <mat-panel-title>
+                  Availabilities
+                </mat-panel-title>
+              </mat-expansion-panel-header>
+              <span fxLayout="row" fxLayoutAlign="start start" fxLayoutGap="10px" formGroupName="availabilities">
```
Hugo-Hontans: Here too for span

PR #836, opened 2019-08-29
add stakeholder status to delivery info

pldespaigne:


libs/material/src/lib/delivery/components/delivery-informations-stakeholders/delivery-informations-stakeholders.component.ts
```diff
@@ -0,0 +1,21 @@
...
+})
+export class DeliveryInformationsStakeholdersComponent {
+  
+  @Input() delivery: Delivery;
+
+  public displayedColumns: string[] = ['logo', 'stakeholder', 'hasSigned'];
+
+  constructor() {}
+
+  hasSigned(stakeholderId: string) {
```
lsenta: `public`

libs/material/src/lib/delivery/pages/delivery-informations-editable/delivery-informations-editable.component.html
```diff
@@ -4,6 +4,7 @@
     <delivery-informations-deadlines-repertory [delivery]="delivery$ | async" (editing)="openSidenav('deadlines')"></delivery-informations-deadlines-repertory>
     <delivery-informations-dates-display [delivery]="delivery$ | async" (editing)="openSidenav('dates')"></delivery-informations-dates-display>
     <delivery-informations-steps-repertory [steps]="(delivery$ | async).steps" (editing)="openSidenav('steps')"></delivery-informations-steps-repertory>
+    <delivery-informations-stakeholders [delivery]="delivery$ | async"></delivery-informations-stakeholders>
```
lsenta: could you wrap the whole thing and `let delivery` once, instead of duplicating all these `delivery$ | async` ?

PR #841, opened 2019-08-29
Issue#801 signed delivery

ioaNikas:
- [x] We can now have signed deliveries (behaving like old ones)
- [x] Materials and Informations controls disabled when delivery is Signed


libs/material/src/lib/delivery/pages/delivery-editable/delivery-editable.component.html
```diff
@@ -10,26 +10,22 @@ <h2>{{ movie.main.title.original }}</h2>
       <a
         mat-icon-button
         matTooltip="Export delivery in a .pdf file"
-        [href]="pdfLink"
+        [href]="['/delivery/contract.pdf?deliveryId=' + delivery.id]"
```
lsenta: could you keep this in the ts code, so that the html stays "dumb"

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -104,14 +102,14 @@ export class MaterialService {
...
-          tx.update(materialRef, { deliveryIds: source.deliveryIds.filter(id => id !== deliveryId) });
+          tx.update(materialRef, { deliveryIds: source.deliveryIds.filter(id => id !== delivery.id) });
         }
       });
     });
   }
 
   /** Update the property status of materials */
-  public updateStatus(materials: Material[], status: MaterialStatus, movieId: string) {
+  public updateMaterialStatus(materials: Material[], status: MaterialStatus, movieId: string) {
```
lsenta: Why do you need to prefix everything with Material? You're in the Material service.

libs/material/src/lib/delivery/pages/delivery-informations-editable/delivery-informations-editable.component.html
```diff
@@ -1,9 +1,21 @@
 <editable-sidenav test-id="content" page-id="delivery-informations" (closed)="opened = false" [opened]="opened" (saved)="updateInformations()">
   <!-- Content -->
   <ng-container content>
-    <delivery-informations-deadlines-repertory [delivery]="delivery$ | async" (editing)="openSidenav('deadlines')"></delivery-informations-deadlines-repertory>
-    <delivery-informations-dates-display [delivery]="delivery$ | async" (editing)="openSidenav('dates')"></delivery-informations-dates-display>
-    <delivery-informations-steps-repertory [steps]="(delivery$ | async).steps" (editing)="openSidenav('steps')"></delivery-informations-steps-repertory>
+    <delivery-informations-deadlines-repertory
```
lsenta: could you have a single `delivery$ | async` and wrap all these components?

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -307,9 +307,9 @@ export class DeliveryService {
   }
 
   /** Remove signatures in array validated of delivery */
-  public unsealDelivery(): Promise<any> {
+  public unsealDelivery(): Promise<void> {
     // TODO(issue#775): ask all stakeholders for permission to re-open the delivery form
-    return this.currentDeliveryDoc.update({ validated: [] });
+    return this.currentDeliveryDoc.update({ validated: [], isSigned: false });
```
lsenta: When do you set `isSigned` to `true`?

PR #843, opened 2019-08-30
Delivery template list bis

Mussuranae:
Table instead of card for template list


libs/material/src/lib/template/pages/template-list/template-list.component.html
```diff
@@ -1,15 +1,57 @@
...
+  <section *ngIf="(templates$ | async) as template">
+    <table mat-table [dataSource]="template" class="mat-elevation-z8">
+
+      <!-- CheckBox -->
+      <ng-container matColumnDef="checkBox">
+        <th mat-header-cell *matHeaderCellDef>
+          <mat-checkbox (change)="masterToggle()"
+          [checked]="selection.hasValue() && isAllSelected()"
+          [indeterminate]="selection.hasValue() && !isAllSelected()"
+          aria-label="select all movies">
```
ioaNikas: You mean "Select all Templates" ?

libs/material/src/lib/template/pages/template-list/template-list.component.html
```diff
@@ -1,15 +1,57 @@
...
+          </mat-checkbox>
+        </th>
+        <td mat-cell *matCellDef="let element">
+          <mat-checkbox
+          (click)="$event.stopPropagation()"
+          (change)="toggle(element)"
+          [checked]="selection.isSelected(element)"
+          aria-label="Select one Movie">
```
ioaNikas: Same as above

libs/material/src/lib/template/pages/template-list/template-list.component.html
```diff
@@ -1,15 +1,57 @@
...
+          </mat-checkbox>
+        </td>
+      </ng-container>
+
+      <!-- Template name -->
+
+      <ng-container matColumnDef="template-name">
+        <th mat-header-cell *matHeaderCellDef>Template name</th>
+
+          <td mat-cell *matCellDef="let element">{{element.name | titlecase}}</td>
```
ioaNikas: Care about indentation, `<td>...</td>` is not wrapped in `<th>...</th>`.

libs/material/src/lib/template/pages/template-list/template-list.component.html
```diff
@@ -1,15 +1,57 @@
...
+
+      </ng-container>
+
+      <!-- Date
+      <ng-container matColumnDef="date">
+        <th mat-header-cell *matHeaderCellDef mat-sort-header>Date</th>
+        <td mat-cell *matCellDef="let element">{{element.name | titlecase}}</td>
+      </ng-container> -->
```
ioaNikas: Template model have a `created` property which is the date when the template is created.
You can call it this way : `{{ element.created.toDate() | date: 'MMM dd yyyy' }}` for example.
If you got bug with this, just delete this code as we don't keep dead code :)

ioaNikas:
If this PR is deprecated, close it.

PR #848, opened 2019-08-30
Copy materials

Hugo-Hontans:
Copy template materials into movie when the delivery can't be signed

### DIFF
--- a/libs/material/src/lib/delivery/+state/delivery.service.ts
+++ b/libs/material/src/lib/delivery/+state/delivery.service.ts
@@ -11,7 +11,7 @@ import {
 } from '@blockframes/movie';
 import { OrganizationQuery, PermissionsService } from '@blockframes/organization';
 import { BFDoc, FireQuery } from '@blockframes/utils';
-import { MaterialQuery, MaterialService } from '../../material/+state';
+import { MaterialQuery, MaterialService, createMaterial } from '../../material/+state';
 import { TemplateQuery } from '../../template/+state';
 import { DeliveryOption, DeliveryWizard, DeliveryWizardKind } from './delivery.store';
 import { AngularFirestoreCollection, AngularFirestoreDocument } from '@angular/fire/firestore';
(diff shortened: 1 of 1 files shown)


libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -211,8 +195,15 @@ export class DeliveryService {
...
-      // Copy movie materials in the delivery
-      await this.copyMaterialsInDelivery(delivery, movieSnap.data() as Movie, tx);
+      // If mustBeSigned is true, copy materials to the delivery
+      if (delivery.mustBeSigned) {
+        await this.copyMaterialsInDelivery(delivery, movieSnap.data() as Movie, tx);
+      }
+
+      // If mustBeSigned is false, add deliveryId in deliveryIds of each material
+      if (!delivery.mustBeSigned) {
+        await this.copyDeliveryId(delivery.id, movieSnap.data() as Movie, tx);
```
ioaNikas: Better name it pushDeliveryId as it push delivery.id into material.deliveryIds

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -406,8 +416,12 @@ export class DeliveryService {
...
 
     materials.forEach(material => {
-      const sameValuesMaterial = movieMaterials.find(movieMaterial => this.materialService.isTheSame(movieMaterial, material));
-      const isNewMaterial = !movieMaterials.find(movieMaterial => movieMaterial.id === material.id) && !sameValuesMaterial;
+      const sameValuesMaterial = movieMaterials.find(movieMaterial =>
+        this.materialService.isTheSame(movieMaterial, material)
+      );
+      const isNewMaterial =
+        !movieMaterials.find(movieMaterial => movieMaterial.id === material.id) &&
+        !sameValuesMaterial;
```
ioaNikas: Keep them in one line please, more easier to read

PR #856, opened 2019-08-30
Issue#840 import member avatar

VicoDraye:


libs/account/src/lib/profile/components/profile-form/profile-form.component.ts
```diff
@@ -13,4 +14,13 @@ export class ProfileFormComponent {
   public get control() {
     return this.controlContainer.control;
   }
+  
+  public get avatar() {
+    return this.control.get('avatar').value;
+  }
+
+  public set avatar(avatarPath: string | undefined){
+    console.log(avatarPath);
```
pldespaigne: remove console.log

libs/organization/src/lib/components/organization-form/organization-form.component.ts
```diff
@@ -13,4 +13,12 @@ export class OrganizationFormComponent {
   public get control() {
     return this.controlContainer.control;
   }
+
+  public get logo() {
+    return this.control.get('logo').value;
+  }
+
+  public set logo(logoPath: string){
+    this.control.get('logo').patchValue(logoPath);
```
pldespaigne: should you not do the same as for the profile form ?
something like `this.control.get('avatar').patchValue(avatarPath || PLACEHOLDER_AVATAR);`

PR #861, opened 2019-09-02
set null in the form for don't update undefined and put the good form…

Hugo-Hontans:
fix material add in delivery


libs/material/src/lib/material/guards/delivery-materials.guard.ts
```diff
@@ -34,7 +34,9 @@ export class DeliveryMaterialsGuard extends StateListGuard<Material> {
         const query = delivery.mustBeSigned
           ? deliveryToBeSignedMaterialsQuery(delivery)
           : deliveryMaterialsQuery(delivery);
-        return this.fireQuery.fromQuery<Material[]>(query)
+          return this.fireQuery.fromQuery<Material[]>(query).pipe(
+            map(materials => materials.map(material => getMaterialStep(material, delivery)))
+          );
```
ioaNikas: You don't need to use this function as we do not use stepId anymore (at the moment).
We already got the full step object in the material.

libs/material/src/lib/delivery/forms/material.form.ts
```diff
@@ -12,10 +12,10 @@ export function createMaterialFormGroup(material: Material, disabled = false) {
     id: new FormControl(material.id),
     value: new FormControl(withDisabled(material.value)),
     description: new FormControl(withDisabled(material.description)),
-    step: new FormControl(withDisabled(material.step)),
+    step: new FormControl(withDisabled(material.step || null)),
     category: new FormControl(withDisabled(material.category)),
-    price: new FormControl(withDisabled(material.price)),
-    currency: new FormControl(withDisabled(material.currency)),
+    price: new FormControl(withDisabled(material.price || null)),
+    currency: new FormControl(withDisabled(material.currency || null)),
```
ioaNikas: Why `|| null` on price and currency ?

Hugo-Hontans: For be sure we're never going to update a material with an undefined

libs/material/src/lib/material/guards/delivery-materials.guard.ts
Hugo-Hontans: This function change the step of a material by the good step of delivery, like this when we update a step, the step of material changes too. It's a quick fix and it will have to be refactored

PR #866, opened 2019-09-02
fields are on salesAgent now

Maxstgt:
Made some changes to the availability search

### DIFF
--- a/apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.service.ts
+++ b/apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.service.ts
@@ -9,7 +9,7 @@ export class BasketService {
   constructor(private db: FireQuery, private organizationQuery: OrganizationQuery) {}
 
   public add(basket: CatalogBasket) {
-
-    }
+    const organization = this.organizationQuery.getValue().org;
+      this.db.doc<Organization>(`orgs/${organization.id}`).update({ catalog: basket });
   }
 }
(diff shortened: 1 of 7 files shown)


apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.service.ts
```diff
@@ -9,7 +9,12 @@ export class BasketService {
   constructor(private db: FireQuery, private organizationQuery: OrganizationQuery) {}
 
   public add(basket: CatalogBasket) {
-    const organizationId = this.organizationQuery.getValue().org.id;
-    this.db.doc<Organization>(`orgs/${organizationId}`).update( { catalog:  basket } )
+    const organization = this.organizationQuery.getValue().org;
+    if (typeof organization.catalog === 'undefined') {
```
ioaNikas: You could either use the basic syntax `if (!organization.catalog)` as `undefined == false`

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.service.ts
```diff
@@ -1,15 +1,24 @@
...
 export class BasketService {
   constructor(private db: FireQuery, private organizationQuery: OrganizationQuery) {}
 
-  public add(basket: CatalogBasket) {
-    const organizationId = this.organizationQuery.getValue().org.id;
-    this.db.doc<Organization>(`orgs/${organizationId}`).update( { catalog:  basket } )
+  public add(distributionRight: DistributionRight) {
+    const organization = this.organizationQuery.getValue().org;
+    const newDistributionRight = [...organization.catalog.rights, distributionRight];
+    console.log(newDistributionRight);
```
phpgeekfr: is console.log mandatory ? ;)

PR #876, opened 2019-09-03
Fix drag event

Hugo-Hontans:


libs/organization/src/test-setup.ts
```diff
@@ -1,7 +1,8 @@
 import 'jest-preset-angular';
 import 'jest';
 
-// Mock DragEvent until jsdom provide a valid implementation
+// Mock any until jsdom provide a valid implementation
 Object.defineProperty(window, 'DragEvent', {
-  value: class DragEvent {}
+  // TODO: issue#875, use DragEvent type
+  value: class Any {}
```
phpgeekfr: try to lunch 2e2 tests if it is ok

PR #880, opened 2019-09-04
Material search / Invitations fix

ioaNikas:
- [x] Added a filter in materials list (delivery and movie)
- [x] Fixed invitations (every invitation are created in the frontend, permissions still in the backend)


libs/material/src/lib/template/pages/template-list/template-list.component.ts
```diff
@@ -34,6 +34,6 @@ export class TemplateListComponent implements OnInit {
   }
 
   public addTemplateDialog(): void {
-    this.dialog.open(TemplateAddComponent, { width: '500px' });
+    this.dialog.open(TemplateAddComponent);
```
Hugo-Hontans: Pay attention, you deleted the ``{ width: '500px' }``, do you try on tablet ?

apps/backend-functions/src/notify.ts
```diff
@@ -40,31 +37,14 @@ export function prepareNotification(notif: BaseNotification): Notification {
     id: db.collection('notifications').doc().id,
     isRead: false,
     date: serverTimestamp(),
-    appIcon: notif.docInformations.type === 'delivery' ? App.mediaDelivering : App.mediaFinanciers,
+    appIcon:
+      notif.docInformations.type === 'delivery' || notif.docInformations.type === null
```
Hugo-Hontans: Create an issue to remove this type null of your interface DocInformations

libs/material/src/lib/template/pages/template-list/template-list.component.ts
ioaNikas: Dialog size will be defined by the size of the image in it. We don't need to specify the size then.

PR #886, opened 2019-09-04
Catalog selection bid

jenna1k:


apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.store.ts
```diff
@@ -5,11 +5,12 @@ import { DistributionRight, Price, BasketStatus } from './basket.model';
...
   price: Price;
   status: BasketStatus;
+  right: DistributionRight;
 }
 
 @Injectable({ providedIn: 'root' })
-@StoreConfig({ name: 'basket' })
-export class BasketStore extends EntityStore<BasketState, DistributionRight> {
+@StoreConfig({ name: 'basket', idKey: 'id' })
+export class BasketStore extends EntityStore<BasketState, any> {
```
phpgeekfr: Why did you changed `DistributionRight` to `any`

Maxstgt: I was working on the guard, and tried several things, I gonna revert this.

PR #896, opened 2019-09-04
Editable sidenav scroll

Hugo-Hontans:


libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -117,6 +117,10 @@ export class DeliveryService {
   // CRUD DELIVERY //
   ///////////////////
 
+  public getDelivery(id: string): Delivery {
```
phpgeekfr: remove this since it's unused

PR #901, opened 2019-09-05
fixed bug with org dist right

Maxstgt:


apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.service.ts
```diff
@@ -6,30 +6,28 @@ import { OrganizationQuery } from '@blockframes/organization';
...
       });
     } else {
-      this.db.doc<Organization>(`orgs/${this.organization.id}/`).update({
+      const newDistributionRight = [...this.organizationQuery.getValue().org.catalog.rights, distributionRight];
+      this.db.doc<Organization>(`orgs/${this.organizationQuery.getValue().org.id}/`).update({
         catalog: {
-          price: { amount: 0, currency: 'us-dollar' },
-          status: BasketStatus.pending,
+          price: this.organizationQuery.getValue().org.catalog.price,
+          status: this.organizationQuery.getValue().org.catalog.status,
```
ioaNikas: Why do you need to update those fields too ? (real question, not a comment)

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.service.ts
```diff
@@ -39,18 +37,18 @@ export class BasketService {
   public addBid(price: Price) {
     const updatedCatalog: CatalogBasket = {
       price: { amount: price.amount, currency: price.currency },
-      status: this.organization.catalog.status,
-      rights: this.organization.catalog.rights
+      status: this.organizationQuery.getValue().org.catalog.status,
+      rights: this.organizationQuery.getValue().org.catalog.rights
```
ioaNikas: Declare a variable before for better readability

`const organization = this.organizationQuery.getValue().org;`

Maxstgt: I guess that somehow was related to the bug. Cause if I'm updating firestore the store/ the variable which is linked to the store, didn't get the latest version of it... that's why I used all the store queries here. If you got a better idea of the readability, let me know

apps/catalog-marketplace/catalog-marketplace/src/app/distribution-right/+state/basket.service.ts
Maxstgt: he doesn't let me update the rights property. I always had to update the whole property. Got a better solution?

PR #911, opened 2019-09-09
Cleaning components

ioaNikas:
- [x] Deleted app and references :  eth-events-server
- [x] Deleted app and references : ip-upload
- [x] Deleted all pages and components in stakeholder folder
- [x] Deleted unused components and pages from material lib
- [x] New style on Save as template dialog to fit our global design


libs/movie/src/lib/stakeholder/stakeholder.module.ts
```diff
@@ -1,41 +1,10 @@
...
-import { MatButtonModule } from '@angular/material/button';
-import { MatDividerModule } from '@angular/material/divider';
-import { StakeholderViewComponent } from './pages/stakeholder-view/stakeholder-view.component';
-import { StakeholderItemComponent } from './components/stakeholder-item/stakeholder-item.component';
-import { MatCheckboxModule } from '@angular/material/checkbox';
-import { MatSidenavModule } from '@angular/material/sidenav';
-import { DirectivesModule, TeamWorkModule } from '@blockframes/ui';
-import { OrganizationModule } from '@blockframes/organization';
 
 @NgModule({
```
GrandSchtroumpf: If you don't use it, remove it.

PR #912, opened 2019-09-09
tx redirect & tx feedback

pldespaigne:


libs/ethers/src/lib/wallet/wallet-add-key/wallet-add-key.component.ts
```diff
@@ -90,7 +83,7 @@ export class WalletAddKeyTunnelComponent implements OnInit {
     if (!!this.walletQuery.getValue().tx) {
       this.router.navigateByUrl('/layout/o/account/wallet/send');
     } else {
-      this.router.navigateByUrl(this.redirectRoute);
+      this.router.navigateByUrl('/layout/o/account/wallet');
```
Maxstgt: I can't see the scope, but why don't use the `walelt.feedback.redirectRoute` for it? I'm not sure if it is possible without an extra import. But you are using it in the HTML

libs/organization/src/lib/components/member-form-role/member-form-role.component.ts
```diff
@@ -63,13 +63,29 @@ export class MemberFormRoleComponent {
       this.permissionsService.updateMembersRole(members);
     };
 
+    const orgName = this.query.getValue().org.name;
+    const orgId = this.query.getValue().org.id;
```
Maxstgt: Just a suggestion but there is getter for it. `this.query.id`

libs/organization/src/lib/components/organization-form-operation/organization-form-operation.component.ts
```diff
@@ -39,14 +40,24 @@ export class OrganizationFormOperationComponent {
...
+    const removedMember = this.control.get('members').value.find(member => member.uid === id) as OrganizationMember;
     const members = this.control.get('members').value.filter(member => member.uid !== id);
     this.control.get('members').patchValue(members);
 
     const memberAddress = await this.service.getMemberAddress(removedMember.email);
     const orgAddress = await this.service.getAddress();
     const operationId = this.control.get('id').value;
-    this.walletService.setTx(CreateTx.removeMember(orgAddress, operationId, memberAddress));
+    const operationName = this.control.get('name').value;
+    const orgId = this.query.getValue().org.id;
```
Maxstgt: Same suggestion as above

libs/organization/src/lib/components/organization-form-operation/organization-form-operation.component.ts
```diff
@@ -59,15 +70,35 @@ export class OrganizationFormOperationComponent {
     const memberAddress = await this.service.getMemberAddress(addedMember.email);
     const orgAddress = await this.service.getAddress();
     const operationId = this.control.get('id').value;
-    this.walletService.setTx(CreateTx.addMember(orgAddress, operationId, memberAddress));
+    const operationName = this.control.get('name').value;
+    const orgId = this.query.getValue().org.id;
```
Maxstgt: ;)

libs/organization/src/lib/components/organization-form-operation/organization-form-operation.component.ts
```diff
@@ -59,15 +70,35 @@ export class OrganizationFormOperationComponent {
...
     this.router.navigateByUrl('/layout/o/account/wallet/send');
   }
 
   public async updateQuorum() {
     const orgAddress = await this.service.getAddress();
     const operationId = this.control.get('id').value;
     const newQuorum = this.control.get('quorum').value;
-    this.walletService.setTx(CreateTx.modifyQuorum(orgAddress, operationId, newQuorum));
+    const operationName = this.control.get('name').value;
+    const orgId = this.query.getValue().org.id;
```
Maxstgt: :P

libs/organization/src/lib/components/organization-signer-form/organization-signer-form.component.ts
```diff
@@ -49,11 +50,32 @@ export class OrganizationSignerFormComponent {
 
     const memberAddress = await this.service.getMemberAddress(this.member.email);
     const orgAddress = await this.service.getAddress();
+    const memberName = this.member.name;
+    const operationName = currentOperation.name;
+    const orgId = this.query.getValue().org.id;
```
Maxstgt: :)

libs/ethers/src/lib/wallet/wallet-add-key/wallet-add-key.component.ts
pldespaigne: I use `wallet.feedback.redirectRoute` to redirect the user at the end of a *send tx tunnel*, here we are in the *add key tunnel*, in this tunnel I want to either go back to the wallet (if the user came from here) or go back to sending a tx (if the user was doing that, if it's the case, then the tx tunnel will redirect correctly at the end). I hope it's clear ^^

PR #919, opened 2019-09-10
CatalogBasket CRUD functions

Maxstgt:
The basket for the distribution rights is now a subcollection of the individual organization.
Inside the basket, there is the `rights` array which holds the distribution rights configurations made by one buyer of an organization. The baskets holds information about the configurations of the distribution rights for example price and status. If there is another buyer for the same organization and he creates his own distribution rights from some movies, this will be a new basket because the price will change.


apps/catalog-marketplace/catalog-marketplace/src/app/movie/selection/selection.component.ts
```diff
@@ -0,0 +1,88 @@
...
+  selector: 'catalog-selection',
+  templateUrl: './selection.component.html',
+  styleUrls: ['./selection.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class CatalogSelectionComponent implements OnInit {
+  public priceControl: FormControl = new FormControl(null);
+  public currencyList: string[];
+  public selectedCurrency;
+  public movieDistributionRights$: MovieData[] = [];
```
ioaNikas: Table data indeed needs to be an Observable of MovieData array. If you can't fix this issue right now, open an issue on ZenHub and refence it here.

During this time, remove the & as movieDistributionRights isn't an observable.

Maxstgt: I added the TODO, but I left the variable, cause otherwise, you won't see anything in the frontend, hope this is okay?

ioaNikas: Sure, don't let an issue block you for too long. Maybe ask a senior to help you on that as I wasn't able to fix it with you :)

PR #925, opened 2019-09-11
e2e catalog-marketplace : signup and logout

jenna1k:


apps/catalog-marketplace/catalog-marketplace-e2e/src/support/pages/LandingPage.ts
```diff
@@ -0,0 +1,10 @@
+import { LoginPage } from ".";
+
+export default class LandingPage {
+  constructor() {}
```
ioaNikas: Don't forget to fill the constructors, its a good practice.

apps/catalog-marketplace/catalog-marketplace-e2e/src/support/pages/OrganizationHomePage.ts
```diff
@@ -0,0 +1,7 @@
+import NavbarPage from "./NavBarPage";
+
+export default class OrganizationHomePage extends NavbarPage{
+    constructor() {
+        super();
+    }
```
ioaNikas: Same as above, you can try to get() an element from OrganizationHomePage to make sure you're on the right page.

PR #940, opened 2019-09-16
Issue#936 delivery editable

Hugo-Hontans:
Use FormBatch in delivery-editable and movie-editable

### DIFF
--- a/libs/material/src/lib/delivery/components/delivery-material-form/delivery-material-form.component.ts
+++ b/libs/material/src/lib/delivery/components/delivery-material-form/delivery-material-form.component.ts
@@ -2,6 +2,8 @@ import { Component, ChangeDetectionStrategy, Output, EventEmitter, Input } from 
 import { ControlContainer } from '@angular/forms';
 import { Delivery, Currencies } from '../../+state';
 import { Material } from '../../../material/+state';
+import { FormElement } from '@blockframes/utils';
+import { MaterialControl } from '../../forms/material.form';
 
 @Component({
   selector: '[formGroupName] delivery-material-form, [formGroup] delivery-material-form, delivery-material-form',
@@ -20,8 +22,8 @@ export class DeliveryMaterialFormComponent{
 
   constructor(public controlContainer: ControlContainer) {}
 
-  public get control() {
-    return this.controlContainer.control;
+  public get control(): FormElement<MaterialControl> {
+    return this.controlContainer.control as FormElement<MaterialControl>;
   }
 
 }
--- a/libs/material/src/lib/delivery/components/delivery-material-form/delivery-material-form.component.ts
+++ b/libs/material/src/lib/delivery/components/delivery-material-form/delivery-material-form.component.ts
@@ -22,8 +22,7 @@ export class DeliveryMaterialFormComponent{
 
   constructor(public controlContainer: ControlContainer) {}
 
-  public get control(): FormElement<MaterialControl> {
+  public get control() {
     return this.controlContainer.control as FormElement<MaterialControl>;
   }
-
 }
--- a/libs/material/src/lib/delivery/forms/material.form.ts
+++ b/libs/material/src/lib/delivery/forms/material.form.ts
@@ -1,5 +1,5 @@
-import { FormBatch } from '@blockframes/utils';
-import { FormControl, FormGroup } from '@angular/forms';
+import { FormBatch, FormElement } from '@blockframes/utils';
+import { FormControl } from '@angular/forms';
 import { Material } from '@blockframes/material';
 import { createMaterial } from '../../material/+state';
 
@@ -31,6 +31,6 @@ export class MaterialForm extends FormBatch<Material, MaterialControl> {
 
   createControl(material: Partial<Material>) {
     const controls = createMaterialControl(material);
-    return new FormGroup(controls);
+    return new FormElement(controls);
   }
 }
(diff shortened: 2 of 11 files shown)


libs/material/src/lib/delivery/pages/delivery-editable/delivery-editable.component.ts
```diff
@@ -10,34 +10,29 @@ import { Router } from '@angular/router';
...
   public pdfLink: string;
-  private selectedMaterialId$ = new BehaviorSubject<string>(null);
-
-  public materialFormGroup$: Observable<FormGroup>;
-  public currentFormList: FormList<Material, any> = createMaterialFormList();
-
-  private destroyed$ = new Subject();
 
+  public form = new MaterialForm();
+  public activeForm$: Observable<AbstractControl>;
```
GrandSchtroumpf: Be more precise on the kind of Control : `Observable<DeliveryControl>` ?

libs/material/src/lib/delivery/pages/delivery-editable/delivery-editable.component.ts
```diff
@@ -54,60 +49,49 @@ export class DeliveryEditableComponent implements OnInit, OnDestroy {
...
-        return this.currentFormList.value.findIndex(material => material.id === materialId);
+    this.materials$ = combineLatest([this.query.selectActive(), this.materialQuery.selectAll()]).pipe(
+      tap(([delivery, materials]) => {
+        this.form.upsertValue(materials);
+        // Disable or enable form depending on delivery isSigned property
+        delivery.isSigned ? this.form.disable() : this.form.enable();
       }),
-      map(index => {
-        return this.currentFormList.at(index);
+      switchMap(([delivery, materials]) => {
```
GrandSchtroumpf: One line : `switchMap(([delivery, materials]) => this.form.selectAll)|`

libs/material/src/lib/delivery/pages/movie-editable/movie-editable.component.ts
```diff
@@ -19,10 +19,8 @@ export class MovieEditableComponent implements OnInit {
   public movie$: Observable<Movie>;
   public opened = false;
 
-  public materialsFormList = createMaterialFormList();
-  public materialFormGroup$: Observable<FormGroup>;
-
-  private selectedMaterialId$ = new BehaviorSubject<string>(null);
+  public form = new MaterialForm();
+  public activeForm$: Observable<AbstractControl>;
```
GrandSchtroumpf: Same here

libs/material/src/lib/delivery/forms/material.form.ts
```diff
@@ -1,29 +1,37 @@
...
-    isOrdered: new FormControl(material.isOrdered),
-    isPaid: new FormControl(material.isPaid),
-    status: new FormControl(material.status),
-    owner: new FormControl(material.owner),
-    storage: new FormControl(material.storage)
-  });
+function createMaterialControl(material: Partial<Material> = {}) {
+  const entity = createMaterial(material);
+  return {
+    id: new FormControl(entity.id),
```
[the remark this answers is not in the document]
Hugo-Hontans: We need the display to return the id of the selected material to setActive the right form with id. So the more simple way is to add id in the form even if we never change it.

libs/material/src/lib/delivery/components/delivery-material-form/delivery-material-form.component.ts
```diff
@@ -20,8 +22,8 @@ export class DeliveryMaterialFormComponent{
 
   constructor(public controlContainer: ControlContainer) {}
 
-  public get control() {
-    return this.controlContainer.control;
+  public get control(): FormElement<MaterialControl> {
+    return this.controlContainer.control as FormElement<MaterialControl>;
```
Hugo-Hontans: I need to force the type of the control here and this is not good.
Whereas the ``getActive()`` of my formBatch returns a FormElement

GrandSchtroumpf: If you remove the output line 25 it should be enough.

libs/material/src/lib/delivery/pages/delivery-editable/delivery-editable.component.ts
```diff
@@ -54,60 +49,47 @@ export class DeliveryEditableComponent implements OnInit, OnDestroy {
...
   /* Update a list of materials **/
   public update() {
     try {
       const delivery = this.query.getActive();
+      const materials = this.form.getAll();
       delivery.mustBeSigned
-        ? this.materialService.updateDeliveryMaterials(this.currentFormList.value, delivery)
-        : this.materialService.update(this.currentFormList.value, delivery);
+        ? this.materialService.updateDeliveryMaterials(materials, delivery)
+        : this.materialService.update(materials, delivery);
```
GrandSchtroumpf: I think this is super bad practice to do that as we are create as many writings as there is materials, event if they didn't change. That one of the reason I think it's actually better to fallback to a single form instead of a list...

GrandSchtroumpf: To do in another PR

Hugo-Hontans: Sure, maybe we can check in a transaction which material we have to update

Hugo-Hontans: No we already do that, in the service we check if the material exists and if it has the same value, we don't update it

PR #942, opened 2019-09-16
issue 607 sort movies - part 1

Maxstgt:
this is part 1 of the issue 607. I need to split it up because I had to rename the movie list guard and the movie active guard, because they are working on the organization movieID property and the naming was not fitting. I did some setup for sort behavior of the search movie page. But this will update when @ioaNikas has worked on the akita-fire lib.


apps/catalog/marketplace/marketplace/src/app/guards/catalog-movie-list.guard.ts
```diff
@@ -0,0 +1,22 @@
+import { Router } from '@angular/router';
+import { Injectable } from '@angular/core';
+import { Query, StateListGuard, FireQuery } from '@blockframes/utils';
+import { Movie, MovieStore } from '@blockframes/movie';
+
+export const catalogMovieListQuery = (): Query<Movie> => ({
+  path: `movies`
+});
+@Injectable({ providedIn: 'root' })
```
pldespaigne: break line please

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.html
```diff
@@ -177,13 +177,31 @@ <h6>You can choose multiple certifications</h6>
       <div class="movies" fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="10px">
         <h3>Films</h3>
         <mat-divider></mat-divider>
+        <button mat-icon-button (click)="cardOrList = !cardOrList" *ngIf="!cardOrList">
+          <mat-icon>list</mat-icon>
+        </button>
+        <button mat-icon-button (click)="cardOrList = !cardOrList" *ngIf="cardOrList">
```
pldespaigne: those two buttons does the same thing, would it not be better to have one button that change its icon ?

pldespaigne: also with the name `cardOrList` it is hard to tell if it's a card when the value is true or when the value is false

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -30,11 +34,21 @@ export class MyErrorStateMatcher implements ErrorStateMatcher {
...
   public movieSearchResults$: Observable<Movie[]>;
+
+  /* Instance of the search form */
   public filterForm = new CatalogSearchForm();
+
+  /* Array of sorting options */
+  public sortOptions: string[] = ['All films', 'Title', 'Director', 'Production Year'];
+
+  /* Flag to indicate either the movies should be presented as a card or a list */
+  public cardOrList: boolean;
```
pldespaigne: prefer a name like `isCard` or `isList`

libs/movie/src/lib/movie/components/display-list/display-list.component.scss
```diff
@@ -0,0 +1,10 @@
+@import 'variables';
```
pldespaigne: unused import

libs/movie/src/lib/movie/components/display-list/display-list.component.ts
```diff
@@ -0,0 +1,19 @@
+import { Movie } from './../../+state/movie.model';
+import { Component, Input } from '@angular/core';
+import { MatTableDataSource } from '@angular/material/table';
+
+@Component({
+  selector: '[movie] movie-display-list',
```
pldespaigne: it's just a question : What does `[movie]` means in the selector ?

libs/movie/src/lib/movie/components/display-list/display-list.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  styleUrls: ['./display-list.component.scss']
+})
+export class MovieDisplayListComponent {
+  public displayedColumns: string[] = ['picture', 'title', 'director', 'productionYear'];
+  public dataSource;
+
+  @Input()
+  set movie(movie: Movie[]) {
+    this.dataSource = new MatTableDataSource(movie);
+    console.log(this.dataSource);
```
pldespaigne: remove `console.log`

libs/movie/src/lib/movie/components/display-list/display-list.component.ts
Maxstgt: you force your parent component to give you an input with that name

PR #947, opened 2019-09-17
E2e save as template

Hugo-Hontans:
e2e: Test save as template functionality

### DIFF
--- a/apps/delivery/delivery-e2e/src/integration/save-as-template.spec.ts
+++ b/apps/delivery/delivery-e2e/src/integration/save-as-template.spec.ts
@@ -1,6 +1,5 @@
 /// <reference types="cypress" />
 import {
-  DeliveryFormPage,
   DeliveryListPage,
   HomePage,
   LandingPage,
@@ -8,23 +7,25 @@ import {
   TemplateDeleteModal,
   TemplateFormPage,
   TemplateListPage,
-  LoginPage
+  LoginPage,
+  DeliveryMaterialsPage
 } from '../support/pages';
-import { User } from '../support/utils/type';
+import { User, Material } from '../support/utils/type';
+import SaveAsTemplateModal from '../support/pages/SaveAsTemplateModal';
 
-const MATERIALS = [
+const MATERIALS: Material[] = [
   {
-    value: 'First Material Value',
+    title: 'First Material Value',
     description: 'First Material Description',
     category: 'Category#1'
   },
   {
-    value: 'Second Material Value',
+    title: 'Second Material Value',
     description: 'Second Material Description',
     category: 'Category#2'
   },
   {
-    value: 'Third Material Value',
+    title: 'Third Material Value',
     description: 'Third Material Description',
     category: 'Category#3'
   }
@@ -38,7 +39,7 @@ const MATERIAL_CHANGED = {
 
 const USER: Partial<User> = {email: 'cytest@blockframes.com', password: 'azerty'}
 
-const TEMPLATE_NAME_1 = 'SaveAsTemplate';
+const TEMPLATE_NAME_1 = 'Saveastemplate';
 const MOVIE_CYTEST = 'I, robot';
 const ORG_CYTEST = 'cytestorg';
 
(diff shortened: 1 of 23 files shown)


apps/delivery/delivery-e2e/src/integration/save-as-template.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    title: 'Second Material Value',
+    description: 'Second Material Description',
+    category: 'Category#2'
+  },
+  {
+    title: 'Third Material Value',
+    description: 'Third Material Description',
+    category: 'Category#3'
+  }
+];
```
ioaNikas: As both delivery CRUD and template CRUD (and soon materials CRUD) use this kind of mock data, we should move it to a new file dedicated to mock data.

apps/delivery/delivery-e2e/src/integration/save-as-template.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+const MATERIAL_CHANGED = {
+  value: 'Value Changed',
+  description: 'Description Changed',
+  category: 'Category#Changed'
+};
```
ioaNikas: Seems like you don't use this const.

apps/delivery/delivery-e2e/src/integration/save-as-template.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    const p12: SaveAsTemplateModal = p11.clickSaveAsTemplate();
+    p12.fillName(TEMPLATE_NAME_1);
+    const p13: DeliveryMaterialsPage = p12.clickUpdate();
+
+    // Verify if the template is overwrote with right materials
+    const p14: HomePage = p13.clickHome();
+    const p15: TemplateListPage = p14.clickContextMenuTemplates();
+    const p16: TemplateFormPage = p15.editTemplate(TEMPLATE_NAME_1);
+    p16.assertMaterial(MATERIALS[1]);
+    p16.assertMaterial(MATERIALS[2]);
```
ioaNikas: You should rename `assertMaterial` into `assertMaterialExists` since if you use it in a sentence, it should sounds like "assert that this material exists"

apps/delivery/delivery-e2e/src/support/pages/NavbarPage.ts
```diff
@@ -1,4 +1,4 @@
-import { HomePage, ViewProfilePage, DeliveryTeamWorkPage, LoginPage, OrganizationMemberPage, OrganizationFormPage } from ".";
+import { HomePage, ViewProfilePage, TemplateListPage, DeliveryTeamWorkPage, LoginPage, OrganizationMemberPage, OrganizationFormPage, DeliveryListPage } from ".";
```
ioaNikas: Seems like we are using different path to import. Some files imports from `'.'` and others imports from `'./FileName'`
We should stick to only one format of import (maybe check if one is heavier than the other and keep the lighter one)

apps/delivery/delivery-e2e/src/support/pages/TemplateFormPage.ts
```diff
@@ -1,9 +1,20 @@
 import TemplateDeleteModal from "./TemplateDeleteModal";
 import TemplateListPage from "./TemplateListPage";
+import { Material } from "../utils/type";
+import NavbarPage from "./NavbarPage";
 
-export default class TemplateFormPage {
+export default class TemplateFormPage extends NavbarPage {
   constructor() {
-    cy.contains('Add a material')
```
ioaNikas: Constructor is missing

libs/movie/src/lib/movie/pages/movie-list/movie-list.component.html
```diff
@@ -1,4 +1,4 @@
-<div fxLayout="row" fxLayoutAlign="center">
+<div page-id="movie-list" fxLayout="row" fxLayoutAlign="center">
```
ioaNikas: As we said yesterday, we should no longer use `page-id` in html. Use `@HostBinding` instead :

```
 export class MovieListComponent {
  @HostBinding('attr.page-id') pageId = 'movie-list';
 }
```

apps/delivery/delivery-e2e/src/integration/save-as-template.spec.ts
Hugo-Hontans: Yes but I would like to wait PR's of Cleclia and Pierre-Louis for do that

apps/delivery/delivery-e2e/src/support/pages/NavbarPage.ts
Hugo-Hontans: In spec files, we need to import by the index.
In pages we can import other pages directly excepted  NavBarPage that needs to import by index

PR #951, opened 2019-09-17
DRY create types on every entry in the static model

Maxstgt:
- Create custom types
- exported slug and label out of static model to reuse them in the dapps.


libs/movie/src/lib/movie/static-model/types.ts
```diff
@@ -0,0 +1,85 @@
...
+export type GenresLabel = typeof GENRESLABEL[number];
+export type GenresSlug = typeof GENRESSLUG[number];
+
+export const STAKEHOLDERROLESLABEL = models['STAKEHOLDER_ROLES'].map(key => key.label);
+export const STAKEHOLDERROLESSLUG = models['STAKEHOLDER_ROLES'].map(key => key.slug);
+
+export type StakeholderRolesLabel = typeof STAKEHOLDERROLESLABEL[number];
+export type StakeholderRolesSlug = typeof STAKEHOLDERROLESSLUG[number];
+
+export const STAKEHOLDERDELIVERYAUTHORIZATIONSLABEL = models[
```
Hugo-Hontans: So long name, I can't read it, can you add ``_`` like :
``STAKEHOLDER_DELIVERY_AUTHORIZATIONS_LABEL``.
And keep the same syntax for all

PR #955, opened 2019-09-18
migration to ethers v5

pldespaigne:


apps/backend-functions/src/generateHash.ts
```diff
@@ -1,8 +1,10 @@
-import { tmpNameSync } from 'tmp';
 import * as admin from 'firebase-admin';
-import { utils } from 'ethers';
-import * as fs from 'fs';
 import { ObjectMetadata } from 'firebase-functions/lib/providers/storage';
+import * as fs from 'fs';
```
GrandSchtroumpf: Import only the methods you need.

apps/backend-functions/src/generateHash.ts
```diff
@@ -21,7 +23,7 @@ function readFile(p: string, encoding: string): Promise<string> {
 
 export const hashFile = async (p: string): Promise<string> => {
```
GrandSchtroumpf: Put "path" instead of "p".
Add comment

apps/backend-functions/src/generateHash.ts
```diff
@@ -21,7 +23,7 @@ function readFile(p: string, encoding: string): Promise<string> {
...
   const content = await readFile(p, 'hex');
-  return utils.keccak256('0x' + content);
+  return keccak256('0x' + content);
 };
 
 export const hashToFirestore = async (object: ObjectMetadata) => {
```
GrandSchtroumpf: Add comment

apps/backend-functions/src/generateHash.ts
```diff
@@ -72,13 +74,16 @@ export const hashToFirestore = async (object: ObjectMetadata) => {
...
 
   console.info('Updating our IP document', ipID, versionID, 'with:', x);
 
   return db.runTransaction(async tx => {
-    const versionRef = db.collection('ip').doc(ipID).collection('version').doc(versionID);
+    const versionRef = db
+      .collection('ip')
+      .doc(ipID)
+      .collection('version')
+      .doc(versionID);
```
GrandSchtroumpf: do : 
```typescript
const versionRef = db.doc(`ip/${ipID}/version/${versionID}`);
```
What is an IP ? The IP application has been put into stand by.

apps/backend-functions/src/relayer.ts
```diff
@@ -1,13 +1,20 @@
...
+import { getDefaultProvider } from 'ethers';
+import { Wallet } from '@ethersproject/wallet';
+import { Contract, ContractFactory } from '@ethersproject/contracts';
+import { TransactionResponse, TransactionReceipt } from '@ethersproject/abstract-provider';
+import { namehash } from '@ethersproject/hash';
+import { keccak256 } from '@ethersproject/keccak256';
+import { toUtf8Bytes } from '@ethersproject/strings';
+import { getAddress } from '@ethersproject/address';
 import { toASCII } from 'punycode';
 import * as CREATE2_FACTORY from './contracts/Factory2.json';
```
GrandSchtroumpf: Except if you're using everything in the json file, only import what you need (I don't think you need the bytecode for example).

GrandSchtroumpf: Might be a good idea to put the contracts inside the "assets" folder. I'm not sure.

apps/backend-functions/src/relayer.ts
```diff
@@ -109,9 +116,9 @@ export async function precomputeAddress(ensDomain: string, config: RelayerConfig
   // CREATE2 address
   let payload = '0xff';
   payload += factoryAddress.substr(2);
-  payload += utils.keccak256(utils.toUtf8Bytes(ensDomain)).substr(2); // salt
-  payload += utils.keccak256(`0x${ERC1077.bytecode}`).substr(2);
-  return `0x${utils.keccak256(payload).slice(-40)}`;
+  payload += keccak256(toUtf8Bytes(ensDomain)).substr(2); // salt
+  payload += keccak256(`0x${ERC1077.bytecode}`).substr(2);
+  return `0x${keccak256(payload).slice(-40)}`;
```
GrandSchtroumpf: I think I already asked that, but it's much easier to read it you create intermediary variables and concat them at the end : 
```typescript
const address = factoryAddress.substr(2);
const salt = keccak256(toUtf8Bytes(ensDomain)).substr(2);
const ... = keccak256(`0x${ERC1077.bytecode}`).substr(2);
const payload = keccack256(`0xff${address}${salt}${...}`);
return `0x${payload.slice(-40)}`;
return `0xff${}`
```

apps/backend-functions/src/generateHash.ts
pldespaigne: I have added a comment to describe the function but I'm not quite sure.
@lsenta you wrote this function 7 month ago do you remember it's purpose ?

apps/backend-functions/src/generateHash.ts
pldespaigne: @lsenta ?

PR #956, opened 2019-09-18
pipe findLabelBySlug

Maxstgt:
Example
 ```
{{ 'actions' | findLabelBySlug: 'GENRES' }}
```
I'm open to a new name for the pipe. Don't know if it is fitting.
I also added some underlines for readability.

### CODE REVIEW

File: libs/utils/src/lib/pipes/find-label-by-slug.pipe.ts [hunk 1]
```diff
@@ -0,0 +1,11 @@
+import { staticModels } from '@blockframes/movie';
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'findLabelBySlug'
+})
+export class FindLabelBySlugPipe implements PipeTransform {
+  transform(value: string, property: string): string {
+    return staticModels[property].find(key => key.slug === value).label;
```
phpgeekfr: You could also have used existing staticModel `getLabelByCode` function instead of `return staticModels[property].find(key => key.slug === value).label;`

And if you try to find a label from a non-existing slug, this function will throw an error (trying to use label of undefined).

+ maybe add trim() and toLowerCase()  on `value`

PR #961, opened 2019-09-19
added sort functionality provided by akita

Maxstgt:
now that we got the movies in the akita store, we can use the built-in functionality.


apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -1,10 +1,17 @@
 // Angular
 import { Router } from '@angular/router';
 import { MatAutocompleteSelectedEvent, MatAutocomplete } from '@angular/material/autocomplete';
-import { ErrorStateMatcher, MatAccordion } from '@angular/material';
-import { Validators, FormGroupDirective, NgForm } from '@angular/forms';
+import { MatAccordion } from '@angular/material';
```
ioaNikas: You got material imports into a component. If possible put them into the module (maybe impossible if used in `@ViewChild`, and use deep import for MatAccordion.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -48,126 +47,110 @@ export class MarketplaceSearchComponent implements OnInit {
...
-    startWith(this.filterForm.value),
-    map(({ languages }) => Object.keys(languages))
-  );
-  public matcher = new MyErrorStateMatcher();
 
-  @ViewChild('territoryInput', { static: false }) territoryInput: ElementRef<HTMLInputElement>;
-  @ViewChild('auto', { static: false }) matAutocomplete: MatAutocomplete;
-  @ViewChild(MatAccordion, { static: false }) accordion: MatAccordion;
+  /* Individual form controls for filtering */
+  public languageControl: FormControl = new FormControl('', [
```
GrandSchtroumpf: Any reason not to put that in a FormGroup with territory ?

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -48,126 +47,110 @@ export class MarketplaceSearchComponent implements OnInit {
...
-    this.movieLanguages = LANGUAGES_LABEL;
-    this.movieCertifications = CERTIFICATIONS_LABEL;
-    this.movieMedias = MEDIAS_LABEL;
-    this.movieTerritories = TERRITORIES_LABEL;
+    this.movieSearchResults$ = combineLatest([
+      this.sortByControl.valueChanges.pipe(
+        startWith('All films'),
+        switchMap(sortIdentifier =>
+          this.movieQuery.selectAll({
+            sortBy: (a, b) => {
```
GrandSchtroumpf: Create an external pure function for that you don't rely on `this` inside. Also type "a" and "b" for better readability.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -48,126 +47,110 @@ export class MarketplaceSearchComponent implements OnInit {
...
+            }
+          })
+        )
+      ),
+      this.filterForm.valueChanges.pipe(startWith(this.filterForm.value))
+    ]).pipe(
+      map(([movies, filterOptions]) => movies.filter(movie => filterMovie(movie, filterOptions)))
+    );
+
+    this.movieSearchResults$.subscribe(movies => (this.availableMovies = movies.length));
```
GrandSchtroumpf: never subscribe without unsubscription.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -48,126 +47,110 @@ export class MarketplaceSearchComponent implements OnInit {
...
 
   constructor(private movieQuery: MovieQuery, private router: Router) {}
 
   ngOnInit() {
-    this.movieGenres = GENRES_LABEL;
-    this.movieLanguages = LANGUAGES_LABEL;
-    this.movieCertifications = CERTIFICATIONS_LABEL;
-    this.movieMedias = MEDIAS_LABEL;
-    this.movieTerritories = TERRITORIES_LABEL;
+    this.movieSearchResults$ = combineLatest([
```
GrandSchtroumpf: Create two variables to make it more explicit  : 
```typescript
const sortBy$ = this.sortByControl.valueChanges...
const filterBy$ = this.filterForm.valueChanges.pipe(startWith(this.filterForm.value));
this.movie$ = combineLatest([sortBy$, filterBy$]).pipe(
  map(([movies, filterOptions]) => movies.filter(movie => filterMovie(movie, filterOptions))),
  tap(movies => this.availableMovies = movies.length), // Or create an observable for that instead.
);
```

libs/utils/src/lib/form/validators/validators.ts
```diff
@@ -56,5 +60,16 @@ export function validMnemonic(control: AbstractControl): ValidationErrors | null
 export async function UniqueOrgName(control: AbstractControl): Promise<ValidationErrors | null> {
   const orgENS = orgNameToEnsDomain(control.value);
   const orgAddress = await ethers.getDefaultProvider(network).resolveName(orgENS);
-  return !orgAddress ? null : { notUnique: true};
+  return !orgAddress ? null : { notUnique: true };
+}
+
+export function languageValidator(control: AbstractControl): { [key: string]: boolean } | null {
```
GrandSchtroumpf: Add comments

libs/utils/src/lib/form/validators/validators.ts
```diff
@@ -56,5 +60,16 @@ export function validMnemonic(control: AbstractControl): ValidationErrors | null
...
+  return !LANGUAGES_LABEL.includes(control.value) ? { languageNotSupported: true } : null;
+}
+
+export class StandardErrorStateMatcher implements ErrorStateMatcher {
```
GrandSchtroumpf: Add comments

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
Maxstgt: yes, because these formControls should be only available inside their own mat-expansion-panel. What would be the benefit of make a formgroup?

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
Maxstgt: there is no this inside of the function.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
GrandSchtroumpf: It's a good reason. It's easier to manage in a formGroup when applicable.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
GrandSchtroumpf: yes, that's why you can create a pure function out of this. It doesn't have to be in the class.
```typescript
function sortMovieBy(a: keyof Movie, b: keyof Movie) {
  ...
}
```

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
Maxstgt: In the html there is already filterformGroup bindend. Then I have to do: 
 ```div [formGroup]=filter  +   <div [formGroup]=languageTerritoryGroup```
which is not good for readbility.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
Maxstgt: why keyof? that will throw an error... cause property doesn't exist.

GrandSchtroumpf: I thought a and b where keys of movie. If not, an enum, or a litteral should be enough.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -19,174 +27,140 @@ import {
...
+
+  /* Observable to combine for the UI */
+  private sortBy$ = this.sortByControl.valueChanges.pipe(
+    startWith('All films'),
+    switchMap(sortIdentifier =>
+      this.movieQuery.selectAll({
+        sortBy: (a: Movie, b: Movie) => {
+          return sortMovieBy(a, b, sortIdentifier);
+        }
+      })
```
GrandSchtroumpf: You can do :
```typescript
private sortBy$ = this.sortBycontrol.valueChanges.pipe(
  startWith('All films'),
  switchMap(sortKey => this.movieQuery.selectAll(sortBy: (a, b) => sortMovieBy(a, b, sortKey ))
)
```
In this case you don't have to type a and b, as it's typed in `sortMovieBy`.

PR #962, opened 2019-09-19
Standardize e2e naming pages

Hugo-Hontans:


apps/delivery/delivery-e2e/src/integration/CRUD-org.spec.ts
```diff
@@ -18,8 +18,8 @@ beforeEach(() => {
   cy.clearLocalStorage();
   cy.visit('/auth');
   cy.viewport('macbook-15');
-  const p1: LandingPage = new LandingPage();
-  const p2: LoginPage = p1.clickCallToAction();
+  const p1: WelcomeViewPAge = new WelcomeViewPAge();
```
ioaNikas: Typo here, WelcomeView**PA**ge

apps/delivery/delivery-e2e/src/support/pages/template/TemplateCreatePage.ts
```diff
@@ -1,4 +1,4 @@
-import NewTemplatePage from './NewTemplatePage';
+import TemplateAddModal from './TemplateAddModal';
```
ioaNikas: Maybe think about renaming it `TemplateTitleForm` as we got the exact same component to add movies. Best thing should be to create a shared component for both case like you did with ConfirmComponent.

I think Deliveries will soon have a name and this component could also be used in that case.

libs/organization/src/lib/pages/organization-editable/organization-editable.component.ts
```diff
@@ -13,6 +13,7 @@ import { startWith, tap } from 'rxjs/operators';
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class OrganizationEditableComponent implements OnInit {
+  @HostBinding('attr.page-id') pageId = 'template-editable';
```
ioaNikas: You mean `'organization-editable'`

Hugo-Hontans: The eye of lynx

apps/delivery/delivery-e2e/src/support/pages/template/TemplateCreatePage.ts
Hugo-Hontans: I don't understand if you are talking about pages in tests or about components in libs

PR #966, opened 2019-09-20
Standardize e2e naming 2

Hugo-Hontans:


apps/delivery/delivery-e2e/src/support/pages/NavbarPage.ts
```diff
@@ -1,56 +1,61 @@
...
+  ////////////////////
+
   public openProfileMenu() {
     cy.get('[page-id=navbar]').get('button[test-id=profile-avatar]').click();
   }
 
+  public clickProfile() {
+    cy.get('button[test-id=profile]').click();
+    return new ProfileEditablePage();
+  }
```
ioaNikas: Is this working without calling `openProfileMenu()` function  to open the menu before clicking on profile ?

apps/delivery/delivery-e2e/src/support/pages/NavbarPage.ts
```diff
@@ -1,56 +1,61 @@
...
+  public clickContextMenuInformation(): DeliveryInformationsEditablePage {
+    cy.get('[page-id=navbar] a').contains('information').click();
+    return new DeliveryInformationsEditablePage();
   }
 
   public clickContextMenuTemplatesCreate() {
-    cy.get('[page-id=navbar]').contains('templates').click();
+    cy.get('[page-id=navbar] a').contains('templates').click();
     return new TemplateCreatePage();
   }
```
ioaNikas: We should either specify the return type of all the functions or never specify it. The choice is your !

libs/material/src/lib/delivery/pages/delivery-informations-editable/delivery-informations-editable.component.ts
```diff
@@ -12,6 +12,8 @@ import { createInformationsFormGroup } from '../../forms/informations.form';
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class DeliveryInformationsEditableComponent implements OnInit {
+  @HostBinding('attr.page-id') pageId = 'delivery-informations-editable';
+
```
ioaNikas: We should rename the entire component to `DeliveryInformationEditableComponent` (without 's'), and do the same for the selector, HostBinding etc.

apps/delivery/delivery-e2e/src/support/pages/NavbarPage.ts
Hugo-Hontans: For me it's not required but if the developer wants add it, he is free. It's just add a specification.

libs/material/src/lib/delivery/pages/delivery-informations-editable/delivery-informations-editable.component.ts
Hugo-Hontans: Yes but it's really not the subject of my PR but I can do it in next

apps/delivery/delivery-e2e/src/support/pages/NavbarPage.ts
Hugo-Hontans: No I can standardize it, you right

PR #977, opened 2019-09-23
Validators marketplace

Maxstgt:
exchange labels to slug
added validators


apps/catalog/marketplace/marketplace/src/app/movie/search/filter.util.ts
```diff
@@ -68,7 +68,7 @@ function certifications(movie: Movie, movieCertification: string[]): boolean {
     }
   }
 }
-
+// TODO #979
```
ioaNikas: You can add some text to summarize what has to be done (like issue title for example).

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.html
```diff
@@ -10,22 +10,35 @@ <h1 class="mat-display-3">Discover our selection</h1>
...
           <!-- Production Year -->
-          <mat-expansion-panel>
+          <mat-expansion-panel expanded>
             <mat-expansion-panel-header>
               <mat-panel-title test-id="production-year-panel">
                 Production Year
               </mat-panel-title>
             </mat-expansion-panel-header>
             <div fxLayout="row" fxLayoutAlign="start start" fxLayoutGap="10px" formGroupName="productionYear">
+              <!-- TODO##968 -->
```
ioaNikas: Same as above.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.html
```diff
@@ -10,22 +10,35 @@ <h1 class="mat-display-3">Discover our selection</h1>
...
+                <mat-error *ngIf="filterForm.get('productionYear').get('from').hasError('max')">
+                  The year can't be above the current year, {{ getCurrentYear }}
+                </mat-error>
               </mat-form-field>
               <mat-form-field>
                 <input test-id="production-year-input-to" matInput type="number" formControlName="to" placeholder="To"
-                  autocomplete="off" />
+                  autocomplete="off" [errorStateMatcher]="matcher" />
+                <mat-error *ngIf="filterForm.get('productionYear').get('to').hasError('pattern')">
+                  Please type in a valid year! Example: <i>{{ getCurrentYear }}</i>
```
ioaNikas: You got an exclamation mark here but not above. Stick to one style.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.html
```diff
@@ -127,9 +143,12 @@ <h6 class="mat-h3">You can choose multiple certifications</h6>
                 <input matInput formControlName="to" [matDatepicker]="pickerAvailableTo" placeholder="to" />
                 <mat-datepicker-toggle test-id="datepicker-to" matSuffix [for]="pickerAvailableTo">
                 </mat-datepicker-toggle>
-                <mat-datepicker #pickerAvailableTo startView="multi-year"></mat-datepicker>
+                <mat-datepicker #pickerAvailableTo startView="multi-year" startAt="2019"></mat-datepicker>
               </mat-form-field>
             </div>
+            <mat-error *ngIf="filterForm.get('availabilities').hasError('invalidRange')">
+              The value <i>To</i> can't be smaller then the value <i>From</i>!
```
ioaNikas: *than the value...

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -132,6 +132,42 @@ export class MarketplaceSearchComponent implements OnInit, OnDestroy {
...
+          this.filterForm.get('productionYear').hasError('invalidRange'))
+        ? true
+        : false;
+    } else {
+      return !this.filterForm
+        .get('availabilities')
+        .get('from')
+        .hasError('min') && this.filterForm.get('availabilities').hasError('invalidRange')
+        ? true
+        : false;
```
ioaNikas: Here you're saying something like : if this condition is false, return true, else return false.
You should use :
```
public showFormGroupError(formGroupName: string): boolean {
    if (formGroupName === 'productionYear') {
      return this.filterForm
        .get('productionYear')
        .get('from')
        .hasError('pattern') &&
        this.filterForm
          .get('productionYear')
          .get('from')
          .hasError('max') &&
        (this.filterForm
          .get('productionYear')
          .get('to')
          .hasError('pattern') &&
          this.filterForm.get('productionYear').hasError('invalidRange'));
    } else {
      return this.filterForm
        .get('availabilities')
        .get('from')
        .hasError('min') && !this.filterForm.get('availabilities').hasError('invalidRange');
    }
}
```

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -152,36 +188,62 @@ export class MarketplaceSearchComponent implements OnInit, OnDestroy {
...
+    /**
+     * We want to exchange the label for the slug,
+     * because for our backend we need to store the slug.
+     */
+    const languageSlug: LanguagesSlug = getCodeIfExists('LANGUAGES', language);
     if (LANGUAGES_LABEL.includes(language)) {
-      this.filterForm.addLanguage(language);
+      this.filterForm.addLanguage(languageSlug);
+    } else {
+      throw new Error('Something went wrong. Please choose a language from the drop down menu');
```
ioaNikas: menu **.**

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -152,36 +188,62 @@ export class MarketplaceSearchComponent implements OnInit, OnDestroy {
...
-    this.filterForm.removeLanguage(language);
-  }
-
-  public deleteLanguage(language: LanguagesLabel) {
-    this.filterForm.removeLanguage(language);
+    /**
+     * We want to exchange the label for the slug,
+     * because for our backend we need to store the slug.
+     */
+    const languageSlug: LanguagesSlug = getCodeIfExists('LANGUAGES', language);
```
ioaNikas: Name should be language**s**Slug as it is an instance of `LanguagesSlug`.
You can also just call it language as you in one of your function like `(language: LanguagesSlug)` :p

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -152,36 +188,62 @@ export class MarketplaceSearchComponent implements OnInit, OnDestroy {
...
   }
 
   public hasGenre(genre: GenresLabel) {
-    if (this.movieGenres.includes(genre) && !this.filterForm.get('type').value.includes(genre)) {
-      this.filterForm.addType(genre);
+    /**
+     * We want to exchange the label for the slug,
+     * because for our backend we need to store the slug.
+     */
+    const genreSlug: GenresSlug = getCodeIfExists('GENRES', genre);
```
ioaNikas: Same as above, but you maybe have a reason for this naming. Ignore those comments if so.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.form.ts
```diff
@@ -5,13 +5,20 @@ import {
   TerritoriesLabel,
   GenresLabel,
   GENRES_LABEL,
-  CERTIFICATIONS_LABEL,
-  MEDIAS_LABEL,
-  TERRITORIES_LABEL
+  LanguagesSlug,
+  GenresSlug,
+  GENRES_SLUG,
```
ioaNikas: You can maybe call it `GENRES` for easier reading.

libs/utils/src/lib/form/validators/validators.ts
```diff
@@ -63,18 +61,42 @@ export async function UniqueOrgName(control: AbstractControl): Promise<Validatio
...
-  return !LANGUAGES_SLUG.includes(control.value.trim().toLowerCase()) ? { languageNotSupported: true } : null;
+  return !LANGUAGES_SLUG.includes(control.value.trim().toLowerCase())
+    ? { languageNotSupported: true }
+    : null;
+}
+/**
+ * @description Form group validator that checks if the two children controls have a valid range
+ * @param firstControl Name of the first control
+ * @param secondControl name of the second control
+ */
```
ioaNikas: Be more explicit in what your params represent. Tell that firstControl is `to` and that secondControl is `from`.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
Maxstgt: The function `getCodeIfExists` only returns one string. So there is no need for plural. And the parameter is already called language / genre. So thats why ...Slug

apps/catalog/marketplace/marketplace/src/app/movie/search/search.form.ts
Maxstgt: No cause there is a difference between label and slug and you have to distinguish it easily

libs/utils/src/lib/form/validators/validators.ts
Maxstgt: But maybe the name is not to or from? I want to have a generic formGroup validator that just checks the range from two formControls...

ioaNikas: If this "range" have other application than a period of time (from => to) yeah it could be generic, but I can't see another use case. As you wish then !

PR #981, opened 2019-09-23
remove ethers & switch to InfuraProvider

pldespaigne:


apps/backend-functions/src/ipHash.ts
```diff
@@ -11,7 +11,7 @@ const abi = [
   'event Timestamp(bytes32 indexed scriptHash, address indexed owner)'
 ];
 
-const i = new ethers.utils.Interface(abi);
+const i = new Interface(abi);
```
GrandSchtroumpf: Use a real name

PR #1011, opened 2019-09-30
Issue#1009 model template material

Hugo-Hontans:
- [x] Create a material raw
- [x] Create an interface for material template
- [x] Convert correctly the type of material when save as template


libs/material/src/lib/delivery/pages/delivery-editable/delivery-editable.component.ts
```diff
@@ -23,6 +23,7 @@ import { FormElement } from '@blockframes/utils';
   styleUrls: ['./delivery-editable.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
+
```
lsenta: what?

lsenta: wtf

PR #1014, opened 2019-10-01
replace model MaterialTemplate in material-state

Hugo-Hontans:
Replace the model MaterialTemplate in Material State


libs/material/src/lib/material/+state/material.model.ts
```diff
@@ -64,3 +70,28 @@ export function getMaterialStep(material: Material, delivery: Delivery) {
...
+export function convertMaterialToMaterialTemplate(material: Partial<Material>): MaterialTemplate {
+  return {
+    id: material.id,
+    category: material.category,
+    value: material.value,
+    description: material.description,
+    price: material.price,
+    currency: material.currency
+  };
+}
```
ioaNikas: Maybe we can use a "one in all" function like :
```
export function createMaterialTemplate(material: Partial<Material>): MaterialTemplate {
  return {
    id: material.id,
    category: material.category || '',
    value: material.value || '',
    description: material.description || '',
    price: material.price || null,
    currency: material.currency || null
  };
}
```

=> if we get only id as a value, it creates a new MaterialTemplate with empty fields, if we pass a material, it will save only the need fields.

Hugo-Hontans: Good idea !

PR #1018, opened 2019-10-01
Global css

Mussuranae:
Review typo and more css


apps/catalog/marketplace/marketplace/src/app/distribution-right/selection-table/form-selection.component.html
```diff
@@ -23,43 +23,57 @@
           <!-- Movie name -->
           <ng-container matColumnDef="movieName">
               <th mat-header-cell *matHeaderCellDef mat-sort-header>Movie name</th>
-              <td mat-cell *matCellDef="let element"><b>{{element.movieName | titlecase}}</b></td>
+              <td mat-cell *matCellDef="let element">
```
Hugo-Hontans: You can directly add a css rules like this :
``td span { 
   font-size: 12px
 }``
without add spans

PR #1023, opened 2019-10-02
Global css

Mussuranae:
CSS Review of forms

### DIFF
--- a/libs/movie/src/lib/movie/form/festival-prizes/festival-prizes.component.html
+++ b/libs/movie/src/lib/movie/form/festival-prizes/festival-prizes.component.html
@@ -3,13 +3,16 @@
   <h3> Festival Prizes </h3>
   <div formArrayName="prizes" *ngFor="let credit of festivalprizes.prizes.controls; let i = index">
     <div [formGroupName]="i" class="full-width">
-      <mat-form-field class="small-margin">
+      <mat-form-field appearance="outline" class="small-margin">
+        <mat-label>Name</mat-label>
         <input matInput formControlName="name" placeholder="Festival name">
       </mat-form-field>
-      <mat-form-field class="small-margin">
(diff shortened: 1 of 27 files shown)


libs/movie/src/lib/movie/form/festival-prizes/festival-prizes.component.html
```diff
@@ -3,13 +3,16 @@
   <h3> Festival Prizes </h3>
   <div formArrayName="prizes" *ngFor="let credit of festivalprizes.prizes.controls; let i = index">
     <div [formGroupName]="i" class="full-width">
-      <mat-form-field class="small-margin">
+      <mat-form-field appearance="outline" class="small-margin">
```
Hugo-Hontans: Do we really need the ``class="small-margin`` ?

libs/movie/src/lib/movie/form/main/main.component.html
```diff
@@ -1,13 +1,15 @@
 <form page-id="movie-form-main" [formGroup]="main">
   <!-- INTERNAL REF -->
-  <mat-form-field class="full-width">
+  <mat-form-field appearance="outline" class="full-width">
```
Hugo-Hontans: Same, do we really need the ``class=full-width`` ?

Mussuranae: It is in the parent component, if we delete it, inputs are all very closed and little. We will refactor this form later I think.

libs/movie/src/lib/movie/form/festival-prizes/festival-prizes.component.html
Mussuranae: Yes for the form, if we delete, all inputs are stuck.

PR #1026, opened 2019-10-02
check selfdestructed wallet

### LINKED ISSUE #859: let user redeploy his smart-wallet
the smart-wallet wallet is automatically redeployed with a key, but it is same key as the previously destroyed smart-wallet, also the user doesn't have any clue that his wallet has been destroyed except for a deploying step during sending tx.

- [x] check if the smart-wallet exist each time the user go to the wallet page
- [x] if the smart-wallet has been destroyed delete all the user's key

Think of a way to prevent the user to re-import old keys

pldespaigne:
fix #859

### DIFF
--- a/libs/ethers/src/lib/key-manager/+state/key-manager.query.ts
+++ b/libs/ethers/src/lib/key-manager/+state/key-manager.query.ts
@@ -44,12 +44,13 @@ export class KeyManagerQuery extends QueryEntity<KeyState, Key> {
     ).toPromise();
   }
   getKeysOfUser(ensDomain: string) {
-    return this.getAll().filter(key => key.ensDomain === ensDomain);
+    // return this.getAll().filter(key => key.ensDomain === ensDomain);
+    return this.getAll({filterBy: key => key.ensDomain === ensDomain});
   }
   getMainKeyOfUser(ensDomain: string) {
     return this.getKeysOfUser(ensDomain).find(key => key.isMainKey);
   }
   getKeyCountOfUser(ensDomain: string) {
-    return this.getAll().filter(key => key.ensDomain === ensDomain).length;
+    return this.getKeysOfUser(ensDomain).length;
   }
 }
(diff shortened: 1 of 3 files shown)


libs/ethers/src/lib/key-manager/+state/key-manager.query.ts
```diff
@@ -43,8 +43,11 @@ export class KeyManagerQuery extends QueryEntity<KeyState, Key> {
       first()
     ).toPromise();
   }
+  getKeysOfUser(ensDomain: string) {
+    return this.getAll().filter(key => key.ensDomain === ensDomain);
+  }
   getMainKeyOfUser(ensDomain: string) {
-    return this.getAll().find(key => key.isMainKey && key.ensDomain === ensDomain);
+    return this.getKeysOfUser(ensDomain).find(key => key.isMainKey);
```
Maxstgt: why not using the akita filterBy property?

libs/ethers/src/lib/wallet/guards/wallet-active.guard.ts
```diff
@@ -22,7 +22,14 @@ export class WalletActiveGuard implements CanActivate, CanDeactivate<Wallet> {
     return new Promise((res, rej) => {
       this.subscription = this.authQuery.user$.pipe(
         filter(user => !!user),
-        tap(async user => await this.walletService.updateFromEmail(user.email))
+        tap(async user => {
+          await this.walletService.updateFromEmail(user.email);
+          try {
+            await this.walletService.checkWallet();
+          } catch (error) {
+            rej(error);
```
Maxstgt: no change request but what is `rej`?

pldespaigne: it's the reject function parameter of the promise, this code means "if the `checkWallet()` throw, reject the promise". This can later be handled by 
```
myPromise
  .then(() => everythingIsOk())
  .catch(() => handleMyError())
  .finally(() => doThisEvenInCaseOfError())
```

libs/ethers/src/lib/key-manager/+state/key-manager.query.ts
pldespaigne: didn't know it exists ^^ will update

PR #1030, opened 2019-10-03
Guard bigger boat

Mussuranae:
Refactor of the catalog basket guard

### DIFF
--- a/apps/catalog/marketplace/marketplace/src/app/guards/catalog-basket-list.guard.ts
+++ b/apps/catalog/marketplace/marketplace/src/app/guards/catalog-basket-list.guard.ts
@@ -11,7 +11,7 @@ export class CatalogBasketGuard extends CollectionGuard<BasketState> {
   }
 
   sync() {
-    // TODO # : change in syncCollection(path, queryFn)
+    // TODO #1027 : change in syncCollection(path, queryFn)
     return this.service.syncQuery();
   }
 }
--- a/apps/catalog/marketplace/marketplace/src/app/guards/catalog-basket-list.guard.ts
+++ b/apps/catalog/marketplace/marketplace/src/app/guards/catalog-basket-list.guard.ts
@@ -1,35 +1,17 @@
-import { CatalogBasket } from '@blockframes/marketplace';
-import { OrganizationQuery } from '@blockframes/organization';
-import { BasketStore } from '../distribution-right/+state/basket.store';
-import { Router } from '@angular/router';
+import { BasketState } from '../distribution-right/+state/basket.store';
 import { Injectable } from '@angular/core';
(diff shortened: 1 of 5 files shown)


apps/catalog/marketplace/marketplace/src/app/guards/catalog-basket-list.guard.ts
```diff
@@ -1,35 +1,17 @@
...
 
-  get query() {
-    return this.orgQuery.select('org').pipe(
-      switchMap(organization => {
-        const query = distributionRightQuery(organization.id);
-        return this.fireQuery.fromQuery<CatalogBasket[]>(query);
-      })
-    );
+  sync() {
+    // TODO #1027 : change in syncCollection(path, queryFn)
```
ioaNikas: It may be worth trying to `npm install akita-ng-fire` to update to the latest version and use it.
Here is the code we tested yesterday :
```
sync() {
  return this.service.syncCollection(`orgs/${organizationId}/baskets`, ref => ref.where('status', '==', 'pending'));
}
```

If it's still not working, just go back to `syncQuery()` and wait for the next update of akita-ng-fire.

apps/catalog/marketplace/marketplace/src/app/movie/selection/selection.component.ts
```diff
@@ -52,10 +55,7 @@ export class MarketplaceSelectionComponent implements OnInit {
...
 
   private getMovieTitle(id: string): string {
-    let movieLookup: string;
-    this.movieQuery
-      .getAll({ filterBy: movie => movie.id === id })
-      .forEach(movie => (movieLookup = movie.main.title.original));
+    const movieLookup = this.movieQuery.getEntity(id).main.title.original;
     if (!movieLookup) {
       throw new Error(`No movie found for this ${id} id`);
     }
```
ioaNikas: Ok we should rewrite this one for more consistency. What do you think about this ?
```
private getMovieTitle(id: string): string {
  const movie = this.movieQuery.getEntity(id);
  return movie
    ? movie.main.title.original
    : throw new Error(`No movie found for this id: ${id}`);
}
```

You can also use a classical `if` statement but as we got very simple return, I choose the ternary way.

apps/catalog/marketplace/marketplace/src/app/guards/catalog-basket-list.guard.ts
Mussuranae: Nice try, but we have to wait the next update :/

apps/catalog/marketplace/marketplace/src/app/movie/selection/selection.component.ts
Mussuranae: I use an `if` as VS Code doesn't want the ternary way :)

PR #1031, opened 2019-10-03
Issue#844 refactor copy materials 2

Hugo-Hontans:
Create 3 sub-functions to copy materials in movie and update a delivery.

### DIFF
--- a/libs/material/src/lib/delivery/+state/delivery.service.ts
+++ b/libs/material/src/lib/delivery/+state/delivery.service.ts
@@ -419,9 +419,19 @@ export class DeliveryService {
         return;
       }
 
-      this.materialService.upsertMaterial(material, sameValuesMaterial, delivery, tx);
-    });
+      // If there already is a material with same properties (but different id), we merge this
+      // material with existing one, and push the new deliveryId into deliveryIds.
+      if (!!sameValuesMaterial) {
+        this.materialService.updateMaterialDeliveryIds(sameValuesMaterial, delivery, tx);
+      }
 
+      // If values are not the same, this material is considered as new and we have to create
+      // and set a new material (with new Id).
+      if (!sameValuesMaterial) {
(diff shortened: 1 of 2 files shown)


libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -402,55 +402,24 @@ export class DeliveryService {
...
-
-    // TODO: issue#844, refactor this function and use updateMaterials() (material service)
+    const materials = await this.db.snapshot<Material[]>(`${document._type}/${document.id}/materials`);
+    const movieMaterials = await this.db.snapshot<Material[]>(`movies/${delivery.movieId}/materials`);
 
     materials.forEach(material => {
-      const sameValuesMaterial = movieMaterials.find(movieMaterial =>
-        this.materialService.isTheSame(movieMaterial, material)
-      );
+      const sameValuesMaterial = movieMaterials.find(movieMaterial => this.materialService.isTheSame(movieMaterial, material));
```
lsenta: Test multiline

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -65,48 +64,54 @@ export class MaterialService {
...
+      const target = createMaterial({
+        ...material,
+        id: this.db.createId()
+      });
+      return this.setNewMaterial(target, delivery, tx);
+    }
+  }
+
+  /** Checks if the material belongs to multiple delivery, if so: update the deliveryIds, otherwise just delete it. */
+  public checkMultipleDelivery(material: Material, delivery: Delivery, sameIdMaterial: Material, tx: firebase.firestore.Transaction) {
```
ioaNikas: You can also call this function `updateOrDeleteMaterial()` to fit some other already existing functions (upsertMaterial(), getOrCreateUserByMail() etc)

Issue #1043, opened 2019-10-07
DRY - Data Type improvements & transformation functions


Jackseed:
#Don'tRepeatYourself

## Objectives
- Refine data types
- Create a layer of functions above the database to transform the data either before using it (f functions) or before storing it (f-1 functions)
- Use the same functions for backend & frontend

## Workflow
1. Split Backend / Frontend Data Type (e.g. _DeliveryFirestore_, _DeliveryFront_)
+ Implement Transformation functions (back and forth)
2. Reuse in backend
+ Share operations over the data (e.g. `addStakeholder`)
3. Verify bundle size
4. Code quality (validate the codebase for use cases of a given type).

## Checklist
 - [ ] Delivery 
      - [ ] Split Data Types 
      - [ ] Implement Transformation functions
      - [ ] Reuse in backend
      - [ ] Share operations over the data 
      - [ ] Verify bundle size
      - [ ] Code quality 

 - [ ] Material 
      - [ ] Split Data Types 
      - [ ] Implement Transformation functions
      - [ ] Reuse in backend
      - [ ] Share operations over the data 
      - [ ] Verify bundle size
      - [ ] Code quality 

 - [ ] Template 
      - [ ] Split Data Types 
      - [ ] Implement Transformation functions
      - [ ] Reuse in backend
      - [ ] Share operations over the data 
      - [ ] Verify bundle size
      - [ ] Code quality 

 - [ ] Movie 
      - [ ] Split Data Types 
      - [ ] Implement Transformation functions
      - [ ] Reuse in backend
      - [ ] Share operations over the data 
      - [ ] Verify bundle size
      - [ ] Code quality 

 - [ ] Stakeholder 
      - [ ] Split Data Types 
      - [ ] Implement Transformation functions
      - [ ] Reuse in backend
      - [ ] Share operations over the data 
      - [ ] Verify bundle size
      - [ ] Code quality 

 - [ ] Invitation 
      - [ ] Split Data Types 
      - [ ] Implement Transformation functions
      - [ ] Reuse in backend
      - [ ] Share operations over the data 
      - [ ] Verify bundle size
      - [ ] Code quality 

 - [ ] Notification 
      - [ ] Split Data Types 
      - [ ] Implement Transformation functions
      - [ ] Reuse in backend
      - [ ] Share operations over the data 
      - [ ] Verify bundle size
      - [ ] Code quality 

 - [ ] Organization 
      - [ ] Split Data Types 
[...]

PR #1063, opened 2019-10-10
DRY : Invitation model

ioaNikas:
Added some comments.
Redefined Invitation model to use it in server and client side (contributes to #1043).

### DIFF
--- a/libs/notification/src/lib/invitation/+state/invitation.model.ts
+++ b/libs/notification/src/lib/invitation/+state/invitation.model.ts
@@ -1,61 +1,58 @@
 import { firestore } from 'firebase/app';
-import { PublicUser } from '@blockframes/auth';
 import { PublicOrganization } from '@blockframes/organization';
-type Timestamp = firestore.Timestamp;
+import { InvitationType, InvitationStatus, InvitationFromUserToOrganization, InvitationFromOrganizationToUser, InvitationToWorkOnDocument } from './invitation.firestore';
+import { PublicUser } from '@blockframes/auth/types';
 
-export interface Invitation {
-  id: string;
-  app: string;
-  type: InvitationType;
-  user?: PublicUser;
-  organization?: PublicOrganization;
-  docId?: string;
-  status: InvitationStatus;
-  date: Timestamp;
-}
 
-export const enum InvitationStatus {
-  accepted = 'accepted',
-  declined = 'declined',
-  pending = 'pending'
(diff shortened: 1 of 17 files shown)


libs/organization/src/lib/components/member-pending/member-pending.component.ts
```diff
@@ -9,8 +9,8 @@ import { Invitation } from '@blockframes/notification';
 })
 
 export class MemberPendingComponent {
-  @Output() accepted = new EventEmitter<Invitation>();
-  @Output() declined = new EventEmitter<Invitation>();
-  @Input() invitations: Invitation[];
+  @Output() accepted = new EventEmitter<InvitationFromUserToOrganization>();
```
lsenta: Are you sure about moving from the generic Invitation type to one specific kind of Invitation? There's no way this code gets another type of invitation?

libs/notification/src/lib/invitation/+state/invitation.model.ts
```diff
@@ -1,61 +1,58 @@
...
+    app: 'main',
+    type: InvitationType.fromOrganizationToUser,
     status: InvitationStatus.pending,
+    date: firestore.Timestamp.now(),
+    ...params
+  };
+}
+
+/** Factory function that create an Invitation of type toWorkOnDocument. */
+export function createInvitationToDocument(params: InvitationToWorkOnDocumentOptions): InvitationToWorkOnDocument {
```
lsenta: Would it make sense to have these create functions in the `firestore.ts` file and reuse them in the backend? (IF we create invites there)

lsenta:
Note "issue#1043" won't link the PR, keep spaces: "issue #1043". I updated the PR description to make sure we reference this.

libs/organization/src/lib/components/member-pending/member-pending.component.ts
ioaNikas: You are right, this object is going right to a function that takes a basic Invitation. Changing it.

libs/notification/src/lib/invitation/+state/invitation.model.ts
ioaNikas: Seems like all of our Invitations are generated client-side. Firebase function are only there to react to status change and do operations (like deleting invitation, writing permissions when we accept invitation to a document etc.)

PR #1064, opened 2019-10-10
create interface MaterialDocument

Hugo-Hontans:
- [x] Create interface MaterialDocument

contributes #1043

### DIFF
--- a/apps/catalog/marketplace/marketplace/src/app/movie/selection/selection.component.ts
+++ b/apps/catalog/marketplace/marketplace/src/app/movie/selection/selection.component.ts
@@ -47,8 +47,10 @@ export class MarketplaceSelectionComponent implements OnInit {
       movieName: this.getMovieTitle(detail.movieId),
       territory: detail.territories[0],
       rights: detail.medias[0],
+      // TODO: correct type issue#1066
       endRights: (detail.duration as any).to.toDate().toDateString(),
-      languages: detail.languages[0],
+      // TODO: correct type issue#1066
+      languages: detail.languages[0] as any
     } as MovieData;
   }
 
--- a/libs/material/src/lib/material/+state/material.model.ts
+++ b/libs/material/src/lib/material/+state/material.model.ts
@@ -1,13 +1,9 @@
-import { Delivery, CurrencyCode } from '../../delivery/+state';
-import { MaterialStatus, MaterialDocument, MaterialRaw } from './material.firestore';
+import { Delivery } from '../../delivery/+state';
+import { MaterialStatus, MaterialDocument, MaterialTemplateDocument } from './material.firestore';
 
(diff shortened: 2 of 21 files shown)


apps/catalog/marketplace/marketplace/src/app/movie/selection/selection.component.ts
```diff
@@ -48,7 +48,7 @@ export class MarketplaceSelectionComponent implements OnInit {
       territory: detail.territories[0],
       rights: detail.medias[0],
       endRights: (detail.duration as any).to.toDate().toDateString(),
-      languages: detail.languages[0],
+      languages: detail.languages[0] as any
```
lsenta: `as any`?

tsconfig.json
```diff
@@ -42,7 +42,11 @@
       "@blockframes/invitation/types": ["libs/notification/src/lib/invitation/+state/invitation.firestore.ts"],
 
       "@blockframes/account": ["libs/account/src/index.ts"],
+
+      // Material
       "@blockframes/material": ["libs/material/src/index.ts"],
+      "@blockframes/material/material/types": ["libs/material/src/lib/material/+state/material.firestore.ts"],
```
lsenta: Inconsistent, check the rest of the file

libs/material/src/lib/material/+state/material.model.ts
```diff
@@ -1,52 +1,22 @@
...
-export interface MaterialRaw {
-  id: string;
-  value: string;
-  description: string;
-  category: string;
-}
+export { MaterialStatus } from './material.firestore';
 
 /** Extends a Material Raw with fields that are specific to Material Template. */
 export interface MaterialTemplate extends MaterialRaw {
```
lsenta: Looks broken, it should extends a `SomethingDocument` from the types.

tsconfig.json
Hugo-Hontans: I will need to create ``@blockframes/materials/delivery/types and @blockframes/material/template/types``

libs/material/src/lib/material/+state/material.model.ts
Hugo-Hontans: I have to export MaterialStatus to the lib material without changes.

apps/catalog/marketplace/marketplace/src/app/movie/selection/selection.component.ts
Hugo-Hontans: Hum, I have to change the type of languages in MovieData, I don't really understand  why the type was a string

Hugo-Hontans: I created an issue to correct that: issue#1066

PR #1065, opened 2019-10-11
Issue#1043 delivery

Hugo-Hontans:
- [x] Create DeliveryDocument
- [x] Create StepDocument
- [x] Create function to convert StepDocument to Step

Contributes to #1043

### DIFF
--- a/apps/backend-functions/src/internals/pdf.ts
+++ b/apps/backend-functions/src/internals/pdf.ts
@@ -5,7 +5,7 @@
  * and in a firebase function.
  */
 import { groupBy, sortBy, isEmpty } from 'lodash';
-import { asIDMap, Delivery, IDMap, MaterialDocument, OrganizationDocument, Stakeholder, Step } from '../data/types';
+import { asIDMap, DeliveryDocument, IDMap, MaterialDocument, OrganizationDocument, Stakeholder, StepDocumentWithDate, convertStepDocumentToStepDocumentWithDate } from '../data/types';
 import { getCollection, getDocument } from '../data/internals';
 
 const PdfPrinter = require('pdfmake');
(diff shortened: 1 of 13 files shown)


apps/backend-functions/src/internals/pdf.ts
```diff
@@ -170,7 +170,7 @@ function rowMaterials(materials: MaterialDocument[], steps: { [id: string]: Step
  * @param materials
  * @param steps
  */
-function rowMaterialsPerCategory(materials: MaterialDocument[], steps: { [id: string]: Step }): any {
+function rowMaterialsPerCategory(materials: MaterialDocument[], steps: { [id: string]: StepDocumentWithDate }): any {
```
lsenta: I'd expect the backend to use `StepDocument` (with timestamp), are you sure we are a transformation to date?

libs/material/src/lib/delivery/+state/delivery.model.ts
```diff
@@ -1,71 +1,24 @@
...
 
-/** Extends a Raw Delivery with fields that are specific to the local data model. */
-export interface Delivery extends DeliveryRaw<Date> {}
-
 /** Syntaxic Sugar: the Delivery type in firestore. */
-export interface DeliveryDB extends DeliveryRaw<Timestamp> {}
-
-/** Syntaxic Sugar: the Delivery Step type used in the frontend. */
-export interface Step extends StepRaw<Date> {}
+export interface DeliveryWithTimestamps extends DeliveryDocument {
```
lsenta: Update the documentation to match what they represent. This one is particularly important since it's not very clear where it lives and what it does.

apps/backend-functions/src/internals/pdf.ts
Hugo-Hontans: We convert StepDocument to StepDocumentwithDate at line 301:
``const steps = asIDMap(convertStepDocumentToStepDocumentWithDate(delivery.steps));``

PR #1070, opened 2019-10-14
Issue#1043 stakeholder

ioaNikas:
Moved Stakeholder folder from Movie lib to Organization lib
Created StakeholderDocument model to be usable both in client side and server side ( issue #1043 ).


libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -150,7 +148,7 @@ export class DeliveryService {
       }
 
       // Create the stakeholder in the sub-collection
-      await this.shService.addStakeholder(delivery, organization.id, true, tx);
+      await this.shService.addStakeholder(delivery.id, organization.id, true, tx);
```
lsenta: You went too far!

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -442,7 +440,7 @@ export class DeliveryService {
   //////////////////////
 
   private makeDeliveryStakeholder(id: string, isAccepted: boolean) {
-    return createDeliveryStakeholder({ id, isAccepted });
+    return createStakeholder({ id, isAccepted });
```
lsenta: This method looks dumb, could we get rid of it altogether?

libs/material/src/lib/delivery/+state/delivery.service.ts
ioaNikas: I get the point, I ll keep it like this and notify @Mussuranae about the changes.

PR #1072, opened 2019-10-14
change movie data creater function

### LINKED ISSUE #1066: Correct types DistributionRight and MovieData to make sens in Basket Model
The function createRightDetail (MarketplaceSelectionComponent) doesn't work correctly.
This function takes an argument of type DistributionRight and convert it to MovieData. 
The returned object is not correct because it misses 2 fields and his type is forced to MovieData.
Care to the conversion with toDate() to detail.duration (type Date forced to any to be converted to Date).

Maxstgt:
close #1066


Maxstgt:
close #1066

PR #1073, opened 2019-10-14
Issue#1043 movie

Hugo-Hontans:
- [x] Create MovieDocument

Contributes to #1043

### DIFF
--- a/apps/backend-functions/src/data/types.ts
+++ b/apps/backend-functions/src/data/types.ts
@@ -2,6 +2,7 @@ import { PublicOrganization } from '@blockframes/organization/types';
 import { NotificationType } from '@blockframes/notification/types';
 import { PublicMovie } from '@blockframes/movie/types';
 
+export { MovieDocument } from '@blockframes/movie/types';
 export { OrganizationDocument, OrganizationStatus } from '@blockframes/organization/types';
 export {
   InvitationDocument,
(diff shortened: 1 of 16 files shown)


libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -0,0 +1,149 @@
+import { DateRangeRaw } from "@blockframes/utils/date-range";
+import { firestore } from "firebase";
```
ioaNikas: You can go deeper in the import with "firebase/app".

apps/backend-functions/src/data/types.ts
```diff
@@ -148,7 +141,7 @@ export interface Notification extends BaseNotification {
 }
 
 export interface SnapObject {
```
phpgeekfr: Can you add a comment to explain what this object is about?

libs/movie/src/lib/movie/guards/movie-organization-active.guard.ts
```diff
@@ -5,12 +5,6 @@ import { Router } from '@angular/router';
 
 export const movieOrganizationActiveQuery = (id: string): Query<Movie> => ({
   path: `movies/${id}`,
```
phpgeekfr: you removed
 ```  stakeholders: (movie: Movie) => ({
    path: `movies/${movie.id}/stakeholders`,
    organization: stakeholder => ({
      path: `orgs/${stakeholder.id}`
    })
  })
```

Is it intentional ? If yes, maybe you need to rename `movieOrganizationActiveQuery` ?

Hugo-Hontans: Sure I rename it

apps/backend-functions/src/data/types.ts
Hugo-Hontans: We will do a point to data/types.ts in other PR

PR #1081, opened 2019-10-15
Moved Delivery from libs to apps

ioaNikas:
Moved Delivery from libs to apps.
Created Home for Delivery (not using movie-list as a page anymore).


GrandSchtroumpf:
I don't really see the point of putting the content inside the apps folder. It just make the code less reusable. For me the only think we should put in the apps is the layout.

Issue #1061, opened 2019-10-09
 DistributionDeal - DistributionRight object


phpgeekfr:
* [x] Move distribution deal section to a sub-collection of movie for security reasons
* [x] check / factorize distributionRights and basket collections and add it to MCD
* [x] Check how to include re-buy requests / proposals functionality on DistributionDeal - DistributionRight object
* [ ] check #1061 in `code`
* [x] rename `sales` into `distributionDeals` or `avails` ?
* [x] clean wishlist stuff

PR #1082, opened 2019-10-15
Move distribution deal section to a sub-collection of movie for security reasons (part of  #1061)

phpgeekfr:
Move distribution deal section to a sub-collection of movie for security reasons.


libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -197,7 +196,7 @@ export function createMovieSale(params: Partial<MovieSale> = {}): MovieSale {
     dubbings: [],
     subtitles: [],
     ...params
-  } as MovieSale;
+  } as MovieSale; // @todo #1061 move to another file (distributionDeal.model.ts ?)
```
Maxstgt: why moving? the sales lives in the movie collection itself, right?

phpgeekfr: Yes, it is a subcollection but I was wondering if spliting movie logic and sales logic was a good idea or not.. I you think that keeping sales logic with movie logic, I'm fine with it ;)

apps/catalog/marketplace/marketplace/src/app/distribution-right/create/create.component.ts
```diff
@@ -180,16 +183,19 @@ export class DistributionRightCreateComponent implements OnInit, OnDestroy {
   //////////////////////
 
   public startResearch() {
+    console.log('todo :)');
```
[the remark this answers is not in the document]
phpgeekfr: Will add one thank

libs/movie/src/lib/movie/+state/movie.model.ts
Maxstgt: yeah maybe you are right, cause this function will only get called in the catalog marketplace.

phpgeekfr: I will let it like this right now, since this is going to have several updates, we will have the time to discuss that later ;)

PR #1083, opened 2019-10-16
Issue#1044 unit - implementing spectator

### LINKED ISSUE #1044: Unit testing framework
# Unit testing framework

## Objectives 
 - [x] Find a good unit testing framework
 - [ ] Implement it

## Description
Angular makes it difficult to create unit tests, we need a specific framework to do so.

Maxstgt:
Spectator is a lib for easy unit testing. It is working, but for the Collectionservice from akita ng fire, there are still some issues. 
Maybe we have to wait for akita ng fire 2.0
close #1044


apps/catalog/marketplace/marketplace/src/app/distribution-right/+state/basket.service.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { AkitaNgRouterStoreModule, RouterService } from '@datorama/akita-ng-router-store';
```
Maxstgt: this is the collectionservice which is kind of tricky to mock. I don't want to trash this code because I guess it is also kind of valuable for template reasons

apps/catalog/marketplace/marketplace/src/jest.config.js
```diff
@@ -0,0 +1,24 @@
+// base config from jest-present-angular
```
GrandSchtroumpf: Why do you have to create this file ?

firestore-debug.log
```diff
@@ -0,0 +1,7 @@
+API endpoint: http://localhost:8080
```
GrandSchtroumpf: Maybe you shouldn't push this file. @lsenta do you have a good practice about that ?

PR #1085, opened 2019-10-16
fix movie form

Hugo-Hontans:
- [x] Remove persistForm of akita
- [x] Fix initialization of movie display
- [x] Fix initialization of movie form

### DIFF
--- a/libs/movie/src/lib/movie/form/root/root.component.ts
+++ b/libs/movie/src/lib/movie/form/root/root.component.ts
@@ -1,8 +1,6 @@
-import { ChangeDetectionStrategy, Component, OnDestroy, OnInit, ViewEncapsulation } from '@angular/core';
-import { createMovie, MovieQuery, MovieService } from '../../+state';
+import { ChangeDetectionStrategy, Component, ViewEncapsulation, Input } from '@angular/core';
+import { MovieQuery, MovieService } from '../../+state';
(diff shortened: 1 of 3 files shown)


libs/movie/src/lib/movie/form/root/root.component.ts
```diff
@@ -11,6 +9,7 @@ import { formatCredit, formatCredits } from '@blockframes/utils/spreadsheet/form
 import { FormControl } from '@angular/forms';
 import { getCodeIfExists } from '../../static-model/staticModels';
 import { CreditFormControl } from '../main/main.form';
+import { Router } from '@angular/router';
 
 @Component({
   selector: 'movie-form-root',
```
phpgeekfr: add input into selector to force user to pass `form`

libs/movie/src/lib/movie/pages/movie-editable/movie-editable.component.ts
```diff
@@ -12,17 +14,20 @@ export class MovieEditableComponent implements OnInit {
   @HostBinding('attr.page-id') pageId = 'movie-editable';
   public fullScreen = false;
   public form$: Observable<Movie>;
+  public form: MovieForm;
```
phpgeekfr: maybe private is enough?

PR #1099, opened 2019-10-21
added dubbings and subtitle support

Maxstgt:
I couldn't add filtering for original language, cause this property doesn't exist on the movie sales interface. So I added support for dubbings and subtitles.

### DIFF
--- a/apps/catalog/marketplace/marketplace/src/app/distribution-right/create/create.component.ts
+++ b/apps/catalog/marketplace/marketplace/src/app/distribution-right/create/create.component.ts
@@ -26,9 +26,9 @@ import { DateRange } from '@blockframes/utils/date-range';
 import { ControlErrorStateMatcher, languageValidator } from '@blockframes/utils';
 import {
   getSalesInDateRange,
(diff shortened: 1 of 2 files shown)


apps/catalog/marketplace/marketplace/src/app/distribution-right/create/create.component.ts
```diff
@@ -189,11 +189,13 @@ export class DistributionRightCreateComponent implements OnInit, OnDestroy {
...
         tap(value => {
+
           //////////////////
           // FORM VALIDATION
           //////////////////
-          if (!(!!value.duration || value.medias.length || value.territories.length)) {
-            console.log('You have to provide value for your research');
+
+          if (!this.form.valid) {
+            console.log('Waiting for the form to be filled');
```
pldespaigne: is it intended to use `console.log` ? is there no other way ?

Maxstgt: there is, but it doesn't make sense to work on the UI if it changes in 2 weeks

Issue #1124, opened 2019-10-22
Update films import


Jackseed:
## Description
Movie view show some new content that we need to introduce to the movie import & excel. 

## Objectives
- [x] List the changes 
- [x] Update Data model
- [x] Implement changes in import 
- [x] Update excel file
- [x] Test it

Jackseed:
I updated the excel here tab AFM version: https://docs.google.com/spreadsheets/u/2/d/1V-cNhhY1AK71dqrl2fYGtkY5lD11PyM9Eb22DatvOns/edit#gid=877090665
Every new info are in green at the end. Some info were mandatory and we won't have it (internal ref or mandate terms), don't change it in the code, we will put random value. I would prefer to integrate those changes in the master rather than making it only in prod since we will need after.

Jackseed:
- [ ] Promo reel & trailer links will be vimeo/video links, we need to keep it in firebase
- [ ] Scenario will be a link to pdf, we will also keep it this way or store it on our bdd and show our link

Jackseed:
@phpgeekfr ok for me, you can merge and close it!

PR #1145, opened 2019-10-23
Update films import #1124

phpgeekfr:


libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -3,6 +3,37 @@ import { firestore } from "firebase/app";
 
 type Timestamp = firestore.Timestamp;
 
+export enum PromotionalElementTypes {
```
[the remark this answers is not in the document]
GrandSchtroumpf: Also, if it's only for typings issue, personnaly I'm more in favor of creating string literals as they are not part of the bundle size and still resolve the typing issue : 
```
type PromotionalElementTypes = 'trailer' | 'poster' | ...
```

libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -12,6 +43,8 @@ interface MovieSalesAgentDealRaw<D> {
   rights: DateRangeRaw<D>;
   territories: string[],
   medias: string[],
+  salesAgent?: Credit,
```
GrandSchtroumpf: Doesn't make so much sens to me why a salesAgent is a Credit...

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -44,6 +46,8 @@ export type MovieSalesAgentDeal = MovieSalesAgentDealDocumentWithDates;
 
 export type Movie = MovieDocumentWithDates;
 
+export type MovieBudget = MovieBudget;
```
GrandSchtroumpf: Hmmm, you shouldn't reexport it. Import directly from `movie.firestore` instead

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -204,6 +211,15 @@ export function createCredit(params: Partial<Credit> = {}): Credit {
...
     ...params
   } as Credit;
 }
+
+
+export function createMovieBudget(params: Partial<MovieBudget> = {}): MovieBudget {
+  return {
+    totalBudget: '',
+    ...params
+  } as MovieBudget;
```
GrandSchtroumpf: Remove `as MovieBudget` and give a default value to all required properties

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -204,6 +211,15 @@ export function createCredit(params: Partial<Credit> = {}): Credit {
     firstName: '',
     lastName: '',
     creditRole: '',
+    logo: '',
     ...params
   } as Credit;
```
GrandSchtroumpf: Remove `as Credit` and give a default value to all required properties

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -181,6 +187,7 @@ export function createMovieSalesAgentDeal(
     },
     territories: [],
     medias: [],
+    reservedTerritories: [],
     ...params
   } as MovieSalesAgentDeal;
```
GrandSchtroumpf: Remove `as MovieSalesAgentDeal` and give a default value to all required properties

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -458,6 +474,119 @@ export class ViewExtractedElementsComponent {
...
+
+          movie.promotionalElements.promotionalElements.push(promotionalElement);
+        } else {
+          importErrors.errors.push({
+            type: 'warning',
+            field: 'promotionalElements',
+            name: 'Promo reel link',
+            reason: 'Optional field is missing',
+            hint: 'Edit corresponding sheet field.'
+          } as SpreadsheetImportError);
```
GrandSchtroumpf: Why do you have to force the type ?

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -797,6 +926,46 @@ export class ViewExtractedElementsComponent {
       } as SpreadsheetImportError);
     }
 
+    if (movie.budget.totalBudget === undefined) {
```
GrandSchtroumpf: What if it's null ?

libs/movie/src/lib/movie/+state/movie.firestore.ts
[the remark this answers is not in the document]
GrandSchtroumpf: Open to discussion on that subject ^^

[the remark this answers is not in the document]
phpgeekfr: `Is it not a duplicate of the slugs you built?` Yes, moved thoses enums into `slug` `label` type

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
phpgeekfr: I think it can't be null because movie fields are initialized and `xlsx` lib returns undefined if a spreadsheet cell is empty.

libs/movie/src/lib/movie/+state/movie.firestore.ts
phpgeekfr: Yes the denomination is ambiguous but credit (like producer) and salesAgent have many attributes in common. I renamed Credit to something more general  and created aliases.

apps/backend-functions/src/delete.ts
```diff
@@ -48,6 +48,10 @@ export async function deleteFirestoreMovie(
     }
   });
 
+  // Delete sub-collections
```
GrandSchtroumpf: Looks like a reusable function. I would recommend to create a pure function for that.
```typescript
export function removeAllSubcollections(snapshot: ..., batch: ...) {
  ....
}
```

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -34,7 +35,9 @@ export type MovieVersionInfo = MovieVersionInfo;
 
 export type Prize = Prize;
 
-export type Credit = Credit;
+export type Credit = Person;
```
GrandSchtroumpf: smart

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
```diff
@@ -64,11 +64,11 @@ export class TableExtractedMoviesComponent implements OnInit {
...
 
-  async createSelectedMovies() : Promise<boolean> {
+  async createSelectedMovies(): Promise<boolean> {
     try {
       const creations = this.selection.selected.filter(importState => !importState.movie.id && !hasImportErrors(importState));
-      for(const movie of creations) {
-        this.processedMovies ++;
+      for (const movie of creations) {
+        this.processedMovies++;
         await this.addMovie(movie);
```
GrandSchtroumpf: Here I think that you are waiting to add the movie before going to the next. If this is not the behavior you're expecting I would recommend to use a `await Promise.all(creations.map(movie => { this.processingMovies++; return this.addMovie(movie)})`.

Better would be tu use the `add()` method from `akita-ng-fire`. It'll create a batch for all movies and manage that for you.

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
```diff
@@ -107,8 +107,8 @@ export class TableExtractedMoviesComponent implements OnInit {
   async updateSelectedMovies(): Promise<boolean> {
     try {
       const updates = this.selection.selected.filter(importState => importState.movie.id && !hasImportErrors(importState));
-      for(const importState of updates) {
-        this.processedMovies ++;
+      for (const importState of updates) {
```
GrandSchtroumpf: Same as above

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
phpgeekfr: At first I was using `Promise.all` to add movies but it was not working for more than 10 movies at the same time.

Your solution with add() sound cool. I'm adding a comment and an issue to remember to do this later

phpgeekfr: https://github.com/blockframes/blockframes/issues/1178

libs/movie/src/lib/movie/components/import/table-extracted-movies/table-extracted-movies.component.ts
phpgeekfr: https://github.com/blockframes/blockframes/issues/1178

Issue #1095, opened 2019-10-21
Update Create your org workflow


Jackseed:
## Description

Users will be able to create organizations and accounts, still with an acceptation required by Cascade8. The flow has to work and needs to be completed.

## Objectives
- [x] Join the community page (https://projects.invisionapp.com/d/main#/console/18708905/389334914/preview)
1. Add the subtitle
2. Change the illustration
3. Css magic
From this:
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/829eca35-4438-4d63-b967-6a8ce29b391a)
to this:
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/cb091a5b-f0ef-4518-88f4-76a7069a7492)

- [x] Find my org page (https://projects.invisionapp.com/d/main#/console/18708905/389334917/preview)
1. Change the buttons: "Let's go" becomes "Send a request"; "Back" becomes "Create my company" and leads to the Create my org page
2. Change the illustration
3. Change titles

Goes from this:
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/01b39ba0-0c7b-49ab-8d2f-1b6eb5904962)
To this:
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/81f318f6-ffab-440e-9766-6311b134869f)

- [x] Congratz page (from joining a company request)
1. Change texts & url > see issue #1057
2. Add a button Edit my profile that leads to the "My account" page
3. Every other links / buttons should send the user to the congratulations page // this will be resolved with issue #1136 

- [x] Request notification (https://projects.invisionapp.com/d/main#/console/18708905/389334918/preview)
Update the notification from:
![image](https://user-images.githubusercontent.com/14964443/67202260-787ec380-f408-11e9-8f18-c34d5f87eb20.png)

to this: 
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/8244bbae-5159-41c0-b0c2-83fb6b8ba194)

pldespaigne:
#1114

Jackseed:
Update of the issue:
- [ ] Block the user on the Join the community page while his request is pending (either for a Cascade8 acceptation when he creates an org or when it's for an org acceptation of the user)
- [ ] Add a cancel button to the pending request @Hugo-Hontans

Jackseed:
[...]

PR #1153, opened 2019-10-25
add feedback message to user request

Mussuranae:
Snackbar message to inform user their request has been sent.
Related to  #1095

### DIFF
--- a/libs/organization/src/lib/pages/organization-find/organization-find.component.ts
+++ b/libs/organization/src/lib/pages/organization-find/organization-find.component.ts
@@ -48,7 +48,7 @@ export class OrganizationFindComponent implements OnInit {
       try {
         await this.invitationService.sendInvitationToOrg(this.selected.objectID);
         this.snackBar.open('Request send !', 'close', { duration: 2000});
-        this.router.navigate(['layout/organization/home']);
+        return this.router.navigate(['layout/organization/home']);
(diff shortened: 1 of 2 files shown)


libs/organization/src/lib/pages/organization-find/organization-find.component.ts
```diff
@@ -47,7 +47,8 @@ export class OrganizationFindComponent implements OnInit {
     if (this.selected) {
       try {
         await this.invitationService.sendInvitationToOrg(this.selected.objectID);
-        await this.router.navigate(['layout/organization/congratulation']);
+        await this.snackBar.open('Request send !', 'close', { duration: 2000});
```
Hugo-Hontans: Don't await the snackBar, it's not a promise

libs/organization/src/lib/pages/organization-find/organization-find.component.ts
```diff
@@ -47,7 +47,8 @@ export class OrganizationFindComponent implements OnInit {
...
+        await this.router.navigate(['layout/organization/home']);
```
Hugo-Hontans: Return directly instead of await

Mussuranae: It was like that before I did some change, is it not problematic ?

Issue #287, opened 2019-04-30
Be able to select image format and to crop images (like meetup does for example)


phpgeekfr:

Mussuranae:
- [ ] User profile picture
- [ ] Organization picture

phpgeekfr:
Assigning this issue to @jenna1k . Have fun ;)

Jackseed:
Implement it on 3 pages:
       - [ ] Account creation page #1104 
       - [ ] Create your org page #1120 
       - [ ] My profile page #1100

Jackseed:
- [ ] Keep the default images for org logo & user picture while there is no image uploaded & saved by the user

Jackseed:
Done ! > closed

PR #1154, opened 2019-10-25
Issue#287 cropper org creation

jenna1k:
related to #287 

implemented cropper in oraganization creation page

### DIFF
--- a/libs/organization/src/lib/components/organization-create/organization-create.component.scss
+++ b/libs/organization/src/lib/components/organization-create/organization-create.component.scss
@@ -48,5 +48,4 @@ button {
 drop-cropper {
   display: block;
   width: 100%;
-  height: 80%;
 }
--- a/libs/organization/src/lib/components/organization-create/organization-create.component.scss
+++ b/libs/organization/src/lib/components/organization-create/organization-create.component.scss
@@ -44,3 +44,9 @@ button {
     width: 300px;
   }
 }
+
+drop-cropper {
+  display: block;
+  width: 100%;
+  height: 80%;
+}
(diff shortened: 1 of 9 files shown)


libs/organization/src/lib/components/organization-create/organization-create.component.scss
```diff
@@ -44,3 +44,9 @@ button {
     width: 300px;
   }
 }
+
+drop-cropper {
+  display: block;
+  width: 100%;
+  height: 80%;
```
GrandSchtroumpf: Hmm you shouldn't need to set the height && the width. I think you only want to set the width here.

libs/ui/src/lib/cropper/cropper/cropper.component.html
```diff
@@ -7,6 +7,9 @@ <h3>Drop something here</h3>
     <image-cropper
       [imageFile]="file"
       [maintainAspectRatio]="true"
+      [containWithinAspectRatio]="true"
+      [onlyScaleDown]="true"
+      [alignImage]="'left'"
```
GrandSchtroumpf: Remove that line

libs/ui/src/lib/cropper/cropper/cropper.component.html
```diff
@@ -7,6 +7,9 @@ <h3>Drop something here</h3>
     <image-cropper
       [imageFile]="file"
       [maintainAspectRatio]="true"
```
GrandSchtroumpf: You can just do : `<image-cropper ... maintainAspectRatio ...>`

libs/ui/src/lib/cropper/cropper/cropper.component.html
jenna1k: removed `[maintainAspectRatio]="true"` & `[alignImage]="'left'"`

PR #1165, opened 2019-10-28
no org id bug fix

### LINKED ISSUE #740: when we create an invitation, the userDoc doesn't exist directly, so the doc is not found
I have no context on that one, comes from a refactoring on https://github.com/blockframes/blockframes/pull/732#discussion_r312456999

pldespaigne:
fix #740


libs/organization/src/lib/no-organization.module.ts
```diff
@@ -19,12 +19,14 @@ import { MatInputModule } from '@angular/material/input';
 import { MatToolbarModule } from '@angular/material/toolbar';
 import { MatListModule } from '@angular/material/list';
 import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
+import { MatProgressBarModule } from '@angular/material';
```
Maxstgt: precise import pls

Issue #1146, opened 2019-10-23
Remove if(AFM_DISABLE) in master


Jackseed:
- [ ] State issue number in the code
- [ ] Search the current issues, for a similar issue.
- [ ] Put relevant labels (which App / Bug / Refactoring )
- [ ] move to icebox (if it's a low priority) or backlog (for future) or in progress
- [ ] Include relevant screenshots, if possible. Drag your screenshots directly into this text box. 
- [ ] State the expected behavior and the steps needed to reproduce the problem.
Will do it after AFM release

Jackseed:
no longer relevant

PR #1167, opened 2019-10-28
Going AFM

lsenta:
Disable the main app for AFM


apps/main/main/src/index.html
```diff
@@ -10,9 +10,11 @@
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons|Material+Icons+Outlined" rel="stylesheet">
   <link rel="manifest" href="manifest.json">
   <meta name="theme-color" content="#1976d2">
+  <!--- Redirect to Catalog app for AMF -->
```
GrandSchtroumpf: Is there a tag, issue or something that we can reference so we can "just" look for it in the search bar to update it after this sprint ?

lsenta: referencing issue #1146

PR #1173, opened 2019-10-29
With intercom

Mussuranae:
Intercom in Marketplace, discuss and get feedback from user directly in the app thanks to a chat


apps/catalog/marketplace/marketplace/src/app/app.module.ts
```diff
@@ -60,6 +61,10 @@ import { LayoutComponent } from './layout/layout.component';
     WalletModule,
     KeyManagerModule,
     NotificationWidgetModule,
+    IntercomModule.forRoot({
+      appId: "srwfltp4", // from Intercom config
```
lsenta: Use a env variable so we can change this value and disable it for dev mode.
Look at sentry, firebase, etc. implementations: relies on `env/env.ts` and `env/env.[prod|ci|staging|etc].ts`

apps/catalog/marketplace/marketplace/src/index.html
```diff
@@ -14,5 +14,13 @@
 <body class="dark-theme mat-app-background">
   <catalog-root></catalog-root>
   <noscript>Please enable JavaScript to continue using this application.</noscript>
+
+  <!-- Intercom Website Visitor -->
+  <script>
+    window.intercomSettings = {
+      app_id: "srwfltp4"
```
lsenta: Looks like you inject intercom twice, as an angular module AND as a raw javascript script.
Use only one, preferably the angular module.

Mussuranae: It was one for non logged user (the html) and the other one was for logged user, I changed it

env/env.prod.ts
```diff
@@ -64,3 +64,7 @@ export const relayer = {
 
 export const backupBucket = 'blockframes-backups';
 export const sendgridAPIKey = ''; // defined in functions.config, see backend-functions/environments
+
+// Intercom
+// ========
+export const intercomId = 'srwfltp4';
```
lsenta: :+1:

env/env.staging.ts
```diff
@@ -80,3 +80,6 @@ export const templateIds = {
   userRequestAccepted: 'd-d32b25a504874a708de6bfc50a1acba7',
   wishlistPending: 'd-e0cd8970d19346eea499a81f67f1a557',
 }
+// Intercom
+// ========
+export const intercomId = 'srwfltp4';
```
GrandSchtroumpf: Do you want to have the same on staging and prod ? As it gives analytics feedbacks it might be better to have it only on Prod ?

libs/utils/src/lib/intercom.module.ts
```diff
@@ -0,0 +1,46 @@
...
+  constructor(
+    query: AuthQuery,
+    public intercom: Intercom,
+    ) {
+      if (intercomId) {
+        query.user$.subscribe(user => {
+          if (!user) {
+            intercom.boot(getIntercomOptions(user))
+          }
+        })
```
GrandSchtroumpf: Let's try that : 
```typescript
constructor(intercom: Intercom, query: AuthQuery) {
  query.user$.pipe(
    distinctUntilChanged((old, new) => !!old !== !!new), // Trigger only when the user login/logout
  ).subscribe(user => {
    intercom.shutdown();  // Stop listening to the current session
    intercom.boot(getIntercomOptions(user));  // Start a new session
  })
}
```

env/env.staging.ts
Mussuranae: We can have the same because it doesn't send any information. You can use Intercom to send message to C8 team but Intercom doesn't get any information.

libs/utils/src/lib/intercom.module.ts
Mussuranae: It doesn't work :'(

Issue #1104, opened 2019-10-21
Account creation page (when invited to an org)


Jackseed:
## Description 
This page already exists but needs to be updated. 
Currently, the user receives an email with the email address to use and his password + a link. 

## Objective
This link should lead to this page:
https://projects.invisionapp.com/d/main#/console/18557321/385911992/preview
including:
- [x] email address being already written and disabled for modification
- [x] First name + last name
- [x] ask for the current password
- [x] ask for a new password
- [x] ask for confirmation of the new password

Also, 
- [x] if the user connects for the first time using the homepage (instead of the link included in the email), he should be redirected to the same page described above.

Jackseed:
Regarding the guard, @Hugo-Hontans worked on the same kind of guard when creating an org.
Also, there will be an image cropper implemented by @jenna1k , issue #287

PR #1179, opened 2019-10-30
Issue#287 implement cropper

jenna1k:
implement cropper in Account creation page
related to #1104 & #287

### DIFF
--- a/libs/auth/src/lib/pages/identity/identity.component.ts
+++ b/libs/auth/src/lib/pages/identity/identity.component.ts
@@ -25,7 +25,6 @@ export class IdentityComponent {
   ) {}
 
   public async update() {
-    console.log(this.form.value)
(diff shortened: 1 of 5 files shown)


libs/auth/src/lib/pages/identity/identity.component.ts
```diff
@@ -13,7 +13,8 @@ import { Router } from '@angular/router';
 export class IdentityComponent {
   public form = new FormGroup({
     name: new FormControl(''),
-    surname: new FormControl('')
+    surname: new FormControl(''),
+    avatar: new FormControl
```
Maxstgt: no parenthesis, no initial value?

jenna1k: added

Issue #1166, opened 2019-10-28
Footer 


Jackseed:
About Bigger Boat / Privacy tabs

PR #1182, opened 2019-10-30
Footer #1166

phpgeekfr:


libs/ui/src/lib/toolbar/left-menu/left-menu.component.html
```diff
@@ -1,7 +1,6 @@
 <div class="sidenav_container">
-  <div fxLayout="column" fxLayoutGap="10px" style="height: 100%;">
-    <mat-toolbar class="mat-elevation-z2 left-sidenav-header" fxLayout="row-reverse"
-      fxLayoutAlign="space-between center">
+  <div style="height: 100%;">
```
Hugo-Hontans: Plz do this in the scss file

styles/style.scss
```diff
@@ -101,4 +101,26 @@ main {
...
+
+.dark-theme {
+  footer {
+    background: mat-color($bf-dark-background, app-bar);
+    border-top: 1px solid mat-color($bf-dark-background, border);
+
+    a {
+      color: mat-color($bf-dark-background, text);
+    }
+  }
```
Hugo-Hontans: Do you see with Clélia to manage it ?
I'm not sure that it is well to change a tag color

libs/ui/src/lib/toolbar/footer/footer.component.scss
```diff
@@ -0,0 +1,29 @@
...
+      li {
+        margin-right: 10px;
+        margin-left: 10px;
+        font-weight: 500;
+
+        a, a:hover, a:visited {
+          color: inherit;
+          font-size: 12px;
+          line-height: 15px;
+          text-decoration: none;
```
Hugo-Hontans: Maybe we have to do this in global style

libs/material/src/lib/delivery/pages/delivery-add-choose-starter/delivery-add-choose-starter.component.html
```diff
@@ -1,9 +1,9 @@
-<main>
+<div>
```
Hugo-Hontans: Do you really need this div ?

libs/material/src/lib/delivery/pages/delivery-add-find-movie/delivery-add-find-movie.component.html
```diff
@@ -1,9 +1,9 @@
-<main>
+<div>
```
Hugo-Hontans: same

libs/ui/src/lib/toolbar/footer/footer.component.scss
phpgeekfr: I would say no because this is specific to footer

styles/style.scss
phpgeekfr: moved to a specific scss file

PR #1188, opened 2019-10-31
Issue#1110 search filters

ioaNikas:
- [x] Add status search filter
- [x] Add sales agents search filter
- [x] Commented unwanted code for AFM (related to issue #1146)
- [ ] Add the Budget filter
- [ ] Add the length filter
- [ ] Add the Festivals filter
- [ ] Add the Color / B&W filter


apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -94,7 +101,18 @@ export class MarketplaceSearchComponent implements OnInit {
   /* Arrays for showing the selected entities in the UI */
   public selectedMovieTerritories: string[] = [];
 
-  /* Flags for the chip input */
+   /* Flags for Sales Agents chip input*/
+   public selectableSAChip = true;
+   public removableSAChip = true;
+   public addOnBlur = true;
```
Maxstgt: I guess if you just do something like `<mat-chip selectableSAChip` it is automatically a boolean = true, so you don't need to specify a variable for that in the ts file which makes it slime.

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -105,23 +123,43 @@ export class MarketplaceSearchComponent implements OnInit {
...
+        } else {
+          return movies.filter(movie => {
+            return filterMovie(movie, filterOptions);
+          });
+        }
+      }),
       tap(movies => (this.availableMovies = movies.length))
     );
 
+    this.salesAgentsSub = this.movieQuery.selectAll().pipe(
```
Maxstgt: nice ;)

apps/catalog/marketplace/marketplace/src/app/movie/search/search.component.ts
```diff
@@ -278,4 +355,29 @@ export class MarketplaceSearchComponent implements OnInit {
...
+    this.salesAgentInput.nativeElement.value = '';
+  }
+
+  public removeSalesAgent(salesAgent: string) {
+    const index = this.selectedSalesAgents.indexOf(salesAgent);
+
+    if (index >= 0) {
+      this.selectedSalesAgents.splice(index, 1);
+    }
+    this.filterForm.removeSalesAgent(salesAgent);
```
Maxstgt: maybe move this also inside of the if statement

apps/catalog/marketplace/marketplace/src/app/movie/search/search.form.ts
```diff
@@ -13,7 +13,10 @@ import {
   MediasSlug,
   MEDIAS_SLUG,
   TerritoriesSlug,
-  TERRITORIES_SLUG
+  TERRITORIES_SLUG,
+  MovieStatusLabel,
+  MovieStatusSlug,
+  MOVIE_STATUS_SLUG
```
Maxstgt: ohm.... which one is for which use case?

ioaNikas: `MovieStatusLabel` is used for `status` type
`MovieStatusSlug` is used for status args type in `addStatus` and `removeStatus`
`MOVIE_STATUS_SLUG` is used to check if the slug as an argument exists in our list of status

Issue #1191, opened 2019-10-31
user can't save form in organization-edit & profile-edit page when image was deleted


jenna1k:
form can't save without uploading image with cropper
![image](https://user-images.githubusercontent.com/20438637/67953474-80e8b280-fbef-11e9-9cfb-c43278dfde2a.png)

PR #1195, opened 2019-11-01
fixed issue#1191

jenna1k:
fixed issue #1191


libs/ui/src/lib/cropper/cropper/cropper.component.ts
```diff
@@ -181,6 +180,7 @@ export class CropperComponent implements ControlValueAccessor{
 
     delete() {
       this.ref.delete().subscribe(() => {
```
Maxstgt: maybe check with Francois if you need to unsubscribe here when component is deactivated

PR #1209, opened 2019-11-04
fix updateWishlist, add label to production status

Hugo-Hontans:


apps/catalog/marketplace/marketplace/src/app/movie/wishlist/wishlist-current-repertory/wishlist-current-repertory.component.html
```diff
@@ -35,7 +35,7 @@ <h4 *ngIf="!isCurrent">Sent {{ date.toDate() | date: 'MMM dd yyyy' }}</h4>
   <!-- Production Status -->
   <ng-container matColumnDef="productionStatus">
     <th mat-header-cell *matHeaderCellDef mat-sort-header>Production Status</th>
-    <td mat-cell *matCellDef="let element">{{ element.main.status | titlecase }}</td>
+    <td mat-cell *matCellDef="let element">{{ getLabel('MOVIE_STATUS', element.main.status) | titlecase }}</td>
```
Maxstgt: there is pipe which we are using Yohann and me in the catalog marketplace. `translateSlug` of course you can the function here but it would be more consistent if we are using the pipe or the function. Also, you wouldn't need the title case pipe here and maybe in the future, if we want to translate the app, this will going to take care by the pipe. The pipe was originally created because Bruce created a ticket.

Hugo-Hontans: Yes sure, simpler with a pipe

PR #1232, opened 2019-11-05
Issue#1218 movie view

Hugo-Hontans:
- [x] Fix movie view
- [x] Fix css


libs/catalog/src/lib/components/promotional-elements/promotional-elements.component.scss
```diff
@@ -0,0 +1,9 @@
+a {
+  margin-top: 8px;
+  span {
```
phpgeekfr: you could have factorize span and b : `span, b {}`

styles/components/mat-button.scss
```diff
@@ -28,10 +28,11 @@ mat-button { // TODO: Get rid of the !important and apply this on mat-raised-but
 .dark-theme {
 
   .mat-stroked-button {
-    border-radius: 50px;
     box-shadow: none;
```
phpgeekfr: check if this is not messing up with other mat-stroked-button

Issue #616, opened 2019-07-10
Configuration required for production


lsenta:
- [x] We need the following api keys / service account for production:
  - [x] sentry (error reporting)
  - [x] sendgrid (mailing)
  - [x] algolia (full-text search)
  - [x] google analytics (via Firebase)
  - [x] intercom (app messaging)

Note: this ticket is likely to get update with more services (analytics, etc).

These should be managed by the cascade8 organisation with billing enabled!

lsenta:
@Jackseed Sending you this as a reminder for getting production ready.

Jackseed:
@lsenta I move this ticket to icebox so we keep it for next services to include in production

Jackseed:
- [ ] Yandex Metrica

GrandSchtroumpf:
- [ ] Firebase Performance
- [ ] Firebase Event / Audiences / ...

Maxstgt:
added performance for catalog marketplace ref #1243

Maxstgt:
added yandex metrika ref #1245

PR #1243, opened 2019-11-06
added perfomance module for analytics - catalog-marketplace

Maxstgt:
ref #616


apps/catalog/marketplace/marketplace/src/app/app.module.ts
```diff
@@ -62,11 +63,12 @@ import { LayoutComponent } from './layout/layout.component';
     WalletModule,
     KeyManagerModule,
     NotificationWidgetModule,
-
+
```
lsenta: remove trailing whitespace

PR #1245, opened 2019-11-06
added yandex metrika

Maxstgt:
Added Yandex Metrika 
ref #616


env/env.prod.ts
```diff
@@ -80,3 +80,6 @@ export const templateIds = {
   userRequestAccepted: 'd-d32b25a504874a708de6bfc50a1acba7',
   wishlistPending: 'd-e0cd8970d19346eea499a81f67f1a557',
 }
+
+// Yandex Metrika Id
+export const yandexId = 56105038;
```
lsenta: add empty value to other envs and check that it works,
this is required for other build configuration and for other developers.

PR #1246, opened 2019-11-06
remove email doublon

pldespaigne:
remove duplicate "Someone wants you to join his org" & user creds emails

### DIFF
--- a/libs/auth/src/lib/+state/auth.service.ts
+++ b/libs/auth/src/lib/+state/auth.service.ts
@@ -153,7 +153,6 @@ export class AuthService {
    * @email find the user with this email. If email doesn't match with an existing user,
    * create a user with this email address.
    */
-  // public async getOrCreateUserByMail(email: string): Promise<User> {
   public async getOrCreateUserByMail(email: string, orgName: string, invitationId?: string): Promise<User> {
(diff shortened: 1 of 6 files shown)


libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -154,9 +154,9 @@ export class AuthService {
    * create a user with this email address.
    */
   // public async getOrCreateUserByMail(email: string): Promise<User> {
```
GrandSchtroumpf: Remove this line if you don't need it

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -143,7 +143,7 @@ export class OrganizationService {
   public async addMember(member: OrganizationMemberRequest) {
     const orgId = this.query.id;
     // get a user or create a ghost user when needed:
-    const { uid } = await this.authService.getOrCreateUserByMail(member.email); // TODO: limit the number of requests per organizations!
+    const { uid } = await this.authService.getOrCreateUserByMail(member.email, this.query.getValue().org.name); // TODO: limit the number of requests per organizations!
```
GrandSchtroumpf: Create a variable `const { name } = this.query.getValue().org` above

PR #1249, opened 2019-11-07
Created reusable component for Terms and Conditions of use checked

ioaNikas:
Reusable component for Terms and Conditions of us checkbox.

### DIFF
--- a/libs/ui/src/lib/terms-conditions/terms-conditions.component.ts
+++ b/libs/ui/src/lib/terms-conditions/terms-conditions.component.ts
@@ -1,6 +1,6 @@
 import { ChangeDetectionStrategy, Component, Output, EventEmitter } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
-import { PrivacyPageComponent } from 'apps/catalog/marketplace/marketplace/src/app/pages/privacy-page/privacy-page.component';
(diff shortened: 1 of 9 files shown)


libs/ui/src/lib/terms-conditions/terms-conditions.component.ts
```diff
@@ -0,0 +1,27 @@
+import { ChangeDetectionStrategy, Component, Output, EventEmitter } from '@angular/core';
+import { MatDialog } from '@angular/material/dialog';
+import { PrivacyPageComponent } from 'apps/catalog/marketplace/marketplace/src/app/pages/privacy-page/privacy-page.component';
```
Hugo-Hontans: Don't hard code the import.
Maybe you have to declare `PrivacyPageComponent` directly in this module if we don't use it elsewhere

ioaNikas: As Terms and Conditions of use are closely related to an app (archipel content won't have the same policy than delivery for example), we need to import them directly from the apps pages.

Later, this component will import Terms and Conditions of each app and `openTermsOfUse()` will use a switch and an `@Input` from parent to load the right page in the dialog.

Issue #1258, opened 2019-11-07
All services / guards should use `akita-ng-fire`


Jackseed:
Guards
- [x] [**Yohann**] Catalog basket 
- [x] [**Yohann**] Catalog movie active (active movie guard is the same for all apps)
- [x] [**Hugo**] Auth
- [ ] [**Hugo**] Delivery
- [x] [**Yohann**] Material
- [x] [**Yohann**] Template
- [x] [**Yohann**] Movie
- [x] Invitation (using dynamic store)
- [x] Notification
- [x] [**Hugo**] Organization

Services
- [x] Basket
- [x] [**Hugo**] Auth
- [ ] [**Hugo**] Delivery
- [x] [**Yohann**] Material
- [x]  [**Yohann**] Template
- [x] Movie
- [x] Invitation
- [x] Notification
- [x] [**Hugo**] Organization
- [x] Stakeholder

Jackseed:
Related to #976

Jackseed:
Needs to be done steps by steps, @ioaNikas will improve this issue

Jackseed:
Updated with #1318

PR #1277, opened 2019-11-13
Issue#1258 guards and services

ioaNikas:
**Guards & Services**
- [x] New Active Movie Guard for all apps
- [x] Template List and Active Guards + service
- [x] Template Materials Guard, Movie Materials Guard and Delivery Materials Guard + service

**Fixes**
- [x] Remove unused guard in wishlist
- [x] Updated context menu
- [x] Fixed mat-table in material list (can't nest columns into ng-containers)
- [x] Notifications message on document delete events is now displayed

ref issue #1258


libs/material/src/lib/template/+state/template.service.ts
```diff
@@ -110,14 +122,15 @@ export class TemplateService {
...
+  // TODO: Find a new way to subscribe on organization templates => ISSUE #1276
   /** Subscribe on organization templates (outside the TemplateListGuard) and set the template store. */
   public subscribeOnTemplates() {
     return this.organizationQuery
       .select(state => state.org.templateIds)
       .pipe(
         switchMap(ids => {
           if (!ids || ids.length === 0) throw new Error('No template yet')
-          const queries = ids.map(id => this.db.fromQuery<Template>(templateQuery(id)))
+          const queries = ids.map(id => this.fireQuery.fromQuery<Template>(templateQuery(id)))
```
GrandSchtroumpf: Use `syncQuery`: https://github.com/dappsnation/akita-ng-fire/tree/v2.0.0#akita-array-with-subcollection

libs/movie/src/lib/movie/guards/movie-active.guard.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from "@angular/core";
+import { CollectionGuard, CollectionGuardConfig } from "akita-ng-fire";
+import { MovieState } from "../+state/movie.store";
+import { MovieService } from "../+state/movie.service";
+import { ActivatedRouteSnapshot } from "@angular/router";
+
+@Injectable({ providedIn: 'root' })
+@CollectionGuardConfig({ awaitSync: true })
```
GrandSchtroumpf: Same as above. Do you always want to apply the `awaitSync` all the time ?

libs/ui/src/lib/terms-conditions/terms-conditions.component.ts
```diff
@@ -16,7 +16,7 @@ export class TermsAndConditionsComponent {
 
   /** Opens a dialog with terms of use and privacy policy. */
   public openTermsOfUse() {
-    this.dialog.open(PrivacyPageComponent, { maxHeight: '100vh' })
+    this.dialog.open(PrivacyPageComponent, { maxHeight: '80vh' })
```
GrandSchtroumpf: remove this change. It's going to create a conflict for me.

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -1,26 +1,34 @@
...
+    super(store)
+  }
 
   //////////////////////////////
   // CRUD MATERIAL (DELIVERY) //
   //////////////////////////////
 
   /** Returns a material to be pushed in a formGroup. */
-  public add(): Material {
+  public addMaterial(): Material {
```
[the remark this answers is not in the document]
ioaNikas: Naming is wrong here. We create an empty material with an Id and return it so we can had it to the form. Changing function name accordingly.

libs/material/src/lib/material/+state/material.model.ts
```diff
@@ -49,3 +48,8 @@ export function createMaterialTemplate(material: Partial<MaterialTemplate>): Mat
     currency: material.currency || null
   };
 }
+
+/** Takes a MaterialDocument and returns it as a MaterialTemplate */
+export function transformMaterialTemplate(material: MaterialDocument): MaterialTemplate {
```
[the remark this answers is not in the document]
ioaNikas: Seems like I can use `createMaterialTemplate` as it take a partial material and returns a template material with only the required fields.

Got you point about type guard, will use it later for more safety.

libs/material/src/lib/material/guards/delivery-materials.guard.ts
```diff
@@ -1,40 +1,31 @@
...
+import { MaterialService } from '../+state/material.service';
+import { ActivatedRouteSnapshot } from '@angular/router';
+import { switchMap } from 'rxjs/operators';
+import { DeliveryQuery } from '../../delivery/+state/delivery.query';
 
 @Injectable({ providedIn: 'root' })
-export class DeliveryMaterialsGuard extends StateListGuard<Material> {
-  urlFallback = 'layout';
-
+@CollectionGuardConfig({ awaitSync: true })
```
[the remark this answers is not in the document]
ioaNikas: It seems that we unsubscribe when we leave the page with the guard. If we wait for the observable to complete, navigation is just a little bit longer. If we don't wait, navigation is faster but we get a blink as the materials appear.

Should we take advantage of `{ awaitSync: false }` + `this.store.loading` for better user experience ?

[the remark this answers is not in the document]
GrandSchtroumpf: That you should ask to the UX team ;). But you can also have more advance strategy. In [this example](https://netbasal.gitbook.io/akita/angular/firebase-integration/collection-guard#getter-parameters) it will wait if the store is no empty.

libs/movie/src/lib/movie/guards/movie-active.guard.ts
ioaNikas: Indeed, just had a discussion with our UX team and it seems that we want the fastest navigation. So we will never wait to sync data. Instead, we will just navigate and use store loading state to display a spinner until we get the data displayed.

libs/material/src/lib/delivery/components/delivery-material-list/delivery-material-list.component.ts
```diff
@@ -5,43 +5,55 @@ import {
...
+  public displayedColumns = this.setDisplayedColumns();
+
+  constructor(private deliveryQuery: DeliveryQuery) {}
+
+  ngOnInit() {
+    this.delivery$ = this.deliveryQuery.selectActive();
+  }
+
+  /* Define an array of columns to be displayed in the list depending on delivery settings **/
+  public setDisplayedColumns() {
```
Maxstgt: TSC detects the return type?

libs/material/src/lib/material/guards/delivery-materials.guard.ts
```diff
@@ -1,41 +1,32 @@
...
-export class DeliveryMaterialsGuard extends StateListGuard<Material> {
-  urlFallback = 'layout';
-
+export class DeliveryMaterialsGuard extends CollectionGuard<MaterialState> {
   constructor(
-    private fireQuery: FireQuery,
-    private deliveryQuery: DeliveryQuery,
-    store: MaterialStore,
-    router: Router
+    service: MaterialService,
```
Maxstgt: not private?

libs/material/src/lib/material/guards/movie-materials.guard.ts
```diff
@@ -1,33 +1,22 @@
...
+  }
 
-  constructor(
-    private fireQuery: FireQuery,
-    private movieQuery: MovieQuery,
-    store: MaterialStore,
-    router: Router
-  ) {
-    super(store, router);
+  get awaitSync() {
```
Maxstgt: the function returns bool but is named awaitSync? It doesn't seem fitting. also if this is async should there be an async in the header of the function so that the return waits for it?

libs/material/src/lib/material/guards/movie-materials.guard.ts
```diff
@@ -1,33 +1,22 @@
...
   }
 
-  get query() {
-    return this.movieQuery.selectActiveId().pipe(
-      switchMap(movieId => {
-        const query = movieMaterialsQuery(movieId);
-        return this.fireQuery.fromQuery<Material[]>(query);
-      })
-    );
+  sync(next: ActivatedRouteSnapshot) {
```
Maxstgt: async?

libs/material/src/lib/material/guards/template-materials.guard.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ActivatedRouteSnapshot } from '@angular/router';
+import { MaterialQuery } from '../+state/material.query';
+
+@Injectable({ providedIn: 'root' })
+export class TemplateMaterialsGuard extends CollectionGuard<MaterialState> {
+  constructor(service: MaterialService, private store: MaterialStore, private query: MaterialQuery) {
+    super(service);
+  }
+
+  get awaitSync() {
```
Maxstgt: maybe same as above

libs/material/src/lib/material/guards/template-materials.guard.ts
```diff
@@ -0,0 +1,22 @@
...
+    return this.query.getCount() === 0;
+  }
+
+  sync(next: ActivatedRouteSnapshot) {
```
Maxstgt: maybe same as above

libs/material/src/lib/template/+state/template.service.ts
```diff
@@ -1,65 +1,68 @@
...
-    await this.db.firestore.runTransaction(async (tx: firebase.firestore.Transaction) => {
-      const organizationSnap = await tx.get(organizationDoc.ref);
-      const templateIds = organizationSnap.data().templateIds || [];
+    // Add the template to the database
+    this.add(template);
 
-      // Create document permissions
-      await this.permissionsService.createDocAndPermissions(template, organization, tx);
+    // Create document permissions
+    // await this.permissionsService.createDocAndPermissions(template, organization, tx);
```
Maxstgt: remove?

libs/material/src/lib/delivery/components/delivery-material-list/delivery-material-list.component.ts
```diff
@@ -5,43 +5,55 @@ import {
...
   @Output() editing = new EventEmitter<string>();
   @Output() selectedMaterial = new EventEmitter<Material>();
   @Output() selectAllMaterials = new EventEmitter<boolean>();
 
   @ViewChild(MatSort, { static: true }) sort: MatSort;
 
   public dataSource: MatTableDataSource<Material>;
-
   public selection = new SelectionModel<Material>(true, []);
+  public delivery$: Observable<Delivery>
```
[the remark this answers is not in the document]
ioaNikas: > Event a semicolon is crucial, Master Laurent, 2019

libs/material/src/lib/material/guards/delivery-materials.guard.ts
ioaNikas: Not private indeed, as we only need native methods from the service.

libs/material/src/lib/material/guards/movie-materials.guard.ts
ioaNikas: It's like asking : Do you want me to wait for synchronization ? yes ? no ?

`awaitSync` is a property of CollectionGuardConfig.

If you set it to true, it won't navigate to the next page until he's synchronized with the data in the store. If you set it to false, it will navigate even if the data is not loaded yet.

Here I'm telling it to wait only if my store is empty. Like that, if I got a store that is already set (like, I've been there before but didn't unsubscribed yet), you will navigate instantly because there is already something in the store.. It's more dynamic.

libs/material/src/lib/material/guards/movie-materials.guard.ts
ioaNikas: `sync` is a native property of CollectionGuard. It subscribe on the document which `id` == `next.params.movieId` and set it active in the store.

PR #1285, opened 2019-11-15
Issue#1258 hooks

ioaNikas:
Using hooks in template and material services.

issue #1258

### DIFF
--- a/libs/material/src/lib/material/+state/material.service.ts
+++ b/libs/material/src/lib/material/+state/material.service.ts
@@ -1,12 +1,10 @@
 import { Injectable } from '@angular/core';
-import { Material, createMaterial, MaterialStatus, createMaterialTemplate, MaterialTemplate } from './material.model';
-import { DeliveryQuery } from '../../delivery/+state/delivery.query';
+import { CollectionConfig, CollectionService } from 'akita-ng-fire';
 import { Delivery } from '../../delivery/+state/delivery.model';
 import { TemplateQuery } from '../../template/+state/template.query';
-import { CollectionConfig, CollectionService } from 'akita-ng-fire';
-import { MaterialState, MaterialStore } from './material.store';
+import { createMaterial, createMaterialTemplate, Material, MaterialStatus, MaterialTemplate } from './material.model';
 import { MaterialQuery } from './material.query';
-import { MaterialDocument } from './material.firestore';
+import { MaterialState, MaterialStore } from './material.store';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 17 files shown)


libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -6,6 +6,7 @@ import { TemplateQuery } from '../../template/+state/template.query';
 import { CollectionConfig, CollectionService } from 'akita-ng-fire';
 import { MaterialState, MaterialStore } from './material.store';
 import { MaterialQuery } from './material.query';
+import { MaterialDocument } from './material.firestore';
```
Maxstgt: shouldn't you export a type from the model first?

libs/material/src/lib/template/pages/template-editable/template-editable.component.ts
```diff
@@ -1,12 +1,14 @@
...
+import { ChangeDetectionStrategy, Component, HostBinding, OnInit } from '@angular/core';
+import { MaterialControl, MaterialForm } from '../../forms/material.form';
+import { MaterialQuery } from '../../../material/+state/material.query';
+import { MaterialService } from '../../../material/+state/material.service';
+import { MaterialTemplate, createMaterialTemplate } from '../../../material/+state/material.model';
+import { filter, map, switchMap, tap } from 'rxjs/operators';
+import { FormEntity } from '@blockframes/utils/form/forms/entity.form';
+import { MatSnackBar } from '@angular/material/snack-bar';
 import { Observable } from 'rxjs';
+import { Template } from '../../+state/template.model';
```
Maxstgt: maybe you want to sort it that everybody can easily read whats being used.

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -1,8 +1,8 @@
 import { Injectable } from '@angular/core';
 import { Movie, createMovie, MovieSale } from './movie.model';
-import { PermissionsService, OrganizationQuery, Organization } from '@blockframes/organization';
+import { OrganizationQuery, Organization } from '@blockframes/organization';
 import { MovieStore, MovieState } from './movie.store';
-import { CollectionService, CollectionConfig } from 'akita-ng-fire';
+import { CollectionService, CollectionConfig, WriteOptions } from 'akita-ng-fire';
 import { switchMap } from 'rxjs/operators';
 import { AngularFirestoreDocument } from '@angular/fire/firestore/document/document';
 import { AngularFirestoreCollection } from '@angular/fire/firestore/collection/collection';
```
Maxstgt: maybe same as above

libs/material/src/lib/material/+state/material.service.ts
ioaNikas: Indeed !

PR #1299, opened 2019-11-20
Delivery routes

GrandSchtroumpf:
Update routes for deliveries


libs/material/src/lib/template/components/template-add/template-add.component.ts
```diff
@@ -15,13 +15,14 @@ export class TemplateAddComponent {
...
-    private router: Router
+    private router: Router,
+    private routes: ActivatedRoute
   ) {}
 
   public addTemplate(templateName: string) {
     const template = this.service.createTemplate(templateName);
     this.close();
-    this.router.navigate([`layout/o/templates/${template.id}`])
+    this.router.navigate([`../${template.id}`], { relativeTo: this.routes })
```
ioaNikas: Calling it `route` without `s` makes more sense.

PR #1305, opened 2019-11-21
Issue#1258 permissions stakeholder

ioaNikas:
- [x] Switch PermissionsService into CollectionService
- [x] Switch PermissionsGuard into CollectionGuard
- [x] Rewrite functions to use CollectionService CRUD
- [ ] Use hooks to create PermissionsDocument when a Document is created
- [x] Stakeholder folder moved into delivery lib


libs/organization/src/lib/permissions/+state/permissions.service.ts
```diff
@@ -40,31 +52,27 @@ export class PermissionsService {
...
+      }
+      // Case of role = 'admin': we remove the member id from superAdmins/members and insert it in admins
+      if (role === UserRole.admin) {
+        delete permissions.roles[uid];
+        permissions.roles[uid] = UserRole.admin;
+      }
+      // Case of role = 'member': we remove the member id from admins/superAdmins and insert it in members
+      if (role === UserRole.member) {
+        delete permissions.roles[uid];
+        permissions.roles[uid] = UserRole.member;
```
Hugo-Hontans: A switch case can be more appropriate

libs/organization/src/lib/permissions/+state/permissions.store.ts
```diff
@@ -1,27 +1,20 @@
...
+import { StoreConfig, EntityStore } from '@datorama/akita';
 import { Injectable } from '@angular/core';
 import { Permissions } from './permissions.model';
+import { CollectionState } from 'akita-ng-fire';
 
 
-export type PermissionsState = Permissions;
+export interface PermissionsState extends CollectionState<Permissions> {}
 
 const initialState: PermissionsState = {
```
Hugo-Hontans: You can import initialState form 'akita-ng-fire' directly

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -15,7 +15,9 @@ import { AngularFirestoreCollection, AngularFirestoreDocument } from '@angular/f
 import { WalletService } from 'libs/ethers/src/lib/wallet/+state';
 import { CreateTx } from '@blockframes/ethers';
 import { TxFeedback } from '@blockframes/ethers/types';
-import { StakeholderDocument, createStakeholder } from '@blockframes/organization/stakeholder/types';
+import { StakeholderService } from '../stakeholder/+state/stakeholder.service';
+import { Stakeholder } from '../stakeholder/+state/stakeholder.model';
+import { StakeholderDocument } from '../stakeholder/+state/stakeholder.firestore';
```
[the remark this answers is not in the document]
ioaNikas: Indeed, but this one refers to a real StakeholderDocument as you can see in the service. It will disappear when we will use 100% of CollectionService CRUD operations.

libs/organization/src/lib/permissions/+state/permissions.store.ts
ioaNikas: As we saw together, initialState is exclusive to Auth.

PR #1312, opened 2019-11-22
With vincents users

lsenta:
- [x] ops: add vincents users for staging tests & golden data,
- [x] ops: add a simple `syncUsers` cmd for manual setup.

### DIFF
--- a/apps/backend-ops/src/main.ts
+++ b/apps/backend-ops/src/main.ts
@@ -1,14 +1,15 @@
-import { prepareFirebase, restore, syncUsers } from './firebaseSetup';
+import { restore, syncUsers } from './firebaseSetup';
 import { MIGRATIONS } from './firestoreMigrations';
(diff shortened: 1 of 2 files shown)


apps/backend-ops/src/main.ts
```diff
@@ -61,4 +61,6 @@ if (cmd === 'prepareForTesting') {
   prepareToronto();
 } else if (cmd === 'upgradeAlgoliaOrgs') {
   upgradeAlgoliaOrgs();
+} else if (cmd === 'syncUsers') {
+  syncUsers(USERS);
```
Mussuranae: Is the variable USERS from the file users.toronto.fixture or from users.fixture ?

lsenta: Good catch, the other variable is named `USERS_TORONTO` now

PR #1333, opened 2019-11-26
added movie update,create and delete hooks

### LINKED ISSUE #1330: Hooks to synchronize Algolia with Firestore movies
Hooks to synchronize Algolia with Firestore movies

Maxstgt:
close #1330 
movies are now synchronized with the algolia index

### DIFF
--- a/apps/backend-functions/src/internals/algolia.ts
+++ b/apps/backend-functions/src/internals/algolia.ts
@@ -58,3 +58,12 @@ export function storeSearchableMovie(
     movie: pick(movie, ALGOLIA_FIELDS)
   });
 }
+
+export function deleteSearchableMovie(movieId: string): Promise<any> {
+  if (!algolia.adminKey) {
+    console.warn('No algolia id set, assuming dev config: skipping');
+    return Promise.resolve(true);
+  }
+
+  return indexMoviesBuilder().deleteObject(movieId);
+}
--- a/apps/backend-functions/src/main.ts
+++ b/apps/backend-functions/src/main.ts
@@ -1,5 +1,4 @@
 // import * as gcs from '@google-cloud/storage';
-import { hashToFirestore } from './generateHash';
 import { onIpHash } from './ipHash';
 import { onDeliveryUpdate } from './delivery';
 import { functions } from './internals/firebase';
@@ -43,18 +42,7 @@ export const onIpHashEvent = functions.pubsub.topic('eth-events.ipHash').onPubli
  *
  * NOTE: we will need to refactor the function to dispatch
  * depending on the file path. Or use different storage buckets
- * (one per app maybe).
- */
-export const generateHash = functions.storage
-  .object()
-  .onFinalize(hashToFirestore);
-
-/**
- * Trigger: when user creates an account.
- *
- * We create a corresponding document in `users/userID`.
- */
-export const onUserCreate = functions.auth
+ * (one per app maybe).onOrganizationUpdateEventfunctions.auth
   .user()
   .onCreate(logErrors(users.onUserCreate));
 
@@ -100,7 +88,7 @@ export const admin = functions.https
 /** Trigger: when signature (`orgId`) is added to or removed from `validated[]`. */
 export const onDeliveryUpdateEvent = onDocumentUpdate('deliveries/{deliveryID}', onDeliveryUpdate);
 
-/** Trigger: when a stakeholder is added to a delivery. */
+/** Trigger: when a stakeholonOrganizationUpdateEventder is added to a delivery. */
 export const onDeliveryStakeholderCreateEvent = onDocumentCreate(
   'deliveries/{deliveryID}/stakeholders/{stakeholerID}',
   onDeliveryStakeholderCreate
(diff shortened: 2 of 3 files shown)


apps/backend-functions/src/movies.ts
```diff
@@ -0,0 +1,49 @@
...
+  const after = change.after.data() as MovieDocument;
+
+  if (before === after) {
+    console.error('No changes detected');
+    throw new Error('movie update function got invalid movie data');
+  }
+
+  if (before.id === after.id) {
+    throw new Error('movie id can not be changed!');
+  }
```
ioaNikas: You'll always throw here. You meant `before.id !== after.id` ?

apps/backend-functions/src/internals/algolia.ts
```diff
@@ -42,3 +42,12 @@ export function storeSearchableMovie(
...
   return indexMoviesBuilder(adminKey).saveObject({ objectID: movie.id, movie });
 }
+
+export function deleteSearchableMovie(movieId: string): Promise<any> {
+  if (!algolia.adminKey) {
+    console.warn('No algolia id set, assuming dev config: skipping');
+    return Promise.resolve(true);
+  }
+
+  return indexMoviesBuilder().deleteObject(movieId);
```
lsenta: Why is this line not using `adminKey` like line 43?

apps/backend-functions/src/movies.ts
```diff
@@ -0,0 +1,49 @@
...
+): Promise<any> {
+  const movie = snap.data();
+  const movieId = context.params.movieId;
+
+  if (!movie || !movieId) {
+    console.error('Invalid movie data:', movie);
+    throw new Error('movie update function got invalid movie data');
+  }
+
+  return Promise.all([
```
lsenta: don't `Promise.all` just return the promise.

apps/backend-functions/src/movies.ts
```diff
@@ -0,0 +1,49 @@
...
+    // Update algolia's index
+    storeSearchableMovie(movie, process.env['ALGOLIA_API_KEY'])
```
lsenta: Can you see me a run of this code? As far as I know `process.env` does not work in google fuctions that's why we have functions.config.

Remove this, and look at `onOrganizationCreate`.

apps/backend-functions/src/movies.ts
```diff
@@ -0,0 +1,49 @@
...
+  ]);
+}
+
+export function onMovieDelete(
+  // First param is needed? @Laurent
```
lsenta: Don't `@people` in comments, we have tools for that.

lsenta: https://firebase.google.com/docs/reference/functions/providers_firestore_.documentbuilder.html#on-delete

keep the argument.

apps/backend-functions/src/movies.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export function onMovieUpdate(
+  change: functions.Change<FirebaseFirestore.DocumentSnapshot>,
+  context: functions.EventContext
+): Promise<any> {
+  const before = change.before.data() as MovieDocument;
+  const after = change.after.data() as MovieDocument;
+
+  if (before === after) {
```
lsenta: This will NEVER happen, the only way for 2 objects to be equal is to point to the same reference.

I made you a jsfiddle:
https://jsfiddle.net/8gLvhy9q/2/

You're looking for Structural Equality: https://basarat.gitbooks.io/typescript/docs/javascript/equality.html

What you should do is compare the before and after fields that you send to algolia, if they are equal, skip the update. Else, send the update.

apps/backend-functions/src/main.ts
```diff
@@ -96,7 +85,7 @@ export const admin = functions.https
 /** Trigger: when signature (`orgId`) is added to or removed from `validated[]`. */
 export const onDeliveryUpdateEvent = onDocumentUpdate('deliveries/{deliveryID}', onDeliveryUpdate);
 
-/** Trigger: when a stakeholder is added to a delivery. */
+/** Trigger: when a stakeholonOrganizationUpdateEventder is added to a delivery. */
```
lsenta: Undo this change

apps/backend-functions/src/main.ts
```diff
@@ -42,18 +42,7 @@ export const onIpHashEvent = functions.pubsub.topic('eth-events.ipHash').onPubli
  *
  * NOTE: we will need to refactor the function to dispatch
  * depending on the file path. Or use different storage buckets
- * (one per app maybe).
- */
-export const generateHash = functions.storage
```
lsenta: that's out of scope for this PR, undo this change.

apps/backend-functions/src/main.ts
```diff
@@ -42,18 +42,7 @@ export const onIpHashEvent = functions.pubsub.topic('eth-events.ipHash').onPubli
...
-  .onFinalize(hashToFirestore);
-
-/**
- * Trigger: when user creates an account.
- *
- * We create a corresponding document in `users/userID`.
- */
-export const onUserCreate = functions.auth
+ * (one per app maybe).onOrganizationUpdateEventfunctions.auth
   .user()
```
lsenta: Looks like a mistake, you commented out a lot of code. Do not keep dead code, remove it.

lsenta:
You merge'd instead of rebasing, that's bad practice.

Have you tested all the code in there?

Maxstgt:
yes, it is tested

PR #1311, opened 2019-11-22
Image display

jenna1k:
relate to #1325

PR #1335, opened 2019-11-26
img reference instead of src

phpgeekfr:


phpgeekfr:
duplicate of https://github.com/blockframes/blockframes/pull/1311

PR #1336, opened 2019-11-26
renamed movieSale into DistributionRight

phpgeekfr:
part of #1061


apps/catalog/marketplace/marketplace/src/app/movie/search/filter.util.ts
```diff
@@ -92,23 +92,23 @@ function certifications(movie: Movie, movieCertification: string[]): boolean {
...
   if (!range || !(range.from && range.to)) {
     return true;
   }
   return (
-    deals.some(sale => {
-      if (sale.rights) {
-        const from: Date = (sale.rights.from as any).toDate();
+    deals.some(deal => {
+      if (deal.rights) {
+        const from: Date = (deal.rights.from as any).toDate();
```
Hugo-Hontans: If you have to force the type of the Date with `as any` to do a `toDate()`,  I think we have a problem with the model

apps/catalog/marketplace/marketplace/src/app/movie/search/filter.util.ts
```diff
@@ -92,23 +92,23 @@ function certifications(movie: Movie, movieCertification: string[]): boolean {
...
          */
         return from.getTime() < range.from.getTime();
       }
     }) &&
-    deals.some(sale => {
-      if (sale.rights) {
-        const to: Date = (sale.rights.to as any).toDate();
+    deals.some(deal => {
+      if (deal.rights) {
+        const to: Date = (deal.rights.to as any).toDate();
```
Hugo-Hontans: same

PR #1340, opened 2019-11-27
User can choose a Theme

GrandSchtroumpf:
Now theme is stored on localStorage.
Each application can specify a default theme. 
User can chose their theme.

### DIFF
--- a/apps/main/main/src/index.html
+++ b/apps/main/main/src/index.html
@@ -11,9 +11,10 @@
   <link rel="manifest" href="manifest.json">
   <meta name="theme-color" content="#1976d2">
   <!-- issue#1146: Redirect to Catalog app for AMF -->
-  <meta http-equiv="Refresh" content="2; url=/catalog-marketplace">
+  <!-- <meta http-equiv="Refresh" content="2; url=/catalog-marketplace"> -->
 </head>
 <body class="light-theme mat-app-background">
+    <main-root></main-root>
   <!-- issue#1146, disabled for AMF: <main-root></main-root> -->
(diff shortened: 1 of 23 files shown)


apps/main/main/src/index.html
```diff
@@ -11,9 +11,10 @@
   <link rel="manifest" href="manifest.json">
   <meta name="theme-color" content="#1976d2">
   <!-- issue#1146: Redirect to Catalog app for AMF -->
-  <meta http-equiv="Refresh" content="2; url=/catalog-marketplace">
+  <!-- <meta http-equiv="Refresh" content="2; url=/catalog-marketplace"> -->
```
pldespaigne: Why do you keep a commented line ?

libs/ui/src/lib/theme/theme-widget/theme-widget.module.ts
```diff
@@ -0,0 +1,19 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { ThemeWidgetComponent } from './theme-widget.component';
+
+// Material
+import { MatIconModule } from '@angular/material/icon';
+import { MatMenuModule } from '@angular/material/menu';
+
+export const material = [MatIconModule, MatMenuModule]
```
pldespaigne: What are the benefits of this instead of putting them directly in the import statement ?

apps/main/main/src/index.html
GrandSchtroumpf: This is the Redirect for AMF, I wasn't sure if I should let it or not... So I just commented

PR #1342, opened 2019-11-28
Issue#1266 firestore rules

### LINKED ISSUE #697: Finalize permissions for the release

### LINKED ISSUE #1266: Implement firestore rules for the prod version
- [x] Re-enable roles in organization views
- [x] Implement new role : member
- [ ] Implement basic rules to securize the prod database (on users, orgs, permissions and movies collections)

### LINKED ISSUE #1343: Create a migration script for organization permissions
Need to transform arrays (superAdmins, admins, members) into a mapping of this sort :

roles : { userId: "role", userId: "role", userId: "role"... }

ioaNikas:
- [x] New rules to fit and securize the production version of the database.
- [x] No permissions loaded in store until user has an orgId.
- [x] Rename interfaces and functions in permissions state folder.
- [x] Create migration script.

close #697 
close #1266
close #1343


firestore.rules
```diff
@@ -1,38 +1,59 @@
...
-    }
 
     // Note: Add a verification of each document ID => {id: ID_SHOULD_BE_EQUAL_TO_PATH}
 
     /// USERS RULES ///
 
     match /users/{userId} {
-    	allow read, update: if isSignedIn(); // TODO: Secure it as soon as we got Inivitations collection
+    	allow read: if isSignedIn() && userId() == existingData().uid;
+    	allow update: if isSignedIn() && userId() == incomingData().uid;
```
lsenta: Check that existingData is also the same, else  `userX` can write to `/users/userY` with `{uid: userX}`

lsenta: You should check the `userId` from `/users/{userId}` too

firestore.rules
```diff
@@ -1,38 +1,59 @@
...
     	allow create: if userId() == incomingData().uid;
+      allow delete: if userId() == existingData().uid;
     }
 
     /// NOTIFICATIONS RULES ///
 
     match /notifications/{notificationId} {
-    	allow read: if currentUser().uid == existingData().userId;
+    	allow read, update: if userId() == existingData().userId
+      	|| userOrgId().orgId == existingData().organization.id;
```
lsenta: `userOrgId().orgId` looks broken, it looks like `user -> orgId -> orgId`

firestore.rules
```diff
@@ -1,38 +1,59 @@
...
 
     match /orgs/{orgId} {
-        // TODO(issue#697): allow create only if state is pending.
-        // TODO(issue#697): allow update only if state is not changed (admin will change it).
-    	allow create: if isSignedIn();
-        allow read, update, delete: if isOrgAdmin(userId(), orgId);
+    	allow create: if isSignedIn() && incomingData().status == 'pending';
+      allow update: if isSignedIn()
+        && isOrgAdmin(userId(), orgId)
+        && incomingData().status == existingData().status;
```
lsenta: tbh I think you should validate all the ids from all incoming data or you might have security issues, if I write `/orgs/org12` with `{id: orgPixar}` I might be able to trick the system into believing I'm part of Pixar down the line.

lsenta: update should prevent changing ids too.

firestore.rules
```diff
@@ -42,56 +63,21 @@ service cloud.firestore {
...
+      allow read: if userHasValidOrg();
+      allow create: if userHasValidOrg();
+      allow update: if getDocumentPermissions(userOrgId(), movieId).ownerId == userOrgId()
+      	&& isOrgAdmin(userId(), userOrgId());
+      allow delete: if getDocumentPermissions(userOrgId(), movieId).ownerId == userOrgId()
+      	&& isSuperAdmin(userId(), userOrgId());
 
       match /{documents=**} {
-        allow read, write;
+      	allow read, write: if userHasValidOrg()
```
lsenta: I can create an org and change the sub collection of every movie on the platform

lsenta:
Nice, some issues, please review & fix what you can.
Open a ticket to list the "weaknesses" you won't fix with this patch (better merge early & improve than hold this PR for months again)

lsenta:
@Mussuranae could you review the migration?

firestore.rules
ioaNikas: Wow, indeed it's a typo ! Thanks

firestore.rules
ioaNikas: I don't get how you could trick the system this way ?
Even if you are free to put the `orgId` you want and event change your `user.orgId`, condition `isOrgAdmin(userId(), orgId)` will always return false as you won't appear as an admin of this org, no ?

I hope I'm not missing the point here

firestore.rules
ioaNikas: Applying exact same conditions as in movies collection to fit the logic.

lsenta:
LGTM, do you need the users' public profile?

Issue #700, opened 2019-07-29
Secure the admin pages


lsenta:
We have a few admin pages in the firebase backend, served at `/admin`.
They are used in #573 and #571 

We should secure these pages, at least with a simble password check for admins, or IAM checks, and at best using firebase auth and adding `cascade8` admin to some users.

Jackseed:
linked to #492

Jackseed:
Really high priority

lsenta:
@Jackseed 
(update after moving on issue #492)

- I can either implement the full firebase auth machinery so that users login with an account like on the main app to be able to trigger a backup / restore and others authentication operations.
   - time to implement: ~3 days, a bunch of decision to make regarding implementing admin as a fully fledged angular app, taking front dev away from the app.
   - works like any other blockframes app.
- Or I can implement a single admin password that we store in the secrets, all the users will copy and paste this password in the admin forms.
  - time to implement: 20min
  - lower security
  - you have to copy and paste this password to auth organization & trigger backups / restore.

Jackseed:
I'll go with your recommandation @lsenta. Admin app is on the roadmap anyway so let's keep the (A) option for the sprint that will launch it.

PR #1345, opened 2019-11-28
admin: add password check on post

lsenta:
Fix issue #700 by adding a simple password form for the admin functions.

### DIFF
--- a/apps/backend-functions/src/admin.ts
+++ b/apps/backend-functions/src/admin.ts
@@ -56,11 +56,12 @@ export async function onRequestAccessToAppWrite(
 
 /**
  * Decorates another function, this will check that the password is valid
- * for admin operations.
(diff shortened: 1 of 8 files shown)


apps/backend-functions/src/admin.ts
```diff
@@ -53,6 +54,26 @@ export async function onRequestAccessToAppWrite(
   );
 }
 
+/**
+ * Decorates another function, this will check that the password is valid
+ * for admin operations.
+ *
+ * See issue#700 for details.
+ *
+ * @param f a function to protect, should be post'd.
```
Hugo-Hontans: Do you mean posted?

lsenta: POST is an http verb, I updated the doc!

PR #1349, opened 2019-12-02
Data migration review

### LINKED ISSUE #1054: Data Model - Migration Script to v1.1
# Data Model - Migration Script to v2.0

## Objective
- [x] Implement a new script to go from 1.0 version (Toronto) to 1.1 (new data model created with #1051 & #1052)

Mussuranae:
Fixes #1054 

migration afm => master


apps/backend-ops/src/firestoreMigrations/0002.ts
```diff
@@ -0,0 +1,134 @@
...
+    delete orgData.catalog;
+    delete orgData.officeAddress;
+    delete orgData.phoneNumber;
+    delete orgData.members;
+
+    const newData = {
+      ...orgData,
+      addresses: {
+        main : {
+          city: address,
```
lsenta: no need for `withoutUndefined` on this?

apps/backend-ops/src/firestoreMigrations/0002.ts
```diff
@@ -0,0 +1,134 @@
...
+            originalRef: '',
+            ref: '',
+            url: prizeData.logo
+          }
+        }))
+      },
+      main: {
+        ...movieData.main,
+        poster: {
+          originalRef: '',
```
lsenta: Could you confirm this is not `optional`?

apps/backend-ops/src/firestoreMigrations/0002.ts
```diff
@@ -0,0 +1,134 @@
...
+        promotionalElements: movieData.promotionalElements.promotionalElements.map(promoData => {
+          const { url } = promoData;
+          delete promoData.url;
+
+          return {
+            ...promoData,
+            media: {
+              originalRef: '',
+              ref: '',
+              url: url
```
lsenta: `url: url` is equivalent to just `url`

apps/backend-ops/src/firestoreMigrations/0002.ts
Mussuranae: No but these fields are mandatory

apps/backend-ops/src/firestoreMigrations/0002.ts
```diff
@@ -0,0 +1,134 @@
...
+  const newUserData = users.docs.map(async (userDocSnapshot: any): Promise<any> => {
+    const userData = userDocSnapshot.data();
+
+    const { avatar } = userData;
+
+    const newData = {
+      ...userData,
+      avatar: {
+        originalRef: '',
+        ref: avatar || '',
```
lsenta: no need for the optional thiny?

apps/backend-ops/src/firestoreMigrations/0002.ts
Mussuranae: It is :/

apps/backend-ops/src/main.ts
```diff
@@ -65,6 +66,12 @@ async function upgradeAlgoliaMovies() {
   return Promise.all(promises);
 }
 
+function migrateToV2() {
+  const { db } = loadAdminServices();
+  updateAdressesOrganizationDocument(db);
+  updatePicturesMovieDocument(db);
```
lsenta: good, now do like https://github.com/blockframes/blockframes/blob/master/apps/backend-ops/src/firestoreMigrations/0001.ts  :
expose a SINGLE `upgrade` function, because you are leaking details about the migration into the main.ts.

As a developer working on the `main.ts` you are forcing me to know how the migration `0002` is implemented and which functions to call, what if I forget one? That's leaky. Keep all the details in the migration itself.

`export async function upgrade(db: Firestore) { ... }`

apps/backend-ops/src/firestoreMigrations/0002.ts
Mussuranae: I don't use this function anymore in the script, because we just change the avatar.
Have I to delete it ?

lsenta:
@phpgeekfr could you review this? The migration code looks fine, but you are the data model owner.

apps/backend-ops/src/firestoreMigrations/0002.ts
lsenta: Nope we still use it in other part of the codebase.

Mussuranae: ok

PR #1350, opened 2019-12-02
Issue#1318 delivery

Hugo-Hontans:
- [x] Use a CollectionService
- [x] Use CollectionGuards
- [x] Update akita ng fire


libs/material/src/lib/delivery/+state/delivery.model.ts
```diff
@@ -41,3 +41,29 @@ export function createDelivery(params: Partial<Delivery>) {
     ...params
   } as Delivery;
 }
+
+export function timestampObjectsToDate(docs: any[]) {
```
GrandSchtroumpf: What about `fromTimestampToDate(docs: { date: Timestamp }[] = [])` ?
Like that you don't have to check if there is no docs

libs/material/src/lib/delivery/+state/delivery.model.ts
```diff
@@ -41,3 +41,29 @@ export function createDelivery(params: Partial<Delivery>) {
...
+  if (!docs) {
+    return [];
+  }
+
+  return docs.map(doc => {
```
GrandSchtroumpf:
```typescript
return docs.map(doc => doc.date ? { ...doc, date: doc.date.toDate() } : doc);
```

libs/material/src/lib/delivery/+state/delivery.model.ts
```diff
@@ -41,3 +41,29 @@ export function createDelivery(params: Partial<Delivery>) {
...
+    if (doc.date) {
+      return { ...doc, date: doc.date.toDate() };
+    } else {
+      return doc;
+    }
+  });
+}
+
+/** Takes a DeliveryDB (dates in Timestamp) and returns a Delivery with dates in type Date */
+export function convertDeliveryWithTimestampsToDelivery(delivery: DeliveryWithTimestamps): Delivery {
```
GrandSchtroumpf: Call it : `createDeliveryFromFirestore` to make it easy to read, and coherent with akita-ng-fire

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -26,36 +26,34 @@ interface AddDeliveryOptions {
...
-  const mgDeadlines = delivery.mgDeadlines || [];
-
-  return {
-    ...delivery,
-    dueDate: delivery.dueDate ? delivery.dueDate.toDate() : null,
-    steps: timestampObjectsToDate(delivery.steps),
-    mgDeadlines: timestampObjectsToDate(mgDeadlines)
-  };
-}
+export const deliveryQuery = (deliveryId: string): Query<DeliveryWithTimestamps> => ({
```
GrandSchtroumpf: There is two const with almost the same name. This is difficult to understand.

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -228,91 +229,56 @@ export class DeliveryService {
...
       oldStep => !stepsWithId.some(newStep => newStep.id === oldStep.id)
     );
 
-    // Remove stepId from the materials according to this array
-    this.removeMaterialsStepId(deletedSteps, batch);
-
-    return batch.update(deliveryRef, { steps: stepsWithId });
-  }
+    // We set the concerned materials stepId to an empty string
+    deletedSteps.forEach(step => {
```
GrandSchtroumpf: I have a hard time to understand what you're doing here. But I don't really like the idea of removing all those documents without using a batch....

libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -228,91 +229,56 @@ export class DeliveryService {
...
+        const materialsWithoutStep = materials.map(material => ({ ...material, stepId: ''}));
+        this.materialService.updateMovieMaterials(materialsWithoutStep, this.movieQuery.getActiveId());
+      }
+    })
 
-  /** Remove stepId of materials of delivery for an array of steps */
-  private removeMaterialsStepId(steps: Step[], batch: firebase.firestore.WriteBatch) {
-    // TODO(issue#773): Use a transaction to make sure we don't lose data
-    const deliveryId = this.query.getActiveId();
+    return this.update(this.query.getActiveId(), {
```
GrandSchtroumpf: You're repeating `this.query.getActive` and `...getActiveId` several time. Create a const `delivery` once and for all in the function and use it everywhere

libs/material/src/lib/delivery/+state/delivery.store.ts
```diff
@@ -27,7 +28,7 @@ export interface DeliveryWizard {
   options: DeliveryOption[];
 }
 
-export interface DeliveryState extends EntityState<Delivery>, ActiveState<string> {
+export interface DeliveryState extends CollectionState<Delivery> {
```
GrandSchtroumpf: Don't use CollectionState use `EntityState` and `ActivateState` instead.

libs/material/src/lib/delivery/+state/delivery.store.ts
```diff
@@ -44,6 +45,16 @@ export class DeliveryStore extends EntityStore<DeliveryState, Delivery> {
     super(initialState);
   }
 
+  // TODO: #issue1288, type correctly delivery
```
GrandSchtroumpf: I think, this has been solve in the latest version of akita

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -135,10 +135,16 @@ export class MaterialService extends CollectionService<MaterialState> {
     materials.forEach(material => this.update(material.id, { isPaid: !material.isPaid }));
   }
 
-  /** Update materials of a movie (specific fields like 'owner' and 'storage'). */
+  /** Update materials of a movie (specific fields like 'owner', 'storage', 'stepId'). */
   public updateMovieMaterials(materials: Material[], movieId: string) {
     this.subcollectionPath = `movies/${movieId}/materials`;
-      materials.forEach(material => this.update(material.id, material));
+    materials.forEach(material => this.update(material.id, material));
```
GrandSchtroumpf: I don't like the idea of doing that many update without a batch. Use `this.update(materials)` directly

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -135,10 +135,16 @@ export class MaterialService extends CollectionService<MaterialState> {
...
+  }
+
+  /** Update stepId of materials of a delivery to empty string. */
+  public removeStepIdDeliveryMaterials(materials: Material[], deliveryId: string) {
+    this.subcollectionPath = `deliveries/${deliveryId}/materials`;
+    materials.forEach(material => this.update(material.id, { stepId: '' }));
```
GrandSchtroumpf: Same here

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -242,6 +248,18 @@ export class MaterialService extends CollectionService<MaterialState> {
     return this.getValue();
   }
 
+  /** Returns a snapshot of movie materials and sets the subcollectionPath. */
+  public async getMovieMaterials(movieId: string) {
+    this.subcollectionPath = `movies/${movieId}/materials`;
```
GrandSchtroumpf: What is subcollectionpath? What if I call getValue from somewhere else just after ? This seems super dangerous.

Hugo-Hontans:
Replaced by PR #1357

PR #1356, opened 2019-12-04
Fixes 1053 : auto migrations script for ops

lsenta:
- [x] remove dead code in ops
- [x] restructure and clean the ops code
- [x] add automated system to run all migrations since your current db version
- [x] update prepare for testing, etc to manage migrations
- [x] refactor the code to generalize the restore flag => MAINTENANCE mode
  - we'll use this in the rules to disable access to the app during maintenance, cc @ioaNikas 
- [x] block database during migrations / preparation


lsenta:
@ioaNikas : I removed some of your migration code and I implemented a flag for maintenance we'll use in the rules, please take a look
@Maxstgt : I moved and reorganized some of your code around ops & algolia, please take a look
@Mussuranae : this is the migration tools we've been waiting for! Please take a look, I'll test it tomorrow.

lsenta:
@phpgeekfr I'd like your feedback and I've been told you'll own writing the migrations, please let me know what you think.

phpgeekfr:
@lsenta no problem, Is there something particular that I need to look at ?
If I understand well, now migrations are automated and script know the current DB version by looking in `_META`  ?

Maxstgt:
@lsenta LGTM

apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -114,29 +43,47 @@ export async function restore(appURL: string) {
...
-    console.error(e);
-  }
+  const url = getBackupURL(appURL);
+  const form = {
+    password: process.env['ADMIN_PASSWORD']
+  };
+
+  // promisified request
+  return new Promise((resolve, reject) => {
+    request.post({ url, form }, (error, response) => {
```
phpgeekfr: There is a very good package [request-promise](https://www.npmjs.com/package/request-promise) for this. Maybe overkill for our usage ..

apps/backend-ops/src/migrations.ts
```diff
@@ -1,8 +1,17 @@
...
+import { last } from 'lodash';
+import { IMigrationWithVersion, MIGRATIONS, VERSIONS_NUMBERS } from './firestoreMigrations';
+import { backup } from './firebaseSetup';
+import { appUrl } from '@env';
+import { endMaintenance, startMaintenance } from '../../backend-functions/src/maintenance';
 
 export const VERSION_ZERO = 0;
 
-export interface VersionDoc {
+export interface IVersionDoc {
```
phpgeekfr: The Apple touch :)

apps/backend-ops/src/migrations.ts
```diff
@@ -31,3 +37,49 @@ export async function updateDBVersion(db: Firestore, version: number): Promise<a
...
+    version,
+    upgrade: MIGRATIONS[version].update
+  }));
+}
+
+export async function migrate(withBackup: boolean = true) {
+  console.info('start the migration process...');
+  const { db } = loadAdminServices();
+
+  // TODO: disable the database updates
```
phpgeekfr: maybe add issue number to your TODOs to retrieve theme easily

apps/backend-ops/src/migrations.ts
```diff
@@ -31,3 +37,49 @@ export async function updateDBVersion(db: Firestore, version: number): Promise<a
...
+    console.info('backup done, moving on to the migrations...');
+  } else {
+    console.warn('⚠️ skipping the backup before running migrations, are you sure?');
+  }
+  try {
+    const currentVersion = await loadDBVersion(db);
+    const migrations = selectAndOrderMigrations(currentVersion);
+    const lastVersion = last(migrations).version;
+
+    console.info(`Running ${migrations.length} between ]${currentVersion}, ${lastVersion}]`);
```
Mussuranae: Is there normal to get 2 ] on this line ?
```between ]${currentVersion}, ${lastVersion}]```

ioaNikas:
My first migration code was so old we can say him "ok boomer". Good thing you got rid of it.

apps/backend-ops/src/migrations.ts
lsenta: Fixed!

apps/backend-ops/src/firebaseSetup.ts
lsenta: Agreed, if I use this a third time I'll add it.

apps/backend-ops/src/migrations.ts
lsenta: :)

apps/backend-ops/src/migrations.ts
lsenta: Yep, that's the range of migrations, between `]2, 5]` means that it contains `3, 4, 5`

Mussuranae: So for this example, the currentVersion 2 will not be included in this range ? 
And if it was [2, 5], the 2nd version would be included in the range ?

apps/backend-ops/src/migrations.ts
lsenta: (it's a default in create react app configs, it became muscle memory: https://palantir.github.io/tslint/rules/interface-name/)

apps/backend-ops/src/migrations.ts
lsenta: If the data is already in version 2, you're not going to apply the migration v2.
This is just logging to help you follow along migrations!

PR #1357, opened 2019-12-04
Issue#1318 material services

Hugo-Hontans:
- [x] Implements akita-ng-fire to delivery
- [x] Create 3 separate service for material


libs/material/src/lib/delivery/+state/delivery.service.ts
```diff
@@ -64,7 +65,8 @@ export class DeliveryService extends CollectionService<DeliveryState> {
     private permissionsService: PermissionsService,
     private shService: StakeholderService,
     private walletService: WalletService,
-    protected store: DeliveryStore
+    protected store: DeliveryStore,
```
GrandSchtroumpf: Put the store at the top to make things more readable

libs/material/src/lib/material/+state/material.service.ts
```diff
@@ -22,9 +22,9 @@ export class MaterialService extends CollectionService<MaterialState> {
     super(store)
   }
 
-  get path() {
-    return this.subcollectionPath;
-  }
+  // get path() {
```
GrandSchtroumpf: Remove this

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,143 @@
...
+import { CollectionConfig, CollectionService } from 'akita-ng-fire';
+import { Material, createMaterial, MaterialStatus } from './material.model';
+import { MaterialState, MaterialStore } from './material.store';
+import { MovieQuery } from '@blockframes/movie';
+import { Delivery } from '../../delivery/+state';
+
+@Injectable({
+  providedIn: 'root'
+})
+@CollectionConfig({ path: 'movies/:movieId/materials' })
```
GrandSchtroumpf: Remove path here if you're overriding it below

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  /** Returns a material to be pushed in a formGroup. */
+  public createMaterial(): Material {
+    const id = this.db.createId();
+    return createMaterial({ id });
+  }
+
+  /** Deletes material of the movie sub-collection in firebase. */
+  public async delete(materialId: string, delivery: Delivery) {
+    const material = await this.getValue(`${materialId}`);
```
GrandSchtroumpf: You should do that in a transaction !

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,143 @@
...
+    // Checks if this material belongs to multiple delivery.
+    // If so, update the deliveryIds, otherwise just delete it.
+    if (material.deliveryIds.length === 1) {
+      this.remove(material.id);
+    } else {
+      this.update(material.id, {
+        deliveryIds: material.deliveryIds.filter(id => id !== delivery.id)
+      });
+    }
+    this.db.doc(`deliveries/${delivery.id}`).update({ validated: [] });
```
GrandSchtroumpf: I would prefer to use the deliveryService.update(delivery.id, { validated: [] }, { write: tx }).
But you can also use the `this.db` is you prefer.
Just do that in a transaction !

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  /** Update materials of a movie (specific fields like 'owner', 'storage', 'stepId'). */
+  public updateMovieMaterials(materials: Material[]) {
+    // TODO: issue#1352 use a multiple update
+    materials.forEach(material => this.update(material.id, material));
+  }
+
+  /** Update materials of a delivery (materials loaded from movie). */
+  public async updateMaterials(materials: Material[], delivery: Delivery) {
+    const movieMaterials = await this.getValue();
```
GrandSchtroumpf: Transaction !!!

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,143 @@
...
+    });
+  }
+
+  /** Create a material in a movie. */
+  public setNewMaterial(material: Material, delivery: Delivery) {
+    this.add({ ...material, deliveryIds: [delivery.id] });
+  }
+
+  /**  Checks properties of two material to tell if they are the same or not. */
+  public isTheSame(matA: Material, matB: Material): boolean {
```
GrandSchtroumpf: This is a pure function (you're not using `this`), put it outside of the class

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,143 @@
...
+      value,
+      description,
+      category,
+      stepId
+    });
+    return JSON.stringify(getProperties(matA)) === JSON.stringify(getProperties(matB));
+  }
+
+  /** Update deliveryIds of a material when this one has the same values that an other. */
+  public updateMaterialDeliveryIds(sameValuesMaterial: Material, delivery: Delivery) {
```
GrandSchtroumpf: Hmmm the name of the material is weird, you imply that the values are already the same, but you're doing a verification below.

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,143 @@
...
+    } else {
+      return this.update(material.id, {
+        deliveryIds: sameIdMaterial.deliveryIds.filter(id => id !== delivery.id)
+      });
+    }
+  }
+
+  /** Update the property status of selected materials. */
+  public updateStatus(materials: Material[], status: MaterialStatus, movieId: string) {
+    materials.forEach(material => this.update(material.id, { status }));
```
GrandSchtroumpf: You can use multiple update here : 
```typescript
this.update(materials.map(m => m.id), { status })
```

GrandSchtroumpf: you are not using `movieId`

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /** Update the property isOrdered of selected materials. */
+  public updateIsOrdered(materials: Material[], movieId: string) {
+    materials.forEach(material => this.update(material.id, { isOrdered: !material.isOrdered }));
```
GrandSchtroumpf: can you try :
```typescript
const ids = materials.map(m => m.id);
this.update(ids, (material) => ({ isOrdered: !material.isOrdered }))
```

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  public updateIsPaid(materials: Material[], movieId: string) {
```
GrandSchtroumpf: same as above

libs/material/src/lib/material/guards/delivery-materials.guard.ts
```diff
@@ -23,10 +25,12 @@ export class DeliveryMaterialsGuard extends CollectionGuard<MaterialState> {
...
   sync(next: ActivatedRouteSnapshot) {
     this.store.reset();
+    const { movieId, deliveryId } = next.params;
+
     return this.deliveryQuery.getActive().mustBeSigned
-      ? this.service.syncCollection(`deliveries/${next.params.deliveryId}/materials`)
-      : this.service.syncCollection(`movies/${next.params.movieId}/materials`, ref =>
-          ref.where('deliveryIds', 'array-contains', next.params.deliveryId)
-          );
+      ? this.service.syncCollection({ params: { deliveryId }})
```
GrandSchtroumpf: It works in the end ?

libs/material/src/lib/material/+state/delivery-material.service.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  /** Deletes material of the delivery sub-collection in firebase. */
+  public deleteDeliveryMaterial(materialId: string) {
+    this.remove(materialId)
+  }
+
+  /** Update materials of a delivery (materials loaded from delivery). */
+  public async updateDeliveryMaterials(materials: Material[]) {
+    // TODO: (ISSUE#773) We should load an update the data within a transaction.
```
GrandSchtroumpf: Transactions!
Look at my comments from the other service for the rest of the code.

libs/material/src/lib/material/+state/movie-material.service.ts
Hugo-Hontans: I have to give a path otherwise I have this error:
```
core.js:4002 ERROR Error: Uncaught (in promise): Error: You should provide a path to the collection
Error: You should provide a path to the collection
```

libs/material/src/lib/material/+state/movie-material.service.ts
Hugo-Hontans: If the values of materials are the same, we just have to add the deliveryId in the array `deliveryIds` of a movie material in the case where the material does not exist yet.

libs/material/src/lib/material/guards/delivery-materials.guard.ts
Hugo-Hontans: Nop, old code

libs/material/src/lib/material/+state/movie-material.service.ts
Hugo-Hontans: I can't use deliveryService in this service cause of circular dependency (deliveryService already import movieMaterialService

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,142 @@
...
+  public async delete(materialId: string, delivery: Delivery) {
+    return this.db.firestore.runTransaction(async tx => {
+      const material = await this.getValue(`${materialId}`);
+
+      // Checks if this material belongs to multiple delivery.
+      // If so, update the deliveryIds, otherwise just delete it.
+      if (material.deliveryIds.length === 1) {
+        this.remove(material.id, { write: tx });
+      } else {
+        this.update(material.id, {
```
GrandSchtroumpf: To make it more readable you could do : 
```typescript
const deliveryIds = material.deliveryIds.filter(id => id !== delivery.id);
this.update(material.id, { deliveryIds }, { write: tx })
```

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  /** Create a material in a movie. */
+  public setNewMaterial(material: Material, delivery: Delivery, tx: firebase.firestore.Transaction) {
+    this.add({ ...material, deliveryIds: [delivery.id] }, { write: tx });
+  }
+
+  /** Update deliveryIds of a material when this one has the same values that an other. */
+  public updateMaterialDeliveryIds(sameValuesMaterial: Material, delivery: Delivery, tx: firebase.firestore.Transaction) {
+    if (!sameValuesMaterial.deliveryIds.includes(delivery.id)) {
+      this.update(sameValuesMaterial.id, {
```
GrandSchtroumpf: Same as above

libs/material/src/lib/material/+state/movie-material.service.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  /** Checks if the material belongs to multiple delivery, if so: update the deliveryIds, otherwise just delete it. */
+  public removeMaterial(material: Material, delivery: Delivery, sameIdMaterial: Material, tx: firebase.firestore.Transaction) {
+    if (sameIdMaterial.deliveryIds.length === 1) {
+      return this.remove(material.id, { write: tx });
+    } else {
+      return this.update(material.id, {
+        deliveryIds: sameIdMaterial.deliveryIds.filter(id => id !== delivery.id)
```
GrandSchtroumpf: Same as above

libs/material/src/lib/material/+state/template-material.service.ts
```diff
@@ -0,0 +1,76 @@
...
+  /** Convert a MaterialDocument into a MaterialTemplateDocument. */
+  formatToFirestore(material: Material) {
+    return createMaterialTemplate(material);
+  }
+
+  async getTemplateMaterials(templateId: string) {
+    const materialsSnapshot = await this.db.firestore.collection(`templates/${templateId}/materials`).get();
+    const materials = [];
+    materialsSnapshot.forEach(material => materials.push(material.data()));
+    return materials as Material[];
```
GrandSchtroumpf:
```typescript
const materials = await this.db.collection(`templates/${templateId}/materials`).ref.get();
return materials.map(material => material.data());
```

libs/material/src/lib/template/+state/template.service.ts
```diff
@@ -68,23 +69,25 @@ export class TemplateService extends CollectionService<TemplateState>{
...
     const templates = this.query.getAll();
     const selectedTemplate = templates.find(template => template.name === name);
-    const templateMaterials = await this.materialService.getTemplateMaterials(selectedTemplate.id);
+    const templateMaterials = await this.templateMaterialService.getTemplateMaterials(selectedTemplate.id);
 
     if (materials.length > 0) {
+      const batch = this.db.firestore.batch();
       // Delete all materials of template
-      templateMaterials.forEach(material => this.materialService.remove(material.id));
+      templateMaterials.forEach(material => this.templateMaterialService.remove(material.id, { write: batch }));
```
GrandSchtroumpf: I would rather having the opposite : 
```typescript
const ids = templateMaterials.map(({ id }) => id);
this.templateMaterialService.remove(ids, { write: batch });
this.templateMaterialService.add(materials, { write: batch });
```

libs/material/src/lib/template/components/template-add/template-add.component.ts
```diff
@@ -15,14 +15,14 @@ export class TemplateAddComponent {
...
-    private router: Router,
-    private routes: ActivatedRoute
+    private router: Router
   ) {}
 
   public addTemplate(templateName: string) {
     const template = this.service.createTemplate(templateName);
     this.close();
-    this.router.navigate([`../${template.id}`], { relativeTo: this.routes })
+    // TODO: issue#1332, relative path doesn't work.
```
GrandSchtroumpf: Have you used the `resolve` function from `@blockframes/utils/routes` to do it relatively ?

libs/material/src/lib/template/components/template-repertory/template-repertory.component.html
```diff
@@ -18,7 +18,7 @@
   <ng-container matColumnDef="action">
     <th mat-header-cell *matHeaderCellDef></th>
     <td mat-cell *matCellDef="let template">
-      <button mat-button (click)="delete.emit(template)"><mat-icon>delete_outline</mat-icon></button>
+      <button mat-button (click)="delete.emit(template); $event.stopPropagation()"><mat-icon>delete_outline</mat-icon></button>
```
GrandSchtroumpf: Create a method on the typescript side instead of packing everything in the click

libs/material/src/lib/template/components/template-add/template-add.component.ts
Hugo-Hontans: No, I added a comment about that on the relative ticket

Issue #1388, opened 2019-12-10
Distribution Deals optimizations


phpgeekfr:
- [x] Update `TERRITORIES` static model to be `ISO-3166` compliant
- [x] Update `LANGUAGES` static model to be `RFC-5646` compliant dedicated issue create #1658
- [x] search for `#1388` in code
- [x] rename to `licensedRights` . checked with @Jackseed

Jackseed:
Best thing for distributionRights would be licensedRights

PR #1399, opened 2019-12-11
Feat 1385 & 1355

phpgeekfr:
This PR builds on top of feat-1061-bis
Feat 1385 & 1355 : some debug on catalog dashboard + contract implementation


apps/catalog/dashboard/dashboard/src/app/dashboard.module.ts
```diff
@@ -6,7 +6,7 @@ import { MovieCreateComponent } from '@blockframes/movie/movie/components/movie-
 const routes: Routes = [
   {
     path: '',
-    redirectTo: 'home',
+    redirectTo: 'catalog',
```
GrandSchtroumpf: Why did you remove the home page ?

apps/catalog/dashboard/dashboard/src/app/pages/dashboard-home/dashboard-home.component.html
```diff
@@ -1,8 +1,53 @@
 <section test-id="content" page-id="catalog-homepage">
-    <article fxLayout="column" fxLayoutAlign="center center">
-        <h1 class="mat-display-3">Welcome to Bigger Boat - Dashboard</h1>
-        <p class="mat-body-1">
-            A consolidation of film libraries to create new opportunities through secure transactions.</p>
-        <button mat-button color="accent" test-id="import" [routerLink]="['/layout/o/home/import']">Upload avails & films doc</button>
-    </article>
-</section>
+  <article fxLayout="column" fxLayoutAlign="center center">
```
GrandSchtroumpf: Is it supposed to match the new design ? Should it go into the mono-catalog app ?

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.firestore.ts
```diff
@@ -0,0 +1,40 @@
...
+
+type Timestamp = firestore.Timestamp;
+
+export const enum ContractStatus {
+  submitted = 'submitted',
+  accepted = 'accepted',
+}
+
+export interface ContractTitleDetail {
+  titleId: string, // ie: movieId
```
GrandSchtroumpf: Why not naming it movieId ?

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.firestore.ts
```diff
@@ -0,0 +1,40 @@
...
+
+interface ContractRaw<D> {
+  id: string,
+  parentContractIds?: string[],
+  childContractIds?: string[],
+  parties: Party[],
+  status: ContractStatus,
+  scope: TermsRaw<D>,
+  signDate?: D,
+  titles: { [titleId: string]: ContractTitleDetail },
```
GrandSchtroumpf: Use native typescript typing `Record<string, ContractTitleDetail>`

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.firestore.ts
```diff
@@ -0,0 +1,40 @@
...
+  price: Price;
+  paymentSchedule?: string; // @todo #1397 change this when creating invoices
+}
+
+export interface InvoiceRaw<D> {
+  // @todo #1397
+}
+
+export interface ContractDocumentWithDates extends ContractRaw<Date> {
```
GrandSchtroumpf: I really don't like this `WithDates` pattern. Why not just `Contract` for Frontend & `ContractDocument` for Backend. We end up with super long name, and if we have more stuff to change between both, the naming `withDates` won't have sense anymore.

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export interface CatalogCart {
+  name: string;
+  status: CartStatus;
+  deals: string[];
+  price: Price;
+}
+
+export interface MovieData { // @TODO (#1388) check this object
```
GrandSchtroumpf: Could we not use a subset of the movie object instead ?

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
```diff
@@ -0,0 +1,103 @@
...
+  movieName: string;
+  duration: DateRange;
+  territory: string;
+  rights: string;
+  languages: string;
+  dubbed: string;
+  subtitle: string;
+}
+
+export type Contract = ContractDocumentWithDates;
```
GrandSchtroumpf: This doesn't change my comment above. I think it's complicated process `ContractRaw -> ContractRaw<Date> -> ContractDocumentWithDates -> Contract`.

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
```diff
@@ -0,0 +1,103 @@
...
+  return {
+    titleId: '',
+    distributionDealIds: [],
+    ...params,
+    price: createPrice(params.price),
+  };
+}
+
+/**
+ * @param contract
```
GrandSchtroumpf: Add comment

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+export function validateContract(contract: Contract): boolean {
+
+  // First, contract must have at least a licensee and a licensor
+
+  if (contract.parties.length < 2) { return false; }
+  const licensees = contract.parties.filter(p => p.role == getCodeIfExists('LEGAL_ROLES', 'licensee'))
```
GrandSchtroumpf: Is ok if you don't have a lot of contract parties (<1000). Else, better use a single loop

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
```diff
@@ -0,0 +1,103 @@
...
+  const licensors = contract.parties.filter(p => p.role == getCodeIfExists('LEGAL_ROLES', 'licensor'))
+
+  if (!licensees.length || !licensors.length) { return false; }
+
+  for(const licensee of licensees) {
+    if(licensee.orgId === undefined) {
```
GrandSchtroumpf: What about null or empty string ?

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
```diff
@@ -0,0 +1,103 @@
...
+      delete licensee.orgId
+    }
+    if(typeof licensee.showName !== 'boolean') {
+      return false;
+    }
+  }
+
+  for(const licensor of licensors) {
+    if(licensor.orgId === undefined) {
```
GrandSchtroumpf: Same here

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.service.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  //////////////////
+  /// CART STUFF
+  //////////////////
+
+  /**
+   * Adds a deal to cart identified by "name"
+   * @param dealId 
+   * @param name
```
GrandSchtroumpf: What name is it?

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.service.ts
```diff
@@ -0,0 +1,189 @@
...
+   * @param name 
+   */
+  public async addDealToCart(dealId: string, name: string): Promise<CatalogCart> {
+    const cart = await this.getCart(name);
+    cart.deals.push(dealId);
+    return this.updateCart(cart);
+  }
+
+  /**
+   *
```
GrandSchtroumpf: Comments

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.service.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   * Adds a deal to cart identified by "name"
+   * @param dealId 
+   * @param name 
+   */
+  public async addDealToCart(dealId: string, name: string): Promise<CatalogCart> {
+    const cart = await this.getCart(name);
+    cart.deals.push(dealId);
+    return this.updateCart(cart);
```
GrandSchtroumpf: You should use a firestore transaction. 
Use native `akita-ng-fire` instead, it does all the good stuff for you :
```typescript
this.update(name, (cart) => ({ deals: [ ...cart.deals, dealId ]})) 
```

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.service.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * 
+   * @param amount 
+   * @param currency
+   * @param _name
+   */
+  public async submitCart(amount: number, currency: MovieCurrenciesSlug, _name?: string) {
```
GrandSchtroumpf: Same here.

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.service.ts
```diff
@@ -0,0 +1,189 @@
...
+      status: CartStatus.submitted
+    };
+    this.updateCart(updatedCart);
+  }
+
+  /**
+   * Creates an empty cart
+   * @param _name
+   */
+  private async initCart(_name?: string): Promise<CatalogCart> {
```
GrandSchtroumpf: Same here.
Also do `initCart(name: string = 'default')`

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.service.ts
```diff
@@ -0,0 +1,189 @@
...
+    const name = _name ? _name : 'default';
+    const cart: CatalogCart = createCart({ name });
+    await this.db.doc<CatalogCart>(`orgs/${this.organizationQuery.getActiveId()}/cart/${name}`).set(cart);
+    return cart;
+  }
+
+  private async updateCart(cart: CatalogCart): Promise<CatalogCart> {
```
GrandSchtroumpf: Remove this function if doesn't do anything more than native `akita-ng-fire`

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.service.ts
GrandSchtroumpf: You'll need to ceate a getter for your path also (see with Hugo)

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.firestore.ts
phpgeekfr: New denomination. To handle movies, series etc .. Check with Vincent ;)

apps/catalog/dashboard/dashboard/src/app/dashboard.module.ts
phpgeekfr: The old one was broken. Just added a new component to be able to use and test the app. This is a quickfix, no css, no specific tests, just a movie list.

apps/catalog/dashboard/dashboard/src/app/pages/dashboard-home/dashboard-home.component.html
phpgeekfr: Nop, just for testing and to have a "running" app

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.firestore.ts
phpgeekfr: Just did what was already used everywhere. Dont really like this WithDates either :)

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
phpgeekfr: Yes of course it should. Don't know where this object come from, but I think this is going to be deleted in #1388

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
phpgeekfr: It will never exceed 4 or 5 parties

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.service.ts
phpgeekfr: This is the name of the cart. `default` if not specified on creation. This will allow us to handle multiple carts for a unique org. Added comment

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.service.ts
phpgeekfr: added comment

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.firestore.ts
GrandSchtroumpf: Maybe you can add a comment `/** movie/serie id */`

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.firestore.ts
GrandSchtroumpf: Could you check with @ioaNikas or @Hugo-Hontans to see if we can improve that. I'm afraid it's become a "thing" ^^

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
GrandSchtroumpf: Can you change the `TODO` naming to "should be removed" or something more explicit ?

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
```diff
@@ -26,15 +29,77 @@ export interface MovieData {
...
+    if(!licensee.orgId) {
+      delete licensee.orgId
+    }
+    if(typeof licensee.showName !== 'boolean') {
+      return false;
+    }
+  }
+
+  for(const licensor of licensors) {
+    if(!licensor.orgId) {
```
GrandSchtroumpf: Same here

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
```diff
@@ -26,15 +29,77 @@ export interface MovieData {
...
+  // First, contract must have at least a licensee and a licensor
+
+  if (contract.parties.length < 2) { return false; }
+  const licensees = contract.parties.filter(p => p.role === getCodeIfExists('LEGAL_ROLES', 'licensee'))
+  const licensors = contract.parties.filter(p => p.role === getCodeIfExists('LEGAL_ROLES', 'licensor'))
+
+  if (!licensees.length || !licensors.length) { return false; }
+
+  for(const licensee of licensees) {
+    if(!licensee.orgId) {
```
GrandSchtroumpf: I'm not sure to understand this line : if there is no "orgId", then delete "orgId".

apps/catalog/marketplace/marketplace/src/app/movie/home/home.component.html
```diff
@@ -47,7 +47,7 @@ <h2>{{ movie.main.title.international }}</h2>
                       {{ country | translateSlug: 'TERRITORIES'}}{{isLast ? '' : ', '}}{{i == 1 && movie.main.originCountries.length > 2 ? ',  ' : ''}}
                     </span>
                     <span *ngIf="movie.main.originCountries.length > 2">... </span>
-                    <span *ngIf="movie.main.length">| {{ movie.main.length }} min </span>
+                    <span *ngIf="movie.main.totalRunTime">| {{ movie.main.totalRunTime }} min </span>
```
GrandSchtroumpf: I'm not sure, but many more readable like that ? What do you think ?
```html
<span *ngIf="movie.main.totalRunTime as time">| {{ time }} min </span>
```

GrandSchtroumpf: You can apply in the other files if you think it's better.

apps/catalog/marketplace/marketplace/src/app/movie/view/view.component.html
```diff
@@ -15,10 +15,10 @@ <h1>{{movie.main.title.international}}</h1>
         <div  class="movie-info" fxLayout="row" fxLayoutAlign="space-between center">
           <div>
             <span *ngIf="movie.main.productionYear">{{ movie.main.productionYear }}</span>
-            <span *ngIf="movie.main.status || movie.main.length"> | </span>
+            <span *ngIf="movie.main.status || movie.main.totalRunTime"> | </span>
```
GrandSchtroumpf: It's kind of difficult to understand what happen here. Would it make sense to create a pipe for that ?

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -227,17 +230,13 @@ export function createDistributionDeal(params: Partial<DistributionDeal> = {}):
...
-    price: createPrice(params.price),
     titleInternalAlias: '',
     formatProfile: FormatProfile.unknown,
     download: false,
     holdbacks: [],
-    fees: [],
-    ...params,
-    licensee: createStakeholder(params.licensee),
-    licensor: createStakeholder(params.licensor),
+    catchUp: createTerms(params.catchUp),
```
GrandSchtroumpf: In the other factories you put the subcreate after the `...params`. Any reason to do it differently here ?

libs/utils/src/lib/common-interfaces/identity.ts
```diff
@@ -81,6 +81,15 @@ export function createStakeholder(params: Partial<Stakeholder> = {}): Stakeholde
   }
 }
 
+export function createParty(params: Partial<Party> = {}): Party {
+  return {
+    ...createStakeholder(params),
```
GrandSchtroumpf: So a "Party" is a superset of the stakeholder ?

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.firestore.ts
phpgeekfr: roger that

apps/catalog/marketplace/marketplace/src/app/distribution-deal/+state/cart.model.ts
phpgeekfr: Lol. Thanks ..

apps/catalog/marketplace/marketplace/src/app/movie/home/home.component.html
phpgeekfr: I agree

libs/movie/src/lib/movie/+state/movie.model.ts
phpgeekfr: in certain circumstances `...params` must be placed before subcreates because subcreates functions output can be erased right after creation by `...params`.
For example:
No need to put `...params` before:
```
export function createTerms(params: Partial<Terms> = {}): Terms {
  return {
    ...params
  };
}
```
but for this yes
```
export function createImgRef(ref: Partial<ImgRef> | string = {}): ImgRef {
  const _ref = typeof ref === 'string' ? { url: ref } : ref;
  return {
    url: '',
    ref: '',
    originalRef: '',
    ..._ref
  };
}
```

libs/utils/src/lib/common-interfaces/identity.ts
phpgeekfr: Party is a Person with a legalRole. Identity => Person => Party

image

PR #1404, opened 2019-12-13
chip and form changes

Mussuranae:
Chips and form field changes

### DIFF
--- a/styles/components/mat-form.scss
+++ b/styles/components/mat-form.scss
@@ -14,6 +14,12 @@ mat-form-field {
       position: static;
     }
   }
+
+  &[no-underline] {
+    .mat-form-field-underline {
+      display: none;
+    }
+  }
 }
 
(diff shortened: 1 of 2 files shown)


styles/components/mat-form.scss
```diff
@@ -14,6 +14,12 @@ mat-form-field {
       position: static;
     }
   }
+
+  &[no-underline] {
```
GrandSchtroumpf: Is it a property of material or a custom one ?
Be careful with custom properties because they won't be documented inside material, so developers might not understand them.

Mussuranae: It's a custom property and I was waiting to merge my PR to write this on notion.

GrandSchtroumpf: don't expect devs to look into notion when using a material component ...

Issue #1463, opened 2019-12-24
Movie form : check how to create a route for each section


phpgeekfr:
use akita to store form inputs between routes

lsenta:
- [ ] decide to save as draft? how? etc.
- [ ] decide how to deal with routes, etc.

Jackseed:
- [ ] how components will get the data from other pages?

Jackseed:
- [ ] can we re-use the form to edit the film?

GrandSchtroumpf:
Tunnel routes : 
- movie-tunnel/:movieId/{tunnel-page-name} (if movideId === 'create' don't fetch existing data)

Store data locally before saving : 
- Option 1: Injectable `MovieForm` (but might not be garbaged collected when you leave tunnel)
- Option 2: Store data in Store, and use `PersistForm` plugin from akita

Save as draft : 
- Should be a flag (public/private) on the document that we can use for queries.

How components will get the data from other pages : 
_Not sure to understand the question_
- Data will be global (either in store or in a MovieForm) and can be reach from all pages.

Can we re-use the form to edit the film :
- Yes with the `:movieId` param in the route. If param is different from "create", then fetch (once) the data from firestore and fill the form with the content.

Jackseed:
- [ ] flag private / public implemented by back team

Jackseed:
ref #1410

PR #1464, opened 2019-12-24
Use route for each movie form section #1463 

phpgeekfr:


GrandSchtroumpf:
I've gave a thought about this and I think it would be better to make the Form injectable: 
```typescript
export class MovieForm extends FormEntity<Movie> {...}
```
Like that you can reuse the same form in all pages and have access to all the validators, reset, and other utils given by the form

PR #1484, opened 2020-01-02
Setup Tunnel budget

GrandSchtroumpf:
Setup the tunnel budget pages with range component.
Good example of how to build the form.


libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -109,10 +109,16 @@ export interface MovieFestivalPrizes {
...
+export interface MovieBudgetRange {
+  from: number;
+  to: number;
+}
+
 export interface MovieBudget {
   totalBudget: string, // WIP #1052 use Price Interface?
   budgetCurrency?: string, // WIP #1052
   detailledBudget?: any // WIP #1052
+  range: MovieBudgetRange // WIP #1053
```
Mussuranae: It's estimatedBudget please, not range.
The value is a static, it is added in this PR https://github.com/blockframes/blockframes/pull/1486/files
I wait after Bruce and Vincent's review to merge it.

libs/movie/src/lib/movie/form/budget/range/range.component.html
```diff
@@ -0,0 +1,10 @@
+<form [formGroup]="range" fxLayout="row">
+  <mat-form-field>
+    <mat-label>From</mat-label>
+    <input matInput placeholder="from" formControlName="from" type="number"/>
```
Mussuranae: With the pending PR, it will change into select input with the static model.

PR #1487, opened 2020-01-03
Issue#1457 shared rules

### LINKED ISSUE #639: Add a stakeholderIds array in delivery model, so we can filter them in guard for shared documents
=> Firestore rules aren't meant to "filter" the documents we can CRUD, we got to do it in the guard.

### LINKED ISSUE #1498: backend function onMovieCreate must use transaction
backend function `onMovieCreate` must use transaction because if we add 10 movies in parallel, some org.movieIds are missing.

ioaNikas:
- [x] Updated rules to fit shared document logic.
- [x] Updated firebase function and cleaned permissions subcollections for better understanding.
- [x] Updated delivery's data model and list query to fetch the right deliveries. (close #639 )
- [x] Moving some movie collection hooks from frontend to backend functions.
- [x] onMovieCreateEvent now use transactions. (close #1498 )


package.json
```diff
@@ -175,8 +175,7 @@
     "tslib": "^1.9.0",
     "tslint": "~5.20.0",
     "typescript": "^3.4.5",
-    "webpack-bundle-analyzer": "^3.4.1",
-    "ts-jest": "24.0.0"
+    "webpack-bundle-analyzer": "^3.4.1"
```
lsenta: why?

libs/notification/src/lib/invitation/+state/invitation.service.ts
```diff
@@ -4,9 +4,10 @@ import { InvitationState } from './invitation.store';
 import { AuthQuery, AuthService } from '@blockframes/auth';
 import { createInvitationToDocument, createInvitationFromUserToOrganization, createInvitationFromOrganizationToUser } from './invitation.model';
 import { CollectionConfig, CollectionService } from 'akita-ng-fire';
-import { Organization, PublicOrganization } from '@blockframes/organization';
+import { Organization } from '@blockframes/organization';
 import { OrganizationService } from '@blockframes/organization/+state/organization.service';
 import { Invitation, InvitationStatus } from './invitation.firestore';
+import { PublicOrganization } from '@blockframes/organization/+state/organization.firestore';
```
lsenta: Doesn't look standard

package.json
ioaNikas: I don't know what's going on here, I'll just take the one from master to be sure to change nothing.

firestore.rules
```diff
@@ -153,24 +180,25 @@ service cloud.firestore {
...
     }
 
+    // Check is the organization is stakeholder on a delivery.
     function isStakeholder(orgId) {
     	return orgId in resource.data.stakeholderIds;
     }
 
     function getDocumentPermissions(orgId, docId) {
-    	return get(/databases/$(database)/documents/permissions/$(orgId)/documentPermissions/$(docId)).data;
+    	return getAfter(/databases/$(database)/documents/permissions/$(orgId)/documentPermissions/$(docId)).data;
```
lsenta: Say there's a movie I don't own, I create a transaction that changes the movie and write the permission document?
Will this pass the `getAfter`?

ioaNikas: Yes it seems that `getAfter` in this case allow every update on the permissions, I maybe should stick with simple `get` for the query in permissions collection to keep it secure ?

PR #1525, opened 2020-01-08
FormList value filter content without value

GrandSchtroumpf:
FormList get rid of the empty value recursively
```typescript
const form = new FormList([
  new FormGroup({
    name: new FormControl(''),
    description: new FormControl(''),
  })
]);
const value = form.value // []
```

### DIFF
--- a/libs/utils/src/lib/form/forms/list.form.ts
+++ b/libs/utils/src/lib/form/forms/list.form.ts
@@ -1,13 +1,31 @@
 import { AbstractControl } from '@angular/forms';
-import { FormArray } from '@angular/forms';
+import { FormArray, FormControl } from '@angular/forms';
 import { FormEntity } from './entity.form';
 import { Validator, AsyncValidator } from './types';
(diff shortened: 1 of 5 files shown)


libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -1,29 +1,61 @@
...
+  } else if (typeof value === 'number') {
+    return true;
+  } else {
+    return !!value
+  }
+}
+
 /** A list of FormField */
 export class FormList<T, Control extends AbstractControl = any> extends FormArray {
+  private _value: T[];
```
Maxstgt: since we got an indicator `private` that determines the accessibility of that property, we don't need underscores here in typescript, in my eyes.

GrandSchtroumpf: `value` already exist. I could call it `whatever` but I want to show that this is the content of the real value of `FormArray`.

PR #1526, opened 2020-01-09
Issue#1505 distribution deals guard & organization cart

### LINKED ISSUE #1440: move some models to their own lib
* [x] move `distributionDeals` related objects to `libs/movie/distribution-deals`
* [x] move `Contract` related objects to `libs/contract`
* [x] move `Cart` related objects to `libs/organization/cart`

ioaNikas:
- [x] Move cart into organization lib as a subfolder with own +state ( close #1440 ).
- [x] Create distribution deals guard for active movie
- [x] Added MatSnackBarModule import into catalog app.module to fix it
And remove some unused code.


libs/movie/src/index.ts
```diff
@@ -19,3 +19,6 @@ export * from './lib/movie/static-model';
 
 // Movie Component Module
 export * from './lib/movie/components/movie-picker/movie-picker.module';
+
+// Distribution Deals guards
+export * from './lib/distribution-deals/guards/active-movie-distribution-deals.guard';
```
GrandSchtroumpf: I'm not sure it should be here. You should put it inside `libs/movie/src/distribution-deal/index.ts` instead (create if it's not there).

libs/organization/src/lib/cart/+state/cart.service.ts
```diff
@@ -137,11 +137,11 @@ export class CartService extends CollectionService<CartState> {
     try {
       const wishlist = orgState.wishlist.map(wish => {
         if (wish.status === 'pending') {
-          const movieIds = wish.movieIds.includes(movieId) ? wish.movieIds.filter(id => id !== movieId) : wish.movieIds;
-          return { ...wish, movieIds };
-        } else {
-          return wish;
+          wish.movieIds = wish.movieIds.includes(movieId)
```
GrandSchtroumpf: This is not a good idea, you're iterating maximum two times on the movieIds. Just do : 
`wish.movieIds = wish.movieIds.filter(id => id !== movieId)`. Like that you iterate only one on the array.

apps/catalog/catalog/src/app/app.module.ts
```diff
@@ -79,7 +82,10 @@ import { FireAnalytics } from '@blockframes/utils/analytics/app-analytics';
       trackLinks: true,
       accurateTrackBounce: true,
       webvisor: true
-    })
+    }),
+
+    // Material
+    MatSnackBarModule
```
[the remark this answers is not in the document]
ioaNikas: I encountered an error when serving catalog app telling me that there was no provider for MatSnackBar a earlier this week. Tried again without it, working fine, so removing this useless import now.

PR #1529, opened 2020-01-09
Fix onMovieCreate function

ioaNikas:
Getting organization in TX to be sure it's updated since last iteration

### DIFF
--- a/apps/backend-functions/src/movie.ts
+++ b/apps/backend-functions/src/movie.ts
@@ -43,9 +43,8 @@ export async function onMovieCreate(
 ) {
   const movie = snap.data() as MovieDocument;
 
-  // Get the user document, and then his organization.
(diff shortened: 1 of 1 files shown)


apps/backend-functions/src/movie.ts
```diff
@@ -55,10 +54,14 @@ export async function onMovieCreate(
   // Create notifications and getting organization snapshot.
   const [notifications, organizationSnap] = await Promise.all([
     createNotificationsForUsers(movie, NotificationType.movieTitleCreated, user),
-    db.doc(`orgs/${organization.id}`).get()
+    db.doc(`orgs/${user.orgId}`).get()
```
lsenta: you get organization twice, once outside the tx and once inside, remove the first one (organizationsnap)

PR #1533, opened 2020-01-10
Migration catalog e2e

Hugo-Hontans:
Fix catalog app and catalog e2e tests


apps/catalog/catalog-e2e/src/support/pages/marketplace/ViewPage.ts
```diff
@@ -15,6 +15,7 @@ export default class ViewPage extends NavbarPage {
 
   public clickWishListButton() {
     cy.get('.heart-button').click();
+    cy.wait(2000);
```
GrandSchtroumpf: Why do you have to wait ?

apps/catalog/catalog/src/app/marketplace/movie/search/search.component.ts
```diff
@@ -264,6 +264,7 @@ export class MarketplaceSearchComponent implements OnInit {
   }
 
   public goToMovieDetails(id: string) {
+    console.log('coucou')
```
GrandSchtroumpf: I think you can remve that ^^

apps/catalog/catalog/src/app/marketplace/movie/search/search.component.ts
```diff
@@ -264,6 +264,7 @@ export class MarketplaceSearchComponent implements OnInit {
...
     this.router.navigateByUrl(`c/o/catalog/${id}`);
```
GrandSchtroumpf: Is it not `c/o/marketplace/${id}` instead ?

libs/organization/src/lib/cart/+state/cart.service.ts
```diff
@@ -137,9 +137,11 @@ export class CartService extends CollectionService<CartState> {
     try {
       const wishlist = orgState.wishlist.map(wish => {
         if (wish.status === 'pending') {
-          wish.movieIds = wish.movieIds.filter(id => id !== movieId)
+          const movieIds = wish.movieIds.filter(id => id !== movieId);
```
GrandSchtroumpf: Even better : 
```typescript
return {
  ...wish,
  movieIds: wish.movieIds.filter(id => id !== movieId)
}
```

apps/catalog/catalog-e2e/src/support/pages/marketplace/ViewPage.ts
Hugo-Hontans: We check when adding a film if the counter on the heart icon increases and cypress goes too fast if we don't wait. I dow it to 500

PR #1536, opened 2020-01-10
create e2e test for movie-tunnel-1

Hugo-Hontans:
Create e2e test for movie-tunnel-page-1


apps/catalog/catalog-e2e/src/integration/marketplace/movie-tunnel-1.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    // Connexion
+    const p1: WelcomeViewPage = new WelcomeViewPage();
+    const p2: LoginViewPage = p1.clickCallToAction();
+    p2.switchMode();
+    p2.fillSignin(LOGIN_CREDENTIALS);
+    const p3: HomePage = p2.clickSignIn();
+
+    // Navigate to movie-tunnel-1
+    cy.visit('c/o/dashboard/titles');
+    const p4 = new TitlesListPage();
```
lsenta: line 31 is strongly tied to the implementation of the titleslistpage and likely to be reused outside, keep all the knowledge INSIDE the TitlesListPage object. The URL path is not something your test-user should care about:

in your class `TitlesListPage` define a static function:
```
 public static jumpByUrl() {
  cy.visit('c/o/dashboard/titles');
  return new TitlesListPage();
}

// in your test:
const p4 = TitleListPage.jumpByUrl();
```

If you have a better name than `jumpByUrl` please use it ;) Also document the function.

apps/catalog/catalog-e2e/src/support/pages/dashboard/StartTunnelPage.ts
```diff
@@ -0,0 +1,10 @@
+export default class StartTunnelPage {
+  constructor() {
+    cy.get('catalog-start-tunnel', { timeout: 5000 });
+  }
+
+  clickBegin() {
+    cy.get('catalog-start-tunnel [test-id=begin]').click();
+    // issue#1407: Will return the movie tunnel page 2
```
lsenta: Could you implement an empty class? So that the test is correct (the page has no feature but the code is correct).

PR #1540, opened 2020-01-10
Implment new icon, update dashboard/tunnel sidenav

### LINKED ISSUE #1175: Update SVG icons for accessibility
update `<title>` and `<desc>` tags into SVG files for accessibility

### LINKED ISSUE #1444: put proper icons on the layout

Mussuranae:
Implement new icons, update dashboard sidenav, update tunnel layout icon
Fix #1444 
Fix #1175


styles/components/bf-icon.scss
```diff
@@ -0,0 +1,11 @@
+@mixin bf-icon-theme($theme) {
+  $primary: map-get($theme, primary);
+  $accent: map-get($theme, accent);
+  $warn: map-get($theme, warn);
+
+  .primary-icon {
```
GrandSchtroumpf: Best would be : 
```scss
.mat-icon {
  &[color="primary"] {
    fill: mat-color($primary, 500);
  }
  &[color="accent"] {
    fill: mat-color($accent, 500);
  }
  ...
}
```
Like that you can do : `<mat-icon color="primary">`

GrandSchtroumpf: The only risk with that is that we might think it's part of Angular material ...

PR #1541, opened 2020-01-10
Issue#1506 contract service and guards

ioaNikas:
- [x] Moved some functions out of ContractService.
- [x] Created some new guards to load the data
- Organization's Contract guard
- Active Contract guard
- Movies from Contract guard


libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -163,10 +181,63 @@ export class ContractService extends CollectionService<ContractState> {
...
+      return false;
+    }
+    const licensees = contract.parties.filter(p => p.party.role === getCodeIfExists('LEGAL_ROLES', 'licensee'))
+    const licensors = contract.parties.filter(p => p.party.role === getCodeIfExists('LEGAL_ROLES', 'licensor'))
+
+    if (!licensees.length || !licensors.length) {
+      return false;
+    }
+
+    for (const licensee of licensees) {
```
GrandSchtroumpf: A small comment above this block of code would be nice :)

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -163,10 +181,63 @@ export class ContractService extends CollectionService<ContractState> {
...
+      if (licensee.party.orgId === undefined) {
```
GrandSchtroumpf: Why not doing ` if (!licensee.party.orgId)` you want to keep it if it's empty string or null ?

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -163,10 +181,63 @@ export class ContractService extends CollectionService<ContractState> {
...
+        delete licensee.party.orgId
```
GrandSchtroumpf: It doesn't make sense to delete something that is already undefined. I think you really want to do `if (!licensee.party.orgId)`

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -163,10 +181,63 @@ export class ContractService extends CollectionService<ContractState> {
...
+      }
+      if (typeof licensee.party.showName !== 'boolean') {
+        return false;
+      }
+    }
+
+    for (const licensor of licensors) {
+      if (licensor.party.orgId === undefined) {
```
GrandSchtroumpf: Same as above

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -163,10 +181,63 @@ export class ContractService extends CollectionService<ContractState> {
...
+        delete licensor.party.orgId
+      }
+      if (typeof licensor.party.showName !== 'boolean') {
+        return false;
+      }
+    }
+
+    // Other contract validation steps goes here
+    // ...
```
GrandSchtroumpf: Maybe a todo ?

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -163,10 +181,63 @@ export class ContractService extends CollectionService<ContractState> {
...
+   * @param contract
+   */
+  public validateContract(contract: Contract): boolean {
+
+    // First, contract must have at least a licensee and a licensor
+
+    if (contract.parties.length < 2) {
+      return false;
+    }
+    const licensees = contract.parties.filter(p => p.party.role === getCodeIfExists('LEGAL_ROLES', 'licensee'))
```
GrandSchtroumpf: Why not using the `getContractParties` method defined below ?

ioaNikas: `getContractParties` getting all parties without looking at their role. Here we only need licensees and licensors. The `LEGAL_ROLES` static model has more than those two roles in it.

If we cannot have another roles than licensee and licensor as parties in contract, then the function and its content doesn't match and we need to be more precise. Gonna ask @phpgeekfr for this one.

libs/contract/src/lib/+state/contract.service.ts
ioaNikas: Since `orgId` can be undefined in some case, we must check it and delete it if so because firebase doesn't accept undefined values.

libs/contract/src/lib/+state/contract.service.ts
ioaNikas: See above

libs/contract/src/lib/+state/contract.service.ts
ioaNikas: See above

libs/contract/src/lib/+state/contract.service.ts
GrandSchtroumpf: But..... If you delete a value then it's `undefined`. It doesn't make sense AT ALL.

GrandSchtroumpf: But..... If you delete a value then it's undefined. It doesn't make sense AT ALL.

phpgeekfr: when you delete a value from an object it"s indeed undefined if you ask it but
```
{
a : 'test',
b : undefined
} 
``` 
is different from (for firebase) :
```
{
a : 'test',
}
 ```

GrandSchtroumpf: Wtf.... How is it possible ? 
And most important, how do you end up with `{ a : 'test', b : undefined }` ?
You should never do `obj[key] = undefined`.

PR #1553, opened 2020-01-13
Issue#1506 contract service crud

ioaNikas:
- [x] Add state and guard for Version subcollection (Contract SubCol)
- [x] Change DistributionDealService functions to use CollectionService built in methods.


libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -32,100 +33,14 @@ export class ContractService extends CollectionService<ContractState> {
...
   /**
    * @dev Fetch contract and last version
    */
-  // @TODO #1389 Use native akita-ng-fire functions
   public async getContractWithLastVersion(contractId: string): Promise<ContractWithLastVersion> {
     try {
       const contractWithVersion = initContractWithVersion();
-      contractWithVersion.doc = await this.getContract(contractId);
-      contractWithVersion.last = await this.getLastVersionContract(contractId);
+      contractWithVersion.doc = this.formatContract(await this.getValue(contractId));
```
GrandSchtroumpf: Create an intermediary variable : 
```typescript
const contract = await this.getValue(contractId);
...
```

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -193,54 +85,52 @@ export class ContractService extends CollectionService<ContractState> {
...
-    // Checking both licensee and licensor to clean the json and tell if contract is valid.
     for (const licensee of licensees) {
-      // Clean the json before sending it to firebase .
-      // If orgId is undefined, delete it (firebase doesn't accept undefined values).
-      if (licensee.party.orgId === undefined) {
-        delete licensee.party.orgId
-      }
-      // Check if showName is a boolean. If not, the contract is invalid and function return false.
+      // Cleaning model to remove undefined properties.
+      cleanModel(licensee);
```
GrandSchtroumpf: cleanModel is immutable. This line does nothing. You should do 
```typescript
const licensee = cleanModel(licenseeRaw);
```

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -0,0 +1,80 @@
...
+    return `contracts/${this.contractQuery.getActiveId()}/versions`;
+  }
+
+  /**
+   *
+   * @param contractId
+   * @param contractVersion
+   */
+  public async addContractVersion(contractVersion: ContractVersion): Promise<string> {
+    const contractVersions = await this.getValue();
```
GrandSchtroumpf: Use a counter + transaction for that. This is super dangerous to have this kind race condition.
Counter: 
- You can have a `__meta__` document that keep track of the count (updated by a function).
- You can have the information on the contract directly but less secure.
- You can use an extension: https://firebase.google.com/products/extensions

Transaction: 
```typescript
firestore.createTransaction(tx => {
  const { count } = tx.get(....)
  this.add(..., { write: tx })
})
```

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -0,0 +1,80 @@
...
+    await this.addContractVersion(version);
+    return contractId;
+  }
+
+  /**
+   * Returns last contract version associated with contractId
+   * @param contractId
+   */
+  public async getLastVersionContract(contractId: string): Promise<ContractVersion> {
+    const contractVersions = await this.getValue();
```
GrandSchtroumpf: You shouldn't call all the collection each time you want to get the count.

libs/contract/src/lib/version/guards/active-contract-versions.guard.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  constructor(service: ContractVersionService) {
+    super(service);
+  }
+
+  /**
+   * Sync on the active contract's versions subcollection.
+   */
+  sync() {
+    return this.service.syncCollection();
```
GrandSchtroumpf: This is the default value.

libs/movie/src/lib/distribution-deals/+state/distribution-deal.service.ts
```diff
@@ -71,40 +54,21 @@ export class DistributionDealService extends CollectionService<DistributionDealS
       // @todo #1397 change this price calculus
       contract.last.titles[movieId].price = contract.last.price;
 
-      const contractId = await this.contractService.addContractAndVersion(contract.doc, contract.last);
+      const contractId = await this.contractVersionService.addContractAndVersion(
```
GrandSchtroumpf: you should directly take the contract as parameter and separate in the service function : 
```typescript
addContractAndVersion({ doc, last }: Contract)
```

libs/movie/src/lib/distribution-deals/create/create.component.ts
```diff
@@ -199,7 +200,7 @@ export class DistributionDealCreateComponent implements OnInit, OnDestroy {
     licensor.party.role = 'licensor';
     contract.doc.parties.push(licensor);
 
-    if(!this.contractService.validateContract(contract.doc)) {
+    if(!this.contractService.isContractValid(contract.doc)) {
```
GrandSchtroumpf:
```typescript
const isValid = this.contractService.isContractValid(contract.doc)
if (!isValid) {...}
```

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -1263,7 +1263,7 @@ export class ViewExtractedElementsComponent {
           }
 
           // Checks if sale already exists
-          if (await this.distributionDealService.existingDistributionDeal(movie.id, distributionDeal)) {
+          if (await this.distributionDealService.getValue(distributionDeal.id)) {
```
GrandSchtroumpf: Create intermediary variable. It's wayyyyyyy more readable. Same for below.

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -193,54 +85,52 @@ export class ContractService extends CollectionService<ContractState> {
...
       }
     }
 
-    // Same for the licensor
     for (const licensor of licensors) {
-      if (licensor.party.orgId === undefined) {
-        delete licensor.party.orgId
-      }
+      // Cleaning model to remove undefined properties.
+      cleanModel(licensor);
```
GrandSchtroumpf: Same as above

libs/movie/src/lib/movie/components/import/table-extracted-contracts/table-extracted-contracts.component.ts
```diff
@@ -94,7 +94,7 @@ export class TableExtractedContractsComponent implements OnInit {
             hint: 'Contract already added'
           });
 
-          return promises.push(this.contractVersionService.addContractAndVersion(importState.contract.doc, importState.contract.last));
+          return promises.push(this.contractVersionService.addContractAndVersion(importState.contract));
```
GrandSchtroumpf: here you return the index of the last element. Is it correct ?

ioaNikas: Yes, index of the last element pushed into promises array.

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -32,100 +33,15 @@ export class ContractService extends CollectionService<ContractState> {
...
    */
-  // @TODO #1389 Use native akita-ng-fire functions
   public async getContractWithLastVersion(contractId: string): Promise<ContractWithLastVersion> {
     try {
       const contractWithVersion = initContractWithVersion();
-      contractWithVersion.doc = await this.getContract(contractId);
-      contractWithVersion.last = await this.getLastVersionContract(contractId);
+      const contract = await this.getValue(contractId)
+      contractWithVersion.doc = this.formatContract(contract);
+      contractWithVersion.last = await this.contractVersionService.getLastVersionContract(contractId);
```
GrandSchtroumpf: You don't need to wait to get the contract to get the version as you already have the id. You should use a Promise.all

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -138,48 +54,25 @@ export class ContractService extends CollectionService<ContractState> {
...
-          .where(`titles.${movieId}.distributionDealIds`, 'array-contains', distributionDealId)
-          .get();
-
-        if (versionSnapshot.size) {
-          const docs = versionSnapshot.docs.map(d => d.data());
-          const sortedDocs = orderBy(docs, 'id', 'desc');
-          contractWithVersion.doc = this.formatContract(contract.data());
-          contractWithVersion.last = this.formatContractVersion(sortedDocs[0]);
+      for (const contract of contracts) {
+        const contractVersions = await this.contractVersionService.getValue(ref =>
```
GrandSchtroumpf: Rarely use `await` in a loop, this is super dangerous as you're going to wait for one iteration to stop before doing the other. Use Promise.all

libs/contract/src/lib/version/+state/contract-version.model.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface ContractWithLastVersion {
+  doc: Contract,
+  last: ContractVersion,
+}
+
+export function createContractVersion(params: Partial<ContractVersion> = {}): ContractVersion {
+  return {
+    id: params.id ? params.id : '1',
+    titles: {},
+    creationDate: new Date(),
```
GrandSchtroumpf: sometime you use new Date() to initialize a date, sometimes you use null (cf Bruce). Chec with him about the best practice.

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -0,0 +1,97 @@
...
+      // Increment count and then assign it to contractVersion id.
+      contractVersion.id = (++_meta.count).toString();
+
+      // Add the version and update _meta.
+      this.add(contractVersion, { write: tx.update(_metaSnap.ref, _meta)})
+    })
+    return contractVersion.id;
+  }
+
+  onDelete() {
```
GrandSchtroumpf: put onDelete (and other hook) always at the top.

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.db.firestore.runTransaction(async tx =>  {
+      // Get the _meta document from versions subcollection.
+      const _metaSnap = await tx.get(this.db.doc(`contracts/${this.contractQuery.getActiveId()}/versions/_meta`).ref);
+      const _meta = _metaSnap.data() as VersionMeta;
```
GrandSchtroumpf: you should be able to do `getValue('_meta')`

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -0,0 +1,97 @@
...
+   * @param contractId
+   */
+  public async getLastVersionContract(contractId: string): Promise<ContractVersion> {
+    try {
+      const _metaSnap = await this.db.doc(`contracts/${contractId}/versions/_meta`).get().toPromise();
+      const { count } = _metaSnap.data() as VersionMeta;
+      const lastVersion = this.getValue(count.toString());
+
+      return lastVersion;
+    } catch (error) {
```
GrandSchtroumpf: I don't see the point of the try/catch here.

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Returns last contract version associated with contractId
+   * @param contractId
+   */
+  public async getLastVersionContract(contractId: string): Promise<ContractVersion> {
+    try {
+      const _metaSnap = await this.db.doc(`contracts/${contractId}/versions/_meta`).get().toPromise();
+      const { count } = _metaSnap.data() as VersionMeta;
```
GrandSchtroumpf: Do `getValue('_meta')`

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -0,0 +1,97 @@
...
+    } catch (error) {
+      throw new Error(error);
+    }
+  }
+
+  /**
+   *
+   * @param contractVersion
+   */
+  public formatContractVersion(contractVersion: any): ContractVersion {
```
GrandSchtroumpf: type the parameter.

GrandSchtroumpf: Why not doing that in akita's [store middleware](https://netbasal.gitbook.io/akita/general/store-middleware) ?

libs/movie/src/lib/distribution-deals/+state/distribution-deal.service.ts
```diff
@@ -119,16 +80,23 @@ export class DistributionDealService extends CollectionService<DistributionDealS
...
+    );
+    return myDeals.map(deal => this.formatDistributionDeal(deal));
   }
 
+  /**
+   * Get distributionDeals from a specific movie.
+   * @param movieId
+   */
+  public async getDistributionDeals(movieId?: string) {
+    const distributionDealsSnap = await this.db.collection(`movies/${movieId}/distributiondeals`).get().toPromise();
```
GrandSchtroumpf: Why not doing `this.get({ params: { movieId }})` ?

libs/contract/src/lib/version/+state/contract-version.service.ts
ioaNikas: If I do ```const _meta = await getValue('_meta')```, `_meta` is considered as a ContractVersion and doesn't have access to `count` property.

I could do ```export class ContractVersionService extends CollectionService<ContractVersionState | VersionMeta>``` but I have to cast the type each time I call `getValue()`

libs/movie/src/lib/distribution-deals/+state/distribution-deal.service.ts
ioaNikas: Because this function can also be called without an active movie (e.g. in search.component.ts aka Line-Up)

ioaNikas: Also, changing it to something more simple 
```
public async getDistributionDeals(movieId: string) {
    return this.getValue(`movies/${movieId}/distributiondeals`);
}
```

libs/contract/src/lib/version/+state/contract-version.model.ts
ioaNikas: Here we use `new Data()` when the version is created to keep a trace of the creation date of the document. When we initialize a date with `null`, it's because we don't know yet the date (like release date for a movie).

libs/contract/src/lib/+state/contract.service.ts
phpgeekfr: This is what we need: make interation one by one because the goal is not to fetch all contacts+their last version but to find the one contract matching our parameters. Once we found it, no need to fetch and check other documents.

Notice the `return` into the loop, once we found what we want, we break out or the `for`

libs/contract/src/lib/version/+state/contract-version.service.ts
ioaNikas: Edit : We can now use getValue('_meta') and cast it into VersionMeta.
But here we still need to have the reference of the snapshot to update it with the transaction, so it seems better to keep it this way. What do you think ?

libs/contract/src/lib/+state/contract.service.ts
ioaNikas: Is it ok if we keep it like this then ?

PR #1566, opened 2020-01-14
Issue#1518 bigquery

Hugo-Hontans:
- [x] connect to google analytics API with a backend function for a single movie
- [x] add security to the function (so that only owner of the movie can load the data)

### DIFF
--- a/apps/backend-functions/src/bigQuery.ts
+++ b/apps/backend-functions/src/bigQuery.ts
@@ -1,16 +1,18 @@
 import { CallableContext } from "firebase-functions/lib/providers/https";
 import { BigQuery } from '@google-cloud/bigquery';
 import { PublicUser, OrganizationDocument, MovieAnalytics, EventAnalytics, CallMovieAnalytics } from "./data/types";
-import { getDocument } from "./data/internals";
+import { getDocument } from './data/internals';
+import { AnalyticsEvents } from '@blockframes/utils/analytics/app-analytics';
+import { bigQueryAnalyticsTable } from "./environments/environment";
 
 function queryAddedToWishlist(movieId: string, from: number, to: number) {
   return `
   SELECT params.value.string_value AS movieId, COUNT(*) AS hits
   FROM
-    \`blockframes-hugo.analytics_195044791.events_*\`,
+    \`${bigQueryAnalyticsTable}*\`,
     UNNEST(event_params) AS params
   WHERE
-    event_name = 'added_to_wishlist'
+    event_name = '${AnalyticsEvents.addedToWishlist}'
     AND _TABLE_SUFFIX BETWEEN FORMAT_DATE("%Y%m%d", DATE_SUB(CURRENT_DATE(), INTERVAL ${from} DAY)) AND FORMAT_DATE("%Y%m%d", DATE_SUB(CURRENT_DATE(), INTERVAL ${to} DAY))
     AND params.key = 'movieId'
     AND params.value.string_value = '${movieId}'
@@ -23,10 +25,10 @@ function queryPromoReelOpened(movieId: string, from: number, to: number) {
   return `
   SELECT params.value.string_value AS movieId, COUNT(*) AS hits
   FROM
-    \`blockframes-hugo.analytics_195044791.events_*\`,
+    \`${bigQueryAnalyticsTable}*\`,
     UNNEST(event_params) AS params
   WHERE
-    event_name = 'promo_reel_opened'
+    event_name = '${AnalyticsEvents.promoReelOpened}'
     AND _TABLE_SUFFIX BETWEEN FORMAT_DATE("%Y%m%d", DATE_SUB(CURRENT_DATE(), INTERVAL ${from} DAY)) AND FORMAT_DATE("%Y%m%d", DATE_SUB(CURRENT_DATE(), INTERVAL ${to} DAY))
     AND params.key = 'movieId'
     AND params.value.string_value = '${movieId}'
@@ -39,10 +41,10 @@ function queryMovieViews(movieId: string, from: number, to: number) {
   return `
   SELECT params.value.string_value AS page_path, COUNT(*) AS hits
   FROM
-    \`blockframes-hugo.analytics_195044791.events_*\`,
+    \`${bigQueryAnalyticsTable}*\`,
     UNNEST(event_params) AS params
   WHERE
-    event_name = 'page_view'
+    event_name = '${AnalyticsEvents.pageView}'
     AND _TABLE_SUFFIX BETWEEN FORMAT_DATE("%Y%m%d", DATE_SUB(CURRENT_DATE(), INTERVAL ${from} DAY)) AND FORMAT_DATE("%Y%m%d", DATE_SUB(CURRENT_DATE(), INTERVAL ${to} DAY))
     AND params.key = 'page_path'
     AND params.value.string_value = '/c/o/marketplace/${movieId}/view'
@@ -59,7 +61,7 @@ async function executeQuery(query: any) {
 
   const options = {
     query,
-    timeoutMs: 100000, // Time out after 100 seconds.
+    timeoutMs: 100000,
     useLegacySql: false,
   };
 
--- a/apps/backend-functions/src/bigQuery.ts
+++ b/apps/backend-functions/src/bigQuery.ts
@@ -1,6 +1,6 @@
 import { CallableContext } from "firebase-functions/lib/providers/https";
 import { BigQuery } from '@google-cloud/bigquery';
-import { PublicUser, OrganizationDocument, MovieAnalytics, EventAnalytics } from "./data/types";
+import { PublicUser, OrganizationDocument, MovieAnalytics, EventAnalytics, CallMovieAnalytics } from "./data/types";
 import { getDocument } from "./data/internals";
 
 function queryAddedToWishlist(movieId: string, from: number, to: number) {
@@ -68,19 +68,19 @@ async function executeQuery(query: any) {
 }
 
 function calculatePercentage(current: number, past: number) {
-  return (current / past - 1) * 100;
+  return past ? (current / past - 1) * 100 : 0;
 }
 
 function calculateIncrease(current: EventAnalytics, past: EventAnalytics) {
-  return past ? calculatePercentage(current.hits, past.hits) : null;
+  return past ? calculatePercentage(current.hits, past.hits) : 0;
 }
 
 /** Call bigQuery with a movieId to get its analytics. */
 export const requestMovieAnalytics = async (
-  data: any,
+  data: CallMovieAnalytics,
   context: CallableContext
 ): Promise<MovieAnalytics> => {
-  const { movieId } = data;
+  const { movieId, currentPeriod, pastPeriod } = data;
   const uid = context.auth!.uid;
   const user = await getDocument<PublicUser>(`users/${uid}`);
   const org = await getDocument<OrganizationDocument>(`orgs/${user.orgId}`);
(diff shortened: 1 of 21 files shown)


apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+    timeoutMs: 100000, // Time out after 100 seconds.
+    useLegacySql: false,
+  };
+
+  // Runs the query
+  return bigqueryClient.query(options);
+}
+
+function calculatePercentage(current: number, past: number) {
+  return (current / past - 1) * 100;
```
phpgeekfr: always avoid potential division by zero ;)

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+function calculateIncrease(current: EventAnalytics, past: EventAnalytics) {
```
phpgeekfr: should it not return  zero instead of null ?

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+      executeQuery(queryPromoReelOpened(movieId, 29, 1)),
+      executeQuery(queryPromoReelOpened(movieId, 59, 30)),
+      executeQuery(queryMovieViews(movieId, 29, 1)),
+      executeQuery(queryMovieViews(movieId, 59, 30))
+    ];
+    const results = await Promise.all(promises);
+    if (results !== undefined && results.length >= 0){
+      return {
+        addedToWishlist: { ...results[0][0][0], increase: calculateIncrease(results[0][0][0], results[1][0][0]) },
+        promoReelOpened: { ...results[2][0][0], increase: calculateIncrease(results[2][0][0], results[3][0][0]) },
```
phpgeekfr: This is not pretty but since I think this BigQuery.ts file will change a lot over time, we will have time to find something more elegant

apps/backend-functions/src/data/types.ts
```diff
@@ -36,6 +36,8 @@ export {
 } from '@blockframes/permissions/types';
 export { PublicUser } from '@blockframes/auth/types';
 export { RequestDemoInformations } from '@blockframes/apps/catalog/app/landing/demo-request.model';
+export { MovieAnalytics } from "@blockframes/movie/movie+state/movie.firestore";
```
phpgeekfr: use the same import for both

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+  const { movieId } = data;
+  const uid = context.auth!.uid;
+  const user = await getDocument<PublicUser>(`users/${uid}`);
+  const org = await getDocument<OrganizationDocument>(`orgs/${user.orgId}`);
+
+  // Security: only owner of the movie can load the data
+  if (org.movieIds.includes(movieId)) {
+    // Request bigQuery
+    const promises = [
+      executeQuery(queryAddedToWishlist(movieId, 29, 1)),
```
phpgeekfr: for my point of view, date interval should be passed has parameters by the front when doing httpsOnCall. Like that if you want to change the period, no need to update functions

apps/backend-functions/src/bigQuery.ts
Hugo-Hontans: sure :)

apps/backend-functions/src/bigQuery.ts
Hugo-Hontans: Yes, it will be more scalable

apps/backend-functions/src/bigQuery.ts
Hugo-Hontans: Yes, bigQuery returns the table to me in 3 nested tables. I will see with @lsenta if I can do otherwise

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
+import { CallableContext } from "firebase-functions/lib/providers/https";
+import { BigQuery } from '@google-cloud/bigquery';
+import { PublicUser, OrganizationDocument, MovieAnalytics, EventAnalytics, CallMovieAnalytics } from "./data/types";
+import { getDocument } from "./data/internals";
+
+function queryAddedToWishlist(movieId: string, from: number, to: number) {
+  return `
+  SELECT params.value.string_value AS movieId, COUNT(*) AS hits
+  FROM
+    \`blockframes-hugo.analytics_195044791.events_*\`,
```
lsenta: Don't hardcode the table! Use a configuration in the env that you load from the `backend-functions/environment....`

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+    UNNEST(event_params) AS params
+  WHERE
+    event_name = 'added_to_wishlist'
```
lsenta: If you used an enum in the that define this event, reuse here `${EVENTS.MOVIE_ADDED_TO_WISHLIST}`

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+    AND _TABLE_SUFFIX BETWEEN FORMAT_DATE("%Y%m%d", DATE_SUB(CURRENT_DATE(), INTERVAL ${from} DAY)) AND FORMAT_DATE("%Y%m%d", DATE_SUB(CURRENT_DATE(), INTERVAL ${to} DAY))
+    AND params.key = 'movieId'
+    AND params.value.string_value = '${movieId}'
```
lsenta: This is bad practice to use string interpolation, a user could inject a funky string and change the meaning of the query,
https://fr.wikipedia.org/wiki/Injection_SQL look at the example. I could make your function spit all the data by sending a weird `from` value for exemple.

There should be a query builder that let you define params in a query and do a safe replacement for you.

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+  GROUP BY
+    movieId
+  `
+}
+
+function queryPromoReelOpened(movieId: string, from: number, to: number) {
+  return `
+  SELECT params.value.string_value AS movieId, COUNT(*) AS hits
+  FROM
+    \`blockframes-hugo.analytics_195044791.events_*\`,
```
lsenta: Don't hardcode, reuse event_name and avoid SQL Injections

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+  GROUP BY
+    movieId
+  `
+}
+
+function queryMovieViews(movieId: string, from: number, to: number) {
+  return `
+  SELECT params.value.string_value AS page_path, COUNT(*) AS hits
+  FROM
+    \`blockframes-hugo.analytics_195044791.events_*\`,
```
lsenta: Don't hardcode, reuse event_name and avoid SQL Injections

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+  `
+}
+
+async function executeQuery(query: any) {
+  // Creates a client
+  const bigqueryClient = new BigQuery();
+
+  const options = {
+    query,
+    timeoutMs: 100000, // Time out after 100 seconds.
```
lsenta: don't need useless comments that adds no information, `timeoutMs: 10000` is pretty self-explanatory, you have the meaning and the unit.

lsenta: You could `const HUNDRED_SECONDS_IN_MS = 100 * 1000` at the top of the file if you REALLY worry someone reading the code won't understand this.

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+    useLegacySql: false,
+  };
+
+  // Runs the query
+  return bigqueryClient.query(options);
+}
+
+function calculatePercentage(current: number, past: number) {
+  return past ? (current / past - 1) * 100 : 0;
```
lsenta: That looks dumb, you have Google's Query Engine, don't invent the wheel on our side.

I bet their query language provide something to do all the statistics you need, quick, precise and without code and bugs to maintain. Google or read its doc.

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,112 @@
...
+  // Security: only owner of the movie can load the data
+  if (org.movieIds.includes(movieId)) {
+    // Request bigQuery
+    const promises = [
+      executeQuery(queryAddedToWishlist(movieId, currentPeriod.from, currentPeriod.to)),
+      executeQuery(queryAddedToWishlist(movieId, pastPeriod.from, pastPeriod.to)),
+      executeQuery(queryPromoReelOpened(movieId, currentPeriod.from, currentPeriod.to)),
+      executeQuery(queryPromoReelOpened(movieId, pastPeriod.from, pastPeriod.to)),
+      executeQuery(queryMovieViews(movieId, currentPeriod.from, currentPeriod.to)),
+      executeQuery(queryMovieViews(movieId, pastPeriod.from, pastPeriod.to))
```
lsenta: I bet these 6 queries can be replaced by one.

apps/backend-functions/src/bigQuery.ts
lsenta: Similar to my previous comment, have you checked the google library?

There has to be a more idiomatic way to deal with your queries. They should provide a query builder AND a result processor.

apps/backend-functions/src/bigQuery.ts
phpgeekfr: SQL injection power

apps/backend-functions/src/bigQuery.ts
Hugo-Hontans: I can use UNION ALL to have one big query that returns an array of my 6 queries

apps/backend-functions/src/bigQuery.ts
Hugo-Hontans: I don't see how this calculation in the SQL query would be better and more maintainable.
It's more complicated or I don't understand what you want and which tool to use.

apps/backend-functions/src/environments/environment.ts
```diff
@@ -31,3 +31,6 @@ export const algolia = {
 
 export const adminEmail = mockConfigIfNeeded('admin', 'email');
 export const adminPassword = mockConfigIfNeeded('admin', 'password');
+
+// BigQuery
+export const bigQueryAnalyticsTable = 'blockframes-hugo.analytics_195044791.events_';
```
Hugo-Hontans: I'm not sure to understand here: everyone will need to put their own table.

Hugo-Hontans: I'm stupid, I have to import it from @env

apps/backend-functions/src/bigQuery.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // Security: only owner of the movie can load the data
+  if (org.movieIds.includes(movieId)) {
+    // Request bigQuery
+    const [rows] = await executeQuery(queryMovieAnalytics, movieId);
+    if (rows !== undefined && rows.length >= 0){
+      return {
+        addedToWishlist: rows.filter(row => row.event_name === AnalyticsEvents.addedToWishlist),
+        promoReelOpened: rows.filter(row => row.event_name === AnalyticsEvents.promoReelOpened),
+        movieViews: rows.filter(row => row.event_name === AnalyticsEvents.pageView)
```
Hugo-Hontans: Tt seems to me to be the simplest for the front

env/env.ci.ts
```diff
@@ -118,3 +118,8 @@ export const quorum = {
     ethAddress: '0xe795245444d459CD0d8e12A26232646B5521e72F',
   },
 }
+
+// BigQuery
+// ========
+// Change it after config BigQuery for blockframes-ci
+export const bigQueryAnalyticsTable = 'blockframes-hugo.analytics_195044791.events_';
```
lsenta: Could you configure it correctly for ci, staging, and prod so it doesn't fail when we try to release?

Hugo-Hontans: I added the right table name for each env with the analytics ID.
Can you confirm `blockframes-prod` ?

apps/backend-functions/src/bigQuery.ts
lsenta: LGTM

env/env.ci.ts
lsenta: Let's have a call and go through the configurations together

PR #1569, opened 2020-01-15
Update data model : customGenre

Mussuranae:
New customGenre field in DB

### DIFF
--- a/libs/movie/src/lib/movie/+state/movie.firestore.ts
+++ b/libs/movie/src/lib/movie/+state/movie.firestore.ts
@@ -1,4 +1,4 @@
-import { 
+import {
   MovieStatusSlug,
   PromotionalElementTypesSlug,
   ResourceRatioSlug,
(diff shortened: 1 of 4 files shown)


libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -170,6 +170,7 @@ export interface MovieOfficialIds {
 }
 
 export interface MovieMain {
+  customGenres?: string[],
```
phpgeekfr: maybe put it bellow the current `genres` attrib

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -87,8 +87,9 @@ export function createMovieMain(params: Partial<MovieMain> = {}): MovieMain {
     stakeholders: [],
     originCountries: [],
     status: '',
+    officialIds: createOfficialIds(params.officialIds),
+    customGenres: [],
     ...params,
```
phpgeekfr: keep `officialIds: createOfficialIds(params.officialIds)` after ...params please

PR #1574, opened 2020-01-15
part of Feat 1478 : move europeanQualification to certifications

phpgeekfr:


libs/movie/src/lib/movie/components/display-film-info-card/display-film-info-card.component.ts
```diff
@@ -25,7 +24,11 @@ export class MovieDisplayFilmInfoCardComponent {
   @Input() stakeholders: Company[];
   @Input() salesAgent;
 
-  public hasTheatricalRelease (originalReleases: MovieOriginalRelease[]) {
-    return originalReleases.some(r => r.media === getCodeIfExists('MEDIAS', 'theatrical'))
+  public hasTheatricalRelease () {
```
GrandSchtroumpf: Whould make sense to create a reusable pure function : 
```typescript
function hasSlug<S extends Scope, code extends ExtractSlug<S>>(array: string[], scope: Scope, key: code) {
  return array.include(getCodeIfExists(scope, key));
}
```
Cf Dajug work on types

phpgeekfr: OK will do this once Dajung PR is ok

PR #1588, opened 2020-01-16
Issue#1503 deal page query

ioaNikas:
- Add queries and update guards
- Little change on Party data model (new field : `address: Location`).

### DIFF
--- a/libs/contract/src/lib/+state/contract.model.ts
+++ b/libs/contract/src/lib/+state/contract.model.ts
@@ -7,22 +7,14 @@ import {
   ContractStatus,
   ContractTitleDetail,
   ContractPartyDetailDocumentWithDates,
-  ContractPartyDetailDocumentWithDatesDocument,
-  ContractVersionDocumentWithDates,
-  ContractDocument
+  ContractPartyDetailDocumentWithDatesDocument
 } from './contract.firestore';
 import { createParty } from '@blockframes/utils/common-interfaces/identity';
-import { Movie } from '@blockframes/movie/movie+state/movie.model';
+import { ContractVersion } from '../version/+state/contract-version.model';
 
 export interface Contract extends ContractDocumentWithDates {
   lastVersion?: ContractVersion;
 };
-
-export interface ContractVersion extends ContractVersionDocumentWithDates {
-  movies?: Movie[];
-}
-
-export type ContractVersion = ContractVersionDocumentWithDates;
 
 export type ContractPartyDetail = ContractPartyDetailDocumentWithDates;
 
@@ -42,7 +34,6 @@ export function createContract(params: Partial<Contract> = {}): Contract {
     parties: [],
     titleIds: [],
     partyIds: [],
-    lastVersionId: '',
     ...params
   };
 }
--- a/libs/contract/src/lib/+state/contract.model.ts
+++ b/libs/contract/src/lib/+state/contract.model.ts
@@ -8,10 +8,19 @@ import {
   ContractTitleDetail,
   ContractPartyDetailDocumentWithDates,
   ContractPartyDetailDocumentWithDatesDocument,
-  ContractVersionDocumentWithDates
+  ContractVersionDocumentWithDates,
+  ContractDocument
 } from './contract.firestore';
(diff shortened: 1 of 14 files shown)


libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -47,10 +48,13 @@ export {
...
   Prize,
   MovieSalesInfoDocumentWithDates as MovieSalesInfo,
-  MovieSalesAgentDealDocumentWithDates as MovieSalesAgentDeal,
-  MovieDocumentWithDates as Movie
+  MovieSalesAgentDealDocumentWithDates as MovieSalesAgentDeal
 } from './movie.firestore';
 
+export interface Movie extends MovieDocumentWithDates {
+  distributionDeals?: DistributionDeal[]
+}
```
phpgeekfr: Same comment than what you did with contract

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -29,18 +46,27 @@ export class MovieService extends CollectionService<MovieState> {
...
+      // Reset the store everytime the movieId changes.
+      tap(_ => this.store.reset()),
+      switchMap(contract => {
+        // Filter movieIds before the query to relieve it.
+        const organizationMovieIds = this.organizationQuery.getActive().movieIds;
+        const movieIds = contract.titleIds.filter(titleId => organizationMovieIds.includes(titleId));
+
+        return awaitSyncQuery.call(this, movieListContractQuery(contract, movieIds))
+      }
+      )
```
phpgeekfr: bad indent

libs/utils/src/lib/common-interfaces/identity.ts
```diff
@@ -87,6 +89,7 @@ export function createParty(params: Partial<Party> = {}): Party {
     ...createStakeholder(params),
     role: '',
     showName: false,
+    address: params.address,
```
phpgeekfr: since `address` is not mandatory on `interface Party` this line is not required since ...params will do it right after. 
If params.address is undefined, this can lead to a firebase error when saving document to db

libs/contract/src/lib/+state/contract.model.ts
```diff
@@ -7,7 +7,11 @@ import {
   ContractPartyDetailDocumentWithDatesDocument
 } from './contract.firestore';
 import { createParty } from '@blockframes/utils/common-interfaces/identity';
-export type Contract = ContractDocumentWithDates;
+import { ContractVersion } from '../version/+state/contract-version.model';
+
+export interface Contract extends ContractDocumentWithDates {
+  lastVersion?: ContractVersion;
+};
```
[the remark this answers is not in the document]
ioaNikas: Everything seems ok. Since lastVersion is an optional field, every other use case of Contract type/interface still works the same as intended. But we make sure to delete this field in the `formatToFirestore()` method before storing it again in the database, changing the object type in the same time.

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -62,13 +62,28 @@ export class ContractVersionService extends CollectionService<ContractVersionSta
...
+   * Returns last contract version id.
+   * @param contractId if not provided, get the last version id of active contract.
+   */
+  public async getLastVersionId(contractId?: string): Promise<string> {
+    const pathOrId = contractId
+      ? `contracts/${contractId}/versions/_meta`
+      : '_meta';
+    const { count } = await this.getValue(pathOrId) as VersionMeta;
+    return count.toString();
+  }
```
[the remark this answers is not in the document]
ioaNikas: Indeed. Deleting the second one instead, as the we can reach the id in the first one already.
Update the first function since `this.getValue(count.toString())` doesn't take in account the contractId and always use the active contract id.

libs/movie/src/lib/movie/+state/movie.model.ts
ioaNikas: Its the same, property is optional and doesn't interfere with old model.

Issue #1562, opened 2020-01-14
Excel Sheet updates


phpgeekfr:
Movies: 
- [x]  Original release (movie.salesInfo.originalRelease) excel should be updated to match model (country + media of originCountryRelease date)
- [ ] Once Max terminated MovieVersionInfo update, update excel import accordingly #1411 
- [x] Be able to handle various Premiere Types (world, market ...)
- [x] rename Production Companie(s) into Stake Holders
- [ ] rating: missing system, country, value [, reason] + rename to `ratings` and use ';' and ','
- [x] European Qualification (Yes or No) can be merge with certifications
- [x] Key Assets (max 3 sentences, use semicolon as separator) => this is only one sentence now
- [ ] missing country in column `Stakeholders (name, role)`
- [ ] missing logo in column `Festival Prizes (Festival name, Year, Selection/Prize, Premiere)` 
- [x] principal cast add role (currently forced to `actor`) May need to split in two columns: crew & cast (CF Anastasiia) #1603 

Contracts:
- [x]  use currency in staticModels `eur` => `euro` => not a big deal
- [x]  when setting child and parents contract Id add information about sub-licensors needed or not
- [x] for our internal use/ developers tests, we should put `OrgId` of `licensors` & `licensee` #1397 
- [x] fix circular issues contract.model & contract-version.model
- [x] check if contract import is OK

Commons:
- [ ] check `#1562` in code comments
- [ ] update `movie.main.stakeholders` after @GrandSchtroumpf  updates

PR #1598, opened 2020-01-16
[WIP] #1562  new excel movie

phpgeekfr:


libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -24,52 +26,51 @@ export class ContractVersionService extends CollectionService<ContractVersionSta
...
 
       // Add the version and update _meta.
-      this.add(contractVersion, { write: tx.update(_metaSnap.ref, _meta)})
-    })
-    return contractVersion.id;
-  }
+      if (_metaSnap.data() === undefined) {
+        await this.add(contractWithLastVersion.last, { write: tx.set(_metaSnap.ref, _meta) })
+      } else {
+        await this.add(contractWithLastVersion.last, { write: tx.update(_metaSnap.ref, _meta) });
```
ioaNikas: `this.add` doesn't seems to work as intended here. Maybe we should stick with classic tx syntax and set/update our documents the safe way

```
tx.set(_metaSnap.ref, _meta);
tx.set(version.ref, contractVersion);
```


+ You won't need to redefine `path` return value if you use tx CRUD, maybe you can get rid of it.

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -165,31 +181,45 @@ export class ContractService extends CollectionService<ContractState> {
...
-      const partiesHavingRoleForChilds = parentContract.parties.filter(p => p.childRoles && p.childRoles.length);
-      partiesHavingRoleForChilds.forEach(parentPartyDetails => {
-        parentPartyDetails.childRoles.forEach(childRole => {
-          const partyDetails = createContractPartyDetail({ party: parentPartyDetails.party });
-          partyDetails.party.role = childRole;
-          contract.parties.push(partyDetails);
-          if (partyDetails.party.orgId) {
-            contract.partyIds.push(partyDetails.party.orgId);
-          }
+      if (parentContract) {
```
ioaNikas: If you correct my mistake, you won't need to check if parentContracts exist. This happens because we call a formating function in formatToFirestore that doesn't save all the needed fields.

In contract.model.ts we got :

```
export function convertToContractDocument(params: Partial<Contract> = {}): ContractDocumentWithDates {
  return {
    id: params.id,
    parties: params.parties || [],
    titleIds: params.titleIds || [],
    partyIds: params.partyIds || []
  };
}
```

And it's missing the fields
```
    parentContractIds: params.parentContractIds || [],
    childContractIds: params.childContractIds || [],
```

Really sorry about that, just figured it out when tried to upload contracts again.

PR #1602, opened 2020-01-16
factorized some static related functions

phpgeekfr:


libs/movie/src/lib/movie/static-model/staticModels.ts
```diff
@@ -834,3 +834,13 @@ export const enum App {
...
   catalog = 'catalog'
 }
+
+/**
+ * Check if data passed into array have at lease one item belonging to scope
+ * @param array 
+ * @param scope 
+ * @param key 
+ */
+export function hasSlug<S extends Scope, code extends ExtractSlug<S>>(array: string[], scope: Scope, key: code): boolean {
```
GrandSchtroumpf: I'm pretty sure I build a function like that : `isInSlug` just above

phpgeekfr: They are different, this one check for is at least one item in a array has a specific key for a given scope.

isInSlug does not handle arrays

Issue #1603, opened 2020-01-16
updates on movie.salesCast section


phpgeekfr:
new sales cast section:
```
producers: Producer[] (interface)
cast: Cast[] (interface)
crew: Crew[] (interface)
```

`directors` stay on movie.main 
Remove `CREDIT_ROLES`
And create :
```
PRODUCER_ROLES
    {
      'slug': 'exectuive-producer',
      'label': 'Executive Producer'
    },
    {
      'slug': 'line-producer',
      'label': 'Line Producer'
    },
    {
      'slug': 'associate-producer',
      'label': 'Associate Producer'
    },
    {
      'slug': 'production-manager',
      'label': 'Production Manager'
    },
CAST_ROLES
{
      'slug': 'lead-role',
      'label': 'Lead Role'
    },
    {
      'slug': 'secondary-role',
      'label': 'Secondary Role'
    }
CREW_ROLES:
 {
      'slug': 'writer',
      'label': 'Writer'
    },    {
      'slug': 'score-composer',
      'label': 'Score Composer'
    },
    {
      'slug': 'dialogue-writer',
      'label': 'Dialogue Writer'
    },
    {
      'slug': 'director-of-photography',
      'label': 'Director of Photography'
    },
    {
      'slug': 'editor',
      'label': 'Editor'
    },
    {
      'slug': 'casting-director',
      'label': 'Casting Director'
    },
    {
      'slug': 'artistic-director',
      'label': 'Artistic Director'
    },
    {
      'slug': 'costume-designer',
      'label': 'Costume Designer'
    },
    {
      'slug': 'make-up-artist',
      'label': 'Make-Up Artist'
    },
    {
      'slug': 'production-designer',
      'label': 'Production Designer'
    },
    {
      'slug': 'first-assistant-director',
      'label': '1st Assistant Director'
    },
    {
      'slug': 'second-assistant-director',
      'label': '2nd Assistant Director'
    },
    {
      'slug': 'post-production-director',
      'label': 'Post-Production Director'
    },
    {
      'slug': 'original-author',
      'label': 'Original Author'
    }
```

Create the various entities that extends `Credit`

phpgeekfr:
![Screenshot_2020-01-16 Data Model - draw io](https://user-images.githubusercontent.com/1194344/72548176-7c4aa680-388e-11ea-8f93-1ede9fd6cda0.png)

PR #1608, opened 2020-01-17
updates on movie.salesCast section

phpgeekfr:
issue  #1603

### DIFF
--- a/libs/movie/src/lib/movie/form/sales-cast/sales-cast.form.ts
+++ b/libs/movie/src/lib/movie/form/sales-cast/sales-cast.form.ts
@@ -4,10 +4,12 @@ import { FormControl } from '@angular/forms';
 import { MovieCreditForm } from '../main/main.form';
 import { createCredit } from '@blockframes/utils/common-interfaces/identity';
 
-function createMovieSalesCastControls(salesCast?: Partial<MovieSalesCast>){
+function createMovieSalesCastControls(salesCast?: Partial<MovieSalesCast>) {
   const entity = createMovieSalesCast(salesCast);
   return {
-    credits: FormList.factory(entity.credits, el => new MovieCreditForm(el)),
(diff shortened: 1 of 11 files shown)


libs/movie/src/lib/movie/form/sales-cast/sales-cast.form.ts
```diff
@@ -35,18 +37,52 @@ export class MovieSalesCastForm extends FormEntity<MovieSalesCastControl>{
...
+    return this.get('producers');
   }
 
-  public removeCredit(i: number): void {
-    this.credits.removeAt(i);
+  get crew() {
+    return this.get('crew');
+  }
+
+  public addCredit(credit?: Partial<Credit>, type: 'cast' | 'crew' | 'producer' = 'cast'): void {
```
GrandSchtroumpf: I'm pretty sure those functions won't be used as we are going to pass the FormList directly (for next time ;) )

phpgeekfr: Yeah I know, update was made to pass the build and to keep a working app

libs/utils/src/lib/common-interfaces/identity.ts
```diff
@@ -103,3 +123,15 @@ export function createCredit(params: Partial<Credit> = {}): Credit {
     ...params
   };
 }
+
+export function createProducer(params: Partial<Credit> = {}): Producer {
```
GrandSchtroumpf: I don't really see the point of those factories. If you want to cast to type from Credit to Proucer you could do : 
```typesrcipt
export function createCredit<T extends Credit>(params: Partial<T> = {}): T
```

phpgeekfr: I'm not familiar with this `T` things, It's not automatic for me :) Do you want me to change this or you will do it yourself ?

PR #1609, opened 2020-01-17
Issue#1568 contract list query

ioaNikas:
- Update contract list query to avoid  fetching unwanted contracts.
- Add new functions to get contracts information when they are displayed as a list.
- Rename `distributiondeals` sub-collection into `distributionDeals`.
- Remove useless `@CollectionConfig` from subcollection services.

### DIFF
--- a/libs/contract/src/lib/+state/contract.service.ts
+++ b/libs/contract/src/lib/+state/contract.service.ts
@@ -12,10 +12,14 @@ import { LegalRolesSlug } from '@blockframes/movie/moviestatic-model/types';
 import { cleanModel } from '@blockframes/utils';
 import { ContractDocumentWithDates } from './contract.firestore';
 
-/** Get all the contracts where user organization is party. */
+/**
+ * Get all the contracts where user organization is party.
+ * Also check that there is no childContractIds to never fetch
+ * contract between organization and Archipel Content.
+ */
 const contractsListQuery = (orgId: string): Query<Contract[]> => ({
   path: 'contracts',
-  queryFn: ref => ref.where('partyIds', 'array-contains', orgId)
+  queryFn: ref => ref.where('partyIds', 'array-contains', orgId).where('childContractIds', '==', [])
 });
 
 /** Get the active contract and put his lastVersion in it. */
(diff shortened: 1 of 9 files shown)


libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -12,10 +12,14 @@ import { LegalRolesSlug } from '@blockframes/movie/moviestatic-model/types';
...
-/** Get all the contracts where user organization is party. */
+/**
+ * Get all the contracts where user organization is party.
+ * Also check that there is no childContractIds to never fetch
+ * contract between organization and Archipel Content.
+ */
 const contractsListQuery = (orgId: string): Query<Contract[]> => ({
   path: 'contracts',
-  queryFn: ref => ref.where('partyIds', 'array-contains', orgId)
+  queryFn: ref => ref.where('partyIds', 'array-contains', orgId).where('childContractIds', '==', [])
```
phpgeekfr: Test this please :)

libs/contract/src/lib/+state/contract.service.ts
```diff
@@ -65,22 +69,25 @@ export class ContractService extends CollectionService<ContractState> {
...
+        : contractOrId
 
       contractWithVersion.doc = this.formatContract(contract);
-      contractWithVersion.last = await this.contractVersionService.getLastVersionContract();
+      contractWithVersion.last = await this.contractVersionService.getLastVersionContract(contract.id);
 
       return contractWithVersion;
     } catch (error) {
-      console.log(`Contract ${contractId} not found`);
+      console.log(`Contract ${contractOrId} not found`);
```
phpgeekfr: is `contractOrId` is an instance of Contact, console.log will not be pretty..

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -72,4 +71,15 @@ export class ContractVersionService extends CollectionService<ContractVersionSta
 
     return lastVersion;
   }
+
+  /**
+   * Returns the creation date of the contract.
+   * @param contractId if not provided, get the last version of active contract.
+   */
+  public async getContractCreationDate(contractId?: string): Promise<Date> {
```
phpgeekfr: getContractCreationDate should be renamed to getContractFirstCreationDate I think.
Otherwise this could be confusing if we have many versions.

phpgeekfr: Or initialCreationDate

PR #1610, opened 2020-01-17
Movie form budget

GrandSchtroumpf:
Second Generation of Form Budget + move static-model.


libs/utils/src/lib/common-interfaces/range.ts
```diff
@@ -16,6 +17,7 @@ export function createRange<D>(params: Partial<RawRange<D>> = {}): RawRange<D> {
   return {
     from: null,
     to: null,
+    label: '',
```
phpgeekfr: need to initialize label even if not required (`label?`)?

libs/utils/src/lib/common-interfaces/range.ts
```diff
@@ -1,6 +1,7 @@
 export interface RawRange<D> {
   from: D | null;
   to: D | null;
+  label?: string;
```
phpgeekfr: please tell me (or Clélia) when you change models.

GrandSchtroumpf: Clélia changed that.

phpgeekfr: OK. Draw.io was not up to date.

PR #1612, opened 2020-01-17
Remove unused guard in catalog dashboard module

ioaNikas:
Quick fix on unused guard.


ioaNikas:
Will be fixed in #1614

PR #1632, opened 2020-01-21
Catalog world map

GrandSchtroumpf:
First generation of map for catalog


libs/movie/src/lib/distribution-deals/selection-map/selection-map.component.html
```diff
@@ -0,0 +1,28 @@
...
+  <mat-card class="map-card">
+    <mat-card-header></mat-card-header>
+    <mat-card-content>
+      <div #leaflet class="leaflet"></div>
+    </mat-card-content>
+  </mat-card>
+
+  <!-- Details -->
+  <mat-card class="view-card" fxLayout="column">
+    <mat-card-header></mat-card-header>
```
jenna1k: same here

libs/movie/src/lib/distribution-deals/selection-map/selection-map.component.html
```diff
@@ -0,0 +1,28 @@
+<section fxLayout="row" fxLayoutAlign="space-between stretch" fxLayoutGap="24px">
+  <!-- World Map -->
+  <mat-card class="map-card">
+    <mat-card-header></mat-card-header>
```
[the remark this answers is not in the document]
GrandSchtroumpf: there is an empty space at the top of the size of a header, and the other card will have a setting button at this place, so to have them at the same level, this is the easiest way to do that.

libs/movie/src/lib/distribution-deals/selection-map/selection-map.component.html
GrandSchtroumpf: There will be a setting button here. But indeed I should not put it now if it's not yet here.

PR #1638, opened 2020-01-21
Movie selection table

GrandSchtroumpf:
Create a reusable table component.
Apply it on titles list.
For Title List: need data from Hugo & Yohann

### DIFF
--- a/apps/catalog/catalog/src/app/dashboard/title/list/list.component.ts
+++ b/apps/catalog/catalog/src/app/dashboard/title/list/list.component.ts
@@ -1,15 +1,75 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, ChangeDetectionStrategy, OnInit } from '@angular/core';
+import { FormControl } from '@angular/forms';
+import { Movie, MovieQuery } from '@blockframes/movie';
+import { startWith, switchMap, share, map } from 'rxjs/operators';
+import { Observable, combineLatest } from 'rxjs';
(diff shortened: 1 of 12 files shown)


apps/catalog/catalog/src/app/dashboard/title/list/list.component.ts
```diff
@@ -1,15 +1,75 @@
...
+  status: 'PUBLISHED' | 'DRAFT';
+}
+
+function createTitleView(movie: Movie, contracts: Contract[]): TitleView {
+  const ownContracts = contracts.filter(c => c.lastVersion.titles[movie.id]);
+  return {
+    title: movie.main.title.international,
+    view: 'View',
+    sales: ownContracts.length,
+    receipt: ownContracts.reduce((sum, contract) => sum + contract.lastVersion.titles[movie.id].price.amount, 0),
```
ioaNikas: `contract.lastVersion` is deprecated since a few hours (sorry about that) 
https://github.com/blockframes/blockframes/blob/master/libs/contract/src/lib/%2Bstate/contract.service.ts

New query will looks like
```
const contractsListQuery = (orgId: string): Query<Contract[]> => ({
  path: 'contracts',
  queryFn: ref => ref.where('partyIds', 'array-contains', orgId).where('childContractIds', '==', []),
    versions: (contract: Contract) => ({
      path: `contracts/${contract.id}/versions`
    })
});
```

Then you can fetch the version with the most recent date as the last contract.
Keep it this way, I'll apply changes myself after rebasing.

PR #1642, opened 2020-01-22
Feat 1562  && feat 1623

phpgeekfr:


libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -33,6 +33,14 @@ export class ContractVersionService extends CollectionService<ContractVersionSta
     return `contracts/${this.contractId}/versions`;
   }
 
+  /**
+   * @dev when we are out the scope of a guard to set the contractId
+   * @param contractId 
+   */
+  public setContractId(contractId: string) {
+    this.contractId = contractId;
```
GrandSchtroumpf: Hmmm it's sound like a very bad idea to set a state inside the service. The service is just here to interact with the Backend. Consider updating the ContractVersionState instead. Else, I think you're doing something wrong.

libs/movie/src/lib/distribution-deals/+state/distribution-deal.service.ts
```diff
@@ -63,11 +63,10 @@ export class DistributionDealService extends CollectionService<DistributionDealS
       // Contract may have been updated along with the distribution deal, we update it
       await this.contractService.add(contract.doc);
       await this.contractVersionService.add(contract.last);
-      return distributionDeal.id;
     }
+    await this.db.doc(`movies/${movieId}`).collection('distributiondeals').doc(distributionDeal.id).set(distributionDeal);
```
GrandSchtroumpf: Consider `return this.add(distributionDeal, { params: { movieId }})`.
If there is an id it will use it, else it will create one. And it returns the id.

libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -34,12 +34,12 @@ export const enum WorkType {
 }
 
 export const enum StoreType {
-  catalog = 'Catalog',
+  library = 'Library',
```
GrandSchtroumpf: Did you update the catalog dashboard titles list ?

libs/movie/src/lib/movie/components/import/table-extracted-deals/table-extracted-deals.component.ts
```diff
@@ -58,53 +63,82 @@ export class TableExtractedDealsComponent implements OnInit {
...
-          return promises.push(this.distributionDealService.addDistributionDeal(movies[importState.movieInternalRef].id, importState.distributionDeal, importState.contract));
-        });
-      this.rows.data = data;
-
-      await Promise.all(promises); // @TODO #1389
-      this.snackBar.open(`${promises.length} distribution deals inserted!`, 'close', { duration: 3000 });
+      const creations = this.selection.selected.filter(importState => !hasImportErrors(importState));
+      for (const contract of creations) {
+        this.processedDeals++;
+        await this.addDeal(contract);
```
GrandSchtroumpf: I understand the idea, but you're sacrificing performance for better UX here. A Promise.all should be faster.

libs/contract/src/lib/version/+state/contract-version.service.ts
phpgeekfr: #1644

libs/movie/src/lib/distribution-deals/+state/distribution-deal.service.ts
phpgeekfr: this.add was not working well ( confirmed by yohann). went to a fallback solution

libs/movie/src/lib/movie/+state/movie.firestore.ts
phpgeekfr: sure will do

libs/movie/src/lib/movie/components/import/table-extracted-deals/table-extracted-deals.component.ts
phpgeekfr: for contracts, I need it to be like that. Can explain you by phone if you want but this is related to child/parents contracts

libs/movie/src/lib/distribution-deals/+state/distribution-deal.service.ts
GrandSchtroumpf: What doesn't work ? So I can update it.
Also consider `this.db.doc('moves/${movieId}/distributiondeals/${id}')` to make it more readable

libs/movie/src/lib/movie/components/import/table-extracted-deals/table-extracted-deals.component.ts
GrandSchtroumpf: ok

libs/movie/src/lib/distribution-deals/+state/distribution-deal.service.ts
phpgeekfr: Did not look at why it did not work in deep. @ioaNikas have the info

PR #1651, opened 2020-01-22
Feat 1388 added territories-ISO-3166

phpgeekfr:
Feat 1388 added territories-ISO-3166


libs/utils/src/lib/static-model/territories-ISO-3166.ts
```diff
@@ -0,0 +1,2099 @@
...
+    throw new Error(`Failed to retreive: ${code}.`);
+  }
+  return territory;
+}
+
+/**
+ * @param code
+ * @param system
+ */
+export function getTerritorySlugFromGeoJson(code: string, system: 'iso_a3' | 'iso_a3' = 'iso_a3'): TerritoriesSlug {
```
GrandSchtroumpf: Something like that would work I think
```typescript
type ExtractCode<S extends 'iso_a2' | 'iso_a3'> = S extends 'iso_a2' | 'iso_a3'
  ? typeof Territory[number][S]
  : never;

export function fromGeoJson<S extends 'iso_a2' | 'iso_a3', C extends ExtractCode<S>>(code: C, system: S = 'iso_a3'): { slug: TerritoriesSlug, label: TerritoriesLabel } {

}
```

PR #1654, opened 2020-01-22
Fixing permissions creation / Update services

### LINKED ISSUE #1644: remove method setContractId in contract.service.ts and find a workaround
remove method `setContractId` in `contract.service.ts` and find a workaround

ioaNikas:
- Add onCreate hooks for Movie, Delivery and Contract document => related permissions are created when a document is added to firestore.
- Update services with CollectionService methods. (fix #1644 )

### DIFF
--- a/libs/contract/src/lib/contract/+state/contract.service.ts
+++ b/libs/contract/src/lib/contract/+state/contract.service.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@angular/core';
 import { ContractStore, ContractState } from './contract.store';
-import { CollectionConfig, CollectionService, awaitSyncQuery, Query } from 'akita-ng-fire';
+import { CollectionConfig, CollectionService, awaitSyncQuery, Query, WriteOptions } from 'akita-ng-fire';
 import { Contract, ContractPartyDetail, convertToContractDocument, createContractPartyDetail, initContractWithVersion, ContractWithLastVersion } from './contract.model';
 import orderBy from 'lodash/orderBy';
 import { OrganizationQuery } from '@blockframes/organization/+state/organization.query';
@@ -11,6 +11,8 @@ import { LegalRolesSlug } from '@blockframes/utils/static-model/types';
 import { cleanModel } from '@blockframes/utils';
 import { ContractDocumentWithDates } from './contract.firestore';
 import { VersionMeta } from '@blockframes/contract/version/+state/contract-version.model';
+import { PermissionsService } from '@blockframes/organization';
+import { firestore } from 'firebase';
 
 /**
  * Get all the contracts where user organization is party.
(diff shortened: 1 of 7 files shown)


libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -67,6 +73,11 @@ export class ContractService extends CollectionService<ContractState> {
 
   }
 
+  onCreate(contract: Contract) {
+    // When a contract is created, we also create a permissions document for each parties.
+    return this.permissionsService.addContractPermissions(contract)
```
GrandSchtroumpf: Is it ok to do that on the client side ?

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -26,36 +26,21 @@ export class ContractVersionService extends CollectionService<ContractVersionSta
...
    * @param contractWithLastVersion
    */
   public async addContractVersion(contractWithLastVersion: ContractWithLastVersion): Promise<string> {
-    this.contractId = contractWithLastVersion.doc.id;
     await this.db.firestore.runTransaction(async tx => {
+      const contractId = contractWithLastVersion.doc.id;
+      const lastVersionId = contractWithLastVersion.last.id;
       // Get the _meta document from versions subcollection.
-      const _metaSnap = await tx.get(this.db.doc(`${this.path}/_meta`).ref);
+      const _metaSnap = await tx.get(this.db.doc(`contracts/${contractId}/versions/_meta`).ref);
```
GrandSchtroumpf: Why not doing `const meta = await this.getValue('_meta', { params: { contractId }})` ?

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -84,9 +66,8 @@ export class ContractVersionService extends CollectionService<ContractVersionSta
    * Returns the creation date of the contract.
    * @param contractId if not provided, get the last version of active contract.
    */
-  public async getContractInitialCreationDate(contractId?: string): Promise<Date> {
-    const documentPath = contractId ? `contracts/${contractId}/versions/` : '';
-    const firstVersion = await this.getValue(documentPath + '1');
+  public async getContractInitialCreationDate(contractId: string): Promise<Date> {
```
GrandSchtroumpf: Looks like something you want to have in the store and fetch from the query no ?

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -62,6 +65,16 @@ export class MovieService extends CollectionService<MovieState> {
     );
   }
 
+  onCreate(movie: Movie, { write }: WriteOptions) {
+    // When a movie is created, we also create a permissions document for it.
+    return this.permissionsService.addDocumentPermissions(movie, write as firestore.WriteBatch)
```
GrandSchtroumpf: Same question as above. Is it safe to do it on the client side ?

libs/contract/src/lib/contract/+state/contract.service.ts
ioaNikas: Yes we talked a lot about it with @lsenta and the conclusion was that it's better do to this way.

libs/contract/src/lib/version/+state/contract-version.service.ts
ioaNikas: Seems I can't get the ref with getValue. Can we  already do it with CollectionService or it needs to be updated ?

libs/movie/src/lib/movie/+state/movie.service.ts
ioaNikas: Same as contract permissions.

phpgeekfr:
@ioaNikas Did you tried to import excel documents to check if it is still working as intended ?

PR #1659, opened 2020-01-23
Issue#1421 story

Hugo-Hontans:
Update summary movie view with:
 - [x] Storyline Elements
 - [x] Credits 
 - [x] Budget, Quota & Critics


libs/movie/src/lib/movie/form/summary/budget/budget.component.html
```diff
@@ -0,0 +1,26 @@
+<ul>
+  <li>
+    <strong>Budget </strong>
+    <ng-container *ngIf="budget.get('estimatedBudget').value.from && budget.get('estimatedBudget').value.from">
```
GrandSchtroumpf:
```html
<ng-container *ngIf="budget.get('estimatedBudget').value as estimatedBudget>
  {{ budgetRange(estimatedBudget) }}
</ng-container>
```
In the component : 
```typescript
budgetRange({ from, to }) {
  return (from && to) ? `$ ${from} - ${to}` : '';
}
```

PR #1737, opened 2020-01-27
Issue#1421 technical information

Hugo-Hontans:
Add components to summary tunnel page:
- [x] Technical Information
- [x] Images (empty)
- [x] Files & links
- [x] Marketplace Evaluation
- [x] Correct slug and label

### DIFF
--- a/libs/movie/src/index.ts
+++ b/libs/movie/src/index.ts
@@ -21,6 +21,8 @@ export * from './lib/movie/form/summary/story/story.module';
 export * from './lib/movie/form/summary/credit/credit.module';
 export * from './lib/movie/form/summary/budget/budget.module';
 export * from './lib/movie/form/summary/technical-information/technical-information.module';
+export * from './lib/movie/form/summary/image/image.module';
+
 // Movie Components
 export { MovieViewComponent } from './lib/movie/pages/movie-view/movie-view.component';
 export { MovieEditableComponent } from './lib/movie/pages/movie-editable/movie-editable.component';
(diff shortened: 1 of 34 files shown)


libs/movie/src/index.ts
```diff
@@ -20,6 +20,11 @@ export * from './lib/movie/form/summary/information/information.module';
 export * from './lib/movie/form/summary/story/story.module';
 export * from './lib/movie/form/summary/credit/credit.module';
 export * from './lib/movie/form/summary/budget/budget.module';
+export * from './lib/movie/form/summary/technical-information/technical-information.module';
```
GrandSchtroumpf: Be careful when reexporting modules that they don't need a third party library, else you'll explode the bundle size.

libs/movie/src/lib/movie/components/display-principal-informations/display-principal-informations.component.ts
```diff
@@ -15,6 +15,6 @@ export class MovieDisplayPrincipalInformationsComponent {
   @Input() salesInfo;
 
   public hasEuropeanQualification() {
-    return this.salesInfo.certifications.some(r => r === getCodeIfExists('CERTIFICATIONS', 'europeanQualification'))
+    return this.salesInfo.certifications.some(r => r === getCodeIfExists('CERTIFICATIONS', 'european-qualification'))
```
GrandSchtroumpf: Same as Yohann, you don't want to recalculate `getCodeIfExists('CERTIFICATIONS', 'european-qualification')` for each iteration.
Best is to do : 
```typescript
const certification = getCodeIfExists('CERTIFICATIONS', 'european-qualification');
return this.salesInfo.certifications.some(c => c === certification);
```

PR #1749, opened 2020-01-28
Fixing permissions again

ioaNikas:
- Remove unused appsPermissions subcollection (until we meet again)
- Add new rule in permissions collection when creating an organization (checks that organization document doesn't exist and checks that userId will be in the organization document to be created).

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -2,7 +2,6 @@ rules_version = '2';
 service cloud.firestore {
   match /databases/{database}/documents {
 
-
     ///////////////////
     /// COLLECTIONS ///
     ///////////////////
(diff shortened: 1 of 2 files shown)


firestore.rules
```diff
@@ -233,6 +233,11 @@ service cloud.firestore {
     	return getAfter(/databases/$(database)/documents/permissions/$(orgId)/documentPermissions/$(docId)).data.canCreate == true;
     }
 
+    function isCreatingOrganization(orgId) {
```
GrandSchtroumpf: You don't have to do that in this PR, but I would like this file to be commented to understand more easily what are the rules. Also a notion article would not be too much.

PR #1758, opened 2020-01-28
Get movie analytics by months

Hugo-Hontans:
- [x] Sorts analytics events into two periods


libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -60,15 +60,24 @@ export const enum StoreStatus {
...
+  event_date: string,
   event_name: AnalyticsEvents,
   hits: number
 }
 
 export interface MovieAnalytics {
-  addedToWishlist: EventAnalytics[],
-  movieViews: EventAnalytics[],
-  promoReelOpened: EventAnalytics[]
+  addedToWishlist: {
```
lsenta: I think you could do
```
export ... {
   [key: EventType]: { current: EventAnalytics[], past: EventAnalytics[] }
}
```

apps/backend-functions/src/bigQuery.ts
```diff
@@ -41,32 +42,44 @@ async function executeQuery(query: any, movieId: string) {
...
+  return {
+    current: events.filter(event => isAfter(parseDate(event), startCurrentRange)),
+    past: events.filter(event => isBefore(parseDate(event), startCurrentRange))
+  };
+};
+
 /** Call bigQuery with a movieId to get its analytics. */
 export const requestMovieAnalytics = async (
-  data: { movieId: string },
+  data: { movieId: string, rangeDays: number },
```
lsenta: `daysPerRange`

libs/movie/src/lib/movie/+state/movie.firestore.ts
Hugo-Hontans: I don't understand how I can do that.

GrandSchtroumpf:
```typescript
type MovieAnalyticsKeys = 'addedToWishlist' | 'movieViews' | 'promoReelOpened';
export type MovieAnalytics = {
  [key in MovieAnalyticsKeys]: {
    current: EventAnalytics[],
    past: EventAnalytics[]
  }
}
```
You ave o use "type" over "interface" to use the "in" operator. Also this work only if you don't want to have optional key values (if addedToWishlist is optional but not movieViews for instance).

Bonus: If you want to reuse the keys as an array you can do that : 
```typescript
const movieAnalyticsKeys = ['addedToWishlist', 'movieViews', 'promoReelOpened'] as const;
type MovieAnalyticsKeys = typeof movieAnalyticsKeys[number];
```

GrandSchtroumpf:
All good for me

PR #1771, opened 2020-01-29
Feat 1657 - invoice / financial report related database models implementation

phpgeekfr:
Feat 1657 - invoice / financial report related database models implementation


libs/utils/src/lib/common-interfaces/terms.ts
```diff
@@ -54,12 +55,22 @@ export interface ScheduledDateRaw<D> extends TermsRaw<D> {
   period: FloatingDuration, // @dev this replace floatingDuration for better readability
 }
 
+/**
+ * A ScheduledDateRaw interface with a counter to keep schedule order
+ */
+export interface ScheduledDateWithCounterRaw<D> extends ScheduledDateRaw<D> {
+  count: 0;
```
GrandSchtroumpf: count has to be "always" 0 ??

libs/utils/src/lib/common-interfaces/terms.ts
```diff
@@ -79,3 +90,11 @@ export function createScheduledDate(params: Partial<ScheduledDate> = {}): Schedu
     period: createFloatingDuration(params ? params.period : undefined),
```
GrandSchtroumpf: Why not doing `createFloatingDuration(params.period)` directly ? I don't understand the difference.

libs/financial-report/src/lib/version/+state/financial-report-version.model.ts
```diff
@@ -0,0 +1,14 @@
...
+/** An interface for a single document to display versions subcollection count. */
+export interface VersionMeta extends FinancialReportVersion {
+  count?: number;
+}
+
+export function createVersionMeta(params: Partial<VersionMeta>): VersionMeta {
+  return {
+    id: '_meta',
+    count: params && params.count || 0,
+  } as VersionMeta;
```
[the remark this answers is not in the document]
phpgeekfr: This is intended as VersionMeta will always contains only `count`

libs/utils/src/lib/common-interfaces/terms.ts
phpgeekfr: no reason ..

PR #1777, opened 2020-01-30
can call analytics for an array of movieIds

Hugo-Hontans:
- [x] Make getMovieAnalytics backend function be usable with an array of movieIds instead of one movieId

### DIFF
--- a/apps/backend-functions/src/bigQuery.ts
+++ b/apps/backend-functions/src/bigQuery.ts
@@ -10,28 +10,30 @@ const queryMovieAnalytics = `
   SELECT
     event_name as event_name,
     COUNT(*) as hits,
-    event_date
+    event_date,
+    value.string_value as movieId,
+    REGEXP_EXTRACT(value.string_value, '.*/marketplace/([^/]+)/view') as movieIdPage
   FROM
     \`${bigQueryAnalyticsTable}*\`,
     UNNEST(event_params) AS params
   WHERE
     (
-      (event_name = @pageView AND key = 'page_path' AND REGEXP_EXTRACT(value.string_value, '.*/marketplace/([^/]+)/view') = @movieId)
+      (event_name = @pageView AND key = 'page_path' AND REGEXP_EXTRACT(value.string_value, '.*/marketplace/([^/]+)/view') in UNNEST(@movieIds))
       OR
-      (event_name = @promoReelOpened AND key = 'movieId' AND value.string_value = @movieId)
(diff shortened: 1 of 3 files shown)


apps/backend-functions/src/bigQuery.ts
```diff
@@ -61,20 +63,25 @@ const aggregateEvents = (events: EventAnalytics[], daysPerRange: number) => {
   };
 };
 
+const mergeMovieIdPageInMovieId = (rows: any[]) => {
+  return rows.map(row => row.movieIdPage ? { ...row, movieId: row.movieIdPage } : row)
```
lsenta: Sauvage,
remove the `movieIdPage` field from the result or someone, somehow, will end up rely on this, which would be a very bad id.
Add a comment.

```
import {omit} from 'lodash';
```

```
// merge the movieIdPage field from bigquery with the movieId field when relevant.
rows.map(row => omit({...row, movieId: row.movieId || row.movieIdPage, ['movieIdPage']);
```

lsenta: (missing `}` left as an exercise to the reader)

libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -62,7 +62,9 @@ export const enum StoreStatus {
 export interface EventAnalytics {
   event_date: string,
   event_name: AnalyticsEvents,
-  hits: number
+  hits: number,
+  movieId: string,
+  movieIdPage: string
```
lsenta: don't, that's a detail of implem.

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -139,6 +139,6 @@ export class MovieService extends CollectionService<MovieState> {
   /** Call a firebase function to get analytics specify to a movie.*/
   public getMovieAnalytics(): Observable<MovieAnalytics> {
     const f = this.functions.httpsCallable('getMovieAnalytics');
-    return f({ movieId: this.query.getActiveId(), daysPerRange: 28 });
+    return f({ movieId: [this.query.getActiveId()], daysPerRange: 28 });
```
lsenta: Nice, 
do you think we should limit the number of movies a user can ask in a query? (in the backend).

Hugo-Hontans: Normally the number of movies that can be called is limited by the number of movies that the organization can access.

PR #1781, opened 2020-01-30
Issue#1636 sort by movie

Hugo-Hontans:
- [x] Sorts analytics event by movieId

### DIFF
--- a/apps/backend-functions/src/bigQuery.ts
+++ b/apps/backend-functions/src/bigQuery.ts
@@ -53,20 +53,25 @@ async function executeQuery(query: any, movieIds: string[], daysPerRange: number
   return bigqueryClient.query(options);
 }
 
-/** Sorts events by movieId and into two periods. */
-const aggregateEvents = (events: EventAnalytics[], movieId: string, daysPerRange: number) => {
+/** Sorts events into two periods. */
+const groupEventsPerDayRange = (events: EventAnalytics[], daysPerRange: number) => {
   const now = new Date();
   const startCurrentRange = subDays(now, daysPerRange);
   const parseDate = (event: EventAnalytics)  => parse(event.event_date, 'yyyyMMdd', new Date());
   return {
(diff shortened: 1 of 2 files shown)


apps/backend-functions/src/bigQuery.ts
```diff
@@ -53,27 +53,28 @@ async function executeQuery(query: any, movieIds: string[], daysPerRange: number
   return bigqueryClient.query(options);
 }
 
-/** Sorts events into two periods. */
-const aggregateEvents = (events: EventAnalytics[], daysPerRange: number) => {
+/** Sorts events by movieId and into two periods. */
+const aggregateEvents = (events: EventAnalytics[], movieId: string, daysPerRange: number) => {
```
lsenta: Don't, implement TWO functions that does ONE thing rather than feeding a monstro-function::

```
groupEventsPerDayRange
groupEventsPerMovies
```

apps/backend-functions/src/bigQuery.ts
```diff
@@ -53,27 +53,28 @@ async function executeQuery(query: any, movieIds: string[], daysPerRange: number
...
-    current: events.filter(event => isAfter(parseDate(event), startCurrentRange)),
-    past: events.filter(event => isBefore(parseDate(event), startCurrentRange))
+    current: events.filter(event => isAfter(parseDate(event), startCurrentRange) && event.movieId === movieId),
+    past: events.filter(event => isBefore(parseDate(event), startCurrentRange) && event.movieId === movieId)
   };
 };
 
 /** Merge the movieIdPage field from bigquery with the movieId field when relevant. */
 const mergeMovieIdPageInMovieId = (rows: any[]) => {
+  // Remove the movieIdPage that it is not relevant to keep.
```
lsenta: "supprimer le movieIdPage qu'il n'est pas pertinent de garder"

apps/backend-functions/src/bigQuery.ts
Hugo-Hontans: I did the fix but if we use the filter function twice, I think it's less optimized.

PR #1784, opened 2020-01-30
Mainenance mode

GrandSchtroumpf:
Implement a maintenance page : 
- Listen on maintenance changes from Firestore
- Reload page when changed
- Guard the entrance on the app if in maintenance mode


libs/ui/src/lib/maintenance/maintenance.component.ts
```diff
@@ -0,0 +1,21 @@
...
+@Component({
+  selector: 'blockframes-maintenance',
+  templateUrl: './maintenance.component.html',
+  styleUrls: ['./maintenance.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class MaintenanceComponent {
+
+  // Use startWth true to avoid flash
+  isInMaintenance$ = this.service.isInMaintenance$.pipe(startWith(true));
```
Hugo-Hontans: Use public variable plz to keep the same consistency.

libs/ui/src/lib/maintenance/maintenance.component.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  constructor(private service: MaintenanceService) { }
+
+  refresh() {
```
Hugo-Hontans: same

PR #1790, opened 2020-01-31
Feat 1591 moved movie.poster to movie.promotionalElements.poster & debug import excel

phpgeekfr:


libs/financial-report/src/lib/version/+state/financial-report-version.model.ts
```diff
@@ -9,6 +9,6 @@ export interface VersionMeta extends FinancialReportVersion {
 export function createVersionMeta(params: Partial<VersionMeta>): VersionMeta {
   return {
     id: '_meta',
-    count: params.count || 0,
+    count: params && params.count ? params.count : 0,
```
phpgeekfr: old method  ` count: params.count || 0,` was breaking app when `params` itself is undefined

libs/contract/src/lib/version/+state/contract-version.model.ts
```diff
@@ -14,7 +14,7 @@ export interface VersionMeta extends ContractVersion {
 export function createVersionMeta(params: Partial<VersionMeta>): VersionMeta {
   return {
     id: '_meta',
-    count: params.count || 0,
+    count: params && params.count ? params.count : 0,
```
phpgeekfr: old method  ` count: params.count || 0,` was breaking app when `params` itself is undefined

PR #1805, opened 2020-02-03
Issue#1552 movie contract list

ioaNikas:
- Add query for contract list of an active movie.
- Use reusable deal list to display the contracts.
- Put back the isTimestamp() function for more consistency.


libs/utils/src/lib/helpers.ts
```diff
@@ -12,13 +12,19 @@ export function cleanModel<T>(data: T): T {
...
 
+/** Checks if the date is a firestore Timestamp. */
+export function isTimeStamp(date: firestore.Timestamp | Date): boolean {
+  return date && date instanceof firestore.Timestamp
+}
+
 /** Takes a Date or a Timestamp and returns it as a Date. */
-export function toDate(date: firestore.Timestamp | Date) {
-  if (date instanceof firestore.Timestamp) {
+export function toDate(date: any): Date {
```
GrandSchtroumpf: date: : firestore.Timestamp | Date

libs/utils/src/lib/helpers.ts
```diff
@@ -12,13 +12,19 @@ export function cleanModel<T>(data: T): T {
   return JSON.parse(JSON.stringify(data));
 }
 
+/** A custom interface for group of dates. Used in notifications/invitations components. */
 export interface DateGroup<T> {
   [date: string]: T[];
 }
 
+/** Checks if the date is a firestore Timestamp. */
+export function isTimeStamp(date: firestore.Timestamp | Date): boolean {
```
GrandSchtroumpf: `export function isTimeStamp(date: firestore.Timestamp | Date): date is firestore.Timestamp`

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -56,9 +66,16 @@ export class ContractService extends CollectionService<ContractState> {
   /** Sync the store with every contracts of the active organization. */
   public syncOrganizationContracts() {
     return this.organizationQuery.selectActiveId().pipe(
-      // Reset the store everytime the movieId changes.
+      switchMap(orgId => awaitSyncQuery.call(this, organizationContractsListQuery(orgId)))
```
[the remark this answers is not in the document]
ioaNikas: Since you can't change the connected user and active organization without leaving this page, I wonder if we really need to reset the store.

After testing it, we unsubscribe when we leave the segment of route where the guard is. Do you think we need to reset the store anyway to be 200% safe ?

libs/utils/src/lib/helpers.ts
ioaNikas: I'm confused, when we type date as `firestore.Timestamp | Date`, and apply `toDate()` if `isTimestamp()` returns true, an error tells me that `toDate()` is does not exist on type 'Timestamp | Date'.

If I put `date instanceof firestore.Timestamp` instead of `isTimestamp(date)`, there is no error, even it's the exact same code within another function.

What do you think is the best solution here ?

PR #1809, opened 2020-02-04
Issue#1549 film list query

ioaNikas:
- Display title list in the reusable table.
- Display a loading spinner until the data is loaded (analytics call takes some seconds).
- Some minor fix in deal view (status format and movie poster path).


libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -334,5 +338,28 @@ export function getMovieTitleList(movies: Movie[]): string[] {
...
+ * Returns the total gross receipts of a movie from the contracts.
+ * @param contracts
+ * @param movieId
+ */
+export function getMovieReceipt(contracts: Contract[], movieId: string) {
+  const currencyPipe = new CurrencyPipe('en-US');
+  const sales = getValidatedContracts(contracts);
+  const amount = sales.reduce((sum, contract) => sum + getContractLastVersion(contract).titles[movieId].price.amount, 0);
+  // We use USD as default currency as we can have different currencies for each deals and contracts.
+  return currencyPipe.transform(amount, 'USD', 'symbol');
```
GrandSchtroumpf: Why the pipe is not use in the template ???

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -334,5 +338,28 @@ export function getMovieTitleList(movies: Movie[]): string[] {
...
 
+/**
+ * Returns the number of views of a movie page.
+ * @param analytics
+ * @param movieId
+ */
+export function getMovieTotalViews(analytics: MovieAnalytics[], movieId: string): number {
+  const movieAnalytic = analytics.find(analytic => analytic.movieId === movieId);
+  const movieHits = movieAnalytic.movieViews.current.map(event => event.hits);
+  return movieHits.reduce((acc, val) => acc + val, 0);
```
GrandSchtroumpf: change "acc" to "sum" so we understand better what it does.

libs/ui/src/lib/list/table-filter/table-filter.component.scss
```diff
@@ -8,4 +8,8 @@
 
 table {
   width: 100%;
+
+  td {
```
GrandSchtroumpf: Should it be the same for all table ? or just this one.
If it's for all, we need to update the theme instead.

apps/catalog/catalog/src/app/dashboard/title/list/list.component.ts
```diff
@@ -1,42 +1,41 @@
...
-import { ContractQuery, Contract } from '@blockframes/contract/contract/+state';
-import { getContractLastVersion } from '@blockframes/contract/version/+state/contract-version.model';
-import { StoreStatus } from '@blockframes/movie/movie+state/movie.firestore';
+import { Contract } from '@blockframes/contract/contract/+state/contract.model';
+import { StoreStatus, MovieAnalytics } from '@blockframes/movie/movie+state/movie.firestore';
+import { ContractQuery } from '@blockframes/contract/contract/+state/contract.query';
+import { getContractLastVersion } from '@blockframes/contract/version/+state';
 
 interface TitleView {
+  id: string;
```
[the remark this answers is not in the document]
ioaNikas: It's a movieId. Here I can't change the parameter name since it is used by the reusable table to create a link using property `id`. I'll add a comment to specify it's an id from a movie.

libs/ui/src/lib/list/table-filter/table-filter.component.scss
ioaNikas: Mathilde told it's the same for all tables. Updating the theme with this font-size for td.

PR #1812, opened 2020-02-04
update icons

Jackseed:


libs/material/src/lib/delivery/components/delivery-new-template/new-template.component.html
```diff
@@ -16,7 +16,7 @@ <h3>Save as a new template</h3>
   <article mat-dialog-actions fxLayout="column" fxLayoutAlign="center center">
     <button
       *ngIf="!isTemplateUpdate; else: updating"
-      test-id="save"
+      test-id="floppy"
```
Hugo-Hontans: Don't change the test-id on this button if you don't update the e2e test :)

PR #1814, opened 2020-02-04
Catalog map

GrandSchtroumpf:
World Map is a reusable component with structural directives


libs/movie/src/lib/distribution-deals/selection-map/selection-map.component.html
```diff
@@ -3,26 +3,31 @@
   <mat-card class="map-card">
     <mat-card-header></mat-card-header>
     <mat-card-content>
-      <div #leaflet class="leaflet"></div>
+      <!-- <div #leaflet class="leaflet"></div> -->
+      <world-map (select)="select($event)">
+        <ng-container *ngFor="let tag of countries$ | async; trackBy: tackByTag">
```
ioaNikas: Did you mean trackByTag

PR #1815, opened 2020-02-04
With errors

lsenta:
At some point we lost errors in the front, this PR adds an error handler that make sure we log the errors!
Only in cases where there are no sentry (we'll keep sentry in prod).

- [x] figure out what's the correct way to display a toast with our frameworks

### DIFF
--- a/libs/utils/src/lib/error-logger.module.ts
+++ b/libs/utils/src/lib/error-logger.module.ts
@@ -5,15 +5,20 @@
  * This ErrorLoggerHandler is a hack to make sure we are actually logging the error.
  */
 import { ErrorHandler, Injectable, NgModule } from '@angular/core';
+import { MatSnackBar, MatSnackBarModule } from '@angular/material/snack-bar';
 
 @Injectable()
 export class ErrorLoggerHandler implements ErrorHandler {
+  constructor(private snackBar: MatSnackBar) {}
+
   handleError(error) {
     console.error(error);
+    this.snackBar.open(`${error}`.substring(0, 100));
   }
 }
 
 @NgModule({
+  imports: [MatSnackBarModule],
   providers: [
     {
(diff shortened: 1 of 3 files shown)


libs/utils/src/lib/error-logger.module.ts
```diff
@@ -0,0 +1,29 @@
...
+ * This ErrorLoggerHandler is a hack to make sure we are actually logging the error.
+ */
+import { ErrorHandler, Injectable, NgModule } from '@angular/core';
+import { MatSnackBar, MatSnackBarModule } from '@angular/material/snack-bar';
+
+@Injectable()
+export class ErrorLoggerHandler implements ErrorHandler {
+  constructor(private snackBar: MatSnackBar) {}
+
+  handleError(error) {
```
GrandSchtroumpf: Can you type it ?

libs/utils/src/lib/error-logger.module.ts
```diff
@@ -0,0 +1,29 @@
...
+    console.error(error);
+    this.snackBar.open(`${error}`.substring(0, 100));
```
GrandSchtroumpf: I think it's `error.message` to get the string out of an Error object

libs/utils/src/lib/error-logger.module.ts
lsenta: `throw 1` is valid javascript, so error will be `any`

libs/utils/src/lib/error-logger.module.ts
lsenta: Same as above, depends on the error object, I'll keep it simple for now.

libs/utils/src/lib/error-logger.module.ts
lsenta: Updated

PR #1817, opened 2020-02-04
Issue#1718 app access

Hugo-Hontans:
- [x] Create App Access component to choose an app in NoOrganizationModule
- [x] Update NoOrgGuard and create new guard for Catalog

### DIFF
--- a/libs/organization/src/lib/+state/organization.firestore.ts
+++ b/libs/organization/src/lib/+state/organization.firestore.ts
@@ -5,8 +5,8 @@ import { Location, BankAccount, createLocation } from '@blockframes/utils/common
 type Timestamp = firestore.Timestamp;
 
 interface AppAccess {
-  dashboard: boolean;
-  marketplace: boolean;
+  catalogDashboard: boolean;
+  catalogMarketplace: boolean;
 }
 
 /** Document model of an Organization */
--- a/libs/organization/src/lib/+state/organization.firestore.ts
+++ b/libs/organization/src/lib/+state/organization.firestore.ts
@@ -4,6 +4,11 @@ import { Location, BankAccount, createLocation } from '@blockframes/utils/common
 
 type Timestamp = firestore.Timestamp;
 
+interface AppAccess {
+  dashboard: boolean;
+  marketplace: boolean;
+}
+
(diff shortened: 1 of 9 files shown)


libs/organization/src/lib/+state/organization.firestore.ts
```diff
@@ -4,6 +4,11 @@ import { Location, BankAccount, createLocation } from '@blockframes/utils/common
 
 type Timestamp = firestore.Timestamp;
 
+interface AppAccess {
+  dashboard: boolean;
```
GrandSchtroumpf: catalogDashboard. As we might have more application marketplace in the future.

libs/organization/src/lib/guard/no-organization.guard.ts
```diff
@@ -21,7 +21,7 @@ export class NoOrganizationGuard {
     }
     const org = await this.service.getValue(orgId);
     return org.status === OrganizationStatus.pending
-      ? this.router.parseUrl('c/organization/congratulations')
+      ? org.appAccess ? this.router.parseUrl('c/organization/congratulations') : this.router.parseUrl('c/organization/app-access')
```
GrandSchtroumpf: This is unreadable. Use a combinaison of `if` statement and ternary condition (see guard above)

libs/organization/src/lib/no-organization.module.ts
```diff
@@ -59,6 +61,11 @@ export const noOrganizationRoutes: Routes = [
     canActivate: [NoOrganizationGuard, NoOrganizationInvitationGuard],
     component: OrganizationCreateComponent,
   },
+  {
+    path: 'app-access',
+    canActivate: [NoOrganizationInvitationGuard],
```
GrandSchtroumpf: Why you don't use the ActiveOrganizationGuard here instead of subscribing/unsubscribing in the component ?

Hugo-Hontans: We have an OrganizationGuard but it has to redirect to `/c/organization` when the user doesn't have organization, this creates an infinite redirection loop.
I used the thing we had on the congratulations page.

PR #1831, opened 2020-02-06
OK feat 1797 Contract object updates

phpgeekfr:
Contract object updates


libs/contract/src/lib/contract/+state/contract.model.ts
```diff
@@ -71,7 +71,8 @@ export function createContractVersion(params: Partial<ContractVersion> = {}): Co
     creationDate: new Date(),
     paymentSchedule: [],
     ...params,
-    status: ContractStatus.submitted,
+    status: ContractStatus.draft,
```
GrandSchtroumpf: hmmm if the status is not draft then you force it to draft here. Are you sure ?

libs/utils/src/lib/common-interfaces/terms.ts
```diff
@@ -1,23 +1,29 @@
-export enum Event {
+export enum PaymentEvent {
   ContractSignatureDate = 'Contract Signature Date',
-  WordlPremiere = 'World Premiere',
   AcceptationAllMaterials = 'Acceptation of all delivery materials',
+  InvoiceEmittedDate ='Invoice emission date',
```
GrandSchtroumpf: small note here, try to keep it consistent in case. Sometime it's camelCase, sometimes it's PascalCase...

libs/contract/src/lib/contract/+state/contract.firestore.ts
```diff
@@ -14,15 +14,28 @@ import { BankAccount } from "@blockframes/utils/common-interfaces/utility";
 type Timestamp = firestore.Timestamp;
 
 export enum ContractStatus {
```
GrandSchtroumpf: What's the rule between static-models and enum ?
Here there is 10 items, and below SUB_LICENSOR_ROLES has two items...

PR #1834, opened 2020-02-06
Reorganize pages for the tunnel in dashboard

GrandSchtroumpf:
Change the routes for the tunnel

### DIFF
--- a/libs/contract/src/lib/contract/tunnel/lobby/lobby.component.ts
+++ b/libs/contract/src/lib/contract/tunnel/lobby/lobby.component.ts
@@ -1,4 +1,6 @@
-import { Component, OnInit, ChangeDetectionStrategy } from '@angular/core';
+import { Component, ChangeDetectionStrategy } from '@angular/core';
+import { ContractService } from '../../+state';
+import { Router, ActivatedRoute } from '@angular/router';
 
 @Component({
   selector: 'contract-lobby',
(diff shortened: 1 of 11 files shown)


libs/contract/src/lib/contract/tunnel/lobby/lobby.component.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  constructor(
+    private service: ContractService,
+    private router: Router,
+    private route: ActivatedRoute,
+  ) { }
+
+  async select(type: 'sale' | 'mandate') {
+    const contractId = await this.service.add({});
+    this.router.navigate([type, contractId], { relativeTo: this.route })
```
Maxstgt: sure this worked, above you specified `/movie/${movieId}/${segment}`;`
and here you use first the type and then the contractId?
But hard make a judgment from this scope here.

libs/contract/src/lib/contract/tunnel/lobby/lobby.module.ts
```diff
@@ -0,0 +1,24 @@
...
+@NgModule({
+  declarations: [LobbyComponent],
+  imports: [
+    CommonModule,
+    ImgAssetModule,
+    FlexLayoutModule,
+    // Material
+    MatCardModule,
+    MatRippleModule,
+    RouterModule.forChild([{ path: '', component: LobbyComponent }])
```
Maxstgt: move above comment

libs/contract/src/lib/contract/tunnel/lobby/lobby.component.ts
GrandSchtroumpf: Yes, the others are in the movie tunnel, this is before the contract tunnel

PR #1845, opened 2020-02-07
Issue#1719 fix

Hugo-Hontans:
- [x] secure remove member (can't remove the last super admin)
- [x] display correctly member role


libs/organization/src/lib/components/member-repertory/member-repertory.component.ts
```diff
@@ -31,4 +32,17 @@ export class MemberRepertoryComponent {
     ? [ 'name', 'surname', 'email', 'position', 'role', 'uid' ]
     : [ 'name', 'surname', 'email', 'position', 'role' ]
   }
+
+  public displayRole(role: UserRole) {
```
GrandSchtroumpf: If we need to do that, for me, this means that this is not a good idea to use enum.

PR #1848, opened 2020-02-07
Reusable genre form

ioaNikas:
- Genre form is now reusable


libs/movie/src/lib/movie/form/summary/information/information.component.html
```diff
@@ -2,13 +2,13 @@
   <li>
     <strong>Original Languages </strong>
     <ng-container *ngFor="let language of main.originalLanguages.controls; let isLast = last">
-      <missing-control [control]="language" label="LANGUAGES" [link]="link" [isLast]="isLast"></missing-control>
+      <missing-control [control]="language" scope="LANGUAGES" [link]="link" [isLast]="isLast"></missing-control>
```
GrandSchtroumpf: You should do that on EVERY "missing-control" of EVERY page that add "label".

PR #1853, opened 2020-02-07
Add docId Index

lsenta:
Fix "a user can give themselves permission for a document owned by another org (R003)"

When a user creates document permission, we index this document (allDocPermission/{docId}) => now we have ONE location with all the documents ids that have permission attached to them.

In the rules, we forbid the creation of a permission doc if there's already a corresponding document id in the index. => document permission for an object that ALREADY exists is created by the backend as admin.

### DIFF
--- a/apps/backend-functions/src/main.ts
+++ b/apps/backend-functions/src/main.ts
@@ -34,6 +34,7 @@ import { onOrganizationCreate, onOrganizationDelete, onOrganizationUpdate } from
 import { adminApp, onRequestAccessToAppWrite } from './admin';
 import { onMovieUpdate, onMovieCreate, onMovieDelete } from './movie';
 import * as bigQuery from './bigQuery';
+import { onDocumentPermissionCreate } from './permissions';
 
 /** Trigger: when eth-events-server pushes contract events. */
 export const onIpHashEvent = functions.pubsub.topic('eth-events.ipHash').onPublish(onIpHash);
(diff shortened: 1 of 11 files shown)


apps/backend-functions/src/main.ts
```diff
@@ -118,6 +119,13 @@ export const onInvitationUpdateEvent = onDocumentWrite(
   onInvitationWrite
 );
 
+/** Trigger: when a permission document is created. */
+export const onDocumentPermissionCreateEvent = onDocumentCreate(
+  'permissions/{orgID}/documentPermissions/{docId}',
+  onDocumentPermissionCreate
+);
+
+
```
phpgeekfr: extra line break ^^

apps/backend-ops/src/firestoreMigrations/0004.ts
```diff
@@ -0,0 +1,5 @@
+import { Firestore } from '../admin';
+
+export async function upgrade(db: Firestore) {
+  // empty, to be defined by Bruce
```
phpgeekfr: coming soon

apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -30,6 +30,10 @@ export async function prepareForTesting() {
   process.exit(0);
 }
 
+export async function restoreShorcut() {
```
ioaNikas: Did you mean shorTcut ?

apps/backend-ops/src/firestoreMigrations/0005.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    // rename orgDocsPermissions
+    const orgDocPerms = await permission.ref.collection('orgDocsPermissions').get();
+    orgDocPerms.forEach(orgDocPerm => {
+      const newRef = permission.ref.collection('documentPermissions').doc(orgDocPerm.id);
+
+      const data = orgDocPerm.data();
+      batch.set(newRef, data);
+
+      const { id, owner } = data;
```
ioaNikas: Care, `owner` in orgDocsPermissions is called `ownerId` in documentPermissions. Nonetheless, I don't know if it breaks something in your code.

lsenta: so I have to rename the key `owner` to `ownerId`, are there any other changes?

PR #1877, opened 2020-02-11
LoginComponent

Hugo-Hontans:
- [x] use fragment url to open LoginComponent


libs/auth/src/lib/pages/login/login.component.ts
```diff
@@ -24,8 +24,15 @@ export class LoginComponent {
     private service: AuthService,
     private router: Router,
     private snackBar: MatSnackBar,
+    private route: ActivatedRoute
   ) {}
 
+  ngOnInit() {
+    if (this.route.snapshot.fragment === 'signin') {
```
GrandSchtroumpf: You don't need the `if`:
```typescript
this.isSignin = this.route.snapshot.fragment === 'signin'
```

PR #1885, opened 2020-02-12
Feat 1872 - public contract collection & feat-1653 debug/update import

phpgeekfr:
functions set to keep up-to-date publicContract documents 
Update on movie spreadsheet for import


apps/backend-functions/src/contract.ts
```diff
@@ -0,0 +1,141 @@
...
+ */
+async function transformContractToPublic(contract: ContractDocument, versionToSkip?: string): Promise<void> {
+  return db.runTransaction(async tx => {
+    const publicContractSnap = await tx.get(db.doc(`publicContracts/${contract.id}`));
+
+    const versionSnap = await tx.get(db.doc(`contracts/${contract.id}/versions/_meta`));
+    const versionDoc = versionSnap.data();
+    let lastVersionDoc;
+    if (!!versionDoc) {
+      const versionToFetch = !versionToSkip || parseInt(versionToSkip, 10) !== parseInt(versionDoc.count, 10) ? versionDoc.count : parseInt(versionToSkip, 10) - 1;
```
ioaNikas: Can you be more explicit or comment what you are doing here ? It's hard to understand from a noob perspective :smile:

apps/backend-functions/src/contract.ts
```diff
@@ -0,0 +1,145 @@
...
+        await tx.set(versionSnap.ref, versionDoc);
+      } else {
+        // There is no remaining contract version we delete _meta document and the corresponding publicContract will be deleted
+        console.log(`deleting _meta document for contractId : ${contract.id}`);
+        tx.delete(versionSnap.ref);
+      }
+    }
+
+    /** @dev public contract document is created only it status is OK */
+    if (lastVersionDoc && ['waitingpayment', 'paid', 'accepted'].includes(lastVersionDoc.status)) {
```
lsenta: define in a const, should be an enum in the firestore model

apps/backend-functions/src/contract.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  return db.runTransaction(async tx => {
+    await tx.delete(db.doc(`publicContracts/${contract.id}`));
+  });
+}
+
+export async function onContractVersionDelete(
+  snap: FirebaseFirestore.DocumentSnapshot,
+): Promise<any> {
+  const contractId = snap.ref.parent.parent ? snap.ref.parent.parent.id : undefined;
```
lsenta: use `const {contractId} = context.params` it matches params in your firestore ref string  (`/contract/{contractId}/etc`)

apps/backend-functions/src/contract.ts
```diff
@@ -0,0 +1,145 @@
...
+  // We retreive the version ID beeing currently removed
```
lsenta: retrieve, being

apps/backend-functions/src/main.ts
```diff
@@ -153,6 +161,59 @@ export const onMovieDeleteEvent = onDocumentDelete(
...
+//------------------------------------------------
+//   Contracts & Contracts Version Management   //
+//------------------------------------------------
+
+/**
+ * Trigger: when a contract is created
+ */
+export const onContractCreateEvent = onDocumentCreate(
+  'contracts/{contractId}',
+  onContractCreate
```
lsenta: use a single `onDocumentWrite` if you do crete, update and delete

apps/backend-functions/src/main.ts
```diff
@@ -153,6 +161,59 @@ export const onMovieDeleteEvent = onDocumentDelete(
...
+  'contracts/{contractId}',
+  onContractDelete
+)
+
+/**
+ * Trigger: when a contractVersion is created
+ */
+export const onContractVersionCreateEvent = onDocumentCreate(
+  'contracts/{contractId}/versions/{versionId}',
+  onContractVersionCreate
```
lsenta: Same, 3 functions => 1 write

apps/backend-functions/src/contract.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+    }
+
+    /** @dev public contract document is created only it status is OK */
+    if (lastVersionDoc && ['waitingpayment', 'paid', 'accepted'].includes(lastVersionDoc.status)) {
+      const publicContract: PublicContractDocument = {
+        id: contract.id,
+        type: contract.type,
+        titleIds: contract.titleIds,
+        scope: lastVersionDoc ? lastVersionDoc.scope : {},
```
lsenta: empty scope, doesn't seems to be handled by the type

apps/backend-functions/src/contract.ts
```diff
@@ -0,0 +1,145 @@
...
+  return db.runTransaction(async tx => {
+    const publicContractSnap = await tx.get(db.doc(`publicContracts/${contract.id}`));
+
+    // Fetching the current version to compare it against versionToSkip value
+    const versionSnap = await tx.get(db.doc(`contracts/${contract.id}/versions/_meta`));
+    const versionDoc = versionSnap.data();
+    let lastVersionDoc;
+    if (!!versionDoc) {
+      // If the current version is the one to skip/delete (versionToSkip), we update _meta version with versionToSkip - 1
+      const versionToFetch = !versionToSkip || parseInt(versionToSkip, 10) !== parseInt(versionDoc.count, 10) ? versionDoc.count : parseInt(versionToSkip, 10) - 1;
```
lsenta: too complex, split.

lsenta: you are relying on the fact that:
- firestore events are triggered in order (delete(x) happens BEFORE delete(x-1)) which is probably not true,
   - you'll might processed delete(x-1) => version = x - 1; then delete(x) => version = x -1 which was deleted right before.
- contract versions are removed in order

lsenta: Why do you need to store a count to get the last version?
just write the version number in the doc and query the last one. You could also simply store a created time and use that to get the last version.

Counting and assuming ids are sequential and ordered has so many requirements, you're going to hide bug everywhere in the app (back and front). It's costly and unreliable if not done correctly.

phpgeekfr: Added comments with #1887. We will stay like that until Berlin but we will work on a better solution after.

apps/backend-functions/src/contract.ts
```diff
@@ -0,0 +1,95 @@
...
+export async function onContractWrite(
+  change: functions.Change<FirebaseFirestore.DocumentSnapshot>,
+): Promise<any> {
+
+  const before = change.before.data();
+  const after = change.after.data();
+
+  if (!after && !!before) { // Deletion
+    return db.runTransaction(async tx => {
+      await tx.delete(db.doc(`publicContracts/${before.id}`));
```
lsenta: `db.doc(`publicContracts/${before.id}`).delete()` a transaction is necessary when you have write operations that depends on reads. There's only one operation here.

apps/backend-functions/src/contract.ts
```diff
@@ -0,0 +1,95 @@
...
+  const before = change.before.data();
+  const after = change.after.data();
+
+  const contractRef = db.doc(`contracts/${contractId}`);
+  const contractSnap = await contractRef.get();
+  const contract = contractSnap.data() as ContractDocument;
+
+  if (!after && !!before) { // Deletion
+    console.log(`deleting ContractVersion : "${versionId}" for : ${contractId}`);
+    return await transformContractToPublic(contract, versionId);
```
lsenta: you could just return, `await` is necessary only if you need to capture an error in the current scope (you don't have try catch here).

apps/backend-functions/src/contract.ts
phpgeekfr: Yes you are right thanks

PR #1893, opened 2020-02-13
Issue 1892

### LINKED ISSUE #1892: update currency slug in static model
update it so that we can use it for the svg icon in the mat icon tag

Maxstgt:
close #1892


libs/contract/src/lib/version/form/price/price.component.html
```diff
@@ -1,13 +1,10 @@
...
-  <mat-label>Currency</mat-label>
-  <mat-select formControlName="currency">
-    <mat-option *ngFor="let currency of staticCurrencies" [value]="currency.slug">
-      {{ currency.label }}
-    </mat-option>
-  </mat-select>
-</mat-form-field>
+<static-select scope="MOVIE_CURRENCIES" [control]="currencyCtrl">
+  <ng-template let-currency>
+   <!--  <mat-icon [svgIcon]="currency.slug" matPrefix></mat-icon> -->
```
phpgeekfr: useless template, did you forget to uncomment?

libs/contract/src/lib/version/form/price/price.component.ts
```diff
@@ -12,17 +18,123 @@ import { ContractVersionPriceForm } from './price.form';
...
+   */
+  public addMovie(event: MatAutocompleteSelectedEvent) {
+    this.tunnelComponent.addTitle(event.option.value.id)
+    this.movieCtrl.reset();
+    this.movies.push(event.option.value)
+  }
+
+  private renderCurrencySymbol(currency: string) {
+    switch (currency) {
+      case 'us-dollar':
```
phpgeekfr: I think you  are using the old slug here.

libs/contract/src/lib/version/form/price/price.component.ts
```diff
@@ -12,17 +18,123 @@ import { ContractVersionPriceForm } from './price.form';
...
+
+  /**
+   * @description helper function to transform a algolia search result into a Movie interface
+   * @param movies movies to transform
+   */
+  private async transformMovie(movies: Promise<MovieAlgoliaResult[]>): Promise<Movie[]> {
+    let movieIds: string[];
+    await movies.then(resMovies => movieIds = resMovies.map(movie => movie.objectID));
+    let moviesFromDB: Movie[];
+    await this.movieService.getValue(movieIds).then(moviesDB => moviesFromDB = moviesDB)
```
phpgeekfr: why not just do :  `const moviesFromDB =  await this.movieService.getValue(movieIds)` ?  same thing for ` let movieIds: string[];` Mayb you should not pass a `Promise` in params and resolve it before calling this method

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -2021,8 +2021,8 @@ export class ViewExtractedElementsComponent {
     }
 
     if (spreadSheetRow[SpreadSheetContractTitle.expenseCurrency + currentIndex]) {
-      if (spreadSheetRow[SpreadSheetContractTitle.expenseCurrency + currentIndex].toLowerCase() === 'eur') {
-        recoupableExpense.price.currency = 'euro';
+      if (spreadSheetRow[SpreadSheetContractTitle.expenseCurrency + currentIndex].toLowerCase() === 'EUR') {
```
phpgeekfr: this condition is not needed anymore. You test if === 'EUR', (with toLowerCase() you have poor chances to pass that..) to put 'EUR' after.
Simply:  `  recoupableExpense.price.currency  = spreadSheetRow[SpreadSheetContractTitle.expenseCurrency + currentIndex].ToUpperCase()`

The best would be to check if value actually belongs to a currency in staticIndex using the appropriate method and to throw a warning if not (like other behaviors in this file). Since this is not in the scope of your PR, if you don't want to do it now just make the minimal change and put a comment for me // @todo (BRUCE) I will do it in my next PR

libs/utils/src/lib/common-interfaces/price.ts
```diff
@@ -118,7 +118,7 @@ export interface PaymentDocument extends PaymentRaw<Timestamp> { }
  * A factory function that creates Price
  */
 export function createPrice(price: Partial<Price> = {}): Price {
-  const defaultCurrency = getCodeIfExists('MOVIE_CURRENCIES', 'euro');
```
phpgeekfr: I think you missed to replace some ` 'euro'  ` in code. see https://github.com/blockframes/blockframes/blob/issue-1892/libs/financial-report/src/lib/financial-report/%2Bstate/financial-report.model.ts

PR #1895, opened 2020-02-13
cherrypicked

Maxstgt:


libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -2021,8 +2021,9 @@ export class ViewExtractedElementsComponent {
     }
 
     if (spreadSheetRow[SpreadSheetContractTitle.expenseCurrency + currentIndex]) {
-      if (spreadSheetRow[SpreadSheetContractTitle.expenseCurrency + currentIndex].toLowerCase() === 'eur') {
-        recoupableExpense.price.currency = 'euro';
+      if (spreadSheetRow[SpreadSheetContractTitle.expenseCurrency + currentIndex].toLowerCase() === 'EUR') {
```
phpgeekfr: you could have at least removed this `.toLowerCase() === 'EUR'` ^^ remove this If.

libs/utils/src/lib/common-interfaces/price.ts
```diff
@@ -118,7 +118,7 @@ export interface PaymentDocument extends PaymentRaw<Timestamp> { }
  * A factory function that creates Price
  */
 export function createPrice(price: Partial<Price> = {}): Price {
-  const defaultCurrency = getCodeIfExists('MOVIE_CURRENCIES', 'euro');
```
phpgeekfr: I think you missed to replace some 'euro' in code. see https://github.com/blockframes/blockframes/blob/issue-1892-cherry/libs/financial-report/src/lib/financial-report/+state/financial-report.model.ts

PR #1898, opened 2020-02-13
Implements submit contract

GrandSchtroumpf:
Submit contract & deals

### DIFF
--- a/libs/contract/src/lib/contract/tunnel/summary-sale/summary-sale.component.ts
+++ b/libs/contract/src/lib/contract/tunnel/summary-sale/summary-sale.component.ts
@@ -1,15 +1,18 @@
 import { Component, OnInit, ChangeDetectionStrategy } from '@angular/core';
+import { FormControl } from '@angular/forms';
+import { AngularFirestore } from '@angular/fire/firestore';
 import { ContractForm } from '../../form/contract.form';
 import { Movie } from '@blockframes/movie';
-import { ContractTunnelComponent, DealControls } from '../contract-tunnel.component';
-import { Observable } from 'rxjs';
-import { FormEntity, TemporalityUnit } from '@blockframes/utils';
-import { FormControl } from '@angular/forms';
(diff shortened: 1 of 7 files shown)


libs/contract/src/lib/contract/tunnel/summary-sale/summary-sale.component.ts
```diff
@@ -1,15 +1,18 @@
 import { Component, OnInit, ChangeDetectionStrategy } from '@angular/core';
+import { FormControl } from '@angular/forms';
+import { AngularFirestore } from '@angular/fire/firestore';
 import { ContractForm } from '../../form/contract.form';
 import { Movie } from '@blockframes/movie';
```
Maxstgt: precise import pls

Maxstgt: should affect bundle size

libs/contract/src/lib/contract/tunnel/summary-sale/summary-sale.component.ts
```diff
@@ -58,5 +67,25 @@ export class SummarySaleComponent implements OnInit {
   }
 
   
-  submit() {}
+  /**
+   * Submit a contract version to Archipel Content
+   * @todo(#1887) should update the version on the contract
+   * @note cannot put this function on the service or you hit cyrcular dependancies
+   */
+  async submit() {
```
Maxstgt: cant find an `await`

libs/contract/src/lib/contract/tunnel/summary-sale/summary-sale.component.ts
GrandSchtroumpf: Bundle size is affected by third party libraries that are not tree-shakable (like the import library). 
So the most important is not the import but the export. You should NOT export in a barrel a component that uses a third party library. So if you import a component that uses a third party library you must have a precise link or it would mean that it hase been exported in a barrel.

For example : the world-map should not be reexported in a barrel. So when you import it you should have a precise import.

PR #1934, opened 2020-02-17
update responsive sidenav with burger

Hugo-Hontans:
- [x] Make navigation bar of marketplace responsive with burger menu


libs/ui/src/lib/tunnel/layout/layout.component.ts
```diff
@@ -35,26 +33,40 @@ function getPage(steps: TunnelStep[], url: string, arithmeticOperator: number):
...
+    private media: MediaMatcher
+  ) {}
 
   ngOnInit() {
     // Share the url by navigation plan
     this.urlBynav$ = combineLatest([this.url$, this.steps$]).pipe(shareReplay());
     this.next$ = this.urlBynav$.pipe(map(([ url, steps ]) => getPage(steps, url, 1)));
     this.previous$ = this.urlBynav$.pipe(map(([ url, steps ]) => getPage(steps, url, -1)));
+
+    this.mobileQuery = this.media.matchMedia('(max-width: 959px)');
```
GrandSchtroumpf: We are not dealing with responsive the same way in the other layouts (see dashboard layout for exemple). I'm fine with any way, but it should be constent.
Let's create a TODO issue (reference it here) to create a unified way to deal with responsive based on FlexLayout : 
- Create a MediaQuery service that match FlexLayout breakingpoint
- Create mixins that match FlexLayout breakingpoint.
- Implement it in the apps.

Hugo-Hontans: I created the issue #1935

PR #1943, opened 2020-02-17
Maintenance App

laurentsenta:
We want to put the app in maintenance for our users and keep accessing it for our admins,
this enable a permanent maintenance page AND deploy to a maintenance app.

- Missing is the ability to auto-refresh the production app is deployed. Seem's to overcomplicate for low to no value.
- Good news is we now know how to use firebase apps structure for multiple sub applications.


fritzschoff:
we got now the crm app

PR #1949, opened 2020-02-17
Issue#1921 members

Hugo-Hontans:
- [x] Change the cards width
- [x] Add the permissions explanation
- [x] Fix invite new members

### DIFF
--- a/libs/organization/src/lib/components/member-add/member-add.component.html
+++ b/libs/organization/src/lib/components/member-add/member-add.component.html
@@ -24,12 +24,7 @@
   <div fxLayout="row" fxLayoutAlign="center">
     <button class="send" mat-flat-button color="accent" test-id="add" [disabled]="isSending" (click)="sendInvitations()">
       <mat-icon svgIcon="send"></mat-icon>
-      <ng-container *ngIf="form.controls.length === 1 else several">
-        <span>Send invitation</span>
-      </ng-container>
-      <ng-template #several>
-        <span>Send invitation(s)</span>
-      </ng-template>
+      <span>Send invitation{form.controls.length, plural, =1 {} other {s}}</span>
     </button>
   </div>
 </mat-card>
--- a/libs/organization/src/lib/components/member-add/member-add.component.html
+++ b/libs/organization/src/lib/components/member-add/member-add.component.html
@@ -2,8 +2,8 @@
   <mat-card-title>Invite Members</mat-card-title>
   <p>Invite your co-workers to join your company so that they can use the application.</p>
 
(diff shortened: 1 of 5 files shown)


libs/organization/src/lib/components/member-add/member-add.component.html
```diff
@@ -12,22 +12,24 @@
           Please enter a <strong>valid</strong> email address.
         </mat-error>
       </mat-form-field>
-    </mat-card-content>
-    <mat-card-actions fxLayout="row" fxLayoutAlign="center">
       <!-- Add / Delete -->
-      <button class="mat-elevation-z0" *ngIf="!last" mat-mini-fab color="warn" (click)="form.removeAt(i)" matTooltip="Remove a Member" aria-label="Remove a Member">
+      <button class="mat-elevation-z0 action" *ngIf="!first" mat-mini-fab color="warn" (click)="form.removeAt(i)" matTooltip="Remove a Member" aria-label="Remove a Member">
```
GrandSchtroumpf: Why you cannot remove the first ? Why not `*ngIf="!last"` instead ?

libs/organization/src/lib/components/member-add/member-add.component.html
```diff
@@ -12,22 +12,24 @@
...
+    </mat-card-content>
   </ng-container>
+  <div fxLayout="row" fxLayoutAlign="center">
+    <button class="send" mat-flat-button color="accent" test-id="add" [disabled]="isSending" (click)="sendInvitations()">
+      <mat-icon svgIcon="send"></mat-icon>
+      <ng-container *ngIf="form.controls.length === 1 else several">
+        <span>Send invitation</span>
+      </ng-container>
+      <ng-template #several>
+        <span>Send invitation(s)</span>
```
GrandSchtroumpf: Why do you put () around the "s" if the whole point it to make it plural ? @Maxstgt has a way to deal with plural you can ask him. Might be a good idea to create a pipe for that if it doesn't exist yet in Angular.

libs/organization/src/lib/components/member-add/member-add.component.html
Hugo-Hontans: I don't want the delete button on the first element (you can't remove the first control).

libs/organization/src/lib/components/member-add/member-add.component.html
Hugo-Hontans: I asked to Max and we use plurial solution:
```<span>Send invitation{form.controls.length, plural, =1 {} other {s}}</span>```

PR #1961, opened 2020-02-18
Identity

Hugo-Hontans:
- [x] Fix identity page
- [x] New design
- [x] Fix terms-conditions


libs/auth/src/lib/auth.module.ts
```diff
@@ -19,6 +19,7 @@ import { MatIconModule } from '@angular/material/icon';
 import { MatBadgeModule } from '@angular/material/badge';
 import { MatTooltipModule } from '@angular/material/tooltip';
 import { MatCheckboxModule } from '@angular/material/checkbox';
+import { MatCardModule } from '@angular/material';
```
Maxstgt: precise import

libs/auth/src/lib/pages/identity/identity.component.ts
```diff
@@ -55,9 +43,8 @@ export class IdentityComponent {
       await this.service.update({
         name: this.form.get('name').value,
         surname: this.form.get('surname').value,
-        avatar: this.form.get('avatar').value
       });
-      this.router.navigateByUrl('/');
+      this.router.navigate(['/c']);
```
Maxstgt: maybe add {relativeTo: this.route}  nad in the constructor ```
private route: ActivatedSnapshot
```

File: libs/auth/src/lib/pages/identity/identity.component.html [hunk 3]
```diff
@@ -1,15 +1,8 @@
 <section fxLayout="column" fxLayoutAlign="center center">
-  <article fxLayout="row" fxLayoutAlign="center center" fxLayout.lt-md="column">
-    <img src="/assets/images/dark-rocket.png">
+  <h1>Join your organization</h1>
+  <img asset="join_orga.svg" alt="Join Organization Illustration"/>
```
Maxstgt: maybe add this icons in the icon component as well, so we can use it also as icon...?

Hugo-Hontans: No it's an illustration in svg.

PR #1968, opened 2020-02-18
Going into production

lsenta:


apps/backend-ops/src/firestoreMigrations/0008.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    if (superAdmins) {
+      superAdmins.forEach(id => (roles[id] = 'superAdmin'));
+    }
+
+    if (admins) {
+      admins.forEach(id => (roles[id] = 'admin'));
+    }
+
+    console.log('newRoles:', roles);
```
Hugo-Hontans: Do you want to let this console.log ?

PR #1970, opened 2020-02-18
resize movie-card

Hugo-Hontans:
- [x] Resize movie-cards

### DIFF
--- a/libs/ui/src/lib/movie-card/movie-card.component.scss
+++ b/libs/ui/src/lib/movie-card/movie-card.component.scss
@@ -3,12 +3,12 @@
 }
 
 img {
-  height: 230px;
+  height: 300px;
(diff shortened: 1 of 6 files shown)


libs/ui/src/lib/movie-card/movie-card.component.scss
```diff
@@ -3,14 +3,18 @@
 }
 
 img {
-  height: 230px;
+  height: 380px;
   object-fit: cover;
 }
 
 .img-container img {
   border-radius: 10px 10px 0 0;
```
jenna1k: can you remove this?
it doesn't fit to card with border-radius

Hugo-Hontans: Yes

PR #1985, opened 2020-02-19
Feat 1906 - Import excel updates - Movie

phpgeekfr:
Import excel updates - Movie


libs/utils/src/lib/spreadsheet/format.ts
```diff
@@ -24,11 +48,12 @@ export function formatCredit(str: string, separator: string = '\\s+'): any {
  * @param str 
  * @param separator 
  * @param subSeparator 
+ * @param thirdItemType
  */
-export function formatCredits(str: string, separator: string = ',', subSeparator: string = '\\s+'): any[] {
+export function formatCredits(str: string, separator: string = ',', subSeparator: string = '\\s+', thirdItemType: string = 'shortBiography'): any[] {
```
GrandSchtroumpf: Format in one argument per line to make it more readable if possible.

PR #1996, opened 2020-02-20
Issue#1806 contract notifications

### LINKED ISSUE #1806: Add New Deal Notifications & Invitations
- [x] Notification when the contract goes from status "Draft" to "Submitted":
-> to: Archipel Content only (centralOrgID)
-> message: activeOrgName submitted a contract.
-> link: document

- [x] Notification when the contract goes from status "Submitted" to "Under negotiation":
-> to: every orgs that are included in the contract parties
-> message: A new offer has been created. 
-> link: document

@Jackseed Can you also tell which logo/image to use on each type of notification ?
We can have a placeholder or use a logo/avatar.

ioaNikas:
Contracts notifications (when submitted to Archipel and when validated by Archipel)
Quick fix for deal view.

fix  #1806


apps/backend-functions/src/contract.ts
```diff
@@ -76,18 +79,57 @@ export async function onContractVersionWrite(
...
     return transformContractToPublic(contract, versionId);
-  } else {
-    return transformContractToPublic(contract);
   }
 
-}
\ No newline at end of file
+  // Prepare data to create notifications
+  const previousVersionId = parseInt(after.id, 10) - 1; // @TODO (#1887)
+  const previousVersion = await getDocument<ContractVersionDocument>(`contracts/${contractId}/versions/${previousVersionId}`)
```
phpgeekfr: I would have add an `exists()` to check if previousVersion exists

apps/backend-functions/src/contract.ts
```diff
@@ -76,18 +79,57 @@ export async function onContractVersionWrite(
...
+
+  if (contractSubmitted) { // Contract is submitted by organization to Archipel Content
+
+    const { id, name } = await getDocument<PublicOrganization>(`orgs/${contract.partyIds[0]}`); // TODO: Find real creator
+
+    const archipelContent = await getDocument<OrganizationDocument>(`orgs/${centralOrgID}`);
+    const notifications = archipelContent.userIds.map(
+      userId => createNotification({
+        userId,
+        organization: { id, name }, // TODO: Add the logo to display if orgs collection is not public to Archipel Content
```
phpgeekfr: TODO with an issue Id, it will be lost forever without a reference

apps/backend-functions/src/contract.ts
```diff
@@ -76,18 +79,57 @@ export async function onContractVersionWrite(
...
+        type: NotificationType.contractInNegotiation,
+        docId: contractId,
+        app: App.biggerBoat
+      });
+    });
+
+    await triggerNotifications(notifications);
+  }
+
+  return transformContractToPublic(contract);
```
phpgeekfr: did you check if public contracts are still created as it was before your ticket?

apps/catalog/catalog/src/app/dashboard/deal/view/view.component.html
```diff
@@ -72,7 +72,7 @@ <h3 i18n>
       <h3>Proposed Payment Schedule</h3>
       <div *ngFor="let payment of lastVersion.paymentSchedule" fxLayout="column" fxLayoutAlign="center start">
         <span>{{ payment.percentage }}% {{ payment.label }}</span>
-        <span class="mat-caption">{{ payment.date }}</span>
+        <span class="mat-caption">{{ payment.date.approxStart }}</span>
```
phpgeekfr: payment.date.approxStart can not exists, it is optional. see `export interface TermsRaw<D>`

Use termToPrettyDate to display it.

libs/contract/src/lib/contract/+state/contract.model.ts
```diff
@@ -203,7 +203,6 @@ export function createContractFromFirestore(contract: any): Contract {
   const c = {
     ...contract,
     signDate: toDate(contract.signDate),
-    titleIds: [],
     parties: contract.parties
```
phpgeekfr: why did you removed `titleIds` ? It will make some functionality (`contract.doc.titleIds.push(titleDetails.titleId);` for example) to crash because it expects that titleIds is initialized with and empty array since this field is not optional.
Please add it back.

ioaNikas: Function `createContractFromFirestore` is used to format the document into something more appropriate to be displayed. It's called in `akitaPreAddEntity` juste before we send it into the store.

Here the function is erasing the titleIds with an empty array.

The factory function to create a contract always initialize the document with empty array if needed
```
export function createContract(params: Partial<Contract> = {}): Contract {
  return {
    id: params.id || '',
    type: ContractType.mandate,
    parties: [],
    titleIds: [],
    partyIds: [],
    documents: createContractLegalDocuments(params.documents),
    ...params
  };
}
```

ioaNikas: Its a formatting function, not a factory, we don't want our titlesIds to be erased with an empty array. 

Its something we call between the moment we fetch the document from firebase and the moment we send it into the store (with `akitaPreAddEntity`)

phpgeekfr: OK my bad sorry. I confused `createContractFromFirestore`  and `createContract`

PR #1997, opened 2020-02-20
Dashboard deal list

### LINKED ISSUE #1910: Deal list - Dashboard - last fixes
## UI 
[Zeplin](https://app.zeplin.io/project/5ca741b3ed21e79faa266664/screen/5e1f233a1ffb15bd95efb518)
- not updated with all requested changes - 

## Updates
- [x] Wait for #1948, if not possible, remove the "Territories" column from the table
- [ ] Change the background color for filter bar, filtering options & table rows 
- [x] Remove the cents from price (.00)
- [x] Display value for contract status, not the key and add a capital letter to the status ("Undernegotiation" > "Under negotiation") @phpgeekfr 
- [ ] The filter bar doesn't filter Buyer names, either make it works or remove it
- [x] Sorting columns don't seem to work, same either fix it or remove it
- [x] Remove the "Past Deals" tab and add contract with status "paid" to "Ongoing Deals" tab
- [x] Fix the filtering from tabs, for now all contracts are displayed in "All" but not in the good tabs (Offers, Ongoing Deals,...)
- [ ] Add a tab "Drafts" for contracts:
- display contracts with "draft" and "submitted" statuses on the last version
- only where the org is owner of the contract
- contracts must be displayed in the list even if there is only creation date
- each rows leads to the creation tunnel loaded with the draft contract
- when the contract is submitted, the form is disabled 
- contract draft link to the tunnel; every other status link to the view
- [ ] Add a "+" button that links to the contract tunnel > sales mode

After Berlin:
- [ ] Dashboard deal list: Add a button "Archipel Mandate":
- links to the mandate tunnel if the mandate is status "draft" or if there is no mandate
- mandate tunnel is hardcoded with Archipel Content in org name + licensee role
- links to the Archipel view page otherwise, blocked by #1913

GrandSchtroumpf:
fix #1910 
Update the deal list
Create contract-table shared by dashboard & marketplace.
Use bf-table for contract-table
Create output "rowClick" on bf-table.
Try to update the style, but failed =="

### DIFF
--- a/libs/contract/src/lib/contract/+state/contract.query.ts
+++ b/libs/contract/src/lib/contract/+state/contract.query.ts
@@ -4,9 +4,15 @@ import { ContractStore, ContractState } from './contract.store';
 import { map, shareReplay, distinctUntilChanged, pluck, switchMap } from 'rxjs/operators';
 import { getVersonView } from './contract.utils';
 import { MovieQuery } from '@blockframes/movie';
+import { ContractType } from './contract.firestore';
+import { Contract } from './contract.model';
 
 @Injectable({ providedIn: 'root' })
 export class ContractQuery extends QueryEntity<ContractState> {
+
+  /** There is only one mandate per organization */
+  mandate$ = this.selectEntity((contract: Contract) => contract.type === ContractType.mandate);
+  sales$ = this.selectAll({ filterBy: contract => contract.type === ContractType.sale });
 
   // @todo(#1887) 
   // don't look for the last version + move it to version query
@@ -36,6 +42,10 @@ export class ContractQuery extends QueryEntity<ContractState> {
 
   constructor(protected store: ContractStore, private movieQuery: MovieQuery) {
     super(store);
+  }
+
+  getMandate() {
+    this.getEntity((contract: Contract) => contract.type === ContractType.mandate)
   }
 
 }
--- a/libs/contract/src/lib/contract/+state/contract.service.ts
+++ b/libs/contract/src/lib/contract/+state/contract.service.ts
@@ -10,7 +10,10 @@ import {
   getContractParties,
   createContractFromFirestore,
   cleanContract,
-  PublicContract
+  PublicContract,
+  createContract,
+  createContractVersion,
+  createVersionMandate
 } from './contract.model';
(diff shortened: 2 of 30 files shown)


libs/contract/src/lib/contract/+state/contract.query.ts
```diff
@@ -38,4 +44,8 @@ export class ContractQuery extends QueryEntity<ContractState> {
     super(store);
   }
 
+  getMandate() {
+    this.getEntity((contract: Contract) => contract.type === ContractType.mandate)
+  }
```
ioaNikas: What is we got more than 1 mandate ?

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -94,6 +80,18 @@ export class ContractService extends CollectionService<ContractState> {
...
+  /////////
+  // GET //
+  /////////
+
+  /** Get the mandate contract of an organization */
+  public async getMandate(orgId: string) {
+    const query = ref => ref.where('partyIds', 'array-contains', orgId).where('type', '==', ContractType.sale);
+    const mandates = await this.getValue(query);
+    return (mandates && mandates.length) ? mandates[0] : undefined;
+  }
```
ioaNikas: There is a function with the same name in query, are they both used ?

libs/contract/src/lib/contract/guards/organization-contract-list.guard.ts
```diff
@@ -1,19 +1,43 @@
...
-import { ContractService, ContractState } from '../+state';
+import { CollectionGuard, CollectionGuardConfig, awaitSyncQuery, Query } from 'akita-ng-fire';
+import { ContractService, ContractState, ContractStore, ContractWithTimeStamp, ContractType } from '../+state';
+import { OrganizationQuery } from '@blockframes/organization';
+import { tap, switchMap } from 'rxjs/operators';
+
+/**
+ * @todo(#1887) Don't import all versions
+ * Get all the contracts where user organization is party.
+ * Also check that there is no childContractIds to never fetch
```
ioaNikas: Comment need an update

Issue #1926, opened 2020-02-16
Movie Tunnel - last fixes


Jackseed:
## UI 
[Zeplin](https://app.zeplin.io/project/5ca741b3ed21e79faa266664/screen/5e0b1449a2ab6098620fecef)

## Updates
- [x] Add a white background everywhere
- [x] When click on save: add a snackbar "Title saved"
- [x] **Close and save button erase information instead of saving** @GrandSchtroumpf 
- [x] Submit to Archipel Content button raise an error while every mandatory fields are filled @GrandSchtroumpf 
![image](https://user-images.githubusercontent.com/14964443/74612726-5d973400-5108-11ea-8217-66e1de29d14d.png)

### Lobby page
- [x] Remove "Files you will need" + text Poster(s), banner image, still photos and the promo reel. You will also need to provide the links to the screener and trailer .
- [x] Change title to "Add a Title" (= remove "and Licensed Rights")
- [x] Change subtitle to :
The 3-step process will take about a half-hour to complete.
- [x] Remove the text upside begin "You can fill it out in one go or over time. You may navigate between steps, but all sections must be completed before submitting to Archipel Content for validation." (it's already written in instructions)
- [x] Remove the card "Licensed Rights" and ask Mathilde for new illustrations
- [x] Remove the text from the cards and change "Upload Media" to "Media"
- [x] Align the 3 cards same as UI
- [x] Add Archipel Content logo on the navbar

### Sidenav
- [x] Rename Main informations to "Main Information" (without s)
- [x] Remove Licensed Rights section
- [x] Rename Uploaded Media to "Media"
- [x] Rename Technical info to "Technical Information" and move it to the Media section on first position
- [x] Rename Synopsis to "Storyline Elements"
- [x] Remove Keywords sub-section
- [x] Rename "Budg., quota, critics" to "Budget, Quotas, Critics"
- [x] Rename Images to "Promotional Images"

### Navbar
- [x] Change logo color to blue
- [x] On title info section, change steps for x/4 and time for "15 min"
- [x] On media section, change steps for x/3 and time for "10 min"
- [x] On legal info section, change steps for x/2 and time for "5 min"
- [x] When the cross top right is clicked and there is no previous page (after a refresh for example), the cross doesn't work, it should lead to the active movie page tab details instead 

### Title information section
_Main info_
- [x] Freshness: use the store type Enum (Library instead of Catalog)
[...]

PR #1998, opened 2020-02-20
some fixes for tunnels

Maxstgt:
ref #1926


apps/catalog/catalog/src/app/dashboard/movie-tunnel/start/start-tunnel.component.html
```diff
@@ -31,7 +31,7 @@ <h2>The 3-step process will take about a half-hour to complete.</h2>
       <ng-container *ngFor="let content of cardContents">
         <mat-card>
           <article fxLayout="column" fxLayoutAlign="center">
-            <img class="card-image" src="http://via.placeholder.com/100x100" alt="Placeholder image">
+            <img class="card-image" [asset]="content.img"  alt="Archipel Content Logo">
```
GrandSchtroumpf: what is .img extension ?

apps/catalog/catalog/src/app/guards/catalog-contract.guard.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Injectable } from '@angular/core';
+import { ContractStore } from '@blockframes/contract';
+
+@Injectable({ providedIn: 'root' })
+export class CatalogContractGuard implements CanActivate {
+
+    constructor(private contractStore: ContractStore, private router: Router) { }
+
+    async canActivate(next: ActivatedRouteSnapshot) {
+        const contract = this.contractStore.getValue().entities[next.paramMap.get('contractId')];
```
GrandSchtroumpf: Don't use the store to access the state. Prefer the query, it's optimized for that.
```typescript
const { contractId } = next.params;
const contract = this.contractQuery.getEntity(contractId);
```

apps/catalog/catalog/src/app/guards/catalog-contract.guard.ts
```diff
@@ -0,0 +1,16 @@
...
+        const lastVersion = getContractLastVersion(contract)
+        return lastVersion.status === 'draft' ? this.router.parseUrl('c/o/dashboard/deals/list') : true;
```
GrandSchtroumpf: Don't hardcode the "dashboard" (else it won't work on the marketplace). If possible use relative path, else use akita routerQuery to manipulate the route as expected.

libs/contract/src/lib/contract/tunnel/contract-tunnel.component.ts
```diff
@@ -63,6 +63,7 @@ export class ContractTunnelComponent implements OnInit {
   private removedDeals: Record<string, string[]> = {};
   public steps$: Observable<TunnelStep[]>;
   public type: ContractType;
+  public exitRoute: Observable<string>;
```
GrandSchtroumpf: Out of convention use "$" at the end when dealing with an observable

libs/contract/src/lib/contract/tunnel/contract-tunnel.component.ts
```diff
@@ -116,19 +129,21 @@ export class ContractTunnelComponent implements OnInit {
...
     for (let i = deals.length - 1; i >= 0; i--) {
       this.removeDeal(movieId, i);
     }
-    this.dealForms.removeControl(movieId);
-    const dealIds = Object.keys(this.dealForms.controls)
-    if (!dealIds.length) {
-      this.router.navigate(['details'], { relativeTo: this.route })
-    } else {
-      this.router.navigate([dealIds[dealIds.length - 1]], { relativeTo: this.route })
+    if (isExploitRight) {
```
GrandSchtroumpf: Add a comment to explain why you do that.

libs/contract/src/lib/contract/tunnel/contract-tunnel.component.ts
```diff
@@ -157,7 +172,7 @@ export class ContractTunnelComponent implements OnInit {
     // Upate Version
     const lastIndex = contract.versions.length - 1;
     const version = createContractVersion({ ...contract.versions[lastIndex] });
-    this.versionService.update({ id: `${lastIndex}`, ...version }, { params: { contractId }, write })
+    this.versionService.update({ id: `${lastIndex} `, ...version }, { params: { contractId }, write })
```
GrandSchtroumpf: NO. You're hadding a whitespace in the id. Don't do that !

libs/contract/src/lib/version/form/price/price.component.ts
```diff
@@ -111,6 +110,10 @@ export class PriceComponent implements OnInit {
     return this.form.last().get('titles').get(movieId).get('price');
   }
 
+  public getMovie(movieId: string) {
```
GrandSchtroumpf: keep the akita synthax to make sure we understand this is an observable : `selectMovie(movieId: string)`

libs/contract/src/lib/contract/tunnel/contract-tunnel.component.ts
Maxstgt: hm thats bad. Why not do lastIndex.toString()

libs/organization/src/lib/components/organization-autocomplete/organization-autocomplete.component.ts
```diff
@@ -49,7 +49,7 @@ export class OrganizationAutocompleteComponent implements OnInit {
        * If the user change the org name to a non existing org,
        * we want to erase the ID fromt he form
        */
-      this.emitSelect.emit('')
+      this.emitSelect.emit('remove')
```
GrandSchtroumpf: Be carefull with magic string like that. You cannot do if($event) for example. Also as there is no documentation, this might create a bug in the futur difficult to debug.

apps/catalog/catalog/src/app/dashboard/movie-tunnel/start/start-tunnel.component.html
Maxstgt: its not an extension, its a key

apps/catalog/catalog/src/app/guards/catalog-contract.guard.ts
```diff
@@ -1,16 +1,24 @@
...
+    constructor(private contractQuery: ContractQuery, private routerQuery: RouterQuery, private router: Router) { }
 
     async canActivate(next: ActivatedRouteSnapshot) {
-        const contract = this.contractStore.getValue().entities[next.paramMap.get('contractId')];
+        const { contractId } = next.params
+        const contract = this.contractQuery.getValue().entities[contractId];
         const lastVersion = getContractLastVersion(contract)
-        return lastVersion.status === 'draft' ? this.router.parseUrl('c/o/dashboard/deals/list') : true;
+        if (lastVersion.status === 'draft') {
+            const splittedRoute = this.routerQuery.getValue().state.url.split('/');
```
GrandSchtroumpf: Add a comment to exlplain why you do that. With an example of value because it's difficult to understand what happen. I understand : 
```typescript
const path = this.routerQuery.getValue().state.url.split('/');
path.shift().pop().pop();
this.router.navigate([path.oin('/')]);
```
But it doesn't make so much sense

PR #2021, opened 2020-02-20
Region chip autocomplete

ioaNikas:
Add region chips autocomplete component.


apps/backend-functions/src/contract.ts
```diff
@@ -86,8 +88,29 @@ export async function onContractVersionWrite(
...
+
+    const notifications = organizations
+    .filter(organizationDocument => !!organizationDocument && !!organizationDocument.userIds)
+    .reduce((ids: string[], { userIds }) => [...ids, ...userIds], [])
+    .map(userId => {
+      return createNotification({
+        userId,
+        type: NotificationType.contractInNegotiation,
+        docId: contractId,
+        app: App.biggerBoat
```
Hugo-Hontans: biggerBoat sure ?

ioaNikas: Yes, biggerBoat = catalog as a string. I don't wanna change that for now because it could break somewhere else just before the release

Issue #1887, opened 2020-02-12
Update version model


GrandSchtroumpf:
This issue is here to reference all piece of code that need to be changed if we put versions inside the contract documents. And discuss how to update the model.

Initial thought: 
we can have a "archive" sub-collection for immutables versions, but at least the last version should be on the document.

---------------------------------

Todos:

- [x]   When a new contract is created, licensee can be set automatically (but can be overrided) with auth.user.orgId. => use akita-ng-fire for this
- [x]   Update front code
- [x]   When a new contract of type mandate is created/updated, the tiltleDetails.price.commissionBase must be set to grossreceipt.
- [x]   When a new contract of type sales is created/updated, the parties can be set automatically with the contract of type mandate of the titles included in contract. The contractParents or the current contract and the contractChild of the contract parent can be updated (only if != draft?).
- [x]   Keep the existing publicContract creation process
- [x]   Add titlesIds to contract.titleIds[] automatically when contract.titles is updated.
- [x]   Add partieIds to contract.partieIds[] automatically when contract.parties is updated and when contract status is not Draft and when parties have an orgId defined.
- [x]   Clean contract signDate, parties signDate and status when a new version is created.
- [x]   don't forget to document everything!
- [x]   check @todo (#1887) in code
- [x]   do we need to create separate collection for each contract type ?
- [x]  ContractVersion are archived only under certain circumstances, for example, contract status must be !== Draft to avoid polluting database.
- [x] Migration script :  moved to #2335

phpgeekfr:
We will have the same kind of issues with `FinancialReport` and `FinancialReportVersion`. 

Added comments `@TODO (#1887)` in code to identify problems.

GrandSchtroumpf:
Here are my arguments about the contract version : 

_If we want all versions on the frontend + frontend can update all versions_ 
There is not reason to have a subcollection. And we just make the code harder to build/maintain. Then we should put the versions on the contract document.

_If we want only the last version on the frontend + frontend cannot update a version_
There is a good reason to have a subcollection. And we need to improve the way we use it : 
- option 1: Have a mutable "version" key on the "contract" document and add it to the "versions" subcollection when we want to make it immutable (archived ?).
- option 2: Query only the latest "version" document from the subcollection with the contract document on the frontend, and frontend should only be allowed to update the latest version.

[...]

PR #2037, opened 2020-02-21
Should solve deal creation

GrandSchtroumpf:
Create distribution deals on selection
Update version with dealIds


libs/contract/src/lib/contract/tunnel/contract-tunnel.component.ts
```diff
@@ -202,6 +192,19 @@ export class ContractTunnelComponent implements OnInit {
     }
     this.removedDeals = {};
 
+
+    const contract = createContract({ id: contractId, ...this.contractForm.value });
+
+    // Upate Version
+    const lastIndex = contract.versions.length - 1;
```
phpgeekfr: can you  add `// @TODO (#1887)`

libs/contract/src/lib/contract/tunnel/contract-tunnel.component.ts
```diff
@@ -202,6 +192,19 @@ export class ContractTunnelComponent implements OnInit {
...
+    const version = createContractVersion({ ...contract.versions[lastIndex] });
+    this.versionService.update({ id: lastIndex.toString(), ...version }, { params: { contractId }, write })
+    delete contract.versions;
```
phpgeekfr: there is a cleanContract method to do that (if you have no circular dependencies ^^)

PR #2053, opened 2020-02-24
Issue#1987 movie notifications

ioaNikas:
Add new notifications for movies (when submitted to and accepted by Archipel Content)

### DIFF
--- a/libs/notification/src/lib/notification/notification-list/notification-list.component.ts
+++ b/libs/notification/src/lib/notification/notification-list/notification-list.component.ts
@@ -37,8 +37,11 @@ export class NotificationListComponent implements OnInit {
     try {
       if (notification.type === NotificationType.newContract || NotificationType.contractInNegotiation) {
         this.router.navigateByUrl(`c/o/dashboard/deals/${notification.docId}`);
-        this.service.readNotification(notification);
       }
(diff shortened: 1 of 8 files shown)


libs/notification/src/lib/notification/notification-list/notification-list.component.ts
```diff
@@ -37,8 +37,11 @@ export class NotificationListComponent implements OnInit {
     try {
       if (notification.type === NotificationType.newContract || NotificationType.contractInNegotiation) {
         this.router.navigateByUrl(`c/o/dashboard/deals/${notification.docId}`);
-        this.service.readNotification(notification);
       }
+      if (notification.type === NotificationType.movieSubmitted || NotificationType.movieAccepted) {
```
phpgeekfr: update If

libs/notification/src/lib/notification/notification-list/notification-list.component.ts
```diff
@@ -37,8 +37,11 @@ export class NotificationListComponent implements OnInit {
     try {
       if (notification.type === NotificationType.newContract || NotificationType.contractInNegotiation) {
```
phpgeekfr: update If

PR #2070, opened 2020-02-24
Feat 2062 - Update OwnerId functionality on Excel Import

phpgeekfr:


libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -93,6 +93,24 @@ export class AuthService extends FireAuthService<AuthState> {
...
+  public async userExists(uid: string): Promise<boolean> {
+    const snap = await this.db.collection('users').doc(uid).get().toPromise();
+    return snap.exists;
+  }
+
+  /**
+   * Fetch an user based on his uid
+   * @param uid
+   */
+  public async getUser(uid: string): Promise<User> {
```
GrandSchtroumpf: Do you want another user or the current one ?
If it's another User I would recommend to use another service as AuthService is used for authentication. Like `UserService` for example. 
If it's the current user, use the query instead.

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -60,9 +61,13 @@ export class MovieService extends CollectionService<MovieState> {
...
   }
 
-  onCreate(movie: Movie, { write }: WriteOptions) {
+  async onCreate(movie: Movie, { write }: WriteOptions) {
     // When a movie is created, we also create a permissions document for it.
-    return this.permissionsService.addDocumentPermissions(movie, write as firestore.WriteBatch)
+    // Since movie can be created on behalf of another user (An admin from admin panel for example)
+    // We use createdBy attribute to fetch OrgId
+    const userId = movie._meta && movie._meta.createdBy ? movie._meta.createdBy : this.authQuery.userId;
+    const user = await this.authService.getUser(userId);
```
GrandSchtroumpf: user is in the store already : `this.authQuery.getValue().profile`

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -60,9 +61,13 @@ export class MovieService extends CollectionService<MovieState> {
...
+    return this.permissionsService.addDocumentPermissions(movie, write as firestore.WriteBatch, user.orgId);
```
GrandSchtroumpf: If you only want the orgId, you can also use orgQuery.getActiveId() as there is only one organization per user.

libs/organization/src/lib/permissions/+state/permissions.service.ts
```diff
@@ -51,8 +51,7 @@ export class PermissionsService extends CollectionService<PermissionsState> {
    * @param doc
    * @param write
    */
-  public addDocumentPermissions(doc: Movie | Delivery | Contract, write: firestore.WriteBatch) {
-    const organizationId = this.organizationQuery.getActiveId();
+  public addDocumentPermissions(doc: Movie | Delivery | Contract, write: firestore.WriteBatch, organizationId: string = this.organizationQuery.getActiveId()) {
```
GrandSchtroumpf: format on three lines.

libs/movie/src/lib/movie/+state/movie.service.ts
phpgeekfr: This is to cover the case when createdBy value :== current logged in user

libs/movie/src/lib/movie/+state/movie.service.ts
phpgeekfr: same, I'm not working with current logged in user

PR #2077, opened 2020-02-25
added month only and chips auto completion

### LINKED ISSUE #2059: Month Datepicker for avails
Make some tests and try to replace the precision of the datepicker from avails from day to month.
For example instead of clicking on 10th January 2021; just limit it to January 2021

Maxstgt:
close #2059
I also added chips autocompletion to this form

### DIFF
--- a/libs/catalog/src/lib/components/avails-filter/avails-filter.component.ts
+++ b/libs/catalog/src/lib/components/avails-filter/avails-filter.component.ts
@@ -1,3 +1,4 @@
+import { RouterQuery } from '@datorama/akita-ng-router-store';
 import { Component, ChangeDetectionStrategy, Input } from '@angular/core';
 import { AvailsSearchForm } from '@blockframes/catalog/form/search.form';
 import { MediasSlug, MEDIAS_SLUG, staticModels } from '@blockframes/utils/static-model';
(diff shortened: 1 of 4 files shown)


libs/catalog/src/lib/components/avails-filter/avails-filter.component.ts
```diff
@@ -82,4 +42,15 @@ export class AvailsFilterComponent implements OnInit{
...
+  public chosenYearHandler(year: Date, type: 'start' | 'end') {
+    type === 'start' ? this.yearStart = year.getFullYear() : this.yearEnd = year.getFullYear();
+  }
+
+  public chosenMonthHandler(month: Date, datepicker: MatDatepicker<NativeDateModule>, type: 'start' | 'end') {
+    const m = month.getMonth() + 1;
+    const date = type === 'start' ? Date.parse(`${m}-1-${this.yearStart}`) : Date.parse(`${m}-1-${this.yearEnd}`)
+    this.availsForm.get('terms').get(type).setValue(new Date(date));
+    datepicker.close();
+  }
```
ioaNikas: It looks like you want to prevent the users to select a specific day. They can't do it anymore with the date picker but they can still write it down in the input. If it's the excepted behavior, ignore this comment, else you should find a way to disable the input or get rid of it and keep the date pickers.

Maxstgt: @Jackseed what do you say about this comment? As @ioaNikas described I read the ticket.

Jackseed: @Maxstgt it's not a problem if the user is able to write a specific day on its own

Maxstgt: good, then this Implementation is the wanted behavior.

PR #2083, opened 2020-02-26
Feat 1955 - Excel import: improvements

phpgeekfr:
moved column "Do you want to show the operator name on a buyer research ?" from sales sheet to contract sheet to avoid useless contracts parties update when saving deals.

Updated price calculus and checks


libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -2074,6 +2060,21 @@ export class ViewExtractedElementsComponent implements OnInit {
         // VALIDATION
         ///////////////
 
+        // Global contract price 
+        const contractPrice = createPrice();
+        importErrors.contract.doc.titleIds.forEach(titleId => {
+          if(importErrors.contract.last.titles[titleId].price) {
```
GrandSchtroumpf: Maybe create a variable here to make it more readable: 
```typescript
const price = importErrors.contract.last.titles[titleId].price
if (price && price.amount) {
  contractPrice.amount += price.amount;
}
if (price && price.currency) {
  contractPrice.currency = price.currency
}
```

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -2183,6 +2186,43 @@ export class ViewExtractedElementsComponent implements OnInit {
         .map(c => c.trim());
     }
 
+    if (spreadSheetRow[SpreadSheetContractTitle.titlePrice + currentIndex]) {
+      const priceParts = spreadSheetRow[SpreadSheetContractTitle.titlePrice + currentIndex].split(this.subSeparator);
+
+      if(priceParts.length >= 2) {
```
GrandSchtroumpf: Add comment many here. Why "2" ?

PR #2085, opened 2020-02-26
Migrate nwrl 9

GrandSchtroumpf:
Update all libraries to nrwl/angular 9
Remove custom Module for Analytics


apps/catalog/catalog/src/app/catalog.module.ts
```diff
@@ -33,6 +33,7 @@ const routes: Routes = createRoutes({
 });
 
 @NgModule({
+  declarations: [],
```
phpgeekfr: now mandatory ?

GrandSchtroumpf: It's a bug with @angular/fire for lazy-loaded module without declarations field.
I opened an [issue on there repo](https://github.com/angular/angularfire/issues/2337).

PR #2092, opened 2020-02-27
Add a specific empty page when there is no notification

### LINKED ISSUE #2064: Empty state for Activity feed
Use an illustration from @vwmathilde

ioaNikas:
Add specific page when there is no notification in activity feed page.
Add count for each notification tabs.

close #2064

### DIFF
--- a/libs/notification/src/lib/notification/activity-feed/activity-feed.component.ts
+++ b/libs/notification/src/lib/notification/activity-feed/activity-feed.component.ts
@@ -8,7 +8,7 @@ import {
 import { Organization } from '@blockframes/organization/+state/organization.model';
 import { OrganizationQuery } from '@blockframes/organization/+state/organization.query';
 import { Observable } from 'rxjs';
-import { startWith, switchMap } from 'rxjs/operators';
+import { startWith, switchMap, map, distinctUntilChanged } from 'rxjs/operators';
 import { FormControl } from '@angular/forms';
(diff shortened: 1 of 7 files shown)


libs/notification/src/lib/notification/activity-feed/activity-feed.component.ts
```diff
@@ -51,4 +51,16 @@ export class ActivityFeedComponent implements OnInit {
...
+
+  /** Returns the number of notifications according to the filter */
+  getCount(filter?: Notification['type'][]): Observable<number> {
+    return this.notificationQuery.selectCount(notification =>
+      filter ? filter.includes(notification.type) : true
+    )
+  }
+
+  /** Returns the number of keys of an object. */
+  public keysLength(notifications: DateGroup<NotificationDocument[]>): number {
```
GrandSchtroumpf: Just create an observable: 
```typescript
hasKey$ = this.allNotifications$.pipe(notifications => !!map(Object.keys(notifications).length));
```

libs/notification/src/lib/notification/no-activity-feed/no-activity-feed.component.html
```diff
@@ -0,0 +1,7 @@
+<section fxLayout="column" fxLayoutAlign="center center">
+  <h1>You have no notifications yet</h1>
+  <p class="mat-body-1">
```
GrandSchtroumpf: mat-body-1 is the base value.

PR #2128, opened 2020-03-03
Feat 2112 - Replace Worktype "Movie" with "Feature Film"

phpgeekfr:
Replace Worktype "Movie" with "Feature Film"


libs/contract/src/lib/contract/+state/contract.firestore.ts
```diff
@@ -40,10 +40,13 @@ export enum ContractStatus {
 /** @dev Valid values of ContractStatus */
 export const ValidContractStatuses = ['waitingpayment', 'paid', 'accepted'];
 
-export enum ContractType {
-  mandate = 'mandate',
-  sale = 'sale'
-}
+export const contractType = {
```
GrandSchtroumpf: I think this is a level 2 : you could use an array (ContractType  === ContractTypeValue) :
```typescript
const contractType = ['mandate', 'sale'] as const;
const ContractType = typeof contractType[number]
```

libs/contract/src/lib/contract/tunnel/lobby/lobby.component.ts
```diff
@@ -16,9 +16,8 @@ export class LobbyComponent {
     private route: ActivatedRoute,
   ) { }
 
-  async select(contractType: 'sale' | 'mandate') {
-    const type = contractType === 'sale' ? ContractType.sale : ContractType.mandate;
+  async select(type: 'sale' | 'mandate') {
```
GrandSchtroumpf: use type `ContractType` here

libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -24,17 +24,20 @@ import { LegalDocument } from "@blockframes/contract/contract/+state/contract.fi
...
+  feature_film: 'Feature Film',
+  short: 'Short',
+  serie: 'Serie',
+  season: 'Season',
+  volume: 'Volume',
+  episode: 'Episode',
+  collection: 'Collection',
+  tv_film: 'TV Film',
+  flow: 'Flow'
+} as const;
```
GrandSchtroumpf: 👍

libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -45,24 +48,30 @@ export type StoreType = keyof typeof storeType;
 export type StoreTypeValue = typeof storeType[StoreType];
 
 export enum PremiereType {
```
GrandSchtroumpf: Do you keep that as an enum or it's for another PR ?

libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -174,12 +183,15 @@ export interface MovieBudget {
...
-  dubbed = 'dubbed',
-  subtitle = 'subtitle',
-  caption = 'caption',
-}
+export const movieLanguageTypes = {
+  original: 'original',
+  dubbed: 'dubbed',
+  subtitle: 'subtitle',
+  caption: 'caption',
+} as const;
```
GrandSchtroumpf: This is a level 2: MovieLanguageTypes === MovieLanguageTypesValue

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -649,8 +656,9 @@ export class ViewExtractedElementsComponent implements OnInit {
             const parseErrors = [];
             if (language) {
               versionParts.map(v => v.trim()).forEach(v => {
-                if (Object.values(MovieLanguageTypes).map(t => t.toLowerCase()).includes(v.toLowerCase())) {
-                  populateMovieLanguageSpecification(movie.versionInfo.languages, language, v.toLowerCase() as MovieLanguageTypes, true);
+                const key = getKeyIfExists(movieLanguageTypes, v as any);
```
GrandSchtroumpf: It doesn't work if you set the type on the "v" : 
```typescript
versionParts.map(v => v.trim()).forEach((v: Language) => ....)
```

libs/movie/src/lib/movie/components/import/view-extracted-elements/view-extracted-elements.component.ts
```diff
@@ -820,10 +828,11 @@ export class ViewExtractedElementsComponent implements OnInit {
         if (spreadSheetRow[SpreadSheetMovie.worldwideBoxOffice]) {
           spreadSheetRow[SpreadSheetMovie.worldwideBoxOffice].split(this.separator).forEach((version: string) => {
             const boxOfficeParts = version.split(this.subSeparator);
-            if (Object.values(UnitBox).map(t => t.toLowerCase()).includes(boxOfficeParts[0].trim().toLowerCase())) {
+            const unit = getKeyIfExists(unitBox, boxOfficeParts[0] as any);
```
GrandSchtroumpf: Same as above

libs/utils/src/lib/pipes/translate-enum.pipe.ts
```diff
@@ -0,0 +1,39 @@
...
+  async transform(value: string, type: string): Promise<string> {
+
+    if (registeredEnums.includes(type)) {
+      let output;
+      switch (type) {
+        // KFH compatible import
+        case 'unitBox':
+        case 'storeType':
+        case 'workType':
+          output = await import('@blockframes/movie/movie/+state/movie.firestore').then(e => e[type]);
```
GrandSchtroumpf: You might want to keep the value in memory once you got them no ?
I'm not sure what the most optimized way to deal with that.

libs/utils/src/lib/pipes/translate-enum.pipe.ts
```diff
@@ -0,0 +1,39 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { getKeyIfExists } from '../helpers';
+
+const registeredEnums = [
+  'unitBox',
+  'storeType',
+  'workType',
+];
+
+export type RegisteredEnums = keyof typeof registeredEnums;
```
GrandSchtroumpf: Is it working ? The key of an array is supposed to be a number. 
I would have done `typeof registeredEnums[number]`

libs/utils/src/lib/pipes/translate-enum.pipe.ts
```diff
@@ -0,0 +1,39 @@
...
+
+@Pipe({
+  name: 'translateEnum'
```
GrandSchtroumpf: I'm not sure to understand this pipe's name. Does it deal with enum or readonly object ?

libs/contract/src/lib/contract/+state/contract.firestore.ts
phpgeekfr: Here contractTypeValue === ContractType but I think it is a mistake. I will change it to 
```
export const contractType = {
  mandate: 'Mandate',
  sale: 'Sale'
} as const;
```

libs/movie/src/lib/movie/+state/movie.firestore.ts
phpgeekfr: for another PR.

libs/movie/src/lib/movie/+state/movie.firestore.ts
phpgeekfr: Same as above. Will transform it into:
```
export const movieLanguageTypes = {
  original: 'Original',
  dubbed: 'Dubbed',
  subtitle: 'Subtitle',
  caption: 'Caption',
} as const;
```

libs/utils/src/lib/pipes/translate-enum.pipe.ts
phpgeekfr: If called many times, the same instance of the pipe is used?

libs/utils/src/lib/pipes/translate-enum.pipe.ts
phpgeekfr: In fact this is not used. removing it

libs/utils/src/lib/pipes/translate-enum.pipe.ts
GrandSchtroumpf: You run the pipe after each rendering cycle. I guess you don't want to get make an HTTP request to a static file each time you run a new rendering cyle.
I guess the best of the best would be to do : 
```typescript
const workType = { ... };
type WorkType = ...;
export const WORK_TYPE = new InjectionToken<WorkType>('work type', {
  providedIn: 'root', // make this instance lazy loaded
  factory: () => workType 
});
```
But it's tedious, and at that point I guess the best would be to import all model on this pipe like for TranslateSlug ^^

libs/utils/src/lib/pipes/translate-enum.pipe.ts
phpgeekfr: renamed it

PR #2137, opened 2020-03-04
Create reusable component for contract version table

### LINKED ISSUE #1952: Component VersionTable
Table to display the negocation history : 
- Used in catalog/marketplace/deal/negociation https://app.zeplin.io/project/5de7944cf85691577aacfbb7/screen/5e46d3919fc526a80b88dfeb
- Used in catalog/dashboard/deal/view

Create a component that format data and display it

ioaNikas:
Create reusable component for contract version table

close #1952


libs/contract/src/lib/version/components/table/table.component.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+@Component({
+  selector: '[versions] contract-version-table',
+  templateUrl: './table.component.html',
+  styleUrls: ['./table.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class VersionTableComponent {
+  columns: Partial<typeof columns> = getColumns(baseColumns);
```
GrandSchtroumpf: I don't understand the point of this line if the keys of the columns are not dynamic.

ioaNikas: Copied and pasted without understanding everything. Now I get the point. Indeed they are static and we don't need that.

PR #2138, opened 2020-03-04
ops: fix deploys with new ngcc

lsenta:


firebase.json
```diff
@@ -38,7 +38,7 @@
   ],
   "functions": {
     "predeploy": [
-      "cp ./package.json dist/apps/backend-functions/package.json",
+      "cat ./package.json | jq \"del(.scripts.postinstall)\" > dist/apps/backend-functions/package.json",
```
GrandSchtroumpf: Is it supposed to stay ?

lsenta: Not really, goal now is to make sure we can deploy (devs, ci, prod, staging) despite the ngcc update.

PR #2139, opened 2020-03-04
Design review #2002

### LINKED ISSUE #2002: Design Review: Library
REALLY IMPORTANT
- [x] not the right color for fab button -> primary 500
image
- [x] please change primary color for dark theme. It's not the good color !!!
- [x] Caption interline : 21px
- [x] 4px between H6 and Caption
image
- [x] use right color for mat-chips 
- [x] use two tone icons please, everywhere
image
- [x] 4 px between H5 and caption
image
- [x] Please, do the same spacing between same elements
image
- [x] Right calendar icon on library
image
- [x] Spacing 24 px
image
- [x] Card font : title -> H6 / Real -> caption
image

It's ok, I may not cry
- [ ] Normally it's a menu drop down, not mat-select
image
- [ ] It's possible to have a more beautiful scrollabar ?
image
- [x] more space between Two elements
image
- [ ] for item "genre", it should be the same component like Country of Origin
- [x] more space between lines (24 px)
image
- [x] Intercom in primary (light theme everywhere)
image

Mussuranae:
Fix #2002

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/movie/search/search.component.scss
+++ b/apps/catalog/catalog/src/app/marketplace/movie/search/search.component.scss
@@ -67,7 +67,7 @@ section.filters {
     }
 
     h6 {
-      margin-bottom: 4px;
+      margin-bottom: 12px;
     }
 
     .checkbox-padding {
--- a/apps/catalog/catalog/src/app/marketplace/movie/search/search.component.scss
+++ b/apps/catalog/catalog/src/app/marketplace/movie/search/search.component.scss
@@ -4,7 +4,7 @@
   margin-top: 0;
   margin-bottom: 0;
   mat-label {
-    margin-left: 20px;;
+    margin-left: 20px;
   }
 }
 
@@ -23,10 +23,6 @@ h5 {
 .language-button {
   margin-top: 15px;
   margin-left: 16px;
-}
-
-mat-checkbox {
-  margin-top: 16px;
 }
 
 .chip-language {
@@ -64,6 +60,18 @@ section.filters {
     mat-card {
       max-height: calc(100vh - var(--toolbarHeight) - 40px);
       overflow: auto;
+    }
+
+    mat-checkbox {
+      margin-top: 16px;
+    }
+
+    h6 {
+      margin-bottom: 4px;
+    }
+
+    .checkbox-padding {
+      padding-left: 16px
     }
 
     .movie-list {
--- a/libs/catalog/src/lib/components/avails-filter/avails-filter.component.html
+++ b/libs/catalog/src/lib/components/avails-filter/avails-filter.component.html
@@ -39,7 +39,7 @@
     </mat-expansion-panel-header>
     <mat-form-field appearance="outline">
       <mat-chip-list #territoryChipList>
-        <mat-chip color="default" *ngFor="let territory of selectedMovieTerritories; let i = index" [selectable]="selectableTerritory"
+        <mat-chip *ngFor="let territory of selectedMovieTerritories; let i = index" [selectable]="selectableTerritory"
           [removable]="!availsForm.value.isActive" (removed)="removeTerritory(territory, i)">
           {{ territory }}
           <mat-icon matChipRemove *ngIf="!availsForm.value.isActive" svgIcon="close"></mat-icon>
--- a/libs/catalog/src/lib/components/avails-filter/avails-filter.component.html
+++ b/libs/catalog/src/lib/components/avails-filter/avails-filter.component.html
@@ -14,6 +14,7 @@
           You only can choose the current year and above.
         </mat-error>
         <mat-datepicker-toggle test-id="datepicker-from" matSuffix [for]="pickerAvailableFrom">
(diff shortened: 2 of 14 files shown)


apps/catalog/catalog/src/app/marketplace/movie/search/search.component.scss
```diff
@@ -66,6 +62,18 @@ section.filters {
...
 
+    mat-checkbox {
+      margin-top: 16px;
+    }
+
+    h6 {
+      margin-bottom: 12px;
+    }
+
+    .checkbox-padding {
```
GrandSchtroumpf: Why not putting it on the checkbox directly ?

libs/catalog/src/lib/components/avails-filter/avails-filter.component.html
```diff
@@ -37,10 +39,10 @@ <h5>AVAILS</h5>
     </mat-expansion-panel-header>
     <mat-form-field appearance="outline">
       <mat-chip-list #territoryChipList>
-        <mat-chip *ngFor="let territory of selectedMovieTerritories; let i = index" [selectable]="selectableTerritory"
+        <mat-chip color="default" *ngFor="let territory of selectedMovieTerritories; let i = index" [selectable]="selectableTerritory"
```
GrandSchtroumpf: What is color="default". If we want a default color it should be when we don't apply any "color" property. If it doesn't work, there is an issue in the theme.

libs/catalog/src/lib/components/avails-filter/avails-filter.component.scss
```diff
@@ -9,3 +9,11 @@ mat-form-field {
 h5 {
   margin-top: 20px;
 }
+
+mat-checkbox {
```
GrandSchtroumpf: It's the second time, should this be in the theme ? All checkbox should behave like that ?

libs/movie/src/lib/movie/form/main/genres/genres.component.html
```diff
@@ -1,9 +1,9 @@
 <!--  Genre -->
 <mat-chip-list aria-label="Genres selection">
   <ng-container *ngFor="let genre of values$ | async; let i = index">
-    <mat-chip selectable removable>
+    <mat-chip selectable removable color="default">
```
GrandSchtroumpf: same as above

libs/ui/src/lib/form/chips-autocomplete/chips-autocomplete.component.ts
```diff
@@ -36,6 +36,7 @@ export class ChipsAutocompleteComponent implements OnInit, OnDestroy {
   @Input() removable = true;
   @Input() disabled = false;
   @Input() placeholder = 'New Items';
+  @Input() color = 'default';
```
GrandSchtroumpf: No. No "default" color

styles/themes/dark.scss
```diff
@@ -140,6 +140,7 @@ $bf-dark-foreground: (
   // Custom variables
   blue-icon:    #9eb1ff,
   dark-text:    #80818a,
+  active:       #b7bcd9,
```
GrandSchtroumpf: What is "active". I really don't like when we add new variables because: 
- There are not retro-compatible with material
- There is no doc about that
- You NEVER duplicate it in light & dark

libs/catalog/src/lib/components/avails-filter/avails-filter.component.html
Mussuranae: It's an issue of Angular Material. We can't apply color on different chips with Angular Material, so when chips have the selected attribute, you can apply a color, but for almost every other use of the chip (in input or autocomplete input) you can't apply a color and it's automatically their default grey color. So I found a way to apply color for every chips we use (you can see the code in styles/components/mat-chips.scss) but after that it took always the primary color by default. So I had a "default" color, so that, we can choose which color we want for chips.

styles/themes/dark.scss
Mussuranae: It's other variables that Mathilde used in both light and dark theme. The 4th  palette including beware color too. It exists in both theme but not implemented in the code.
https://app.zeplin.io/project/5de7944cf85691577aacfbb7/screen/5e0f712e213d5298b590e201

apps/catalog/catalog/src/app/marketplace/movie/search/search.component.scss
Mussuranae: Because some section with checkboxes are well positioned and some aren't. The problem isn't on the checkboxes but rather on the component where checkboxes are (for this example, some checkboxes are in a list, and the mat-list-item-content has a padding I didn't success to cancel for this case, but we don't want to cancel this padding for all mat-list-item-content we have in the project). So I put a padding for others checkboxes that aren't in an other component.

libs/catalog/src/lib/components/avails-filter/avails-filter.component.scss
Mussuranae: No, when checkboxes will be in an other component (like table or mat-list) they will be well placed without additional padding or margin.

PR #2140, opened 2020-03-04
ok feat 2130 - create a pipe for termToPrettyDate method

phpgeekfr:
create a pipe for termToPrettyDate method


libs/utils/src/lib/pipes/to-pretty-date.pipe.ts
```diff
@@ -0,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { termToPrettyDate, Terms } from '../common-interfaces/terms';
+
+@Pipe({
+  name: 'termToPrettyDate'
```
GrandSchtroumpf: Add `pure: true` when a pipe is a pure function.

GrandSchtroumpf: Could you name it `termDate` ?

libs/utils/src/lib/pipes/to-pretty-date.pipe.ts
```diff
@@ -0,0 +1,11 @@
...
+})
+export class TermToPrettyDatePipe implements PipeTransform {
+  transform(term: Terms , type: 'end' | 'start'): string {
```
GrandSchtroumpf: Would it make sense to set a value by default ? Or you think it's better not to ?

PR #2155, opened 2020-03-05
angular: prod with source maps

lsenta:
Adding sourcemaps during production build.

Sourcemaps: The mapping between typescript code and the production compiled code, that'll make the stack trace for errors more readable. This is the simple approach: generate and deploy the source maps in production.

Usually we don't, we keep them hidden so only devs can see them, and use some tools to sync the data / logs (using sentry for example).

But that's more complexity for no reason: the code is open source anyway, nothing to obfuscate.


GrandSchtroumpf:
I thought adding source map to the code increases bundle size & reduce performance as the javascript executed is larger. Do you have documentation or any ressource concerning this ?

lsenta:
(that's not the case, proceeding with merging).

PR #2156, opened 2020-03-05
Feat 2090 - Be able to edit contrat.last.titles<titleId> Price & debug

phpgeekfr:


libs/admin/src/lib/admin-panel/pages/contract/contract.component.html
```diff
@@ -35,15 +35,20 @@ <h4>Party : {{ partyDetails.party.displayName }} ( org: {{ partyDetails.party.or
...
+          <li><strong>Status : </strong>{{ partyDetails.status | toLabel: 'contractStatus' | async  }}</li>
           <li><strong>Sign date : </strong>{{ partyDetails.signDate ? (partyDetails.signDate | date: 'short') : ' -- ' }}</li>
           <li><strong>Role : </strong>{{ partyDetails.party.role }}</li>
           <li><strong>Show name : </strong>{{ partyDetails.party.showName ? 'yes' : 'no' }}</li>
           <li>
             <span>Child roles : </span>
-            <ng-container *ngFor="let childRole of partyDetails.childRoles; let isLast = last">
-              {{ childRole }}{{isLast ? '' : ' - '}}
+            <ng-container *ngIf="partyDetails.childRoles.length else nochildrole">
+              <ng-container *ngFor="let childRole of partyDetails.childRoles; let isLast = last">
```
GrandSchtroumpf: The translateSlug manage an array of slug directly. This should be enough
```html
<ng-container *ngIf="partyDetails.childRoles.length else nochildrole">
  {{ childRole | translateSlug: 'SUB_LICENSOR_ROLES' }}
</ng-container>
```

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
```diff
@@ -127,27 +130,42 @@ export class ContractComponent implements OnInit {
...
-        });
-
-        this.titles = [...this.titles];
-      })
     });
+  }
 
+  private loadTitles() {
+    this.titles = [];
+    Object.keys(this.contract.last.titles).forEach(async id => {
```
GrandSchtroumpf: Use map instead : 
```typescript
this.titles = Object.keys(this.contract.last.titles).map(...)
```
Like that you don't have to do `this.titles = [...this.titles];` to force the rerendering

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
```diff
@@ -127,27 +130,42 @@ export class ContractComponent implements OnInit {
...
+      });
+
+      this.titles = [...this.titles];
+    })
+  }
+
+  private async reloadContractAndVersions(newVersionId: string, reloadTitles = false) {
+    this.version = parseInt(newVersionId, 10);
+    this.contractVersions = await this.contractVersionService.getContractVersions(this.contractId);
+    this.contract = await this.contractService.getContractWithLastVersion(this.contractId);
```
GrandSchtroumpf: Use `Promise.all([])` for calls to the server

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
```diff
@@ -243,6 +261,69 @@ export class ContractComponent implements OnInit {
...
+        titleId,
+        subtitle: 'If you leave now, your changes will not be saved.',
+        titleDetail: this.contract.last.titles[titleId]
+      },
+      disableClose: true
+    });
+    return dialogRef.afterClosed().subscribe((output: ContractTitleDetail | { remove: boolean }) => this.updateTitle(output, titleId));
+  }
+
+  private async updateTitle(output: ContractTitleDetail | { remove: boolean }, titleId?: string, ): Promise<boolean> {
```
GrandSchtroumpf: Add "@todo(....) don't use contract.last, but contract.version"

GrandSchtroumpf: Also add a comment, I don't understand with you have two totally different type of output.

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
```diff
@@ -243,6 +261,69 @@ export class ContractComponent implements OnInit {
...
+    // A title have been updated, added or removed. Need to re-calculate contract price.
+    this.calculatePrice(writeableVersion);
+
+    // Udpate titleIds array
+    writeableContract.titleIds = Object.keys(writeableVersion.titles);
+
+    // @TODO (#1887)
+    // Update contract and create a new version
+    await this.contractService.update(writeableContract);
+    const newVersionId = await this.contractVersionService.addContractVersion({ doc: this.contract.doc, last: writeableVersion });
```
GrandSchtroumpf: I think we should put as much logic from above a possible in a service. We might forget to update it everywhere when we update it, and might lead to some debaviour difficult to debug.

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
```diff
@@ -260,21 +341,20 @@ export class ContractComponent implements OnInit {
...
     const newVersionId = await this.contractVersionService.addContractVersion({ doc: this.contract.doc, last: update });
-    this.version = parseInt(newVersionId, 10);
-    this.contractVersions = await this.contractVersionService.getContractVersions(this.contractId);
-    this.cdRef.detectChanges();
+    await this.reloadContractAndVersions(newVersionId);
 
     this.snackBar.open('Contract global price updated !', 'close', { duration: 5000 });
   }
+
+  private calculatePrice(version :ContractVersion) {
```
GrandSchtroumpf: Look like tyou don't use `this`. You could extract this method as a pure function (maybe make it reusable somehow later)

GrandSchtroumpf: With some luck it might already exist at 2 others places ^^

PR #2163, opened 2020-03-06
yandex tracking script in head tag

Maxstgt:
smal fix
Screenshot from 2020-03-06 15-09-28


apps/catalog/catalog/src/index.html
```diff
@@ -1,19 +1,22 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
```
lsenta: What is happening here? It looks like all you did was to change the indentation

Maxstgt: yes, I was trying to hard code the script tag to see if it is working. Shouldn't have pushed it to the PR!

lsenta:
What's the rationale for this patch? The PR description is only "small fix" and there's no ticket associated.

apps/catalog/catalog/src/index.html
lsenta: Remove it, `git reset HEAD~2 -- path/to/filename.html`, `git add -u` `git commit --amend`

Maxstgt:
> What's the rationale for this patch? The PR description is only "small fix" and there's no ticket associated.

The problem was, that we used a npm package for connecting our app with yandex metrika, but this package wasnt maintained anymore and was simply not working. So I did it by myself. And there is no issue about it. Vincent was asking me to do it.

apps/catalog/catalog/src/index.html
Maxstgt: thanks

PR #2172, opened 2020-03-09
correct timestamp type in notications

Hugo-Hontans:
- [x] Fix backend build issue on the CI


libs/notification/src/lib/notification/+state/notification.firestore.ts
```diff
@@ -42,6 +42,6 @@ export interface NotificationOptions {
 /** Generic informations for a Notification. */
 export interface NotificationDocument extends NotificationOptions {
   id: string;
-  isRead: boolean;
+  isRead: boolean
```
lsenta: You had it all perfect until that one useless and wrong edit, SO CLOSE!

Hugo-Hontans: Eyes of Lynx

PR #2176, opened 2020-03-09
added akita ng fire hook when logging out

### LINKED ISSUE #320: Clean /reset all stores on user logout

Maxstgt:
close #320

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/layout/layout.component.ts
+++ b/apps/catalog/catalog/src/app/marketplace/layout/layout.component.ts
@@ -44,10 +44,4 @@ export class LayoutComponent implements OnInit, AfterViewInit, OnDestroy {
   ngOnDestroy() {
     this.sub.unsubscribe();
   }
-
-  public async logout() {
-    await this.authService.signOut();
-    // TODO: issue#879, navigate with router
-    window.location.reload();
-  }
 }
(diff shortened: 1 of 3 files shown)


apps/catalog/catalog/src/app/marketplace/layout/layout.component.ts
```diff
@@ -44,10 +44,4 @@ export class LayoutComponent implements OnInit, AfterViewInit, OnDestroy {
   ngOnDestroy() {
     this.sub.unsubscribe();
   }
-
-  public async logout() {
-    await this.authService.signOut();
-    // TODO: issue#879, navigate with router
-    window.location.reload();
-  }
```
Maxstgt: function was never called

libs/auth/src/lib/+state/auth.service.ts
```diff
@@ -38,6 +38,16 @@ export class AuthService extends FireAuthService<AuthState> {
...
   }
 
+  /**
+   * @description function that gets triggered when
+   * AuthService.signOut is called
+   */
+  public onSignout() {
+    localStorage.clear();
+    indexedDB.deleteDatabase('firebase-installations-database');
+    indexedDB.deleteDatabase('firebaseLocalStorageDb');
```
phpgeekfr: Can you add some comments so other can understand what this function is doing?

phpgeekfr: Once logged out there is no remaining user data on browser side ?

Maxstgt: yes that is how I understood the requirements of the issue. Like the example you said that maybe the user is on a shared desktop and we dont want to have any data remaining.

Issue #879, opened 2019-09-04
Can't use Angular Router after the disconnexion of the user (current fix is ​​to reload the page)


Hugo-Hontans:

Maxstgt:
since we are using akita ng fire, we either can do some redirecting when the sync function fails in the guard, or we just call window.location.reload(), in the hook of onSignout, which is the current state. And I talked to @GrandSchtroumpf and he thinks it is okay to cal window.localtion.reload(). So this issue can be closed now. Reopen if you want to have a different behavior

PR #2182, opened 2020-03-10
removed TODO

Maxstgt:
ref #879


libs/account/src/lib/profile/profile-menu/profile-menu.component.ts
```diff
@@ -29,8 +29,6 @@ export class ProfileMenuComponent implements OnInit{
 
   public async logout() {
     await this.service.signOut();
-    // TODO: issue#879, navigate with router
-    window.location.reload();
```
Hugo-Hontans: Why do you remove the `    window.location.reload();` here ?

Maxstgt: becuase there is a akita ng fire hook onSignout, where this is called, so you dont DRY everywhere

PR #2187, opened 2020-03-10
Issue#1935 service

Hugo-Hontans:
- [x] Create a breakpointService
- [x] Use it for components


libs/ui/src/lib/tunnel/layout/layout.component.html
```diff
@@ -1,13 +1,13 @@
 <mat-toolbar fxLayout="row" fxLayoutAlign="space-between center">
-  <ng-container *ngIf="mobileQuery.matches">
+  <ng-container *ngIf="isSmall$ | async">
```
GrandSchtroumpf: Could you give the name of the breakpoint instead ? ltMd

Hugo-Hontans: Sure

PR #2190, opened 2020-03-11
Issue#2104 design review illustration

### LINKED ISSUE #2104: DESIGN REVIEW - Illustrations
Change all illutrations -> svg to webp
[ ] light theme
[ ] dark theme

jenna1k:
resolve #2104


libs/ui/src/lib/media/cropper/cropper/cropper.component.scss
```diff
@@ -7,6 +7,10 @@ section {
   }
 }
 
+.add-files-icon {
```
Maxstgt: you can use img tag here, cause I can't see any other img tag in the cropper.html

jenna1k: there is another img tag for showing cropped image

PR #2195, opened 2020-03-12
Issue#2111 view

Hugo-Hontans:
- [x] Show the view without all the data being loaded


libs/movie/src/lib/movie/components/movie-analytics-chart/movie-analytics-chart.component.ts
```diff
@@ -97,23 +103,23 @@ export class MovieAnalyticsChartComponent {
...
+      axisBorder: {show: false},
       axisTicks: {show: false}
     };
   }
 
   totalHitsOnCurrentMonth(eventName: MovieAnalyticsEventName) {
-    const total = this.chartData.find(chart => chart.eventName === eventName).y
-    return getSum(total);
+    const total = this.chartData.find(chart => chart.eventName === eventName).y;
+    return total ? getSum(total) : undefined;
```
Maxstgt: why undefined and not `0`

Hugo-Hontans: When the analytics is not loaded, we want to display nothing and especially not 0

GrandSchtroumpf: Same as previous PR. You don't need to specify "return undefined" : 
```typescript
if (total) {
  return getSum(total);
}
```

Hugo-Hontans: Yes sure !

PR #2203, opened 2020-03-12
Issue#2111 analytics

Hugo-Hontans:
- [x] Create analytics store and query in movie store and query.
- [x] Create functions in movie service to sync on analytics
- [x] Implements them in dashboard titles and home pages

### DIFF
--- a/libs/movie/src/lib/movie/+state/movie.service.ts
+++ b/libs/movie/src/lib/movie/+state/movie.service.ts
@@ -6,7 +6,7 @@ import {
   WriteOptions,
   syncQuery
 } from 'akita-ng-fire';
-import { switchMap } from 'rxjs/operators';
+import { switchMap, filter, tap, map } from 'rxjs/operators';
 import { createMovie, Movie, MovieAnalytics } from './movie.model';
 import { MovieState, MovieStore } from './movie.store';
 import { AuthQuery, AuthService } from '@blockframes/auth';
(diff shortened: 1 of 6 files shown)


libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -37,11 +38,44 @@ export class MovieService extends CollectionService<MovieState> {
...
+        const f = this.functions.httpsCallable('getMovieAnalytics');
+        return f({ movieIds, daysPerRange: 28 });
+      }),
+      tap(analytics => this.store.analytics.upsertMany(analytics))
+    )
+  }
+
+  /** Gets every analytics for all movies with status accepted and sync them. */
+  public syncAnalyticsWithStatusAccepted() {
+    return combineLatest([
```
GrandSchtroumpf: As say about create on logic  : 
You can start with : 
```typescript
public syncAnalytics(filterBy?: (movie: Movie) => boolean) {
  const movieIds$ = this.query.selectAll({fitlerBy}).pipe(movies => movies .map(m => m.id));
  return combineLatest([movieIds$, this.query.analytics.select('ids')]).pipe(

  )
}
```
**Later** you might want to have more options
```typescript
interface SyncAnalyticsOptions {
  filterMovie: (movie: Movie) => boolean;
  daysPerRange: number;
}
syncAnalytics(options: SyncAnalyticsOptions = { daysPerRange: 28 }) {
 ...
}
```

libs/movie/src/lib/movie/+state/movie.store.ts
```diff
@@ -7,6 +7,7 @@ export interface MovieState extends EntityState<Movie, string>, ActiveState<stri
 @Injectable({ providedIn: 'root' })
 @StoreConfig({ name: 'movies', idKey: 'id' })
 export class MovieStore extends EntityStore<MovieState> {
+  analytics = new EntityStore<MovieState>(null, { name: 'movieAnalytics', idKey: 'movieId' });
```
GrandSchtroumpf: You're sure you don't need any initial state, like `{ ids: [], entities: {} }` ?

Hugo-Hontans: Yes

PR #2207, opened 2020-03-12
added movie tunnel guard

Maxstgt:


apps/catalog/catalog/src/app/dashboard/movie-tunnel/summary/summary.component.ts
```diff
@@ -40,6 +40,8 @@ export class TunnelSummaryComponent {
         this.router.navigate(['../../../../titles', movieId, 'details'], { relativeTo: this.route })
       })
     } else {
+      // Log the invalid forms
+      console.error(this.findInvalidControlsRecursive(this.form))
```
Maxstgt: dev purposes

libs/movie/src/lib/movie/guards/movie-tunnel.guard.ts
```diff
@@ -0,0 +1,20 @@
...
+export class MovieTunnelGuard implements CanActivate {
+    constructor(private service: MovieService, private routerQuery: RouterQuery,
+        private router: Router, private orgQuery: OrganizationQuery, private orgService: OrganizationService) { }
+
+    async canActivate() {
+        const movieId: string = this.routerQuery.getValue().state.root.params.movieId;
+        const movie = await this.service.getValue(movieId)
+        const org = await this.orgService.getValue(this.orgQuery.getActiveId())
+        return movie?.main.storeConfig.status !== 'submitted' && org.movieIds.includes(movieId) ?
+            true : this.router.navigateByUrl('c/o/dashboard/titles');
```
Maxstgt: navigate function didn't worked.

apps/catalog/catalog/src/app/dashboard/movie-tunnel/start/start-tunnel.component.ts
```diff
@@ -46,11 +50,14 @@ export class StartTunnelComponent implements OnInit {
...
-      this.loading = false;
-      this.router.navigate([movieId], { relativeTo: this.route });
+      await this.movieService.runTransaction(async (write) => {
+        const createdBy = this.auth.getValue().uid;
+        const movie = createMovie({ _meta: { createdBy } });
+        const movieId = await this.movieService.add(movie, { write });
+        await this.orgService.update(this.orgQuery.getActiveId(), (org) => ({ movieIds: [...org.movieIds, movieId] }), { write })
+        this.loading = false;
+        this.router.navigate([movieId], { relativeTo: this.route });
+      })
```
GrandSchtroumpf: Put all the logic that is not related to the transaction outside of the transaction (only keep "add" & "update" inside)

libs/movie/src/lib/movie/guards/movie-tunnel.guard.ts
GrandSchtroumpf: So far as I understand you cannot navigate from an injected class, only from a routed component.
BUT
canActive support RouterUrl output to redirect. So the best is to do : 
```typescript
const isSubmitted = movie?.main.storeConfig.status === 'submitted';
const orgHasMovie = org.movieIds.includes(movieId);
const redirect = this.router.parseUrl(`c/o/dashboard/titles/${movieId}`);
return (!isSubmitted && orgHasMovie) || redirect;  
```

Issue #2222, opened 2020-03-16
[Algolia] Use algolia autocomplete


GrandSchtroumpf:
Following the creation of the algolia autocomplete (see https://github.com/blockframes/blockframes/issues/2088), let's use it !

Here is a list of component that could use it: 
**catalog**: 
- [x] dashboard/layout/layout.component
- [x] marketplace/movie/search/search.component

**organization**:
- [x] organization/.../components/organization-search
- [x] organization/.../pages/organization-find

**festival** (wait for PR on festival to be merged)
- [x] dashboard/dashboard.component
- [x] marketplace/title/list/list.component

PR #2228, opened 2020-03-17
added algolia autocomplete component to other components

Maxstgt:
ref #2222
part 1, festival app is coming soon


apps/catalog/catalog/src/app/dashboard/layout/layout.component.html
```diff
@@ -29,6 +29,12 @@
       <button mat-icon-button (click)="sidenav.toggle()">
         <mat-icon>menu</mat-icon>
       </button>
+      <algolia-autocomplete [indexName]="movieIndex" pathToValue="objectID" mode="legacy" (selectionChange)="navigate($event)">
+        <!-- TODO #2225 -->
+        <ng-template let-result>
+          <span>{{ result.movie.main.title.international }}</span>
```
GrandSchtroumpf: could you use an anchor with routLink here, instead of the selectionChange to navigate ?

Maxstgt: yep, I will try

Issue #2220, opened 2020-03-16
[User] Refactorize libs "account" & "auth" into "user"


GrandSchtroumpf:
The lib "account" has never been really clear.

Let's create a library "user" with two subfolder : 
- **auth**: _folder dedicated to the connected user_.
- **user**: _folder dedicated for any other user (like the members of an organization)_

## Auth
Let's keep the same structure as the current "auth" folder.
Add components from the current "account/profile" folder.

## User
Create a state with a CollectionService.
Add component from "organization/member".

PR #2236, opened 2020-03-18
Landing page redirection guard

ioaNikas:
Create specific guard to redirect users out of landing page when needed.

close #2236

### DIFF
--- a/apps/catalog/catalog/src/app/catalog.module.ts
+++ b/apps/catalog/catalog/src/app/catalog.module.ts
@@ -5,13 +5,11 @@ import { createRoutes } from '@blockframes/utils/routes';
 
 // Guards
 import { CatalogAppGuard } from './guards/catalog-app.guard';
-import { UserRedirectionGuard } from '@blockframes/auth/guard/user-redirection.guard';
 
 const routes: Routes = createRoutes({
   appName: 'catalog',
   landing: {
     path: '',
-    canActivate: [UserRedirectionGuard],
     loadChildren: () => import('./landing/landing.module').then(m => m.LandingModule)
   },
   appsRoutes: [{
--- a/apps/catalog/catalog/src/app/catalog.module.ts
+++ b/apps/catalog/catalog/src/app/catalog.module.ts
@@ -2,8 +2,10 @@ import { NgModule } from '@angular/core';
 import { RouterModule, Routes } from '@angular/router';
 import { MovieCollectionGuard } from '@blockframes/movie/movie/guards/movie-collection.guard';
 import { createRoutes } from '@blockframes/utils/routes';
+
+// Guards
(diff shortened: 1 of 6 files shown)


apps/catalog/catalog/src/app/guards/user-redirection.guard.ts
```diff
@@ -0,0 +1,42 @@
+import { Injectable } from '@angular/core';
+import { Router, CanActivate } from '@angular/router';
+import { OrganizationService } from '@blockframes/organization/+state/organization.service';
+import { AngularFireAuth } from '@angular/fire/auth';
+
+@Injectable({ providedIn: 'root' })
+export class UserRedirectionGuard implements CanActivate {
```
GrandSchtroumpf: Maybe it make more sense to be it in the auth library.

apps/catalog/catalog/src/app/catalog.module.ts
```diff
@@ -3,11 +3,13 @@ import { RouterModule, Routes } from '@angular/router';
 import { MovieCollectionGuard } from '@blockframes/movie/movie/guards/movie-collection.guard';
 import { createRoutes } from '@blockframes/utils/routes';
 import { CatalogAppGuard } from './guards/catalog-app.guard';
+import { UserRedirectionGuard } from './guards/user-redirection.guard';
 
 const routes: Routes = createRoutes({
   appName: 'catalog',
   landing: {
     path: '',
+    canActivate: [UserRedirectionGuard],
```
GrandSchtroumpf: Could you integrate that in the `createRoutes` app directly ?

libs/organization/src/lib/+state/organization.service.ts
```diff
@@ -119,6 +118,15 @@ export class OrganizationService extends CollectionService<OrganizationState> {
     });
   }
 
+  /** Returns the organization of a specific user. */
+  public async getUserOrganization(userId: string) {
+    const userDoc = await this.db.doc(`users/${userId}`).get().toPromise();
+    const { orgId } = userDoc.data() as User;
```
GrandSchtroumpf: Can you add a `@todo(#2220) use userService.getValue(userId) when created` here ?

apps/catalog/catalog/src/app/catalog.module.ts
ioaNikas: Since it is something specific to landing page, I don't find an elegant way to integrate it in `createRoutes`. If you got a clue about that, I'm open.

GrandSchtroumpf: Maybe : 
```typescript
children: [{
  ...landing,
  canActivate: [...landing.canActive, UserRedirectionGuard],
}, ....]
```

PR #2257, opened 2020-03-20
Create user card + update auth service

GrandSchtroumpf:
Create User card under the "account" library
Update akita-ng-fire
Update auth service to match new version of akita-ng-fire

### DIFF
--- a/libs/account/src/lib/profile/components/card/card.component.html
+++ b/libs/account/src/lib/profile/components/card/card.component.html
@@ -1,0 +1,4 @@
+<img [imgRef]="user.avatar" placeholderAsset="profil_user.webp" alt="User avatar" />
+<h5>{{ user.surname }}</h5>
+<h4>{{ user.position }}</h4>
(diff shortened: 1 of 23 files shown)


libs/account/src/lib/profile/components/card/card.component.html
```diff
@@ -0,0 +1,4 @@
+<img [imgRef]="user.avatar" placeholderAsset="profil_user.webp" alt="User avatar" />
+<h5>{{ user.surname }}</h5>
```
Mussuranae: Is user's surname is enough ? We have name and surname for user in the DB, don't you want the both ?

GrandSchtroumpf: I'll change to name & surname. Would be nice to change to regular firstName, lastName though ^^

Mussuranae: Yes, I will see with Bruce to harmonise all this kind of data.

PR #2264, opened 2020-03-23
load theme directly in style app file

Mussuranae:
Change the load of the theme.


apps/catalog/catalog/src/styles.scss
```diff
@@ -3,3 +3,27 @@
 
 // @todo(#1617) Lazy Load this
 @import "~leaflet/dist/leaflet.css";
+
+.light-theme {
+  @include angular-material-theme($bf-light-theme);
+  @include sat-datepicker-theme($bf-light-theme);
```
GrandSchtroumpf: I don't think we are using the datepicker here.

apps/delivery/delivery/src/styles.scss
```diff
@@ -1,2 +1,20 @@
 /* You can add global styles to this file, and also import other style files */
 @import "../../../../styles/style.scss";
+
+.light-theme {
```
GrandSchtroumpf: don't modify this file as it's removed by Yohann in a parallel PR

apps/catalog/catalog/src/styles.scss
```diff
@@ -3,3 +3,27 @@
...
+  @include bf-material-theme($bf-light-theme);
```
GrandSchtroumpf: I think you want to load the world-map style here too. We don't use it in the festival (yet)

styles/theme.scss
```diff
@@ -46,13 +46,13 @@
   @include mat-bf-snack-bar-theme($theme);
   @include bf-footer-theme($theme);
   @include bf-widget-theme($theme);
-  @include bf-tunnel-theme($theme);
-  @include bf-table-theme($theme);
   @include bf-color-font($theme);
+  @include bf-table-theme($theme);
   @include bf-missing-control-theme($theme);
+  // TODO the two following include can be separate if we have a style file for dashboard and a style file for marketplace
```
GrandSchtroumpf: Not really. Both of them are used in dashboard & marketplace. But world-map is not part of the festival.

PR #2266, opened 2020-03-23
Issue 1198 - dyn title

### LINKED ISSUE #1198: Change page titles according to the page the user is in
- [x] Make some research
- [x] Implement it on marketplace first

Maxstgt:
close #1198


apps/catalog/catalog/src/app/dashboard/title/view/view.component.ts
```diff
@@ -25,12 +26,34 @@ export class TitleViewComponent implements OnInit {
...
   ngOnInit() {
     this.getMovie();
   }
 
+  /**
+* We need to dinstinguish between page load and route change
+* from mat tab component.
+* @param link optional param when the function is getting called from the template 
+*/
+  public refreshTitle(link?: string) {
```
GrandSchtroumpf: Could you do it directly in the ngOnInit with an observable ? 
```typescript
someCustomObservable$.subscribe(title => this.title.setTitle(title))
```

apps/catalog/catalog/src/app/marketplace/deal/list/list.component.html
```diff
@@ -3,7 +3,9 @@ <h1>Offers & Deals</h1>
 <mat-tab-group>
   <mat-tab *ngFor="let tab of tabs$ | async">
     <ng-template mat-tab-label>
-      {{ tab.name }} ({{ tab.contracts.length }})
+      <div (click)="refreshTitle(tab.name)">
```
GrandSchtroumpf: Same as above, if you could put on mat-tab it's better, else could you change div for a span ?

apps/catalog/catalog/src/app/marketplace/title/view/view.component.html
```diff
@@ -41,7 +43,8 @@ <h3>{{ movie.main.title.original }}</h3>
       *ngFor="let link of navLinks"
       [routerLink]="link.path"
       routerLinkActive #rla="routerLinkActive"
-      [active]="rla.isActive">
+      [active]="rla.isActive"
+      (click)="refreshTitle(link.path)">
```
GrandSchtroumpf: Why not managing the title at the child component level ? I don't think it's a good idea to update title on the parent when you click on a link. Same for above.

lsenta:
⚠️ don't merge to master, use develop or release.

Maxstgt:
> It's a good draft. Consider having some title more generic so we can use a simple service. Like that it's going to be more reusable & reused ;).
> A simple pattern would be :
> 
> * list page:  (${notification.count}) ${count ? '' : 'No'} ${entityName}s - ${app}
> * view page: (${notification.count}) ${entityName} ${subpage} - ${app}
> * other: (${notification.count}) ${title} - ${app}

About the naming convention is not my business, the titles are coming from @2amarkina .
About the service, we can implement such service when remote config is ready.

GrandSchtroumpf:
You can start to create a service that we could improve later. 
The goal is to update the title only in the `ngOnInit` of the component (no `refreshTitle` in the HTML)
For example : 
```typescript
const staticTitle = {
  festival: {
    home: 'Home Page',
    movieList: (count) => count ? 'Lot of Movie' : 'No Movie'
  }
}
export class TitleService {
  constructor(private title: Title) {}
  set(app, page) {
    this.title.setTitle(staticTitle[app][page])
  }
  setCount(app, page, count) {
    this.title.setTitle(staticTitle[app][page](count))
  }
}
```
Then in the component use observables when listening on changes of values
```typescript
export class MovieListComponent {
  constructor(private title: TitleService, private query: MovieQuery) {}
  ngOnInit() {
    this.query.selectCount().subscribe(count => this.title.setCount('festivale', 'movieList', count));
  }
}
```

Maxstgt:
I see your point, but this service ist just a wrapper around the title service from angular. And the only benefit I can see is that you store your text some where central, but if we implement remote config sooner or later, this benefit will get obsolet. We have to rewrite this service again or even have to delete it. Because this feature will get handled by transloco, if we using this lib.

apps/catalog/catalog/src/app/marketplace/title/view/view.component.html
Maxstgt: if you doing it in the child you need to import the service multiple time, no need for that

libs/organization/src/lib/pages/view/view.component.ts
```diff
@@ -22,23 +24,39 @@ const   navLinks = [{
...
-    private authQuery: AuthQuery
-  ) { }
+    private authQuery: AuthQuery,
+    private dynTitle: DynamicTitleService,
+    private routerQuery: RouterQuery
+  ) {
+    this.sub = this.routerQuery.select('state').subscribe(data => {
+      if (data.url.includes('members')) {
+        this.dynTitle.setPageTitle('Members', `${this.query.getActive().name}`,
+         { appName: { slug: 'catalog', label: 'Blockframes' }, showAppName: false })
```
GrandSchtroumpf: Hmmm I don't understand why "catalog" should be display differently. If you want to display "Blockframes", create a "blockframes" slug.

libs/user/src/lib/auth/pages/profile-view/profile-view.component.ts
```diff
@@ -27,8 +28,15 @@ export class ProfileViewComponent implements OnInit {
     private organizationQuery: OrganizationQuery,
     private tunnelService: TunnelService,
     private authService: AuthService,
-    private snackBar: MatSnackBar
-  ) { }
+    private snackBar: MatSnackBar,
+    private dynTitle: DynamicTitleService
+  ) {
+    this.dynTitle.setPageTitle(`
+    ${this.authQuery.getValue().profile.surname}
```
GrandSchtroumpf: Could you do that ?
```typescript
const { surname, name } = this.authQuery.getValue().profile;
```

libs/utils/src/lib/dynamic-title/dynamic-title.service.ts.service.ts
```diff
@@ -0,0 +1,85 @@
...
+   * This variable holds the value that this service gets when 
+   * he gets initialized. This is useful as a fallback option 
+   */
+  private initTitle = '';
+
+  private pages: Record<string, any>;
+
+  constructor(private title: Title, private routerQuery: RouterQuery) {
+    this.sub = this.routerQuery.select('state').subscribe(data => {
+      this.setAppName(data.root.data.app)
```
GrandSchtroumpf: Create a simple object above the service:
```typescript
const appNames = {
  catalog: 'Archipel Content',
  festival: 'Archipel Market',
  blockframes: 'Blockframes',
}
type AppName = keyof typeof appNames;
```
And do : 
```typescript
this.routerQuery.select('state').subscribe(data => {
  this.app = appName[data.root.data.app];
})
```

libs/utils/src/lib/dynamic-title/dynamic-title.service.ts.service.ts
```diff
@@ -0,0 +1,85 @@
...
+      case 'catalog':
+        this.app = { slug: 'catalog', label: 'Archipel Content' }
+    }
+  }
+
+  /**
+   * @description
+   * @param section can be the name of a tab or a filter
+   * @param entityWithSection can be movie title name
+   * @param config.appName by default it uses the ActivatedRoute interface to determine the app name,
```
GrandSchtroumpf: appName should be a string litteral with the key of the appNames object (see comment above): 
```typescript
type AppName = keyof typeof appNames;
```

libs/utils/src/lib/dynamic-title/dynamic-title.service.ts.service.ts
```diff
@@ -0,0 +1,85 @@
...
+  private pages: Record<string, any>;
+
+  constructor(private title: Title, private routerQuery: RouterQuery) {
+    this.sub = this.routerQuery.select('state').subscribe(data => {
+      this.setAppName(data.root.data.app)
+    })
+
+    // First, we need the app name
+    this.pages = {
+      catalog: {
```
GrandSchtroumpf: Nothing depends on `this` here. Create an object above your service

libs/organization/src/lib/pages/view/view.component.ts
```diff
@@ -22,23 +24,39 @@ const   navLinks = [{
...
     private query: OrganizationQuery,
     private tunnelService: TunnelService,
-    private service: OrganizationService,
-    private authQuery: AuthQuery
-  ) { }
+    private authQuery: AuthQuery,
+    private dynTitle: DynamicTitleService,
+    private routerQuery: RouterQuery
+  ) {
+    this.sub = this.routerQuery.select('state').subscribe(data => {
```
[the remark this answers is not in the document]
Maxstgt: cause the url change when the tabs are clicked

PR #2278, opened 2020-03-24
Feat 1887 bis - Update version model 

phpgeekfr:
- When a new contract of type mandate is created/updated, the tiltleDetails.price.commissionBase is  set to grossreceipt.
- Add titlesIds to contract.titleIds[] automatically when contract.titles is updated.
- Add partieIds to contract.partieIds[] automatically when contract.parties is updated and when contract status is not Draft and when parties have an orgId defined.
- Clean contract signDate, parties signDate and status when a new version is created.


apps/backend-functions/src/contract.ts
```diff
@@ -115,6 +115,72 @@ async function checkAndTriggerNotifications(current: ContractDocument) {
...
+ * This method is in charge of updating contract document on DB.
+ * It updates some of document attributes.
+ * @param tx 
+ * @param ref 
+ * @param current 
+ */
+function updateContract(tx: FirebaseFirestore.Transaction, ref: DocumentReference, current: ContractDocument, newVersion: boolean = false) {
+
+  // We create an array of title ids for querying purposes
+  current.titleIds = [];
```
GrandSchtroumpf: Maybe rename "current" by "currentContract" (I know we are in the contract file, but still ^^)

apps/backend-functions/src/contract.ts
```diff
@@ -115,6 +115,72 @@ async function checkAndTriggerNotifications(current: ContractDocument) {
...
+      if (p.party.orgId) { // Only if orgId is defined on party
+        current.partyIds.push(p.party.orgId);
+      }
+    })
+  }
+
+  // If true, a new version is beeing created.
+  // We need to remove previous parties signDate and reset status
+  if (newVersion) {
+    delete current.signDate;
```
GrandSchtroumpf: Why you need to do that ? You're not using it later

phpgeekfr: It is for cleaning signDate since a new version is created. If for example a new title is added to contract, previous signatures / status should be removed.

PR #2282, opened 2020-03-25
Implement meta for events

GrandSchtroumpf:
Update form + types for event's meta field


libs/event/src/lib/+state/event.firestore.ts
```diff
@@ -0,0 +1,28 @@
+import { firestore } from 'firebase';
+
+type Timestamp = firestore.Timestamp;
+
+// Event types
+export type EventTypes = 'standard' | 'meeting' | 'screening' | 'local';
+export type EventMeta = Meeting | Screening | {};
+export interface Meeting {
+  callUrl: string;
```
phpgeekfr: I thinks this was for example and it will change soon.

libs/event/src/lib/+state/event.firestore.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+export interface Screening {
+  movieId: string;
```
phpgeekfr: same

PR #2287, opened 2020-03-25
Removing some index.ts files and use deeper imports

### LINKED ISSUE #2230: Ethers is loaded when it should not
When serving the app (catalog/festival) on marketplace or dashboard some libs from ethersprojects are loaded. This is likely an issue of import.
**I think** it comes from the organization-form that import @blockframes/ethers/helpers.
We need to track why the component is loaded while if should not (basically lazy-loading is failing here).
How to see it : comment out the `(window as any).global = window;` in the polyfills and go to marketplace, you should see a trace.
Maybe looking at the "network" tab in the console could help seeing all module loaded.

ioaNikas:
Delete some libs index to avoid importing external libs by mistake

- notification index
- auth index
- utils index
- ui index
- ethers index

Slightly better loading time on landing page

close #2230


libs/user/src/lib/user/account/account.module.ts
```diff
@@ -1,20 +1,17 @@
...
   { path: '',
     children: [
       { path: '', redirectTo: 'profile', pathMatch: 'full' },
       { path: 'profile', loadChildren: () => import('../user-display.module').then(m => m.UserDisplayModule) },
-      {
-        path: 'wallet',
-        canActivate: [ActiveDaoGuard],
-        loadChildren: () => import('@blockframes/ethers').then(m => m.WalletModule)
-      }
+      // {
```
GrandSchtroumpf: Maybe add a comment here to explain

PR #2290, opened 2020-03-25
move function to user.service

Mussuranae:
Move every function that concern user into user.service, deleted them from auth.service


libs/user/src/lib/user/+state/user.service.ts
```diff
@@ -18,15 +21,113 @@ export class UserService extends CollectionService<UserState> {
...
+  // TODO #1389 we can replace a lot of these functions by native Akita-ng-fire functions
+
+    /** Call a firebase function to get or create a user.
+   * @email find the user with this email. If email doesn't match with an existing user,
+   * create a user with this email address.
+   */
+  public async getOrCreateUserByMail(email: string, orgName: string, invitationId?: string): Promise<User> {
+    const f = this.functions.httpsCallable('getOrCreateUserByMail');
+    return f({ email, orgName }).toPromise();
+  }
```
phpgeekfr: I think this one should stay on auth service because the `getOrCreateUserByMail` function called is only handling auth and does not interact with db.

PR #2292, opened 2020-03-25
Feat 1887 ter - front cleaning

phpgeekfr:


apps/catalog/catalog/src/app/dashboard/deal/list/list.component.ts
```diff
@@ -122,11 +121,11 @@ export class DealListComponent {
   async createMandate() {
     const type = 'mandate';
     const orgId = this.orgQuery.getActiveId();
-    const mandate = await this.service.getMandate(orgId);
+    const mandate = await this.contractService.getMandateByOrgId(orgId);
```
GrandSchtroumpf, line 124: Any reason to rename it ? If you give a good name to the params (+ comment) you should need to specify `byOrgId`

apps/catalog/catalog/src/app/dashboard/deal/negotiation/negotiation.component.html
```diff
@@ -44,13 +44,13 @@
   </mat-card>
 </section>
 
-<section *ngIf="oldVersions$ | async as oldVersions">
-  <mat-card *ngIf="oldVersions.length">
+<section *ngIf="historizedVersionsView$ | async as historizedVersions">
```
GrandSchtroumpf, line 47: I think you can sticked with "versions" XD.
Or "negociationHistory". But "versions" is still better I think ^^

apps/catalog/catalog/src/app/dashboard/deal/negotiation/negotiation.component.ts
```diff
@@ -21,7 +21,7 @@ export class NegociationComponent {
   licensees = this.query.getActiveParties('licensee');
   subLicensors$ = this.query.subLicensors$;
 
-  oldVersions$ = this.query.oldVersionsView$;
+  historizedVersionsView$ = this.query.historizedVersionsView$;
```
GrandSchtroumpf, line 24: I think for this case you should use AngularFirestore directly (because this is the only component where you should have access to the versions): 
```typescript
versions$ = this.query.selectActiveId().pipe(
  switchMap(contractId => this.db.collection(`contracts/${contractId }/versions`).valueChanges())
);
```

apps/catalog/catalog/src/app/marketplace/deal/negotiation/negotiation.component.html
```diff
@@ -34,13 +34,13 @@
   </mat-card>
 </section>
 
-<section *ngIf="oldVersions$ | async as oldVersions">
-  <mat-card *ngIf="oldVersions.length">
+<section *ngIf="historizedVersionsView$ | async as historizedVersions">
```
GrandSchtroumpf, line 37: Same as above

libs/admin/src/lib/admin-panel/components/contract-tree/contract-tree.component.ts
```diff
@@ -46,39 +46,38 @@ export class ContractTreeComponent implements OnInit {
...
     const processed = this.tree[contract.id] ? true : false;
     if (!processed) {
       this.tree[contract.id] = {
         id: contract.id,
         contract,
         level,
-        parents: contract.parentContractIds,
-        childs: contract.childContractIds,
+        parents: contract.parentContractIds ? contract.parentContractIds : [],
+        childs: contract.childContractIds ? contract.childContractIds : [],
```
GrandSchtroumpf, line 60: Use `contract.parentContractIds || []`

libs/admin/src/lib/admin-panel/pages/contract/contract.component.html
```diff
@@ -111,16 +111,16 @@ <h2>Create new version</h2>
             </mat-form-field>
           </div>
           <mat-card-actions align="end">
-          <button type="submit" mat-flat-button color="primary" (click)="updateVersion()">Update to v-{{ version +1 }}</button>
+          <button type="submit" mat-flat-button color="primary" (click)="updateVersion()">Update to v-{{ contract.lastVersion.id +1 }}</button>
```
GrandSchtroumpf, line 114: Hmmm are you sure that the id is a number ? I'm pretty sure it's a string

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
```diff
@@ -124,56 +125,37 @@ export class ContractComponent implements OnInit {
...
-      this.contract = await this.contractService.getContractWithLastVersion(this.contractId);
-      this.contractForm = new ContractAdminForm(this.contract.doc);
-      this.contractVersionForm = new ContractVersionAdminForm(this.contract.last);
-      this.version = parseInt(this.contract.last.id, 10);
       this.publicContract$ = this.contractService.listenOnPublicContract(this.contractId);
-      this.contractVersions = await this.contractVersionService.getContractVersions(this.contractId);
 
-      this.loadTitles();
+      this.contract$ = this.contractService.listenOnContract(this.contractId).pipe(map(c => {
+        this.contract = c;
```
GrandSchtroumpf, line 131: You should avoid to mutate the state inside a subsription that is aims to do something else.
I think there is a design issue here on what should be reactive or not.
Like you shouldn't set "this.contract" if you have an observable "this.contract$" for that.

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
```diff
@@ -124,56 +125,37 @@ export class ContractComponent implements OnInit {
...
       this.contractId = params.contractId;
-      this.contract = await this.contractService.getContractWithLastVersion(this.contractId);
-      this.contractForm = new ContractAdminForm(this.contract.doc);
-      this.contractVersionForm = new ContractVersionAdminForm(this.contract.last);
-      this.version = parseInt(this.contract.last.id, 10);
       this.publicContract$ = this.contractService.listenOnPublicContract(this.contractId);
-      this.contractVersions = await this.contractVersionService.getContractVersions(this.contractId);
 
-      this.loadTitles();
+      this.contract$ = this.contractService.listenOnContract(this.contractId).pipe(map(c => {
```
GrandSchtroumpf, line 130: what's the difference with `syncDoc(this.contractId)` ?

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
```diff
@@ -124,56 +125,37 @@ export class ContractComponent implements OnInit {
...
+        this.contract = c;
+        this.contractForm = new ContractAdminForm(this.contract);
+        this.contractVersionForm = new ContractVersionAdminForm(this.contract.lastVersion);
+        this.cdRef.markForCheck();
+        this.titles = [];
```
GrandSchtroumpf, line 135: Consider using map instead of mutating an array. immutable modification are easier to be detected by angular (then more performant)

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
```diff
@@ -124,56 +125,37 @@ export class ContractComponent implements OnInit {
...
+        if (c.titleIds && c.lastVersion.titles) {
+          c.titleIds.forEach(async titleId => {
+            const movie = await this.movieService.getValue(titleId);
```
GrandSchtroumpf, line 138: Blocking the thread for each doesn't seems like a good idea

libs/admin/src/lib/admin-panel/pages/contracts/contracts.component.ts
```diff
@@ -55,13 +51,11 @@ export class ContractsComponent implements OnInit {
     }
 
     const promises = contracts.map(async contract => {
-      const contractWithLastVersion = await this.contractService.getContractWithLastVersion(contract.id);
-      const row = { ...contractWithLastVersion } as any;
-
+      const row = { ...contract } as any;
```
GrandSchtroumpf, line 54: map create a copy of each item, you shouldn't need to copy it again.

libs/contract/src/lib/contract/+state/contract.model.ts
```diff
@@ -12,33 +12,24 @@ import {
...
- * used only in front
- */
-export interface ContractWithLastVersion { // @TODO (#1887) remove this
-  doc: Contract,
-  last: ContractVersion,
-}
-
 export interface Contract extends ContractDocumentWithDates {
-  versions?: ContractVersion[];
+  historizedVersions?: ContractVersion[];
```
GrandSchtroumpf, line 26: I think you can go with "versions" directly

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -46,113 +47,119 @@ export class ContractService extends CollectionService<ContractState> {
...
-  /** Get the mandate contract of an organization */
-  public async getMandate(orgId: string) {
-    const query = ref => ref.where('partyIds', 'array-contains', orgId).where('type', '==', 'sale');
-    const mandates = await this.getValue(query);
-    return (mandates && mandates.length) ? mandates[0] : undefined;
+  /**
+   * Gets the mandate contract of an organization
+   * @param orgId 
+   */
+  public async getMandateByOrgId(orgId: string): Promise<Contract> {
```
GrandSchtroumpf, line 60: You can use `limit(1)`
Maybe I'm a little pushy with the DRY here ^^, but here is another suggestion : 
```typescript
/**
 * @example
 *```typescrpit
 * service.getMandate(ref => ref.where('partyIds', 'array-contains', orgId) // mandateByOrgId
 * service.getMandate(ref => ref.where('titleIds', 'array-contains', movieId) // mandateByMovieId
 */
async getMandate(queryFn: QueryFn) {
  const query = ref => queryFn(ref).where('partyIds', 'array-contains', orgId).where('type', '==', 'mandate').limit(1);
  const contracts = await this.getValue(query);
  return contracts[0];
}
```
Like that you can do : `service.getMandate(ref => ref.where('partyIds', 'array-contains', orgId))`

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -46,113 +47,119 @@ export class ContractService extends CollectionService<ContractState> {
...
-        }
-      }
-    }
+  public async getContractFromDeal(movieId: string, distributionDealId: string): Promise<Contract> {
+    const contracts = await this.getValue(ref =>
+      ref.where(`lastVersion.titles.${movieId}.distributionDealIds`, 'array-contains', distributionDealId),
+    );
+
+    // Can have only one result
+    return contracts.length ? contracts.pop() : undefined;
```
GrandSchtroumpf, line 91: Maybe use "limit(1)" to not get more than we need

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -46,113 +47,119 @@ export class ContractService extends CollectionService<ContractState> {
...
-      ? createVersionMandate({ id: '1', ...version })
-      : createContractVersion({ id: '1', ...version });
-    // Create contract + verions + _meta version
-    const contractId = await this.add(_contract, { write });
-    this.contractVersionService.add(_version, { params: { contractId }, write });
-    this.contractVersionService.add({ id: '_meta', count: 1 }, { params: { contractId }, write });
-    await write.commit();
-    return contractId;
+  public async create(contract: Partial<Contract>): Promise<string> {
+    return await this.add(createContract({ ...contract, }));
```
GrandSchtroumpf, line 101: `return this.add(createContract(contract));`

apps/catalog/catalog/src/app/dashboard/deal/list/list.component.ts
phpgeekfr: Because there is another method `getMandateByMovieId`. Don't know how to create method with same name but different params

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -46,113 +47,119 @@ export class ContractService extends CollectionService<ContractState> {
...
+      licensor.party.orgId = centralOrgID
+      licensor.party.role = 'licensor';
+      contract.parties.push(licensor);
+    }
+
+    for (const titleId of Object.keys(titlesAndDeals)) {
+      const deals = titlesAndDeals[titleId];
+      const dealDocs = deals.map(d => {
+        const deal = createDistributionDeal({ ...d, contractId: contract.id })
+        if (!deal.id) { deal.id = objectHash(deal) };
```
GrandSchtroumpf, line 139: Any reason to use objectHash instead of createId ? If you don't really need the hash, consider using the createId instead.

apps/catalog/catalog/src/app/dashboard/deal/negotiation/negotiation.component.html
phpgeekfr: Like that it is clear that we are not talking about lastVersion but only archived ones

libs/contract/src/lib/contract/+state/contract.service.ts
GrandSchtroumpf: Also akita-ng-fire will take care of this normally. So you could just do : 
```typescript
for (const movieId of Object.keys(titlesAndDeals)) {
const deals = titlesAndDeals[movieId];
const dealDocs = deals.map(d => createDistributionDeal({ ...d, contractId: contract.id }));
const distributionDealIds= await this.distributionDealService.add(dealDocs, { params: { movieId }, write });
contract.lastVersion.titles[titleId] = createContractTitleDetail({ distributionDealIds, titleId });
}
```

libs/admin/src/lib/admin-panel/pages/contract/contract.component.html
phpgeekfr: I changed it to a number

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -46,113 +47,119 @@ export class ContractService extends CollectionService<ContractState> {
...
+        return deal;
+      });
+      const distributionDealIds = await this.distributionDealService.add(dealDocs, { params: { movieId: titleId }, write });
+      contract.lastVersion.titles[titleId] = createContractTitleDetail({ distributionDealIds, titleId: titleId });
+    }
+
+    // Add contract
+    const isValid = await this.validateAndConsolidateContract(contract);
+    if (isValid) {
```
GrandSchtroumpf, line 148: Could you not check the inValidity before preparing the batch ?

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -187,42 +194,18 @@ export class ContractService extends CollectionService<ContractState> {
...
-     */
-    if (parentContracts.length === 0) {
-      const promises = contract.parentContractIds.map(id => this.getValue(id));
-      if (promises.length) {
-        parentContracts = await Promise.all(promises);
+    // Fetch displayName for parties that have orgId defined
+    for (const p of contract.parties) {
+      if (p.party.orgId) {
+        const org = await this.organizationService.getValue(p.party.orgId);
+        p.party.displayName = org.denomination.publicName ? org.denomination.publicName : org.denomination.full;
```
GrandSchtroumpf: Use "||"

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -273,7 +256,13 @@ export class ContractService extends CollectionService<ContractState> {
     this.update({ ...contract, parties: [...updatedParties, updatedParty] })
   }
 
-  public listenOnPublicContract(contractId: string) : Observable<PublicContract>{
+  public listenOnContract(contractId: string): Observable<Contract> {
```
GrandSchtroumpf, line 259: Any reason not to use `syncDoc(contractId)` and get the value from the Query ?

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -300,4 +289,24 @@ export class ContractService extends CollectionService<ContractState> {
...
+   * Changes contract status to submitted. Also updates distribution deals state.
+   * @param _contract 
+   */
+  public async submit(_contract: Contract) {
+    const write = this.db.firestore.batch();
+    const contract = { ..._contract };
+    contract.lastVersion = { ...contract.lastVersion, status: 'submitted' }; // For read-only purposes
+    this.update(contract, { write });
+
+    Object.keys(contract.lastVersion.titles).forEach(titleId => {
```
GrandSchtroumpf, line 302: You're using `for (const ... in ...)` for most of your PR. Many keep with this for consistency

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -300,4 +289,24 @@ export class ContractService extends CollectionService<ContractState> {
     return contractsSnap.docs.map(c => createContractFromFirestore(c.data()));
   }
 
+  /**
+   * Changes contract status to submitted. Also updates distribution deals state.
+   * @param _contract 
+   */
+  public async submit(_contract: Contract) {
+    const write = this.db.firestore.batch();
+    const contract = { ..._contract };
```
GrandSchtroumpf, line 298: I don't underdant why you nee to copy here. You could mutate the status of the lastVersion of the contract given as parameter no ?

libs/contract/src/lib/contract/guards/active-contract.guard.ts
```diff
@@ -5,10 +5,10 @@ import { ContractState, ContractStore, ContractWithTimeStamp, ContractService }
 import { tap } from 'rxjs/operators';
 
 /** Get the active contract and put his lastVersion in it. */
-// todo(#1887) remove versions in query
 const contractQuery = (contractId: string): Query<ContractWithTimeStamp> => ({
   path: `contracts/${contractId}`,
-  versions: contract => ({
+  /** @dev This is used to fetch all archived versions along with contract (KFH) */
+  historizedVersions: contract => ({
```
GrandSchtroumpf, line 11: Same as above

libs/contract/src/lib/version/+state/contract-version.service.ts
```diff
@@ -1,85 +1,41 @@
...
-    const contractsSnap = await this.db
-      .collection(`contracts/${contractId}/versions`)
-      .get()
-      .toPromise();
-    return contractsSnap.docs.filter(v => v.id !== '_meta').map(c => createContractVersionFromFirestore(c.data()));
+  public listenOnContractVersions(contractId: string): Observable<ContractVersion[]> {
+    return this.db.collection(`contracts/${contractId}/versions`)
+      .valueChanges()
+      .pipe(
+        map((versions : ContractVersion[]) => versions.filter(v => v.id !== undefined ).map(v => createContractVersionFromFirestore(v)))
```
GrandSchtroumpf, line 38: I don't understand why you need to filter on undefined ids. 
You can do `valueChanges({ idField: 'id' })` to force firestore to the the id in the key "id", like that you're sure that you always have a unique id.

apps/catalog/catalog/src/app/dashboard/deal/negotiation/negotiation.component.ts
phpgeekfr: In fact there is two components using this

libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts
phpgeekfr: Yes, the difference is that `createContractFromFirestore` is called on listening. And i think that syncDoc does not dot it yet

libs/admin/src/lib/admin-panel/pages/contracts/contracts.component.ts
phpgeekfr: I need to cast it to any because I add data to row that is not part of contract model

libs/contract/src/lib/contract/+state/contract.service.ts
phpgeekfr: Because the incoming _contract is readonly. same for .lastVersion

libs/contract/src/lib/version/+state/contract-version.service.ts
phpgeekfr: Good catch. It is because we used _meta.count and this object did not had any id

libs/admin/src/lib/admin-panel/pages/contracts/contracts.component.ts
GrandSchtroumpf: What about `.map((row: any) => { ... })`

PR #2294, opened 2020-03-26
Festival search filter clean

pldespaigne:
- [x] add `orgName` to the algolia movie index
- [x] create a seller filter linkt to the algolia movie index
- [x] refactor the genres filter to use algolia
- [x] refactor the country filter to use algolia
- [x] refactor the language filter to use algolia
  - [x] refactor language key in the algolia movie index
- [x] refactor the status filter to use algolia
- [x] refactor the budget filter to use algolia
  - [x] create the budget key in the algolia movie index
- [x] add search by keywords
- [x] clean the movie index from useless data
- [x] add auto-configuration of index to the backend-ops (see #2299)
- [x] prevent empty page at the beginning (display some movies from the store)
- [x] clean the code

### DIFF
--- a/apps/backend-functions/src/internals/algolia.ts
+++ b/apps/backend-functions/src/internals/algolia.ts
@@ -1,6 +1,7 @@
 import algoliasearch from 'algoliasearch';
 import { algolia } from '../environments/environment';
-import pick from 'lodash/pick';
+import { MovieDocument } from '../data/types';
+import { LanguagesSlug } from '@blockframes/utils/static-model';
 
 const indexOrganizationsBuilder = (adminKey?: string) => {
   const client = algoliasearch(algolia.appId, adminKey || algolia.adminKey);
@@ -33,8 +34,7 @@ const indexMoviesBuilder = (adminKey?: string) => {
 };
 
 export function storeSearchableMovie(
-  movie: FirebaseFirestore.DocumentData,
-  orgId: string,
+  movie: MovieDocument,
   orgName: string,
   adminKey?: string
 ): Promise<any> {
@@ -42,24 +42,51 @@ export function storeSearchableMovie(
     console.warn('No algolia id set, assuming dev config: skipping');
     return Promise.resolve(true);
   }
-  const ALGOLIA_FIELDS: string[] = [
-    'main.genres',
-    'main.title.international',
-    'main.title.original',
-    'main.directors',
-    'main.originalLanguages',
-    'main.status',
-    'main.originCountries',
-    'promotionalDescription.keywords',
-    'salesAgentDeal.salesAgent.displayName',
-    'versionInfo.dubbings',
-    'versionInfo.subtitles'
-  ];
+
+  // TODO un-nest the data
+  // const ALGOLIA_FIELDS: string[] = [
+    // 'main.genres',
+    // 'main.title.international',
+    // 'main.title.original',
+    // 'main.directors', // TODO clean useless fields in this array
+    // 'main.originalLanguages',
+    // 'main.status',
+    // 'main.originCountries',
+    // 'promotionalDescription.keywords',
+    // 'salesAgentDeal.salesAgent.displayName',
+    // 'versionInfo.dubbings', // TODO implement correctly the languages
+    // 'versionInfo.subtitles', // TODO implement correctly the languages
+    // 'budget.estimatedBudget.from',
+    // 'budget.estimatedBudget.to'
+  // ];
   return indexMoviesBuilder(adminKey).saveObject({
     objectID: movie.id,
-    movie: pick(movie, ALGOLIA_FIELDS),
-    orgId,
-    orgName
+    // movie: pick(movie, ALGOLIA_FIELDS),
+
+    // searchable keys
+    title: {
+      international: movie.main.title.international,
+      original: movie.main.title.original,
+    },
+    directors: movie.main.directors!.map((director) => `${director.firstName} ${director.lastName}`),
+    keywords: movie.promotionalDescription.keywords,
+
+    genres: movie.main.genres,
+    originCountries: movie.main.originCountries,
+    languages: {
+      original: movie.main.originalLanguages,
+      dubbed: Object.keys(movie.versionInfo.languages).filter(lang => movie.versionInfo.languages[lang as LanguagesSlug]?.dubbed),
+      subtitle: Object.keys(movie.versionInfo.languages).filter(lang => movie.versionInfo.languages[lang as LanguagesSlug]?.subtitle),
+      caption: Object.keys(movie.versionInfo.languages).filter(lang => movie.versionInfo.languages[lang as LanguagesSlug]?.caption),
+    },
+    status: movie.main.status,
+    budget: {
+      from: movie.budget.estimatedBudget!.from,
+      to: movie.budget.estimatedBudget!.to,
+    },
+    orgName,
+
+    // orgId,
   });
 }
 
--- a/apps/backend-functions/src/internals/algolia.ts
+++ b/apps/backend-functions/src/internals/algolia.ts
@@ -43,25 +43,8 @@ export function storeSearchableMovie(
     return Promise.resolve(true);
   }
 
-  // TODO un-nest the data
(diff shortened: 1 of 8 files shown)


apps/backend-functions/src/internals/algolia.ts
```diff
@@ -33,29 +34,41 @@ const indexMoviesBuilder = (adminKey?: string) => {
...
   return indexMoviesBuilder(adminKey).saveObject({
     objectID: movie.id,
-    movie: pick(movie, ALGOLIA_FIELDS)
+
+    // searchable keys
+    title: {
+      international: movie.main.title.international,
+      original: movie.main.title.original,
+    },
+    directors: movie.main.directors!.map((director) => `${director.firstName} ${director.lastName}`),
```
GrandSchtroumpf: Are you sure you always have a directors list ?

apps/backend-functions/src/internals/algolia.ts
```diff
@@ -33,29 +34,41 @@ const indexMoviesBuilder = (adminKey?: string) => {
...
+    keywords: movie.promotionalDescription.keywords,
+
+    // facets
+    genres: movie.main.genres,
+    originCountries: movie.main.originCountries,
+    languages: {
+      original: movie.main.originalLanguages,
+      dubbed: Object.keys(movie.versionInfo.languages).filter(lang => movie.versionInfo.languages[lang as LanguagesSlug]?.dubbed),
```
GrandSchtroumpf: do `filter((lang: LanguagesSlug) => ..` instead of using "as", if possible.

apps/backend-functions/src/movie.ts
```diff
@@ -167,7 +167,13 @@ export async function onMovieUpdate(
     return triggerNotifications(notifications);
   }
 
-  return storeSearchableMovie(after);
+  // insert orgName & orgID to the algolia movie index (this is needed in order to filter on the frontend)
+  const creatorSnapshot = await db.doc(`users/${after._meta!.createdBy}`).get();
```
GrandSchtroumpf: Except if I'm wrong, but I thing you can set the data you'll get as a generic parameter for "doc". So you don't have to use "as" later: 
```typescript
db.doc<PublicUser>(...)
```

apps/backend-ops/src/algolia.ts
```diff
@@ -23,7 +24,14 @@ export async function upgradeAlgoliaMovies() {
 
   const promises = [];
   movies.forEach(movie => {
```
GrandSchtroumpf: Use a map : 
```typescript
const promises = movies.map(async movie => {
  const userSnap = await db.doc<User>(`users/${movieData._meta.createdBy}`).get();
  const orgSnap = await db.doc<Organization>(orgs${userSnap.date().orgId}`).doc(user.orgId).get();
  return storeSearchableMovie(movieData, orgSnap.date().name, process.env['ALGOLIA_API_KEY']);
});
return Promise.all(promises);
```

apps/festival/festival/src/app/marketplace/title/list/list.component.html
```diff
@@ -47,117 +42,140 @@ <h6>Filters</h6>
...
-                    [disabled]="!languageControl.valid">
-                    <mat-icon>search</mat-icon>
+            <!-- Languages -->
+            <mat-expansion-panel>
+              <mat-expansion-panel-header>
+                <span class="mat-body-1">Language & Version</span>
+              </mat-expansion-panel-header>
+              <form formGroupName="languages" fxLayout="column">
+                <div fxLayout="row" fxLayoutAlign="start start">
+                  <mat-form-field appearance="outline">
```
GrandSchtroumpf: Is there no reusable component for this ? Something like static-autocomplete
@Maxstgt I think you worked on something like that no ?

apps/festival/festival/src/app/marketplace/title/list/list.component.html
```diff
@@ -47,117 +42,140 @@ <h6>Filters</h6>
...
-            <mat-expansion-panel-header>
-              <span class="mat-body-1">Budget</span>
-            </mat-expansion-panel-header>
-            <div fxLayout="column" fxLayoutAlign="start start" class="checkbox-padding">
-              <mat-checkbox (change)="hasBudget(budget)" color="primary" *ngFor="let budget of budgetList">
-                {{ budget.label }}
-              </mat-checkbox>
-            </div>
-          </mat-expansion-panel>
+            <!-- Sellers -->
```
GrandSchtroumpf: Add a large comment : 
```html
<!-- SPECIFIC TO FESTIVAL APPLICATION -->
```
Like that we know what we can remove when we build a reusable movie-list layout.

libs/movie/src/lib/distribution-deals/form/search.form.ts
```diff
@@ -134,7 +136,7 @@ function createCatalogSearchControl(search: CatalogSearch) {
   return {
     productionYear: createTermsControl(search.productionYear),
     genres: new FormStaticArray(search.genres, 'GENRES', [Validators.required]),
-    status: new FormControl(search.status),
+    productionStatus: new FormControl(search.productionStatus),
```
GrandSchtroumpf: Did you talk with @phpgeekfr & @Mussuranae about this change in the data structure ?

libs/movie/src/lib/distribution-deals/form/search.form.ts
```diff
@@ -201,16 +216,16 @@ export class CatalogSearchForm extends FormEntity<CatalogSearchControl> {
         `Production status ${status} is not part of the defined status, here is the complete list currently available: ${MOVIE_STATUS_SLUG}`
       );
     } else {
-      this.get('status').setValue([...this.get('status').value, status]);
+      this.get('productionStatus').setValue([...this.get('productionStatus').value, status]);
```
GrandSchtroumpf: I don't see the point of creating a getter if you don't use it...
It looks like the "productionStatus" is an array of string. Did you consider using a FormList ?
Here is the rule :
- if control is "multiple select", "autocomplete chips", "toggle button": FormControl(string[])
- if control is a list of input/select/autocomplete: FormList

libs/movie/src/lib/distribution-deals/form/search.form.ts
```diff
@@ -201,16 +216,16 @@ export class CatalogSearchForm extends FormEntity<CatalogSearchControl> {
...
     }
   }
 
   removeStatus(status: MovieStatusSlug) {
     if (MOVIE_STATUS_SLUG.includes(status)) {
-      const newControls = this.get('status').value.filter(
+      const newControls = this.get('productionStatus').value.filter(
```
GrandSchtroumpf: Use your getter.

libs/movie/src/lib/distribution-deals/form/search.form.ts
pldespaigne: status was a conflicting name in form with the validation status

apps/backend-functions/src/movie.ts
pldespaigne: Nope, doing that throw me an error : `Expected 0 type arguments, but got 1`

apps/backend-functions/src/internals/algolia.ts
pldespaigne: I have tried that, but I get a huuuge `No overload matches this call` error

apps/backend-ops/src/algolia.ts
pldespaigne: `movies` is not an array but a `QuerySnapshot`, so it doesn't have a `map` method

libs/movie/src/lib/distribution-deals/form/search.form.ts
pldespaigne: this control is a "multiple select" check boxes, so I should leave it like that as a `FormControl` right ?

apps/festival/festival/src/app/marketplace/title/list/list.component.html
pldespaigne: I'll see with Max to do that in an other PR

PR #2299, opened 2020-03-26
algolia auto config

pldespaigne:
Add auto-configuration of the Algolia movie index to the backend-ops


pldespaigne:
related to #2294

apps/backend-functions/src/internals/algolia.ts
```diff
@@ -33,6 +33,39 @@ const indexMoviesBuilder = (adminKey?: string) => {
   return client.initIndex(INDEX_NAME_MOVIES);
 };
 
+export function setMovieConfiguration(adminKey?: string): Promise<any> {
```
GrandSchtroumpf: Add comments to explain what this function is doing & what is adminKey

PR #2321, opened 2020-03-27
migrations: add missing v12

lsenta:


lsenta:
When this is merged, could you remove the v2 dataset in blockframes-ci?

PR #2324, opened 2020-03-29
Add animations

GrandSchtroumpf:
Create animations for : 
- having local header on a page.
- transition between pages.

### DIFF
--- a/apps/festival/festival/src/app/marketplace/event/list/list.component.ts
+++ b/apps/festival/festival/src/app/marketplace/event/list/list.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, OnDestroy, ChangeDetectionStrategy } from '@angular/core';
+import { Component, OnInit, OnDestroy, ChangeDetectionStrategy, ViewChild, ElementRef } from '@angular/core';
 import { Subscription } from 'rxjs';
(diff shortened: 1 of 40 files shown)


apps/festival/festival/src/app/marketplace/event/list/list.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, OnInit, OnDestroy, ChangeDetectionStrategy } from '@angular/core';
+import { Component, OnInit, OnDestroy, ChangeDetectionStrategy, ViewChild, ElementRef } from '@angular/core';
```
pldespaigne: I might be wrong, but I think those are unused imports

libs/movie/src/lib/movie/layout/view/view.component.ts
```diff
@@ -1,8 +1,10 @@
-import { Component, Input, ChangeDetectionStrategy, Directive, ViewEncapsulation } from '@angular/core';
+import { Component, Input, ChangeDetectionStrategy, Directive, ViewEncapsulation, ViewChild, ElementRef } from '@angular/core';
```
pldespaigne: same thing

PR #2332, opened 2020-03-30
User migration

### LINKED ISSUE #2281: User data rework
Rename fields 'surname' and 'name' in 'firstname' and 'lastname' for user.

Mussuranae:
Change on user collection : firstname and lastname instead of name and surname, to get more consistency in the DB
Fix #2281

### DIFF
--- a/apps/backend-ops/src/firestoreMigrations/0014.ts
+++ b/apps/backend-ops/src/firestoreMigrations/0014.ts
@@ -1,0 +1,29 @@
+import { Firestore } from '../admin';
+
+
+/**
+ * Update name and surname in user documents
+ */
+export async function upgrade(db: Firestore) {
+  const users = await db.collection('users').get();
+
+  const newUserData = users.docs.map(async (userDocSnapshot: any): Promise<any> => {
+    const userData = userDocSnapshot.data();
+    const { name, surname } = userData;
+
+    if (userData.name && userData.surname) {
+      delete userData.name;
+      delete userData.surname;
+
+      const newData = {
+        ...userData,
+        firstName: name,
+        lastName: surname
+      };
+      return userDocSnapshot.ref.set(newData);
+    }
+  });
+
+  await Promise.all(newUserData);
+  console.log('Updating user collection done.');
+}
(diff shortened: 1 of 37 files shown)


apps/backend-ops/src/firestoreMigrations/0014.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Update name and surname in user documents
+ */
+export async function upgrade(db: Firestore) {
+  const users = await db.collection('users').get();
+
+  const newUserData = users.docs.map(async (userDocSnapshot: any): Promise<any> => {
+    const userData = userDocSnapshot.data();
+    const { name, surname } = userData;
+
+    if (userData.name && userData.surname) {
```
phpgeekfr: What if only one is defined ? We will still keep the other one on DB.
You should check theme separatly

libs/organization/src/lib/organization/+state/organization.firestore.ts
```diff
@@ -25,7 +25,7 @@ interface OrganizationRaw<D> {
   cart: CatalogCart[];
   created: D;
   denomination: Denomination;
-  description: string;
+  description?: string;
```
phpgeekfr: why this change ? (update draw.io if need to keep this)

libs/organization/src/lib/organization/+state/organization.firestore.ts
```diff
@@ -12,7 +12,7 @@ interface AppAccess {
 
 interface Denomination {
   full: string;
-  publicName?: string;
+  public?: string;
```
phpgeekfr: Don't forget to update draw.io

phpgeekfr: and migrations ?

libs/organization/src/lib/organization/+state/organization.firestore.ts
Mussuranae: As no one org has a description currently, I though to let this one no mandatory for now, or I create one by default for every organization.

libs/organization/src/lib/organization/+state/organization.firestore.ts
Mussuranae: It's because I put only "public" on the migration that I change that in this PR -_- My bad for this one

apps/backend-ops/src/firestoreMigrations/0014.ts
Mussuranae: Sure ! I will update that !

PR #2340, opened 2020-03-31
Get rid of timestamp in maintenance mode

lsenta:
Problem:
when we restore / upgrade the db we delete, recreate, and update a lot of documents.
It triggers all the functions' onCreate, onDelete, onUpdate, etc.
Which triggers notifications among other things.

We want to disable the functions for the time of the restore / migrations. BUT some functions are triggered a long time after the migration is complete (because they were cold and firebase took some time to start them).

This patch:
- replace the maintenance time with a system that checks the event timestamp.
- if the event occurred within a function, it'll skip it.
- even if the function is trigger AFTER the restore was completed.

We still have one issue: if a function is triggered before a maintenance mode.

Say:
a user updates a movie,
the upgrade starts,
the function starts,
=> this cloud functions might be skipped or run on the wrong data version.

The solution is to disable the app a few minutes BEFORE starting the upgrade / migration process. So that no events can be produced right before the migrate. So basically we'll always have to wait a few minutes for google cloud functions. Whether it's at the start or the end of the process.


lsenta:
Assigning Bruce since he's the most experienced of the reviewer list.

lsenta:
no tested, careful

PR #2346, opened 2020-03-31
Extract wishlist btn

pldespaigne:
- [x] move wishlist-button to `libs/organization/src/lib/organization/components`
- [x] updating ref to the button
- [x] add a flag to display the button as a heart icon button instead of the big outlined version


libs/organization/src/lib/organization/components/wishlist-button/wishlist-button.component.ts
```diff
@@ -18,6 +18,7 @@ export class WishlistButtonComponent implements OnInit {
   toggle$: Observable<boolean>;
 
   @Input() movieId: string;
+  @Input() small = false;
```
GrandSchtroumpf: Use `coerseBooleanProperty` from @angular/cdk : 
```typescript
_small: boolean;
@Input()
get small() {
  return this._small;
}
set small(isSmall: boolean) {
  this._small = coersBooleanProperty(isSmall);
}
```
Like that we can do : 
```html
<wishlist-button small></wishlit-button>
```

PR #2399, opened 2020-03-31
Issue#2308 cropper

ioaNikas:
Update cropper ratios to be easier to maintain


libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
```diff
@@ -12,6 +12,14 @@ import { sanitizeFileName } from '@blockframes/utils/file-sanitizer';
 
 type CropStep = 'drop' | 'crop' | 'upload' | 'upload_complete' | 'show';
 
+const CROPPER_RATIOS = {
```
GrandSchtroumpf: Keep regular naming convention: `mediaRatio`.

libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
```diff
@@ -12,6 +12,14 @@ import { sanitizeFileName } from '@blockframes/utils/file-sanitizer';
...
+  avatar: 1/1,
+  logo: 1/1,
```
GrandSchtroumpf: You shouldn't have twice the same ratio.

libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
```diff
@@ -12,6 +12,14 @@ import { sanitizeFileName } from '@blockframes/utils/file-sanitizer';
 
 type CropStep = 'drop' | 'crop' | 'upload' | 'upload_complete' | 'show';
 
+const CROPPER_RATIOS = {
+  avatar: 1/1,
```
GrandSchtroumpf: I don't really like the sementics here.

PR #2428, opened 2020-04-01
Feat 2244 - Invitations to event

phpgeekfr:


apps/backend-functions/src/permissions.ts
```diff
@@ -11,11 +11,11 @@ import { db, functions } from './internals/firebase';
  * This index might be used for other features, will come later.
  */
 export async function onDocumentPermissionCreate(
-  snapshot: FirebaseFirestore.DocumentSnapshot,
+  _: FirebaseFirestore.DocumentSnapshot,
```
Mussuranae, line 14: just an underscore ? Oo'

libs/event/src/lib/+state/event.service.ts
```diff
@@ -33,4 +38,39 @@ export class EventService extends CollectionService<EventState> {
...
+    // It will check if userId is allowed to push data
+    // (ie: owner of event)
+    // If yes, it will save the url into docIndex collection with docId==eventId
+    const f = this.functions.httpsCallable('setEventUrl');
+    const eventConfig = createPrivateEventConfig({ url });
+    return f({ eventId, eventConfig, userId }).toPromise();
+  }
+
+  /**
+   * Set event private url
```
Mussuranae, line 59: Get*

apps/backend-functions/src/permissions.ts
GrandSchtroumpf: It's a convention to put an underscore when you don't use the first parameter but you need the second. It'll sut the linter up.

libs/event/src/lib/+state/event.service.ts
```diff
@@ -33,4 +38,39 @@ export class EventService extends CollectionService<EventState> {
...
     return syncQuery.call(this, screeningsQuery(queryFn));
   }
+
+  /**
+   * Set event private url
+   * @param eventId 
+   * @param url 
+   * @param userId
+   */
+  public setEventUrl(eventId: string, url: string, userId: string): Promise<any> {
```
GrandSchtroumpf, line 48: I don't see the point of using the cloud function for that. You can let the admin of the event write on the document.

libs/event/src/lib/+state/event.service.ts
```diff
@@ -33,4 +38,39 @@ export class EventService extends CollectionService<EventState> {
...
+    const eventConfig = createPrivateEventConfig({ url });
+    return f({ eventId, eventConfig, userId }).toPromise();
+  }
+
+  /**
+   * Get event private url
+   * @param eventId 
+   * @param userId 
+   */
+  public getEventUrl(eventId: string, userId: string): Promise<string> {
```
GrandSchtroumpf, line 63: Same, you could use a normal snapshot. The most important is the rules.

libs/notification/src/lib/invitation/+state/invitation.firestore.ts
```diff
@@ -22,21 +27,54 @@ export type InvitationOrUndefined = InvitationDocument | undefined;
...
+ * @see #2244 When a invitation is created
+ * A backend function will check if:
+ * If we have an user or an org we can create a notification.
+ * If we have an email, the function will send an email.
+ * If user created an account, we will then be able to replace email by user
+ * */
+export interface InvitationToAnEvent extends Invitation {
+  mode: InvitationMode,
+  type: 'event';
+  fromOrg?: PublicOrganization,
```
GrandSchtroumpf, line 51: I'm still not 100% convinced, but I don't have better in mind XD. So let's go for that

libs/event/src/lib/+state/event.service.ts
phpgeekfr: If we let the admin write on docsIndex we need to make rules for that but it was intended to be internal only. I think it will be easier this way but we can discuss it.

libs/event/src/lib/+state/event.service.ts
phpgeekfr: How can I use rules to do this ? A rule can only check if you have access to a document or not before displaying it, it cannot send data back to user, only check.
The rule will be much harder to setup than a https callback but we can discuss it

phpgeekfr: If we do It like you want, we need to setup a service for docIndex collection. And what if user need to get the URL but cannot access other private field?
With my way, there is no such potential issue and the backend query for checking and returning url will be much easier to setup than a in rule.

libs/event/src/lib/+state/event.service.ts
phpgeekfr: OK, just looked at the existing rules for DocIndex:
```
    match /docsIndex/{docId} {
      // index of every doc create with a permission doc associated:
      // essential for isBrandNewDocument.
      allow read, write: if false;
    }
```
This mean we cannot use docsIndex since it must no be readble

phpgeekfr: So we can make a regular call for this like you said. But we need to setup a whole service for this ? My opinion is that a https callable is still better cuz we can integrate it to existing eventService ;)

libs/event/src/lib/+state/event.service.ts
phpgeekfr: Event with my new comment above, I still think that a function for this is much easier to do with https callable than in the rule.

libs/event/src/lib/+state/event.service.ts
phpgeekfr: Sorry I'm not well awaken yet ^^ The current rules for DocIndex is what we need

libs/notification/src/lib/notification/notification-widget/notification-widget.component.ts
```diff
@@ -53,10 +53,7 @@ export class NotificationWidgetComponent implements OnInit {
   }
 
   private memberInvitations(invitation: Invitation) {
-    return (
-      invitation.status === 'pending' &&
-      invitation.type === 'toWorkOnDocument'
-    );
+    return invitation.status === 'pending';
```
ioaNikas, line 56: I did something wrong here as I wanted to prevent organization common members to have access to invitations from users or from their organization.
We should add another condition `invitation.type !== 'fromUserToOrganization` to fix this.

libs/user/src/lib/user/index.ts
```diff
@@ -2,6 +2,7 @@ export * from './account/account.module';
 export * from './+state/user.store';
 export * from './+state/user.service';
 export * from './+state/user.query';
+export * from './+state/user.firestore';
```
ioaNikas, line 5: We can get rid of this index.ts since there is no path to it in `tsconfig.json` and we don't seem to use it elsewhere. You should check if it is the case.

libs/notification/src/lib/notification/notification-widget/notification-widget.component.ts
phpgeekfr: I had a doubt about this ! Thanks

libs/admin/src/lib/admin-panel/pages/event/event.component.ts
```diff
@@ -0,0 +1,99 @@
...
+    private eventService: EventService,
+    private movieService: MovieService,
+    private snackBar: MatSnackBar,
+    private userService: UserService,
+    private orgService: OrganizationService,
+  ) {
+  }
+
+  ngOnInit() {
+    this.route.params.subscribe(async params => {
```
GrandSchtroumpf, line 39: If we follow the akita pattern (service -> store -> query -> component) it would looks like that : 
```typescript
// Sync Firestore with the akita store
this.sub = this.route.params.pipe(
  pluck('eventId'),
  switchMap(id => this.eventService.syncScreening({ id })) // special sync that join event with org & movie
).subscribe();
// Get the event from the store
this.event$ = this.eventQuery.selectActive().pipe();
// Get the screening url from the callable function
this.privateLink$ = this.event$.pipe(
  distinctUntilChanges((a, b) => a.id === b.id), // trigger only when eventId changes
  switchMap(({ movie }) => this.movieService.getMoviePrivateConfig(movie.id, 'screeningUrl'))
);
```

libs/admin/src/lib/admin-panel/pages/event/event.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  public async setEventOwner() {
+    const user = await this.userService.getUser(this.event.ownerId);
+    if(!!user){
+      this.eventOwner= `User ${user.firstName} ${user.lastName} ( ${user.uid} )`;
+    } else { // should be an OrgId
+      const org = await this.orgService.getValue(this.event.ownerId);
+      this.eventOwner= `Organization ${org.denomination.full} ( ${org.id} )`;
+    }
+    this.cdRef.markForCheck();
```
GrandSchtroumpf, line 96: I really think you shouldn't mark for check, you're updating a string here. There is no mutation.

libs/admin/src/lib/admin-panel/pages/event/event.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  public getMovieAdminPath(movieId: string) {
+    return `/c/o/admin/panel/movie/${movieId}`;
+  }
+
+  public getCalendarPath() {
+    return `/c/o/dashboard/event`;
+  }
+
+  public async setEventOwner() {
```
GrandSchtroumpf, line 88: Each type of event has it's own type of owner (right now). So you could disciminate the ownerId type (user/org) with the type of event (screening => org / meeting => user).

libs/admin/src/lib/admin-panel/pages/events/events.component.html
```diff
@@ -0,0 +1,20 @@
+<h1><a routerLink="/c/o/admin/panel/">Administration panel</a> >Events</h1>
+
+<bf-table-filter 
+  [source]="rows" 
+  [columns]="versionColumns" 
+  [initialColumns]="initialColumns" 
+  [filterPredicate]="filterPredicate"
+  showFilter="true"
```
GrandSchtroumpf, line 8: I'm pretty sure the bf-table-filter coerc the boolean for `showFilter` & this other, you don't need to pass "true".
If not, this wouldn't work as you're passing the string "true", and not the boolean ;)

libs/admin/src/lib/admin-panel/pages/events/events.component.ts
```diff
@@ -0,0 +1,62 @@
...
+      row.edit = {
+        id: row.id,
+        link: `/c/o/admin/panel/event/${row.id}`,
+      }
+      return row;
+    });
+
+    this.rows = await Promise.all(promises);
+
+    this.cdRef.markForCheck();
```
GrandSchtroumpf, line 50: Same here as above. You're reassigning the value, ngZone should see it.

libs/admin/src/lib/admin-panel/pages/movie/movie.component.ts
```diff
@@ -53,6 +57,9 @@ export class MovieComponent implements OnInit {
     this.movie = await this.movieService.getValue(this.movieId);
     this.movieForm = new MovieAdminForm(this.movie);
 
+    const privateConfig = await this.movieService.getMoviePrivateConfig(this.movieId);
+    this.privateConfigForm = new PrivateConfigForm(privateConfig !== false ? privateConfig : {});
```
GrandSchtroumpf, line 61: It there any change the privateConfig is "0" or empty string ?
Else you can use "||" : `new PrivateConfigForm(privateConfig || {})`

libs/admin/src/lib/admin-panel/pages/overview/overview.component.html
```diff
@@ -9,12 +9,17 @@ <h2>Pages</h2>
   </li>
   <li>
     <a routerLink="../deals">
-      <mat-icon>theaters</mat-icon><span matLine>Distribution deals</span>
+      <mat-icon svgIcon="video_library"></mat-icon><span matLine>Distribution deals</span>
```
GrandSchtroumpf, line 12: 👍

libs/event/src/lib/+state/event.service.ts
```diff
@@ -33,4 +37,37 @@ export class EventService extends CollectionService<EventState> {
...
+  /*public getEventUrl(eventId: string): Promise<string> {
+    const f = this.functions.httpsCallable('getEventUrl');
+    return f({ eventId }).toPromise();
+  }*/
+
+  /**
+   * @dev ADMIN method
+   * Get all contracts.
+   */
+  public async getAllEvents(): Promise<Event[]> {
```
GrandSchtroumpf, line 66: Normally this is the same as `service.getValue()` without params.

libs/notification/src/lib/invitation/+state/invitation.firestore.ts
```diff
@@ -9,40 +9,61 @@ export interface Invitation {
   id: string;
   app: App;
   type: InvitationType;
+  mode: InvitationMode,
   status: InvitationStatus;
   date: Timestamp;
+  fromOrg?: PublicOrganization,
+  fromUser?: PublicUser,
```
GrandSchtroumpf, line 16: Maybe add some comments to explain what it is.

libs/utils/src/lib/common-interfaces/utility.ts
```diff
@@ -17,6 +17,24 @@ export interface BankAccount {
...
+  /** @dev private custom config for a doc. Setted by an https callable function */
+  config?: PrivateConfig
+}
+
+export interface PrivateConfig {
+  url?: string
+};
+
+// Private event config
+export function createPrivateConfig(config: Partial<PrivateConfig>): PrivateConfig {
```
GrandSchtroumpf, line 32: This function is doing nothing XD

apps/catalog/catalog/src/app/marketplace/layout/layout.component.ts
```diff
@@ -36,7 +36,7 @@ export class LayoutComponent implements OnInit, AfterViewInit, OnDestroy {
       map(wishlists => wishlists.find(wishlist => wishlist.status === 'pending'))
     );
     this.wishlistCount$ = this.currentWishlist$.pipe(
-      map(wishlist => wishlist.movieIds.length)
+      map(wishlist => wishlist ? wishlist.movieIds.length : 0)
```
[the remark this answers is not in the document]
phpgeekfr: The problem here is that wishlist is undefined, not wishlist.movieIds. Does your change handles this?

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -125,4 +126,22 @@ export class MovieService extends CollectionService<MovieState> {
...
+  public async setMoviePrivateConfig(movieId: string, privateConfig: PrivateConfig): Promise<any> {
+    const f = this.functions.httpsCallable('setDocumentPrivateConfig');
+    return f({ docId: movieId, config: privateConfig }).toPromise();
+  }
+
+  /**
+   * @dev ADMIN method
+   * Https callable function to get privateConfig for a movie.
+   */
+  public async getMoviePrivateConfig(movieId: string): Promise<PrivateConfig | false> {
```
GrandSchtroumpf, line 143: I think you should need to precise a key or an array of keys as second parameters in case there are different rules on the the privateConfig.

apps/backend-functions/src/privateConfig.ts
```diff
@@ -0,0 +1,96 @@
...
+
+
+/**
+ * Generic https callable function to set privateConfig for a document.
+ * Can only be called by blockframes admin
+ * @param data 
+ * @param context 
+ */
+export const setDocumentPrivateConfig = async (data: { docId: string, config: PrivateConfig }, context: CallableContext) => {
+  if (!context || !context.auth) { return false; }
```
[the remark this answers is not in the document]
phpgeekfr: It makes no differences. Updating to throw errors instead

apps/catalog/catalog/src/app/marketplace/layout/layout.component.ts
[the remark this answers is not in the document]
phpgeekfr: tested, this way `wishlist.movieIds.length || 0` does not work  if wishlist is undefined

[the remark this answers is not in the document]
GrandSchtroumpf: You can do : `wishlist?.movieIds?.length || 0 `

libs/utils/src/lib/common-interfaces/utility.ts
phpgeekfr: Pretty cool isn't it?

libs/movie/src/lib/movie/+state/movie.service.ts
phpgeekfr: good catch

libs/admin/src/lib/admin-panel/pages/events/events.component.html
phpgeekfr: In fact both are working (true or "true"). Changde to true to be more logical

apps/backend-functions/src/privateConfig.ts
```diff
@@ -0,0 +1,98 @@
...
+  const admin = await db.doc(`blockframesAdmin/${context.auth.uid}`).get();
```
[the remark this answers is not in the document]
phpgeekfr: Added a comment to  #2460 here since the setEventUrl is currently commented.

PR #2430, opened 2020-04-01
Add small updates after Fred review

GrandSchtroumpf:


apps/festival/festival/src/app/marketplace/marketplace.component.html
```diff
@@ -8,19 +8,19 @@
...
       </a>
       <a mat-list-item routerLink="wishlist" routerLinkActive="active">
         <mat-icon mat-list-icon  svgIcon="heart_fill"></mat-icon>
         <span mat-line>Wishlist</span>
       </a>
       <a mat-list-item routerLink="event" routerLinkActive="active">
-        <mat-icon mat-list-icon  svgIcon="video_library"></mat-icon>
-        <span mat-line>Sceenings</span>
+        <mat-icon mat-list-icon  svgIcon="screening"></mat-icon>
+        <span mat-line>My Sceenings</span>
```
Mussuranae, line 23: Typo : scReenings

libs/utils/src/lib/static-model/staticModels.ts
```diff
@@ -8,7 +8,7 @@ const models = {
     { 'slug': 'horror', 'label': 'Horror' },
     { 'slug': 'science-fiction', 'label': 'Science Fiction' },
     { 'slug': 'thriller', 'label': 'Thriller' },
-    { 'slug': 'coming-of-age', 'label': 'Coming of Age' },
+    { 'slug': 'coming-of-age', 'label': 'Young Adult' },
```
Mussuranae, line 11: Can you also rename slug ?

GrandSchtroumpf: It would change the database for a change of label. It's not worst it I think

pldespaigne: good catch !

PR #2463, opened 2020-04-07
[WIP] Add prerendering

chrisguttandin:
This pull request adds prerendering for the festival app. It's basically the result of running `ng generate universal --client-project festival` and `npm install angular-prerender --save-dev`.

It also adds a script new script which can be executed by executing `npm run prerender:festival`.


angular.json
```diff
@@ -878,6 +878,30 @@
...
+          "options": {
+            "outputPath": "dist/apps/festival/server",
+            "main": "apps/festival/festival/src/main.server.ts",
+            "tsConfig": "apps/festival/festival/tsconfig.server.json",
+            "externalDependencies": [
+              "@firebase/firestore"
+            ]
+          },
+          "configurations": {
+            "production": {
```
GrandSchtroumpf, line 893: Could you add the config the staging ("env/env.staging.ts") & ci ("env/env.ci.ts") ?

apps/festival/festival/src/app/app.server.module.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@NgModule({
+  imports: [
+    AppModule,
+    ServerModule,
+  ],
+  bootstrap: [AppComponent],
+  providers: [
+    {
+        provide: YandexMetricaService,
```
GrandSchtroumpf, line 16: As the YandexMetricaService is our own module I would recommend to add the exception for server side there. Like that it would work on every apps out-of-the-box. 
It's low priority though, I can take care of that :)

libs/ui/src/lib/icon-service.ts
```diff
@@ -260,8 +260,14 @@ export type IconSvg = typeof icons[number]['name'];
 @Injectable({ providedIn: 'root' })
 export class IconService {
   constructor(private matIconRegistry: MatIconRegistry, private domSanitizer: DomSanitizer) {
-    icons.forEach(({ name, url }) => {
-      this.matIconRegistry.addSvgIcon(name, this.domSanitizer.bypassSecurityTrustResourceUrl(url));
-    });
+    if (typeof window === 'undefined') {
```
GrandSchtroumpf, line 263: Could you reference the link to the issue here so we have a reference in the code of why we do that ?

libs/ui/src/lib/theme/theme.service.ts
```diff
@@ -35,7 +35,7 @@ export class ThemeService {
 
   /** Get the current value of the theme */
   initTheme(mode: 'dark' | 'light') {
-    const theme = localStorage.getItem('theme') || mode;
+    const theme = (typeof window !== 'undefined') ? localStorage.getItem('theme') || mode : mode;
```
GrandSchtroumpf, line 38: Can we use `plateformId` instead ? I think it's more explicit than `(typeof window !== 'undefined')` especially when working with junior developers who don't know what are the issues with angular Universal.

package.json
```diff
@@ -45,7 +45,8 @@
     "pwm": "pwmetrics --config",
     "bundlesize": "bundlesize",
     "postinstall": "ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points",
-    "storybook:ui": "npx ng run ui:storybook"
+    "storybook:ui": "npx ng run ui:storybook",
+    "prerender:festival": "npx ng build festival --prod && npx ng run festival:server && npx angular-prerender --browser-target festival:build --server-target festival:server"
```
GrandSchtroumpf, line 49: 👍

pldespaigne:
🎉

libs/ui/src/lib/theme/theme.service.ts
chrisguttandin: Yes, I agree `isPlatformBrowser(PLATFORM_ID)` is much more self explanatory.

libs/ui/src/lib/icon-service.ts
chrisguttandin: I added a comment.

apps/festival/festival/src/app/app.server.module.ts
chrisguttandin: Okay, I moved the branching inside the YandexMetricaService.

angular.json
chrisguttandin: I added the targets by copying all settings besides the file replacement. I hope that's fine.

PR #2477, opened 2020-04-08
refactor algolia & fix it's use every where

pldespaigne:
- [x] refactor Algolia autocomplete component
- [x] search an facet feature
- [x] fix use of the component everywhere
- [x] use it in seller-filter


libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.html
```diff
@@ -1,35 +1,66 @@
...
+      <ng-template [ngTemplateOutlet]="template || default" [ngTemplateOutletContext]="{ $implicit: result }"></ng-template>
       <ng-template #default> {{ result | json }}</ng-template>
+
     </mat-option>
   </mat-autocomplete>
 </ng-container>
+
+
+<!-- NATIVE VERSION -->
 <ng-template #nativeControl>
```
GrandSchtroumpf, line 37: It doesn't feels like it requires a full new template for just the property "matNativeControl". Maybe you can ask @Maxstgt why it was required.

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.ts
```diff
@@ -51,132 +62,127 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
...
+  /** Set your own placeholder */
   @Input() placeholder = 'Search...'
 
-  /**
-   * Can set to false if control should display the value
-   */
-  @Input() resetInput = false;
+  /** Can set to false if control should display the value */
+  private _resetInput: boolean;
+  @Input()
```
GrandSchtroumpf, line 76: @Maxstgt has created a `@boolean` decorator for that. It should make your life easier.

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.ts
```diff
@@ -51,132 +62,127 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
...
-
+  /** Sets the input to a native input control */
   private _native: boolean;
-
-  /**
-   * Sets the input to a native input control
-   */
   @Input()
-  get native() { return this._native }
+  get native() { return this._native; }
```
GrandSchtroumpf, line 88: same here.

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.ts
```diff
@@ -51,132 +62,127 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
...
   };
 
-  /**
-   * Holds information for showing which icons
-   */
+  /** Holds information for showing which icons */
   private _icons: Array<'cross' | 'magnifying_glasses'> = [];
-
   @Input()
   get icons() { return this._icons }
```
GrandSchtroumpf, line 96: I don't see the point of having a getter/setter format here. You can do : 
```typescript
@Input() icons: ('cross' | 'magnifying_glasses')[];
```
Maybe a prefix/suffix interface would make more sense btw.

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.ts
```diff
@@ -51,132 +62,127 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
...
+  /** Renders the template coming from the parent component */
+  @ContentChild(TemplateRef) template: TemplateRef<any>;
+
+  @ViewChild('input') input: ElementRef<HTMLInputElement>;
 
   ngOnInit() {
-    this.indexSearch = this.config.searchClient.initIndex(this.config.indexName)
+
+    if (!!this.facetName.trim()) {
+      this.pathToValue = 'value'; // for facet the result object only contain a 'value' key
```
GrandSchtroumpf, line 127: Maybe add a `@note` in the comment of the pathToValue to say that you shouldn't specify faceName & pathToValue.

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.ts
```diff
@@ -51,132 +62,127 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
...
-      switchMap(text => this.indexSearch.search(text)),
-      pluck('hits')
+      switchMap(text => {
+        if (!this.facetName.trim()) {
+          return this.indexSearch.search(text);
+        } else {
+          return this.indexSearch.searchForFacetValues({facetName: this.facetName, facetQuery: text});
+        }
+      }),
+      map((result: any) => {
```
GrandSchtroumpf, line 145: Might be interesting at some point to have a generic type for algolia-autocomplete component

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.ts
```diff
@@ -51,132 +62,127 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
...
+    this.indexSearch = searchClient.initIndex(this.indexName);
+
     this.algoliaSearchResults$ = this.control.valueChanges.pipe(
       debounceTime(300),
-      filter(text => typeof text === 'string'),
+      filter(text => typeof text === 'string' && !!text.trim()),
       distinctUntilChanged(),
-      switchMap(text => this.indexSearch.search(text)),
-      pluck('hits')
+      switchMap(text => {
```
GrandSchtroumpf, line 138: I think it's more readable like that: 
```typescript
const textSearch = (text: string) => this.indexSearch.search(text).pipe(pluck('hits'));
const facetSearch = (text: string) => this.indexSearch.searchForFacetValues({facetName: this.facetName, facetQuery: text}).pipe(pluck('facetHits'));
this.algoliaSearchResults$ = this.control.valueChanges.pipe(
  ...,
  switchMap(text => iif(this.facetName.trim(), facetSearch(text), textSearch(text))),
  tap(data => this.lastValue$.next(data))
);
```
[Documentation](https://www.learnrxjs.io/learn-rxjs/operators/conditional/iif)

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.ts
```diff
@@ -51,132 +62,127 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
...
     }
   }
 
   /**
-  * Since we input the path we need to initalize the function after the input gets handled,
+  * Since we input the path we need to initialize the function after the input gets handled,
   * otherwise displayWithPath is undefined and this will throw an error
   */
-  public displayFn() {
+  public displayFn = () => {
```
GrandSchtroumpf, line 185: I prefer binding the call than makeing this method an attribute.

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.html
Maxstgt: It was required because with form field you still get the underline that animates when clicked in. That's why it's a new template

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.ts
Maxstgt: Yes but not merged yet.

libs/ui/src/lib/algolia-autocomplete/algolia-autocomplete.component.ts
pldespaigne: I have tried `iff` but the problem is that it fire both search then evaluate the boolean then subsribe, but because both search are fired it always result in an error on one of them

pldespaigne:
- [x] refactor `seller-filter` into `algolia-chips-autocomplete`

apps/catalog/catalog/src/app/dashboard/layout/layout.component.html
```diff
@@ -34,9 +34,9 @@
       <button mat-icon-button (click)="sidenav.toggle()">
         <mat-icon svgIcon="menu"></mat-icon>
       </button>
-      <algolia-autocomplete [indexName]="movieIndex" native [icons]="['cross', 'magnifying_glasses']" [resetInput]="true">
+      <algolia-autocomplete icon native clearable resetInput [indexName]="movieIndex" keyToDisplay="title.international">
```
GrandSchtroumpf, line 37: I really don't like the "icon" input. It's not super readable. I would rather go for `prefix="iconName"`.

Maybe later we could have something like the chips & datepicker : 
```html
<algolia-autocomplete (clear)="">
  <mat-icon algoliaClear svgIcon=""></mat-icon>
</algolia-autocomplete>
```

apps/festival/festival/src/app/marketplace/title/list-new/list.component.html
```diff
@@ -114,7 +114,15 @@
       </ng-template>
 
       <ng-template filter label="Seller">
-        <title-filter-seller [form]="sellersForm"></title-filter-seller>
+        <algolia-chips-autocomplete
```
GrandSchtroumpf, line 117: I would recommand to change "indexName" to "index" & "facetName" to "facet".
Also "keyToDisplay" should be "displayWith". Something like that :
```typescript
@Input() displayWith(display: string | function) {
  delete this.keyToDisplay; 
  delete this.displayFn;
  typeof display === 'string'
    ? this.keyToDisplay = display
    :  this.displayFn = displayFn
}
```

libs/ui/src/lib/algolia/autocomplete/algolia-autocomplete.component.html
```diff
@@ -0,0 +1,64 @@
...
+      autocapitalize="off"
+      autocorrect="off"
+      spellcheck="false"
+    />
+    <mat-icon matSuffix svgIcon="cross" *ngIf="clearable && control.value" (click)="control.reset()"></mat-icon>
+    <mat-icon matPrefix [svgIcon]="svgIcon" *ngIf="icon" class="pointer"></mat-icon>
+  </mat-form-field>
+  <mat-autocomplete
+    #auto="matAutocomplete"
+    [displayWith]="displayFn.bind(this)"
```
GrandSchtroumpf, line 22: 👍

PR #2494, opened 2020-04-09
Issue#2392 titles flash

Hugo-Hontans:
- [x] Fix flash on titles page
- [x] Implement empty image in marketplace/deals

### DIFF
--- a/libs/contract/src/lib/contract/guards/organization-contract-list.guard.ts
+++ b/libs/contract/src/lib/contract/guards/organization-contract-list.guard.ts
@@ -20,7 +20,7 @@ const organizationContractsListQuery = (orgId: string): Query<ContractWithTimeSt
 
 /** Sync all contract of the active organization */
 @Injectable({ providedIn: 'root' })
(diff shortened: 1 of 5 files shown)


libs/contract/src/lib/contract/guards/organization-contract-list.guard.ts
```diff
@@ -20,7 +20,7 @@ const organizationContractsListQuery = (orgId: string): Query<ContractWithTimeSt
 
 /** Sync all contract of the active organization */
 @Injectable({ providedIn: 'root' })
-@CollectionGuardConfig({ awaitSync: true })
+@CollectionGuardConfig({ awaitSync: false })
```
ioaNikas, line 23: You can delete the line as default value of awaitSync is false.

Hugo-Hontans: I prefer to let false because the person who sees the code understands that it is done on purpose

PR #2505, opened 2020-04-10
Contract tunnel fix

### LINKED ISSUE #2400: [Release 1.3] Small Contract Tunnel fixes
- Contract - payment terms: change ‘every’ for ‘After` and add `Within` before duration and Period
- Exploitation rights: open the first accordion from start if possible
- Terms were not imported from the cart
- Dates going red without explanation:
image
- submit your contract offer > link to congrats page
- wrong status on submit 
image

### LINKED ISSUE #2404: Title Price is not saved in contract tunnel
Screenshot_2020-03-31 Create a contract offer - Overview - Archipel Content
Example route :
http://localhost:4200/c/o/marketplace/tunnel/contract/5ZRTalVn5W6dHrMa3QhE/sale/details

The forms get the value from DB but when saving, the price.amount is always zero.

Maxstgt:
close #2400 

we want to update the DB with the contrat form value and not with the store value, store value is old and not updated. This was the reason why we had the price issue that Bruce stated in ref #2404 
close #2404


libs/contract/src/lib/contract/tunnel/summary-sale/summary-sale.component.ts
```diff
@@ -69,8 +72,8 @@ export class SummarySaleComponent implements OnInit {
    * Submit a contract version to Archipel Content
    */
   async submit() {
-    const contract = this.query.getActive();
     await this.tunnel.save();
-    return this.contractService.submit(contract);
+    await this.contractService.submit(this.tunnel.contractForm.value);
+    this.router.navigate(['success'], {relativeTo: this.route})
```
GrandSchtroumpf, line 77: You might want to try/catch before navigating no ?

PR #2507, opened 2020-04-10
Pre updates on invitations

GrandSchtroumpf:
Some update on invitation before a larger PR

### DIFF
--- a/apps/festival/festival/src/app/dashboard/event/view/view.component.ts
+++ b/apps/festival/festival/src/app/dashboard/event/view/view.component.ts
@@ -1,4 +1,4 @@
-import { Component, ChangeDetectionStrategy } from '@angular/core';
+import { Component, ChangeDetectionStrategy, OnInit } from '@angular/core';
 import { EventQuery } from '@blockframes/event/+state/event.query';
 
 @Component({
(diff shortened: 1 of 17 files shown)


apps/festival/festival/src/app/dashboard/event/view/view.component.ts
```diff
@@ -7,10 +7,15 @@ import { EventQuery } from '@blockframes/event/+state/event.query';
...
 })
-export class EventViewComponent {
+export class EventViewComponent implements OnInit {
 
   event$ = this.query.selectActive();
 
+
   constructor(private query: EventQuery) { }
 
+  ngOnInit() {
```
pldespaigne, line 17: Why do you import and implement `OnInit` for an empty `ngOnInit()` ??

libs/event/src/lib/+state/event.service.ts
```diff
@@ -44,10 +44,10 @@ export class EventService extends CollectionService<EventState> {
    * @param eventId 
    */
   // @TODO (#2460)  Waiting for a decision on screening flow before uncomment
```
pldespaigne, line 46: Maybe know you can remove the TODO comment

PR #2519, opened 2020-04-14
Title css

pldespaigne:
css & responsiveness of title page and components

### DIFF
--- a/apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
+++ b/apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
@@ -1,0 +1,9 @@
+
+mat-form-field {
+  width: 100%;
+}
+
+movie-card {
+  max-width: 250px;
+  padding: 0 0 24px 0;
+}
--- a/apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
+++ b/apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
@@ -1,9 +1,27 @@
+@import '~@angular/flex-layout/mq';
 
 mat-form-field {
   width: 100%;
 }
 
+title-filter {
+  static-check-boxes {
+    padding: 0 16px;
+  }
+}
+
 movie-card {
   max-width: 250px;
   padding: 0 0 24px 0;
+}
+
+@include layout-bp(xs) {}
+
+@include layout-bp(sm) {}
+
+@include layout-bp(md) {
+  movie-card {
+    max-width: 200px;
+    padding: 0 0 16px 0;
+  }
 }
--- a/apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
+++ b/apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
@@ -11,17 +11,25 @@ title-filter {
 }
 
 movie-card {
-  max-width: 250px;
+  max-width: 25%;
   padding: 0 0 24px 0;
 }
 
-@include layout-bp(xs) {}
+@include layout-bp(xs) {
+  movie-card {
+    max-width: 65%;
+  }
+}
 
-@include layout-bp(sm) {}
+@include layout-bp(sm) {
+  movie-card {
+    max-width: 45%;
+  }
+}
 
 @include layout-bp(md) {
   movie-card {
-    max-width: 200px;
+    max-width: 35%;
     padding: 0 0 16px 0;
   }
 }
--- a/apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
+++ b/apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
@@ -12,7 +12,7 @@ title-filter {
 
 movie-card {
   max-width: 25%;
-  padding: 0 0 24px 0;
+  padding-bottom: 24px;
 }
 
 @include layout-bp(xs) {
@@ -30,6 +30,6 @@ movie-card {
 @include layout-bp(md) {
   movie-card {
     max-width: 35%;
-    padding: 0 0 16px 0;
+    padding-bottom: 16px;
   }
 }
--- a/libs/movie/src/lib/movie/form/filters/languages/language-filter.component.html
+++ b/libs/movie/src/lib/movie/form/filters/languages/language-filter.component.html
@@ -1,18 +1,16 @@
 
 <!-- LANGUAGE AUTO COMPLETE -->
-<div fxLayout="row" fxLayoutAlign="start start" fxLayoutGap="16px">
+<div fxLayout="row" fxLayoutAlign="start start" fxLayoutGap="16px" class="select-language">
 
   <input-autocomplete
-    fxFlex="grow"
+    fxFlex
     label="New language"
     scope="LANGUAGES"
     [control]="languageControl"
   ></input-autocomplete>
 
   <button
-    fxFlex="auto"
     test-id="languages"
-    class="language-button"
     mat-mini-fab
     [disabled]="!languageControl.valid"
     color="primary"
--- a/libs/movie/src/lib/movie/form/filters/languages/language-filter.component.html
+++ b/libs/movie/src/lib/movie/form/filters/languages/language-filter.component.html
@@ -1,14 +1,16 @@
 
 <!-- LANGUAGE AUTO COMPLETE -->
-<div fxLayout="row" fxLayoutAlign="start start">
+<div fxLayout="row" fxLayoutAlign="start start" fxLayoutGap="16px">
 
   <input-autocomplete
+    fxFlex="grow"
     label="New language"
     scope="LANGUAGES"
     [control]="languageControl"
   ></input-autocomplete>
 
   <button
+    fxFlex="auto"
     test-id="languages"
     class="language-button"
     mat-mini-fab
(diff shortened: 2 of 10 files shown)


apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
```diff
@@ -0,0 +1,27 @@
...
+}
+
+title-filter {
+  static-check-boxes {
+    padding: 0 16px;
+  }
+}
+
+movie-card {
+  max-width: 250px;
```
GrandSchtroumpf, line 14: Better set a % and change it with the responsive

libs/movie/src/lib/movie/form/filters/languages/language-filter.component.html
```diff
@@ -1,14 +1,16 @@
...
 
   <input-autocomplete
+    fxFlex="grow"
     label="New language"
     scope="LANGUAGES"
     [control]="languageControl"
   ></input-autocomplete>
 
   <button
+    fxFlex="auto"
```
GrandSchtroumpf, line 13: I think you can do : 
```html
<input-autocomplete fxFlex label="New language" label="New language" [control]="languageControl"></input-autocomplete>
<button test-id="languages" mat-mini-fab>..</button>
```

libs/movie/src/lib/movie/form/filters/languages/language-filter.component.html
```diff
@@ -1,14 +1,16 @@
...
     test-id="languages"
     class="language-button"
```
GrandSchtroumpf, line 15: Are you using this class in the css ?

libs/movie/src/lib/movie/form/filters/languages/language-filter.component.scss
```diff
@@ -0,0 +1,6 @@
+
+div {
```
GrandSchtroumpf, line 2: Better use class than targeting the children of a div I think. I would say that for article/section, but it's weird for me with div

libs/movie/src/lib/movie/layout/list/title-list.component.html
```diff
@@ -3,24 +3,28 @@
...
-          <mat-icon matPrefix svgIcon="list"></mat-icon>
-        </ng-container>
-        <ng-container *ngIf="listView">
-          <mat-icon matPrefix svgIcon="grid"></mat-icon>
-        </ng-container>
-      </button>
-      <ng-template [ngTemplateOutlet]="titleSortTemplate"></ng-template>
+    <nav fxLayout="row" fxLayoutGap.gt-sm="10px">
+      <h3 fxFlex>{{titleType}} List</h3>
+      <span fxFlex="1 1 100%">
```
GrandSchtroumpf, line 8: span is for an inline text. Use div for layouting.
Same below

libs/movie/src/lib/movie/layout/list/title-list.component.html
```diff
@@ -3,24 +3,28 @@
...
+        <ng-template [ngTemplateOutlet]="titleSearchTemplate"></ng-template>
+      </span>
+      <span fxFlex fxLayout="row" fxLayoutAlign="start center">
```
GrandSchtroumpf, line 11: Why do you need to use 3 fxFlex? I never had to do that before, so I fill like there is something weird going on here ^^

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,47 +1,81 @@
 @import '~@angular/flex-layout/mq';
 
+page-nav {
+  nav {
+    border: none;
```
GrandSchtroumpf, line 5: Does the nav have a border by default ?

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,47 +1,81 @@
...
+    width: 90vw;
```
GrandSchtroumpf, line 6: Why ? I don't think you should do that

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,47 +1,81 @@
...
+    align-items: center;
+    h3 {
+      text-transform: capitalize;
+    }
+    span {
+      padding: 20px 0 0 0;
```
GrandSchtroumpf, line 12: do `padding-top: 24px` instead. It's more readable.

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,47 +1,81 @@
...
-      width: 100%;
-      position: static;
-      mat-card {
-        height: 100%;
-      }
-      .filter {
-        margin-top: 50px;
+  page-nav {
+    nav {
+      width: 80vw;
```
GrandSchtroumpf, line 37: same as above

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,47 +1,81 @@
...
+      h3 {
+        display: none;
```
GrandSchtroumpf, line 39: Better having *ngIf or something like that if possible with fxLayout

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,47 +1,81 @@
...
+  aside {
+    position: static;
+    mat-card {
+      width: 100%;
+      max-height: fit-content;
     }
   }
 }
 
 @include layout-bp(sm) {
```
GrandSchtroumpf, line 52: It's the same as xs, use `@include layout-bp(lt-md)` or something like that

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,47 +1,81 @@
...
       }
     }
-    article mat-form-field-wrapper {
-      width: 80%;
+  }
+  aside {
+    position: static;
+    mat-card {
+      width: 100%;
+      max-height: fit-content;
```
GrandSchtroumpf, line 47: Why you need to do that here and do some calculation above ?

libs/movie/src/lib/movie/layout/list/title-list.component.html
pldespaigne: this is like inline layout in the nav and somehow using div instead of span trash the vertical alignment

libs/movie/src/lib/movie/layout/list/title-list.component.scss
pldespaigne: I think so, this is the only way I found to remove an ugly bottom border floating above the page

libs/movie/src/lib/movie/layout/list/title-list.component.scss
pldespaigne: I can use % if you prefer, but this is to prevent the right edge of the sort select to go out of the screen in the nav

libs/movie/src/lib/movie/layout/list/title-list.component.scss
pldespaigne: because when the screen is big the sidenav is sticky so we want to be able to scroll inside, and when the screen is small the sidenav is static and above the cards and we just want to scroll the whole page instead of scrolling inside the sidenav

apps/festival/festival/src/app/marketplace/title/list-new/list.component.scss
```diff
@@ -0,0 +1,35 @@
...
+
+title-filter {
+  static-check-boxes {
+    padding: 0 16px;
+  }
+}
+
+movie-card {
+  max-width: 25%;
+  padding: 0 0 24px 0;
```
GrandSchtroumpf, line 15: Use `padding-bottom` to make it easier to read.

libs/movie/src/lib/movie/form/filters/languages/language-filter.component.scss
```diff
@@ -0,0 +1,6 @@
+
+.select-language {
+  button {
+    margin: 16px 0 0 0;
```
GrandSchtroumpf, line 4: Use `margin-top` to make it easier to read.

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,47 +1,47 @@
 @import '~@angular/flex-layout/mq';
 
+page-nav {
+  mat-toolbar {
+    div {
+      padding: 24px 0 0 0;
```
GrandSchtroumpf, line 6: I dont' understand why you need that.
+ Use `padding-top` to make it easier to read.

libs/movie/src/lib/movie/layout/list/title-list.component.html
```diff
@@ -1,11 +1,9 @@
 <page-nav targetId="nav">
-  <mat-toolbar>
-    <button mat-icon-button>
-      <mat-icon svgIcon="menu"></mat-icon>
-    </button>
-    <nav mat-tab-nav-bar>
-      <h3>{{titleType}} List</h3>
+  <mat-toolbar fxLayout="row" fxLayoutGap.gt-sm="10px">
```
GrandSchtroumpf, line 2: Do : `<mat-toolbar fxLayout fxLayoutAlign="start center" fxLayoutGap="16px">`

PR #2534, opened 2020-04-15
Create Layout for marketplace/dashboard

GrandSchtroumpf:
Create a layout for marketplace & dashboard
Implement it inside the festival app.

### DIFF
--- a/apps/festival/festival/src/app/dashboard/dashboard.component.html
+++ b/apps/festival/festival/src/app/dashboard/dashboard.component.html
@@ -1,75 +1,31 @@
-<mat-sidenav-container>
-  <mat-sidenav #sidenav id="dashboard-navbar" [attr.role]="(ltMd$ | async) ? 'dialog' : 'navigation'"
-    [mode]="(ltMd$ | async) ? 'over' : 'side'" [opened]="(ltMd$ | async) === false">
-    <aside fxLayout="column"> <!-- We need another layer because mat-sidenav is not a direct parent -->
-      <mat-toolbar>
-        <a class="logo" routerLink=".">
-          <img src="assets/logo/dark/festival.svg" alt="Archipel Content Logo" />
-        </a>
-      </mat-toolbar>
-      <mat-nav-list fxFlex>
-        <a mat-list-item routerLink="home" routerLinkActive="active">
-          <mat-icon matListIcon svgIcon="dashboard"></mat-icon>
-          <span matLine><b>Dashboard</b></span>
-        </a>
-        <a mat-list-item routerLink="title" routerLinkActive="active">
-          <mat-icon matListIcon svgIcon="video_library"></mat-icon>
-          <span matLine><b>My Titles</b></span>
-        </a>
-        <a mat-list-item routerLink="event" routerLinkActive="active">
-          <mat-icon matListIcon svgIcon="calendar"></mat-icon>
-          <span matLine><b>My Schedule</b></span>
-        </a>
-      </mat-nav-list>
-      <mat-nav-list *ngIf="org$ | async as org">
-        <mat-divider></mat-divider>
-        <a mat-list-item [routerLink]="['/c/o/organization', org.id]" routerLinkActive="active">
-          <img matListAvatar [imgRef]="org.logo" alt="Organization logo"/>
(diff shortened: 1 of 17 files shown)


apps/festival/festival/src/app/dashboard/dashboard.component.html
```diff
@@ -1,75 +1,31 @@
...
+  </mat-nav-list>
+  <mat-nav-list *ngIf="org$ | async as org">
+    <mat-divider></mat-divider>
+    <a mat-list-item [routerLink]="['/c/o/organization', org.id]" routerLinkActive="active">
+      <img matListAvatar [imgRef]="org.logo" alt="Organization logo"/>
+      <span matLine>{{ org.denomination.full }}</span>
+      <span matLine>
+        <small class="mat-caption">{{ org.activity }}</small>
+      </span>
+    </a>
```
Maxstgt, line 29: wouldn't it be reusable and cleaner if you got something like `tunnel-nav` component? you write your steps like so you did with an interface:
```
const steps = [{
  title: 'Step 1',
  icon: 'document',
  routes: [{
    path: 'details',
    label: 'Contract Details'
  }]
}, {
  title: 'Summary',
  icon: 'send',
  routes: [{
    path: 'summary',
    label: 'Summary'
  }]
}]
```
and you put it in. That way you got a clean html in your dashboard.component and it makes sure that we got the same layout in all of our apps

apps/festival/festival/src/app/marketplace/marketplace.module.ts
```diff
@@ -113,24 +115,26 @@ const routes: Routes = [{
...
+    // MatSidenavModule,
     MatIconModule,
-    MatButtonModule,
-    MatDividerModule,
-    MatBadgeModule,
-    MatMenuModule,
+    // MatButtonModule,
+    // MatDividerModule,
+    // MatBadgeModule,
+    // MatMenuModule,
```
Maxstgt, line 128: remove it or add TODO

libs/ui/src/lib/layout/dashboard/dashboard.component.ts
```diff
@@ -0,0 +1,54 @@
...
+    private invitationQuery: InvitationQuery,
+    private notificationQuery: NotificationQuery,
+    private router: Router
+  ) { }
+
+  ngAfterViewInit() {
+    // TODO #2502: https://github.com/angular/components/issues/4280
+    this.sub = this.router.events.pipe(
+      filter(event => event instanceof NavigationEnd)
+    ).subscribe(() => this.sidenavContent.scrollTo({ top: 0}))
```
Maxstgt, line 48: you can add a event on Navigation start and save the scrollY position so if you navigating back you can restore the position

apps/festival/festival/src/app/dashboard/dashboard.component.html
Maxstgt: it is more the configurable approach but since it is not that generic the steps, it seems the better way. For me content projection is for highly unpredictable data you want to display. Which is not the case here, as far as I can predict.

GrandSchtroumpf: I prefere using ng-content when possible. I think html is more readable than JSON.
In the case of tunnel we whre using the expension panel which are quite heavy to implement. Here it's just a list. But I agree I could have gone with a configurable JSON too

libs/ui/src/lib/layout/dashboard/dashboard.component.ts
GrandSchtroumpf: Do you have code example ? Could you open an issue for this ?

Maxstgt:
```typescript 
private prevYPosition = 0;

this.sub = this.router.events.pipe(
      tap(event => event instanceOf NavigationStart ? this.prevYPostion = window.scrollY),
      filter(event => event instanceof NavigationEnd)
    ).subscribe(() => this.sidenavContent.scrollTo({ top: prevYPosition || 0}))
```

Maxstgt: Since layout is always the top component the "local store" won't get eraised

PR #2536, opened 2020-04-15
Feat 2521 - import cleaning and organization

phpgeekfr:


libs/movie/src/lib/movie/components/import/table-extracted-elements/table-extracted-elements.module.ts
```diff
@@ -0,0 +1,72 @@
...
+    MatDialogModule,
+    MatCheckboxModule,
+    MatExpansionModule,
+
+    // Librairies
+    ImageReferenceModule,
+    TermDateModule,
+    ToLabelModule,
+  ],
+  entryComponents: [
```
Mussuranae, line 63: If you let it empty, is it useful to have it here ?

PR #2541, opened 2020-04-15
2521 - bis reserved territories for festival app

phpgeekfr:


libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -133,7 +133,7 @@ export class MovieService extends CollectionService<MovieState> {
    */
   public async setMoviePrivateConfig(movieId: string, privateConfig: PrivateConfig): Promise<any> {
     const f = this.functions.httpsCallable('setDocumentPrivateConfig');
-    return f({ docId: movieId, config: privateConfig }).toPromise();
+    return f({ docId: movieId, config: privateConfig }).toPromise().catch(_ => false);
```
GrandSchtroumpf, line 136: Do you have specific actions on "false" ?
I think you sould let the component decide what happen if the method failed.

libs/movie/src/lib/movie/components/import/table-extracted-elements/movies/movies.component.ts
```diff
@@ -125,6 +129,14 @@ export class TableExtractedMoviesComponent implements OnInit {
       hint: 'Movie already saved'
     });
     this.rows.data = data;
+
+    if (importState.distributionRights) {
+      const orgId = this.orgQuery.getActiveId();
+      const titlesAndRights = {} as TitlesAndRights;
+      titlesAndRights[importState.movie.id] = importState.distributionRights;
```
GrandSchtroumpf, line 136: The would be more readable:
```typescript
const movieId = importState.movie.id;
const titlesAndRights = { [movieId]: importState.distributionRights };
```

libs/movie/src/lib/movie/components/import/view-extracted-elements/movies/view-extracted-movies.component.ts
```diff
@@ -859,6 +865,34 @@ export class ViewExtractedMoviesComponent implements OnInit {
...
+        //////////////////
+        // FESTIVAL FIELDS
+        //////////////////
+        if (spreadSheetRow[SpreadSheetMovie.reservedTerritories]) {
+          // Here we need to create 'lite' version of distribution deals with only the reserved territories.
+          // This feature is used on festival app.
+
+          const distributionRight = createDistributionRight();
+          distributionRight.territory = [];
+          spreadSheetRow[SpreadSheetMovie.reservedTerritories].split(this.separator).forEach((c: ExtractCode<'TERRITORIES'>) => {
```
GrandSchtroumpf, line 877: Usually I use : 
- for ... of ... : iterate over array, I don't need the index
- forEach: iterate over array, I need the index
- map: I want to create a new array out of another.

PR #2543, opened 2020-04-16
movie-header

### LINKED ISSUE #2352: [Component] Header
# Component

[Light](https://zpl.io/2Gl5pYr)
[Dark](https://zpl.io/Vk09rrG)

## UI review
_remove this section once everything is checked_
- [x] UI doesn't already have a existing component.
- [ ] Name display in Zeplin is the same as the one in the issue (see below).
- [x] Data can be display like in the UI (see notion).
- [x] A link to this issue in the zeplin object

## Implement Component
- [x] Create story with storybook
- [x] Connect component to Zeplin with zeplin-connect

## Description

- name: catalog-footer
- lib/app: catalog - app

## Type
- [ ] **Layout**: _Structure a page with content projection (ex: MovieView)._
- [ ] **Page**: _Loaded by the router. Usually used in an app (ex: CatalogDashboardMovieView)._
- [x] **Component**: _Structure a piece of data in the page (ex: MovieTable)._
- [ ] **Display**: _Render data in a page. Can use a component-type (ex: MovieMain, MovieBudget)._
- [ ] **Form**: _Get data from the user (ex: MovieFormBudget)._

### Input
_can be data (movie, distribution right, ...), or UI related (dense, appearance)_
movie

### Data view
_data displayed in the UI, derived from the inputs (ex: **title**: movie.main.title.international)

------

## Where is it used
-links and snapshots where the component is used_
https://app.zeplin.io/project/5e82f6ca1c987015ec21c1d7/screen/5e830f0a69814511d8862b8b

Maxstgt:
close #2352 
the name in zeplin is a bit different from our because Mathilde cant rename it otherwise she will get conflicts with her other components in zeplin.
added poster mixin
added movie title features component


libs/movie/src/lib/movie/components/header/header.component.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+function tailorMovieForHeader(movie: Movie): MovieHeaderData {
+  const statusLabel = getLabelBySlug('MOVIE_STATUS', movie.main.status)
+  const convertedGenres: string[] = movie.main.genres.map(genre => getLabelBySlug('GENRES', genre))
+  const convertedOriginalLanguages = movie.main.originalLanguages.map(language => getLabelBySlug('LANGUAGES', language))
+  const convertedOriginCountries = movie.main.originCountries.map(country => getISO3166TerritoryFromSlug(country));
+  // Convert the values here for the tool tip, it saves a lot of resources
+  // We convert the array to a string and replace the commas with whitespace
+  const tooltipGenres = convertedGenres.splice(1, convertedGenres.length - 1).toString().replace(/[, ]/g, ' ');
```
GrandSchtroumpf, line 45: Put everything, don't leave the first element out.
Why not just doing `genres.join(', ')`?

libs/movie/src/lib/movie/components/header/header.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  tooltipOriginalLanguages: string,
+  productionYear: number,
+  productionStatus: MovieStatusLabel,
+  banner: PromotionalElement,
+  poster: PromotionalElement[],
+}
+
+function tailorMovieForHeader(movie: Movie): MovieHeaderData {
+  const statusLabel = getLabelBySlug('MOVIE_STATUS', movie.main.status)
+  const convertedGenres: string[] = movie.main.genres.map(genre => getLabelBySlug('GENRES', genre))
```
GrandSchtroumpf, line 40: All the code related to the feature line should be done in a specific component.

libs/ui/src/lib/storybook/toolkit/toolkit.component.ts
```diff
@@ -16,8 +16,9 @@ export class ToolkitComponent {
   view: Views = 'component';
   theme$ = this.themeService.theme$;
 
-  constructor(icons: IconService, private themeService: ThemeService, private cdr: ChangeDetectorRef) {
+  constructor(public icons: IconService, private themeService: ThemeService, private cdr: ChangeDetectorRef) {
     this.themeService.theme = 'light';
+    this.icons.init()
```
GrandSchtroumpf, line 21: remove public and do `icons.init()` without the this

libs/movie/src/lib/movie/components/header/header.component.html
```diff
@@ -0,0 +1,48 @@
...
+                        {{ tailoredMovie.originalLanguages[0] + ' |' }}
+                    </span>
+                    <span [matTooltip]="tailoredMovie.tooltipCountries" *ngIf="tailoredMovie.countries.length">
+                        <!-- Display only the first value, rest goes into the tooltip -->
+                        {{ tailoredMovie.countries[0].iso_a2 + ' |' }}
+                    </span>
+                    <span *ngIf="tailoredMovie.productionYear">
+                        {{ tailoredMovie.productionYear + ' |' }}
+                    </span>
+                    <span *ngIf="tailoredMovie.productionStatus">
```
GrandSchtroumpf, line 36: If there is no productionStatus then the line ends up with "|". It should be ok though.

libs/movie/src/lib/movie/components/header/header.component.html
```diff
@@ -0,0 +1,48 @@
...
+            </div>
+            <!-- Descriptions -->
+            <div fxLayout="column">
+                <h4>{{ tailoredMovie.directors | displayName }}</h4>
+                <h1 class="mat-display-2">{{ tailoredMovie.title.international }}</h1>
+                <h2>{{ tailoredMovie.title.original }}</h2>
+                <h4 class="mat-body">
+                    <span *ngIf="tailoredMovie.workType">
+                        {{ tailoredMovie.workType + ' |'}}
+                    </span>
```
GrandSchtroumpf, line 16: It's cleaner like that
`<span *ngIf="tailoredMovie.workType">{{ tailoredMovie.workType }} | </span>`

libs/movie/src/lib/movie/components/header/header.component.scss
```diff
@@ -0,0 +1,24 @@
+.poster {
+    width: 33%;
+    background-size: cover;
+}
+
+.inner-header {
```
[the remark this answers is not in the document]
GrandSchtroumpf: Applay the same padding on the right then

libs/movie/src/lib/movie/components/header/header.component.ts
```diff
@@ -0,0 +1,89 @@
...
+  templateUrl: './header.component.html',
+  styleUrls: ['./header.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class HeaderComponent {
+
+  public movieView: MovieHeaderView;
+
+  @HostBinding('style.backgroundImage') background: string;
+  @HostBinding('style.backgroundSize') size: string;
```
GrandSchtroumpf, line 59: Why not set the size on css ?
```css
:host {
  bakground-size: cover;
}
```

libs/movie/src/lib/movie/components/header/header.component.ts
```diff
@@ -0,0 +1,89 @@
...
+  @Input()
+  set movie(movie: Movie) {
+    if (movie) {
+      this.movieView = createMovieView(movie);
+      this.background = `url(${this.movieView.banner.media.url})`;
```
GrandSchtroumpf, line 64: What happen is there is no banner ?
I didn't try, but if you do `@HostBinding('backgroundRef') background: ImgRef` I think it'll work.

libs/movie/src/lib/movie/components/header/header.component.html
```diff
@@ -0,0 +1,22 @@
+<header fxLayout="column" fxLayoutAlign="start start" *ngIf="movieView">
```
GrandSchtroumpf, line 1: You don't need this now that you apply the background on the :host.
You can merge the three tags (in the code below I don't take the breadcrumbs into account yet): 
```html
<header class="overlay" fxLayout="row" fxLayoutGap="20px" *ngIf="movieView">
  <img [imgRef]="poster" placeholderAsset="empty_poster.webp" />
  <section fxLayout="column">
     ...
  </section>
</header>
```

libs/movie/src/lib/movie/components/header/header.component.html
```diff
@@ -0,0 +1,22 @@
...
+    <div class="overlay">
+        <!-- Breadcrumbs and Poster -->
+        <div fxLayout="row" fxLayoutGap="20px" class="inner-header">
+            <!-- ATM we use the first image, but we will use geo localisation to determine which one to use -->
+            <div [backgroundRef]="movieView.poster[0].media" placeholderAsset="empty_poster.webp" class="poster">
```
GrandSchtroumpf, line 6: Use img tag here.

libs/movie/src/lib/movie/components/header/header.component.ts
```diff
@@ -0,0 +1,89 @@
...
+  const statusLabel = getLabelBySlug('MOVIE_STATUS', movie.main.status)
+  return {
+    directors: movie.main.directors,
+    title: {
+      original: movie.main.title.original,
+      international: movie.main.title.international
+    },
+    banner: movie.promotionalElements.banner,
+    poster: movie.promotionalElements.poster,
+    titleFeatures: [
```
GrandSchtroumpf, line 37: I think this should be managed by the titleFeature. Just pass it the whole movie object as input. Else it means that we need to do this operation each time we need a feature line.

libs/movie/src/lib/movie/components/header/header.component.ts
```diff
@@ -0,0 +1,89 @@
...
+  const convertedOriginCountries = movie.main.originCountries.map(country => getISO3166TerritoryFromSlug(country)).map(country => country.iso_a2);
+  const statusLabel = getLabelBySlug('MOVIE_STATUS', movie.main.status)
+  return {
+    directors: movie.main.directors,
+    title: {
+      original: movie.main.title.original,
+      international: movie.main.title.international
+    },
+    banner: movie.promotionalElements.banner,
+    poster: movie.promotionalElements.poster,
```
GrandSchtroumpf, line 36: Get the first poster here directly.

libs/movie/src/lib/movie/components/title-features/title-features.component.html
```diff
@@ -0,0 +1,7 @@
+<h4 class="mat-body" *ngIf="features.length" fxLayout="row wrap">
+    <span *ngFor="let feature of features; let isLast = last" 
+        [matTooltip]="feature.tooltip">
+        <!-- Use name code here, otherwise a evil formatter can destroy the layout -->
+        {{ feature.view }}{{ isLast ? '' : '&nbsp;|&nbsp;'}}
```
GrandSchtroumpf, line 5: What about using a mat-divider vertical instead of "|" ? I think it would be nicer. Also you can specify the margin between each elements.

libs/movie/src/lib/movie/components/title-features/title-features.component.ts
```diff
@@ -0,0 +1,28 @@
...
+            return { view: feature[0], tooltip: feature.join(', ') }
+        }
+        return { view: feature, tooltip: '' }
+    })
+}
+
+@Component({
+    selector: '[titleFeatures] movie-title-features',
+    templateUrl: './title-features.component.html',
+    styleUrls: ['./title-features.component.scss']
```
GrandSchtroumpf, line 21: User changeDetection

libs/movie/src/lib/movie/components/title-features/title-features.component.ts
```diff
@@ -0,0 +1,28 @@
+import { Component, Input } from '@angular/core';
+
+interface TitleFeature {
+    view: string,
+    tooltip: string
+}
+
+function createTitleFeatureView(features: Array<string | string[]>): TitleFeature[] {
```
GrandSchtroumpf, line 8: use (string | string[])[]

libs/movie/src/lib/movie/components/title-features/title-features.component.html
```diff
@@ -0,0 +1,7 @@
+<h4 class="mat-body" *ngIf="features.length" fxLayout="row wrap">
+    <span *ngFor="let feature of features; let isLast = last" 
+        [matTooltip]="feature.tooltip">
```
GrandSchtroumpf, line 3: Does it show something if the content of "tooltip" is empty ? If yes, it shouldn't

styles/images.scss
```diff
@@ -0,0 +1,43 @@
...
+  md: (
+    begin: 960px,
+    end: 1279.9px
+  ),
+  lg: (
+    begin: 1280px,
+    end: 1919.9px
+  )
+*/
+@mixin poster($breakpoint) {
```
GrandSchtroumpf, line 19: Hmmm I think it makes things more difficult to understand if you use breakpoints to describe different size of poster that have nothing to do with the breakpoint. For example on a mobile you might wand to display the poster in large because you show it on all the screen. I would recommand to use different naming.

styles/images.scss
```diff
@@ -0,0 +1,43 @@
...
+        height: 1280px;
+    }
+}
+
+@mixin banner {
+    height: 1080px;
+    width: 1920px;
+}
+
+@mixin avatar($width: "auto", $height: $width) {
```
GrandSchtroumpf, line 40: 👍

styles/images.scss
Maxstgt: the breakpoints property are the same then from `@import '~@angular/flex-layout/mq';` you just pass down the param you got and you get the corresponding size. Can't get easier then that.
```
@include layout-bp(xs) {
    .poster {
        @include poster(xs);
    }
```

GrandSchtroumpf: Let's say I want my movie-card with poster to take the whole poster on a mobile I must do : 
```scss
@include layout-bp(xs) {
  .poster {
    @include poster(lg);
  }
}
```
Which makes things weird for me.

libs/movie/src/lib/movie/components/header/header.component.ts
Maxstgt: nope, it doesnt

GrandSchtroumpf: ok too bad ^^. Let's keep it with backgroundImage and manage the case when there is no banner.

libs/movie/src/lib/movie/components/header/header.component.html
Maxstgt: img gets a white broder.

libs/movie/src/lib/movie/components/header/header.component.ts
Maxstgt: missing module, event thou its imported

libs/movie/src/lib/movie/components/header/header.component.html
GrandSchtroumpf: o.O, where does it come from ? I don't see any reason to have a white border on every image of the app. This might be a old code. Can you update it please ?
We need an img tag here, not a background.

Maxstgt: Will have a look!

.vscode/launch.json
```diff
@@ -10,6 +10,21 @@
...
+            "program": "${workspaceFolder}/node_modules/.bin/ng",
+            "args": [
+                "test",
+                "movie",
+                "--runInBand",
+                "true"
+            ],
+            "cwd": "${workspaceFolder}",
+            "console": "integratedTerminal",
+            "internalConsoleOptions": "neverOpen"
```
Maxstgt, line 27: debugger for movie lib

angular.json
```diff
@@ -523,7 +523,7 @@
           "style": "scss"
         }
       },
-      "root": "apps/catalog/catalog",
+      "root": "",
```
Maxstgt, line 526: debugger for chrome

libs/ui/src/lib/slider/slider.stories.ts
```diff
@@ -4,12 +4,12 @@ import { ToolkitModule } from '@blockframes/ui/storybook';
...
 
 export default {
-  title: 'Bf Slider'
+  title: 'Slider'
 };
 
 export const bfSlider = () => ({
   moduleMetadata: { imports: [SliderModule, ToolkitModule] },
-  name: 'Bf Slider',
+  name: 'Slider',
```
Maxstgt, line 12: otherwise zeplin CLI will throw an error

libs/movie/src/lib/movie/components/header/header.component.html
```diff
@@ -0,0 +1,17 @@
+<header class="overlay" fxLayout="row" fxLayoutGap="20px" *ngIf="movieView">
+    <!-- Breadcrumbs and Poster -->
+    <!-- ATM we use the first image, but we will use geo localisation to determine which one to use -->
+    <img [imgRef]="movieView.poster.media" placeholderAsset="empty_poster.webp" class="poster">
+    <!-- Descriptions -->
+    <div fxLayout="column">
```
GrandSchtroumpf, line 6: I think we want to have the actions at the bottom of the header. So you sould use space-between here. and group everything except the actions.

libs/movie/src/lib/movie/components/header/header.component.html
```diff
@@ -0,0 +1,17 @@
...
+        <h4>{{ movieView.directors | displayName }}</h4>
+        <h1 class="mat-display-2">{{ movieView.title.international }}</h1>
+        <h2>{{ movieView.title.original }}</h2>
+        <movie-title-features [movie]="titleFeatures"></movie-title-features>
+        <ng-content select="movie-header-preferences"></ng-content>
+        <ng-content select="movie-header-cta"></ng-content>
+        <div fxLayout="row">
```
GrandSchtroumpf, line 13: This will have no impact on the children because they will be wrap inside a `movie-header-actions` tag, so the row only apply for the `movie-header-actions` here. 
You can remove this tag.

libs/movie/src/lib/movie/components/header/header.component.html
```diff
@@ -0,0 +1,17 @@
...
+            <ng-content select="movie-header-actions"></ng-content>
```
GrandSchtroumpf, line 14: If you want the movie-header-actions to be a row, you'll need to apply the style directly on the `.movie-header-actions` class. See comment below.
**This is not mandatory**. You can apply the row on each implementation of `movie-header-actions` if you prefer.

libs/movie/src/lib/movie/components/header/header.component.scss
```diff
@@ -0,0 +1,43 @@
+@import '~@angular/flex-layout/mq';
+@import 'images';
+
+:host {
+    background-size: contain;
+}
+
+.poster {
+    @include poster('large');
+    background-size: contain;
```
GrandSchtroumpf, line 10: Remove background related style.

libs/movie/src/lib/movie/components/header/header.component.scss
```diff
@@ -0,0 +1,43 @@
+@import '~@angular/flex-layout/mq';
+@import 'images';
+
+:host {
+    background-size: contain;
+}
+
+.poster {
+    @include poster('large');
```
GrandSchtroumpf, line 9: You should never set the width & height of an image tag. Only set the width or height here (I thing the width would be safer).

libs/movie/src/lib/movie/components/header/header.component.ts
```diff
@@ -0,0 +1,75 @@
...
+  selector: '[movie] movie-header',
+  templateUrl: './header.component.html',
+  styleUrls: ['./header.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class HeaderComponent {
+
+  public movieView: MovieHeaderView;
+
+  public titleFeatures: Movie;
```
GrandSchtroumpf, line 43: Just call it "movie".

libs/movie/src/lib/movie/components/header/header.component.scss
```diff
@@ -0,0 +1,43 @@
...
+    // TODO #2548
+    color: white !important;
+}
+
+@include layout-bp(xs) {
+    .poster {
+        @include poster('small');
+    }
+
+    .inner-header {
```
GrandSchtroumpf, line 34: This class doesn't exist anymore

libs/movie/src/lib/movie/components/header/header.component.scss
```diff
@@ -0,0 +1,43 @@
+@import '~@angular/flex-layout/mq';
+@import 'images';
+
+:host {
+    background-size: contain;
```
GrandSchtroumpf, line 5: If you do that you'll have some part of the background that won't be covered by the banner. Use "cover" instead.

libs/movie/src/lib/movie/components/header/header.component.scss
```diff
@@ -0,0 +1,43 @@
...
+}
+
+header {
+    padding: 50px 24px;
+}
+
+.overlay {
+    width: 100%;
+    background: linear-gradient(to right, black, transparent);
+}
```
GrandSchtroumpf, line 21: Here create a class `.movie-header-actions` with the display: flex.
Like that it will apply on the childrens of this directive. 
You'll need to set the ViewEncapsulation to None, so you'll need to wrap all the code here with `movie-header` :
```scss
movie-header {
  background-size: cover;
  backgorund-attachement: fixed;
  .poster {
    width: 30%; // or somehting like that
  }
  header {
    ...
  }
  ...
}
```

libs/movie/src/lib/movie/components/header/header.component.ts
```diff
@@ -0,0 +1,75 @@
...
+    },
+    banner: movie.promotionalElements.banner,
+    poster: movie.promotionalElements.poster[0]
+  }
+}
+
+@Component({
+  selector: '[movie] movie-header',
+  templateUrl: './header.component.html',
+  styleUrls: ['./header.component.scss'],
```
GrandSchtroumpf, line 36: To set the style on movie-header-actions you'll need to set `ViewEncapsulation.None`

libs/movie/src/lib/movie/components/title-features/title-features.component.html
```diff
@@ -0,0 +1,6 @@
+<h4 class="mat-body" *ngIf="features.length" fxLayout="row">
+    <span *ngFor="let feature of features; let isLast = last" 
+        [matTooltip]="feature.tooltip">
+        {{ feature.view }}<mat-divider *ngIf="!isLast" vertical></mat-divider>
+    </span>
```
GrandSchtroumpf, line 5: To make things nice & clean : 
```html
<ng-container *ngFor="let feature of features; let isLast = last" >
  <span [matTooltip]="feature.tooltip">
    {{ feature.view }}
  </span>
  <mat-divider *ngIf="!isLast" vertical></mat-divider>
</ng-container>
```

libs/movie/src/lib/movie/components/header/header.component.scss
Maxstgt: no why, its mandatory here!

GrandSchtroumpf: If the poster is an img tag you don't need a background size

libs/movie/src/lib/movie/components/header/header.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  public movieView: MovieHeaderView;
+
+  public titleFeatures: Movie;
+
+  @HostBinding('style.backgroundImage') background: string;
+  @Input()
+  set movie(movie: Movie) {
+    if (movie) {
+      this.titleFeatures = movie;
```
GrandSchtroumpf, line 45: Please call it "movie", as it is the movie and not the titleFeatures.
Variables name should describe what things are, not what they do.

Maxstgt: will collide with set input function. 
``` 
@Input('movie')
  set movieInput(movie: Movie) {
```

GrandSchtroumpf: Then `this.title = movie` should be good ^^

PR #2544, opened 2020-04-16
Feat 1589 - cleaning data model & unused code

phpgeekfr:


libs/movie/src/lib/movie/components/import/view-extracted-elements/movies/view-extracted-movies.component.ts
```diff
@@ -734,17 +732,19 @@ export class ViewExtractedMoviesComponent implements OnInit {
 
             switch (currency) {
               case '$':
-                movie.budget.budgetCurrency = 'USD';
+                movie.budget.totalBudget.currency = 'USD';
```
GrandSchtroumpf, line 735: We don't have a static model for that ?

phpgeekfr: sure we do

PR #2551, opened 2020-04-16
feat 2152 -  Prevent admin/devs to push data to analytics

phpgeekfr:


libs/utils/src/lib/analytics/app-analytics.ts
```diff
@@ -1,13 +1,26 @@
...
 
-@Injectable({providedIn: 'root'})
+@Injectable({ providedIn: 'root' })
 export class FireAnalytics {
-  constructor(public analytics: AngularFireAnalytics, private authQuery: AuthQuery) {}
+  constructor(public analytics: AngularFireAnalytics, private authQuery: AuthQuery) { }
 
   public event(name: AnalyticsEvents, params: Record<string, any>) {
+
+    if (this.authQuery.user && this.authQuery.orgId && (
```
GrandSchtroumpf, line 13: Create variables to me is more explicit : 
```typescript
const { user, orgId, isBlockframesAdmin } = this.authQuery;
const isOperator = isBlockframesAdmin || orgId === centralOrgID ;
if (user && orgId && isOperator)
```

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -15,6 +17,16 @@ export class AuthService extends FireAuthService<AuthState> {
     super(store);
   }
 
+  /**
+   * @dev This populates the RoleState part of the AuthState.
+   * Used to check if logged in user is blockframesAdmin or not.
+   */
+  selectRoles(user: FireBaseUser) {
+    return this.db.collection('blockframesAdmin').doc(user.uid).valueChanges().pipe(
+      map(doc => ({ blockframesAdmin: doc !== undefined ? true : false }))
```
GrandSchtroumpf, line 26: Maybe `doc(user.uid).snapshotChanges().pipe(map(doc => ({ blockframesAdmin: doc.exists }))`

PR #2553, opened 2020-04-17
new media model

Mussuranae:
New static object for picture' size.

### DIFF
--- a/libs/ui/src/lib/media/media.model.ts
+++ b/libs/ui/src/lib/media/media.model.ts
@@ -1,0 +1,21 @@
+const formats = {
+  avatar: {
+    height: 100,
+    width: 100
+  },
+  banner: {
+    height: 1080,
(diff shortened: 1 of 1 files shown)


libs/ui/src/lib/media/media.model.ts
```diff
@@ -0,0 +1,21 @@
+const formats = {
+  avatar: {
+    height: 100,
+    width: 100
+  },
+  banner: {
+    height: 1080,
+    width: 1920
```
Maxstgt, line 8: in the chat he wrote 190 px?

Maxstgt: but this seems better :P

Mussuranae: I guess it's a mistyping, a banner can't have a width of 190 :)

PR #2561, opened 2020-04-20
Feat 2461 email

### LINKED ISSUE #2461: Handle created invitation to event on the backend side
in `apps/backend-functions/src/invitation.ts` : update `onInvitationWrite` write to handle new type "event".

 * If we have an user or an org we can create a notification.
 * If we have an email, the function will send an email.
 * If user created an account, we will then be able to replace email by user

- [ ] @Mussuranae implement mails in `apps/backend-functions/src/internals/invitations/events.ts`
- [x] @phpgeekfr check `#2461` in code
- [x] @phpgeekfr deploy new functions and test the whole invitation/notifications flow.
- [x] @phpgeekfr implement cron backend function

Mussuranae:
Write a new email function to send email when an user is invited on an event and has no account on Archipel Content
Fix #2461


firestore.rules
```diff
@@ -52,7 +52,7 @@ service cloud.firestore {
     }
 
     /// INVITATIONS RULES ///
```
phpgeekfr, line 54: Since there is only linting changes, can you skip it ? The file is going to change a lot and we could avoid conflicts

libs/notification/src/lib/invitation/+state/invitation.service.ts
```diff
@@ -8,6 +8,7 @@ import { AuthService } from '@blockframes/auth/+state/auth.service';
 import { InvitationDocument } from './invitation.firestore';
 import { toDate } from '@blockframes/utils/helpers';
 
+
```
phpgeekfr, line 11: yes good catch ^^

apps/backend-functions/src/internals/invitations/events.ts
```diff
@@ -47,11 +49,11 @@ async function onInvitationToAnEventCreate({
     switch (mode) {
       case 'invitation':
         console.log(`Sending invitation email for a meeeting event (${eventId}) from ${senderEmail} to : ${recipient}`);
-        // @TODO (#2461) send invitation to a meeting email template
+        await sendMailFromTemplate(userInvitedToEvent(recipient, senderEmail, eventId));
```
phpgeekfr, line 52: can you rename `userInvitedToEvent` into `invitationToMeetingFromUser` to be more like `invitationToScreeningFromOrg` and also because, even if we don't have the case right now, recipient can be user or org email ?

apps/backend-functions/src/internals/invitations/events.ts
```diff
@@ -47,11 +49,11 @@ async function onInvitationToAnEventCreate({
...
         break;
       case 'request':
         console.log(`Sending request email to attend an event (${eventId}) from ${senderEmail} to : ${recipient}`);
-        // @TODO (#2461) send (invitation or screening) request email template
+        await sendMailFromTemplate(userRequestToEvent(senderEmail, recipient, eventId));
```
phpgeekfr, line 56: same, maybe something like requestToAttendEventFromUser ?

PR #2565, opened 2020-04-20
Event analytics view

Hugo-Hontans:
- [x] Table to see users of event analytics
- [x] Card to see the count of users
- [x] Create sync for one event

### DIFF
--- a/apps/festival/festival/src/app/dashboard/event/view/view.component.html
+++ b/apps/festival/festival/src/app/dashboard/event/view/view.component.html
@@ -7,6 +7,20 @@
     <invitation-guest-list [invitations]="invitations$ | async"></invitation-guest-list>
   </section>
 </ng-container>
+<h2>Statistics</h2>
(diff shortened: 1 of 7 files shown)


apps/festival/festival/src/app/dashboard/event/view/view.component.html
```diff
@@ -7,6 +7,34 @@ <h1>{{ event.title }}</h1>
...
+        showPaginator
+        [initialColumns]="initialColumns"
+        [columns]="columns"
+        [source]="analytics.eventUsers">
+      </bf-table-filter>
+    </section>
+  </ng-container>
+  <ng-template #noAnalytics>
+    <article fxLayout="column" fxLayoutAlign="center center">
+      <img asset="wait.webp" alt="waiting image"/>
```
Maxstgt, line 33: don't set words like `image` or `logo` in alt properties. Because the screen reader already know that this tag is about an image.

PR #2573, opened 2020-04-21
movie-slider

### LINKED ISSUE #2357: [Component] movie-slider
# Component

[Dark](https://app.zeplin.io/styleguide/5e0f7156cf562c1cb478f0f5/components?seid=5e8466c4dd077544c2a20867)
[Light](https://app.zeplin.io/styleguide/5e05fbe68925c3177b82ebd1/components?seid=5e846604489a7f48bfd5f3d8)

## UI review
_remove this section once everything is checked_
- [x] UI doesn't already have a existing component.
- [x] Name display in Zeplin is the same as the one in the issue (see below).
- [x] Data can be display like in the UI (see notion).
- [x] A link to this issue in the zeplin object

## Implement Component
- [x] Create story with storybook
- [x] Connect component to Zeplin with zeplin-connect

## Description

- name: SliderComponent - "movie-slider"
- lib/app: Movie

## Type
- [ ] **Layout**: _Structure a page with content projection (ex: MovieView)._
- [ ] **Page**: _Loaded by the router. Usually used in an app (ex: CatalogDashboardMovieView)._
- [x] **Component**: _Structure a piece of data in the page (ex: MovieTable)._
- [ ] **Display**: _Render data in a page. Can use a component-type (ex: MovieMain, MovieBudget)._
- [ ] **Form**: _Get data from the user (ex: MovieFormBudget)._

### Input
_can be data (movie, distribution right, ...), or UI related (dense, appearance)_
- [ ] movie

### Data view
_data displayed in the UI, derived from the inputs (ex: **title**: movie.main.title.international)
- [ ] title: movie.main.titles
- [ ] director: movie.main.directors[] 
- [ ] feature: ....
- [ ] description: movie...logline

## Where is it used
-links and snapshots where the component is used_

fritzschoff:
close #2357


libs/movie/src/lib/movie/components/slider/slider.component.html
```diff
@@ -0,0 +1,15 @@
+<bf-slider autoplay loop hideIndicators>
+    <bf-slide *ngFor="let view of movieViews; let i = index" overlayColor="linear-gradient(to right, black, transparent)">
+        <div fxLayout="column" fxLayoutAlign="start start">
+            <h4 *ngIf="view.directors">{{ view.directors | displayName }}</h4>
+            <h1 class="mat-display-2" *ngIf="view.titles.international">{{ view.titles.international }}</h1>
```
GrandSchtroumpf, line 5: Why do you need to specify `mat-display-2`. Is it not the default style of h1 ?
Also why do you need to use ngIf here ?

libs/movie/src/lib/movie/components/slider/slider.component.html
```diff
@@ -0,0 +1,15 @@
...
+            <h2 *ngIf="view.titles.original">{{ view.titles.original }}</h2>
+            <movie-title-features [movie]="titles[i]"></movie-title-features>
+            <ng-content select="movie-slider-cta"></ng-content>
+            <p class="mat-body" *ngIf="view.logline">
```
GrandSchtroumpf, line 9: Isn't `mat-body` the default style of "p" ?

libs/movie/src/lib/movie/components/slider/slider.component.ts
```diff
@@ -0,0 +1,55 @@
+// Angular
+import { Component, ChangeDetectionStrategy, Input, Directive } from '@angular/core';
+
+// Blockframes
+import { Movie } from '@blockframes/movie/+state';
+
+function createMovieSliderView(movies: Movie[]) {
```
GrandSchtroumpf, line 7: Couldn't you use the same as the header ?
I'm wondering if we need to recreate one each time or if we can have one in the movie.model that we could reuse.

libs/movie/src/lib/movie/components/slider/slider.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    return movies.map(movie => {
+        // UX decided to cut the logline to 335 characters
+        const shortedLogline = movie.story?.logline.length > 335
```
GrandSchtroumpf, line 10: I think it would be better to create a Pipe for that : `{{ logline | maxlength: 335 }}`

libs/movie/src/lib/movie/components/slider/slider.component.ts
```diff
@@ -0,0 +1,55 @@
...
+            ? movie.story?.logline.substr(0, 332).concat('...')
+            : movie.story?.logline
+        console.log(shortedLogline)
```
GrandSchtroumpf, line 13: remove console.log

libs/movie/src/lib/movie/components/slider/slider.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+@Component({
+    selector: '[movies] movie-slider',
+    templateUrl: './slider.component.html',
+    styleUrls: ['./slider.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class MovieSliderComponent {
+    titles: Movie[]
+    movieViews: MovieSliderView[];
```
GrandSchtroumpf, line 36: The ReturnType of createMovieSliderView is already a list.

libs/movie/src/lib/movie/components/slider/slider.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    @Input() set movies(movies: Movie[]) {
+        this.movieViews = createMovieSliderView(movies) as any[];
```
GrandSchtroumpf, line 38: Why do you need to do `as any[]` ?

libs/movie/src/lib/movie/components/title-features/title-features.component.html
```diff
@@ -1,4 +1,4 @@
-<h4 *ngIf="features.length" fxLayout fxLayoutGap="8px">
+<h4 *ngIf="features.length" fxLayout>
```
GrandSchtroumpf, line 1: I don't understand why you remove it. I you set `margin: 5px 0` to mat-divider it should work.

libs/movie/src/lib/movie/components/slider/slider.component.html
Maxstgt: No. 
If value is set to an empty string by the factory function, dont show it

libs/movie/src/lib/movie/components/slider/slider.component.ts
Maxstgt: I'm using logline here which we dont use in the header, also I don't use the poster here which we use in the header. I guess it is a question about where to draw the line. You can create such function and if no value is provided set it to an empty string. But this feels like beating the purpose of the new way we want to build the component? The component should only care about the properties which it needs.

libs/movie/src/lib/movie/components/title-features/title-features.component.html
Maxstgt: after the span there are some centimeters that give the illusion of that is not totally centered(the mat divider)

libs/movie/src/lib/movie/components/slider/slider.component.ts
```diff
@@ -0,0 +1,50 @@
+// Angular
+import { Component, ChangeDetectionStrategy, Input, Directive } from '@angular/core';
+
+// Blockframes
+import { Movie } from '@blockframes/movie/+state';
+
+function createMovieSliderView(movies: Movie[]) {
+    return movies.map(movie => {
+        return {
```
GrandSchtroumpf, line 9: I think you should have a function for one movie instead. It's not a good idea to make the array nature of the type implicit.
```typescript
function createView(movie: Movie) {
  return ...
}
type MovieSlideView = ReturnType<typeof createView>;
...
views: MovieSlideView[];

File: libs/movie/src/lib/movie/components/title-features/title-features.component.html (hunk 8 above)
Comment (GrandSchtroumpf) [reply libs/movie/src/lib/movie/components/title-features/title-features.component.html:1]: Because your span is not `text-aign: center` I think

File: libs/utils/src/lib/pipes/max-length.pipe.ts [hunk 10]
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+/**
+ * Cuts the text to provided length
+ * and replace it with 3 dots
+ */
+@Pipe({
+  name: 'maxLength'
```
GrandSchtroumpf, line 8: add `pure: true`

Maxstgt: pipes are pure by default

libs/movie/src/lib/movie/components/title-features/title-features.component.html
Maxstgt: it comes from the letter spacing.

Maxstgt: thats why I will work with the margin with the mat divider

libs/movie/src/lib/movie/components/slider/slider.component.html
```diff
@@ -0,0 +1,15 @@
+<bf-slider autoplay loop hideIndicators>
+    <bf-slide *ngFor="let view of movieViews; let i = index" overlayColor="gradient">
+        <div fxLayout="column" fxLayoutAlign="start start">
```
GrandSchtroumpf, line 3: Give it a class "overlay". To avoid targeting a "div" in CSS, it might lead to unexpected style error when maintaining the component.

libs/movie/src/lib/movie/components/slider/slider.component.scss
```diff
@@ -0,0 +1,11 @@
+h1,
+h2,
+h4,
+p {
+    // TODO #2548
+    color: white !important;
+}
+
+div {
+    padding: 80px 0 0 100px;
```
GrandSchtroumpf, line 10: Mathilde rarely use unsymetric padding. When the content is on the top/left corner Zeplin only gives you the padding on left & top. But think of it as a rect with the same padding top/bottom & left/right: 
`padding: 80px 100px`;

Alson, try it on different size of screen to make sure it doesn't break on responsive.

libs/movie/src/lib/movie/components/title-features/title-features.component.html
GrandSchtroumpf: Ok, it's weird, but ok

PR #2577, opened 2020-04-21
ops: add missing algolia upgrade users

lsenta:
add this new operation to the code that upgrade testing envs & production.


pldespaigne:
good catch !
but too late 😉 

https://github.com/blockframes/blockframes/blob/abbba550069c11c8667f9182a0c252db17a3275e/apps/backend-ops/src/firebaseSetup.ts#L28

PR #2594, opened 2020-04-21
Signed video url firebase function

pldespaigne:
- [x] add a firebase functions that check if a user is authorized to view a video and if so, return a signed jwplayer url
- [x] add a basic player page
- [x] add a fire storage -> jwplayer upload function, the function is commented due to #2643


libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -290,6 +290,9 @@ interface MovieRaw<D> {
   festivalPrizes: MovieFestivalPrizes;
   budget: MovieBudget;
   movieReview: MovieReview[];
+
+  // TODO discuss of what is the better way to store the JWPlayer id
```
GrandSchtroumpf, line 294: When ? where ? with who ? Did you invite people to a meeting ?

apps/backend-functions/src/player.ts
```diff
@@ -0,0 +1,169 @@
...
+import { isUserInvitedToScreening } from './internals/invitations/events';
+import { MovieDocument } from './data/types';
+import { jwplayerSecret } from './environments/environment';
+import { createHash } from 'crypto';
+
+// TODO issue#2643
+// No typing
+// const JWPlayerApi = require('jwplatform');
+
+interface ReadVideoParams {
```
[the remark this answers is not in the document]
pldespaigne: I don't know if we will need more field or not, in the future.
But the way `https.onCall` functions works is by passing a single object as a parameter, so I thought it was better to define an interface to enforce typing.

apps/backend-functions/src/player.ts
```diff
@@ -0,0 +1,169 @@
...
+      error: 'TOO_LATE',
+      result: `The event ${data.eventId} is finished`
+    };
+  }
+
+  if (!('titleId' in event.meta)) {
+    throw new Error(`Event ${data.eventId} is a screening but doesn't have a 'titleId' !`);
+  }
+
+  const movieSnapshot = await db.collection('movies').doc(event.meta.titleId).get();
```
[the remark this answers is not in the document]
pldespaigne: nope,  `db.collection<MovieDocument>(...)` give `Expected 0 type arguments, but got 1`

apps/backend-functions/src/player.ts
```diff
@@ -0,0 +1,169 @@
...
+
+interface ErrorResultResponse {
+  error: string;
+  result: any;
+}
+
+export const getPrivateVideoUrl = async (
+  data: ReadVideoParams,
+  context: CallableContext
+): Promise<ErrorResultResponse> => {
```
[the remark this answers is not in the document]
pldespaigne: always returning `{error: string, result: string}` we can then check if `error` is empty or not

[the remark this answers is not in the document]
GrandSchtroumpf: You cannot catch in a frontend if the function throw ?
```typescript
try {
  callfunction(params)
} catch(err)  {
  console.log('Something rong happen in the backend')
}
```

apps/backend-functions/src/player.ts
```diff
@@ -0,0 +1,169 @@
...
+      result: `You have not been invited to see this movie`
+    };
+  }
+
+  const toSign = `videos/${movie.hostedVideo}.mp4:${end.getTime()}:${jwplayerSecret}`;
+  const md5 = createHash('md5');
+
+  const signature = md5.update(toSign).digest('hex');
+
+  const signedUrl = `http://cdn.jwplayer.com/videos/${movie.hostedVideo}.mp4?exp=${end.getTime()}&sig=${signature}`;
```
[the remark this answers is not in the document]
pldespaigne: Might be over engineered as we will only use this function here.

apps/backend-functions/src/player.ts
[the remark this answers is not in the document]
GrandSchtroumpf: there is a method in data/internal that does that : 
```typescript
const movie = await getDocument<MovieDocument>(`movies/${event.meta.titleId}`)
```

apps/backend-functions/src/player.ts
```diff
@@ -0,0 +1,183 @@
...
+  if (!data.fileName) {
+    throw new Error(`No 'fileName' params, this parameter is mandatory !`);
+  }
+
+  if (!data.movieId) {
+    throw new Error(`No 'movieId' params, this parameter is mandatory !`);
+  }
+
+  const movieRef = db.collection('movies').doc(data.movieId);
+  const movieSnap = await movieRef.get();
```
GrandSchtroumpf, line 134: You can use `getDocument`. See above.

apps/backend-functions/src/player.ts
```diff
@@ -0,0 +1,183 @@
...
+      error: 'UNKOWN_MOVIE',
+      result: `There is no movie with id ${data.movieId}`
+    }
+  }
+
+  if (!context.auth) {
+    throw new Error(`Unauthorized call !`);
+  }
+
+  // TODO perform further check to see if user is authorized to upload a video for a given movie
```
GrandSchtroumpf, line 147: Create an issue for that and add a reference here.

apps/backend-functions/src/player.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  const storage = admin.storage();
+  const videoFile = await storage.bucket().file(`uploads/${data.fileName}`);
+  const [exists] =  await videoFile.exists(); // this line throw an unexpected Error
```
GrandSchtroumpf, line 151: `this line throw an unexpected Error` what should we do about that ?

apps/festival/festival/src/app/dashboard/dashboard.module.ts
```diff
@@ -77,6 +77,9 @@ const routes: Routes = [{
           },{
             path: 'edit',
             loadChildren: () => import('./event/edit/edit.module').then(m => m.EventEditModule)
+          }, {
+            path: 'screening',
```
GrandSchtroumpf, line 81: Why do you add the screening page in the dashboard 🤨

apps/festival/festival/src/app/marketplace/marketplace.module.ts
```diff
@@ -73,13 +74,18 @@ const routes: Routes = [{
...
         data: { animation: 'event-view' },
         children: [{
             path: '',
             loadChildren: () => import('./event/view/view.module').then(m => m.EventViewModule),
           }, {
             path: 'session',
-            loadChildren: () => import('./event/session/session.module').then(m => m.SessionModule),
+            loadChildren: () => import('@blockframes/event/components/session/session.module').then(m => m.EventSessionModule),
+          }, {
+            path: 'screening',
```
GrandSchtroumpf, line 87: I think (but I'm not sure) that if would be better to have a `jw-player` component and use it inside the session page if type of event is "screening"

libs/event/src/lib/components/player/player.component.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    // start loading the script
+    this.loadScript();
+  }
+
+  loadScript() {
+    const id = 'jwplayer-script';
+
+    // check if the script tag already exists
+    if (!document.getElementById(id)) {
```
GrandSchtroumpf, line 44: Never use document like that. Inject it in the constructor : `@Inject(DOCUMENT) private document: Document`

libs/event/src/lib/components/player/player.component.html
```diff
@@ -0,0 +1,2 @@
+<p>Player !</p>
```
GrandSchtroumpf, line 1: Can you manage the state when the player is not loaded yet ?

libs/event/src/lib/components/player/player.component.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  constructor(
+    private eventQuery: EventQuery,
+    private functions: AngularFireFunctions,
+  ) {}
+
+  ngOnInit() {
+    // listen for when the script is loaded AND the component html exists
+    this.sub = combineLatest([this.loaded, this.afterView]).pipe(
+      startWith([false, false]),
```
GrandSchtroumpf, line 31: You shouldn't need to startWith. The BehaviorSubject emit a first value always.

libs/event/src/lib/components/player/player.component.ts
```diff
@@ -0,0 +1,79 @@
...
+      filter(([loaded, afterView]) => loaded && afterView),
+      tap(() => this.initPlayer())
```
GrandSchtroumpf, line 33: Do this tap in the subscription

libs/event/src/lib/components/player/player.component.ts
```diff
@@ -0,0 +1,79 @@
...
+      script.onload = () => {
+        this.loaded.next(true);
+      }
+    } else {
+      this.loaded.next(true); // already loaded
+    }
+  }
+
+  ngAfterViewInit() {
+    this.afterView.next(true);
```
GrandSchtroumpf, line 59: Why not starting the subscription here instead ?
```typescript
this.isLoaded$.pipe(
  filter(isLoaded => isLoaded),
  first(), // Stop listening once it's loaded (you don't need to unsubscribe)
).subscribe(() => this.initPlayer())
```

GrandSchtroumpf: But honestly I whould just do everything here: 
```typescript
async ngAfterViewInit() {
  await this.loadScript();
  this.initPlayer();
}

loadScript() {
    if (this.document.getElementById(id)) {
      res()
    } else {
       ...
       script.onload = () => res()
    }
  })
}
```

apps/backend-functions/src/player.ts
pldespaigne: old error #2643 it's fixed now and I will close it with this PR, I just forget to remove the comment

PR #2599, opened 2020-04-22
Improve calendar & algolia-chips-autocomplete

GrandSchtroumpf:


apps/festival/festival/src/app/marketplace/title/list/list.component.html
```diff
@@ -37,7 +37,7 @@
 
     <ng-template filter label="Sales Agents">
       <algolia-chips-autocomplete
-        [index]="movieIndexName"
+        index="indexNameMovies"
```
pldespaigne, line 40: please also change this everywhere in the codebase, otherwise it will break several pages

GrandSchtroumpf: I found it only once

PR #2605, opened 2020-04-22
Issue#2308 cloud function

ioaNikas:
- Create resizing function for images uploaded to the storage
- Force image-cropper to generate webp images
- Update uuidv4() function to generate shorter string (google cloud storage doesn't handle very long names)

### DIFF
--- a/apps/backend-functions/src/internals/resize-images.ts
+++ b/apps/backend-functions/src/internals/resize-images.ts
@@ -1,0 +1,29 @@
+const gcs = require('@google-cloud/storage')();
+const os = require('os');
+const path = require('path');
+
+export async function onFileUploadEvent(event: any) {
+  const object = event.data;
+  const bucket = object.bucket;
+  const contentType = object.contentType;
+  const filePath = object.name;
+  console.log('File change detected, function execution started');
+
+  if (path.basename(filePath).startsWith('renamed--')) {
+    console.log('We already renamed that file !');
+    return;
+  }
+
+
+  const distBucket = gcs.bucket(bucket);
+  const tmpFilePath = path.join(os.tmpdir(), path.basename(filePath));
+  const metadata = { contentType };
+  return distBucket.file(filePath).download({
+    destination: tmpFilePath
+  }).then(() => {
+    return distBucket.upload(tmpFilePath, {
+      destination: 'renamed-' + path.basename(filePath),
+      metadata
+    });
+  })
+}
--- a/apps/backend-functions/src/internals/resize-images.ts
+++ b/apps/backend-functions/src/internals/resize-images.ts
@@ -1,29 +1,68 @@
-const gcs = require('@google-cloud/storage')();
-const os = require('os');
-const path = require('path');
+import { functions } from './firebase';
+import { tmpdir } from 'os';
+import { join, dirname } from 'path';
+import * as gcs from '@google-cloud/storage';
+import * as fs from 'fs-extra';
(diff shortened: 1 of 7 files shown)


apps/backend-functions/src/internals/resize-images.ts
```diff
@@ -0,0 +1,68 @@
+import { functions } from './firebase';
+import { tmpdir } from 'os';
+import { join, dirname } from 'path';
+import * as gcs from '@google-cloud/storage';
+import * as fs from 'fs-extra';
```
GrandSchtroumpf, line 5: It's usually a good practice to only import what you need : `import { ..., ... } form 'fs-extra'` to let Webpack know what he can treeshaked

apps/backend-functions/src/internals/resize-images.ts
```diff
@@ -0,0 +1,68 @@
...
+import sharp from 'sharp';
+
+const storage = new gcs.Storage();
+
+export async function onFileUploadEvent(data: functions.storage.ObjectMetadata) {
```
GrandSchtroumpf, line 10: A global comment to explain what this method is doing should be good.

apps/backend-functions/src/internals/resize-images.ts
```diff
@@ -0,0 +1,68 @@
...
+  // Get all the needed informations from the data (bucket, path, file name and directory)
+  const bucket = storage.bucket(data.bucket);
+  const filePath = data.name;
+
+  if (!filePath) {
+    return false;
+  }
+
```
GrandSchtroumpf, line 18: Filtrer by type. We don't have only images in the storage

apps/backend-functions/src/internals/resize-images.ts
```diff
@@ -0,0 +1,68 @@
...
+      break;
+    default:
+      console.warn('No bucket directory, exiting function');
+      return false;
+  }
+
+  // Iterate on each item of sizes array to generate all wanted resized images
+  const uploadPromises = sizes.map(async size => {
+    // Define a new name with prefix 'bf@' to recognize already resized images in the future
+    const resizedImgName = `bf@${size}_${fileName}`;
```
GrandSchtroumpf, line 52: I would have done the opposite `bf@${fileName}_${size}` to make it more readable.

apps/backend-functions/src/internals/resize-images.ts
```diff
@@ -0,0 +1,68 @@
...
+    const resizedImgPath = join(workingDir, resizedImgName);
+
+    // Use sharp to resize : take a path, resize to wanted size, save file to a new path.
+    await sharp(tmpFilePath).resize(size).toFile(resizedImgPath);
```
GrandSchtroumpf, line 56: Would be nice to extract everything related to the resize outside of the file/directory operations.
The goal would be to do somehting like that in the end : 
```typescript
try {
  const master = uploadMaster();
  resize();
  blur();
} catch (err) {
  console.log(err.message);
  return false;
}
```

libs/ui/src/lib/media/cropper/cropper/cropper.component.html
```diff
@@ -13,6 +13,7 @@
     <image-cropper
       [imageFile]="file"
       [aspectRatio]="cropRatio"
+      format= "webp"
```
GrandSchtroumpf, line 16: remove whitespace

PR #2607, opened 2020-04-22
Dark contrast theme

### LINKED ISSUE #2548: create high contrast theme dark/light
We came across this issue when we work on the movie header. We the background is a mix between linear-gradient and movie banner. In the design the font color didn't match the typography we got. So we need to create a high contrast theme.

Mussuranae:
Beginning of the contrast theme, actually only for the dark.
Fix #2548

### DIFF
--- a/apps/festival/festival/src/styles.scss
+++ b/apps/festival/festival/src/styles.scss
@@ -30,8 +30,6 @@
 .dark-contrast-theme {
   @include angular-material-theme($bf-dark-contrast-theme);
   @include bf-material-theme($bf-dark-contrast-theme);
-  @include bf-world-map-theme($bf-dark-contrast-theme);
-  @include sat-datepicker-theme($bf-dark-contrast-theme);
 
   // Not supported yet by browser. Here for future support
   scrollbar-color: dark;
--- a/apps/festival/festival/src/styles.scss
+++ b/apps/festival/festival/src/styles.scss
@@ -26,3 +26,13 @@
   // Not supported yet by browser. Here for future support
   scrollbar-color: dark;
 }
+
+.dark-contrast-theme {
+  @include angular-material-theme($bf-dark-contrast-theme);
+  @include bf-material-theme($bf-dark-contrast-theme);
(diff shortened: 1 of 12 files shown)


apps/festival/festival/src/styles.scss
```diff
@@ -26,3 +26,13 @@
   // Not supported yet by browser. Here for future support
   scrollbar-color: dark;
 }
+
+.dark-contrast-theme {
+  @include angular-material-theme($bf-dark-contrast-theme);
+  @include bf-material-theme($bf-dark-contrast-theme);
+  @include bf-world-map-theme($bf-dark-contrast-theme);
+  @include sat-datepicker-theme($bf-dark-contrast-theme);
```
GrandSchtroumpf, line 34: We don't use world-map & sat-darepicker in festival.

libs/movie/src/lib/movie/components/slider/slider.component.html
```diff
@@ -1,15 +1,15 @@
...
+<bf-slider loop hideIndicators class="dark-contrast-theme">
     <bf-slide *ngFor="let view of movieViews; let i = index" overlayColor="gradient">
         <div fxLayout="column" class="overlay">
             <h4 *ngIf="view.directors">{{ view.directors | displayName }}</h4>
             <h1 class="mat-display-2" *ngIf="view.titles.international">{{ view.titles.international }}</h1>
             <h2 *ngIf="view.titles.original">{{ view.titles.original }}</h2>
             <movie-title-features [movie]="titles[i]"></movie-title-features>
             <ng-content select="movie-slider-cta"></ng-content>
-            <p *ngIf="view.logline">
+            <p *ngIf="view.logline" class="mat-body">
```
GrandSchtroumpf, line 9: Is mat-body not the default value for p ?

Mussuranae: I saw no difference when I applied the contrast-theme but with the class, it was good.
I will add the `<p>` to the font-color default.

PR #2619, opened 2020-04-24
Data movie app access

### LINKED ISSUE #2591: Movie Model - App access
We need to differentiate movies from Content of those from Market with something like
```
appAccess: {
  catalog: boolean;
  festival: boolean;
}
```
- [x] Migration script for already existing films on Content (all the current movies can be tagged Content but need to be based on a list for future migrations, can't be by default)
- [x] Add this when creating a film on content or festival 
- [ ] Update the query so the films are loaded correctly

Mussuranae:
Add an AppAccess to movie
Fix #2591

### DIFF
--- a/apps/backend-ops/src/firestoreMigrations/0019.ts
+++ b/apps/backend-ops/src/firestoreMigrations/0019.ts
@@ -1,5 +1,21 @@
 import { Firestore } from '../admin';
 
+const updateStoreConfig = (oldStoreConfig) => {
+  if (oldStoreConfig) {
+    if (!oldStoreConfig.appAccess) {
+      return {
+        ...oldStoreConfig,
+        appAccess: {
+          catalog: true,
+          festival: false
+        }
+      };
+    } else {
+      return oldStoreConfig;
+    }
+  }
+};
+
 /**
  * Add appAccess with catalog and festival booleans in storeConfig in movie documents.
  */
@@ -9,22 +25,6 @@ export async function upgrade(db: Firestore) {
   const newMovieData = movies.docs.map(
     async (movieDocSnapshot: any): Promise<any> => {
       const movieData = movieDocSnapshot.data();
-
-      const updateStoreConfig = (oldStoreConfig) => {
-        if (oldStoreConfig) {
-          if (!oldStoreConfig.appAccess) {
-            return {
-              ...oldStoreConfig,
-              appAccess: {
-                catalog: true,
-                festival: false
-              }
-            };
-          } else {
-            return oldStoreConfig;
-          }
-        }
-      };
 
       const storeConfig = updateStoreConfig(movieData.main.storeConfig);
 
--- a/apps/backend-ops/src/firestoreMigrations/0019.ts
+++ b/apps/backend-ops/src/firestoreMigrations/0019.ts
@@ -1,0 +1,45 @@
+import { Firestore } from '../admin';
+
+/**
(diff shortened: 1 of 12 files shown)


apps/backend-ops/src/firestoreMigrations/0019.ts
```diff
@@ -0,0 +1,45 @@
...
+              appAccess: {
+                catalog: true,
+                festival: false
+              }
+            };
+          } else {
+            return oldStoreConfig;
+          }
+        }
+      };
```
phpgeekfr, line 27: Weird way of declaring a function. It will be re-created for each loop of movies

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -1,40 +1,64 @@
...
     protected store: MovieStore,
   ) {
     super(store);
   }
 
-  async onCreate(movie: Movie, { write }: WriteOptions) {
-    // When a movie is created, we also create a permissions document for it.
-    // Since movie can be created on behalf of another user (An admin from admin panel for example)
-    // We use createdBy attribute to fetch OrgId
-    const userId = movie._meta?.createdBy ? movie._meta.createdBy : this.authQuery.userId;
```
phpgeekfr, line 35: Please do not remove code that have a comment explaining  why it is here

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -1,40 +1,64 @@
...
+    await this.runTransaction(async (tx) => {
+
+      // Add movie and update the organization to add the movie ID to the org
+      movieId = await this.add(cleanModel(movie), { write: tx });
+      await this.orgService.update(this.orgQuery.getActiveId(), (org) => ({ movieIds: [...org.movieIds, movieId] }), { write: tx });
+
+      // When a movie is created, we also create a permissions document for it.
+      // Since movie can be created on behalf of another user (An admin from admin panel for example)
+      // We use createdBy attribute to fetch OrgId
+      this.permissionsService.addDocumentPermissions(movieId, tx, this.orgQuery.getActiveId());
```
phpgeekfr, line 59: You should use onCreate for this.

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -1,40 +1,64 @@
...
+import { OrganizationQuery } from '@blockframes/organization/+state/organization.query';
 
 @Injectable({ providedIn: 'root' })
 @CollectionConfig({ path: 'movies' })
 export class MovieService extends CollectionService<MovieState> {
+
+
   constructor(
     private authQuery: AuthQuery,
     private userService: UserService,
```
phpgeekfr, line 26: remove unused

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -1,40 +1,65 @@
...
+    private routerQuery: RouterQuery,
     private functions: AngularFireFunctions,
     private query: MovieQuery,
     protected store: MovieStore,
   ) {
     super(store);
   }
 
+  async create(movieImported?: Movie): Promise<string> {
+    const createdBy = this.authQuery.getValue().uid;
```
phpgeekfr, line 37: you should use `this.authQuery.userId` instead of retrieving the whole document

Issue #2637, opened 2020-04-27
Debug issue - Archipel Content


Jackseed:
- [x] Add a pipe for running time: add « min » only on numbers, should be use everywhere running time is displayed
![image](https://user-images.githubusercontent.com/14964443/80374589-29be5480-8897-11ea-825c-f3faf69d0390.png)

- [x] Add the org activity under the org name on c/o/organization/orgId/view/org, cf 
￼
![image](https://user-images.githubusercontent.com/14964443/80374736-72760d80-8897-11ea-88e6-5d45b16260ca.png)
Vs 
![image](https://user-images.githubusercontent.com/14964443/80374742-74d86780-8897-11ea-8b54-fc0368fdc26a.png)

- [x] Align first name column with the « First Name » label on org tab members
![image](https://user-images.githubusercontent.com/14964443/80374766-7e61cf80-8897-11ea-857b-90d5e50114dc.png)

- [x] Sort by title / director not working on /search page, fix it and sort by title by default
- [x] Center the « close » button on snackbar
￼
![image](https://user-images.githubusercontent.com/14964443/80374814-91749f80-8897-11ea-8bf6-6d667873b2fe.png)

- [x] `marketplace/selection` move to wishlist button should remove the title from the cart (and all distribution rights attached) in addition to adding it to wishlist

- [x] When I go from `marketplace/deals` to `marketplace/search` by clicking in « Explore the Library » button  got 0 titles, titles are loading fine when reloading the page
￼
![image](https://user-images.githubusercontent.com/14964443/80374921-b8cb6c80-8897-11ea-89ed-57895a10cdca.png)
￼
![image](https://user-images.githubusercontent.com/14964443/80374933-bb2dc680-8897-11ea-8baa-4651711b0880.png)

- [x] `marketplace/selection` Add a delete button on each line of the exploitation rights table on the right
￼
![image](https://user-images.githubusercontent.com/14964443/80374959-c680f200-8897-11ea-9a86-6475a3afec57.png)

- [ ] After org creation, add a "Go to Dashboard" or "Go to Marketplace" CTA button depending on the org access. Button must be disabled while org is not accepted

- [x] FESTIVAL: Use the new banner image from Mathilde (used [here](https://projects.invisionapp.com/d/main#/console/19607961/413916192/preview))
- [x] Display only films with main.storeConfig.status « accepted » on marketplace/search
[...]

PR #2644, opened 2020-04-28
Issue 2637

fritzschoff:
ref #2637


libs/contract/src/lib/contract/+state/contract.query.ts
```diff
@@ -16,6 +16,11 @@ export class ContractQuery extends QueryEntity<ContractState> {
   public activeVersion$ = this.selectActive(contract => contract.lastVersion);
 
   public activeVersionView$ = this.activeVersion$.pipe(
+    /*
+    * If user goes back, this will call the getVersionView
+    * with an undefined value, and this throws an error 
+     */
```
GrandSchtroumpf, line 22: Put this comment on `// on line` please.

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -53,6 +53,13 @@ export class ContractService extends CollectionService<ContractState> {
     return cleanContract(contract);
   }
 
+  formatFromFirestore(contract: Contract) {
+    contract.versions.forEach(version => {
```
GrandSchtroumpf, line 57: Use `forEach` only when you need to have the index & element. Prefer for ... of  ... to do some mutations: 
```typescript
for (const version of contract.versions) {
  ..
}
```

libs/movie/src/lib/movie/guards/movie-collection.guard.ts
```diff
@@ -15,15 +15,12 @@ const movieListWithRightsQuery = () => ({
 });
 
 @Injectable({ providedIn: 'root' })
+@CollectionGuardConfig({ awaitSync: true })
```
GrandSchtroumpf, line 18: Why would you wait if there is already movies in the store ?

libs/notification/src/lib/invitation/guard/invitations.guard.ts
```diff
@@ -15,7 +16,11 @@ export class InvitationGuard extends CollectionGuard<InvitationState> {
   /** This sync on invitations where userId is the same as the connected user id */
   sync() {
     return this.authQuery.user$.pipe(
-      switchMap(user => this.service.syncCollection(ref => ref.where('toUser.uid', '==', user.uid)))
+      // If we logout we will need to make sure user exist, otherwise we get an error
```
GrandSchtroumpf, line 19: If the user logout we should not be in this guard anymore. You can add a `filter(user => !!user.uid)` to avoid error "while navigating to login page".

libs/utils/src/lib/helpers.ts
```diff
@@ -27,7 +27,10 @@ export function toDate(date: firestore.Timestamp | Date): Date {
   if (isTimeStamp(date)) {
     return date.toDate();
   }
-  if (typeof date === 'string' || typeof date === 'number') {
+  if (typeof date === 'number') {
+    return new Date(date * 1000)
```
GrandSchtroumpf, line 31: Why would you multiple by 1000 ?
This is quite dangerous as we expect the date number to be in milliseconde already.

libs/utils/src/lib/pipes/runtime.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'runtime'
+})
+
+export class RuntimePipe implements PipeTransform {
+  transform(value: string | number): string {
```
GrandSchtroumpf, line 8: Right now, this specific filter should be in @blockframes/movies/pipes as it's specific to runtime (due to 'TBC').
BUT, I think we can make it generic like that :
```typescript
transform(value: number | string | { start: number | Date, end: number | Date }, base: 'ms' | 's' | 'min' = 'ms', ..., placeholder: string) {
  if (!value) {
    return placeholder;
  }
  if (typeof value === 'string') {
    return value;
  }
  let duration = typeof value === 'object'
    ? value.end - value.start
     : value;
  // Do the operation to transform a number into 1h 31min 32sec depending on the base.
  // 132 base "min" -> 2h 12min
  // 120 000 no base ('ms') -> 2min
}
```

styles/components/mat-button.scss
```diff
@@ -48,6 +48,11 @@
     padding: 14px 32px;
   }
 
+  // Center close button
+  .mat-simple-snackbar .mat-button {
```
GrandSchtroumpf, line 52: I don't think it's a good idea. This make sense only when there is no text.

libs/utils/src/lib/pipes/runtime.module.ts
```diff
@@ -0,0 +1,11 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+
+import { RuntimePipe } from './runtime.pipe';
+
+@NgModule({
+  imports: [CommonModule],
```
GrandSchtroumpf, line 7: You don't need to import CommonModule

libs/utils/src/lib/pipes/runtime.module.ts
```diff
@@ -0,0 +1,11 @@
...
+  exports: [RuntimePipe],
+  declarations: [RuntimePipe],
+})
+export class RuntimeModule { }
```
GrandSchtroumpf, line 11: I would recommend to put the module in the same file as the pipe to make things more light.

libs/organization/src/lib/organization/pages/view/view.component.html
```diff
@@ -4,6 +4,7 @@
       <img [imgRef]="org.logo" placeholderAsset="Organization_250.png" alt="organization logo" />
       <div class="org-info">
         <h1>{{ org.denomination.full }}</h1>
+        <h3>{{ org.activity }}</h3>
         <ng-container *ngIf="user$ | async as user">
```
Jackseed, line 8: I hadn't seen this before, but it shouldn't be here, we only display org info here, not the user's position

apps/catalog/catalog/src/app/marketplace/movie/selection/selection.component.html
```diff
@@ -12,7 +12,7 @@ <h1 i18n *ngIf="count$ | async as count">{{count}} title{count, plural, =0 {} =1
     <article *ngFor="let title of titles">
       <movie-banner [movie]="selectMovie(title.movieId) | async">
         <movie-banner-actions fxLayout="column" fxLayoutAlign="start end" fxLayoutGap="16px">
-          <wishlist-button [movieId]="title.movieId"></wishlist-button>
+          <wishlist-button [movieId]="title.movieId" (click)="removeTitle(title.movieId)"></wishlist-button>
```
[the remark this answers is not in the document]
fritzschoff: vincent wants to remove the movie when you add it to the wishlist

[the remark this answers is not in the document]
fritzschoff: remove it from the cart so to say

libs/movie/src/lib/movie/guards/movie-collection.guard.ts
fritzschoff: as you can see below I removed the awaitSync function which result into false when one movie is active. This will lead to only one movie present in `search` . So we need to wait for all of them

fritzschoff: and we need them in the store, cause on the search page we startWith(this.query.selectAll()). But depends on how we proceed if we want to exclude akita in some situation

libs/utils/src/lib/helpers.ts
fritzschoff: cause if it is a unix timestamp you need to *1000 when you want to have a valid js date. otherwise 
``` 
new Date(unixTimestamp) = 1/19/1970

File: styles/components/mat-button.scss (hunk 7 above)
Comment (fritzschoff) [reply styles/components/mat-button.scss:52]: I guess it is center in his column

File: apps/catalog/catalog/src/app/marketplace/movie/selection/selection.component.html (hunk 11 above)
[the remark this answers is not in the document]
Comment (GrandSchtroumpf) [reply apps/catalog/catalog/src/app/marketplace/movie/selection/selection.component.html:15]: Create an output on the whishlis-button (added) / (removed)

File: libs/movie/src/lib/movie/guards/movie-collection.guard.ts (hunk 3 above)
Comment (GrandSchtroumpf) [reply libs/movie/src/lib/movie/guards/movie-collection.guard.ts:18]: I think this is typically a situation where we make thinks worth by accumulating a bunch of small fixes. I would recommand to remove the guard & do `movieService.valueChanges(...)` in the component directly.

File: libs/utils/src/lib/helpers.ts (hunk 5 above)
Comment (GrandSchtroumpf) [reply libs/utils/src/lib/helpers.ts:31]: The ("typeof date === "number") case is when you do something like `Date.now()` and in this case, this is already in ms.
In the case of Firestore Timestamp it's in the (isTimestam(date)) case.

File: libs/organization/src/lib/organization/pages/view/view.component.html (hunk 10 above)
Comment (GrandSchtroumpf) [reply libs/organization/src/lib/organization/pages/view/view.component.html:8]: If you look here (scroll to Anastasiia wireframe) there is the company role :
https://github.com/blockframes/blockframes/issues/2640

File: libs/utils/src/lib/helpers.ts (hunk 5 above)
Comment (fritzschoff) [reply libs/utils/src/lib/helpers.ts:31]: image
image
If I remove the if statement === number

Comment (fritzschoff) [reply libs/utils/src/lib/helpers.ts:31]:
```
new Date(1588172208)
Mon Jan 19 1970 10:09:32 GMT+0100 (Central European Standard Time)
``` 
``` 
new Date(1588172208*1000)
Wed Apr 29 2020 16:56:48 GMT+0200 (Central European Summer Time)
```

Comment (fritzschoff) [reply libs/utils/src/lib/helpers.ts:31]: I see your concern about checking for number, but you can check for length of the unixTimestamp, maybe?

File: apps/catalog/catalog/src/app/marketplace/movie/search/search.component.ts [hunk 12]
```diff
@@ -32,23 +32,22 @@ export class MarketplaceSearchComponent implements OnInit {
...
     this.filterForm.storeConfig.add('accepted');
-    this.dynTitle.setPageTitle('Titles')
-    this.movieSearchResults$ = this.filterForm.valueChanges.pipe(
+    this.dynTitle.setPageTitle('Titles');
+    this.movieSearchResults$ = combineLatest([this.filterForm.valueChanges, this.sortByControl.valueChanges])
+    .pipe(
       debounceTime(300),
       distinctUntilChanged(),
-      switchMap(() => this.filterForm.search()),
+      switchMap(observables => observables[0] = this.filterForm.search()),
```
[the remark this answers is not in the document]
fritzschoff: https://rxjs-dev.firebaseapp.com/api/operators/withLatestFrom 
this won't work, unfortunately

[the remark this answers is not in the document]
fritzschoff: cause it emits only when the two source emits. but maybe the user doenst click on sort select.

libs/movie/src/lib/movie/guards/movie-collection.guard.ts
fritzschoff: I will remove it in the route, cause we also need to use akita for home page. For the slider.

apps/catalog/catalog/src/app/dashboard/title/view/view.module.ts
```diff
@@ -1,9 +1,15 @@
+// Angular
 import { NgModule } from '@angular/core';
 import { CommonModule } from '@angular/common';
 import { RouterModule } from '@angular/router';
 import { FlexLayoutModule } from '@angular/flex-layout';
+
+// Blockframes
 import { ImageReferenceModule } from '@blockframes/ui/media/image-reference/image-reference.module';
 import { TranslateSlugModule } from '@blockframes/utils/pipes/translate-slug.module';
+import { DurationModule } from '@blockframes/utils/pipes/duration.pipe';
```
GrandSchtroumpf, line 10: 👍

libs/utils/src/lib/akita-helper/sort-movie-by.ts
```diff
@@ -5,7 +5,7 @@ export function sortMovieBy(a: Movie, b: Movie, sortIdentifier: string) {
     case 'Title':
       return a.main.title.international.localeCompare(b.main.title.international);
     case 'Director':
-      return a.main.directors[0].lastName.localeCompare(b.main.directors[0].lastName);
+      return a.main.directors[0]?.lastName.localeCompare(b.main.directors[0]?.lastName);
```
GrandSchtroumpf, line 8: Are we still using this file ? Looks like it's deprecated now.

PR #2646, opened 2020-04-28
Improve invitation process

GrandSchtroumpf:


libs/notification/src/lib/invitation/+state/invitation.service.ts
```diff
@@ -27,6 +32,14 @@ export class InvitationService extends CollectionService<InvitationState> {
     }
   }
 
+  formatToFirestore(invitation: Invitation) {
+    for (const key in invitation) {
+      if (typeof invitation[key] === 'undefined') delete invitation[key];
+    }
+    console.log(invitation);
```
phpgeekfr, line 39: forgot to remove ?

PR #2663, opened 2020-05-01
Cleaning 

phpgeekfr:


libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -45,8 +45,18 @@ export class OrganizationService extends CollectionService<OrganizationState> {
...
+   * This converts the OrganizationDocument into an Organization
+   * @param org
+   * @dev If this method is implemented, remove akitaPreAddEntity and akitaPreUpdateEntity on store
+   */
+  formatFromFirestore(org: OrganizationDocument): Organization {
+    return {
+      ...org,
+      created: toDate(org.created), // prevent error in case the guard is wrongly called twice in a row
+      updated: toDate(org.updated),
+      wishlist: convertWishlistDocumentToWishlistDocumentWithDate(org.wishlist)
```
GrandSchtroumpf, line 58: Could you rename the function ? Looks like the .net library ^^

Issue #2539, opened 2020-04-15
Organization acceptance or rejection and organization access on application level


phpgeekfr:
Currently, the organization permission at application level is disabled and the code is commented because we only handle dashboard/marketplace level.

If we need to enable it again, here is what should be done.
- [x] Move the admin button to accept or reject an org from the backend express server to the admin-panel
- [x] Move the admin button to accept or reject an org application request from the backend express server to the admin-panel
- [x] Check for `#2539` in code to uncomment

example:
```
appAccess: {
  catalog {
    marketplace: true, 
    dashboard: false
  }
  festival {
    marketplace: true, 
    dashboard: false
  }
}
```

- [x] Also clean invitations related to organizations (no need to create an interface for each notification type) and remove createInvitationXXX

PR #2667, opened 2020-05-01
WIP #2539 bis

phpgeekfr:


apps/backend-functions/src/orgs.ts
```diff
@@ -73,6 +76,25 @@ async function notifyOnOrgMemberChanges(before: OrganizationDocument, after: Org
   }
 }
 
+/** Checks if new org admin updated app access (possible only when org.status === 'pending' for a standard user ) */
+function hasOrgAppAccessChanged(before: OrganizationDocument, after: OrganizationDocument): boolean {
+  if (!!after.appAccess && before.status === 'pending' && after.status === 'pending') {
```
GrandSchtroumpf, line 81: Why do you check is status of both is pending ? If one is accepted for example and not the other what should happen ?

apps/backend-functions/src/orgs.ts
```diff
@@ -73,6 +76,25 @@ async function notifyOnOrgMemberChanges(before: OrganizationDocument, after: Org
...
+    for (const a of app) {
+      if (after.appAccess[a].dashboard === true && (!before.appAccess[a] || before.appAccess[a].dashboard === false)) { return true; }
```
GrandSchtroumpf, line 83: Create a function with a good naming : 
```typescript
const accessChanges = (module: 'dashboard' | 'marketplace') => {
  return after.appAccess[a][module] === true && (!before.appAccess[a] || before.appAccess[a][module]  === false)
}
if (accessChanges('dashboard') || accessChanges('marketplace')) {
  return true;
}
```

apps/backend-functions/src/templates/mail.ts
```diff
@@ -149,40 +149,37 @@ export function requestToAttendEventFromUser(fromUserFirstname: string, fromUser
...
+/**
+ * @param orgId
+ */
 const organizationCreatedTemplate = (orgId: string) =>
   `
   A new organization was created on the blockframes project,
 
-  Visit ${appUrl}${ADMIN_ACCEPT_ORG_PATH}/${orgId} to enable it.
-  `; // @TODO (#2539) => redirect to admin panel
+  Visit ${appUrl}${ADMIN_ACCEPT_ORG_PATH}/${orgId} or go to ${ADMIN_ACCEPT_ORG_PATH}/${orgId} to view it.
```
GrandSchtroumpf, line 163: why not adding `${appUrl}` in front for the path ?

apps/backend-functions/src/templates/mail.ts
```diff
@@ -193,17 +190,15 @@ export function organizationCreated(orgId: string): EmailRequest {
...
- * It sends and email to admin to accept or reject the request
+ * It sends an email to admin to accept or reject the request
  */
-/*export function organizationRequestedAccessToApp(orgId: string, appId: string): EmailRequest {
+export function organizationRequestedAccessToApp(orgId: string): EmailRequest {
   return {
     to: adminEmail,
     subject: 'An organization requested access to an app',
-    text: organizationRequestAccessToAppTemplate(orgId, appId)
+    text: organizationRequestAccessToAppTemplate(orgId)
```
GrandSchtroumpf, line 199: Could we do something like in the frontend later ? `request(orgId).from('org').to('accessApp')` or something like that

libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -107,32 +109,13 @@ export class OrganizationService extends CollectionService<OrganizationState> {
...
-    });
-  }*/
-
-
   public async setBlockchainFeature(value: boolean) {
     const orgId = this.query.getActiveId();
     return this.update(orgId, { isBlockchainEnabled: value });
   }
 
+  public async notifyAppAccessChange(orgId: string){
```
GrandSchtroumpf, line 117: yu don't need async here

libs/utils/src/lib/apps.ts
```diff
@@ -45,4 +37,14 @@ export function createInnerAppAccess(innerAppAccess: Partial<InnerAppAccess> = {
     marketplace: false,
     ...innerAppAccess
   }
+}
+
+export function setAppName(slug: App) {
+  switch (slug) {
+    case 'catalog':
```
GrandSchtroumpf, line 44: Put catalog with default at the end.

libs/utils/src/lib/apps.ts
```diff
@@ -45,4 +37,14 @@ export function createInnerAppAccess(innerAppAccess: Partial<InnerAppAccess> = {
     marketplace: false,
     ...innerAppAccess
   }
+}
+
+export function setAppName(slug: App) {
```
GrandSchtroumpf, line 42: It looks like a getter, not a setter.
Why not using a map ?
```typescript
const appName = {
  catalog: 'Archipel Content',
  ...
};
```

libs/utils/src/lib/apps.ts
```diff
@@ -7,17 +7,17 @@ import { RouterQuery } from "@datorama/akita-ng-router-store";
...
 
-export interface InnerAppAccess { marketplace: boolean, dashboard: boolean };
-export type OrgAppAccess = Record<App, InnerAppAccess>;
-export type MovieAppAccess = Record<App, boolean>;
+export const module = ['dashboard', 'marketplace'] as const;
+export type Module = typeof module[number];
+
+const appName = {
+  catalog: 'Archipel Content',
+  festival: ''
```
Mussuranae, line 15: `festival: 'Archipel Market'` ?

apps/backend-functions/src/templates/mail.ts
phpgeekfr: Yes we could, but right now, org can change it's app access at creation only

PR #2677, opened 2020-05-04
scroll bug

### LINKED ISSUE #2502: scroll to top in mat-sivenav-content
since we layouting  all of our apps in a mat sidenav the router cant set the scrollY position to 0 because the content you want to be on top is the content of the mat side nav but not the mat sidenav itself. 
We have to investigate how we want to restore the scroll position.
- Either we can manage to inject the DOCUMENT element in the VIEWPORTSCOLLER from angular and say, this is what we care about for scrolling.
- Or we manage do notify the router from the component level and say "hey, this is the main content you want to care about"
There are also some helpful links to github issues about that in this issue ref #2123 
Since we are before release 1.3 there is no time to investigate but should be done in the future to handle this scroll position issue.

fritzschoff:
close #2502


libs/ui/src/lib/layout/dashboard/dashboard.component.html
```diff
@@ -41,8 +41,10 @@
       </header>
     </app-bar>
     <!-- CONTENT -->
-    <main fxFlex>
-      <router-outlet></router-outlet>
-    </main>
+    <cdk-virtual-scroll-viewport itemSize="1">
```
GrandSchtroumpf, line 44: You don't need a virtual scroll viewport, you need `cdkScrollable` : `<main cdkScrollable fxFlex>`

libs/ui/src/lib/layout/dashboard/dashboard.component.ts
```diff
@@ -31,8 +37,7 @@ export class DashboardComponent implements AfterViewInit, OnDestroy {
   public algoliaSearchResults$: Observable<SearchResult[]>;
 
   @ViewChild(MatSidenav) sidenav: MatSidenav;
-  @ViewChild(MatSidenavContent) sidenavContent: MatSidenavContent;
-
+  @ViewChild(CdkVirtualScrollViewport) scrollViewport: CdkVirtualScrollViewport;
```
GrandSchtroumpf, line 40: After comment from above, use `@ViewChild(CdkScrollable)`

libs/ui/src/lib/tunnel/layout/layout.component.ts
```diff
@@ -43,23 +45,38 @@ export class TunnelLayoutComponent implements OnInit {
...
     // Share the url by navigation plan
     this.urlBynav$ = combineLatest([this.url$, this.steps$]).pipe(shareReplay());
-    this.next$ = this.urlBynav$.pipe(map(([ url, steps ]) => getPage(steps, url, 1)));
-    this.previous$ = this.urlBynav$.pipe(map(([ url, steps ]) => getPage(steps, url, -1)));
+    this.next$ = this.urlBynav$.pipe(map(([url, steps]) => getPage(steps, url, 1)));
+    this.previous$ = this.urlBynav$.pipe(map(([url, steps]) => getPage(steps, url, -1)));
+    // https://github.com/angular/components/issues/4280
+    this.sub = this.router.events.pipe(
+      filter(event => event instanceof NavigationEnd))
+      .subscribe(() => {
```
GrandSchtroumpf, line 73: Do it on one line `subscribe(() =>  this.sidenavContent.scrollTo({ top: 0 }))`

PR #2705, opened 2020-05-06
watermark

pldespaigne:
- [x] generate a watermark in the backend
- [x] display the watermark in the player


apps/backend-functions/src/internals/resize-images.ts
```diff
@@ -18,6 +18,11 @@ export async function onFileUploadEvent(data: functions.storage.ObjectMetadata)
     return false;
   }
 
+  if (data.contentType === 'image/svg+xml') {
```
GrandSchtroumpf, line 21: Add a comment to explain why we shouldn't have that.
What about the other contentType (text, script, ...) ?

apps/backend-functions/src/internals/watermark.ts
```diff
@@ -0,0 +1,36 @@
...
+import { admin, db, getStorageBucketName } from './firebase';
+
+/**
+ * - Generate a svg file with the name & email of the user
+ * - Store the watermark file in the storage bucket
+ * - Update the user's firestore doc to link it to the watermark file
+ */
+export async function upsertWatermark(user: PublicUser) {
+
+  if (!user.firstName || !user.lastName) {
```
GrandSchtroumpf, line 12: You don't check for the email ?

apps/backend-functions/src/internals/watermark.ts
```diff
@@ -0,0 +1,36 @@
...
+  const ref = `watermark/${user.uid}.svg`;
+  const file = admin.storage().bucket(getStorageBucketName()).file(ref);
+
+  await new Promise(res => {
+    file.createWriteStream({contentType: 'image/svg+xml'}).end(watermark, () => res());
+  });
+
+  const signedUrl = await file.getSignedUrl({action: 'read', expires: '01-01-3000', version: 'v2'});
+
+  user.watermark = {
```
GrandSchtroumpf, line 27: Be carefull the ImgRef model will evolve wih @ioaNikas's PR on multile size image.

apps/backend-functions/src/templates/watermark.ts
```diff
@@ -0,0 +1,14 @@
+
+export function getWatermark(firstName: string, lastName: string, email: string) {
+  return `
+    <svg viewBox="0 0 640 360" xmlns="http://www.w3.org/2000/svg">
+      <style>
```
GrandSchtroumpf, line 5: I'm not sure, but does this only apply on the current svg scope or is going to apply everywhere ?

apps/backend-functions/src/users.ts
```diff
@@ -77,17 +78,43 @@ export const onUserCreate = async (user: UserRecord) => {
   // update Algolia index
   const userSnap = await userDocRef.get();
   const userData = userSnap.data() as PublicUser;
-  return storeSearchableUser(userData);
+
+  const promises: Promise<any>[] = [];
+  promises.push(storeSearchableUser(userData));
+  promises.push(upsertWatermark(userData));
```
GrandSchtroumpf, line 84: Ok. Just an alternative here
```typescript
return Promise.all([
  storeSearchableUser(userData),
  upsertWatermark(userData)
])
```

apps/backend-functions/src/users.ts
```diff
@@ -77,17 +78,43 @@ export const onUserCreate = async (user: UserRecord) => {
...
+  }
+
+  // if name or email has changed : update watermark
+  if (
+    before.firstName !== after.firstName ||
+    before.lastName !== after.lastName ||
+    before.email !== after.email
+  ) {
+    promises.push(upsertWatermark(after));
+  }
```
GrandSchtroumpf, line 115: Better : 
```typescript
const changed = (keys: string[]) => keys.some(key => before[key] !== after[key]);
if (changed['firstName', 'lastName', 'email']) {
  if (before.avatar?.url !== after.avatar?.url) {
    promises.push(storeSearchableUser(after));
  }
  promises.push(upsertWatermark(after));
}
```

libs/event/src/lib/components/player/player.component.ts
```diff
@@ -1,14 +1,16 @@
...
 import { DOCUMENT } from '@angular/common';
+import { AuthQuery } from '@blockframes/auth/+state';
 
 declare const jwplayer: any;
 
 @Component({
   selector: 'event-player',
   templateUrl: './player.component.html',
   styleUrls: ['./player.component.scss'],
+  encapsulation: ViewEncapsulation.None,
```
GrandSchtroumpf, line 13: Add comment to explain why

apps/backend-functions/src/internals/watermark.ts
pldespaigne: email is mandatory in `PublicUser` where `firstName` & `lastName` are not

apps/backend-functions/src/internals/resize-images.ts
pldespaigne: other `contentType` are checked by the if just above

apps/backend-functions/src/templates/watermark.ts
pldespaigne: Oh sh*t you are right it's global scope

apps/backend-functions/src/users.ts
pldespaigne: not a big fan of this, it's not easily understandable, and moreover if the `firstName` change but not the `avatar` we will not update algolia

GrandSchtroumpf: Ah sorry I wrote too quickly : 
```typescript
const changed = (keys: string[]) => keys.some(key => before[key] !== after[key]);
const nameChanged = changed['firstName', 'lastName', 'email']
if (nameChanged || before.avatar?.url !== after.avatar?.url) {
  promises.push(storeSearchableUser(after));
}
if (nameChanged) {
  promises.push(upsertWatermark(after));
}
```

PR #2712, opened 2020-05-06
debug on org creation

phpgeekfr:


libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -76,10 +76,10 @@ export class OrganizationService extends CollectionService<OrganizationState> {
       roles: { [user.uid]: 'superAdmin' }
     });
 
-    // Update user with orgId.
-    const userDoc = this.db.doc(`users/${user.uid}`);
-    write.update(userDoc.ref, { orgId });
-    return this.permissionsService.add({ id: orgId, ...permissions }, { write });
+    return Promise.all([
+      this.permissionsService.add(permissions, { write }),
+      this.userService.update(user.uid, { orgId, ...user }, { write }),
```
GrandSchtroumpf, line 81: If you just update the orgId you shouldn't need to pass the whole user: 
```typescript
this.userService.update(user.uid, { orgId }, { write }),
```

libs/organization/src/lib/permissions/guard/permissions.guard.ts
```diff
@@ -20,7 +20,11 @@ export class PermissionsGuard extends CollectionGuard<PermissionsState> {
 
   sync() {
     return this.authQuery.user$.pipe(
-      switchMap(user => {
+      // @TODO(#2710) This async is needed to prevent permission denied
+      // in the case where user.orgId exists but not the permissions document.
```
GrandSchtroumpf, line 24: By adding a reference to an issue I expect the description to be inside the issue, not in the code XD

PR #2720, opened 2020-05-06
Implements meeting + fix bugs

GrandSchtroumpf:
Fix bunch of bugs from maketplace
Add meeting creation
Improve organization page
List only organizations from the app
Remove notification link
Add


libs/notification/src/lib/invitation/guard/invitations.guard.ts
```diff
@@ -11,21 +11,40 @@ import { InvitationService } from '../+state/invitation.service';
...
+import { PermissionsQuery } from '@blockframes/permissions/+state';
+import { combineLatest } from 'rxjs';
+import { OrganizationQuery } from '@blockframes/organization/+state';
 
 @Injectable({ providedIn: 'root' })
 @CollectionGuardConfig({ awaitSync: false })
 export class InvitationGuard extends CollectionGuard<InvitationState> {
-  constructor(service: InvitationService, private authQuery: AuthQuery) {
+  constructor(
+    service: InvitationService,
```
phpgeekfr, line 22: missing private. not a big deal

GrandSchtroumpf: It's not missing it's used in the super

Issue #2719, opened 2020-05-06
Set the correct domain when sending emails


phpgeekfr:
Currently we use the same domain for links in every mail we send.
But in fact there can be two domains and the user should be redirected to the good one when he clicks on a link of an email.

- [ ] Identify the mails that uses links and check which domain it should use.
- [ ] If user to who we are sending the mail have an orgId, we can guess the domain using org.appAcess

Jackseed:
We should also send different welcome emails depending on the app where the user created an account.
For now we have one for Archipel Content only ID: d-eb8e1eb7c5a24eb8af1d2d32539ad714

Jackseed:
Emails where we should change the url according to the org appAccess:
- [ ] Request to join your org received Template ID: d-b1ab5d21def145ccb759520e2d984436
- [ ] Organization Created Template ID: d-8c5f7009cd2f4f1b877fa168b4efde48
// Looks like url was hardcoded, I removed the link for now
- [ ] Join an org request - Accepted Template ID: d-d32b25a504874a708de6bfc50a1acba7

Emails with different ids depending on the app:
- [ ] Welcome email
- Archipel Content: Template ID: d-eb8e1eb7c5a24eb8af1d2d32539ad714 
- Archipel Market: tbc
[ ] Invited user credentials
- Archipel Content: Template ID: d-a34ce9ea59c5477f9feae8f556157b6b 
- Archipel Market: tbc

PR #2726, opened 2020-05-07
#2719 release 1.3

phpgeekfr:


env/env.ci.ts
```diff
@@ -17,7 +17,8 @@ export const sentryDsn = undefined;
 // Firebase
 // ========
 
-export const appUrl = 'https://blockframes-ci.web.app';
+export const appUrlContent = 'hhttps://blockframes-ci.web.app';
```
lsenta, line 20: typo

apps/backend-functions/src/data/internals.ts
```diff
@@ -9,6 +9,8 @@ import { PermissionsDocument } from '@blockframes/permissions/types';
 import { ContractDocument } from '@blockframes/contract/contract/+state/contract.firestore';
 import { createImgRef } from '@blockframes/utils/image-uploader';
 import { createDenomination } from '@blockframes/organization/+state/organization.firestore';
+import { App, getOrgAppAccess } from '@blockframes/utils/apps';
```
lsenta, line 12: So far we tried avoiding importing libraries & code that depends on Akita and the frontend libs.

apps/backend-functions/src/data/internals.ts
```diff
@@ -91,3 +93,21 @@ export async function getAdminIds(organizationId: string): Promise<string[]> {
   return adminIds;
 }
 
+/**
+ * Return the first app name that an org have access to
+ * @param _org 
+ */
+export async function getOrgAppName(_org: OrganizationDocument | string): Promise<App> {
+  if (typeof _org === 'string') {
+    const org = await getDocument<OrganizationDocument>(`orgs/${_org}`);
```
pldespaigne, line 102: what happen if `_org` is an org id that doesn't exists ? like if `org` become `undefined`
maybe you should check for this case

firestore.rules
```diff
@@ -383,7 +383,7 @@ service cloud.firestore {
     }
 
     function getDocumentPermissions(orgId, docId) {
-    	return get(/databases/$(database)/documents/permissions/$(orgId)/documentPermissions/$(docId)).data;
+    	return getAfter(/databases/$(database)/documents/permissions/$(orgId)/documentPermissions/$(docId)).data;
```
lsenta, line 386: Looks dangerous: if a user transact / batch to a document and its document permission, they may elevate their authorizations (getAfter looks at what's in the DB and what's written at the same time).

apps/backend-functions/src/orgs.ts
```diff
@@ -149,10 +151,11 @@ export async function onOrganizationUpdate(
   const blockchainBecomeEnabled = before.isBlockchainEnabled === false && after.isBlockchainEnabled === true;
 
   const { id, userIds } = before as OrganizationDocument;
-  const admin = await db.collection('users').doc(userIds[0]).get().then(adminSnapShot => adminSnapShot.data()! as PublicUser); // TODO use laurent's code after the merge of PR #698
+  const admin = await getDocument<PublicUser>(`users/${userIds[0]}`);
```
pldespaigne, line 154: same here, what if `admin` is undefined

libs/movie/src/lib/movie/components/movie-analytics-chart/movie-analytics-chart.component.ts
```diff
@@ -45,7 +45,7 @@ function getLastDays(from: number, to: number = 0) {
 function toYMD(date: Date) {
   const m = date.getMonth()
   const d = date.getDate()
-  return `${date.getFullYear()}${m < 10 ? `0${m+1}` : m+1}${d < 10 ? `0${d}` : d}`;
+  return `${date.getFullYear()}${m < 10 ? `0${m + 1}` : m + 1}${d < 10 ? `0${d}` : d}`;
```
pldespaigne, line 48: it's up to you but I find it more readable to use `padStart()` string method

firestore.rules
phpgeekfr: I agree but since documentPermissions documents can be created only if the document related to this permissions is brand new and can be updated only by org admin.

libs/movie/src/lib/movie/components/movie-analytics-chart/movie-analytics-chart.component.ts
phpgeekfr: Not my code, not touching it ^^

apps/backend-functions/src/data/internals.ts
phpgeekfr: You mean that I need to move those functions ?

lsenta: Yea, if this is shared code keep it into the .firestore.ts that relates to orgs

PR #2738, opened 2020-05-08
With emulator

laurentsenta:

### DIFF
--- a/apps/catalog/catalog/src/app/app.component.ts
+++ b/apps/catalog/catalog/src/app/app.component.ts
@@ -3,6 +3,8 @@ import { ChangeDetectionStrategy, Component } from '@angular/core';
 import { ThemeService } from '@blockframes/ui/theme';
 import { IconService } from '@blockframes/ui/icon-service';
 import { TunnelService } from '@blockframes/ui/tunnel';
+import firebase from 'firebase';
+
 
 @Component({
   selector: 'catalog-root',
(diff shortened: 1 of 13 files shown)


apps/catalog/catalog/src/app/app.component.ts
```diff
@@ -3,6 +3,8 @@ import { ChangeDetectionStrategy, Component } from '@angular/core';
 import { ThemeService } from '@blockframes/ui/theme';
 import { IconService } from '@blockframes/ui/icon-service';
 import { TunnelService } from '@blockframes/ui/tunnel';
+import firebase from 'firebase';
```
GrandSchtroumpf, line 6: The error in the console says that firebase SDK should be imported like that 
```typescript
import * as firebase from 'firebase/app';
import 'firebase/<PACKAGE>';
```

apps/catalog/catalog/src/app/app.module.ts
```diff
@@ -73,24 +78,37 @@ import { ErrorLoggerModule } from '@blockframes/utils/error-logger.module';
...
+      }
+    )
   ],
   providers: [ScreenTrackingService, UserTrackingService],
   bootstrap: [AppComponent]
 })
 export class AppModule {
-
   constructor(private router: Router, private analytics: FireAnalytics) {
+    if (window.location.hostname === 'localhost') {
```
GrandSchtroumpf, line 103: Add "?" after window in `window?.location` because the app can be rendered in a server (SSR/prerendering) where window doesn't exist.

apps/catalog/catalog/src/app/app.module.ts
```diff
@@ -73,24 +78,37 @@ import { ErrorLoggerModule } from '@blockframes/utils/error-logger.module';
...
+      console.log('testing locally -- hitting local functions and firestore emulators');
+      theFirebase.functions().useFunctionsEmulator('http://localhost:5001');
+      theFirebase.firestore().settings({
```
GrandSchtroumpf, line 106: There is this issue thread about the emulator : 
https://github.com/angular/angularfire/issues/2183

According to the last comment you should be able to add host & ssl directly in the env config.

fritzschoff:
looks to me stale

PR #2740, opened 2020-05-11
festival: enable prod sourcemaps

lsenta:

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -765,7 +765,7 @@
               ],
               "optimization": true,
               "outputHashing": "all",
-              "sourceMap": false,
+              "sourceMap": true,
               "extractCss": true,
               "namedChunks": false,
               "extractLicenses": true,


lsenta:
@GrandSchtroumpf 

diff configuration in prod:

```
› git diff                                                                            ☺
diff --git a/angular.json b/angular.json
index d21e8afd1..663ee8e4b 100644
--- a/angular.json
+++ b/angular.json
@@ -765,9 +765,10 @@
               ],
               "optimization": true,
               "outputHashing": "all",
-              "sourceMap": false,
+              "sourceMap": true,
               "extractCss": true,
               "namedChunks": false,
+              "aot": true,
               "extractLicenses": true,
               "vendorChunk": false,
               "buildOptimizer": true,
@@ -782,7 +783,9 @@
                   "maximumWarning": "6kb",
                   "maximumError": "10kb"
                 }
-              ]
+              ],
+              "serviceWorker": true,
+              "ngswConfigPath": "apps/catalog/catalog/ngsw-config.json"
             },
             "ci": {
               "fileReplacements": [
```

I'm just enabling sourcemaps for the release but we might be missing something else.

GrandSchtroumpf:
LGTM. Is sourcemap already enabled on festival app ?
@Jackseed  Maybe we should check the icons for the service worker, I don't recall we changed them.

lsenta:
> LGTM. Is sourcemap already enabled on festival app ?

No, that's the point of this PR. The lines with a `+` are things that where not ported from the catalog build config to festival.

PR #2750, opened 2020-05-12
fix algolia movies index

pldespaigne:
- [x] add `name` field as searchable attribute to the Org index's config
- [x] fix bug about `_meta.createdBy` in the Movie index


apps/backend-ops/src/algolia.ts
```diff
@@ -69,13 +71,21 @@ export async function upgradeAlgoliaMovies() {
...
+      if (querySnap.size === 0) {
+        throw new Error(`Movie ${movie.id} is not part of any orgs`);
+      }
+
+      // TODO : here we might decide to arbitrary choose first org
+      if (querySnap.size > 1) {
+        throw new Error(`Movie ${movie.id} is part of several orgs (${querySnap.docs.map(doc => doc.id).join(', ')})`);
+      }
+
+      const orgName = (querySnap.docs[0].data() as OrganizationDocument).denomination.full;
```
phpgeekfr, line 86: I Think you want to add public data on algolia ? If so, you can do something like this : 
```
export class OrgNamePipe implements PipeTransform {
  transform(value: Organization) {
    return value.denomination.public || value.denomination.full;
  }
}
````

Comment (lsenta) [reply apps/backend-ops/src/algolia.ts:87]: Get that angular stack away from the ops! One framework and a class for an or operator?! 

```
const {denomination}= (querySnap.docs[0].data() as OrganizationDocument)
const orgName = denomination.public || denomination.full
```

PR #2767, opened 2020-05-13
Event view

GrandSchtroumpf:
Create an event view page
Improve queries

### DIFF
--- a/libs/event/src/lib/layout/edit/edit.component.html
+++ b/libs/event/src/lib/layout/edit/edit.component.html
@@ -65,7 +65,7 @@
         </button>
       </div>
 
-      <p>Here the list of invited persons</p>
+      <!-- @todo(#2763) Add loading indication here -->
       <invitation-guest-list [invitations]="invitations"></invitation-guest-list>
     </mat-card>
   </section>
--- a/libs/event/src/lib/layout/edit/edit.component.html
+++ b/libs/event/src/lib/layout/edit/edit.component.html
@@ -42,7 +42,7 @@
     <ng-content></ng-content>
 
 
-    <!-- <h2>Invitations</h2>
+    <h2>Invitations</h2>
     <mat-card class="invitations" fxLayout="column"  fxFlex="35">
       <div fxLayout fxLayout="space-between center" fxLayoutGap="16px">
         <algolia-chips-autocomplete fxFlex
@@ -65,9 +65,9 @@
         </button>
       </div>
 
-      Here the list of invited persons
+      <p>Here the list of invited persons</p>
       <invitation-guest-list [invitations]="invitations"></invitation-guest-list>
-    </mat-card> -->
+    </mat-card>
   </section>
 </form>
 
--- a/libs/event/src/lib/layout/view/view.component.html
+++ b/libs/event/src/lib/layout/view/view.component.html
@@ -1,5 +1,5 @@
 <ng-container *ngIf="event$ | async as event">
-  <header fxLayout="column" fxLayoutAlign="center center">
+  <header class="dark-contrast-theme" fxLayout="column" fxLayoutAlign="center center">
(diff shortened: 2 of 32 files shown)


libs/notification/src/lib/invitation/+state/invitation.service.ts
```diff
@@ -36,6 +38,27 @@ export class InvitationService extends CollectionService<InvitationState> {
...
 
+  /////////////
+  // QUERIES //
+  /////////////
+  /**
+   * Query all invitation on a docId created by from an orgId (Should be updated if )
+   * @param docId The id liked to the invitation
+   * @param orgId The organization owner of the docId
+   */
+  public queryGuest(docId: string, orgId: string) {
```
phpgeekfr, line 49: Query must reflect rules so:
If you query only invitations related to events: .where(type == 'attendEvent')  should be enough (and .where('docId', '==', docId can be optional) because rule is : `existingData().type == 'attendEvent' && isSignedIn()`

If you query invitation related to joinOrg   : .where(type == 'joinOrganization') and .where('docId', '==', orgId) should be enough because rule is :  `( existingData().type == 'joinOrganization' && (isGuest() || isHostAdmin()) )`

libs/notification/src/lib/invitation/+state/invitation.service.ts
```diff
@@ -36,6 +38,27 @@ export class InvitationService extends CollectionService<InvitationState> {
     return cleanInvitation(invitation);
   }
 
+  /////////////
+  // QUERIES //
+  /////////////
+  /**
+   * Query all invitation on a docId created by from an orgId (Should be updated if )
+   * @param docId The id liked to the invitation
+   * @param orgId The organization owner of the docId
```
phpgeekfr, line 47: Typos

libs/event/src/lib/layout/edit/edit.component.html
```diff
@@ -65,9 +65,9 @@ <h2>Details</h2>
         </button>
       </div>
 
-      Here the list of invited persons
+      <p>Here the list of invited persons</p>
```
phpgeekfr, line 68: Typo ?

libs/event/src/lib/layout/view/view.component.html
```diff
@@ -1,31 +1,17 @@
 <ng-container *ngIf="event$ | async as event">
-  <header [backgroundAsset]="">
-    <div fxLayout="column" fxLayoutAlign="space-around center" fxLayoutGap="8px">
-      <h1>{{ event.title }}</h1>
-      <p>{{ event | eventRange }}</p>
-    </div>
+  <header class="dark-contrast-theme" fxLayout="column" fxLayoutAlign="center center">
+    <h1 class="mat-display-2">{{ event.title }}</h1>
+    <pre class="mat-body-1">{{ event | eventRange }}</pre>
     <ng-container *ngIf="event.isPrivate; else isPublic">
```
phpgeekfr, line 5: this container is useless

libs/notification/src/lib/invitation/guard/invitations.guard.ts
```diff
@@ -38,11 +38,18 @@ export class InvitationGuard extends CollectionGuard<InvitationState> {
...
+            // Invitation to the user & Request to the user
+            this.service.syncCollection(ref => ref.where('toUser.uid', '==', user.uid)),
+            this.service.syncCollection(ref => ref.where('fromUser.uid', '==', user.uid))
           ])
         } else {
-          return this.service.syncCollection(ref => ref.where('toUser.uid', '==', user.uid))
+          return combineLatest([
+            this.service.syncCollection(ref => ref.where('toUser.uid', '==', user.uid)),
+            this.service.syncCollection(ref => ref.where('fromUser.uid', '==', user.uid))
+          ])
```
phpgeekfr, line 52: This should work even with new rules, I will try that.

PR #2769, opened 2020-05-13
Review card update

### LINKED ISSUE #2752: add creationDate and Author to MovieReview
We need that information for the movie review card

Mussuranae:
Fix #2752 
Add critic name and publication date on movie review card


libs/ui/src/lib/review-card/review-card.component.scss
```diff
@@ -1,3 +1,9 @@
 :host {
   display: block;
 }
+
+mat-card {
+  width: 330px;
+  min-height: 340px;
+  max-height: 474px;
+}
```
fritzschoff, line 9: where does the hard coded value come from? Did you check it on mobile?

libs/ui/src/lib/review-card/review-card.component.ts
```diff
@@ -5,8 +5,9 @@ function createMovieReviewView(review: Partial<MovieReview>) {
   return {
     journalName: review.journalName || 'Journale name couldn\'t be loaded',
     criticQuote: review.criticQuote || 'Quote couldn\'t be loaded',
-    link: review.revueLink
-    // Wait for TODO #2752 add author and date, and maybe img
+    link: review.revueLink,
+    criticName: review.criticName,
+    publicationDate: review.publicationDate
```
fritzschoff, line 10: if the data model is optional then provide fallback value.
`publicationDate: review.publicationDate ||  'Date unkown'`

libs/ui/src/lib/review-card/review-card.component.scss
Mussuranae: From the mockup of Mathilde, but I didn't check in mobile view

libs/ui/src/lib/review-card/review-card.component.ts
Mussuranae: I put an if in the html, if we have no information from the DB, it won't be display, isn't it better than unknow information ?

libs/ui/src/lib/review-card/review-card.component.scss
fritzschoff: remove it, you set the dimensions of the component in the parent page. This component shouldn't care how big it is

libs/ui/src/lib/review-card/review-card.component.ts
fritzschoff: the ng if is okay but then do 
```
publicationDate: review?.publicationDate || ''
```
otherwise you get an error `cannot read publicationDate of undefined`. And the empty string will result in false if you put it in the ngIf and then you wont render any text.

maybe you can set the `?` after the publicationDate as well, try it out, not sure

Mussuranae: Ah ok, I'll add it so.

PR #2771, opened 2020-05-13
Bf carousel

fritzschoff:


libs/ui/src/lib/carousel/carousel.component.html
```diff
@@ -0,0 +1,5 @@
+<div class="row" cdkScrollable>
+  <ng-content></ng-content>
+</div>
+<mat-icon mat-icon-button svgIcon="arrow_back" (click)="scrollTo('left')" class="previous" *ngIf="showBackwardBtn"></mat-icon>
```
GrandSchtroumpf, line 4: Let's keep a separation of concern & improve a11y by using the right tag. 
```html
<!-- Only show if there is something on the left
<ng-container *ngIf="!beginning">
  <button mat-icon-button (click)="scrollTo('left')" class="previous">
    <mat-icon svgIcon="arrow_back"></mat-icon>
  </button>
</ng-container>
```

libs/ui/src/lib/carousel/carousel.component.scss
```diff
@@ -0,0 +1,29 @@
...
+  overflow: hidden;
+  scroll-behavior: smooth;
+}
+
+mat-icon {
+  position: absolute;
+  outline: none;
+  cursor: pointer;
+  border-radius: 50%;
+  background-color: rgba($color: #928f8f, $alpha: 0.3);
```
GrandSchtroumpf, line 18: I think you're rewriting the more of the style of 'button [mat-icon-button]' here. Remove, update your tag as explained about, and add only what you need (I think only position: absolute)

GrandSchtroumpf: Also, use css synthax when possible: `rgba(#928f8f, 0.3);`

libs/ui/src/lib/carousel/carousel.component.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.previous {
+  left: 0px;
+  bottom: 50%;
```
GrandSchtroumpf, line 23: I don't think it's going to be centered. You'll need to remove half of the height of the mat-icon-button

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,60 @@
...
+  selector: 'bf-carousel',
+  templateUrl: 'carousel.component.html',
+  styleUrls: ['./carousel.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class CarouselComponent implements AfterViewInit {
+
+  public currentPosition: number;
+
+  public showBackwardBtn = false;
```
GrandSchtroumpf, line 23: I would recommand : 
```typescript
beginning = true;
end = false;
```

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(private layout: Layout) { }
+
+  ngAfterViewInit() {
+    this.currentPosition = this.scrollable.measureScrollOffset('left');
+  }
+
+  scrollTo(direction: 'left' | 'right') {
+    /*  If right === 0 it means that we are at the end and want to start from the beginning again. */
+    if (!this.scrollable.measureScrollOffset('right')) {
```
GrandSchtroumpf, line 35: I don't think so ^^. For me if you're at the end you just movie from one set of card, not going back to the beginning.
Maybe @2amarkina has an opinion on that.

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,60 @@
...
+      ? this.showBackwardBtn = true
+      : this.showBackwardBtn = false
+
+    direction === 'right'
+      ? this.scrollable.scrollTo({ left: this.currentPosition + this.layout.width.getValue() })
+      : this.scrollable.scrollTo({ left: this.currentPosition - this.layout.width.getValue() })
+  }
+}
+
+@Directive({ selector: '[shrink]' })
```
GrandSchtroumpf, line 57: call it `carousel-item` instead.

libs/ui/src/lib/carousel/carousel.component.ts
fritzschoff: I talked with @2amarkina and we agreed on 
```
yes so practically like netflix. beginning:  just the forward button, after on click the backward button appears but only until you reach the "start" again. If you hit the end of the slide, you go back to the start again
```

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,60 @@
...
+    */
+    this.currentPosition === 0
+      ? this.showBackwardBtn = true
+      : this.showBackwardBtn = false
+
+    direction === 'right'
+      ? this.scrollable.scrollTo({ left: this.currentPosition + this.layout.width.getValue() })
+      : this.scrollable.scrollTo({ left: this.currentPosition - this.layout.width.getValue() })
+  }
+}
```
GrandSchtroumpf, line 55: I would recommend you to do : 
```typescript
if (direction === 'right') {
  if (this.scrollable.measureScrollOffset('left') === 0) {
    this.beginning = false;
  }
  this.scrollable.scrollTo({ left: this.currentPosition + this.width });
  if (this.scrollable.measureScrollOffset('right') === 0) {
    this.end = true;
  }
} else {
  ...same for left
}
```

libs/ui/src/lib/carousel/carousel.component.ts
fritzschoff: I would stick to the old naming, cause you are not going to the beginning or the end, you jsut go to the next section

libs/ui/src/lib/carousel/carousel.component.ts
fritzschoff: this would disable the back button for instance when you scroll to the right, and we dont want that

fritzschoff: But I guess after our small netflix detour, we have sorted it out

GrandSchtroumpf: No if you have : 
```html
<ng-container *ngIf="!beginning">
  <!-- back button -->
</ng-container>
```
And 
```typescript
if (this.scrollable.measureScrollOffset('left') === 0) {
  this.beginning = false;
}
```
It will work as expected.

GrandSchtroumpf: Btw if scrollTo is a promise you wanna await to be sure to have the right measurementoffset

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  templateUrl: 'carousel.component.html',
+  styleUrls: ['./carousel.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class CarouselComponent {
+
+  public currentPosition: number;
+
+  // Indicators to show arrow buttons
+  public showBack: Observable<boolean>;
```
GrandSchtroumpf, line 27: Put that with the comment below

PR #2774, opened 2020-05-15
WIP test emails spam score

phpgeekfr:
added an interface to send test email from admin panel


apps/backend-functions/src/internals/email.ts
```diff
@@ -1,54 +1,65 @@
...
   };
 
   return SendGrid.send(msg);
 }
+
+export const sendTestMail = async (
+  data: { request: EmailRequest, from: string },
+  context: CallableContext
+): Promise<ErrorResultResponse> => {
+  if (!context || !context.auth) { throw new Error('Permission denied'); }
```
GrandSchtroumpf, line 52: Maybe be a little bit more specific. Just to avoid thinking that it comes from firestore.

apps/backend-functions/src/internals/email.ts
```diff
@@ -1,54 +1,65 @@
...
     to,
-    from: 'admin@blockframes.io',
     templateId,
     dynamicTemplateData: data,
   };
 
   return SendGrid.send(msg);
 }
+
+export const sendTestMail = async (
```
GrandSchtroumpf, line 48: I'm not sure to understand what the goal of this method. Can you add a comment ?

libs/admin/src/lib/admin/admin.service..ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from '@angular/core';
+import { AngularFireFunctions } from '@angular/fire/functions';
+import { EmailRequest } from '@blockframes/utils/emails';
+import { ErrorResultResponse } from '@blockframes/utils/utils';
+
+@Injectable({ providedIn: 'root' })
+export class AdminService  {
```
GrandSchtroumpf, line 7: Do you really need this service ? Looks like you don't need to share this logic with many components.

apps/backend-functions/src/internals/email.ts
phpgeekfr: You have been traumatized too ;)

libs/admin/src/lib/admin/admin.service..ts
phpgeekfr: As you want

PR #2775, opened 2020-05-15
Session page

### LINKED ISSUE #2707: Festival - Session page
## UI 
[Screening - Before ](https://projects.invisionapp.com/d/main?origin=v7#/console/19607961/409972823/preview?scrollOffset=1065.5555419921875)
[Screening - Now](https://projects.invisionapp.com/d/main?origin=v7#/console/19607961/409964813/preview?scrollOffset=1065.5555419921875)
Meeting - to come

## Description
The session page has 2 states depending on the time of the event:
1. Before the event
2. During the event
We use same template for meeting & screening.
This page will be accessible through:
- The event page when click on the CTA button
- The calendar when click on the event if it's live. 

## Acceptance criteria
- [ ] Session page - Screening - Before
- [ ] Session page - Screening - Now
- [ ] Session page - Screening - After
- [ ] Session page - Meeting - Before
- [ ] Session page - Meeting - Now
- [ ] Session page - Meeting - After

Mussuranae:
Session page for festival app
Fix #2707


apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -1,7 +1,44 @@
 :host {
+  height: 100%;
   display: flex;
-  flex-direction: column;
   align-items: center;
   justify-content: center;
-  height: 100%;
+}
+
+.uppercase {
```
GrandSchtroumpf, line 8: Do "h4" directly here. I find it better that giving a class.

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -1,7 +1,44 @@
 :host {
+  height: 100%;
   display: flex;
```
GrandSchtroumpf, line 3: I think you can remove this style

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -1,7 +1,44 @@
...
+  text-transform: uppercase;
+}
+
+h2, h4, h6 {
+  margin: 0;
+}
+
+section {
+  width: 1392px;
+  height: 687px;
```
GrandSchtroumpf, line 18: Never set the width & height in px for a section. You're going to break the responsive. This section should take all the page here: "height: 100%"

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -1,7 +1,44 @@
...
+  margin-top: 70px;
+  background-color: var(--background-slider);
+
+  .screen {
+    margin-top: 24px;
+    width: 860px;
```
GrandSchtroumpf, line 24: either we use a image tag to keep the ratio or we set the width/height for each screen size but we cannot set the width/height fix like that.

Mussuranae: I added size for every screen size.

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
Mussuranae: If I remove the `display: flex`, it's not in the middle of the page :/

apps/festival/festival/src/app/marketplace/event/session/session.component.html
```diff
@@ -1,8 +1,53 @@
...
+            <button mat-flat-button color="accent" (click)="playVideo()">
+              <mat-icon svgIcon="play"></mat-icon>
+              <span>Play</span>
+            </button>
+          </div>
+        </article>
+      </ng-container>
+
+      <ng-template #showVideo>
+        <event-player></event-player>
```
GrandSchtroumpf, line 23: There is no input on this component ?

apps/festival/festival/src/app/marketplace/event/session/session.component.html
```diff
@@ -1,8 +1,53 @@
...
+      </ng-template>
+
+      <img asset="screening_now.webp" alt="Screening now picture">
```
GrandSchtroumpf, line 26: In `alt` don't say "picture", the screen reader already know that. 
Also `screening now` doesn't describe what it is.

apps/festival/festival/src/app/marketplace/event/session/session.component.html
```diff
@@ -1,8 +1,53 @@
...
+          <span class="mat-display-2">{{ event.title }}</span>
+          <span class="mat-caption">Meeting now</span>
+          <a mat-flat-button color="accent" [href]="event.meta.callUrl">
+            <mat-icon svgIcon="play"></mat-icon>
+            <span>Join the meeting</span>
+          </a>
+        </div>
+      </article>
+
+      <img asset="screening_now.webp" alt="Screening now picture">
```
GrandSchtroumpf, line 44: Is there a specific image for the meetings ? Also the `alt` is not good, this is not about screening. It might confuse someone using a screen reader.

apps/festival/festival/src/app/marketplace/event/session/session.component.html
```diff
@@ -1,8 +1,53 @@
...
+    </section>
+
+    <!-- MEETING TYPE -->
+    <section *ngSwitchCase="'meeting'" fxLayout="column" fxLayoutAlign="space-between center">
+
+      <article class="dark-contrast-theme screen" [style.background-image]="background">
+        <div class="overlay" fxLayout="column" fxLayoutAlign="center center">
+          <span class="mat-display-2">{{ event.title }}</span>
+          <span class="mat-caption">Meeting now</span>
+          <a mat-flat-button color="accent" [href]="event.meta.callUrl">
```
GrandSchtroumpf, line 37: add attribute `target="_blank"` to be sure it's open a new tab.

apps/festival/festival/src/app/marketplace/event/session/session.component.html
```diff
@@ -1,8 +1,53 @@
...
+      </ng-template>
+
+      <img asset="screening_now.webp" alt="Screening now picture">
+
+    </section>
+
+    <!-- MEETING TYPE -->
+    <section *ngSwitchCase="'meeting'" fxLayout="column" fxLayoutAlign="space-between center">
+
+      <article class="dark-contrast-theme screen" [style.background-image]="background">
```
GrandSchtroumpf, line 33: Why not using `[backgroundRef]` ?

apps/festival/festival/src/app/marketplace/event/session/session.component.ts
```diff
@@ -1,19 +1,37 @@
...
+    private route: ActivatedRoute,
+  ) { }
 
   ngOnInit(): void {
-    this.url = this.sanitizer.bypassSecurityTrustResourceUrl('https://player.vimeo.com/video/391939808');
+    this.event$ = this.route.params.pipe(
+      pluck('eventId'),
+      switchMap(eventId => this.service.queryDocs(eventId))
+    );
+    this.background = `url(https://images.pexels.com/photos/3183150/pexels-photo-3183150.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=750&w=1260)`;
```
GrandSchtroumpf, line 30: Do we have the right on this image ? Do we have to quote the author ? If yes where should be do that ?
I think it would be better to not use an external image.

apps/festival/festival/src/app/marketplace/event/session/session.component.ts
```diff
@@ -1,19 +1,37 @@
...
   templateUrl: './session.component.html',
-  styleUrls: ['./session.component.scss']
+  styleUrls: ['./session.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class SessionComponent implements OnInit {
 
-  url: SafeResourceUrl;
+  public background: string;
+  event$: Observable<Event>;
```
GrandSchtroumpf, line 17: keep the same structure. If you use `public`, put it everywhere.

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -1,7 +1,105 @@
...
+      max-width: 100%;
+    }
+    event-player {
+      width: 80%;
+      height: 80%;
+    }
+  }
+}
+
+@include layout-bp(sm) {
```
GrandSchtroumpf, line 70: I think you can use `lt-sm` for everything that is `lower-than sm` so you don't have to repeat yourself
```scss
@include layout-bp(lt-sm)
```

You have the same with `gt-**` for `greater than **`

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
GrandSchtroumpf: For the screen we can set the exact value. But I still think you shouldn't set the size like that for the section. Use %.

apps/festival/festival/src/app/marketplace/event/session/session.component.ts
```diff
@@ -1,19 +1,37 @@
...
+    this.event$ = this.route.params.pipe(
+      pluck('eventId'),
+      switchMap(eventId => this.service.queryDocs(eventId))
+    );
+    this.background = `url(https://images.pexels.com/photos/3183150/pexels-photo-3183150.jpeg?auto=compress&cs=tinysrgb&dpr=2&h=750&w=1260)`;
   }
 
+  playVideo() {
+    this.showSession = false;
+    return;
```
GrandSchtroumpf, line 35: You don't need to return anything.

apps/festival/festival/src/app/marketplace/event/session/session.component.html
```diff
@@ -1,8 +1,53 @@
...
+
+      <ng-container *ngIf="showSession else showVideo">
+        <article class="dark-contrast-theme screen" [backgroundRef]="event.movie.promotionalElements.banner.media">
+          <div class="overlay" fxLayout="column" fxLayoutAlign="center center">
+            <img [imgRef]="event.org.logo" placeholderAsset="empty_organization.webp" type="logo" alt="organisation logo">
+            <h4>{{ event.org | orgName }} presents</h4>
+            <span class="mat-display-2">{{ event.movie.main.title.international }}</span>
+            <h2>by {{ event.movie.main.directors | displayName }}</h2>
+            <span class="mat-caption">Screening now</span>
+            <button mat-flat-button color="accent" (click)="playVideo()">
```
GrandSchtroumpf, line 14: You can do `(click)="showSession = false"` if the method `playVideo` is only going to do that.

apps/festival/festival/src/app/marketplace/event/session/session.component.html
Mussuranae: Nope.

Mussuranae: It gets the active event itself.

apps/festival/festival/src/app/marketplace/event/session/session.component.html
Mussuranae: Not yet, I wanted to update it with Mathilde's picture when I had it.

apps/festival/festival/src/app/marketplace/event/session/session.component.html
GrandSchtroumpf: Ok, I don't think it's a good design to have this like that. Leave it like that, I'll see wit @pldespaigne.

apps/festival/festival/src/app/marketplace/event/session/session.component.html
Mussuranae: I put a todo inside the code for the one who will be in charge of the page.

apps/festival/festival/src/app/marketplace/event/session/session.component.html
Mussuranae: I hadn't the picture at this moment, and it's one picture for the both theme, so have I to put it in both folder for both theme ?

apps/festival/festival/src/app/marketplace/event/session/session.component.ts
Mussuranae: It was for the example, I hadn't the picture at this moment.

GrandSchtroumpf: Ok, make sure you'll have a picture as soon as possible. We don't want to have that in prod.

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
Mussuranae: I tried it, and the result wasn't good. It is with a `lt-md`.

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
Mussuranae: I did it because of the background-color on the section. It is okay if you don't go in the extreme, like if you have a small height size but big width size (just played with the window).

apps/festival/festival/src/app/marketplace/event/session/session.component.html
```diff
@@ -1,8 +1,55 @@
...
+    </section>
+
+    <!-- MEETING TYPE -->
+    <section *ngSwitchCase="'meeting'" fxLayout="column" fxLayoutAlign="space-between center">
+
+      <article class="dark-contrast-theme screen" [style.background-image]="'url(assets/images/meeting.webp)'">
+        <div class="overlay" fxLayout="column" fxLayoutAlign="center center">
+          <span class="mat-display-2">{{ event.title }}</span>
+          <span class="mat-caption">Meeting now</span>
+          <a mat-flat-button color="accent" [href]="event.meta.callUrl" target="_blank">
```
GrandSchtroumpf, line 39: Maybe check if there is a callUrl. If not, display a message like : 
```html
<p>
Contact <a [mailTo]="event.oganizedBy.email">{{ event.oganizedBy | displayName }}</a> to provide a link for the meeting
</p>
```

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -1,7 +1,65 @@
...
+  background-color: var(--background-slider);
+  text-align: center;
+  img {
+    width: 100%;
+  }
+}
+
+.screen {
+  width: 75%;
+  padding-top: calc(56.25% * 0.75); /* 16:9 Aspect Ratio: 9 / 16 * 50% (screen size) */
```
GrandSchtroumpf, line 29: You might want to change the comment from 50% to 75%

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -1,7 +1,65 @@
...
+      width: 80px;
+      border-radius: 50%;
+    }
+    button, a {
+      margin-top: 24px;
+    }
+  }
+}
+
+@include layout-bp(md) {
```
GrandSchtroumpf, line 54: You don't need this one, it doesn't change the default

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -1,7 +1,65 @@
...
+  .screen {
+    width: 75%;
+    padding-top: calc(56.25% * 0.75); /* 16:9 Aspect Ratio: 9 / 16 * 50% (screen size) */
+  }
+}
+@include layout-bp(lt-md) {
+  .screen {
+    width: 100%;
+    padding-top: calc(56.25% * 1); /* 16:9 Aspect Ratio: 9 / 16 * 50% (screen size) */
```
GrandSchtroumpf, line 63: No need to repeat the comment. But if you do, adapt it with `* 100%`

PR #2778, opened 2020-05-15
Event view

GrandSchtroumpf:
Add organizedBy to the meeting
Bunch of small updates

### DIFF
--- a/libs/event/src/lib/layout/view/view.component.html
+++ b/libs/event/src/lib/layout/view/view.component.html
@@ -7,7 +7,5 @@
     <pre class="mat-body-1">{{ event | eventRange }}</pre>
     <invitation-action [invitation]="invitation$ | async" [event]="event"></invitation-action>
   </header>
-  <section>
-    <ng-content></ng-content>
-  </section>
+  <ng-content></ng-content>
 </ng-container>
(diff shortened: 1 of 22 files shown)


libs/event/src/lib/layout/view/view.component.html
```diff
@@ -1,13 +1,27 @@
...
+        <h6>{{ event.type }} done</h6>
+        <pre class="mat-body-1">{{ event | eventRange }}</pre>
+      </ng-container>
+
+    </ng-container>
   </header>
-  <section>
-    <ng-content></ng-content>
-  </section>
+  <ng-content></ng-content>
```
phpgeekfr, line 26: typo

libs/user/src/lib/user/+state/user.store.ts
```diff
@@ -1,8 +1,8 @@
 import { Injectable } from "@angular/core";
 import { StoreConfig, EntityStore, EntityState } from "@datorama/akita";
-import { OrganizationMember } from "./user.model";
```
phpgeekfr, line 3: can you remove OrganizationMember from the whole app  ?

GrandSchtroumpf: Clélia is going to do it. I think it worth an whole issue.

phpgeekfr: Should I create the issue ?

PR #2786, opened 2020-05-18
added storybook to user lib

fritzschoff:
refactored user component


libs/user/src/lib/user/components/card/card.component.html
```diff
@@ -1,4 +1,6 @@
-<img [imgRef]="user.avatar" placeholderAsset="profil_user.webp" alt="User avatar" />
-<h5>{{ user.firstName }} {{ user.lastName }}</h5>
-<h4>{{ user.position }}</h4>
-<a class="mat-caption" [href]="'mailto:' + user.email">{{ user.email }}</a>
+<section fxLayout="column" fxLayoutAlign="start center">
```
GrandSchtroumpf, line 1: You don't need this tag, you can apply the flexbox on the `:host` instead. Let's not add unnecessary tags.

package.json
```diff
@@ -48,6 +48,7 @@
     "postinstall": "ngcc --properties es2015 browser module main --first-only --create-ivy-entry-points",
     "storybook:ui": "npx ng run ui:storybook",
     "storybook:movie": "npx ng run movie:storybook",
+    "storybook:user": "npx nx run user:storybook",
```
GrandSchtroumpf, line 51: Are you using those script ? If not remove it, I think we only need one example to remember how to run it.

Issue #2586, opened 2020-04-21
Upload heavy files (films) on Firestorage


Jackseed:
- [x] remove the old `hostedVideo` key + new movie video data model + corresponding migration
- [x] Add a way to upload videos **& screening** to a movie through the admin panel
- [x] Upload video should go to the private folder of the corresponding movie
- [x] refactor the `getPrivateVideo` backend function to handle any kind of video and not only the screening
- [x] refactor the `uploadVideoToJWPlaye`r backend function to handle the new file architecture
- [x] Use the backend function created by @pldespaigne to upload the video from firebase storage to JW Player
- [ ] check for #2586 in code
- [x] wait for `jwPlayer` response for api_key error

link to master issue #3639

Jackseed:
UX work to make the user wait @2amarkina

phpgeekfr:
can we close this issue ?

pldespaigne:
We don't have neither movie upload nor storage rules.
But maybe those 2 things are in others issues ?

Jackseed:
This issue is still to be done but it will be integrated with @2amarkina's work on #2832 so we wait for wireframes before implementing the upload

Jackseed:
@phpgeekfr should give the right path to upload the movie on firebase

pldespaigne:
@Jackseed @2amarkina where should I put the upload in the front-end ?

Jackseed:
@pldespaigne it goes on the dashboard movie view with a new tab, here is the corresponding [invision](https://projects.invisionapp.com/d/main?origin=v7#/console/19607961/418106483/preview?scrollOffset=7564)
(+ following pages)
Not sure it has been integrated into Zeplin.. @stephanie-cascade8 do you know ?

PR #2792, opened 2020-05-19
Upload heavy files (films) on Firestorage #2586

pldespaigne:
- [x] add an upload tab to dashboard movie view
- [x] prevent user to navigate away or unload/reload during upload
- [ ] automate upload from fire storage to JWPlayer
- [ ] prevent user to re-upload movie if it already exists


pldespaigne:
This PR is heavily outdated, video upload is part of the road map, and will be done later with the new media architecture

PR #2797, opened 2020-05-19
Debug#2762

ioaNikas:
Festival and Catalog production debug


libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.html
```diff
@@ -0,0 +1,23 @@
+<section *ngIf="hasLink" fxLayout="row" fxLayout.lt-md="column" fxLayoutGap="24px">
+  <wishlist-button [movieId]="movie.id" small></wishlist-button>
+  <ng-container *ngFor="let link of links | slice:0:2">
+    <a [href]="link.url" target="_blank" fxLayout="row" fxLayoutAlign="start center">
+      <mat-icon [svgIcon]="link.icon"></mat-icon>
+      <span>{{ link.label }}</span>
+    </a>
+  </ng-container>
+  <ng-container>
+    <a *ngIf="links.length > 3" [matMenuTriggerFor]="more" fxLayout="row" fxLayoutAlign="start center">
```
GrandSchtroumpf, line 10: Pretty nice 👍 
Maybe show only 2 buttons is not enough for large screen though.

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.ts
```diff
@@ -0,0 +1,43 @@
...
+];
+
+@Component({
+  selector: '[movie] movie-promotional-links',
+  templateUrl: './promotional-links.component.html',
+  styleUrls: ['./promotional-links.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class PromotionalLinksComponent {
+  public promoLinks = promoLinks;
```
GrandSchtroumpf, line 21: This should be an input because we don't want to always display the same things on different apps

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  @Input() movie: Movie;
+
+  get hasLink(): boolean {
+    return this.promoLinks.some(link => !!this.movie.promotionalElements[link].media.url);
+  }
+
+  get links() {
+    return promoLinks.map(link => {
```
GrandSchtroumpf, line 30: Maybe creating a pipe could be useful for this kind of logic.

libs/organization/src/lib/organization/pages/organization-app-access/organization-app-access.component.ts
```diff
@@ -41,7 +43,7 @@ export class OrganizationAppAccessComponent implements OnInit, OnDestroy {
     await this.service.update(this.query.getActiveId(), {
       appAccess: orgAppAccess
     });
-
+    this.snackBar.open(`The organization ${this.query.getActive().denomination.full} has been created`, 'close', { duration: 2000 });
```
GrandSchtroumpf, line 46: It doesn't looks like a good UX for me. If we are redirected to `create-congratulations` we already have a feedback of success.

libs/organization/src/lib/organization/pages/organization-create-feedback/organization-create-feedback.component.html
```diff
@@ -1,10 +1,13 @@
...
+    <p>You’ll receive a confirmation link by mail once your organization is validated.
       <br> Thank you for your patience.  </p>
   </article>
   <article fxLayout="column" fxLayoutAlign="center center">
     <img asset="wait.webp" />
   </article>
-</section>
\ No newline at end of file
+  <ng-container *ngIf="isAccepted$ | async">
+    <a routerLink="/c/o" mat-flat-button color="accent">Go to {{ workspace }}</a>
```
GrandSchtroumpf, line 11: Maybe add a small animation here to make it more enjoyable (if you see how to do it).

libs/organization/src/lib/organization/pages/organization-create-feedback/organization-create-feedback.component.ts
```diff
@@ -1,10 +1,36 @@
...
 @Component({
   selector: 'organization-create-feedback',
   templateUrl: './organization-create-feedback.component.html',
   styleUrls: ['./organization-create-feedback.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class OrganizationCreateFeedbackComponent {
+export class OrganizationCreateFeedbackComponent implements OnInit, OnDestroy {
+  private subscription: Subscription;
+  public workspace = this.query.getActive().appAccess[getCurrentApp(this.routerQuery)].dashboard
```
GrandSchtroumpf, line 16: We should create a pipe for that. Maybe inside utils/pipe

libs/organization/src/lib/organization/pages/organization-create-feedback/organization-create-feedback.component.ts
```diff
@@ -1,10 +1,36 @@
...
+  );
+
+  constructor(
+    private query: OrganizationQuery,
+    private routerQuery: RouterQuery,
+    private service: OrganizationService
+  ) {}
+
+  ngOnInit() {
+    this.subscription = this.service.syncOrgActive().subscribe();
```
GrandSchtroumpf, line 30: I'm not sure what happen here. If you listen only on this page (not inside a guard), I think the best would be : 
```typescript
this.isAccepted$ = this.routerQuery.selectParams('orgId').pipe(
  switchMap(orgId => this.service.valueChanges(orgId)),
  map(organization => organization.status === 'accepted')
);
```
Like that you don't have to unsubscribe

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.html
ioaNikas: Wishlist button (wich is mandatory) was meant to have text and "be the third button". Let's have 3 buttons until we figure out how to make it !

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.html
```diff
@@ -1,18 +1,18 @@
...
-  <ng-container *ngFor="let link of links | slice:0:2">
-    <a [href]="link.url" target="_blank" fxLayout="row" fxLayoutAlign="start center">
-      <mat-icon [svgIcon]="link.icon"></mat-icon>
-      <span>{{ link.label }}</span>
-    </a>
-  </ng-container>
-  <ng-container>
-    <a *ngIf="links.length > 3" [matMenuTriggerFor]="more" fxLayout="row" fxLayoutAlign="start center">
+  <ng-container *ngIf="(promoLinks | promotionalLinks: movie) as links">
+    <ng-container *ngFor="let link of links | slice:0:3">
```
GrandSchtroumpf, line 4: I think it's better to do :
```html
<ng-container *ngFor="let link of links | promotionalLinks: movie | slice:0:3">
``` 
(and update the name of the input to `links`)

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.html
```diff
@@ -1,18 +1,18 @@
...
+      <a [href]="link.url" target="_blank" fxLayout="row" fxLayoutAlign="start center">
```
GrandSchtroumpf, line 5: I really don't think it's a good idea to force the flex layout here as you're overriting the default behavior of the button. Can you explain why you did that ?

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.html
```diff
@@ -1,18 +1,18 @@
 <section *ngIf="hasLink" fxLayout="row" fxLayout.lt-md="column" fxLayoutGap="24px">
   <wishlist-button [movieId]="movie.id" small></wishlist-button>
```
GrandSchtroumpf, line 2: I'm surprise it's `small`. For me `small` is suppose to be the icon without the text

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.html
```diff
@@ -1,18 +1,18 @@
...
+        <span>{{ link.label }}</span>
+      </a>
+    </ng-container>
+    <a *ngIf="links.length > 4" [matMenuTriggerFor]="more" fxLayout="row" fxLayoutAlign="start center">
       <mat-icon svgIcon="more_horiz"></mat-icon>
       <span>More</span>
     </a>
     <mat-menu #more>
-      <ng-container *ngFor="let link of links | slice:3:5">
+      <ng-container *ngFor="let link of links | slice:4:5">
```
GrandSchtroumpf, line 15: You shouldn't set a limit to the slice here. What happen if there are 10 links ?

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.ts
```diff
@@ -18,26 +8,11 @@ const promoLinks = [
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class PromotionalLinksComponent {
-  public promoLinks = promoLinks;
 
   @Input() movie: Movie;
+  @Input() promoLinks: string[];
```
GrandSchtroumpf, line 13: Call it `links`.

libs/organization/src/lib/organization/pages/organization-app-access/organization-app-access.component.ts
```diff
@@ -22,8 +21,7 @@ export class OrganizationAppAccessComponent implements OnInit, OnDestroy {
     private query: OrganizationQuery,
     private router: Router,
     private route: ActivatedRoute,
-    private routerQuery: RouterQuery,
-    private snackBar: MatSnackBar
+    private routerQuery: RouterQuery
```
GrandSchtroumpf, line 24: Are you using it ?

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.html
ioaNikas: Forgot to use mat-button here, my mistake.
I see that when I use mat-button, layout is good, but typo is not what I need (text is white and in meant to be grey'ish)

libs/organization/src/lib/organization/pages/organization-app-access/organization-app-access.component.ts
ioaNikas: Yes, I use it to get the app name. Could I use it directly in the pipe ?

GrandSchtroumpf: No, all good I think

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.html
ioaNikas: Shall we have a third way to display then ? Because without small the button style doesn't match what we have on invision.

libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.html
GrandSchtroumpf: Hmm this should depends on the local theme. If there is a backgorund image it should be white,  else it should be grey. But don't hardcode it here.

PR #2822, opened 2020-05-19
Invit users function

GrandSchtroumpf:
Create a callable function to invit many user by email.

### DIFF
--- a/apps/backend-functions/src/internals/users.ts
+++ b/apps/backend-functions/src/internals/users.ts
@@ -1,0 +1,54 @@
+
+import { App } from '@blockframes/utils/apps';
+import { templateIds } from '@env';
+import { generate as passwordGenerator } from 'generate-password';
+import { OrganizationDocument } from '../data/types';
+import { getOrgAppName, getAppUrl, getDocument } from '../data/internals';
+import { userInvite } from '../templates/mail';
+import { auth } from './firebase';
+import { sendMailFromTemplate } from './email';
+
+
+interface UserProposal {
+  uid: string;
+  email: string;
+}
+
+
+const generatePassword = () =>
+  passwordGenerator({
+    length: 12,
+    numbers: true
+  });
+
+
+/** 
+ * Get user by email & create one if there is no user for this email
+ */
+export const getOrCreateUserByMail = async (data: { email: string, orgId: string }): Promise<UserProposal> => {
+  const { email, orgId } = data;
+
+  try {
+    const user = await auth.getUserByEmail(email);
+    return { uid: user.uid, email };
+  } catch {
+    const password = generatePassword();
+
+    // User does not exists, send them an email.
+    const user = await auth.createUser({
+      email,
+      password,
+      emailVerified: true,
+      disabled: false
+    });
+
+    const org = await getDocument<OrganizationDocument>(`orgs/${orgId}`);
+    const appName: App = await getOrgAppName(org);
+    const urlToUse = await getAppUrl(org);
+    const templateToUse = appName === 'festival' ? templateIds.userCredentialsMarket : templateIds.userCredentialsContent;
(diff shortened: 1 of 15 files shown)


apps/backend-functions/src/internals/users.ts
```diff
@@ -0,0 +1,54 @@
...
+      emailVerified: true,
+      disabled: false
+    });
+
+    const org = await getDocument<OrganizationDocument>(`orgs/${orgId}`);
+    const appName: App = await getOrgAppName(org);
+    const urlToUse = await getAppUrl(org);
+    const templateToUse = appName === 'festival' ? templateIds.userCredentialsMarket : templateIds.userCredentialsContent;
+
+    await sendMailFromTemplate(userInvite(email, password, org.denomination.full, urlToUse, templateToUse));
```
phpgeekfr, line 50: I will add the dynamic "from" here

apps/backend-functions/src/internals/users.ts
```diff
@@ -0,0 +1,54 @@
...
+  passwordGenerator({
+    length: 12,
+    numbers: true
+  });
+
+
+/** 
+ * Get user by email & create one if there is no user for this email
+ */
+export const getOrCreateUserByMail = async (data: { email: string, orgId: string }): Promise<UserProposal> => {
```
phpgeekfr, line 28: this method in auth.service is still called `public async getOrCreateUserByMail(email: string, orgId: string): Promise<User> {`  please remove it

apps/backend-functions/src/internals/watermark.ts
```diff
@@ -9,11 +9,12 @@ import { admin, db, getStorageBucketName } from './firebase';
  */
 export async function upsertWatermark(user: PublicUser) {
 
-  if (!user.firstName || !user.lastName) {
-    throw new Error(`Cannot generate watermark for user ${user.uid} (${user.email}) because 'firstName' and/or 'lastName' is/are not set!`);
+  if (!user.email) {
```
phpgeekfr, line 12: Why did you removed the firstName and LastName ? What append if they are undefined?

phpgeekfr: I think this is ok because you added `export function getWatermark(email: string, firstName: string = '', lastName: string = '') {` et watermark.ts

libs/notification/src/lib/invitation/+state/invitation.firestore.ts
```diff
@@ -45,3 +45,17 @@ export type InvitationStatus = 'accepted' | 'declined' | 'pending';
 export type InvitationType = 'attendEvent' | 'joinOrganization';
 
 export type InvitationMode = 'request' | 'invitation';
+
+
+/** Create an Invitation */
+export function createInvitation(params: Partial<InvitationBase<Date>> = {}): InvitationBase<Date> {
```
phpgeekfr, line 51: By convention in the project, we usually put `createModel` function in the .model.ts file. This lead us to re-create another `createModel` in the cloud functions.  You want to change a bit this convention (I'm Ok with this, this is just informative)?

libs/user/src/lib/user/+state/user.firestore.ts
```diff
@@ -1,4 +1,20 @@
-import { ImgRef } from "@blockframes/utils/image-uploader";
+import { ImgRef, createImgRef } from "@blockframes/utils/image-uploader";
+
+export interface User {
```
phpgeekfr, line 3: UserBase or no need here because we don't have to handle dates and timestamps ?

GrandSchtroumpf: I just copy/paste the current User interface to be able to share it with the backend. I didn't thought further.

PR #2839, opened 2020-05-20
Movie list maintenance

pldespaigne:
- [x] css quick fix

### DIFF
--- a/libs/movie/src/lib/movie/layout/list/title-list.component.scss
+++ b/libs/movie/src/lib/movie/layout/list/title-list.component.scss
@@ -27,7 +27,6 @@ aside {
     width: 380px;
     max-height: calc(100vh - var(--toolbarHeight) - 40px);
     overflow: auto;
-    border-radius: 16px;
   }
 }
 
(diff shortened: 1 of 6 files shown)


libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,28 +1,33 @@
...
-    text-transform: capitalize;
-  }
 }
 
 aside {
   position: sticky;
   top: calc(var(--toolbarHeight) + 20px);
   mat-card {
-    width: 450px;
+    width: 380px;
```
GrandSchtroumpf, line 27: Why here the mat-card is 380px and in the festival it's 30% ?

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,28 +1,33 @@
...
     max-height: calc(100vh - var(--toolbarHeight) - 40px);
     overflow: auto;
+    border-radius: 16px;
```
GrandSchtroumpf, line 30: I'm really really really not found of overriding the default style of the mat-card here. Please remove.

PR #2846, opened 2020-05-22
Cloud functions work with Node 10

GrandSchtroumpf:
Have a specific package.json for cloud functions.
Remove bad dependancies in cloud functions.


firebase.json
```diff
@@ -52,7 +52,6 @@
   ],
   "functions": {
     "predeploy": [
-      "cat ./package.json | jq \"del(.scripts.postinstall)\" > dist/apps/backend-functions/package.json",
```
phpgeekfr, line 55: This is not needed anymore ?

apps/backend-functions/src/orgs.ts
```diff
@@ -10,8 +10,9 @@ import { deleteObject, storeSearchableOrg } from './internals/algolia';
 import { sendMail, sendMailFromTemplate } from './internals/email';
 import { organizationCreated, organizationWasAccepted, organizationRequestedAccessToApp, organizationCanAccessApp } from './templates/mail';
 import { OrganizationDocument, PublicUser, PermissionsDocument } from './data/types';
+import { algolia } from './environments/environment';
+import { mnemonic, relayer } from './environments/environment';
```
phpgeekfr, line 14: you can import the 3 in one line

firebase.json
GrandSchtroumpf: no this put the main package.json in the dist folder. We should only use the local cloud function package.json

PR #2856, opened 2020-05-25
Org view maintenance

### LINKED ISSUE #2804: MAINTENANCE - Org view

Mussuranae:
Review of the org page and the tab contact, line-up, screening and meeting
Fix #2804


apps/backend-functions/src/internals/invitations/events.ts
```diff
@@ -54,7 +54,7 @@ async function onInvitationToAnEventCreate({
      * No need to create a notification because fromOrg and user recipient
      * will already get the invitation displayed on front end.
      */
-    const senderEmail = fromOrg.denomination.public;
+    const senderEmail = fromOrg.denomination.public ? fromOrg.denomination.public : fromOrg.denomination.full;
```
phpgeekfr: François likes more `fromOrg.denomination.public || fromOrg.denomination.full` but this form work also well.

apps/backend-functions/src/internals/invitations/events.ts
```diff
@@ -54,7 +54,7 @@ async function onInvitationToAnEventCreate({
      * No need to create a notification because fromOrg and user recipient
      * will already get the invitation displayed on front end.
      */
-    const senderEmail = fromOrg.denomination.public;
+    const senderName = fromOrg.denomination.public ? fromOrg.denomination.public : fromOrg.denomination.full;
```
phpgeekfr, line 57: Francois likes mokre `fromOrg.denomination.public || fromOrg.denomination.full` but this work also well.

apps/festival/festival/src/app/marketplace/organization/title/title.component.ts
```diff
@@ -15,7 +15,8 @@ import { scaleIn } from '@blockframes/utils/animations/fade';
 export class TitleComponent implements OnInit, OnDestroy {
   @HostBinding('@scaleIn') private animePage;
   private sub: Subscription;
-  public titles$ = this.query.selectAll();
+  // Todo Try to filter movie before sync with the store
```
phpgeekfr, line 18: I'm not sure this is a good idea because in some pages we may display movies that are in draft. Or is this only in catalog?
If this comment is relevant, please add an issue number

Mussuranae: I asked to Vincent because it could be true, but on the marketplace, we will display only movies that are accepted, never draft. So we can let it like that

PR #2858, opened 2020-05-25
Invitations fix

Hugo-Hontans:
- [x] Fix invitation-item and invitation-action
- [x] Load invitations without invitations from the user connected

### DIFF
--- a/apps/backend-functions/src/invitation.ts
+++ b/apps/backend-functions/src/invitation.ts
@@ -111,8 +111,12 @@ export const inviteUsers = (data: UserInvitation, context: CallableContext): Pro
     for (const email of data.emails) {
       getOrCreateUserByMail(email, user.orgId, invitation.type, data.app)
         .then(u => createPublicUser(u))
-        .then(toUser => invitation.toUser = toUser)
-        .then(_ => db.collection('invitations').add(invitation))
+        .then(toUser => {
+          invitation.toUser = toUser;
+          const id = db.collection('invitations').doc().id;
+          invitation.id = id;
+        })
+        .then(_ => db.collection('invitations').doc(invitation.id).set(invitation))
         .then(result => promises.push({ result, error: '' }))
         .catch(error => promises.push({ result: undefined, error }))
         .then(lastIndex => {
--- a/apps/backend-functions/src/invitation.ts
+++ b/apps/backend-functions/src/invitation.ts
@@ -97,7 +97,7 @@ interface UserInvitation {
 
 /**
  * Invite a list of user by email.
- * @dev this function polyfills the Promise.allSettled  
+ * @dev this function polyfills the Promise.allSettled
  */
 export const inviteUsers = (data: UserInvitation, context: CallableContext): Promise<any> => {
   return new Promise(async (res) => {
--- a/apps/festival/festival/src/app/dashboard/invitation/invitation.component.ts
+++ b/apps/festival/festival/src/app/dashboard/invitation/invitation.component.ts
@@ -1,5 +1,6 @@
 import { Component, ChangeDetectionStrategy } from '@angular/core';
 import { InvitationQuery } from '@blockframes/invitation/+state';
+import { AuthQuery } from '@blockframes/auth/+state/auth.query';
 
 @Component({
   selector: 'festival-invitation',
@@ -12,6 +13,6 @@ export class InvitationComponent {
   // Invitation that require an action
   invitations$ = this.query.toMe();
 
-  constructor(private query: InvitationQuery) { }
+  constructor(private query: InvitationQuery, private authQuery: AuthQuery) { }
 
 }
--- a/apps/festival/festival/src/app/dashboard/invitation/invitation.component.ts
+++ b/apps/festival/festival/src/app/dashboard/invitation/invitation.component.ts
@@ -1,6 +1,5 @@
 import { Component, ChangeDetectionStrategy } from '@angular/core';
 import { InvitationQuery } from '@blockframes/invitation/+state';
-import { AuthQuery } from '@blockframes/auth/+state/auth.query';
 
 @Component({
   selector: 'festival-invitation',
@@ -13,6 +12,6 @@ export class InvitationComponent {
   // Invitation that require an action
   invitations$ = this.query.toMe();
 
-  constructor(private query: InvitationQuery, private authQuery: AuthQuery) { }
+  constructor(private query: InvitationQuery) { }
 
 }
--- a/apps/festival/festival/src/app/marketplace/invitation/invitation.component.ts
+++ b/apps/festival/festival/src/app/marketplace/invitation/invitation.component.ts
@@ -1,5 +1,6 @@
 import { Component, ChangeDetectionStrategy } from '@angular/core';
 import { InvitationQuery } from '@blockframes/invitation/+state';
+import { AuthQuery } from '@blockframes/auth/+state/auth.query';
 
 @Component({
   selector: 'festival-invitation',
--- a/apps/festival/festival/src/app/marketplace/invitation/invitation.component.ts
+++ b/apps/festival/festival/src/app/marketplace/invitation/invitation.component.ts
@@ -1,6 +1,5 @@
 import { Component, ChangeDetectionStrategy } from '@angular/core';
 import { InvitationQuery } from '@blockframes/invitation/+state';
-import { AuthQuery } from '@blockframes/auth/+state/auth.query';
 
 @Component({
   selector: 'festival-invitation',
--- a/libs/event/src/lib/components/screening-item/screening-item.component.ts
+++ b/libs/event/src/lib/components/screening-item/screening-item.component.ts
@@ -2,6 +2,8 @@ import { Component, ChangeDetectionStrategy, Input } from '@angular/core';
 import { InvitationQuery, Invitation } from '@blockframes/invitation/+state';
 import { ImgRef } from '@blockframes/utils/media/media.firestore';
 import { ScreeningEvent } from '../../+state';
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
 
 @Component({
   selector: 'event-screening-item',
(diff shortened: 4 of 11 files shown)


apps/festival/festival/src/app/dashboard/invitation/invitation.component.ts
```diff
@@ -9,8 +10,8 @@ import { InvitationQuery } from '@blockframes/invitation/+state';
 })
 export class InvitationComponent {
 
-  invitations$ = this.query.selectAll();
+  invitations$ = this.query.selectAll({ filterBy: invitation => !!invitation.fromOrg || (invitation.fromUser.uid !== this.authQuery.userId) });
```
ioaNikas, line 13: If you do `invitation.fromUser.uid !== this.authQuery.userId`, won't it get you all the invitation from other users ?

Did you mean to use `&&` instead of `||` as an operator ?

apps/festival/festival/src/app/marketplace/invitation/invitation.component.ts
```diff
@@ -9,6 +10,6 @@ import { InvitationQuery } from '@blockframes/invitation/+state';
 })
 export class InvitationComponent {
 
-  invitations$ = this.query.selectAll();
-  constructor(private query: InvitationQuery) { }
+  invitations$ = this.query.selectAll({ filterBy: invitation => !!invitation.fromOrg || (invitation.fromUser.uid !== this.authQuery.userId) });
```
ioaNikas, line 13: Same as above.

libs/notification/src/lib/invitation/components/action/action.component.ts
```diff
@@ -1,26 +1,41 @@
...
+  public isFromUser: boolean;
+
+  constructor(
+    private service: InvitationService,
+    private authQuery: AuthQuery,
+    private snackBar: MatSnackBar) {}
+
+  ngOnInit() {
+    this.isFromUser = !!this.invitation && this.invitation.fromUser.uid === this.authQuery.userId;
+  }
```
ioaNikas, line 27: Instead of using on init, you can maybe set it with the input, like this 
```
@Input() set invitation(invitation: Invitation) {
  this.isFromUser = !!invitation && invitation.fromUser.uid === this.authQuery.userId;
  // You can also use "invitation?.fromUser.uid === this.authQuery.userId" to make it shorter
}

File: libs/notification/src/lib/invitation/components/action/action.component.html [hunk 4]
```diff
@@ -1,13 +1,22 @@
...
 <ng-container *ngIf="invitation; else noInvitation">
-  <button mat-button test-id="invitation-accept" [color]="invitation.status === 'accepted' ? 'primary' : ''" (click)="accept(invitation)">
-    <mat-icon svgIcon="check_circle"></mat-icon>
-    <span>Accept Invitation</span>
-  </button>
-  <button mat-button test-id="invitation-refuse" [color]="invitation.status !== 'accepted' ? 'primary' : ''" (click)="decline(invitation)">
-    <mat-icon svgIcon="close"></mat-icon>
-    <span>Decline Invitation</span>
-  </button>
+  <ng-container *ngIf="(invitation.status === 'pending' && canAction) else displayStatus">
```
GrandSchtroumpf, line 3: I think you can decline an invitation that has been accepted already.

libs/notification/src/lib/invitation/components/action/action.component.html
```diff
@@ -1,13 +1,22 @@
...
+    <button mat-button test-id="invitation-accept" [color]="invitation.status === 'accepted' ? 'primary' : ''" (click)="accept(invitation)">
+      <mat-icon svgIcon="check_circle"></mat-icon>
+      <span>Accept Invitation</span>
+    </button>
+    <button mat-button test-id="invitation-refuse" [color]="invitation.status !== 'accepted' ? 'warn' : ''" (click)="decline(invitation)">
```
GrandSchtroumpf, line 8: Better do `[color]="invitation.status === 'decline' ? 'warn' : ''"`. Like that we can support more status in the future

libs/notification/src/lib/invitation/components/action/action.component.ts
```diff
@@ -1,26 +1,41 @@
...
-  constructor(private service: InvitationService) {}
+  public canAction: boolean;
+
+  constructor(
+    private service: InvitationService,
+    private authQuery: AuthQuery,
+    private snackBar: MatSnackBar) {}
+
+  ngOnInit() {
+    this.canAction = !!this.invitation && ((!(this.invitation.fromUser?.uid === this.authQuery.userId) && this.invitation.toOrg?.id === this.authQuery.orgId) || this.invitation.toUser?.uid === this.authQuery.userId);
```
GrandSchtroumpf, line 26: We shouldn't do that on the invitation action component. You should only use the action on invitation that need action. This is not the job of the component to tell if it should be used.

Also, this is super difficult to read ^^. When you do that, you need to create human readable variables that explains what happens.

libs/notification/src/lib/invitation/components/action/action.component.ts
```diff
@@ -1,26 +1,41 @@
...
+  }
 
   accept(invitation: Invitation) {
     this.service.acceptInvitation(invitation);
+    // Todo try to redirect user on the event page when user clics on close button (rename it into More details)
+    this.snackBar.open('Invitation accepted', 'close', { duration: 2000 });
```
GrandSchtroumpf, line 32: I don't think it make sense to add aggressive feedback like "snackBar" when you already change the color of the invitation

libs/notification/src/lib/invitation/components/action/action.component.html
Hugo-Hontans: useless code because now we don't have buttons when we did a choice

libs/notification/src/lib/invitation/components/action/action.component.html
Hugo-Hontans: We decided it Monday with @Jackseed and @2amarkina

apps/backend-functions/src/invitation.ts
```diff
@@ -111,8 +111,12 @@ export const inviteUsers = (data: UserInvitation, context: CallableContext): Pro
     for (const email of data.emails) {
       getOrCreateUserByMail(email, user.orgId, invitation.type, data.app)
         .then(u => createPublicUser(u))
-        .then(toUser => invitation.toUser = toUser)
-        .then(_ => db.collection('invitations').add(invitation))
+        .then(toUser => {
+          invitation.toUser = toUser;
+          const id = db.collection('invitations').doc().id;
```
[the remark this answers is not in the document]
Hugo-Hontans: I saw with @phpgeekfr we only need to create an id here in the backend functions, so I don't think that is usefull to create a function for that now

libs/notification/src/lib/invitation/components/action/action.component.ts
Hugo-Hontans: I remove it

libs/notification/src/lib/invitation/components/action/action.component.ts
Hugo-Hontans: Change for readable variables :)
But I did this because we use the same component for too many different cases I think.
If you don't find it clear, I can open a issue for that

libs/event/src/lib/components/screening-item/screening-item.component.ts
```diff
@@ -12,12 +14,14 @@ import { ScreeningEvent } from '../../+state';
...
   public poster: ImgRef;
   public screening: ScreeningEvent;
-  public invitation: Invitation;
+  public invitation$: Observable<Invitation>;
 
   @Input() set event(screening: ScreeningEvent) {
     this.screening = screening;
     this.poster = screening.movie?.promotionalElements.poster[0].media;
-    this.invitation = this.invitationQuery.getAll().find(e => e.docId === screening.id);
+    this.invitation$ = this.invitationQuery.selectAll().pipe(
```
GrandSchtroumpf, line 22: If you want only one, use `selectEntity(invitation => invitation.docId === screening.id)`

libs/notification/src/lib/invitation/components/action/action.component.html
```diff
@@ -1,13 +1,22 @@
...
-  <button mat-button test-id="invitation-accept" [color]="invitation.status === 'accepted' ? 'primary' : ''" (click)="accept(invitation)">
-    <mat-icon svgIcon="check_circle"></mat-icon>
-    <span>Accept Invitation</span>
-  </button>
-  <button mat-button test-id="invitation-refuse" [color]="invitation.status !== 'accepted' ? 'primary' : ''" (click)="decline(invitation)">
-    <mat-icon svgIcon="close"></mat-icon>
-    <span>Decline Invitation</span>
-  </button>
+<ng-container *ngIf="invit; else noInvitation">
+  <ng-container *ngIf="(invit.status === 'pending' && canAction) else displayStatus">
```
GrandSchtroumpf, line 3: Create a pipe for that : `*ngIf="invitation | hasAction else displayStatus"`.
The code of the pipe in the comment below

libs/notification/src/lib/invitation/components/action/action.component.ts
```diff
@@ -11,9 +12,25 @@ import { Invitation, InvitationService } from '../../+state';
...
+  public canAction: boolean;
+
+  @Input() set invitation(invit: Invitation) {
+    this.invit= invit;
+    if (!!this.invit) {
+      const userId = this.authQuery.userId;
+      const isNotFromUser = !(this.invit.fromUser?.uid === userId);
+      const isToOrg = this.invit.toOrg?.id === this.authQuery.orgId;
+      const isToUser = this.invit.toUser?.uid === userId;
+      this.canAction = (isNotFromUser && isToOrg) || isToUser;
```
GrandSchtroumpf, line 26: Create a pipe to improve performance and make the code cleaner. Put it in the same file than the component and declare it (without exporting it) in the same module.
```typescript
@Pipe({ name: 'hasAction', pure: true })
export class InvitationHasActionPipe implements PipeTransform {
  constructor(private authQuery: AuthQuery) {}
  transform(invitation: Invitation) {
      const userId = this.authQuery.userId;
      const isNotFromUser = !(this.invit.fromUser?.uid === userId);
      const isToOrg = this.invit.toOrg?.id === this.authQuery.orgId;
      const isToUser = this.invit.toUser?.uid === userId;
      return (isNotFromUser && isToOrg) || isToUser;
  }
}
```

PR #2861, opened 2020-05-26
Feat 2818 - mail maintenance

phpgeekfr:


apps/backend-functions/src/templates/mail.ts
```diff
@@ -64,40 +67,40 @@ export function organizationWasAccepted(email: string, userFirstName?: string, a
...
   const data = {
     userFirstName,
     orgName
   };
-  return { to: email, templateId: templateIds.joinAnOrgPending, data };
+  return { to: email, templateId: templateIds.request.joinOrganization.pending, data };
 }
 
-export function organizationCanAccessApp(admin: PublicUser, appName: App, appUrl: string): EmailTemplateRequest {
+export function organizationAppAccessChanged(admin: PublicUser, appLabel: String, appUrl: string): EmailTemplateRequest {
```
GrandSchtroumpf, line 81: use the type `string` instead of the class definition `String`

apps/backend-functions/src/users.ts
```diff
@@ -24,6 +26,20 @@ export const startVerifyEmailFlow = async (data: any) => {
...
+  const { email, app, firstName } = data;
+  const from = getSendgridFrom(app);
+
+  if (!email) {
+    throw new Error('email is a mandatory parameter for the "sendVerifyEmail()" function');
+  }
+
+  const verifyLink = await admin.auth().generateEmailVerificationLink(email);
+  const template = accountCreationEmail(email, verifyLink, firstName);
+  await sendMailFromTemplate(template, from);
```
GrandSchtroumpf, line 41: What happens if one of these 3 statements fail ?

apps/backend-functions/src/users.ts
```diff
@@ -34,7 +50,8 @@ export const startResetPasswordEmail = async (data: any) => {
   }
 
   const resetLink = await admin.auth().generatePasswordResetLink(email);
-  await sendMailFromTemplate(userResetPassword(email, resetLink), from);
+  const template = userResetPassword(email, resetLink);
+  await sendMailFromTemplate(template, from);
```
GrandSchtroumpf, line 54: Same as above about the fail. Should you add a try/catch ?

libs/utils/src/lib/apps.ts
```diff
@@ -82,11 +84,25 @@ export function getMoviePublishStatus(a: App): StoreStatus {
   return a === 'festival' ? 'accepted' : 'submitted';
 }
 
+/**
+ * Returns the domain that should be used in emails depending on the current app
+ * @param a 
+ */
+export function getSendgridUrl(a?: App): string {
```
GrandSchtroumpf, line 91: Why not using a mapping : 
```typescript
export const sendgridUrl: Record<App, string> = {
  festival: appUrlMarket,
  catalog: appUrlContent
}
```

apps/backend-functions/src/internals/invitations/events.ts
```diff
@@ -37,63 +50,75 @@ async function onInvitationToAnEventCreate({
...
+    /**
+     * @dev We wants to send this email only if user have an orgId. If not, this means that he already received an
+     * email inviting him along with his credentials.
+    */
+    const user = await getDocument<PublicUser>(`users/${toUser.uid}`);
+    if (!user.orgId) {
+      console.log('Invitation have already been sent along with user credentials');
+      return;
+    }
+    recipients.push({ email: toUser.email, name: toUser.firstName });
```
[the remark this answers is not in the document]
phpgeekfr: Name is the name of the recipient used in email (Hi {{firstName}}).  It is only firstName because I think the email template was designed like that.

apps/backend-functions/src/internals/invitations/events.ts
```diff
@@ -37,63 +50,75 @@ async function onInvitationToAnEventCreate({
...
-    }
-    console.log(`Sending invitation email for an event (${docId}) from ${senderName} to : ${recipient}`);
-    const urlToUse = await getAppUrl(org);
-    const from = await getFromEmail(org);
-    return await sendMailFromTemplate(invitationToScreeningFromOrg(toUser, fromOrg.denomination.full, event.title, link, urlToUse), from);
+    const org = await getDocument<OrganizationDocument>(`orgs/${fromOrg.id}`);
+    const senderName = org.denomination.public || org.denomination.full;
+    const link = getEventLink(org);
+    const urlToUse = getSendgridUrl(appSlug);
+    const appName = getAppName(appSlug);
```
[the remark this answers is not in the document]
phpgeekfr: renamed slug to label

apps/backend-functions/src/data/internals.ts
```diff
@@ -111,16 +111,24 @@ export async function getOrgAppSlug(_org: OrganizationDocument | string): Promis
...
-  return getSendgridFrom(slug);
+  const key = await getOrgAppKey(_org);
+  return getSendgridFrom(key);
+}
+
+/**
+ * This check if org have access to marketplace in at least one app
+ * @param org 
+ */
+export function canAccessMarketplace(org: OrganizationDocument): boolean {
```
GrandSchtroumpf, line 131: You may want to have the same for dashboard ^^. You can also fatorize if you want : 
```typescript
function canAccess(module: Module, org Organization) {
  return app.some(a => org.appAccess[a]?[module])
}
```

apps/backend-functions/src/internals/users.ts
```diff
@@ -39,20 +39,11 @@ export const getOrCreateUserByMail = async (email: string, orgId: string, invita
...
-      case 'joinOrganization':
-      default:
-        templateToUse = slug === 'festival' ? templateIds.user.credentials.market : templateIds.user.credentials.content;
-        break;
-    }
-    const template = userInvite(email, password, org.denomination.full, urlToUse, templateToUse);
+    const urlToUse = sendgridUrl[app];
+    const from = getSendgridFrom(app);
+
+    const templateId = templateIds.user.credentials[invitationType][app];
```
GrandSchtroumpf, line 45: 😍

Issue #2809, opened 2020-05-19
MAINTENANCE - Login / Creation flow


Jackseed:
- [ ] Login 
- [x] Forgot password
- [x] session terminated on logout? cannot create account or login after
- [x] If you are not logged and try to access a page with auth guard (https://archipelcontent.com/c/o/marketplace/search for example) you get a blank page instead of beeing redirected to /

Jackseed:
- [ ] change the url `/auth/connexion#signin` for `/auth/connexion#login`

maltmayer:
- [ ] Correct text + go to line at the end of sentences
« It seems to be your first time using one of the Blockframes applications. 
To create your account, we need to make sure that you’re attached to a company. 
Please choose an option below: » 
![image](https://user-images.githubusercontent.com/55785823/82688048-7f98d900-9c58-11ea-863c-5060bace776c.png)

maltmayer:
- [ ] Correct text:
"Thank you for creating your organization!
You’ll receive an email confirmation once your organization is validated.
Thank you for your patience."

- [ ] Delete CTA 'Go to marketplace"

![image](https://user-images.githubusercontent.com/55785823/82688466-3bf29f00-9c59-11ea-8276-eeb4c0a3b714.png)

![image](https://user-images.githubusercontent.com/55785823/82688213-c4247480-9c58-11ea-84c2-2619347bc075.png)

Jackseed:
- [ ] Redirect connected user that goes to landing page to homepage

phpgeekfr:
`If you are not logged and try to access a page with auth guard (https://archipelcontent.com/c/o/marketplace/search for example) you get a blank page instead of beeing redirected to /` This seems to be corrected already

phpgeekfr:
@Jackseed `Redirect connected user that goes to landing page to homepage`. What is the landing page please ?

Jackseed:
- [ ] Add this message on the snackbar error when a user tries to create an already existing account:
"An account already exists for this email address. 
If you don't remember your password, please reset it or try searching in your past emails. You might have received an invitation containing credentials to join Archipel."
+ if possible link to the reset password page on "please reset it"

Jackseed:
> @Jackseed `Redirect connected user that goes to landing page to homepage`. What is the landing page please ?
@phpgeekfr 
it's www.archipelmarket.com or www.archipelcontent.com

phpgeekfr:
[...]

PR #2867, opened 2020-05-26
 #2809 Login maintenance

phpgeekfr:


libs/landing/src/lib/toolbar/toolbar.component.html
```diff
@@ -6,7 +6,7 @@
     <a fxHide.lt-md mat-stroked-button [color]="color" href="#demo" id="demo-link">Schedule a demo</a>
   </nav>
   <div fxLayout="row" fxLayoutAlign="space-around center" fxLayoutGap="16px">
-    <a test-id="login" mat-button [color]="color" routerLink="/auth/connexion" fragment="signin">Log in</a>
+    <a test-id="login" mat-button [color]="color" routerLink="/auth/connexion" fragment="login">Log in</a>
```
GrandSchtroumpf, line 9: Hmm would be better to keep signin everywhere is it's possible. We use `signin` because it's the key used by firebase authentication.

libs/organization/src/lib/organization/guard/pending-organization.guard.ts
```diff
@@ -36,7 +39,9 @@ export class PendingOrganizationGuard extends CollectionGuard<OrganizationState>
               }
 
               if (org.status === 'accepted') {
-                return '/c/o/dashboard/home';
+                const app = getCurrentApp(this.routerQuery);
+                const moduleAccess = getOrgModuleAccess(org, app);
```
GrandSchtroumpf, line 43: If you only want the first one you can do `const [ moduleAccess ] = getOrgModuleAccess(org, app)`.
Or even `const [ moduleAccess = 'dashboard' ] = getOrgModuleAccess(org, app)` to have a default value.
It's up to you. Maybe you think it's not more readable ^^.

libs/user/src/lib/auth/components/widget/widget.component.ts
```diff
@@ -25,6 +25,7 @@ export class AuthWidgetComponent {
 
   public async logout() {
     await this.service.signOut();
+    window.location.reload();
```
GrandSchtroumpf, line 28: I thought we didn't needed to do that anymore. Could we just `router.navigate(['/'])` ? or `navigateByUrl`

libs/user/src/lib/auth/guard/no-auth.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+    private afAuth: AngularFireAuth,
+    private routerQuery: RouterQuery,
+  ) {
+    super(service);
+  }
+
+  sync() {
+    return this.afAuth.authState.pipe(
+      switchMap(userAuth => {
+        if (!userAuth) { return new Promise(r => r(true)); }
```
GrandSchtroumpf, line 28: `if (!userAuth) return of(true)` or `Promise.resolve(true)` if you want to keep with the Promise

libs/user/src/lib/auth/guard/no-auth.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+        return this.service.sync().pipe(
+          catchError(() => new Promise(r => r(true))),
```
GrandSchtroumpf, line 30: Same here

libs/user/src/lib/auth/guard/no-auth.guard.ts
```diff
@@ -0,0 +1,43 @@
...
+          map(_ => this.query.orgId),
+          switchMap(orgId => orgId ? this.orgService.getValue(orgId) : new Promise<false>(r => r(false))),
+          map(org => {
+            if (!org) { return '/c/organization'; }
+            const app = getCurrentApp(this.routerQuery);
+            const moduleAccess = getOrgModuleAccess(org, app);
```
GrandSchtroumpf, line 36: Same as above for the first access if you want

libs/user/src/lib/auth/pages/identity/identity.component.ts
```diff
@@ -58,7 +67,12 @@ export class IdentityComponent {
         .where('type', '==', 'joinOrganization')
         .where('toUser.uid', '==', this.query.userId));
       const pendingInvitation = invitations.find(invitation => invitation.status === 'pending');
-      await this.db.doc(`invitations/${pendingInvitation.id}`).update({ status: 'accepted' });
+      if (!!pendingInvitation) {
+        // We put invitation to accepted only if the invitation.type is joinOrganization.
+        // Otherwise, user will have to create or join an org before accepting the invitation (to attend event for example)
+        await this.db.doc(`invitations/${pendingInvitation.id}`).update({ status: 'accepted' });
```
GrandSchtroumpf, line 73: Could you use the invitationService here instead for consistency ?

libs/user/src/lib/auth/pages/login/login.component.scss
```diff
@@ -45,3 +47,11 @@ mat-sidenav {
     width: 100%;
   }
 }
+
+.custom-snackbar-message {
+  text-align: center;
+  a {
+    color: mat-color($bf-dark-foreground, blue-icon);
```
GrandSchtroumpf, line 54: Don't import theme related stuff in component scss. Either create a component theme file under styles/components, or use css variables `--primary`.
This might lead to bundle size leak. We should be carefull with that. If it doesn't exactly match the expected design it's not the most important here.

libs/user/src/lib/auth/components/widget/widget.component.ts
phpgeekfr: Page needs to be reloaded completely to re-open closed connection with firebase

libs/organization/src/lib/organization/guard/pending-organization.guard.ts
phpgeekfr: I like this KFH way

PR #2870, opened 2020-05-26
Add message for maxlenght

GrandSchtroumpf:
Add a message when there is too many invitations

### DIFF
--- a/libs/notification/src/lib/invitation/form/user/user.component.html
+++ b/libs/notification/src/lib/invitation/form/user/user.component.html
@@ -18,7 +18,7 @@
 <ng-container *ngIf="form.getError('maxlength') as error">
   <article class="maxlength" @scaleIn fxLayout fxLayoutAlign="space-between start" fxLayoutGap="24px">
     <p class="mat-body-1">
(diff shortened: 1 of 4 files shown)


libs/notification/src/lib/invitation/form/user/user.component.html
```diff
@@ -14,13 +14,35 @@
   </button>
 </div>
 
+<!-- Too many emails -->
+<ng-container *ngIf="form.getError('maxlength') as error">
+  <article class="maxlength" @scaleIn fxLayout fxLayoutAlign="space-between start" fxLayoutGap="24px">
+    <p class="mat-body-1">
+      Unfortunately, the current version do not allow you to send more than {{ error.requiredLength }} invitations at once for now.
```
fritzschoff, line 21: current version of the app? of the our product?

fritzschoff: why requiredLength? should be more like maxLength or something?

GrandSchtroumpf: `requiredLength` is the value given by Validators.maxlength

PR #2876, opened 2020-05-27
Merging the imgRef and asset into one, yay :) Issue #2528

joanofdart:
Merged both imgRef and asset directives in one.
Changed the **imgRef** naming convention to just **ref**
Removed **placeholderUrl** since we don't use it anymore and the images are already within each theme.

Issue #2528

PR #2877, opened 2020-05-27
Landing img

Hugo-Hontans:
- [x] Fix app logo festival
- [x] Fix images in landing page

### DIFF
--- a/libs/landing/src/lib/landing.module.ts
+++ b/libs/landing/src/lib/landing.module.ts
@@ -26,6 +26,7 @@ import { MatDividerModule } from '@angular/material/divider';
 import { MatButtonModule } from '@angular/material/button';
 import { MatSnackBarModule } from '@angular/material/snack-bar';
 import { MatListModule } from '@angular/material/list';
+import { ImgAssetModule } from '@blockframes/ui/theme/img-asset.module';
 
 @NgModule({
   declarations: [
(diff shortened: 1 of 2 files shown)


libs/landing/src/lib/landing.module.ts
```diff
@@ -42,6 +43,7 @@ import { MatListModule } from '@angular/material/list';
     AppLogoModule,
     TelInputModule,
     RoleModule,
+    ImgAssetModule,
```
joanofdart, line 46: We are not gonna be using the image asset module anymore. Please do await for https://github.com/blockframes/blockframes/pull/2876 since we got rid of imgAssetModule and now we should use just imageReferenceModule instead.

libs/ui/src/lib/layout/app-logo/app-logo.component.ts
```diff
@@ -3,7 +3,7 @@ import { RouterQuery } from '@datorama/akita-ng-router-store';
 
 const appLogos = {
   catalog: 'LogoArchipelContentPrimary.svg',
-  festival: 'archipel_market.png'
+  festival: 'logo_archipel_market_home.svg'
```
joanofdart, line 6: Could we have the same naming convention for both? either camel case or kebab 🤔 
it feels weird having one NamedLikeThis and the other named_like_this. What do you think? 😄

PR #2878, opened 2020-05-27
Migration

### LINKED ISSUE #2868: MAINTENANCE - Migration script
- [x] Correct typo on invitations from migration script 15
- [x] Change the old activity to match the list (see with @Jackseed for what activity should be given to existing orgs)
- [x] Add a country to every org addresses

Mussuranae:
Bye bye organizaton on notification and invitation.
Updated activity for organization with the new static (const in the code).
Fix #2868

### DIFF
--- a/apps/backend-ops/src/firestoreMigrations/0023.ts
+++ b/apps/backend-ops/src/firestoreMigrations/0023.ts
@@ -1,0 +1,89 @@
+import { Firestore } from '../admin';
+
+
+export async function upgrade(db: Firestore) {
+  const invitations = await db.collection('invitations').get();
+  const notifications = await db.collection('notifications').get();
+  const orgs = await db.collection('orgs').get();
+  const batch = db.batch();
+
+  // Update invitations with organization
+  invitations.docs.forEach(doc => {
+    const invitation = doc.data();
+    const { organizaton } = invitation;
+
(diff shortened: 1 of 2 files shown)


apps/backend-ops/src/firestoreMigrations/0023.ts
```diff
@@ -0,0 +1,89 @@
...
+    const { organizaton } = notification;
+
+    if (organizaton) {
+      delete notification.organizaton;
+      const newData = {
+        ...notification,
+        organization: organizaton,
+      };
+
+      if (!newData) {
```
phpgeekfr, line 42: Is this check important? If so maybe you should throw an error because as it is right now, you will try to set an empty doc in `return batch.set(doc.ref, newData);`

apps/backend-ops/src/firestoreMigrations/0023.ts
```diff
@@ -0,0 +1,89 @@
...
+    const { organizaton } = invitation;
+
+    if (organizaton) {
+      delete invitation.organizaton;
+      const newData = {
+        ...invitation,
+        organization: organizaton,
+      };
+
+      if (!newData) {
```
phpgeekfr, line 22: same as bellow

PR #2880, opened 2020-05-28
Issue#2810 maintenance org settings

ioaNikas:
Update organization form (comment bankAccount parts, logo as ImgRefForm)
Update image resizing function to generate original url
Change formControlName used by organization form in the cropper


apps/backend-functions/src/internals/image.ts
```diff
@@ -98,58 +97,64 @@ async function resize(data: functions.storage.ObjectMetadata) {
...
 
-    const value = {
-      ref: filePath,
-
-      urls: {
-        ...uploadedUrl,
-        original,
-      }
-    }
+    const value = { ref: filePath, urls };
```
fritzschoff, line 153: so if you update the original control as well now, how do you trigger your local form to update the formControls. This only happend when you update the original formControl locally?

libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
```diff
@@ -95,9 +95,43 @@ export class CropperComponent implements OnDestroy {
...
+  // update the parent form field when there is change in the component (component -> parent)
+  registerOnChange(fn: any): void {
+    this.uploaded = (ref: string) => fn(ref);
+    this.deleted = () => fn(createImgRef());
+  }
+
+  registerOnTouched(fn: any): void {
+    return;
+  }
+
```
fritzschoff, line 134: you want the controlValueAccessor back again? why? This will crash in movie form tunnel

apps/backend-functions/src/internals/image.ts
```diff
@@ -98,58 +97,64 @@ async function resize(data: functions.storage.ObjectMetadata) {
...
 
     if (docData === undefined) {
       throw new Error('Data is undefined');
     }
 
-    const original = (get(docData, fieldToUpdate) as ImgRef).urls.original;
-
     // format an array of {key, value}[] into a record of {key1: value1, key2: value2, ...}
-    const uploadedUrl: Record<string, string> = {}
+    const urls: Record<string, string> = {};
```
phpgeekfr, line 148: can you type the "key" with a custom type or interface ? so we know what is inside the first string of the record?

apps/backend-functions/src/internals/image.ts
```diff
@@ -61,7 +59,7 @@ async function resize(data: functions.storage.ObjectMetadata) {
 
   if (uploadedSize !== 'original') {
     console.info('skipping upload that is not "original": ' + uploadedSize);
-    return;
+    return false;
```
phpgeekfr, line 62: does it make any difference ?

ioaNikas: Nope, it's something I added when tried to debug to fit the other return cases (all are them are false or true)

apps/backend-functions/src/internals/image.ts
ioaNikas: Opened an issue for this (#2882), I lack knowledge on this and will try to clean it later

libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
ioaNikas: Removed this code

PR #2887, opened 2020-05-28
Invitations ui

Hugo-Hontans:

### DIFF
--- a/libs/notification/src/lib/invitation/components/action/action.component.scss
+++ b/libs/notification/src/lib/invitation/components/action/action.component.scss
@@ -1,4 +1,17 @@
 :host {
   display: flex;
-  
+}
+
+button {
+  span {
+    font-size: 13px;
+  }
+}
+
+.primary {
+  color: var(--primary)
+}
+
+.warn {
+  color: var(--warn)
 }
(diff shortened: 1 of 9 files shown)


libs/notification/src/lib/invitation/components/item/item.component.html
```diff
@@ -2,12 +2,8 @@
   <img matListAvatar *ngIf="invitation.fromUser" placeholderAsset="profil_user.webp" [imgRef]="invitation.fromUser.avatar" alt="Profile picture" />
   <img matListAvatar *ngIf="invitation.fromOrg" placeholderAsset="profil_user.webp" [imgRef]="invitation.fromOrg.logo" alt="Organization logo" />
   <p matLine class="mat-body-1">
```
GrandSchtroumpf, line 4: I think you don't need matLine anymore if you have only one

libs/notification/src/lib/invitation/components/action/action.component.scss
```diff
@@ -1,4 +1,17 @@
 :host {
   display: flex;
-  
-}
\ No newline at end of file
+}
+
+button {
+  span {
+    font-size: 13px;
```
GrandSchtroumpf, line 7: Prefere existing class like "mat-body-1" or "mat-body-2" or "mat-caption" that apply theme style.
We want the button to always behave the same.

PR #2888, opened 2020-05-28
added filter count for movies

fritzschoff:


libs/ui/src/lib/layout/layout.module.ts
```diff
@@ -104,7 +104,7 @@ export class Flex implements OnInit, OnDestroy {
   constructor(private layout: Layout) {}
 
   /* Use this function if you need the number value of calculation */
-  get marginOffset() {
+  marginOffset() {
```
fritzschoff, line 107: sometimes the getter was undefined

libs/movie/src/lib/movie/+state/movie.query.ts
```diff
@@ -14,12 +14,10 @@ export class MovieQuery extends QueryEntity<MovieState, Movie> {
     super(store);
   }
 
-  /** @deprecated Use orgQuery.hasMovies$ instead */
   hasMovies(predicate?: Parameters<QueryEntity<MovieState, Movie>['selectCount']>[0]) {
```
fritzschoff, line 17: why should we use the org query if we want to see if movies are available ?

apps/festival/festival/src/app/marketplace/home/home.component.ts
```diff
@@ -6,9 +6,11 @@ import { MovieQuery, MovieMain, MovieService, Movie } from '@blockframes/movie/+
 
 // RxJs
 import { Observable, Subscription } from 'rxjs';
+import { map } from 'rxjs/operators';
 
 interface CarouselSection {
   title: string;
+  hasEnoughMovies$: Observable<boolean>;
```
GrandSchtroumpf, line 13: I think `movieCount$` would be better. Like that we could display the amount in the title for example. And you can set the minium amount in the HTML directly.

libs/movie/src/lib/movie/+state/movie.query.ts
GrandSchtroumpf: The final goal is to use akita only for organization own movies (like for the invitation & notification). 
Like that you have a direct query to firestore for `home `& `title`. Which would solve some `reset `issues between thoses pages.

libs/ui/src/lib/layout/layout.module.ts
GrandSchtroumpf: What about returning 0 if _margin is not set yet ? For me it would make more sense. But a function is fine too (I wonder why it would work better though).

fritzschoff: super weird, but the variable should be always defined, cause I'm setting the directive in the homepage...

Issue #2848, opened 2020-05-22
we should use callbable functions to get the proper application used instead of guessing it with org id


phpgeekfr:

phpgeekfr:
try to remove use of  `getFromEmail` and `getAppUrl`

phpgeekfr:
idea: create _meta.updatedFrom ?

Jackseed:
@phpgeekfr still an open issue? 
Looks connected to #5049

phpgeekfr:
This is more a reminder than a real issue but yes it can be linked to #5049

phpgeekfr:
@Mussuranae did a part of this issue in https://github.com/blockframes/blockframes/pull/5464 . Since this one is old and that main issues are solved in @Mussuranae PR, I'm closing this one

PR #2893, opened 2020-05-28
movie import maintenance

phpgeekfr:


libs/ui/src/lib/upload/file-upload/file-upload.component.scss
```diff
@@ -16,6 +16,16 @@ section {
   transition: border 0.2s ease-in-out;
 }
 
+.upload-button-container {
+  height: 60px;
+  button {
+    width: 60px;
+    .add-files-icon {
```
GrandSchtroumpf, line 23: It would be easier to read if you write `img`. It shouldn't create a conflict with the other one because you're inside a button.

libs/utils/src/lib/apps.ts
```diff
@@ -67,11 +67,12 @@ export function getAppName(slug: App) {
 /**
  * Returns the apps that the org have access to
  * @param org
+ * @param first
  * @example
  * getOrgAppAccess(orgA); // ['catalog', 'festival']
  * getOrgAppAccess(orgB); // ['festival']
  */
-export function getOrgAppAccess(org: OrganizationDocument | OrganizationDocumentWithDates): App[] {
+export function getOrgAppAccess(org: OrganizationDocument | OrganizationDocumentWithDates, first: App = 'festival'): App[] {
```
GrandSchtroumpf, line 75: Any reason why you set `first` to festival ? I think it might be confusing. I would rather leave it required :

libs/utils/src/lib/apps.ts
```diff
@@ -81,7 +82,12 @@ export function getOrgAppAccess(org: OrganizationDocument | OrganizationDocument
     }
   }
 
-  return Object.keys(allowedApps).map(k => k as App);
+  const apps = Object.keys(allowedApps).map(k => k as App);
+  if (apps.length > 1 && apps.includes(first)) {
```
GrandSchtroumpf, line 86: It's not super clear what you're doing here, maybe add a comment.

GrandSchtroumpf: And honestly it looks a little bit "fait avec des ficelles" ^^

libs/utils/src/lib/apps.ts
phpgeekfr: Vincent asked me to put festival in first if an org have access to the two apps. It is a fix waiting for another method to retrieve the proper app

libs/utils/src/lib/apps.ts
phpgeekfr: we could also say "avec la bite et le couteau"

phpgeekfr: (this is temporary, waiting for #2848 )

phpgeekfr: added a comment

libs/utils/src/lib/apps.ts
GrandSchtroumpf: maybe create a todo or something if it's a fix waiting for something else

PR #2896, opened 2020-05-29
multiple fixes

pldespaigne:
- [x] fix event notification not displaying first/last name
- [x] fix automatic filter of festival movies
- [x] fix sales agents filter bug (related to orgName in algolia)
- [x] implementation of adaptative streaming for JWPlayer

### DIFF
--- a/apps/backend-functions/src/movie.ts
+++ b/apps/backend-functions/src/movie.ts
@@ -1,10 +1,11 @@
 import { functions, db } from './internals/firebase';
-import { MovieDocument, OrganizationDocument, PublicUser, StoreConfig, PublicOrganization } from './data/types';
+import { MovieDocument, OrganizationDocument, PublicUser, StoreConfig } from './data/types';
 import { triggerNotifications, createNotification } from './notification';
 import { getDocument, getOrganizationsOfMovie } from './data/internals';
 import { removeAllSubcollections } from './utils';
 import { storeSearchableMovie, deleteObject } from './internals/algolia';
 import { centralOrgID, algolia } from './environments/environment';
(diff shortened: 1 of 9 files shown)


apps/backend-functions/src/movie.ts
```diff
@@ -21,7 +21,7 @@ export async function onMovieCreate(
   const organization = await getDocument<OrganizationDocument>(`orgs/${user.orgId}`);
 
   // Update algolia's index
-  return storeSearchableMovie(movie, organization.denomination.full);
+  return storeSearchableMovie(movie, organization.denomination.public || organization.denomination.full);
```
GrandSchtroumpf, line 24: I'm pretty sure there is a `orgName` function in `organization.firestore.ts`. It'll do the job. If not, create it and use it also in the org-name.pipe.ts

GrandSchtroumpf: Update the rest of the PR with that

libs/notification/src/lib/notification/+state/notification.store.ts
```diff
@@ -22,7 +22,7 @@ export class NotificationStore extends EntityStore<NotificationState, Notificati
 
   public formatNotification(notification: Notification): Partial<Notification> {
     const displayName = notification.user ? `${notification.user.firstName} ${notification.user.lastName}` : 'Someone';
-    const orgName = notification.organization?.denomination.full;
+    const orgName = notification.organization?.denomination?.public || notification.organization?.denomination?.full || 'unknown organization';
```
GrandSchtroumpf, line 25: What about "Organization with no name" ?

PR #2898, opened 2020-05-29
change date of invitation-item

Hugo-Hontans:
- [x] Change date

### DIFF
--- a/libs/notification/src/lib/invitation/components/item/item.component.html
+++ b/libs/notification/src/lib/invitation/components/item/item.component.html
@@ -1,13 +1,14 @@
 <mat-list-item>
   <img matListAvatar *ngIf="invitation.fromUser" asset="profil_user.webp" [ref]="invitation.fromUser.avatar" alt="Profile picture" />
   <img matListAvatar *ngIf="invitation.fromOrg" asset="profil_user.webp" [ref]="invitation.fromOrg.logo" alt="Organization logo" />
-  <div matLine fxLayout="row" fxLayoutAlign="space-between center">
-    <p  class="mat-body-1">
-      {{ invitation.message }} {{ invitation.date | date: 'medium' }}
-    </p>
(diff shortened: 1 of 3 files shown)


libs/notification/src/lib/invitation/components/item/item.component.html
```diff
@@ -1,13 +1,14 @@
...
+  <p matLine class="mat-body-1">
+    {{ invitation.message }}
+  </p>
+  <span matLine>
+    <small class="mat-caption">
+      {{ invitation.date | date: 'medium' }}
+    </small>
+  </span>
+  <invitation-action [invitation]="invitation"></invitation-action>
+  <a mat-button [routerLink]="['../event/', invitation.docId]">More details</a>
```
GrandSchtroumpf, line 13: Depending where is the item used this is going to break. I think we are only using this in the invitation list so we can go with that now. But double check if we are not using is somewhere else first.

libs/notification/src/lib/invitation/components/item/item.component.scss
```diff
@@ -0,0 +1,3 @@
+invitation-action {
+  width: 250px;
```
GrandSchtroumpf, line 2: Hmmm it looks quite dangerous to do that. Did you try the responsive ? With the sidenav open & close ?
Maybe a % be might be more appropriate.

libs/notification/src/lib/invitation/components/item/item.component.html
Hugo-Hontans: We are only use it in invitation list, I verified

libs/notification/src/lib/invitation/components/item/item.component.scss
Hugo-Hontans: I use this to fix the link button on the right because of it is not the same size between invitations.
The responsive version is broke in all cases

PR #2899, opened 2020-05-29
Issue 2835

### LINKED ISSUE #2835: ChangeDetectino Bug
it seems to be an issue about change detection.
For example:
- In the Bf carousel component, we want to show or hide arrows depending on the position of the carousel. But you need to click somewhere on the page to trigger the rerendering from angular. Even thou you called changeDetectionRef.markForCheck() in the component. 
- Also if disable buttons when user is scrolling, otherwise the offset will calculate the wrong way

fritzschoff:
close #2835 
would be nice to disable the buttons when it is scrolling for any miss calculation for sliding. but not important right now

### DIFF
--- a/apps/festival/festival/src/app/marketplace/home/home.component.html
+++ b/apps/festival/festival/src/app/marketplace/home/home.component.html
@@ -37,9 +37,9 @@
               <movie-slide [movie]="movie">
                 <movie-slide-cta>
                   <ng-container *ngIf="movie.promotionalElements.trailer">
-                    <a mat-flat-button color="accent" [routerLink]="['/c/o/marketplace/title/' + movie.id]"
+                    <a mat-flat-button color="accent" [routerLink]="['/c/o/marketplace/title/', movie.id]"
                       class="detail">
-                      More details</a>
+                      More Details</a>
                   </ng-container>
                 </movie-slide-cta>
                 <movie-slide-actions>
--- a/apps/festival/festival/src/app/marketplace/home/home.component.html
+++ b/apps/festival/festival/src/app/marketplace/home/home.component.html
@@ -2,7 +2,7 @@
   <header fxLayout="column" fxLayoutAlign="center start" fxLayoutAlign.xs="center center" fxLayoutGap="20px"
     class="dark-contrast-theme">
     <h4>WELCOME TO ARCHIPEL MARKET</h4>
-    <h1>The endless film market</h1>
+    <h1>The ongoing film market</h1>
     <a mat-flat-button color="accent" test-id="library" routerLink="../title">What's on the market today</a>
   </header>
 
--- a/apps/festival/festival/src/app/marketplace/home/home.component.html
+++ b/apps/festival/festival/src/app/marketplace/home/home.component.html
@@ -8,7 +8,7 @@
 
   <section fxLayout="column" fxLayoutGap="48px">
     <ng-container *ngFor="let section of sections; let i = index">
-      <article fxLayout="column" *ngIf="(section.movieCount$ | async) > 4">
+      <article fxLayout="column" *ngIf="(section.movieCount$ | async) >= 4">
         <div fxLayout="row" fxLayoutAlign="space-between start">
           <h1>{{ section.title }}</h1>
           <a mat-button routerLink="../title">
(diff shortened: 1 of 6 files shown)


libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -9,51 +9,62 @@ import {
...
-      this.showForward = this.onScrolling('right')
-    }
-    this.scroll =  this.scrollable.elementScrolled();
-  } */
+  ngAfterViewInit() {
+    this.showBack$ = this.onScrolling('left');
+    this.showForward$ = this.onScrolling('right');
+
+    this.subBack = this.showForward$.subscribe(value => {
+      this.showForward = value;
```
fritzschoff, line 59: like that change detection is working. No idea why

apps/festival/festival/src/app/marketplace/home/home.component.html
```diff
@@ -23,26 +23,33 @@ <h1>{{ section.title }}</h1>
...
+              <mat-icon svgIcon="arrow_forward"></mat-icon>
+            </a>
+          </div>
+          <bf-slider hideIndicators loop class="dark-contrast-theme">
+            <bf-slide *ngFor="let movie of section.movies$ | async" overlayColor="gradient"
+              [image]="movie.promotionalElements.banner.media">
+              <movie-slide [movie]="movie">
+                <movie-slide-cta>
+                  <ng-container *ngIf="movie.promotionalElements.trailer">
+                    <a mat-flat-button color="accent" [routerLink]="['/c/o/marketplace/title/' + movie.id]"
```
GrandSchtroumpf, line 40: use an array format with "," instead of "+" : `[routerLink]="['/c/o/marketplace/title', movie.id]"`

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -64,16 +75,19 @@ export class CarouselComponent /* implements AfterViewInit */ {
...
+  onScrolling(direction: 'right' | 'left') {
     return this.scrollable.elementScrolled().pipe(
       debounceTime(50),
       map(_ => !!this.scrollable.measureScrollOffset(direction)),
       distinctUntilChanged(),
-      tap(_ => this.ngZone.run(() => this.cdr.detectChanges())),
       startWith(direction === 'right'),
-    )
-  } */
+      tap(_ => this.ngZone.run(() => this.cdr.detectChanges())))
```
GrandSchtroumpf, line 84: Remove that if you need to force the change detection after anyway

libs/ui/src/lib/carousel/carousel.component.ts
GrandSchtroumpf: Do something like that then : 
```typescript
this.subForward = this.onScrolling('left').subscribe(showforward => {
  this.showForward = showforward;
  this.zone.run(() => this.cdr.markforCheck()) // or detectChange
})
```
Same for subBack

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -9,51 +9,62 @@ import {
...
+  ngAfterViewInit() {
+    this.showBack$ = this.onScrolling('left');
+    this.showForward$ = this.onScrolling('right');
+
+    this.subBack = this.showForward$.subscribe(value => {
+      this.showForward = value;
+      this.cdr.detectChanges();
+    })
+
+    this.subForward = this.showBack$.subscribe(value => {
```
GrandSchtroumpf, line 63: It's weird to name it `subForward ` for `showBack` I think ^^

libs/ui/src/lib/carousel/carousel.component.scss
```diff
@@ -10,16 +10,18 @@
   scroll-behavior: smooth;
 }
 
-button {
+#button {
```
GrandSchtroumpf, line 13: Id are unique. Use class instead : `class="direction"`

libs/ui/src/lib/carousel/carousel.component.scss
```diff
@@ -10,16 +10,18 @@
...
   position: absolute;
+  background: rgba(255, 255, 255, 0.226);
```
GrandSchtroumpf, line 15: Maybe use mat-mini-fab instead if you want to have a background. and you can have an opacity on hover : 
```typescript
button {
  opacity: 0.5;
}
button:hover {
  opacity: 1;
}
```

GrandSchtroumpf: If it doesn't work, keep the hardcoded background. But I don't like that so much ^^

PR #2908, opened 2020-05-29
Issue#2810 maintenance org settings

ioaNikas:
Update on sales agent cards style


libs/organization/src/lib/organization/components/card/card.component.scss
```diff
@@ -1,3 +1,39 @@
...
+.tab-content {
+  padding: 16px 0;
+}
+
 p {
   text-align: center;
+  margin-bottom: 4px;
+  width: 100%;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
GrandSchtroumpf, line 37: 👍

libs/ui/src/lib/form/country/country.component.ts
```diff
@@ -14,11 +14,15 @@ type TERRITORIES = typeof staticModel.TERRITORIES;
 })
 export class FormCountryComponent implements OnInit {
   @Input() form: FormStaticValue<'TERRITORIES'>;
-  countries = staticModel.TERRITORIES;
+  @Input() noWorld = false;
```
GrandSchtroumpf, line 17: use the decorator `@boolean` here like that you can do `<form-country noWorld ...>` directly : 
```typescript
import { boolean } from '@blockframes/utils/decorators'; // or something like that
...
@Input() @boolean noWorld = false;
```

PR #2909, opened 2020-05-29
Add session guard + stuff

GrandSchtroumpf:


libs/event/src/lib/layout/view/view.component.ts
```diff
@@ -31,7 +31,8 @@ export class EventViewComponent implements OnInit {
   ngOnInit(): void {
     this.invitation$ = this.event$.pipe(
       switchMap(event => this.invitationQuery.selectByDocId(event.id)),
-    )
+      shareReplay(1)
```
phpgeekfr, line 34: what is this ?

libs/event/src/lib/pipes/event-time.pipe.ts
```diff
@@ -1,18 +1,25 @@
 import { Pipe, PipeTransform, NgModule } from '@angular/core';
 import { Event } from '../+state/event.model';
 
+export type EventTime = 'early'|'onTime'|'late';
```
phpgeekfr, line 4: linter

libs/ui/src/lib/app-bar.ts
```diff
@@ -123,6 +128,7 @@ export class PageBarDirective implements AfterViewInit, OnDestroy {
   ngOnDestroy() {
     this.appContainer.appBar.detach();
     if (this.targetId && this.observer) {
+      console.log('Unobserve');
```
phpgeekfr, line 131: forgot to remove ?

libs/event/src/lib/layout/view/view.component.ts
GrandSchtroumpf: To share the result multiple time. else each "| async" will call the database.

PR #2927, opened 2020-06-01
Feat 2866 - general debug

phpgeekfr:


libs/utils/src/lib/apps.ts
```diff
@@ -12,9 +12,10 @@ export type App = typeof app[number];
 export const module = ['dashboard', 'marketplace'] as const;
 export type Module = typeof module[number];
 
-const appName = {
+export const appName = {
   catalog: 'Archipel Content',
-  festival: 'Archipel Market'
+  festival: 'Archipel Market',
+  blockframes: 'Blockframes'
```
ioaNikas, line 18: Does this is still a thing ?

libs/utils/src/lib/dynamic-title/dynamic-title.service.ts
```diff
@@ -3,39 +3,39 @@ import { Subscription } from 'rxjs';
...
       return `${section} ${suffix}`;
     },
-    entityWithSection: (section: string, titleName: string, showAppName: boolean, appName: string) => {
-      const suffix = showAppName ? `- ${appName}` : '';
+    entityWithSection: (section: string, titleName: string, showAppName: boolean, _appName: string) => {
+      const suffix = showAppName ? `- ${_appName}` : '';
       return `${titleName} - ${section} ${suffix}`;
     }
   },
   blockframes: {
```
ioaNikas, line 20: Same as above

libs/utils/src/lib/apps.ts
phpgeekfr: I don't think so but since it was there, I did not wanted to remove it just before release

libs/utils/src/lib/dynamic-title/dynamic-title.service.ts
phpgeekfr: I don't think so but since it was there, I did not wanted to remove it just before release

PR #2930, opened 2020-06-01
Org-chip and space

Mussuranae:
Org-chip is now in dark-contrast-theme (movie-view) and org address has no more space (org-view)

### DIFF
--- a/libs/organization/src/lib/organization/layout/view/view.component.html
+++ b/libs/organization/src/lib/organization/layout/view/view.component.html
@@ -29,13 +29,13 @@
       <h4>{{ org.activity | toLabel: 'orgActivity' | async }}</h4>
       <span class="mat-display-2">{{ org.denomination.full }}</span>
       <h2>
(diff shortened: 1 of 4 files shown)


libs/organization/src/lib/organization/layout/view/view.component.html
```diff
@@ -29,13 +29,13 @@
       <h4>{{ org.activity | toLabel: 'orgActivity' | async }}</h4>
       <span class="mat-display-2">{{ org.denomination.full }}</span>
       <h2>
```
fritzschoff, line 31: don't you want to use a div? now every comma is in a h2 with the style of h2 and the rest is a span with the style of a span?

Mussuranae: That's the strange part, but everything is in the h2 style
image
I think it's because I put no `class="mat-body-1"` in the span so it takes the h2 style

PR #2931, opened 2020-06-01
Fix form error in profile view

ioaNikas:
Fix form error in profile view


apps/backend-functions/src/users.ts
```diff
@@ -80,7 +80,7 @@ export const onUserCreate = async (user: UserRecord) => {
     if (userDoc.exists) {
       if (!user.emailVerified) {
         const u = userDoc.data() as PublicUser;
-        /** 
+        /**
```
fritzschoff, line 83: yes, sorry good point

libs/ui/src/lib/media/image-reference/image-reference.form.ts
```diff
@@ -1,14 +1,13 @@
...
 import { FormEntity } from '@blockframes/utils/form/forms/entity.form';
-import { urlValidators } from '@blockframes/utils/form';
 
 function createImgRefControl(reference: ImgRef) {
   const { ref, urls } = createImgRef(reference);
   return {
     ref: new FormControl(ref),
     urls: new FormGroup({
-      original: new FormControl(urls.original, urlValidators),
+      original: new FormControl(urls.original),
```
fritzschoff, line 10: why removing, it should be always a url?

PR #2943, opened 2020-06-02
Org fixes

ioaNikas:
Fixes small bugs and errors, add filters on organizations and their movies.

### DIFF
--- a/libs/organization/src/lib/organization/components/card/card.component.ts
+++ b/libs/organization/src/lib/organization/components/card/card.component.ts
@@ -17,7 +17,9 @@ export class OrganizationCardComponent {
   @Input()
   set org(org: Organization) {
     this._org = org;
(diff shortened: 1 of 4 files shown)


lsenta:
closing, already in hotfxies

libs/organization/src/lib/organization/components/card/card.component.ts
```diff
@@ -17,7 +17,9 @@ export class OrganizationCardComponent {
   @Input()
   set org(org: Organization) {
     this._org = org;
-    this.movies = org.movieIds.map(movieId => this.movieQuery.getEntity(movieId));
+    this.movies = org.movieIds
+      .map(movieId => this.movieQuery.getEntity(movieId))
```
Hugo-Hontans, line 21: you can't do an selectAll with filter or something like that ?

PR #2944, opened 2020-06-02
worked on design review from Mathilde

### LINKED ISSUE #2938: DESIGN REVIEW - Log in / Sign up
image
- [x] Wrong fab button. It should be like this
image

image
- [x] change check box color in primary
- [x] align center check box and text
- [x] wrong link color : primary 500, not primary 200

image
- [x] too much margin bottom. 80px is enough

image
- [x] use link for "forget password", not a basic button.

image
- [x] form background color in #ffffff
- [x] image background color #F4F4F4

fritzschoff:
close #2938


libs/user/src/lib/auth/components/signin-form/signin-form.component.html
```diff
@@ -26,7 +26,7 @@ <h1>Log in</h1>
           Password is <strong>required</strong>
         </mat-error>
       </mat-form-field>
-      <a type="button" color="primary" mat-button routerLink="../email-verification">Forgot your password ?</a>
+      <a type="button" color="primary" routerLink="../email-verification" class="forgot-password">Forgot your password ?</a>
```
GrandSchtroumpf, line 29: is `color="primary"` really working here ?
What about using a `mat-button` directive or something to keep it consistent with the rest of the code ?

libs/user/src/lib/auth/components/accept-conditions/accept-conditions.component.html
```diff
@@ -1,8 +1,9 @@
-<div fxLayout="row" fxLayoutAlign="center" fxLayoutGap="10px">
-  <mat-checkbox test-id="checkbox" (change)="toggleTermsOfUse()"></mat-checkbox>
-  <span>I accept the
-    <a (click)="openTermsOfUse(displayConditions)" class="terms-link">Terms and Conditions of Use</a>.
-  </span>
+<div fxLayout="row" fxLayoutAlign="center">
+  <mat-checkbox test-id="checkbox" (change)="toggleTermsOfUse()" color="primary">
+    <span>I accept the
+      <a mat-button (click)="openTermsOfUse(displayConditions)" class="terms-link">Terms and Conditions of Use</a>.
```
GrandSchtroumpf, line 4: Because `a` is a mat-button, I would recommend puttin closing the span after the "I accept the"

styles/components/mat-button.scss
```diff
@@ -76,26 +96,36 @@
...
+  .mat-fab,
+  .mat-mini-fab {
     @include _mat-theme-overridable-elevation(0, $theme);
     margin-bottom: 8px;
   }
 
+  .mat-fab {
+    &.active {
+      opacity: 0.5;
+      background-color: #606f87;
```
GrandSchtroumpf, line 108: Does this color already exist in the theme ? Would be much better for consistency.

styles/components/mat-button.scss
```diff
@@ -76,26 +96,36 @@
...
-  .mat-fab, .mat-mini-fab {
+  .mat-fab,
+  .mat-mini-fab {
     @include _mat-theme-overridable-elevation(0, $theme);
     margin-bottom: 8px;
   }
 
+  .mat-fab {
+    &.active {
+      opacity: 0.5;
```
GrandSchtroumpf, line 107: Do you really want to lower the opacity of everything inside the mat-fab ? Icon and all ?
Else apply transparency on the background-color only.

styles/components/mat-button.scss
```diff
@@ -76,26 +96,36 @@
...
   }
 
-  .mat-fab, .mat-mini-fab {
+  .mat-fab,
+  .mat-mini-fab {
     @include _mat-theme-overridable-elevation(0, $theme);
     margin-bottom: 8px;
   }
 
+  .mat-fab {
```
GrandSchtroumpf, line 105: Only mat-fab ? not mat-mini-fab ?

styles/style.scss
```diff
@@ -67,6 +67,7 @@ a {
 .terms-link {
```
GrandSchtroumpf, line 67: Thi should not be in the style.scss. It should be move to the component stylesheet.
Also, as you're using mat-button, I think some of the code here is deprecated.

PR #2947, opened 2020-06-02
Merge back hotfixes 1.4.1

lsenta:


apps/festival/festival/src/app/marketplace/marketplace.component.html
```diff
@@ -3,30 +3,40 @@
...
+        <span mat-line><b>My Wishlist</b></span>
       </a>
     </mat-nav-list>
   </aside>
 
+<<<<<<< HEAD
   <header fxFlex fxLayout fxLayoutAlign="end end" class="marketplace-header">
     <a mat-icon-button test-id="heart-icon" routerLink="event/calendar">
       <mat-icon svgIcon="calendar"></mat-icon>
+=======
```
GrandSchtroumpf, line 35: Can you accept incoming value ?

lsenta: Thanks for the catch, my vscode doesn't autosave yet.

PR #2949, opened 2020-06-02
Invitation css

Hugo-Hontans:


libs/notification/src/lib/invitation/components/list/list.component.scss
```diff
@@ -1,4 +1,12 @@
...
-  padding: 24px;
-}
\ No newline at end of file
+}
+
+p {
+  margin-bottom: 8px;
+}
+
+.divider {
```
Mussuranae, line 9: You can do :
```
.divider {
  margin: 16px 0;
}
```
to apply margin on top and bottom in the same time.

PR #2953, opened 2020-06-02
design review from mathilde

fritzschoff:


apps/festival/festival/src/app/marketplace/home/home.component.html
```diff
@@ -26,9 +26,9 @@ <h1>{{ section.title }}</h1>
           </ng-container>
         </bf-carousel>
         <ng-container *ngIf="i === 0">
-          <div fxLayout="row" fxLayoutAlign="space-between start">
-            <h1 class="margin-title">Featured films</h1>
-            <a mat-button routerLink="../title">
+          <div fxLayout="row" fxLayoutAlign="space-between start" flex class="movie-header-slider">
+            <h1>Featured films</h1>
+            <a mat-button routerLink="../title" class="active" color="primary">
```
GrandSchtroumpf, line 31: How is this button active ? The only reason it would be active is if the the user is on the "../title" page, which cannot be the case. There is a problem in the semantic here. For me there is no reason to apply the same "active" style on this button than on other "a" navigation.

PR #2958, opened 2020-06-03
Feat 2920 - Be able to import orgs from excel in admin panel

phpgeekfr:
Feat 2920 - Be able to import orgs from excel in admin panel


libs/import/src/lib/components/view-extracted-elements/organizations/view-extracted-organizations.component.html
```diff
@@ -0,0 +1,24 @@
+<article>
+  <h1>Import Organizations</h1>
+  <section class="help-section" fxLayout="column" fxLayoutAlign="center">
+    <p>Please select the organizations and admins to create.</p>
+    <a [routerLink]="['./']" (click)="openIntercom()">Need help ?</a>
```
GrandSchtroumpf, line 5: use `routerLink="./"`
But I don't understand what it does...

libs/import/src/lib/components/view-extracted-elements/organizations/view-extracted-organizations.component.ts
```diff
@@ -0,0 +1,312 @@
...
+        ///////////////
+        // VALIDATION
+        ///////////////
+
+
+        const orgWithErrors = await this.validate(importErrors);
+
+        if (!orgWithErrors.newOrg) {
+          this.orgsToUpdate.data.push(orgWithErrors);
+          this.orgsToUpdate.data = [... this.orgsToUpdate.data];
```
GrandSchtroumpf, line 217: Why do you need to do that ? as you're doing markForCheck it should be ok (I guess 🤔)

libs/import/src/lib/components/view-extracted-elements/rights/view-extracted-rights.component.scss
```diff
@@ -0,0 +1,14 @@
+article {
+    padding: 20px;
+    min-height: calc(100vh - (var(--toolbarHeight) * 2) - 30px); // TODO #1212
```
GrandSchtroumpf, line 3: I think you can remove at least 1 --toolbarHeight

libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -69,11 +69,10 @@ export class OrganizationService extends CollectionService<OrganizationState> {
    * create related documents (permissions, apps permissions, user...).
   */
   async onCreate(org: Organization, { write }: WriteOptions) {
-    const user = this.authQuery.user;
     const orgId: string = org.id;
     const permissions = createPermissions({
       id: orgId,
-      roles: { [user.uid]: 'superAdmin' }
+      roles: { [org.userIds[0]]: 'superAdmin' }
```
GrandSchtroumpf, line 75: Why ?

Issue #2655, opened 2020-04-29
Implement image-set directive to load the right size of image in the component


ioaNikas:
[Ressource](https://css-tricks.com/responsive-images-css/)

GrandSchtroumpf:
## Current state
There is 4 directives that deal with images : 
- `ImgAsset`: Load local image asset into a `img` tag depending on the current theme
- `BackgroundAsset`: Set background with local asset depending on the current theme
- `ImgReference`: Load Firestorage reference image into an `img` tag (+ fallback on local asset)
- `BackgroundRef`: Same but for background image.

> This issue focuses on the `ImgReference`, but should lead to a unified version to deal with one or two directive max in the end.

Currently the `ImageReference` directive get the original image from Firestorage. Instead it should use the `srcset` property to load the right size.

## Use ImgRef object
Each document stores the reference to their images through the `ImgRef` object interface. It provided several sizes (generated by a backend function): original / xs / md / lg.
The goal it to update the `ImageReference` directive to use `srcset` instead of `src`. 
For `srcset` to work we need to specify the size of the image we are using, this depends on the "type" of the image.

Here is how it should looks like in the end : 
`<img [imgRef]="profile" asset="empty_profile" size="xs" />`

This means : 
- If profile existes, check it's type (logo, avatar, banner, poster, still) & the size to get the width
- Set `srcset` with url & width of the image
- Else, use the asset in the local project `empty_profile.webp`, depending on the current theme.

## Future
In this end we should have only two directive to deal with images : 
`<img [imgRef]="profile" asset="empty_profile" size="xs" />`
& 
`<div [bgRef]="banner" asset="empty_banner" size="lg"></div>`

If we want to display an asset (and not a reference), we should only write : 
`<img asset="empty_profile" size="xs" />`
Currently local images are not resized, but this is a future work we'll need to do

PR #2959, opened 2020-06-03
APP-2655 Added srcset attribute to the image tag. We need to filter t…

joanofdart:
…he original one from the other sizes because if not, it will add another line with a 0w.

- We need to check if path.ref has any value, to avoid decriptor errors.
- We only add srcset to those who have a ref, else we avoid adding it to the img element.

#2655 

image

As you can see in the image, the browser is using the "lg" version of the image.
🎉


libs/utils/src/lib/media/media.firestore.ts
```diff
@@ -24,3 +24,20 @@ export function createImgRef(ref: Partial<ImgRef> | string = {}): ImgRef {
...
+  } else if (url.includes('logo')) {
+    return { original: 0, xs: 50, md: 100, lg: 300 };
+  } else if (url.includes('poster')) {
+    return { original: 0, xs: 200, md: 400, lg: 600 };
+  } else if (url.includes('banner')) {
+    return { original: 0, xs: 300, md: 600, lg: 1200 };
+  } else if (url.includes('still')) {
+    return { original: 0, xs: 50, md: 100, lg: 200 };
+  } else {
+    console.warn('No bucket directory, exiting function');
```
ioaNikas, line 40: Here you can throw an error so our error logger can directly show to the dev the error in a snackbar

PR #2966, opened 2020-06-03
Privacy policy

pldespaigne:
- [x] add privacy policy
- [x] add tick box for it on both pages


libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.scss
```diff
@@ -0,0 +1,31 @@
...
+
+header {
+  padding: 100px 0;
+  text-align: center;
+}
+
+mat-card {
+  margin-top: 50px;
+  margin-bottom: 50px;
+  width: 1000px;
```
GrandSchtroumpf, line 11: Could you set a % instead ? What happen on 4k screen for example ?

libs/user/src/lib/auth/components/signup-form/signup-form.component.scss
```diff
@@ -22,6 +22,12 @@ mat-checkbox {
   margin-right: 5px;
 }
 
+.terms-link {
```
GrandSchtroumpf, line 25: I don't see where this class is used.

libs/user/src/lib/auth/pages/identity/identity.component.html
```diff
@@ -29,16 +29,29 @@ <h1>Join {{ appName }}</h1>
         </mat-form-field>
       </ng-container>
       <password-confirm formControlName="newPassword"></password-confirm>
-      <auth-accept-conditions (termsOfUseChecked)="checkTermsOfUse($event)">
-        <ng-template authConditions>
-          <auth-terms-conditions></auth-terms-conditions>
-        </ng-template>
-      </auth-accept-conditions>
+      <div>
```
GrandSchtroumpf, line 32: You don't need those div, you can remove them.

Issue #2782, opened 2020-05-18
Debug - festival / Archipel Market


Jackseed:
This issue will gather on comments all the bugs to fix on `festival`.

maltmayer:
MARKETPLACE > TITLE LIST
- [ ] "Sort by" doesn't work

maltmayer:
MARKETPLACE > SALES AGENT PAGE > LINE UP

- [ ] the directors don't appear on the movie cards in the line-up
![image](https://user-images.githubusercontent.com/55785823/82302837-696ded00-99ba-11ea-97ce-bc32692c30bd.png)

MARKETPLACE > ORG PAGE 

- [ ] the org cards don't all have the same size 
![image](https://user-images.githubusercontent.com/55785823/82302927-8efaf680-99ba-11ea-9119-e2c695ab14a1.png)

maltmayer:
- [ ] My Wishlist should be in the top bar

- [ ] The icons in the top bar (calendar / invitations / notifications) have weird proportions. They should have the same space between one another. And is it possible make them the same size ?? (calendar icon is 2 times bigger thant invitations)

![image](https://user-images.githubusercontent.com/55785823/82304255-607e1b00-99bc-11ea-9c5d-05c7a6634e91.png)

- [ ] My Calendar should be in the side menu 
![image](https://user-images.githubusercontent.com/55785823/82304284-6c69dd00-99bc-11ea-9639-da8d288bb098.png)

maltmayer:
- [ ] Wishlist doesn't work

maltmayer:
DASHBOARD NO TITLE

- [ ] wrong "import your titles in bulk"  logo for dark theme

- [ ] put period points at the end of sentences. 

- [ ] Correct text : Title : "You have no title yet." / after that : "You haven't imported titles yet. You may add several titles at once through bulk import, or just one at a time."

![image](https://user-images.githubusercontent.com/55785823/82425227-83750180-9a86-11ea-842b-06f6ea9a7726.png)

maltmayer:
- [ ] Here my event is over and I see that. If there is no statistic to show, then there should either be an illustration saying there is no statistic to show, or juste delete this area.

![image](https://user-images.githubusercontent.com/55785823/82425818-59700f00-9a87-11ea-9477-61e37f58cd3c.png)

maltmayer:
- [ ] Archipel Market logo is blurred. change to better quality
[...]

PR #2967, opened 2020-06-03
APP-2782 Fixed an issue where resizeObserver was non existent on Safari

joanofdart:
Related to #2782 
* ResizeObserver issue.


libs/ui/src/lib/layout/layout.module.ts
```diff
@@ -35,22 +35,19 @@ export class Layout implements OnInit, OnDestroy {
...
-  ngOnInit() {
+  async ngOnInit() {
     const el = this.el.nativeElement;
-    if (ResizeObserver) {
-      this.observer = new ResizeObserver(([entry]) => this.width.next(entry.contentRect.width));
-      this.observer.observe(el);
-    } else {
-      // do something with window:resize
-    }
+    const resizeObs = (ResizeObserver in window) ? (await import('@juggle/resize-observer')).ResizeObserver : ResizeObserver;
```
GrandSchtroumpf, line 40:
```typescript
if (!window) {
  return;
}
const resizeObs = ResizeObserver in window) ? ...
```

libs/utils/src/lib/media/media.firestore.ts
```diff
@@ -37,7 +37,6 @@ export function getImgSize(url: string) {
   } else if (url.includes('still')) {
     return { original: 0, xs: 50, md: 100, lg: 200 };
   } else {
-    console.warn('No bucket directory, exiting function');
-    return false;
+    throw new Error('No bucket directory, exiting function');
```
phpgeekfr, line 40: This message does not seems very relevant

libs/ui/src/lib/layout/layout.module.ts
```diff
@@ -35,22 +35,19 @@ export class Layout implements OnInit, OnDestroy {
...
+    this.observer = new resizeObs(([entry]) => this.width.next(entry.contentRect.width));
+    this.observer.observe(el);
```
phpgeekfr, line 42: LGTM but wait for @GrandSchtroumpf point of view

libs/utils/src/lib/media/media.firestore.ts
joanofdart: Any idea on what message should be there instead? :)

libs/ui/src/lib/layout/layout.module.ts
joanofdart: LGTM? ahh looks good to me?

libs/utils/src/lib/media/media.firestore.ts
joanofdart: @phpgeekfr

PR #2968, opened 2020-06-03
fix create-screening e2e test

Hugo-Hontans:


libs/e2e/src/lib/utils/functions.ts
```diff
@@ -47,3 +47,9 @@ export function formatAmPm(date: Date, toLowerCase?: boolean) {
   hours = hours ? hours : 12; // the hour 0 should be 12
   return toLowerCase ? hours + ' ' + amPM.toLowerCase() : hours + ' ' + amPM;
 }
+
+export function getTomorrowDay(date: Date) {
+  const days = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
```
ioaNikas, line 52: Unused, you can delete.

PR #2979, opened 2020-06-04
cookie banner

pldespaigne:
- [x] add a cookie banner to the `app.component`
- [x] add simple accepted cookie to know if we should display the banner or not

### DIFF
--- a/libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.scss
+++ b/libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.scss
@@ -1,0 +1,8 @@
+:host {
+  position: fixed;
+  bottom: 0;
+  z-index: 1;
+  a {
+    text-decoration: underline;
+  }
+}


libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.html
```diff
@@ -0,0 +1,11 @@
+<mat-card *ngIf="!hasAccepted">
+  <mat-card-content fxLayout="row" fxLayout.lt-md="column" fxLayoutAlign="center center">
+    <p fxFlex="shrink">By continuing to browse this site, you are agreeing with our
+      <a href="javascript:void(0)" (click)="openPrivacyPolicy()">privacy policy</a>
+      and the use of cookies,
+      whose purpose is to provide personalize content, deliver our services and monitor analytics.
+      You can change your cookie preferences at any time.</p>
+    <button fxFlex="1 0 auto" mat-button>Change preferences</button>
```
GrandSchtroumpf, line 8: Why do you need to apply the fxFlex ? The normal style is not good enough ?

Issue #2951, opened 2020-06-02
Admin email notification when first connexion of a user


Jackseed:
Add an email notification to team@cascade8.com when first connexion of a user
> internal email so easy on text, ask @maltmayer if needed

Jackseed:
@lsenta : 
```onUserUpdate: if before.firstName === undefined || before.firstName === "" AND after.firstname !== before.firstname { BOOM, user just filed their info }```

Issue #2826, opened 2020-05-20
Set the from part for the mail sent in `onUserCreate` backend trigger


phpgeekfr:
Currently this mail is sent by `team@cascade8.com` but we can update the account process creation to retrieve the application used for account creation and set the correct "from" part.

PR #2985, opened 2020-06-04
 #2951 Admin email notification when first connexion of a user 

phpgeekfr:


apps/backend-functions/src/templates/mail.ts
```diff
@@ -164,6 +165,17 @@ const organizationRequestAccessToAppTemplate = (orgId: string, appUrl: string =
   Visit ${appUrl}${ADMIN_ACCEPT_ORG_PATH}/${orgId} or go to ${ADMIN_ACCEPT_ORG_PATH}/${orgId} to enable it.
   `;
 
+
+/**
+ * @param user
+ */
+const userFirstConnexionTemplate = (user: PublicUser) =>
+  `
+  User ${user.firstName} ${user.lastName} connected for the first time to the app.
```
Mussuranae, line 174: Maybe add on which application he connected for the first time ?

phpgeekfr: I wanted to do so, but we don't have this information yet. Will add a comment with #2826

PR #2988, opened 2020-06-04
Img cropper fix master

phpgeekfr:


libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
```diff
@@ -97,6 +97,16 @@ export class CropperComponent implements OnDestroy {
...
   constructor(private storage: AngularFireStorage, private _renderer: Renderer2, private _elementRef: ElementRef) { }
 
+  ngOnInit() {
+    this.form.valueChanges.pipe(
+      startWith(this.form.value),
+      filter(imgRef => imgRef.ref)
+    ).subscribe(imgRef => {
+      this.previewUrl = of(imgRef.urls.original);
+      this.step.next('show');
+    });
```
GrandSchtroumpf, line 107: We should unsubscribe here : 
```typescript
const sub = this.for.valueChanges.pipe(....);
this.sub.add(sub);
```

phpgeekfr: @GrandSchtroumpf ok, can you check the new commit ?

PR #2989, opened 2020-06-05
Implemented footer in Dashboard and Marketplace Layout

RemcoSimonides:
- [x] Implement footer in Marketplace and Dashboard layout

### DIFF
--- a/libs/ui/src/lib/layout/dashboard/dashboard.component.html
+++ b/libs/ui/src/lib/layout/dashboard/dashboard.component.html
@@ -50,6 +50,7 @@
     <!-- CONTENT -->
       <main cdkScrollable fxFlex>
         <router-outlet></router-outlet>
+        <bf-footer></bf-footer>
       </main>
   </mat-sidenav-content>
 </mat-sidenav-container>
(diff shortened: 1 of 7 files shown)


libs/ui/src/lib/layout/dashboard/dashboard.component.html
```diff
@@ -50,6 +50,7 @@
     <!-- CONTENT -->
       <main cdkScrollable fxFlex>
         <router-outlet></router-outlet>
+        <bf-footer></bf-footer>
```
GrandSchtroumpf, line 53: Did you check with @2amarkina if we want a footer on dashboard ?

libs/ui/src/lib/layout/footer/footer.component.scss
```diff
@@ -5,4 +5,5 @@
   flex-wrap: wrap;
   justify-content: space-between;
   align-items: center;
+  margin-top: auto;
```
GrandSchtroumpf, line 8: I don't understand what this is suppose to do. Are you sure it's usefull style ?

PR #3003, opened 2020-06-08
💣 DO NOT MERGE 💣 - Upload and delete image executed on update profile

RemcoSimonides:
- [x] Uploading and deleting image is now done when profile is updated instead of when the image is cropped

### DIFF
--- a/apps/catalog/catalog/src/app/dashboard/movie-tunnel/movie-tunnel.component.ts
+++ b/apps/catalog/catalog/src/app/dashboard/movie-tunnel/movie-tunnel.component.ts
@@ -1,5 +1,6 @@
 // Angular
 import { Component, ChangeDetectionStrategy, OnInit } from '@angular/core';
+import { AngularFireStorage } from '@angular/fire/storage';
 // Blockframes
 import { MovieService, MovieQuery, Movie } from '@blockframes/movie/+state';
 import { MovieForm } from '@blockframes/movie/form/movie.form';
@@ -9,6 +10,7 @@ import { TunnelStep, TunnelRoot, TunnelConfirmComponent } from '@blockframes/ui/
 import { switchMap } from 'rxjs/operators';
 import { of } from 'rxjs';
 import { mergeDeep } from '@blockframes/utils/helpers';
+import { MoviePromotionalElementsForm } from '@blockframes/movie/form/promotional-elements/promotional-elements.form';
 
 const steps: TunnelStep[] = [{
   title: 'Title Information',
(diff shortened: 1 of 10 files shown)


apps/catalog/catalog/src/app/dashboard/movie-tunnel/movie-tunnel.component.ts
```diff
@@ -111,4 +113,54 @@ export class MovieTunnelComponent implements TunnelRoot, OnInit {
       switchMap(shouldSave => shouldSave ? this.save() : of(false))
     );
   }
+
+  private async handleUpdatePromotionalImages(moviePromotionalElementsForm: MoviePromotionalElementsForm, promotionalElements) {
```
GrandSchtroumpf, line 117: All of this should be managed by the MediaService. You shouldn't do that here.
You should have something like : 
```typescript
const value = this.form.value;
const medias = ['promotionalElements.still_photo', 'promotionalElements.banner', ...]; // this should be exported by the movie lib so we don't have to write it several time
this.media.upload(value, medias);  // Upload your files
const updates = removePaths(value, medias); // Remove files path from update
const movie  = mergeDeep(this.query.getActive(), this.form.value); // deepmerge with current value
await this.service.update(movie); // Update everything except media fields.
```

apps/festival/festival/src/app/dashboard/tunnel/media-image/media-image.component.html
```diff
@@ -48,7 +48,7 @@ <h3 type>Poster</h3>
         <ng-container *ngFor="let control of stillPhoto.controls; let i = index; let last = last">
           <span class="still mat-body-2">Still # {{i + 1}}</span>
           <drop-cropper
-            [storagePath]="stillPath"
+            [storagePath]="'movies/' + movie.id + '/promotionalElements.still_photo.' + i + '.media'"
```
GrandSchtroumpf, line 51: Same as above

apps/festival/festival/src/app/dashboard/tunnel/movie-tunnel.component.ts
```diff
@@ -100,4 +102,54 @@ export class MovieTunnelComponent implements TunnelRoot, OnInit {
       switchMap(shouldSave => shouldSave ? this.save() : of(false))
     );
   }
+
+  private async handleUpdatePromotionalImages(moviePromotionalElementsForm: MoviePromotionalElementsForm, promotionalElements) {
```
GrandSchtroumpf, line 106: Same as above

libs/ui/src/lib/media/cropper/cropper/cropper.component.html
```diff
@@ -42,7 +42,7 @@
   <article class="show" *ngSwitchCase="'show'" @finalZoom>
     <ng-container>
       <div class="show">
-        <img class="croppedImage" [src]="previewUrl | async" />
+        <img class="croppedImage" [src]="croppedImage || (previewUrl | async)" />
```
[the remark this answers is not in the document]
RemcoSimonides: There is no previewUrl until it has been uploaded to storage.

PR #3005, opened 2020-06-08
APP-2782 Making sure we show either TBD or the time.

joanofdart:
- Fixed a typo.

Related to #2782


joanofdart:
Added **My Dashboard** link which should only be seen by an org who has access to Festival and Dashboard :)

image

apps/festival/festival/src/app/marketplace/marketplace.component.html
```diff
@@ -25,6 +25,13 @@
         <mat-icon mat-list-icon svgIcon="heart_fill"></mat-icon>
         <span mat-line><b>My Wishlist</b></span>
       </a>
+      <ng-container *ngIf="canAccessDashboard(org$ | async)">
```
GrandSchtroumpf, line 28: Never use method for an input prefere pipes. I'm pretty sure there a pipe like that in the code (you can ask Bruce or PL if you don't find it): 
```html
<ng-container *ngIf="org$ | async | canAccess: 'dashboard'">
```
We use pipes because, if they are pure, they do not reload the part of the code on every change detection.

apps/festival/festival/src/app/marketplace/marketplace.component.ts
```diff
@@ -1,9 +1,19 @@
...
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class MarketplaceComponent {}
+export class MarketplaceComponent {
+
+  public org$ = this.orgQuery.selectActive();
+
+  constructor(private orgQuery: OrganizationQuery) { }
+
+  canAccessDashboard(org: OrganizationDocumentWithDates): boolean {
```
GrandSchtroumpf, line 16: remove this method in favor of pipe.

libs/movie/src/lib/movie/components/title-features/title-features.component.ts
```diff
@@ -11,7 +11,7 @@ interface TitleFeature {
 
 function createTitleFeatureView(movie: Movie): TitleFeature[] {
     const convertedGenres = movie.main?.genres.map(genre => getLabelBySlug('GENRES', genre));
-    const convertedRunTime = `${movie.main?.totalRunTime} min`;
+    const convertedRunTime = typeof movie.main?.totalRunTime === 'string' ? 'TBD' : `${movie.main?.totalRunTime} min`;
```
GrandSchtroumpf, line 14: If it's a string what would be the string ?
Also you can refactor like that too make it more readable : 
```typescript
if (movie.main) {
  const { genres, totalRunTime, originalLanguages, originCountries, status } = movie.main;
  const convertedGenres = genres.map(genre => getLabelBySlug('GENRES', genre));
  const convertedRunTime = typeof totalRunTime === "string" ? totalRunTime : `${totalRunTime}min`;
  ...
} else {
  return '';
}

```

joanofdart: Done.

apps/festival/festival/src/app/marketplace/marketplace.component.ts
joanofdart: Amazing, thanks. Done :)

apps/festival/festival/src/app/marketplace/marketplace.component.html
joanofdart: Gotchu! Done

apps/festival/festival/src/app/marketplace/marketplace.component.html
```diff
@@ -25,6 +25,13 @@
         <mat-icon mat-list-icon svgIcon="heart_fill"></mat-icon>
         <span mat-line><b>My Wishlist</b></span>
       </a>
+      <ng-container *ngIf="org$ | async | canAccess: 'dashboard'">
```
GrandSchtroumpf, line 28: Modify it also in the dashboard.component.html

joanofdart: Done!

PR #3007, opened 2020-06-08
gdpr banner & form 🍪

pldespaigne:
On going implementation of GDPR cookie banner & form
- cookie banner works but with a simple "accept" button
- cookie form is only html for now

- [ ] Add cookie preferences to profile page
- [ ] add to catalog and financiers


apps/festival/festival/src/app/app.component.ts
```diff
@@ -2,11 +2,10 @@ import { Component, ChangeDetectionStrategy } from '@angular/core';
 import { ThemeService } from '@blockframes/ui/theme';
 import { TunnelService } from '@blockframes/ui/tunnel';
 import { IconService } from '@blockframes/ui/icon-service';
-import { YandexMetricaService } from '@blockframes/utils/yandex-metrica/yandex-metrica.service';
 
 @Component({
   selector: 'festival-root',
-  template: '<router-outlet></router-outlet><safari-banner><safari-banner>',
+  template: '<router-outlet></router-outlet><cookie-banner></cookie-banner><safari-banner><safari-banner>',
```
GrandSchtroumpf, line 8: So we've got two banner ... ? @pldespaigne should we remove the safari banner now ?

apps/festival/festival/src/app/app.module.ts
```diff
@@ -79,22 +82,45 @@ import { SafariBannerModule } from '@blockframes/utils/safari-banner/safari-bann
...
-    const navEnds = this.router.events.pipe(filter(event => event instanceof NavigationEnd));
+  constructor(
+    router: Router,
+    analytics: FireAnalytics,
+    intercom: IntercomService,
+    yandex: YandexMetricaService,
+    @Inject(YM_CONFIG) ymConfig: number
+  ) {
+
+    const googleCookieAccepted = localStorage.getItem('c8-gdpr-google-analytics');
```
GrandSchtroumpf, line 100: localStorage is a slow process, don't use it several time. Store an object with the keys instead.
No need to specify "c8", localstorage is specific per page :
```typescript
const { googleAnalytics, intercom, } = localStorage.getItem('gdpr');
```
also use camelCase for the keys

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.html
```diff
@@ -0,0 +1,9 @@
+<mat-card *ngIf="!hasAccepted">
+  <mat-card-content fxLayout="row" fxLayout.lt-md="column" fxLayoutAlign="center center">
+    <p fxFlex="shrink">By continuing to browse this site, you are agreeing with the use of cookies,
+      whose purpose is to provide personalize content, deliver our services and monitor analytics.
+      You can change your cookie preferences at any time.</p>
+    <button fxFlex="1 0 auto" mat-button (click)="changePreferences()">Change preferences</button>
```
GrandSchtroumpf, line 6: no need to specify 1 0 auto. Just `fxFlex` works

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.ts
```diff
@@ -0,0 +1,85 @@
...
+    const cookieNames = cookies.map(cookie => cookie.split('=')[0].trim());
+    this.hasAccepted = cookieNames.some(cookieName => cookieName === 'archipel');
+  }
+
+  /** Opens a dialog with terms of use and privacy policy given by the parent. */
+  public openPrivacyPolicy() {
+    this.dialog.open(PrivacyPolicyComponent, { maxHeight: '80vh' })
+  }
+
+  public userAcceptedCookies() {
```
GrandSchtroumpf, line 46: Rename to `acceptCookies`

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.ts
```diff
@@ -0,0 +1,85 @@
...
+  ngOnInit() {
+    const cookieString = this.document.cookie;
+    const cookies = cookieString.split(';');
+    const cookieNames = cookies.map(cookie => cookie.split('=')[0].trim());
+    this.hasAccepted = cookieNames.some(cookieName => cookieName === 'archipel');
+  }
+
+  /** Opens a dialog with terms of use and privacy policy given by the parent. */
+  public openPrivacyPolicy() {
+    this.dialog.open(PrivacyPolicyComponent, { maxHeight: '80vh' })
```
GrandSchtroumpf, line 43: Also set a maxWidth

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.ts
```diff
@@ -0,0 +1,85 @@
...
+    private intercom: IntercomService,
+    private yandex: YandexMetricaService,
+    @Inject(YM_CONFIG) private ymConfig: number,
+    @Inject(DOCUMENT) private document: Document
+  ) {}
+
+  ngOnInit() {
+    const cookieString = this.document.cookie;
+    const cookies = cookieString.split(';');
+    const cookieNames = cookies.map(cookie => cookie.split('=')[0].trim());
```
GrandSchtroumpf, line 37: You might have a "cannot read trim of undefined" here if there is no first item.
Use `shift` instead: `cookies.map(cookie => cookie.split('=').shift()?.trim());`

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.hasAccepted = cookieNames.some(cookieName => cookieName === 'archipel');
```
GrandSchtroumpf, line 38: Will this work for every app ? As media financiers doesn't seem to be an archipel app... I don't know.

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.ts
```diff
@@ -0,0 +1,85 @@
...
+        this.confirmCookies();
+        if (settings.google) this.enableAnalytics();
+        if (settings.intercom) this.enableIntercom();
+        if (settings.yandex) this.enableYandex();
+      }
+    })
+  }
+
+  confirmCookies() {
+    this.document.cookie = 'archipel=';
```
GrandSchtroumpf, line 66: Maybe use "blockframes" instead of archipel.

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.html
```diff
@@ -0,0 +1,61 @@
...
+      <h3>CASCADE 8, publisher of ARCHIPEL MARKET</h3>
+      <p>We value your privacy.</p>
+
+      <p>You can set your consent preferences and determine how you want your data to be used based on the purposes below.
+      You may set your preferences for us independently from those of third-party partners.
+      Each purpose has a description so that you know how we and partners use your data.</p>
+    </section>
+
+    <section>
+      <mat-divider></mat-divider>
```
GrandSchtroumpf, line 14: Put mat-divider between the section.

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.html
```diff
@@ -0,0 +1,61 @@
...
+      </div>
+    </section>
+
+    <section fxLayout="row" fxLayoutAlign="center center">
+      <button mat-flat-button color="primary" (click)="save()">Save and quit</button>
+    </section>
+
+    <mat-divider></mat-divider>
+
+    <a href="javascript:void(0)" class="mat-caption" (click)="openPrivacyPolicy()">Privacy Policy</a>
```
GrandSchtroumpf, line 58: Use button instead of "a"

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.scss
```diff
@@ -0,0 +1,15 @@
...
+  mat-card-content {
+    a {
+      text-decoration: underline;
+    }
+    mat-radio-button {
+      margin: 16px 32px 16px 0px;
+    }
+    mat-divider {
+      margin: 16px;
+      position: inherit;
```
GrandSchtroumpf, line 11: remove "position" you don't need it.

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.scss
```diff
@@ -0,0 +1,15 @@
+mat-card {
+  mat-card-content {
+    a {
+      text-decoration: underline;
+    }
+    mat-radio-button {
+      margin: 16px 32px 16px 0px;
```
GrandSchtroumpf, line 7: Organize your code so you don't have unsymetrics margins.

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.scss
```diff
@@ -0,0 +1,15 @@
...
+    }
+    mat-divider {
+      margin: 16px;
+      position: inherit;
+      width: auto;
```
GrandSchtroumpf, line 12: this is the default value. don't specify it.

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  form = new CookiesConsentForm();
+
+  constructor(
+    private dialog: MatDialog,
+    private dialogRef: MatDialogRef<any>,
+  ) { }
+
+  masterToggle(event: MatRadioChange) {
+    Object.keys(this.form.controls)
```
GrandSchtroumpf, line 24: Cleaner : 
```typescript
for (const key in this.form.controls) {
  this.form.get(key).setValue(event.value);
}
```

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Component({
+  selector: 'cookie-form',
+  templateUrl: './cookie-form.component.html',
+  styleUrls: ['./cookie-form.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class CookieFormComponent {
+
+  form = new CookiesConsentForm();
```
GrandSchtroumpf, line 16: For small form like that use a simple FormGroup. We create specific forms when we need typescript autocompletion and type checking. But here it's more a burden than an help : 
```typescript
form = new FormGroup({
    google: new FormControl(null, Validators.required),
    intercom: new FormControl(null, Validators.required),
    yandex: new FormControl(null, Validators.required),
})
```

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie.form.ts
```diff
@@ -0,0 +1,34 @@
+import { FormControl, Validators } from '@angular/forms';
+import { EntityControl, FormEntity } from '@blockframes/utils/form/forms/entity.form';
+
+export interface CookiesConsent {
+  google: boolean,
+  intercom: boolean,
+  yandex: boolean,
+}
+
+function createCookiesConsent(params?: Partial<CookiesConsent>): CookiesConsent {
```
GrandSchtroumpf, line 10: remove this file.

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.form.get(event.source.name as keyof CookiesConsent).setValue(event.checked);
+  }
+
+  save() {
+    this.dialogRef.close(this.form.value);    
+  }
+
+  /** Opens a dialog with terms of use and privacy policy given by the parent. */
+  public openPrivacyPolicy() {
+    this.dialog.open(PrivacyPolicyComponent, { maxHeight: '80vh' });
```
GrandSchtroumpf, line 38: set `maxWidth` too

libs/utils/src/lib/analytics/app-analytics.ts
```diff
@@ -3,18 +3,24 @@ import { Injectable } from '@angular/core';
 import { AuthQuery } from '@blockframes/auth/+state/auth.query';
 import { AnalyticsEvents } from './analyticsEvents';
 import { centralOrgID } from '@env';
+import { GDPRService } from '../gdpr-cookie/gdpr-service/gdpr-service';
 
 @Injectable({ providedIn: 'root' })
-export class FireAnalytics {
-  constructor(public analytics: AngularFireAnalytics, private authQuery: AuthQuery) { }
+export class FireAnalytics extends GDPRService {
```
GrandSchtroumpf, line 9: Don't extends. Inject your GdprService into the contructor instead.
Composition is usually better than inheritance.

libs/utils/src/lib/intercom/intercom.module.ts
```diff
@@ -0,0 +1,13 @@
+import { NgModule } from '@angular/core';
+import { IntercomModule } from 'ng-intercom';
+import { intercomId } from '@env';
+
+@NgModule({
+  imports: [
+    IntercomModule.forRoot({
+      appId: intercomId,
+      updateOnRouterChange: true // will automatically run `update` on router event changes. Default: `false`
```
GrandSchtroumpf, line 9: Might have a performance issue if it runs for each route change instead of only once...

libs/utils/src/lib/intercom/intercom.service.ts
```diff
@@ -0,0 +1,57 @@
...
+@Injectable({ providedIn: 'root' })
+export class IntercomService extends GDPRService {
+
+  constructor(
+    private query: AuthQuery,
+    public intercom: Intercom
+    ) {
+    super('c8-gdpr-intercom');
+    if (intercomId) {
+      query.user$.pipe(
```
GrandSchtroumpf, line 18: Never leave a long term subscription. You should never subscribe in a service. Always in a component

libs/utils/src/lib/intercom/intercom.service.ts
```diff
@@ -0,0 +1,57 @@
...
+        distinctUntilChanged((oldUser, newUser) => !!oldUser === !!newUser) // Trigger only when the user login/logout
+        ).subscribe(user => {
```
GrandSchtroumpf, line 20: This logic should be a method triggered by the login / logout methods of the auth service.
Using a observable here is counter efficient.

libs/utils/src/lib/intercom/intercom.service.ts
```diff
@@ -0,0 +1,57 @@
...
+    const user = this.query.user;
+    this.intercom.boot(this.getIntercomOptions(user));
+  }
+
+  disable() {
+    this.intercom.shutdown();
+  }
+
+  private getIntercomOptions(user: User) {
+    return user
```
GrandSchtroumpf, line 42: Cleaner : 
```typescript
const widget = { activator: "#intercom" };
if (user) {
  const { email, uid, firstName } = user;
  return { email, widget, user_uid: uid, name: firstName };
}
return { widget };
```

GrandSchtroumpf (changes requested):

apps/festival/festival/src/app/app.component.ts
RemcoSimonides: They are indeed on top of each other now;
image

I don't think it's a good idea to remove safari banner. Also because it may be even more important once we know how many users use old versions of Safari (https://github.com/blockframes/blockframes/issues/3580)

RemcoSimonides: Maybe it's an idea to place them on above each other, or to show the Safari banner underneath the cookie banner?

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.html
RemcoSimonides: image

apps/festival/festival/src/app/app.component.ts
pldespaigne: I thought Safari banner was made because of some weird behavior of the browser.
And we said something about it & Twillio during the visio meeting no ?

I guess it's better to keep it to be sure.

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.html
RemcoSimonides: I can't seem to figure out how I can make it look like the design when I use `button`. Do you know?
image

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.scss
RemcoSimonides: Position and width are actually useful because they overwrite other values. I don't know where they come from, or what to do against this. I had the same situation recently somewhere else. So if there is something I could do, I'd like to know :)
image

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.html
GrandSchtroumpf:
```html
<button mat-button ...>
```

RemcoSimonides: Not the same, but looks good!

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.ts
RemcoSimonides: I'll refactor this when I start on my next PR https://projects.invisionapp.com/d/main#/console/19607961/420299224/preview
For this one I also need to set the initial values with whatever is already saved

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -18,7 +23,9 @@ export class AuthService extends FireAuthService<AuthState> {
     private http: HttpClient,
     private query: AuthQuery,
     private functions: AngularFireFunctions,
-    private routerQuery: RouterQuery
+    private routerQuery: RouterQuery,
+    public ngIntercom: Intercom,
+    private gdprService: GDPRService
```
RemcoSimonides, line 28: I don't like to have to use `Intercom`, `GDPRService`, or `IntercomService` in AuthService....

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -121,4 +141,13 @@ export class AuthService extends FireAuthService<AuthState> {
       ip: ip
     }
   }
+
+  private updateIntercom(userCredential: UserCredential) {
+    const { intercom } = this.gdprService.cookieConsent;
+    if (!intercom || !intercomId) return;
+
+    this.db.doc<User>(`users/${userCredential.user.uid}`).valueChanges().pipe(take(1)).subscribe(user => {
+      this.ngIntercom.update(getIntercomOptions(user));
```
RemcoSimonides, line 150: `ngIntercom.update` should do the trick instead of rebooting intercom like we did before. Now we should even keep the conversation before and after sign in/up. I haven't tested yet... I'll warn Marie first ^^

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -59,14 +66,27 @@ export class AuthService extends FireAuthService<AuthState> {
...
    * @description function that gets triggered when
    * AuthService.signOut is called
    */
   async onSignout() {
+    // Keep cookieConsent in localStorage
+    const gdpr = localStorage.getItem('gdpr');
+
     localStorage.clear();
     sessionStorage.clear();
     window.location.reload();
```
RemcoSimonides, line 87: No need to reboot Intercom on signout because the page is reloaded on signout, so it's reinitiated anyway

libs/user/src/lib/auth/+state/auth.service.ts
RemcoSimonides: It does work!

apps/festival/festival/src/app/app.module.ts
```diff
@@ -79,22 +83,37 @@ import { SafariBannerModule } from '@blockframes/utils/safari-banner/safari-bann
...
+    analytics: FireAnalytics,
+    intercomService: IntercomService,
+    yandexService: YandexMetricaService,
+    gdprService: GDPRService,
+    @Inject(YM_CONFIG) ymConfig: number
+  ) {
+
+    const { googleAnalytics, intercom, yandex } = gdprService.cookieConsent;
+    if (!googleAnalytics) analytics.analytics.setAnalyticsCollectionEnabled(false);
+    !intercom ? intercomService.disable() : intercomService.enable();
```
GrandSchtroumpf, line 104: Do the opposite : `intercom ? enable() : disable()`
And just for the beauty of the code, put the ternary after the yandex if statement.

libs/user/src/lib/auth/+state/auth.service.ts
GrandSchtroumpf: why ?

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -59,14 +66,29 @@ export class AuthService extends FireAuthService<AuthState> {
...
+    // Keep cookieConsent in localStorage
+    const gdpr = localStorage.getItem('gdpr');
+
+    this.ngIntercom.shutdown();
+
     localStorage.clear();
     sessionStorage.clear();
     window.location.reload();
+
+    localStorage.setItem('gdpr', gdpr);
```
GrandSchtroumpf, line 91: I don't think this will do anything. When you reload the page the Javascript thread is shutted down. Also I don't understand why you need this line as you didn't mutated the gdpr object

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -59,14 +66,29 @@ export class AuthService extends FireAuthService<AuthState> {
...
+    this.updateIntercom(userCredential);
+  }
+
   /**
    * @description function that gets triggered when
    * AuthService.signOut is called
    */
   async onSignout() {
+    // Keep cookieConsent in localStorage
+    const gdpr = localStorage.getItem('gdpr');
```
GrandSchtroumpf, line 83: You're not using it. You should remove it.

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -121,4 +143,13 @@ export class AuthService extends FireAuthService<AuthState> {
       ip: ip
     }
   }
+
+  private updateIntercom(userCredential: UserCredential) {
+    const { intercom } = this.gdprService.cookieConsent;
+    if (!intercom || !intercomId) return;
+
+    this.db.doc<User>(`users/${userCredential.user.uid}`).valueChanges().pipe(take(1)).subscribe(user => {
```
GrandSchtroumpf, line 151: Do you really need the whole user for the intercom options or it would work with the user credential data ?

If you nede the whole user, use the `UserService` : 
```typesript
const user = await this.userService.getValue(this.user.uid);
const options = getIntercomOptions(user);
this.ngIntercom.update(options);
```

libs/utils/src/lib/gdpr-cookie/gdpr-service/gdpr-service.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from "@angular/core";
+import { CookiesConsent } from "../cookie-form/cookie.form";
+
+@Injectable({ providedIn: 'root' })
+export class GDPRService {
+
+  get cookieConsent(): CookiesConsent {
+    return JSON.parse(localStorage.getItem('gdpr')) ?? {};
```
GrandSchtroumpf, line 8: do `JSON.parse(localStorage.getItem('gdpr') ?? '{}');` instead.

libs/utils/src/lib/intercom/intercom.module.ts
```diff
@@ -0,0 +1,10 @@
+import { NgModule } from '@angular/core';
+import { IntercomModule } from 'ng-intercom';
+import { intercomId } from '@env';
+
+@NgModule({
+  imports: [
+    IntercomModule.forRoot({ appId: intercomId })
+  ],
+})
+export class IntercomAppModule {}
```
GrandSchtroumpf, line 10: I don't see the benefice of this module. Do you plan to add more stuff ? Else remove it

libs/utils/src/lib/intercom/intercom.service.ts
```diff
@@ -0,0 +1,33 @@
+import { Injectable } from '@angular/core';
+import { GDPRService } from '../gdpr-cookie/gdpr-service/gdpr-service';
+import { AuthQuery } from '@blockframes/auth/+state/auth.query';
+import { Intercom } from 'ng-intercom';
+import { User } from '@blockframes/user/types';
+
+@Injectable({ providedIn: 'root' })
+export class IntercomService {
```
GrandSchtroumpf, line 8: Couldn't you implement this service into the gdpr service instead of having two classes ? Looks overkill to me

GrandSchtroumpf (approved):

libs/user/src/lib/auth/+state/auth.service.ts
RemcoSimonides: Almost! I do have the uid and email in user credential data, but the displayName is empty. I could edit some code somewhere which also updates the displayName on credentialData when the User data is updated? Then I don't need to get the whole user.

Using the UserService in AuthService will give a circular dependency warning :/

libs/user/src/lib/auth/+state/auth.service.ts
RemcoSimonides: To prevent circular dependencies. It's not the case now, but I try to avoid using services in services because of this

libs/user/src/lib/auth/+state/auth.service.ts
RemcoSimonides: I've refactored it a little bit so the `localStorage.setItem('gdpr', gdpr)` is at least before `window.location.reload()`. But getting and setting the gdpr from localStorage is to prevent the settings to dissapear in the `localStorage.clear()` ;)

libs/utils/src/lib/intercom/intercom.service.ts
RemcoSimonides: Obviously this is possible; but I think their purpose are for different things. Whereas the IntercomService is to control the intercom pop-up on the website, and the GDPR service is just for the cookie settings. I think we should keep these separate from each other

PR #3013, opened 2020-06-09
merge back from master to develop

phpgeekfr:


apps/backend-functions/src/internals/image.ts
```diff
@@ -152,13 +153,25 @@ export async function onImageDeletion(data: functions.storage.ObjectMetadata) {
...
+  } catch (e) {
+    console.log(`Error while updating image references in document "${collection}/${id}" : ${e.message}`);
+  }
+
   // By filtering out the uploadedSize path, we make sure, that we don't try to delete an already deleted image
-  imgSizeDirectory.forEach(async (sizeDir : string) => {
+  imgSizeDirectory.forEach(async (sizeDir: string) => {
     const path = `${filePathElements[0]}/${filePathElements[1]}/${filePathElements[2]}/${sizeDir}/${filePathElements[4]}`;
     const exists = await bucket.file(path).exists();
     if (exists) {
```
pldespaigne, line 177: `file(path).exists()` return an array containing a boolean, therefore `const exists` will be either `[true]` or `[false]` and the if will only test that the array exists witch is always `true`.

Should be replaced by `const [exists] = await bucket.file(path).exists();` to work as intended

phpgeekfr: @pldespaigne will apply this update in another commit

PR #3014, opened 2020-06-09
fix analytics in EventReview Page

Hugo-Hontans:

### DIFF
--- a/apps/festival/festival/src/app/dashboard/event/review/review.component.html
+++ b/apps/festival/festival/src/app/dashboard/event/review/review.component.html
@@ -24,17 +24,16 @@
         </mat-card>
       </section>
     </ng-container>
-    <ng-template #noAnalytics>
-      <article class="no-analytics" fxLayout="column" fxLayoutAlign="center center">
-        <img asset="wait.webp" alt="waiting"/>
-        <p class="mat-body-2">There is no statistics available here yet.</p>
-      </article>
-    </ng-template>
   </ng-container>
+  <ng-template #noAnalytics>
+    <article class="no-analytics" fxLayout="column" fxLayoutAlign="center center">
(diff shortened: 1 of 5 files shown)


libs/event/src/lib/+state/event.service.ts
```diff
@@ -66,6 +67,18 @@ export class EventService extends CollectionService<EventState> {
...
+      const f = this.functions.httpsCallable('getEventAnalytics');
+      const analytics = await f({ eventIds: [eventId] }).toPromise();
+      const eventAnalytics = analytics.find(a => a.eventId === eventId);
+      this.analytics[eventId] = eventAnalytics;
+      return eventAnalytics;
+    }
+  }
+
   /** Gets analytics for one event and sync them. */
   public syncEventAnalytics() {
```
GrandSchtroumpf, line 83: You can remove that now.

apps/festival/festival/src/app/dashboard/event/review/review.component.html
```diff
@@ -1,31 +1,40 @@
...
+            showFilter
+            showPaginator
+            [initialColumns]="initialColumns"
+            [columns]="columns"
+            [source]="analytics">
+          </bf-table-filter>
+        </mat-card>
+      </section>
+    </ng-container>
+    <ng-template #noAnalytics>
```
GrandSchtroumpf, line 27: Put this outside of the ng-container it's more readable.
```html
<ng-container *ngIf="analytics$ | async as analytics; else loading">
  <ng-container *ngIf="analytics.length else noAnalytics">
    ...
  <ng-container>
<ng-container>
<ng-template #noAnalytics>...</ng-template>
<ng-template #loading>...</ng-template>
```

PR #3016, opened 2020-06-09
onDelete quick fix

pldespaigne:
apply some quick fix on the `onFileDelete` backend function

### DIFF
--- a/apps/backend-functions/src/internals/image.ts
+++ b/apps/backend-functions/src/internals/image.ts
@@ -145,7 +145,7 @@ async function resize(data: functions.storage.ObjectMetadata) {
 
 export async function onFileDeletion(data: functions.storage.ObjectMetadata) {
 
-  // TODO here we might need to handle deletion of other file types
+  // TODO here we might need to handle deletion of other file types (issue#3017)
 
   // If the type of the data is not an image, exit the function
   if (!data.contentType?.includes('image')) {
@@ -153,9 +153,7 @@ export async function onFileDeletion(data: functions.storage.ObjectMetadata) {
     return false;
   }
 
-  // we don't want to resize a vector image because:
-  // 1) vector are resizable at will by design
-  // 2) it will crash the resize program
+  // we don't want to execute this function on the watermark
   if (data.contentType === 'image/svg+xml') {
     console.log('File is an SVG image, exiting function');
     return false;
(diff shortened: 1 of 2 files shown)


apps/backend-functions/src/internals/image.ts
```diff
@@ -143,7 +143,23 @@ async function resize(data: functions.storage.ObjectMetadata) {
   return remove(workingDir);
 }
 
-export async function onImageDeletion(data: functions.storage.ObjectMetadata) {
+export async function onFileDeletion(data: functions.storage.ObjectMetadata) {
+
+  // TODO here we might need to handle deletion of other file types
```
phpgeekfr, line 148: can you add an issue number ?

apps/backend-functions/src/internals/image.ts
```diff
@@ -143,7 +143,23 @@ async function resize(data: functions.storage.ObjectMetadata) {
...
+    console.log(`File ${data.contentType} is not an image, exiting function`);
+    return false;
+  }
+
+  // we don't want to resize a vector image because:
+  // 1) vector are resizable at will by design
+  // 2) it will crash the resize program
+  if (data.contentType === 'image/svg+xml') {
+    console.log('File is an SVG image, exiting function');
+    return false;
```
phpgeekfr, line 161: We are on the delete side. Why do we make such tests? If we already have svg or pdf on the prod, it will stay forever

pldespaigne: ok sorry comment is miss leading I will change it.
but the thing is that this function is triggered on every file delete, so we should check that file actually an image before trying to delete the other size.

For example when you change the name of a user that will trigger a function that will overwrite the watermark with the new user name causing a `onDelete` event to fire

phpgeekfr: but like this you will skip the deletion of references after `// Clean document that reference this image``

GrandSchtroumpf: But the file is already delete at this point no ?

phpgeekfr: not the references in movie or org document for example

pldespaigne: what ?!

phpgeekfr:
``` 
 try {
    // Clean document that reference this image
    const docData: any = await getDocument(`${collection}/${id}`);
    const value = { ref: '', urls: [] };
    const updated = set(docData, fieldToUpdate, value);
    const docRef = db.collection(collection).doc(id);
    await docRef.update(updated);
  } catch (e) {
    console.log(`Error while updating image references in document "${collection}/${id}" : ${e.message}`);
  }
```

PR #3020, opened 2020-06-09
design event-review page

Hugo-Hontans:


apps/festival/festival/src/app/dashboard/event/review/review.component.html
```diff
@@ -20,6 +20,7 @@ <h3>Attendees</h3>
             [initialColumns]="initialColumns"
             [columns]="columns"
             [source]="analytics">
+            <ng-template colRef="name" let-name>{{ name | titlecase }}</ng-template>
```
GrandSchtroumpf, line 23: Could you use the `user | displayName | titlecase` instead ? Or something similar

apps/festival/festival/src/app/dashboard/event/review/review.component.ts
```diff
@@ -41,7 +38,11 @@ export class EventReviewComponent implements OnInit {
     this.analytics$ = this.eventId$.pipe(
       switchMap((eventId: string) => this.service.queryAnalytics(eventId)),
       filter(analytics => !!analytics),
-      map(analytics => analytics.eventUsers)
+      map(analytics => {
+        return analytics.eventUsers.map(event =>
+          ({ ...event, name: `${event.firstName} ${event.lastName}` })
+        );
```
GrandSchtroumpf, line 44: **If you cannot use the `displayName` pipe**, rewrite this like that : 
```typescript
map(analytics => {
  const transformEvent = event => ({ ...event, name: `${event.firstName} ${event.lastName}` });
  return analytics.eventUsers.map(transformEvent);
})
```

PR #3028, opened 2020-06-10
allowed more RAM and time to admin  (app express)

phpgeekfr:


apps/backend-functions/src/main.ts
```diff
@@ -29,8 +29,22 @@ import { getPrivateVideoUrl, uploadToJWPlayer } from './player';
...
+
+//--------------------------------
+//    Configuration             //
+//--------------------------------
+
+/** 
+ * Runtime options for heavy functions 
+ * @dev linked to #2531 (Changing functions REGION)
+ */
+const runtimeOpts = {
```
GrandSchtroumpf, line 41: Maybe give a better name like `heavyConfig` or someting like that.

Issue #2987, opened 2020-06-04
Img sizes references in documents are overwritten when form is saved to db


phpgeekfr:
Img sizes references in documents are overwritten when form is saved to db:

- [ ] org
- [ ] movie
- [ ] user

There is also an issue with excel upload

pldespaigne:
This issue will be partly solve with merge of the `release/media` branch, but we will still need to tackle the import excel part

pldespaigne:
Closed : Everything will be solved with the merge of #3142

PR #3029, opened 2020-06-10
creating media service

pldespaigne:
Creating a media service that handle multiple upload to storage alogn with play/pause/cancel and display of progress percentage

### DIFF
--- a/libs/utils/src/lib/media/image-uploader.service.ts
+++ b/libs/utils/src/lib/media/image-uploader.service.ts
@@ -1,0 +1,32 @@
+import { Injectable } from "@angular/core";
+import { AngularFireStorage, AngularFireUploadTask } from "@angular/fire/storage";
+import { HttpClient } from "@angular/common/http";
+import { createImgRef, ImgRef } from "./media.model";
+import { sanitizeFileName } from "../file-sanitizer";
+
+@Injectable({ providedIn: 'root' })
+export class ImageUploader {
+
+  constructor(
+    private afStorage: AngularFireStorage,
+    private httpClient: HttpClient,
+  ) { }
+
+  /**
+   * Fetchs a remote image and uploads it to firestore
+   * @param imageUrl
+   * @param afPath
+   */
+  public async upload(imageUrl: string, afPath: string = 'movies'): Promise<ImgRef | undefined> {
+    try {
+      const data = await this.httpClient.get(imageUrl, { responseType: 'blob' }).toPromise();
+      const snapshot = await this.afStorage.upload(`${afPath}/${sanitizeFileName(imageUrl)}`, data);
+      const url = await snapshot.ref.getDownloadURL();
+      const meta = await snapshot.ref.getMetadata();
+      return createImgRef({ urls: { original: url }, ref: meta.fullPath });
+    } catch (error) {
+      return;
+    }
+  }
+
+}
--- a/libs/utils/src/lib/media/image-uploader.service.ts
+++ b/libs/utils/src/lib/media/image-uploader.service.ts
@@ -25,6 +25,7 @@ export class ImageUploader {
       const meta = await snapshot.ref.getMetadata();
       return createImgRef({ urls: { original: url }, ref: meta.fullPath });
     } catch (error) {
+      console.error(`Upload fail because of`, error);
       return;
     }
   }
--- a/libs/utils/src/lib/media/media.query.ts
+++ b/libs/utils/src/lib/media/media.query.ts
@@ -1,6 +1,7 @@
 import { QueryEntity } from '@datorama/akita';
 import { Injectable } from '@angular/core';
-import { UploadState, MediaState, MediaStore } from './media.store';
+import { UploadState, MediaState, MediaStore, isUploading } from './media.store';
+import { map } from 'rxjs/operators';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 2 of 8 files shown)


libs/utils/src/lib/media/media.service.ts
```diff
@@ -1,32 +1,96 @@
...
-      const meta = await snapshot.ref.getMetadata();
-      return createImgRef({ urls: { original: url }, ref: meta.fullPath });
-    } catch (error) {
-      return;
+  /** Check if a file exists in the **Firebase storage** */
+  exists(path: string) {
+    return this.storage.ref(path).getDownloadURL().toPromise().then(() => true).catch(() => false);
+  }
+
+  async uploadBlob(path: string, data: Blob, fileName: string) {
```
GrandSchtroumpf, line 25: You should be able to upload many files at once. Best would be to have an object with the right format and use this format in the forms.
```typescript
uploadFormBlob(uploadFiles: UploadFile | UploadFile[]) {
  const files = Array.isArray(uploadFiles) ? uploadFiles : [uploadFiles];
  for (const file of files) {  ..  }
}
```

GrandSchtroumpf: If path is the refPath, call it `ref` or `refPath`

libs/utils/src/lib/media/media.service.ts
```diff
@@ -1,32 +1,96 @@
...
+
+  pause(fileName: string) {
+    if (this.query.hasEntity(fileName)) {
+      this.tasks[fileName].pause();
+      this.store.update(fileName, { status: 'paused' });
     }
   }
 
+  resume(fileName: string) {
+    if (this.query.hasEntity(fileName)) {
```
GrandSchtroumpf, line 79: I would rather look `if (fileName in this.tasks)`.

libs/utils/src/lib/media/media.query.ts
```diff
@@ -0,0 +1,21 @@
...
+
+@Injectable({
+  providedIn: 'root'
+})
+export class MediaQuery extends QueryEntity<MediaState, UploadState> {
+  constructor(protected store: MediaStore) {
+    super(store);
+  }
+
+  isUploading(fileName: string) {
```
GrandSchtroumpf, line 13: You need an observable version of this also.

libs/utils/src/lib/media/media.query.ts
```diff
@@ -0,0 +1,21 @@
...
+    const upload = this.getEntity(fileName);
+    if (!!upload && (upload.status === 'uploading' || upload.status === 'paused')) {
```
GrandSchtroumpf, line 15: You're using this a lot, maybe create a pure function 
```typescript
function isUploading(upload) {
  return upload?.status === 'uploading' || upload.status === 'paused'
}
function isDone(upload) {
  return upload?.status === 'succeed' || upload.status === 'canceled';
}
```

libs/utils/src/lib/media/media.query.ts
```diff
@@ -0,0 +1,21 @@
...
+import { UploadState, MediaState, MediaStore } from './media.store';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class MediaQuery extends QueryEntity<MediaState, UploadState> {
+  constructor(protected store: MediaStore) {
+    super(store);
+  }
+
```
GrandSchtroumpf, line 12: If would be good to have something like : 
```typescript
hasUploadingFiles() {
  return this.selectAll({ filterBy: isUploadind }).pipe(list => !!list.length);
}
```

libs/utils/src/lib/media/image-uploader.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    private afStorage: AngularFireStorage,
+    private httpClient: HttpClient,
+  ) { }
+
+  /**
+   * Fetchs a remote image and uploads it to firestore
+   * @param imageUrl
+   * @param afPath
+   */
+  public async upload(imageUrl: string, afPath: string = 'movies'): Promise<ImgRef | undefined> {
```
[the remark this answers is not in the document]
pldespaigne: Ok after discussion with @phpgeekfr we decided to leave the `uploadFromUrl` part related to excel import out of this PR as it's a big thing.

So it will be better with it's own PR

this is linked to #2987

libs/utils/src/lib/media/image-uploader.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    try {
+      const data = await this.httpClient.get(imageUrl, { responseType: 'blob' }).toPromise();
+      const snapshot = await this.afStorage.upload(`${afPath}/${sanitizeFileName(imageUrl)}`, data);
+      const url = await snapshot.ref.getDownloadURL();
+      const meta = await snapshot.ref.getMetadata();
+      return createImgRef({ urls: { original: url }, ref: meta.fullPath });
+    } catch (error) {
+      return;
```
Hugo-Hontans, line 28: maybe add a console.error or something to show the error

libs/utils/src/lib/media/media.service.ts
```diff
@@ -1,32 +1,98 @@
...
-    } catch (error) {
-      return;
+  /** Check if a file exists in the **Firebase storage** */
+  exists(path: string) {
+    return this.storage.ref(path).getDownloadURL().toPromise().then(() => true).catch(() => false);
+  }
+
+  uploadBlob(uploadFiles: UploadFile | UploadFile[]): Promise<void> | Promise<void>[] {
+    if (Array.isArray(uploadFiles)) {
+      return uploadFiles.map(uploadFile => this.uploadBlob(uploadFile) as Promise<void>);
```
Hugo-Hontans, line 25: You shouldn't have to force the type

PR #3036, opened 2020-06-10
expire 5h & player reload

### LINKED ISSUE #3032: Reload event session page when url expires
### Description
We want to redirect the user when the url of the film expires so he doesn't stay on an ongoing spinner. A reload of the page is enough as it should trigger the guard that will redirect the user to the event page. 

### Acceptance criteria
- [x] Add the url duration variable on env (+ be sure that staging + prod are updated accordingly) so back and front call the same variable
- [x] Reload the page when the page has been open for the same as the url duration variable
The duration to be set to 5 hours

pldespaigne:
fix #3032 asked by @Jackseed

### DIFF
--- a/apps/backend-functions/src/environments/environment.prod.ts
+++ b/apps/backend-functions/src/environments/environment.prod.ts
@@ -16,7 +16,6 @@ export {
   sentryDsn,
   bigQueryAnalyticsTable,
   centralOrgID,
-  linkDuration
 } from '@env';
 
 import { firebase } from '@env';
--- a/apps/backend-functions/src/environments/environment.prod.ts
+++ b/apps/backend-functions/src/environments/environment.prod.ts
@@ -7,7 +7,7 @@
 import * as functions from 'firebase-functions';
 
 import { algolia as algoliaClient } from '@env';
-export { 
+export {
   factoryContract,
   backupBucket,
   relayer,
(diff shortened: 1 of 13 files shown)


apps/backend-functions/src/environments/environment.prod.ts
```diff
@@ -7,15 +7,16 @@
...
   factoryContract,
   backupBucket,
   relayer,
   appUrlContent,
   appUrlMarket,
   sentryDsn,
   bigQueryAnalyticsTable,
-  centralOrgID
+  centralOrgID,
+  linkDuration
```
GrandSchtroumpf, line 19: Is it different depending on the build (dev, prod, ...) if not don't put that in the env

env/demo/env.demo1.ts
```diff
@@ -162,3 +162,7 @@ export const bigQueryAnalyticsTable = 'blockframes-demo1.analytics_200039147.eve
 // Archipel Content OrgId
 // ======================
 export const centralOrgID = "jnbHKBP5YLvRQGcyQ8In";
+
+// JWPlayer video link duration (in seconds)
+// ======================
+export const linkDuration = 18000; // 60 seconds * 60 min * 5 hours
```
GrandSchtroumpf, line 168: Better to write `1000 * 60 * 60 * 5`

PR #3039, opened 2020-06-11
App 2782 debug archipel

joanofdart:
Related to #2782


apps/festival/festival/src/app/marketplace/title/main/main.component.ts
```diff
@@ -51,6 +51,7 @@ export class MainComponent {
   }
 
   public budgetRange({ from, to }) {
-    return (from && to) ? `$ ${from} - ${to}` : '';
+    const formatter = new Intl.NumberFormat();
```
GrandSchtroumpf, line 54: Check it is works on Safari 11. I think Inlt has a bad browser coverage.
You could use a custom pipe for that (if it doesn't already exist). Check utils/pipes.

joanofdart: I have decided against intl and used DecimalPipe.

apps/festival/festival/src/app/marketplace/title/main/main.component.ts
```diff
@@ -51,6 +52,6 @@ export class MainComponent {
   }
 
   public budgetRange({ from, to }) {
-    return (from && to) ? `$ ${from} - ${to}` : '';
+    return (from && to) ? `$ ${this.numberPipe.transform(from)} - ${this.numberPipe.transform(to)}` : '';
```
GrandSchtroumpf, line 55: Use the method `formatNumber` instead (it's used under the hood by DecimalPipe but it's a pure function) : 
https://angular.io/api/common/formatNumber

apps/festival/festival/src/app/marketplace/title/main/main.component.ts
```diff
@@ -12,7 +13,7 @@ import { Movie } from '@blockframes/movie/+state/movie.model';
 export class MainComponent {
   public movie$ = this.movieQuery.selectActive();
 
-  constructor(private movieQuery: MovieQuery) { }
+  constructor(private movieQuery: MovieQuery, private numberPipe: DecimalPipe) { }
```
GrandSchtroumpf, line 16: Don't inject a pipe, angular always provides a pure functions that do the same job instead (see comment below)

libs/event/src/lib/layout/view/view.component.scss
```diff
@@ -16,4 +16,9 @@ pre {
 
 section {
   padding: 24px;
-}
\ No newline at end of file
+}
+
+.join-btn {
+  background: #FF6E61;
```
GrandSchtroumpf, line 22: Do not override the style of a global component. Here you're using mat-button and changing the background & border-radius, it means that it's not a mat-button anymore.
If it's on the design, it means that there is another component / directive that should do the work.
In this case I think it's `mat-flat-button` or `mat-stroked-button`.

The takeover here is that each time you are overriding the style of an existing component/directive, you should ask yourself if there is not another component that already do that. If not, you can ask me ;). But there is 5% chance that I let you override it in this end XD

PR #3047, opened 2020-06-11
Issue#3006 cleaning script

### LINKED ISSUE #3006: Clean db migration / script
### Description
We need to delete old docs that are related to deleted docs (such as notifications, users, orgs,...). 
One good advice from Laurent is to have a function onDelete that takes cares of automatically delete these docs, and better to tag the docs "deleted" instead of really deleting it. 
But this is a big work and we need first a patch to prevent prod errors currently happening. 

As we will need to play the script often, a migration script may be blocking as it will be played only once. Better would be to create a script that we can run whenever we want or a special migration scripts that is played on all migrations. 
I let the devs decide

### Acceptance criteria 
- [x] Delete notifications related to users that don't exist anymore
- [x] Delete invitations related to users that don't exist anymore
- [x] Delete userIds on org related to users that don't exist anymore + connected permissions
- [x] Delete movieIds on org related to movies that don't exist anymore + connected permissions
- [x] Delete notifications related to movies that don't exist anymore
- [x] Delete invitations related to events that don't exist anymore
- [x] Delete notifications related to orgs that don't exist anymore
- [x] Delete orgId on users related to orgs that don't exist anymore

ioaNikas:
Add cleaning script to remove or update deprecated data.

close #3006


apps/backend-ops/src/clean.ts
```diff
@@ -0,0 +1,186 @@
+import { loadAdminServices } from './admin';
```
GrandSchtroumpf, line 1: Maybe the name of the file can be more explicit. What are you cleaning ? Is it a script we can use several time or only once ?

apps/backend-ops/src/clean.ts
```diff
@@ -0,0 +1,186 @@
...
+import { NotificationDocument } from '@blockframes/notification/types';
```
GrandSchtroumpf, line 2: Use `@blockframes/notification/+state/notification.firestore` instead.

apps/backend-ops/src/clean.ts
```diff
@@ -0,0 +1,186 @@
...
+import { InvitationDocument } from '@blockframes/invitation/+state/invitation.firestore';
+import { User } from '@blockframes/user/types';
```
GrandSchtroumpf, line 4: Same here

apps/backend-ops/src/clean.ts
```diff
@@ -0,0 +1,186 @@
...
+    cleanNotifications(notifications, existingIds),
+    cleanInvitations(invitations, existingIds),
+    cleanUsers(users, organizationIds),
+    cleanOrganizations(organizations, userIds, movieIds),
+    cleanPermissions(permissions, organizationIds)
+  ]);
+}
+
+function cleanNotifications(
+  notifications: FirebaseFirestore.QuerySnapshot<FirebaseFirestore.DocumentData>,
```
GrandSchtroumpf, line 51: Where did you get the reference to the type `FirebaseFirestore`?
Could you do :
```typescript
import { QuerySnapshot, DocumentData } from 'firestore/firestore';
```

apps/backend-ops/src/clean.ts
```diff
@@ -0,0 +1,186 @@
...
+/**
+ * Check each type of invitation and return false if a referenced document doesn't exist
+ * @param invitation the invitation to check
+ * @param existingIds the ids to compare with invitation fields
+ */
+function isInvitationValid(invitation: InvitationDocument, existingIds: string[]): boolean {
+  switch (invitation.type) {
+    case 'attendEvent':
+      return (
+        (existingIds.includes(invitation.fromOrg?.id) &&
```
GrandSchtroumpf, line 171: Create a variable for each of these assessement :
```typescript
const toOrgExists = existingIds.includes(invitation.fromOrg?.id);
const toUserExists  = existingIds.includes(invitation.toUser?.uid);
...
```

apps/backend-ops/src/clean.ts
ioaNikas: `FirebaseFirestore` is a namespace, maybe because of that we don't need to import it ?

`QuerySnapshot, DocumentData` don't exist in firebase/firestore :(

PR #3050, opened 2020-06-11
blockframes admin guard and image sizes display for org, movie and users

phpgeekfr:


libs/admin/src/lib/admin-panel/admin-panel.module.ts
```diff
@@ -26,6 +26,7 @@ import { EditTitleModule } from './components/edit-title/edit-title.module';
 import { TranslateSlugModule } from '@blockframes/utils/pipes/translate-slug.module';
 import { OrgNameModule } from '@blockframes/organization/pipes/org-name.pipe';
 import { MemberPermissionsModule } from '@blockframes/organization/components/member-permissions/member-permissions.module';
+import { ImageSizesModule } from './components/image-sizes/image-sizes.module';
```
GrandSchtroumpf, line 29: Would be nice to have that in ui/media. Like that we could use it in different places. And it'll be part of the future media library.

libs/admin/src/lib/admin-panel/components/image-sizes/image-sizes.component.html
```diff
@@ -0,0 +1,28 @@
+<mat-card>
+  <mat-card-header>
+    <mat-card-title>{{ title }}</mat-card-title>
+  </mat-card-header>
+  <mat-card-content fxLayout="column" fxLayoutAlign="center center">
+    <ng-container *ngIf="imgRef else noimgref">
+      <ng-container *ngIf="imgRef.urls.original">
```
GrandSchtroumpf, line 7: What about : 
```html
<ng-container *ngFor="imgRef.urls | keyvalue as url">
  <img [src]="url.value" [alt]="url.key + ' version'" />
  <a [href]="url.value" target="_blank">{{ url.key | imgSize }}</a>
</ng-container>
```

libs/admin/src/lib/admin-panel/guard/blockframes-admin.guard.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@angular/core';
+import { AuthQuery } from '@blockframes/auth/+state/auth.query';
+import { Router } from '@angular/router';
+
+@Injectable({ providedIn: 'root' })
+export class BlockframesAdminGuard {
```
GrandSchtroumpf, line 6: add `implements CanActivate`

libs/admin/src/lib/admin-panel/guard/blockframes-admin.guard.ts
```diff
@@ -0,0 +1,20 @@
...
+  constructor(
+    private authQuery: AuthQuery,
+    protected router: Router
+  ) { }
+
+  async canActivate() {
+    const isBlockframesAdmin = this.authQuery.isBlockframesAdmin;
```
GrandSchtroumpf, line 13: What about : 
```typescript
return this.authQuery.isBlockframesAdmin
  ? true
  : this.router.parseUrl('c/o');
```

libs/admin/src/lib/admin-panel/components/image-sizes/image-sizes.component.ts
```diff
@@ -0,0 +1,14 @@
...
+
+@Component({
+  selector: '[imgRef] admin-image-sizes',
+  templateUrl: './image-sizes.component.html',
+  styleUrls: ['./image-sizes.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ImageSizeComponent  {
+  @Input() imgRef: ImgRef;
+  @Input() title = 'Image sizes';
```
GrandSchtroumpf, line 13: Instead of passing an input, could you use `<ng-content>` it would looks nicer on the html

libs/admin/src/lib/admin-panel/pages/movie/movie.component.html
```diff
@@ -91,7 +91,15 @@ <h2>Rights :</h2>
       </bf-table-filter>
     </mat-card>
   </article>
+  <mat-card *ngIf="movie">
+    <h2>Promotional Elements :</h2>
 
+    <admin-image-sizes *ngIf="movie.promotionalElements.banner.media" title="Movie banner" [imgRef]="movie.promotionalElements.banner.media"></admin-image-sizes>
```
GrandSchtroumpf, line 97: I think there is a pipe something like `[imgRef]="movie | ref: 'movie': 'banner'"` Like that we can update the path at one place only (as we are planning to move banner to main, it might make sense).

libs/admin/src/lib/admin-panel/pages/movie/movie.component.html
```diff
@@ -91,7 +91,15 @@ <h2>Rights :</h2>
...
+    <ng-container *ngFor="let poster of movie.promotionalElements.poster; let i = index; let last = last">
+      <admin-image-sizes [title]="posterName(i)" [imgRef]="poster.media"></admin-image-sizes>
```
GrandSchtroumpf, line 99: Ideally I would rather have something like : 
```html
<admin-image-sizes [imgRef]="poster.media">
  <h3>Movie poster n°{{ i + 1 }}</h3> <!-- Or whatever tag you want -->
</admin-image-sizes>
```

PR #3054, opened 2020-06-12
safari-banner

pldespaigne:


libs/utils/src/lib/safari-banner/safari-banner.component.ts
```diff
@@ -0,0 +1,50 @@
...
+    );
+}
+
+@Component({
+  selector: 'safari-banner',
+  templateUrl: './safari-banner.component.html',
+  styleUrls: ['./safari-banner.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class SafariBannerComponent implements OnInit {
```
GrandSchtroumpf, line 29: Many we should call it 'info-footer' and have several components inside to display different type of information (like safari, or cookies, or whatever).

PR #3055, opened 2020-06-12
Invitation link

Hugo-Hontans:

### DIFF
--- a/libs/notification/src/lib/invitation/components/item/item.component.html
+++ b/libs/notification/src/lib/invitation/components/item/item.component.html
@@ -12,7 +12,8 @@
   <invitation-action [invitation]="invitation"></invitation-action>
   <a
     mat-button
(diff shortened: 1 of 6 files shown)


libs/notification/src/lib/invitation/components/item/item.component.html
```diff
@@ -10,7 +10,10 @@
     </small>
   </span>
   <invitation-action [invitation]="invitation"></invitation-action>
-  <a mat-button [routerLink]="['../event/', invitation.docId]">
+  <a
+    mat-button
+    [routerLink]="invitation.type === 'joinOrganization' ? ['../../organization', invitation.toOrg.id, 'view', 'members'] : ['../event/', invitation.docId]"
```
GrandSchtroumpf, line 15: Could you create a pipe for that ? I've got an example on how to do it in `libs/events/pipes/event-link.pipe`.

PR #3059, opened 2020-06-12
Quick media fix

GrandSchtroumpf:
- merge image from the server
- upload scenario

### DIFF
--- a/libs/movie/src/lib/movie/+state/movie.service.ts
+++ b/libs/movie/src/lib/movie/+state/movie.service.ts
@@ -51,6 +51,20 @@ export class MovieService extends CollectionService<MovieState> {
     return movie;
   }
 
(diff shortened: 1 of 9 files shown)


libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -51,6 +51,20 @@ export class MovieService extends CollectionService<MovieState> {
     return movie;
   }
 
+  // @TODO (#2987): Update image on save instead of merging then on save
+  save(movie: Movie) {
+    return this.runTransaction(async write => {
+      const ref = this.getRef(movie.id);
+      const snap = await write.get(ref);
+      const movieDoc = snap.data() as Movie;
```
phpgeekfr, line 59: maybe add createMovie(snap.data()) to be sure that movieDoc.promotionalElements exists ?

PR #3073, opened 2020-06-15
fix indexes for events

Hugo-Hontans:
- [x] Fix query with indexes to load events


apps/festival/festival/src/app/marketplace/event/list/list.component.ts
```diff
@@ -17,9 +18,10 @@ export class ListComponent implements OnInit {
 
   ngOnInit(): void {
     const query = ref => ref
-      .orderBy('meta.titleId').where('meta.titleId', '>', '')
       .orderBy('end').startAt(new Date());
-    this.events$ = this.service.queryByType(['screening'], query);
+    this.events$ = this.service.queryByType(['screening'], query).pipe(
```
GrandSchtroumpf, line 22: Can you add a comment to explain why we can't do that on queryFn ?

PR #3078, opened 2020-06-16
[admin-panel] Org name bigger on org page & reset password email

phpgeekfr:


libs/organization/src/lib/organization/pipes/org-name.pipe.ts
```diff
@@ -3,13 +3,18 @@ import { Organization, orgName } from '../+state';
 
 @Pipe({ name: 'orgName', pure: true })
 export class OrgNamePipe implements PipeTransform {
-  transform(value: Organization) {
-    return orgName(value); 
+  transform(org: Organization, display: 'public' | 'full' | 'long' = 'public') {
+    switch (display) {
```
GrandSchtroumpf, line 7: Don't use a switch when you can do a simple `if / else` statement.

libs/admin/src/lib/admin/+state/admin.service.ts
```diff
@@ -27,4 +28,14 @@ export class AdminService {
...
     return f({}).toPromise();
   }
+
+  /**
+  * Send email from Google Firebase to reset password
+  * @dev This is the "hard" reset password link, sent directly from Google.
+  * @param email 
+  */
+  public async sendPasswordResetEmail(email: string): Promise<void> {
+    const auth = firebase.auth();
```
GrandSchtroumpf, line 38: It would be better if you use `AngularFireAuth`: 
```typescript
constructor(private auth: AngularFireAuth) {}
public async sendPasswordResetEmail(email: string): Promise<void> {
  return this.auth.sendPasswordResetEmail(email);
}
```

libs/admin/src/lib/admin/+state/admin.service.ts
```diff
@@ -27,4 +28,14 @@ export class AdminService {
...
+    return await auth.sendPasswordResetEmail(email);
```
GrandSchtroumpf, line 39: remove `await`

Issue #3092, opened 2020-06-16
Admin Panel - Org import debug


Jackseed:
Currently the email sent when importing orgs is in the name of the blockframesAdmin organisation and not the created org. 
Eg. I create org X with user A using blockframesAdmin from org Archipel
A will receive an invitation to join org Archipel instead of X

Issue #2952, opened 2020-06-02
Admin Panel - Analytics


Jackseed:
Update admin panel to add some analytics on users for the business development.
Here [are some examples](https://projects.invisionapp.com/d/main#/console/19564186/411037156/preview).
Check with @maltmayer the priorities 

- [x] be able to export orgs list into excel or csv
- [x] be able to export users list into excel or csv @phpgeekfr missing orgname & role
- [x] be able to export films list into excel or csv
- [x] be able to export event list into excel or csv
- [x] check https://docs.google.com/spreadsheets/d/1jckFAorLiD2TaZMG_FdoYO7ktCIStBRKgQP8eDKWmIA/edit#gid=7312071

- [x] @maltmayer reco in #2090 https://github.com/blockframes/blockframes/issues/2090#issuecomment-639395819

phpgeekfr:
@Jackseed @maltmayer Can we schedule a demo so I can show you what I have done so far ?

PR #3100, opened 2020-06-17
Feat #3092 - corrected org sender for invitation to org in admin panel

phpgeekfr:


libs/admin/src/lib/admin-panel/pages/event/event.component.html
```diff
@@ -9,8 +9,8 @@ <h2>
   <h3 *ngIf="eventOwner">Owner : {{ eventOwner }}</h3>
   <ul>
     <li><strong>Type :</strong> {{ event.type }}</li>
-    <li><strong>From :</strong> {{ event.start.toDate() | date: 'medium'}}</li>
-    <li><strong>To :</strong> {{ event.end.toDate() | date: 'medium'}}</li>
+    <li><strong>From :</strong> {{ event.start | date: 'medium'}}</li>
```
GrandSchtroumpf, line 12: 😍

libs/admin/src/lib/admin-panel/pages/event/event.component.ts
```diff
@@ -41,15 +42,18 @@ export class EventComponent implements OnInit {
...
-            .then(t => {
-              this.movie = t;
-              return t;
-            })
-            .then(_ => this.cdRef.markForCheck())
-            .catch(_ => {
-              this.snackBar.open('Error while loading movie private config', 'close', { duration: 5000 });
-            });
+          if (titleId) {
+            this.movieService.getValue(titleId as string)
```
GrandSchtroumpf, line 46: Could you format it to `async/await` ? I think it would be much clearer : 
```typescript
const titleId = this.event.meta.titleId as string;
if (titleId) {
  try {
    this.movie = await this.movieService.getValue(titleId);
    this.cdRef.markforCheck(); // I'm not even sure you'll need it with the async/await snthax
  } catch (err) {
    this.snackBar.open('Error while loading movie', 'close', { duration: 5000 });
  }
}
```

phpgeekfr: Will add a comment with issue number because this update was in fact part of #2952 and was not intended to be committed yet

PR #3106, opened 2020-06-17
[WIP] feat #2952 analytics info & statistics

phpgeekfr:


libs/admin/src/lib/admin/toolbar-top/toolbar-top.component.scss
```diff
@@ -0,0 +1,3 @@
+.fill-remaining-space {
+ flex: 1 1 auto;
```
GrandSchtroumpf, line 2: Why not using `fxFlex` as attribute direction in the html? It should do exactly that.

libs/admin/src/lib/admin/admin/admin.component.ts
```diff
@@ -0,0 +1,8 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'admin-admin',
```
GrandSchtroumpf, line 4: change selector to `admin-home` or `admin-layout` or something else.

libs/admin/src/lib/admin/admin/admin.component.ts
```diff
@@ -0,0 +1,8 @@
...
+  templateUrl: './admin.component.html',
+  styleUrls: ['./admin.component.scss']
```
GrandSchtroumpf, line 6: add changeDetectionStrategy.OnPush

libs/admin/src/lib/admin/toolbar-top/toolbar-top.component.ts
```diff
@@ -0,0 +1,8 @@
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'admin-toolbar-top',
+  templateUrl: './toolbar-top.component.html',
+  styleUrls: ['./toolbar-top.component.scss']
```
GrandSchtroumpf, line 6: add changeDetectionStrategy.OnPush

PR #3110, opened 2020-06-18
Original filename for upload component

### LINKED ISSUE #3108: Show original name of image file in upload component
Cropper component sanitizes filename into something that isn't recognized by the user. This looks confusing when seeing the file upload status. The idea is to show the original filename of the file which is being uploaded.

RemcoSimonides:
fix #3108

### DIFF
--- a/libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
+++ b/libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
@@ -124,12 +124,12 @@ export class CropperComponent implements OnInit, OnDestroy {
       const blob = b64toBlob(this.croppedImage);
 
       this.nextStep('show');
-
+      const fileName = this.file.name.replace(/(\.[\w\d_-]+)$/i, '.webp');
       this.form.patchValue({
         path: `${this.storagePath}/original/`,
         blob: blob,
         delete: false,
-        fileName: this.file.name
+        fileName: fileName
       })
       this.form.markAsDirty();
 
--- a/libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
+++ b/libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
@@ -124,6 +124,8 @@ export class CropperComponent implements OnInit, OnDestroy {
       const blob = b64toBlob(this.croppedImage);
 
       this.nextStep('show');
+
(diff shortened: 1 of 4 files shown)


libs/utils/src/lib/media/media.service.ts
```diff
@@ -72,8 +76,8 @@ export class MediaService {
    * @param fileOrBlob
    * @param fileName
    */
-  private async upload(path: string, fileOrBlob: Blob | File, fileName: string) {
-    const exists = await this.exists(path.concat(fileName));
+  private async upload(path: string, fileOrBlob: Blob | File, fileName: string, sanitizedFileName: string) {
```
pldespaigne, line 79: don't add a `sanitizeFileName` parameter, you can sanitize only once inside the function instead of doing it in every parent function

libs/utils/src/lib/media/media.service.ts
```diff
@@ -43,7 +44,8 @@ export class MediaService {
     if (Array.isArray(uploadFiles)) {
       uploadFiles.forEach(uploadFile => this.uploadBlob(uploadFile));
     } else {
-      this.upload(uploadFiles.ref, uploadFiles.data, uploadFiles.fileName);
+      const sanitizedFileName = sanitizeFileName(uploadFiles.fileName).replace(/(\.[\w\d_-]+)$/i, '.webp');
```
pldespaigne, line 47: wtf is `.replace(/(\.[\w\d_-]+)$/i, '.webp');` what happen if my blob is not an image ?

libs/ui/src/lib/media/cropper/cropper/cropper.component.ts
```diff
@@ -124,12 +124,12 @@ export class CropperComponent implements OnInit, OnDestroy {
       const blob = b64toBlob(this.croppedImage);
 
       this.nextStep('show');
-
+      const fileName = this.file.name.replace(/(\.[\w\d_-]+)$/i, '.webp');
```
pldespaigne, line 127: can you try to understand what this line does, and add a comment to explain it please

RemcoSimonides: Ah right, of the file name it takes the part after the "." and changes it to ".webp" so; with "abc.def.gfh" it takes ".gfh" and changes it to ".webp" making "abc.def.webp" in the end.

PR #3112, opened 2020-06-18
build(npm): update firebaserc with my env

george43g:


george43g:
> You might want to add the festival hosting too

Done.

PR #3122, opened 2020-06-19
Fix deleting other media values when uploading new image

RemcoSimonides:
- [x] Uploading new image on MovieForm deletes the other images on the form
- [ ] The "Delete" button of a Still Photo on the MovieForm doesn't delete it from the array in the database and therefore it is shown again when the page is refreshed.
- [ ]  The "Delete" button of a Still Photo empties the Still Photo Array on the database and doesn't delete any image from the storage.
- [ ] Last Still Photo doesn't have a delete button

### DIFF
--- a/libs/media/src/lib/+state/media.model.ts
+++ b/libs/media/src/lib/+state/media.model.ts
@@ -2,20 +2,20 @@ import { ImgRef, OldImgRef } from './media.firestore';
 
 export * from './media.firestore';
 
-export function extractMedia(origin: any) {
+export function extractToBeUpdatedMedia(origin: any) {
   const value = Object.assign({}, origin);
-  const media = extractMediaValue(value);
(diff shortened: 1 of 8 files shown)


libs/media/src/lib/+state/media.model.ts
```diff
@@ -26,6 +26,9 @@ function isMedia(obj: any): boolean {
   return typeof obj === 'object' && !!obj && 'ref' in obj && 'urls' in obj;
 }
 
+function mediaNeedsUpdate(obj: ImgRef): boolean {
+  return obj.delete || (obj.path && obj.blob);
```
pldespaigne, line 30: `obj.delete` is a boolean, but not `obj.path` and `obj.blob` it's better to write `(!!obj.path && !!obj.blob)`

PR #3124, opened 2020-06-19
last fix for create wishlist tests

Hugo-Hontans:


Hugo-Hontans:
> why deleting it ? dont you need at some point the code again?

I lighten the tests as much as possible, we test the same functionality several times for nothing like removing a movie from the wishlist

PR #3127, opened 2020-06-19
Update the background ref to support Safari

GrandSchtroumpf:


libs/event/src/lib/components/card/card.component.html
```diff
@@ -6,13 +6,15 @@
 
   <!-- Screening -->
   <ng-container *ngSwitchCase="'screening'">
-    <ng-container *ngIf="event.movie as movie">
-      <div class="dark-contrast-theme background" [bgRef]="movie | screeningBackground : size">
+    <ng-container *ngIf="event.movie as movie; else noMovie">
+      <div class="dark-contrast-theme background" [bgRef]="movie | screeningBackground : size" bgAsset="empty_poster.webp">
```
pldespaigne, line 10: Oh yeah ! what about all our illustrations that are `webp` ?

GrandSchtroumpf: Max updated them.

PR #3128, opened 2020-06-19
added spec file

### LINKED ISSUE #3087: Unit test the Media service
link to the master issue #3084

fritzschoff:
close #3087

### DIFF
--- a/libs/media/src/lib/+state/media.service.spec.ts
+++ b/libs/media/src/lib/+state/media.service.spec.ts
@@ -1,20 +1,35 @@
 import { createServiceFactory, SpectatorService, mockProvider } from '@ngneat/spectator/jest';
 import { MediaService } from './media.service';
-import { MediaStore } from './media.store';
+import { AngularFireStorage, AngularFireStorageModule } from '@angular/fire/storage';
+import { Overlay, OverlayRef } from '@angular/cdk/overlay';
+import { FirebaseApp, AngularFireModule } from '@angular/fire';
+import { firebase } from '@env';
 import { MediaQuery } from './media.query';
-import { AngularFireStorage } from '@angular/fire/storage';
-import { Overlay } from '@angular/cdk/overlay';
+import { MediaStore } from './media.store';
 
+const random = () => (Math.random() + 1).toString(36).substring(7);
 
-// this.overlay.position().global().bottom('16px').left('16px')
+const TESTDATA = {
+  path: 'unit-test/',
+  blob: new Blob([JSON.stringify({ blockframes: 'movies' })], { type: 'application/json' }),
+  dataName: 'test',
+  file: new File([JSON.stringify({ blockframes: 'movies' })], 'test'),
+}
+
+async function prepareStorage(afStorage: AngularFireStorage) {
+  return afStorage.ref(TESTDATA.path.concat(TESTDATA.dataName)).delete().toPromise()
+}
 
 describe('Media Service', () => {
-
+  let app: FirebaseApp;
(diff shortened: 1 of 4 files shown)


libs/media/src/lib/+state/media.service.spec.ts
```diff
@@ -29,14 +44,46 @@ describe('Media Service', () => {
...
+
+  test('Upload a blob to storage and validate it exists', async done => {
+    prepareStorage(afStorage)
+    spectator.service.uploadBlob({ path: TESTDATA.path, data: TESTDATA.blob, fileName: TESTDATA.dataName });
+    let result: boolean;
+    setTimeout(async () => {
+      result = await spectator.service.exists(TESTDATA.path.concat(TESTDATA.dataName));
+      expect(result).toBe(true)
+      done();
+    }, 500)
```
george43g, line 72: What if it takes longer than 500ms for the file to upload? I feel like this isn't guaranteed to pass.

I feel like in `libs/media/src/lib/+state/media.service.ts` methods like uploadBlob and UploadFile should return their promise, rather than just leave it hanging.

george43g:
https://github.com/blockframes/blockframes/pull/3128/commits/37af3042ae54faf2cd5d1e10cf904af02d30cc48#r444827714

george43g:
I feel like otherwise, `libs/media/src/lib/+state/media.service.spec.ts` is quite beautifully done. It also seems that you followed what we discussed quite closely, which is really great to see as well. Good teamwork :) 

There would have been a lot of problem-solving to get that working. I've left some comments on certain lines to discuss, some of the suggestions may not be trivial to implement if we need to get this working quickly - however, I think it's also important to get our testing structure and strategy correct from the get-go so we don't have to go back later and rewrite the tests.

fritzschoff:
@george43g I had a talk with Pierre about returning the promise and we agreed on that it would be a good idea to return the promise. So I update and clean the test accordingly. But now the test is failing, cause if you upload a blob and check it if it exists, you won't found it. But if you check then by yourself on firebase storage the file is there. So I guess, firebase needs a split second to validate that file in order then to let the service find it. Do you got an idea how to fix that?

george43g:
> @george43g I had a talk with Pierre about returning the promise and we agreed on that it would be a good idea to return the promise. So I update and clean the test accordingly. But now the test is failing, cause if you upload a blob and check it if it exists, you won't be able to find it. But if you check then by yourself on firebase storage the file exist. So I guess, firebase needs a split second to validate that file in order then to let the service find it. Like when you rerun the test, it will passes cause he can find the file. Do you got an idea how to fix that?

Sorry - I must not have pressed the "submit review" button so a lot of my comments must not have been visible earlier.

To answer your question, as per one of the comments, I think we need to first do the assertions properly, and then if it's still an issue, we'll go from there.

Usually I like to clear out backend services at the beginning of the test suite to ensure that the generated data is consistent. In our case that might not work though... let's take it one step at a time and see how it goes. I'll be merging a PR soon that introduces a `.env` file to the repo, so that could potentially help solve env related issues here as well.

pldespaigne:
bump
@george43g @fritzschoff @xmano

pldespaigne:
@george43g @xmano

fritzschoff:
this is getting stale @pldespaigne

pldespaigne:
what is the state of this PR ?
should we close it or should we try to merge it ?
@xmano @george43g what are your thoughts on this ?

xmano:
Created a new PR
https://github.com/blockframes/blockframes/pull/3999

Issue #3117, opened 2020-06-18
[Hotfix] Update directive to support fallback image


GrandSchtroumpf:
We need to support fallback images 
- Use "bf-img" component instead of imgRef directive.
- Update backgroundRef directives.

0. **checkout from `hotfix/1.6.1`**
1. Modify every img tag with `bg-img`
2. Check css style that target img tags
3. push on `hotfix/1.6.1`
4. push `hotfix/1.6.1` on staging & test.

PR #3130, opened 2020-06-19
App 3117 img directive issue

joanofdart:
Related to #3117


libs/organization/src/lib/organization/components/card/card.component.scss
```diff
@@ -9,9 +9,12 @@ a {
   border-radius: 50%;
   margin-bottom: 16px;
 
-  img {
-    width: 100%;
-    border-radius: 50%;
+  picture {
```
GrandSchtroumpf, line 12: you should use bf-img here I think

fritzschoff:
wait for the apps, dont merge yet

PR #3132, opened 2020-06-19
Issue 3117 max

fritzschoff:
ref #3117

### DIFF
--- a/libs/event/src/lib/components/screening-item/screening-item.component.scss
+++ b/libs/event/src/lib/components/screening-item/screening-item.component.scss
@@ -6,11 +6,9 @@
 a {
   height: 300px;
   margin-right: 24px;
-  
-  picture {
-    bf-img {
-      height: 100%;
-    }
+
+  bf-img {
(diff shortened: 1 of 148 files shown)


libs/event/src/lib/components/screening-item/screening-item.component.scss
```diff
@@ -6,8 +6,11 @@
 a {
   height: 300px;
   margin-right: 24px;
-  img {
-    height: 100%;
+  
+  picture {
```
GrandSchtroumpf, line 10: You might want to update to `bf-img` instead.

libs/organization/src/lib/organization/components/card/card.component.scss
```diff
@@ -9,9 +9,12 @@ a {
   border-radius: 50%;
   margin-bottom: 16px;
 
-  img {
-    width: 100%;
-    border-radius: 50%;
+  picture {
```
GrandSchtroumpf, line 12: Same here

Issue #3120, opened 2020-06-18
[Hotfix] Update resize image cloud function to create fallback


GrandSchtroumpf:
webp doesn't work on Safari ! We need to create a png fallback version of the image when someone create one.

0. **Checkout from `hotfix/1.6.1`**
1. Update the cloud function of resize to convert the original image into png.
2. Update firestore to add the downloadUrl of this fallback image into the field `urls.fallback` of the imgRef.
3. Update the cloud function hook on delete to remove the fallback image also.
4. Create a PR on hotfix/1.6.1.
5. Push hotfix/1.6.1 on staging (call @GrandSchtroumpf first)
6. Test on staging: https://staging.archipelmarket.com on Safari through browserstack (ask @joanofdart if you need help).
7. Grab a glass 🥂

PR #3133, opened 2020-06-19
add fallback image

pldespaigne:
#3120

### DIFF
--- a/apps/backend-functions/src/internals/image.ts
+++ b/apps/backend-functions/src/internals/image.ts
@@ -1,6 +1,6 @@
 import { db, functions } from './firebase';
 import { tmpdir } from 'os';
-import { join, dirname } from 'path';
+import { join, dirname, basename, extname } from 'path';
 import * as admin from 'firebase-admin';
 import { ensureDir, remove } from 'fs-extra';
 import sharp from 'sharp';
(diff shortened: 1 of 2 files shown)


apps/backend-functions/src/internals/image.ts
```diff
@@ -75,19 +76,40 @@ async function resize(data: functions.storage.ObjectMetadata) {
...
+        version: 'v2'
+      });
+
+      uploaded.push({ key, url: signedUrl });
+
+    // Fallback : we need to convert the uploaded file into a png
+    // and also generate an access url
+    } else if (key === 'fallback') {
+
+      const pngFileName = basename(resizedImgName, extname(resizedImgName)) + '.png'
```
GrandSchtroumpf, line 105: Do you need to add ".png" in the end ? I'm not sure we are doing that for webp. But that's fine for me. Maybe we should.

libs/utils/src/lib/media/media.firestore.ts
```diff
@@ -26,23 +26,23 @@ export function createImgRef(ref: Partial<ImgRef> | string = {}): ImgRef {
...
+      lg: '',
     },
     ..._ref
   };
 }
 
 export function getImgSize(url: string) {
   if (url.includes('avatar')) {
-    return { original: 0, xs: 50, md: 100, lg: 300 };
+    return { original: 0, xs: 50, md: 100, lg: 300, fallback: 0 };
```
GrandSchtroumpf, line 37: Hmmm do we need that ?

pldespaigne: unfortunately yes, that's how we iterate over the different files we need to create

apps/backend-functions/src/internals/image.ts
pldespaigne: yes every other image have the `.webp` extension, if we don't do this our png will be named `myPng.webp`

PR #3134, opened 2020-06-19
Removed support for OldImgRef

### LINKED ISSUE #3063: Remove old path for media
Because of retrocompatibility issues we keep old path for media "media.url" as a condition.
We would remove the use of this old format for `media.urls.original`

RemcoSimonides:
Fix #3063

### DIFF
--- a/libs/media/src/lib/+state/media.model.ts
+++ b/libs/media/src/lib/+state/media.model.ts
@@ -1,4 +1,4 @@
-import { ImgRef, OldImgRef } from './media.firestore';
+import { ImgRef } from './media.firestore';
 
 export * from './media.firestore';
 
@@ -50,14 +50,4 @@ export type Formats = keyof typeof formats;
 export function getRatio(format: Formats) {
   const { height, width } = formats[format];
   return width / height;
-}
-
-// @todo(#3063) Remove this verifier
-export function isOldImgRef(ref: ImgRef | OldImgRef): ref is OldImgRef {
-  return 'url' in ref;
-}
-
-// @todo(#3063) Update this function to unsupport oldImgRef
-export function getMediaUrl(ref: ImgRef | OldImgRef) {
-  return isOldImgRef(ref) ? ref.url : ref.urls.original;
 }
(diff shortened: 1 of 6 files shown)


libs/media/src/lib/+state/media.model.ts
```diff
@@ -51,13 +51,3 @@ export function getRatio(format: Formats) {
...
-
-// @todo(#3063) Remove this verifier
-export function isOldImgRef(ref: ImgRef | OldImgRef): ref is OldImgRef {
-  return 'url' in ref;
-}
-
-// @todo(#3063) Update this function to unsupport oldImgRef
-export function getMediaUrl(ref: ImgRef | OldImgRef) {
-  return isOldImgRef(ref) ? ref.url : ref.urls.original;
-}
```
phpgeekfr, line 63: You removed this because we don't use ref.url anymore ?   I'm afraid we had to update this method in production to make the safari fix.

The new one looks like this  ```
// @todo(#3063) Remove this verifier
export function isOldImgRef(ref: ImgRef | OldImgRef): ref is OldImgRef {
  return 'url' in ref;
} 

// @todo(#3063) Update this function to unsupport oldImgRef
export function getMediaUrl(ref: ImgRef | OldImgRef) {
  if (ref) {
    if (isOldImgRef(ref)) {
      return ref.url;
    } else {
      return isSafari() ? ref.urls.fallback : ref.urls.original;
    }
  }
}
```
So can you wait the merge back from production to develop so you can update the new `getMediaUrl` ? If you want I can also remove the `f(isOldImgRef(ref)) ` while merging back.

phpgeekfr: @RemcoSimonides  Since I had to face merge conflicts, I took your update into account for #3063 when merging back hotfix/1.6.1 into develop.

phpgeekfr: I also removed the old `OldImgRef` interface

RemcoSimonides: Ah I am glad I asked you to review this! I can remove this PR now?

phpgeekfr: Yes I think you can remove it

PR #3148, opened 2020-06-22
Fix circular dependency and upload widget auto closes after 5 seconds

RemcoSimonides:
Tasks are now in an InjectionToken in order to be able to use them in both the MediaService and UploadWidgetComponent. The const for this InjectionToken needs to live in a different file than MediaService.

Media Store is possibly not needed anymore

The UploadWidgetComponent now closes automatically after 5 seconds if all uploads have succeeded. If the upload has failed, then the UploadWidgetComponent stays. The widget can then only be closed by clicking the cancel button on the failed uploads.


libs/media/src/lib/+state/media.service.ts
```diff
@@ -83,33 +88,31 @@ export class MediaService {
...
       throw new Error(`Upload Error : A file named ${fileName} is already uploading!`);
     }
 
     const task = this.storage.upload(path.concat(fileName), fileOrBlob);
 
-    this.store.upsert(fileName, {
+    this.tasks.add({
+      id: fileName,
       status: 'uploading',
       progress: 0,
```
GrandSchtroumpf, line 101: I think the kind of information the uploadTask has.

libs/media/src/lib/+state/media.service.ts
```diff
@@ -83,33 +88,31 @@ export class MediaService {
...
+      uploadtask: task
     });
 
-    task.percentageChanges().subscribe(p => this.store.update(fileName, { progress: p }));
-
-    this.tasks[fileName] = task;
+    task.percentageChanges().subscribe(p => this.tasks.updateProgress(fileName, p));
```
GrandSchtroumpf, line 105: You don't need to do that anymore.

RemcoSimonides:
Closing in favor o version2 https://github.com/blockframes/blockframes/pull/3269

PR #3178, opened 2020-06-24
Added Organization Minimal Component + fix storybook

RemcoSimonides:
- [x] Add Organization Minimal Component
- [ ] Add Featured Sales Agent Carousel to Homepage Festival

### DIFF
--- a/libs/movie/src/lib/movie/components/card/card.stories.ts
+++ b/libs/movie/src/lib/movie/components/card/card.stories.ts
@@ -2,13 +2,14 @@ import { object } from '@storybook/addon-knobs';
 import { MovieCardModule } from './card.module';
 import { ToolkitModule, MOVIES } from '@blockframes/ui/storybook';
 import { CardComponent } from './card.component';
+import { RouterTestingModule } from '@angular/router/testing';
 
 export default {
   title: 'Movie Card'
 };
 
 export const movieCard = () => ({
-  moduleMetadata: { imports: [MovieCardModule, ToolkitModule] },
+  moduleMetadata: { imports: [MovieCardModule, ToolkitModule, RouterTestingModule] },
   name: 'Movie Card',
   component: CardComponent,
   template: `
--- a/libs/organization/src/lib/organization/components/card-minimal/card-minimal.component.html
+++ b/libs/organization/src/lib/organization/components/card-minimal/card-minimal.component.html
@@ -1,0 +1,6 @@
+<a matRipple [routerLink]="'/c/o/marketplace/organization/' + org.id">
+  <img [ref]="org.logo" asset="empty_organization.webp" alt="organization logo" />
+</a>
+<div fxLayout="column" fxLayoutAlign="center center">
+  <h5>{{ org | orgName }}</h5>
+</div>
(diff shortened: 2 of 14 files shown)


libs/movie/src/lib/movie/components/card/card.stories.ts
```diff
@@ -2,13 +2,14 @@ import { object } from '@storybook/addon-knobs';
...
 import { CardComponent } from './card.component';
+import { RouterTestingModule } from '@angular/router/testing';
 
 export default {
   title: 'Movie Card'
 };
 
 export const movieCard = () => ({
-  moduleMetadata: { imports: [MovieCardModule, ToolkitModule] },
+  moduleMetadata: { imports: [MovieCardModule, ToolkitModule, RouterTestingModule] },
```
fritzschoff, line 12: nice!

libs/organization/src/lib/organization/components/card-minimal/card-minimal.component.html
```diff
@@ -0,0 +1,6 @@
+<a matRipple [routerLink]="'/c/o/marketplace/organization/' + org.id">
```
fritzschoff, line 1: make it a button? like `<a mat-button>` 
Should not intersect with the design

fritzschoff: and remove then mat ripple from module

libs/organization/src/lib/organization/components/card-minimal/card-minimal.component.scss
```diff
@@ -0,0 +1,23 @@
+:host {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+a {
+  width: 180px;
+  height: 180px;
```
fritzschoff, line 9: nope, forbidden. The size gets determine by the layout directive. There is layout => flex => col, you can see an example at the bf-carousel component in storybook. So you don't need fixed sizes here

libs/organization/src/lib/organization/components/card-minimal/card-minimal.module.ts
```diff
@@ -0,0 +1,24 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
```
fritzschoff, line 2: not mandatory, but if you want you can sort your imports and comment the section.

libs/organization/src/lib/organization/components/card-minimal/card-minimal.component.scss
fritzschoff: https://material.io/design/layout/responsive-layout-grid.html#columns-gutters-and-margins

libs/organization/src/lib/organization/components/card-minimal/card-minimal.component.scss
```diff
@@ -0,0 +1,21 @@
...
+}
+
+a {
+  padding: 24px 16px 16px;
+  border-radius: 50%;
+  margin-bottom: 16px;
+
+  img {
+    width: 100%;
+    border-radius: 50%;
```
fritzschoff, line 14: why two times border-radius? I guess in the image it should be enough!

libs/organization/src/lib/organization/components/card-minimal/card-minimal.component.scss
```diff
@@ -0,0 +1,21 @@
...
+  }
+}
+
+h5 {
+  text-align: center;
```
fritzschoff, line 19: you already set it to center via the fxLayout directive in the div, I guess you can remove it here

Issue #3180, opened 2020-06-24
Change link for "A new organization has been created" email


Jackseed:
- [ ] link to `https://archipelmarket.com/c/o/admin/panel/organization/orgId` instead of `https://archipelcontent.com/c/o/admin/panel/organization/orgId`

phpgeekfr:
@maltmayer @Jackseed I changed default from content to market. But keep in mind that admin panel will soon be a standalone app, so the URL choice content/market will not be an issue anymore.

PR #3185, opened 2020-06-25
#3180 changed default app url in admin emails

phpgeekfr:


apps/backend-functions/src/templates/mail.ts
```diff
@@ -51,7 +51,13 @@ export function userResetPassword(email: string, link: string): EmailTemplateReq
  * @param pageURL
  * @param templateId
  */
-export function userInvite(email: string, password: string, orgName: string, pageURL: string = appUrlContent, templateId: string = templateIds.user.credentials.joinOrganization.catalog): EmailTemplateRequest {
+export function userInvite(
+  email: string,
+  password: string,
+  orgName: string,
+  pageURL: string = appUrlMarket, // @TODO #3081 appUrlAdmin
+  templateId: string = templateIds.user.credentials.joinOrganization.catalog
```
Mussuranae, line 59: joinOrganization.catalog ? not festival ? Is there a template for every app or just one for the moment ?

phpgeekfr: This is a default value that should be overrided. Since "festival" as become the main app, I updated the default

phpgeekfr: There is 4 templates for this mail : 
```
    credentials: {
      attendEvent: {
        catalog: 'd-ce3e57248a694cefacad49bc4c820078',
        festival: 'd-ce3e57248a694cefacad49bc4c820078',
      },
      joinOrganization: {
        catalog: 'd-a34ce9ea59c5477f9feae8f556157b6b',
        festival: 'd-f0c4f1b2582a4fc6ab12fcd2d7c02f5c',
      }
    },
```

PR #3199, opened 2020-06-26
Migration Watermark to new ImRef interface

RemcoSimonides:
Migration Watermark

### DIFF
--- a/apps/backend-ops/src/firestoreMigrations/0003.ts
+++ b/apps/backend-ops/src/firestoreMigrations/0003.ts
@@ -87,9 +87,9 @@ export async function updatePosterStructure(db: Firestore) {
     const movieData = movieDocSnapshot.data();
 
     if (movieData.main && movieData.main.poster) {
-      const media = createImgRef(movieData.main.poster);
+      const media = createOldImgRef(movieData.main.poster);
 
-      const moviePoster = createPromotionalElement({
+      const moviePoster = createOldPromotionalElement({
         label: 'Poster',
         media,
       });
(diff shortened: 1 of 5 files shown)


apps/backend-ops/src/firestoreMigrations/0003.ts
```diff
@@ -115,3 +114,47 @@ export async function upgrade(db: Firestore) {
...
+      original: '',
+      xs: '',
+      md: '',
+      lg: '',
+    },
+    ..._ref
+  };
+}
+
+interface PromotionalElement {
```
pldespaigne, line 142: could you rename it to something like `OldPromotionalElement`

pldespaigne: just to be super clear in the migration ^^

apps/backend-ops/src/firestoreMigrations/0025.ts
```diff
@@ -208,3 +205,35 @@ const updateImgRef = async (
     return EMPTY_REF;
   }
 }
+
+interface ImgRef {
```
pldespaigne, line 209: pleas also add `Old` for all these interfaces

apps/backend-ops/src/firestoreMigrations/0028.ts
```diff
@@ -0,0 +1,40 @@
+import { Firestore } from "../admin";
+import { PublicUser } from "@blockframes/user/types";
+import { createHostedMedia } from "@blockframes/media/+state/media.model";
+
+/**
+ * Migrate old ImgRef objects to new one.
```
pldespaigne, line 6: `Migrate old watermarks into new ones (HostedMedia)`

PR #3214, opened 2020-06-27
Debug work

Mussuranae:
Reload all tabs when logout from one, display organization card only if org has movie available, little stuff
Related to #2782


libs/organization/src/lib/organization/components/card/card.component.ts
```diff
@@ -25,6 +24,6 @@ export class OrganizationCardComponent {
     return this._org;
   }
 
-  constructor(private movieQuery: MovieQuery){}
+  constructor(){}
```
joanofdart, line 27: If there's nothing to be consumed, we shouldn't need a constructor at all

PR #3218, opened 2020-06-30
Feat 3081 - create crm (former admin panel) app

phpgeekfr:


angular.json
```diff
@@ -1193,6 +1193,137 @@
...
+            ],
+            "styles": [
+              "node_modules/angular-calendar/css/angular-calendar.css",
+              "apps/crm/crm/src/styles.scss"
+            ],
+            "scripts": []
+          },
+          "configurations": {
+            "production": {
+              "fileReplacements": [
```
GrandSchtroumpf, line 1237: check how it's done with the other app, but I think you'll need to update that.

angular.json
```diff
@@ -1193,6 +1193,137 @@
...
+              "apps/crm/crm/src/favicon.ico",
+              "apps/crm/crm/src/assets",
+              {
+                "glob": "**/*",
+                "input": "./libs/ui/src/lib/assets",
+                "output": "./assets"
+              }
+            ],
+            "styles": [
+              "node_modules/angular-calendar/css/angular-calendar.css",
```
GrandSchtroumpf, line 1230: Do you need the calendar ?

angular.json
```diff
@@ -1193,6 +1193,137 @@
...
+              "apps/crm/crm/src/styles.scss"
+            ],
+            "scripts": []
```
GrandSchtroumpf, line 1233: You might need to add `@firebase` in somewhere here. Something like `externalDependancies`

apps/crm/crm/src/app/app.module.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import { MatNativeDateModule } from '@angular/material/core';
+
+
+@NgModule({
+  declarations: [AppComponent],
+  imports: [
+    
+    // Angular
+    BrowserModule.withServerTransition({ appId: 'crm-app' }),
```
GrandSchtroumpf, line 41: Did you created a ssr or prerendering ?

apps/crm/crm/src/styles.scss
```diff
@@ -0,0 +1,36 @@
+/* You can add global styles to this file, and also import other style files */
+@import "../../../../styles/style.scss";
+
+.light-theme {
+  @include angular-material-theme($bf-light-theme);
+  @include sat-datepicker-theme($bf-light-theme);
```
GrandSchtroumpf, line 6: You can remove that

apps/crm/crm/src/styles.scss
```diff
@@ -0,0 +1,36 @@
...
+  @include bf-material-theme($bf-light-theme);
+  @include bf-calendar-theme($bf-light-theme);
```
GrandSchtroumpf, line 8: Same here I think.

angular.json
phpgeekfr: It seams that this is only used with festival "server".

Issue #3200, opened 2020-06-26
[Festival] Implement elaborated featured sales agent section on homepage  


RemcoSimonides:
## Description

![image](https://user-images.githubusercontent.com/27687382/85863086-ef9fff00-b7c2-11ea-8437-038b0ce9db3a.png)

## Acceptance Criteria
- [x] Add new component
- [x] Add elaborated featured sales agent section on homepage

## UI
[Invision](https://projects.invisionapp.com/d/#/console/19607961/413926229/preview)

RemcoSimonides:
There is no 'detailed company description' field yet in the database. This will be later added and then needs to be added to this component too

PR #3224, opened 2020-07-01
Sales agent banner component

RemcoSimonides:
Part 1 issue #3200
- [x] Add Featured Sales Agent Component


libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,40 @@
+<div fxLayout="row" fxLayoutAlign="space-between start" flex class="movie-header-slider">
+    <h1>{{ _org | orgName }}</h1>
+    <a mat-button [routerLink]="'/c/o/marketplace/organization/' + _org.id" color="primary">
+        <span>See all ({{ _movies.length }})</span>
+    </a>
+</div>
+<div fxLayout="row" fxLayout.lt-md="column" fxLayoutAlign="start start" flex>
```
fritzschoff, line 7: make this a section

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,40 @@
...
+    <section [col]="6">
```
fritzschoff, line 8: this an article

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,40 @@
...
+        <section fxLayoutAlign="start center">
```
fritzschoff, line 9: this a div

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,40 @@
...
+            <a mat-button [routerLink]="'/c/o/marketplace/organization/' + _org.id">
+                <bf-img [ref]="_org.logo" asset="empty_organization.webp" alt="organization logo"></bf-img>
+            </a>
+            <article>
```
fritzschoff, line 13: div as well

libs/organization/src/lib/organization/components/banner/banner.component.scss
```diff
@@ -0,0 +1,13 @@
+bf-img {
```
fritzschoff, line 1: I don't think that works for every sizes. Have you tried to remove it and let the col, layout and flex do their job?

libs/organization/src/lib/organization/components/banner/banner.component.scss
```diff
@@ -0,0 +1,13 @@
...
+    height: 120px;
+    width: 120px;
+}
+
+section {
+    width: 100% !important;
```
fritzschoff, line 7: why important?

libs/organization/src/lib/organization/components/banner/banner.component.ts
```diff
@@ -0,0 +1,27 @@
...
+@Component({
+  selector: '[org] org-banner',
+  templateUrl: './banner.component.html',
+  styleUrls: ['./banner.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class OrganizationBannerComponent {
+  
+  _org: Organization;
+  _movies: Movie[] = []
```
fritzschoff, line 14: do something like that ``` 

  @Input()
  get slideDirection() { return this._slideDirection }
  set slideDirection(value: Slider['slideDirection']) {
    this.slideDirection$.next(value);
    this._slideDirection = value;
  }
``` 
and make the values that starts with an underline, private

File: libs/organization/src/lib/organization/components/banner/banner.component.ts [hunk 8]
```diff
@@ -0,0 +1,27 @@
...
+  @Input()
+  set org(org: Organization) {
+    this._org = org;
+    if (!!org.movies) {
+      this._movies = org.movies
+        .filter(movie => movie?.main.storeConfig.status === 'accepted' && movie?.main.storeConfig.appAccess.festival);
+    }
+  }
+
+  constructor() {}
```
fritzschoff, line 25: remove if you don't need it

libs/organization/src/lib/organization/components/banner/banner.module.ts
```diff
@@ -0,0 +1,27 @@
+// Angular
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { RouterModule } from '@angular/router';
+import { FlexLayoutModule } from '@angular/flex-layout';
+
+// BlockFrames
```
fritzschoff, line 7: nice!!!

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,40 @@
+<div fxLayout="row" fxLayoutAlign="space-between start" flex class="movie-header-slider">
+    <h1>{{ _org | orgName }}</h1>
+    <a mat-button [routerLink]="'/c/o/marketplace/organization/' + _org.id" color="primary">
+        <span>See all ({{ _movies.length }})</span>
```
GrandSchtroumpf, line 4: You don't need a span inside an anchor, except if there is an icon with it.

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,40 @@
+<div fxLayout="row" fxLayoutAlign="space-between start" flex class="movie-header-slider">
+    <h1>{{ _org | orgName }}</h1>
+    <a mat-button [routerLink]="'/c/o/marketplace/organization/' + _org.id" color="primary">
```
GrandSchtroumpf, line 3: Use the standard array synthax with using [routerLink] with variables : 
`[routerLink]="[/c/o/marketplace/organization/', _org.id]"`

Are we sure it'll always be used in the marketplace ?

libs/organization/src/lib/organization/components/banner/banner.stories.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export const organizationMinimalCard = () => ({
+  moduleMetadata: { imports: [OrganizationBannerModule, ToolkitModule, MatLayoutModule, RouterTestingModule] },
+  name: 'Featured Sales Agent Banner',
+  component: OrganizationBannerComponent,
+  template: `
+    <storybook-toolkit>
+      <h1 title>Featured Sales Agent Banner</h1>
+      <div layout flex>
```
GrandSchtroumpf, line 19: the org-banner should be self suffisant, it should integrate the `layout` inside itself, else we'll get issues in the future

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,40 @@
...
+        <span>See all ({{ _movies.length }})</span>
+    </a>
+</div>
+<div fxLayout="row" fxLayout.lt-md="column" fxLayoutAlign="start start" flex>
+    <section [col]="6">
+        <section fxLayoutAlign="start center">
+            <a mat-button [routerLink]="'/c/o/marketplace/organization/' + _org.id">
```
GrandSchtroumpf, line 10: Same an above.
I don't think you want a mat-button around an image. A simple `matRipple` should be enough.

libs/organization/src/lib/organization/components/banner/banner.stories.ts
RemcoSimonides: `layout` has been added to avoid the following error:
`ERROR NullInjectorError: StaticInjectorError(DynamicModule)[Column -> Layout]: 
  StaticInjectorError(Platform: core)[Column -> Layout]: 
    NullInjectorError: No provider for Layout!`

Do you know how to resolve this otherwise?

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,27 @@
...
+    <h1>{{ org | orgName }}</h1>
+    <a mat-button [routerLink]="orgLink" color="primary">
+        <span>See all ({{ movies.length }})</span>
+        <mat-icon svgIcon="arrow_forward"></mat-icon>
+    </a>
+</header>
+<section fxLayout fxLayout.lt-sm="column" flex>
+    <article>
+        <div fxLayoutAlign="start center">
+            <a [col]="3" matRipple [routerLink]="orgLink">
```
GrandSchtroumpf, line 11: remove col="3"

libs/organization/src/lib/organization/components/banner/banner.component.ts
```diff
@@ -0,0 +1,17 @@
...
+@Component({
+  selector: '[org] org-banner',
+  templateUrl: './banner.component.html',
+  styleUrls: ['./banner.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class OrganizationBannerComponent {
+  
+  @Input() org;
+  @Input() orgLink: string;
```
GrandSchtroumpf, line 14: I don't think it's a good idea to path the link. The page to an organisation should always be the same. From what I understand of this component, what you got before was good.

libs/organization/src/lib/organization/components/banner/banner.stories.ts
GrandSchtroumpf: You don't need this layout anymore, you can remove it.

RemcoSimonides: Ah of course!

libs/organization/src/lib/organization/components/banner/banner.component.ts
RemcoSimonides: As I understood from Anastasiia, this component will be reused on home page of other apps too, so then `/c/o/marketplace/organization` won't be valid...

GrandSchtroumpf: Yes it will because it'll always be the same route for every app

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,27 @@
...
+    <h1>{{ org | orgName }}</h1>
+    <a mat-button [routerLink]="orgLink" color="primary">
+        <span>See all ({{ movies.length }})</span>
+        <mat-icon svgIcon="arrow_forward"></mat-icon>
+    </a>
+</header>
+<section fxLayout fxLayout.lt-sm="column">
+    <article>
+        <div fxLayoutAlign="start center">
+            <a matRipple [routerLink]="orgLink">
```
fritzschoff, line 11: out of consistency, I guess it would be nice if you can use `mat-button` instead of mat-ripple. Since we use it in our repo like that.

fritzschoff: image
Maybe you can try to give the organization description a width of 100% and the upper div with the image and the movies length a flex shrink. Since all the organization photos should have the same size, because we push them through the resize cloud funcrtions, you shouldn't have any unexpected behavior.

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,27 @@
+<header fxLayout fxLayoutAlign="space-between start">
+    <h1>{{ org | orgName }}</h1>
+    <a mat-button [routerLink]="orgLink" color="primary">
+        <span>See all ({{ movies.length }})</span>
+        <mat-icon svgIcon="arrow_forward"></mat-icon>
+    </a>
+</header>
+<section fxLayout fxLayout.lt-sm="column">
+    <article>
```
fritzschoff, line 9: you sure you need this article?

libs/organization/src/lib/organization/components/banner/banner.component.scss
```diff
@@ -0,0 +1,45 @@
...
+bf-img {
+    width: 100%;
+    border-radius: 50%;
+}
+
+article {
+    width: 50%;
+
+    &.org-info {
+        margin-left: 24px;
```
fritzschoff, line 21: why don't you use the flex directive instead of hardcoding it?

libs/organization/src/lib/organization/components/banner/banner.module.ts
```diff
@@ -0,0 +1,39 @@
...
+    CommonModule,
+    RouterModule,
+    FlexLayoutModule,
+
+    // Blockframes
+    OrgNameModule,
+    ImageReferenceModule,
+    MovieCardModule,
+    MatLayoutModule,
+    MatButtonModule,
```
fritzschoff, line 30: nice layouting!!! but move the mat into the material section

libs/organization/src/lib/organization/components/banner/banner.module.ts
```diff
@@ -0,0 +1,39 @@
...
+import { CommonModule } from '@angular/common';
+import { RouterModule } from '@angular/router';
+import { FlexLayoutModule } from '@angular/flex-layout';
+
+// BlockFrames
+import { ImageReferenceModule } from '@blockframes/media/directives/image-reference/image-reference.module';
+import { OrganizationBannerComponent } from './banner.component';
+import { OrgNameModule } from '../../pipes/org-name.pipe';
+import { MovieCardModule } from '@blockframes/movie/components/card/card.module';
+import { MatLayoutModule } from '@blockframes/ui/layout/layout.module';
```
fritzschoff, line 12: here as well

libs/organization/src/lib/organization/components/banner/banner.component.scss
```diff
@@ -0,0 +1,45 @@
+@import '~@angular/flex-layout/mq';
+
+:host {
+    display: block;
+}
+
+header {
+    margin-left: 24px;
```
fritzschoff, line 8: seems a case for flex directive, since it will not always be 24 px

libs/organization/src/lib/organization/components/banner/banner.component.scss
```diff
@@ -0,0 +1,45 @@
...
+
+article {
+    width: 50%;
+
+    &.org-info {
+        margin-left: 24px;
+        padding-right: 24px;
+
+        .org-image {
+            width: 20%;
```
fritzschoff, line 25: how does it look on small screens?

RemcoSimonides: Pretty good since it leaves enough space for the organization name next to it! :)

libs/organization/src/lib/organization/components/banner/banner.component.scss
```diff
@@ -0,0 +1,32 @@
+@import '~@angular/flex-layout/mq';
+
+:host {
+    display: block;
+}
+
+bf-img {
+    width: 100%;
```
GrandSchtroumpf, line 8: Are you sure ? It feels weird to set a % on the width in this case. I think a fixed value in px that you update in responsive would more efficient.

libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -0,0 +1,27 @@
+<header fxLayout fxLayoutAlign="space-between start" flex>
+    <h1>{{ org | orgName }}</h1>
+    <a mat-button [routerLink]="['/c/o/marketplace/organization/', org.id]" color="primary">
+        <span>See all ({{ movies.length }})</span>
+        <mat-icon svgIcon="arrow_forward"></mat-icon>
+    </a>
+</header>
+<section fxLayout fxLayout.lt-sm="column" flex>
```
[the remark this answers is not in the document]
RemcoSimonides: You've suggested to use a scss variable instead of flex. Could you elaborate on that idea?

libs/movie/src/lib/movie/pipes/movies-from-org.pipe.ts
```diff
@@ -0,0 +1,19 @@
+import { Pipe, NgModule } from "@angular/core";
+import { Movie } from "../+state";
+import { Organization } from "@blockframes/organization/+state";
+
+@Pipe({
+    name: 'fromOrg',
+})
+export class MovieFromOrgPipe {
```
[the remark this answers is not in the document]
RemcoSimonides: It depends on how the component will be implemented. I guess that's for the PR in which the Component is added to the homepage so I remove it for now.

libs/organization/src/lib/organization/components/banner/banner.component.html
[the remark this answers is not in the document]
GrandSchtroumpf: In the styles/style.scss you can create a css variable : 
```css
:root {
  --layout-margin: 24px;
}
```
Then you can use it in any scss file: 
```css
section {
  margin: 0 var(--layout-margin);
}
```
We could talk with Stéphanie concerning the list of variables we would like to keep.

libs/movie/src/lib/movie/pipes/movies-from-org.pipe.ts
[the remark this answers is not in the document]
GrandSchtroumpf: If you don't use the pipe anymore, remove it. It's better than keeping and forgetting it.

PR #3226, opened 2020-07-01
Feat 3081 - added hosting target 

phpgeekfr:


apps/backend-functions/src/templates/mail.ts
```diff
@@ -168,22 +167,22 @@ export function requestToAttendEventFromUser(
 /**
  * @param orgId
  */
-const organizationCreatedTemplate = (orgId: string, appUrl: string = appUrlMarket) => // @TODO #3081 appUrlAdmin
+const organizationCreatedTemplate = (orgId: string) =>
   `
   A new organization was created on the blockframes project,
 
-  Visit ${appUrl}${ADMIN_ACCEPT_ORG_PATH}/${orgId} or go to ${ADMIN_ACCEPT_ORG_PATH}/${orgId} to view it.
+  Visit ${appUrlCrm}${ADMIN_ACCEPT_ORG_PATH}/${orgId} or go to ${ADMIN_ACCEPT_ORG_PATH}/${orgId} to view it.
```
GrandSchtroumpf, line 174: what about appUrl.crm ?

phpgeekfr: OK

Issue #3211, opened 2020-06-26
[Bug] Cannot read property 'toUpperCase' of undefined


Jackseed:
## Description
One of the most common error we have (1.1k events, 175 users).
[Sentry details](https://sentry.io/organizations/cascade-8/issues/1733337807/?project=1540126&referrer=slack)

## Acceptance criteria
- [ ] fix this :)

PR #3228, opened 2020-07-01
APP-3211 Uppercase a string.

joanofdart:
Related to #3211


libs/utils/src/lib/pipes/display-name.pipe.ts
```diff
@@ -6,7 +6,7 @@ import { Person } from '../common-interfaces';
 })
 export class DisplayNamePipe implements PipeTransform {
   transform(value: Person | Person[]): string {
-    const capitalize = (str: string) => str[0].toUpperCase() + str.slice(1);
+    const capitalize = (str: string) => str.toUpperCase();;
```
GrandSchtroumpf, line 9: Remove double ";;" & maybe add a "?" in str?.toUpperCase

Issue #3206, opened 2020-06-26
Change email address for CI emails


Jackseed:
## Description
Currently the email used is team@cascade8.com which is quite disturbing cause the emails are mixed up with prod ones. 

## Acceptance criteria
Use developers@cascade8.com:
- [ ]  for CI admin emails (org creation, first user connexion,..)
- [ ] for staging emails
- [ ] more generally for every emails that don't come from prod

GrandSchtroumpf:
Currently the value is in [libs/utils/src/lib/apps.ts](https://github.com/blockframes/blockframes/blob/develop/libs/utils/src/lib/apps.ts#L21).
We need to movie it to the `env.***.ts` with specific addresses depending on the environement.

GrandSchtroumpf:
Find where this email is used. Might be in: 
- Circle CI environement variable
- secret.sh from staging

joanofdart:
Hey,
I've been doing some research and found some stuff but I need clarification:
````typescript
export const sendgridEmailsFrom = {
  catalog: { email: 'team@archipelcontent.com', name: 'Archipel Content' },
  festival: { email: 'team@archipelmarket.com', name: 'Archipel Market' },
  default: { email: 'team@cascade8.com', name: 'Cascade 8' }
} as const;
```
We are defining a list of emails from which we will send. Theres a default block but I don't know if this is related to the CI stuff?
Also..
```typescript
export function getSendgridFrom(a?: App): EmailData {
  if (!a) {
    return sendgridEmailsFrom.default;
  } else {
    return sendgridEmailsFrom[a] || sendgridEmailsFrom.default;
  }
}
```
Again, we are using the default email but I don't think this is related to CI at all. Just commenting here to make sure.

Last but not least, there's an ENV variable in circleCI which is **CASCADE8_ADMIN**, I'm wondering if this is the email @Jackseed was referring to?

FYI,
@GrandSchtroumpf

Jackseed:
@joanofdart sorry I your comment just now. 
In fact I'm not sure your fix was the correct one. We still receive emails from CI to team@cascade8.com. I have the feeling that now the emails sent from dev projects are received to developers@cascade8.com (because we do receive some of them to that address now).
Also to that issue we should include mails from staging, e.g. every mails not coming from prod should go to developers@cascade8.com so we don't mix up with client emails.

phpgeekfr:
Not closing this until we decide to use blockframes.io for email other than prod.

[...]

PR #3236, opened 2020-07-02
fix(adminEmail): added a **dev** check to see wether we are currently…

joanofdart:
… in dev and if not, we are setting the email to 'developers@cascade8.com'

Related to #3206


env/env.prod.ts
```diff
@@ -1,4 +1,5 @@
 export const production = true;
+export const dev = false;
```
Mussuranae, line 2: So for prod, it will be also developers@cascade8.com ?

apps/backend-functions/src/environments/environment.ts
```diff
@@ -41,7 +41,9 @@ export const algolia = {
   adminKey: mockConfigIfNeeded('algolia', 'api_key')
 };
 
-export const adminEmail = mockConfigIfNeeded('admin', 'email');
+
+
+export const adminEmail = dev ? mockConfigIfNeeded('admin', 'password'): 'developers@cascade8.com';
```
GrandSchtroumpf, line 46: isn't it the opposite ?

joanofdart: local => **mockConfigIfNeeded('admin', 'password');**
staging => **'developers@cascade8.com'**
prod => environment.prod.ts => **functions.config().admin.email;**

env/env.prod.ts
joanofdart: In prod we have the email assigned directly, e.g:
- **adminEmail = functions.config().admin.email;**

Issue #3229, opened 2020-07-02
Extract medias from Movie/Org/User forms


pldespaigne:
Right now we have a separate form for the firestore data (user/org/movie) and medias.
We need to merge those forms, then extract medias before updating, so as to save data without media and only upload the medias.

- [x] merge form for the user
- [x] merge form for the org
- [x] merge form for the movie

link to main issue #3084

PR #3240, opened 2020-07-03
Extract medias from Movie/Org/User forms

pldespaigne:
solve #3229 
- `HostedMediaForm` is now included inside user/org/movie forms
- we extract needed info before update & upload

### DIFF
--- a/libs/media/src/lib/components/upload/file-upload/file-upload.component.ts
+++ b/libs/media/src/lib/components/upload/file-upload/file-upload.component.ts
@@ -8,6 +8,7 @@ import {
 import { HostedMediaForm } from '@blockframes/media/directives/media/media.form';
 import { MatSnackBar } from '@angular/material/snack-bar';
 import { getMimeType } from '@blockframes/utils/file-sanitizer';
+import { getFileNameFromPath } from '@blockframes/media/+state/media.model';
 
 @Component({
   selector: '[form] [storagePath] file-upload',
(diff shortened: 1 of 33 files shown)


libs/media/src/lib/components/upload/file-upload/file-upload.component.ts
```diff
@@ -59,7 +60,11 @@ export class FileUploadComponent implements OnInit {
   public selected(files: FileList) {
     if (!files.item(0)) {
       this.snackBar.open('No file found', 'close', { duration: 1000 });
-      this.state = 'waiting';
+      if (!!this.form.oldRef.value) {
```
GrandSchtroumpf, line 63: What about : `this.state = this.form.oldRef.value ? 'file' : 'waiting'` ?

libs/media/src/lib/components/upload/file-upload/file-upload.component.ts
```diff
@@ -92,14 +97,30 @@ export class FileUploadComponent implements OnInit {
     this.form.markAsDirty();
   }
 
-  public reset() {
+  public delete() {
     this.state = 'waiting';
 
+    this.form.patchValue({
```
GrandSchtroumpf, line 103: Can you do it on one line ?

libs/media/src/lib/components/upload/file-upload/file-upload.component.ts
```diff
@@ -92,14 +97,30 @@ export class FileUploadComponent implements OnInit {
...
       blobOrFile: undefined,
       delete: false,
-      fileName: '',
+      fileName: !!this.form.oldRef.value ? getFileNameFromPath(this.form.oldRef.value) : '',
     })
     this.form.markAsDirty();
+
+    fileExplorer.value = null;
+
+    if (!!this.form.oldRef.value) {
```
GrandSchtroumpf, line 120: Same as above.

PR #3242, opened 2020-07-03
George/configure emulator testing

xmano:
Setup unit-testing with local emulator and some test helpers.


libs/testing/firebase/src/lib/functions.ts
```diff
@@ -0,0 +1,37 @@
+import * as admin from 'firebase-admin';
+import * as firebaseTesting from '@firebase/testing'
```
GrandSchtroumpf, line 2: Where is this used ?

package.json
```diff
@@ -180,6 +181,7 @@
     "cross-env": "^7.0.2",
     "cypress": "^4.4.0",
     "dotenv": "8.2.0",
+    "firebase-functions-test": "^0.2.1",
```
GrandSchtroumpf, line 184: there is no official library to instead ?
Be careful not to add too many dependancies in the project. It's difficult to maintain.

tools/scripts/check-google-application-credentials.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const CREDENTIALS_DIR = 'tools/credentials/';
+const ENV_FILE = '.env';
+const KEY = 'GOOGLE_APPLICATION_CREDENTIALS';
+
+if (process.env[KEY]) {
+  // * Exists
+  console.log(`Your ${KEY} are successfully set.`);
+} else {
+  // ! Not found
```
GrandSchtroumpf, line 15: I'm not sure to understand what you're doing here, looks like you look for json file under `'tools/credentials/'` and supposed it's the google credentials. I don't see so much in which case this would happen. 

Currently everybody have put their credential somewhere in their computer. So for me if we want to force dev to put their credentials in the tools/credentials folders the best is to : 
```typescript
if (process.env[KEY]) {
  const path = process.env[KEY];
  // If there is another path, copy the file & update the .env.
  if (path !== `./tools/credentials/google-credential.json`) {
    const from = resolve(process.cwd(), path), resolve(process.cwd());
    const to = resolve(process.cwd()), `./tools/credentials/google-credential.json`);
    copyFile(from, to);
    // update the .env file
  }
} else {
  console.error(`No ${KEY} found in ${ENV_FILE}, you can find the documentation here : ${docLink}`);
}
```

tools/scripts/check-google-application-credentials.ts
```diff
@@ -0,0 +1,33 @@
+import { config } from 'dotenv';
+import { readdirSync, appendFileSync } from 'fs';
+import { resolve, join } from 'path';
+
+config();
+
+const CREDENTIALS_DIR = 'tools/credentials/';
+const ENV_FILE = '.env';
+const KEY = 'GOOGLE_APPLICATION_CREDENTIALS';
```
GrandSchtroumpf, line 9: Would be nice to do something generic for each key : 
```typescript
const variables = [{
  key: 'GOOGLE_APPLICATION_CREDENTIALS',
  doc: // link to documentation
}, {
  key: 'ALGOLIA_API_KEY',
  doc: // link to documentation
},...];
for (const variable of variables) {
  if (!process.env[variable.key]) {
    console.error(`No ${variable.key} found in ${ENV_FILE}, you can find the documentation here : ${variable.doc}`);
  }
}
```

tsconfig.json
```diff
@@ -58,7 +58,10 @@
       "@blockframes/e2e/*": ["libs/e2e/src/lib/*"],
       "@blockframes/landing/*": ["libs/landing/src/lib/*"],
       "@blockframes/import": ["libs/import/src/index.ts"],
-      "@blockframes/media/*": ["libs/media/src/lib/*"]
+      "@blockframes/media/*": ["libs/media/src/lib/*"],
+      "@blockframes/testing/seed-data": ["libs/testing/seed-data/src/index.ts"],
```
GrandSchtroumpf, line 62: We can do `"@blockframes/testing/*": ["libs/testing/seed-data/src/*"],`

GrandSchtroumpf:
@george43g try to keep your branches separated. Here there is some work on `.env` variables & some work on the emulator.
Let's remove all sublib files before merging. 
You can run `ng generate @nrwl/workspace:remove --projectName=testing-firebase` to let nx remove all the generated files for you (first backup the content of the ts files you want to keep ^^).

george43g:
Currently working on this. Will provide feedback shortly.

george43g:
this library was not ready for a PR. This was just some preliminary work.

package.json
george43g: this is a necessary official library to test firebase functions.

tools/scripts/check-google-application-credentials.ts
george43g: Regarding docs for each var - that's for a future PR where we manage vars.

tools/scripts/check-google-application-credentials.ts
george43g: Yup, this is a good idea - I should definitely do this - forgot about backwards compatibility.

tsconfig.json
george43g: will discuss this with you shortly

libs/testing/firebase/src/lib/functions.ts
george43g: this is incomplete, but yes will be used to configure DB. 

If each function took the DB as a parameter, or if a module was used to init firebase, it would be much easier to test. Certain structures are impossible to test - so I don't think we can escape improving code in order to make testing possible.

libs/testing/firebase/jest.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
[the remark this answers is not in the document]
george43g: will discuss this with you

tsconfig.json
george43g: I have pointed it to src - that way the default barrel will still work, and to do a deep import you'd say
`@blockframes/testing`
`@blockframes/testing/lib/firebase`
And even deeper
`@blockframes/testing/lib/firebase/functions`

PR #3243, opened 2020-07-03
Migration and unit test

Mussuranae:
First unit test on a migration script


apps/backend-ops/src/firestoreMigrations/0029.ts
```diff
@@ -0,0 +1,50 @@
...
+    return batch.set(movie.ref, newData);
+  })
+
+  await batch.commit();
+  console.log('Create new technical section on movie.');
+}
+
+
+export function createTechnicalSection(movie: Partial<Movie>) {
+  const { salesInfo } = movie;
```
xmano, line 29: Is it possible that movie may not have salesInfo property set? Should you check for the possibility of this not being set?

Issue #3247, opened 2020-07-06
Find way to test migration scripts via emulator


george43g:
George + Mano - discovered this issue and Want to document it.

Find a way to setup unit tests for migration scripts.

The problem is that the migration scripts have hardcoded connecting to the env in env.ts. We need to override this config for tests to ensure they connect to the emulator instead.

xmano:
Issues with current code in **george/configure-emulator-testing** branch

1.  Connection cannot be made to local emulator.
     Debugged this issue and found the cause to be GOOGLE_APPLICATION_CREDENTIALS not being set properly in .env file : check the quotes and path. Path should be relative.

2. After fixing this manually in .env, encounter another config issue.
    
```
    Cannot find module './operations_client_config' from 'operationsClient.js'

    However, Jest was able to find:
        './operations_client_config.json'
```

3. The current setup sometimes seems to download a jar file - this can slowdown tests and may hog the bandwidth. Verify this.

4.  Need a way to inject local emulator config into apps for unit-test mode.

PR #3257, opened 2020-07-08
update(test-case): add a test case to directly connect local-emulator.

xmano:
Add testing library infrastructure #3247 

Currently I am trying to configure the tests to directly exercise the local emulator.

Testing with just single spec functions.specs.ts

Connection is made.
Verifying with other test case to write and read a document.

Debugging the failure :

`
    > 25 |     await expect(db.doc('testCollection/testDoc').get())                                                          
         |           ^                                                                                                       
      26 |           .resolves.toBe({testStr});                                                                              
      27 |   });                                                                                                             
      28 |                                                                                                                   
                                                                                                                             
      at expect (node_modules/expect/build/index.js:138:15)                                                                  
      at libs/testing/src/lib/firebase/functions.spec.ts:25:11                                                               
      at step (libs/testing/src/lib/firebase/functions.spec.ts:33:23)                                                        
      at Object.next (libs/testing/src/lib/firebase/functions.spec.ts:14:53)                                                 
      at libs/testing/src/lib/firebase/functions.spec.ts:8:71                                                                
      at new ZoneAwarePromise (node_modules/zone.js/dist/zone.js:913:33)                                                     
      at Object.<anonymous>.__awaiter (libs/testing/src/lib/firebase/functions.spec.ts:4:12)                                 
      at libs/testing/src/lib/firebase/functions.spec.ts:18:40                                                               
      at ZoneDelegate.Object.<anonymous>.ZoneDelegate.invoke (node_modules/zone.js/dist/zone.js:386:30)                      
      at ProxyZoneSpec.Object.<anonymous>.ProxyZoneSpec.onInvoke (node_modules/zone.js/dist/proxy.js:117:43)                 
      at ZoneDelegate.Object.<anonymous>.ZoneDelegate.invoke (node_modules/zone.js/dist/zone.js:385:36)                      
      at Zone.Object.<anonymous>.Zone.run (node_modules/zone.js/dist/zone.js:143:47)                                         
                                                                                                                             
  console.error node_modules/zone.js/dist/zone.js:690                                                                        
    Unhandled Promise rejection:                                                                                             
    Null value error. for 'get' @ L21 ; Zone: <root> ; Task: setTimeout ; Value: Error [FirebaseError]:                      
    Null value error. for 'get' @ L21                                                                                        
`


libs/testing/src/lib/firebase/functions.spec.ts
```diff
@@ -1,8 +1,31 @@
 import { initFunctionsTestMock } from './functions';
+import * as firebaseTesting from '@firebase/testing'
 import { firestore, initializeApp } from 'firebase-admin';
 
+const project_ID = "blockframes-mano";
```
GrandSchtroumpf, line 5: You don't want to give your real project ID here. You want a random value for each projectId I think, else I fear you might have concurrency issues between tests that are using the same projectId

libs/testing/src/lib/firebase/functions.spec.ts
```diff
@@ -1,8 +1,31 @@
...
+
 describe('firebase testing library', () => {
   test.todo('Hello World!');
+
+  it('should access emulator directly', async () => {
+    const db = firebaseTesting.initializeTestApp({
```
GrandSchtroumpf, line 11: Where are you saying that you're using the emulator for this test ?

libs/testing/src/lib/firebase/functions.spec.ts
```diff
@@ -1,8 +1,31 @@
 import { initFunctionsTestMock } from './functions';
+import * as firebaseTesting from '@firebase/testing'
 import { firestore, initializeApp } from 'firebase-admin';
 
+const project_ID = "blockframes-mano";
+
 describe('firebase testing library', () => {
   test.todo('Hello World!');
```
GrandSchtroumpf, line 8: Please remove that.

libs/testing/src/lib/firebase/functions.spec.ts
```diff
@@ -1,16 +1,54 @@
...
-    const test = 5;
-    await db
-      .collection('test')
-      .doc('test')
-      .set({ test });
-    await expect(db.doc('test/test').get()).resolves.toBe(5);
+
+  describe('Direct access to emulator', () => {
+    it.skip('should access emulator directly', async () => {
+      //TODO : This is for reference
```
GrandSchtroumpf, line 9: In the project, whenever we add a TODO, if you don't solve it in the same PR, you need to create an issue related to this TODO & add the id of the issue in the comment. For example : 
```typescript
// @todo(#1234) Make it reusable
```

libs/testing/src/lib/firebase/functions.spec.ts
```diff
@@ -1,16 +1,54 @@
...
+      //TODO: Debug the test failure.                
+      await expect(db.doc('testCollection/testDoc').get())
+            .resolves.toBe({testStr});
+    });
+  })
+
+  describe('Access emulator through library', () => {
+    let db;
+
+    beforeAll(() => {
```
GrandSchtroumpf, line 33: I think you want that to be `beforeEach` to make sure the DB is clean between each test.
Or you can clean it with `afterEach`. Maybe it's faster 🤔

libs/testing/src/lib/firebase/functions.spec.ts
```diff
@@ -1,16 +1,54 @@
...
+    }) 
+
+    it('should be able to access the emulator', async () => {
+      const testDoc = await db.collection("testCollection").doc("testDoc");
+      firebaseTesting.assertSucceeds(testDoc.get());    
+    });
+
+    it('should be able to read the document', async () => {
+      const testStr = "blockframes";
+      const testDoc = db.collection("testCollection")
```
GrandSchtroumpf, line 45: As you're going to write that a lot I would recommand you to name collection "col", document "doc", subcollection "subCol", subdocument : "subdoc": 
```typescript
const doc = db.doc('col/doc').set({ value });
```

libs/testing/src/lib/firebase/functions.spec.ts
```diff
@@ -1,16 +1,54 @@
...
+                      .doc("testDoc").set({testStr});
+      //TODO: Debug the test failure.      
+      //Expectation failure : Received promise rejected instead of resolved          
+      await expect(db.doc('testCollection/testDoc').get())
```
GrandSchtroumpf, line 49: I think you should do something like : 
```typescript
const doc = await db.doc('testCollection/testDoc').get();
expect(doc.data()).toEqual({ testStr });
```
You cannot use `toBe` as `toBe` check the memory reference and not the deep equality.

george43g: `await expect(db.doc.get).resolves....(use deep equality compare with written object)`

jest.config.js
```diff
@@ -15,14 +15,15 @@ module.exports = {
   transform: {
     '^.+\\.(ts|js|html)$': 'ts-jest'
   },
-  testEnvironment: 'jest-environment-jsdom-thirteen',
+  testEnvironment: 'node',
```
GrandSchtroumpf, line 18: I'm not sure it's a good idea to put it here. What will happen when we run angular unit test ?

libs/testing/src/lib/firebase/functions.spec.ts
```diff
@@ -1,16 +1,17 @@
-import { initFunctionsTestMock } from './functions';
+﻿import * as firebaseTesting from '@firebase/testing'
 import { firestore, initializeApp } from 'firebase-admin';
 
 describe('firebase testing library', () => {
-  test.todo('Hello World!');
   it.skip('should be able to access the emulator', async () => {
-    const firebaseTest = initFunctionsTestMock();
+    //Init test DB
+    const projectId = String(Math.random());
```
GrandSchtroumpf, line 7: Why are you not using the code from George ?

libs/testing/jest.config.js
```diff
@@ -1,9 +1,14 @@
...
   name: 'testing',
   preset: '../../jest.config.js',
   transform: {
     '^.+\\.[tj]sx?$': 'ts-jest'
   },
-  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'html'],
+  //Note: testEnvironment may not be appropriate for some app tests
+  testEnvironment: 'node',
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'html', 'json'],
+  testPathIgnorePatterns: [
```
GrandSchtroumpf, line 10: Remove if not needed.

jest.config.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
     'jest-preset-angular/build/HTMLCommentSerializer.js'
   ],
   testPathIgnorePatterns: [
+    'node_modules/',
```
GrandSchtroumpf, line 26: remove if not needed.

libs/testing/src/lib/firebase/functions.spec.ts
```diff
@@ -1,16 +1,16 @@
...
-    const test = 5;
-    await db
-      .collection('test')
-      .doc('test')
-      .set({ test });
-    await expect(db.doc('test/test').get()).resolves.toBe(5);
+    const docRef = db.collection('testCollection').doc('testDoc');
+    const testData = { name: 'blockframes' };
+    await docRef.set(testData);
+    const doc = await db.doc('testCollection/testDoc').get();
```
GrandSchtroumpf, line 13: You can do `const doc = await docRef.get()` to not repeat yourself.

jest.config.js
xmano: removed and moved to lib config

xmano:
Closes #3247

PR #3260, opened 2020-07-08
Poster banner

pldespaigne:
- [x] poster is not an array anymore but just a single image
- [x] poster & banner are moved from `promotionalElement` to `main`


pldespaigne:
I screwed the git history, closing in favor of #3261

PR #3277, opened 2020-07-13
Issue 3209

fritzschoff:


libs/notification/src/lib/invitation/components/message/message.component.html
```diff
@@ -0,0 +1,14 @@
+<ng-container *ngIf="isEvent else noEvent">
+  <ng-container *ngIf="(invitation | async) as parsedInvitation">
+    <span class="mat-body-1">{{ orgName }} invited you to
```
GrandSchtroumpf, line 3: As written in the [doc](https://material.angular.io/guide/typography), "mat-body-1" is the default style. You shouldn't need it.

libs/notification/src/lib/invitation/components/message/message.component.ts
```diff
@@ -0,0 +1,53 @@
...
+          this._invitation = `Your organization sent an invitation to this user email: ${invitation.toUser.email}.`;
+        }
+        break;
+      case 'attendEvent':
+        this.isEvent = true;
+        if (invitation.mode === 'request') {
+          let from: string;
+          if (invitation.fromOrg) {
+            from = invitation.fromOrg.denomination.public ? invitation.fromOrg.denomination.public : invitation.fromOrg.denomination.full;
+          } else if (invitation.fromUser) {
```
GrandSchtroumpf, line 38: In this specific case it's always going to be fromOrg

libs/notification/src/lib/invitation/components/message/message.component.html
```diff
@@ -0,0 +1,14 @@
...
+      <a color="primary" [routerLink]="['/c/o/marketplace/event/' + parsedInvitation.id]" id="link">
```
GrandSchtroumpf, line 4: Use the array format with ","

libs/notification/src/lib/invitation/components/message/message.component.ts
```diff
@@ -0,0 +1,53 @@
...
+  @Input()
+  get invitation() { return (this._invitation as any) }
+  set invitation(invitation: Invitation) {
+    switch (invitation.type) {
+      case 'joinOrganization':
+        this.isEvent = false;
+        if (invitation.mode === 'request') {
+          this._invitation = `${invitation.fromUser.firstName} ${invitation.fromUser.lastName} wants to join your organization.`;
+        } else {
+          this._invitation = `Your organization sent an invitation to this user email: ${invitation.toUser.email}.`;
```
GrandSchtroumpf, line 29: This is not an invitation message. I'm pretty sure you should not cover this case.

libs/notification/src/lib/invitation/components/message/message.component.ts
```diff
@@ -0,0 +1,53 @@
...
+  styleUrls: ['message.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class MessageComponent {
+
+  public orgName: string;
+
+  public isEvent = false;
+
+  private _invitation: Observable<Event> | string;
```
GrandSchtroumpf, line 18: This is not a good practice to have either an Observable or a string. It's super difficult to understand what happens.

libs/notification/src/lib/invitation/components/message/message.component.html
GrandSchtroumpf: Also, what happens if you're on dashboard ? You want to redirect to dashboard. I think there is a pipe for link like that.

Issue #3230, opened 2020-07-02
Create a HostedMediaFormList


pldespaigne:
We need a generic way to handle list of hosted medias so as to hide the complexity from the front.

link to main issue #3084

PR #3279, opened 2020-07-13
Media list

pldespaigne:
solve issue #3230


apps/festival/festival/src/app/dashboard/tunnel/media-image/media-image.component.ts
```diff
@@ -11,8 +12,14 @@ import { MovieTunnelComponent } from '../movie-tunnel.component';
 export class MediaImageComponent {
   form = this.tunnel.form;
 
+  public movie = this.movieQuery.getActive();
```
GrandSchtroumpf, line 15: Looks like you only need the id. Also, I'm wondering is using the routerquery wouldn't be a better idea if we get rid of akita for movie at this level.

apps/festival/festival/src/app/dashboard/tunnel/media-image/media-image.component.ts
```diff
@@ -11,8 +12,14 @@ import { MovieTunnelComponent } from '../movie-tunnel.component';
...
+
   constructor(private tunnel: MovieTunnelComponent, private movieQuery: MovieQuery) {}
 
-  public movie = this.movieQuery.getActive();
+  addStill() {
+    this.form.promotionalElements
+      .get('still_photo')
+      .push(new MoviePromotionalImageForm({ label: 'Still' }));
```
GrandSchtroumpf, line 22: Maybe you could create a `add` method on the form so you can do : 
```typescript
this.form.promotionalElements.get('still_photo').add({ label: 'Still' });
```
Also I thought we got rid of the label ..

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -120,10 +120,17 @@ export function createMoviePromotionalElements(
   params: Partial<MoviePromotionalElements> = {},
   initDefault: boolean = true
 ): MoviePromotionalElements {
+
+  const newStills: Record<string, PromotionalImage> = {};
+  Object.keys(params.still_photo)
```
GrandSchtroumpf, line 125: Use for in if you do `Object.keys().forEach`
```typescript
for (const key in params.still_photo) {
  newStills[key] = createPromotionalImage(params.still_photo[key]);
}
```

libs/utils/src/lib/form/forms/entity-list.form.ts
```diff
@@ -0,0 +1,23 @@
+import { FormEntity, EntityControl } from './entity.form';
+
+export class FormEntityList<C extends EntityControl<Record<string, T>>, T = any> extends FormEntity<C, Record<string, T>> {
```
GrandSchtroumpf, line 3: I think you should create a class specific for list of media. I don't see the point of making it generic, it'll create more garbage code than anything.

libs/movie/src/lib/movie/form/promotional-elements/promotional-elements.form.ts
```diff
@@ -84,10 +85,10 @@ function createMoviePromotionalElementsControls(promotionalElements?: Partial<Mo
 
   return {
     // Images
-    still_photo: new FormGroup(stillPhotoControls),  // TODO issue#3230
+    still_photo: new FormEntityList<Record<string, MoviePromotionalImageForm>>(stillPhotoControls),
 
     // Hosted Media
-    trailer: new FormGroup(trailerControls), // TODO issue#3230
+    trailer: new FormGroup(trailerControls), // TODO issue#3230 // TODO SHOULD WE DELETE THAT ????
```
GrandSchtroumpf, line 91: Yes we should

apps/backend-functions/src/internals/image.ts
```diff
@@ -102,3 +102,7 @@ export async function handleImageChange(after: ImgRef) {
   }
 
 }
+
+export function isEmptyImage(image: ImgRef) {
+  return ['original', ...imgSizeDirectory].every(key => !image[key as keyof ImgRef]?.ref);
```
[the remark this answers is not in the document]
pldespaigne: this doesn't work because `every()` expect a `key: string` and typescript says that `keyof ImgRef` and `string` are incompatible

libs/utils/src/lib/form/forms/entity-list.form.ts
pldespaigne: I made a generic class because we use this for images (aggregation of 5 medias (original, fallback, lg, ...) ) but we could also need it for simple media (like a list of pdf for example).

Thus we don't know what kind of media form to put in the list, I thought it was better to make it generic.

I can still move this code to the media lib though.

libs/media/src/lib/directives/media/media-list.form.ts
```diff
@@ -0,0 +1,24 @@
+import { EntityControl, FormEntity } from "@blockframes/utils/form/forms/entity.form";
+
+// TODO when we will transition to ImgIx we can remove the generic part of this form issue#3283
+export class MediaFormList<C extends EntityControl<Record<string, T>>, T = any> extends FormEntity<C, Record<string, T>> {
```
GrandSchtroumpf, line 4: Maybe it makes more sense to put it in the media/forms folder instead of media/directives.

Issue #3096, opened 2020-06-16
Migration script


Jackseed:
Mirgate every things !

## Database Migration
Only migrate the db schema, here we don't care about the storage, or about wrong ref & ulr.
Those will be tackled in other migrations steps.

- [x] `user`
  - [x] `avatar` : `ImgRef` [@RemcoSimonides ]
  - [x] `watermark` : `HostedMedia` [@RemcoSimonides ]
- [x] `org.logo` : `ImgRef` [@RemcoSimonides ]
- [x] `movie.promotionalElements`
  - [x] `banner.media` : `ImgRef` -> move to `main`  [@RemcoSimonides ]
  - [x] `poster.media` : Array [0] -> single ImgRef -> move to `main` migration [@RemcoSimonides ]
  - [x] `presentation_deck.media` : `HostedMedia` [@RemcoSimonides ]
  - [x] `scenario.media` : `HostedMedia` [@RemcoSimonides ]
  - [x] `promo_reel_link.media` : `ExternalMedia` [@fritzschoff  ]
  - [x] `screener_link.media` : `ExternalMedia` [@fritzschoff  ]
  - [x] `teaser_link.media` : `ExternalMedia` [@fritzschoff  ]
  - [x] `trailer_link.media` : `ExternalMedia` [@fritzschoff  ]
  - [x] `trailer.media` : Remove [@RemcoSimonides ]
  - [x] `still_photo.media`  : Record migration, then ImgRef migration [@RemcoSimonides ]
  

### Records migration
Transform every `movie.promotionalElements` arrays into `Record<string, ...>`.
You should not change the inner elements, this will be done later depending on their types.
You can use the index teh string key of the record :
```TypeScript
poster: [
  { poster0 },
  { poster1 },
  { poster2 },
  // ...
]
```
Should become :
```TypeScript
poster: {
  '0': { poster0 },
  '1': { poster1 },
  '2': { poster2 },
  // ...
}
```
### ImgRef migration
**Old format**
```TypeScript
avatar { // depending on the document, this can also be org.logo or movie.promotionalElements.banner.media, etc... 
  ref: 'users/ID/avatar/', // this become orginal.ref
  urls: {
    original: 'https://bablallba/orgininal-access-url',  // this become original.url
    xs: 'https://bablallba/xs-access-url',   // this become xs.url (xs.ref stay empty)
    md: 'https://bablallba/md-access-url',   // this become md.url (md.ref stay empty)
    // ... other sizes
  }
}
[...]

PR #3281, opened 2020-07-14
Remove trailer

RemcoSimonides:
issue #3096
- [x] Remove trailer from code
- [x] Remove trailer in migration

### DIFF
--- a/apps/festival/festival/src/app/marketplace/home/home.component.html
+++ b/apps/festival/festival/src/app/marketplace/home/home.component.html
@@ -51,23 +51,6 @@
               <mat-icon svgIcon="arrow_forward"></mat-icon>
             </a>
           </div>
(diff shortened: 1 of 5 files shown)


apps/festival/festival/src/app/marketplace/home/home.component.html
```diff
@@ -51,23 +51,6 @@ <h1>Featured films</h1>
...
-                      class="detail">
-                      More Details</a>
-                  </ng-container>
-                </movie-slide-cta>
-                <movie-slide-actions>
-                  <wishlist-button [movieId]="movie.id"></wishlist-button>
-                </movie-slide-actions>
-              </movie-slide>
-            </bf-slide>
-          </bf-slider>
```
pldespaigne, line 70: That's not a good idea at all !
just remove the `<ng-container *ngIf ...` but not it's content

PR #3285, opened 2020-07-14
Unit Tests(notification): Fixes #3284

### LINKED ISSUE #3284: UnitTest : Notification Service
[o] Unit Test the notification service

[o] Identify best practice for angular service tests involving firestore
- Test structure
- Configs

[o] Expectations:
- Should meaningfully test the service
- 

nx test notification should be green
npm run test & ng test should also pass.

xmano:
Unit tests for notification service.


libs/notification/src/lib/notification/+state/notification.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { NotificationService } from './notification.service';
+import { NotificationStore } from './notification.store';
+import { AngularFirestore } from '@angular/fire/firestore';
+import { Notification } from './notification.model';
+import * as firebase from '@firebase/testing';
+
+const NotificationStoreStub = {
+  formatNotification : (notification) => ({
+    format: JSON.stringify(notification)
```
GrandSchtroumpf, line 11: You don't need to stringify I think, just : `format: notification`. Like that you don't need to parse later.

libs/notification/src/lib/notification/+state/notification.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  it('should create notification service', () => {
+    expect(notifService).toBeTruthy();
+  });
+
+  it('reads notification', () => {
+    notifService.update = jest.fn();
+    notifService.readNotification(notif);
+    expect(notifService.update).toHaveBeenCalled();
```
GrandSchtroumpf, line 54: You're not testing against the emulator here.
You need to do someting like : 
```typescript
const doc = await firestore().doc('notifications/'+id).get();
expect(doc.data().read).toBeTruthy);
```

libs/notification/tsconfig.spec.json
```diff
@@ -1,9 +1,10 @@
-{
+﻿{
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "outDir": "../../dist/out-tsc",
     "module": "commonjs",
     "types": ["jest", "node"]
   },
+  "files": ["src/test-setup.ts"],
```
GrandSchtroumpf, line 8: I'm suprised you need that...

xmano:
Addressing the issues with a different PR #3334

Issue #3207, opened 2020-06-26
[Festival] Improve public org list loading


Jackseed:
## Description
This page is super long to load: `/c/o/marketplace/organization`
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/eb638fb2-219f-47af-9f6a-625d12270208)

## Acceptance criteria
- [ ] Look at the queries and find a way to speed up loading time

PR #3299, opened 2020-07-15
Fix and improve featured sales agent banner component

RemcoSimonides:
- [x] Reinstert html code lost during media migration
- [x] Apply organization with movie query in [issue #3207](https://github.com/blockframes/blockframes/pull/3297) 
- [x] Remove unused code

### DIFF
--- a/apps/festival/festival/src/app/marketplace/home/home.component.html
+++ b/apps/festival/festival/src/app/marketplace/home/home.component.html
@@ -66,6 +66,12 @@
               </movie-slide>
             </bf-slide>
           </bf-slider>
+          <!-- Featured Sales Agent Banner -->
+          <ng-container *ngIf="featuredOrg$ | async as org">
+            <ng-container *ngIf="org | orgMovies:0:3 | async as movies">
+              <org-banner [org]="org" [movies]="movies"></org-banner>
+            </ng-container>
+          </ng-container>
         </ng-container>
       </article>
     </ng-container>
--- a/apps/festival/festival/src/app/marketplace/home/home.component.html
+++ b/apps/festival/festival/src/app/marketplace/home/home.component.html
@@ -68,9 +68,7 @@
           </bf-slider>
           <!-- Featured Sales Agent Banner -->
           <ng-container *ngIf="featuredOrg$ | async as org">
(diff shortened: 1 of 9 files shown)


libs/organization/src/lib/organization/+state/organization.model.ts
```diff
@@ -20,9 +20,7 @@ export {
 
 export type AppStatus = 'none' | 'requested' | 'accepted';
 
-export interface Organization extends OrganizationDocumentWithDates {
-  movies?: Movie[];
-}
+export interface Organization extends OrganizationDocumentWithDates {}
```
RemcoSimonides, line 23: I could remove this line but then we don't have a simple Organization interface

apps/festival/festival/src/app/marketplace/home/home.component.html
```diff
@@ -66,6 +66,12 @@ <h1>Featured films</h1>
               </movie-slide>
             </bf-slide>
           </bf-slider>
+          <!-- Featured Sales Agent Banner -->
+          <ng-container *ngIf="featuredOrg$ | async as org">
+            <ng-container *ngIf="org | orgMovies:0:3 | async as movies">
+              <org-banner [org]="org" [movies]="movies"></org-banner>
```
GrandSchtroumpf, line 72: you can do even better by removing the [movies] input and do the pipe "orgMovie" directly in the component.

apps/festival/festival/src/app/marketplace/home/home.component.ts
```diff
@@ -69,7 +69,7 @@ export class HomeComponent implements OnInit, OnDestroy {
     ];
 
     this.orgs$ = this.organizationService
-      .queryWithMovies(ref => ref
+      .valueChanges(ref => ref
```
RemcoSimonides, line 72: I guess Akita works like this? I am not familiar with it yet

libs/organization/src/lib/organization/+state/organization.model.ts
GrandSchtroumpf: keep it like that it's fine.

RemcoSimonides:
Moved changes to already existing implement featured sales agent banner branch (https://github.com/blockframes/blockframes/pull/3303)

PR #3303, opened 2020-07-16
Implement featured sales agent component in homepage

RemcoSimonides:
- [x] Reinstert html code lost during media migration
- [x] Apply organization with movie query in [issue #3207](https://github.com/blockframes/blockframes/pull/3297) 
- [x] Remove unused code

### DIFF
--- a/libs/organization/src/lib/organization/components/banner/banner.component.html
+++ b/libs/organization/src/lib/organization/components/banner/banner.component.html
@@ -1,7 +1,7 @@
 <header fxLayout fxLayoutAlign="space-between start">
   <h1>{{ org | orgName }}</h1>
   <a mat-button [routerLink]="['/c/o/marketplace/organization/', org.id]" color="primary">
-    <span>See all ({{ movies.length }})</span>
+    <span>See all ({{ org.movieIds.length }})</span>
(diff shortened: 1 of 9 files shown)


libs/organization/src/lib/organization/components/banner/banner.component.html
```diff
@@ -1,7 +1,7 @@
 <header fxLayout fxLayoutAlign="space-between start">
   <h1>{{ org | orgName }}</h1>
   <a mat-button [routerLink]="['/c/o/marketplace/organization/', org.id]" color="primary">
-    <span>See all ({{ movies.length }})</span>
+    <span>See all ({{ org.movieIds.length }})</span>
```
RemcoSimonides, line 4: This may not be the way to go since it includes movies that have not been accepted. Is there perhaps another way to get this number?

GrandSchtroumpf: good point. In a next PR we could do something like : 
```html
<ng-container *ngIf="org | orgMovies | onlyAccepted | async as movies">
  <header>...</header>
  <section>...</section>
</ng-container>
```

Issue #3267, opened 2020-07-09
Improve communication between UploadWidgetComponent, MediaService and Storage


RemcoSimonides:
The [fix for the circular dependency](https://github.com/blockframes/blockframes/pull/3148) introduced a few new bugs.

- [x] If the state of the uploadTask is "paused" then the cancel button in the UploadWidgetComponent doesn't work
- [x] Canceling an upload in the UploadWidgetComponent while it's state is "running" shows an error but at least the cancel button does work (catch error).
- [x] If the UploadWidgetComponent is already open and a new file needs to be uploaded, this task isn't added to the component.

link to main issue #3084

PR #3305, opened 2020-07-16
Fix upload cancel and updatable items in UploadWidgetComponent

RemcoSimonides:
issue #3267

- [x] If the state of the uploadTask is "paused" then the cancel button in the UploadWidgetComponent doesn't work
- [x] Canceling an upload in the UploadWidgetComponent while it's state is "running" shows an error but at least the cancel button does work (catch error).
- [x] If the UploadWidgetComponent is already open and a new file needs to be uploaded, this task isn't added to the component.

### DIFF
--- a/libs/media/src/lib/+state/media.service.ts
+++ b/libs/media/src/lib/+state/media.service.ts
@@ -51,7 +51,9 @@ export class MediaService {
     const files = Array.isArray(uploadFiles) ? uploadFiles : [uploadFiles];
     const tasks = files.map(file => this.storage.upload(`${file.path}${file.fileName}`, file.data));
     this.addTasks(tasks);
-    (Promise as any).allSettled(tasks as AngularFireUploadTask[]).then(() => delay(5000).then(() => this.detachWidget()));
+    (Promise as any).allSettled(tasks)
+      .then(() => delay(5000))
+      .then(() => this.detachWidget());
     this.showWidget();
   }
 
@@ -72,7 +74,9 @@ export class MediaService {
     }
 
     this.addTasks(tasks);
-    (Promise as any).allSettled(tasks as AngularFireUploadTask[]).then(() => delay(5000).then(() => this.detachWidget()));
+    (Promise as any).allSettled(tasks)
+      .then(() => delay(5000))
+      .then(() => this.detachWidget());
     this.showWidget();
   }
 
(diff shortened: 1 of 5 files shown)


libs/media/src/lib/+state/media.service.ts
```diff
@@ -87,16 +91,36 @@ export class MediaService {
...
+    const tasks = this._tasks.getValue();
+    for (const task of tasks) {
+      const state = task.task.snapshot.state
+      if (state !== 'success') {
+        canClose = false;
+        break;
+      }
+    }
+
+    if (canClose) {
```
GrandSchtroumpf, line 113: Because detachWidget is only triggered when Promise.allSettle is done, then you might have issues if you add more task in between don't you ?

libs/media/src/lib/+state/media.service.ts
```diff
@@ -87,16 +91,36 @@ export class MediaService {
...
+    }
   }
 
-  private async showWidget(tasks?: AngularFireUploadTask[]) {
+  private async showWidget() {
     if (!this.overlayRef) {
       this.overlayRef = this.overlay.create(this.overlayOptions);
       const instance = new ComponentPortal(UploadWidgetComponent);
-      instance.injector = Injector.create({ providers: [{ provide: 'tasks', useValue: tasks }]});
+      instance.injector = Injector.create({ providers: [{ provide: 'tasks', useValue: this._tasks }]});
```
GrandSchtroumpf, line 123: give the `this._tasks.asObservable` here instead of doing it in the component

RemcoSimonides: I call tasks.next() in the UploadWidgetComponent when a task is removed. This won't be possible anymore if I pass the observable only, right?

libs/media/src/lib/+state/media.service.ts
```diff
@@ -87,16 +91,36 @@ export class MediaService {
...
+  private addTasks(tasks: AngularFireUploadTask[]) {
+    const t = this._tasks.getValue();
+    t.push(...tasks);
+    this._tasks.next(t);
+  }
+
+  private detachWidget() {
+    if (!this.overlayRef) return;
+
+    let canClose: boolean = true;
```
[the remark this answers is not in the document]
RemcoSimonides: Nice! I didn't know about `every`

libs/media/src/lib/+state/media.service.ts
RemcoSimonides: No, because there will be a new Promise.allSettle active. the detachWidget is called for all the Promise.allSettled, but checks whether it can actually close looking at all the tasks.

I did however find another issue when trying this; I don't clean up finished uploads so they appear again when the widget is shown

libs/media/src/lib/+state/media.service.ts
GrandSchtroumpf: task shouldn't be removed. We use cancel now ;)

RemcoSimonides: Yet there was always the ability to remove a task once the upload has succeeded or failed :/

Issue #3283, opened 2020-07-14
Transition to ImgIx


pldespaigne:
Transition from custom image system to ImgIx.

- [x] backend
- [x] form
- [x] frontend

link to main issue #3084

PR #3311, opened 2020-07-16
Transition to ImgIx 

pldespaigne:
solve #3283


libs/media/src/lib/+state/media.firestore.ts
```diff
@@ -38,47 +38,27 @@ export function createHostedMedia(media?: Partial<HostedMedia>): HostedMedia {
   }
 }
 
-
 /**
  * This array contains the different image sizes.
- * @note it **does not** contains the `original` key
  */
-export const imgSizeDirectory = ['lg', 'md', 'xs', 'fallback'] as const;
+export const imgSizeDirectory = ['lg', 'md', 'xs'] as const;
```
GrandSchtroumpf, line 44: If it's only used in the frontend can you move it into the directive direcly ?

libs/media/src/lib/+state/media.firestore.ts
```diff
@@ -38,47 +38,27 @@ export function createHostedMedia(media?: Partial<HostedMedia>): HostedMedia {
...
-    return { xs: 50, md: 100, lg: 300, fallback: 1024 };
-  } else if (url.includes('logo')) {
-    return { xs: 50, md: 100, lg: 300, fallback: 1024 };
-  } else if (url.includes('poster')) {
-    return { xs: 200, md: 400, lg: 600, fallback: 1024 };
-  } else if (url.includes('banner')) {
-    return { xs: 300, md: 600, lg: 1200, fallback: 1024 };
-  } else if (url.includes('still')) {
-    return { xs: 50, md: 100, lg: 200, fallback: 1024 };
+export function getImgSize(ref: string) {
```
GrandSchtroumpf, line 49: Same as above.

apps/catalog/catalog/src/app/dashboard/movie-tunnel/chain-of-titles/chain-of-titles.component.html
```diff
@@ -8,17 +8,13 @@ <h1 subtitle>Chain of Titles</h1>
...
       </mat-card-header>
       <mat-card-content>
-        <ng-container *ngFor="let control of chainOfTitles.controls; let i = index; let last = last">
+        <ng-container *ngFor="let control of chainOfTitles; let i = index; let last = last">
           <file-upload
             test-id="chain-titles"
-            path="pdf"
-            uploadOnFirestore="true"
-            return="imgRef"
+            storagePath=""
```
GrandSchtroumpf, line 14: Could we make it optional with a default value in the component if we only give empty string ?

libs/media/src/lib/directives/image-reference/background-reference.directive.ts
```diff
@@ -19,12 +19,13 @@ export class BackgroundReferenceDirective implements OnInit, OnDestroy {
...
   /** Set background-image attribute in any html tag with the url stored in firestore.
    *  If path is wrong, src will be set with provided placeholder or empty string */
-  @Input() set bgRef(image: ImgRef) {
+  @Input() set bgRef(image: HostedMedia) {
     if(!image){
       this.ref$.next('');
     } try {
-      const url = getImageUrl(image);
-      this.ref$.next(url);
+      // TODO IMPLEMENT IMGIX issue#3283
```
GrandSchtroumpf, line 26: Shoudn't it be done in this PR ?

libs/media/src/lib/directives/image-reference/image-reference.directive.ts
```diff
@@ -62,16 +78,44 @@ export class ImageReferenceDirective implements OnInit, OnDestroy {
...
+
+      this.cdr.markForCheck()
+    });
   }
 
   ngOnDestroy() {
     this.sub.unsubscribe();
   }
+
+  @HostListener('error')
```
GrandSchtroumpf, line 110: Put the hostlistener just after the constructor

libs/media/src/lib/directives/image-reference/imgix-helpers.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @example
+ * const param: ImageParameters = {
+ *   fit: 'crop',
+ *   width: 100,
+ *   height: 100
+ * };
+ * formatParameters(param); // '?fit=crop&w=100&h=100&'
+ */
+export function formatParameters(parameters: ImageParameters): string {
+  let query = '';
```
GrandSchtroumpf, line 26: Personaly I would have done : 
```typescript
const query = Object.entries(paramters)
  .filter(([key, value]) => !!value)
  .map(([key, value]) => `${key}=${value}`)
  .join('&')
```
But then "width" becomes "w" & same for height.
But I think it's nice to have the same keys than imgix in the end.

libs/utils/src/lib/breakpoint/breakpoints.service.ts
```diff
@@ -1,35 +1,35 @@
 import { Injectable } from "@angular/core";
 import { BreakpointObserver, Breakpoints } from "@angular/cdk/layout";
 import { shareReplay, distinctUntilChanged, map } from "rxjs/operators";
-import { Observable } from "rxjs";
+import { Observable, combineLatest } from "rxjs";
```
GrandSchtroumpf, line 4: Looks like you don't need combineLatest anymore

libs/media/src/lib/+state/media.model.ts
```diff
@@ -59,6 +59,18 @@ function mediaNeedsUpdate(media: HostedMediaFormValue) {
   return media.delete || (!!media.ref && !!media.blobOrFile);
 }
 
+export function getFileNameFromPath(path: string) {
+  return path.split('/').pop()
+}
+
+
+// TODO issue#3283 DO WE  REALLY NEED ALL THE CODE BELLOW ???
```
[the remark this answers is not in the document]
pldespaigne: I really like this kind of comment ! ^^

apps/catalog/catalog/src/app/dashboard/movie-tunnel/chain-of-titles/chain-of-titles.component.html
pldespaigne: No, we need to know where we will upload the file when the user hit the save button, here it's a weird behavior.
We should either fix it or maybe it's the sign that we need another component.

libs/media/src/lib/directives/image-reference/background-reference.directive.ts
pldespaigne: indeed, but you are too fast ⚡

PR #3315, opened 2020-07-16
add notification unit testing

GrandSchtroumpf:
Setup an Unit test example that connect to the emulator.

### DIFF
--- a/libs/notification/src/lib/notification/+state/notification.service.spec.ts
+++ b/libs/notification/src/lib/notification/+state/notification.service.spec.ts
@@ -17,7 +17,7 @@ describe('Notifications Test Suite', () => {
   let service: NotificationService;
   let db: AngularFirestore
 
-  beforeAll(async () => {
+  beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


libs/notification/src/lib/notification/+state/notification.service.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+  formatNotification(notification) {
+    return { format: JSON.stringify(notification) };
+  }
+};
+
+describe('Notifications Test Suite', () => {
+  let service: NotificationService;
+  let db: AngularFirestore
+
+  beforeAll(async () => {
```
xmano, line 20: This should be beforeEach() We need to setup / tear down for each test

GrandSchtroumpf: I'm not sure to see the point here. We clear the database everytime and every method is pure, so you cannot have any size effect between tests.

PR #3322, opened 2020-07-17
Catch error thrown on upload cancel

RemcoSimonides:
issue #3267

- [x] If the state of the uploadTask is "paused" then the cancel button in the UploadWidgetComponent doesn't work
- [x] Canceling an upload in the UploadWidgetComponent while it's state is "running" shows an error but at least the cancel button does work (catch error).


RemcoSimonides:
Continueing with https://github.com/blockframes/blockframes/pull/3305

PR #3334, opened 2020-07-17
Test(notif): Add more unit tests for notification service.

xmano:

### DIFF
--- a/libs/notification/src/lib/notification/+state/notification.service.spec.ts
+++ b/libs/notification/src/lib/notification/+state/notification.service.spec.ts
@@ -6,19 +6,10 @@ import { AngularFireModule } from '@angular/fire';
 import { SETTINGS, AngularFirestoreModule, AngularFirestore } from '@angular/fire/firestore';
 import { loadFirestoreRules, clearFirestoreData } from '@firebase/testing';
 import { readFileSync } from 'fs';
-import { Notification } from './notification.model';
 
 describe('Notifications Test Suite', () => {
   let service: NotificationService;
   let db: AngularFirestore;
-  const notif: Notification = {
-    type: 'newContract',
-    toUserId: "",
-    id: "",
-    isRead: false,
-    message: "Jest Notify",
-    date: new Date()
-  };
 
   beforeEach(async () => {
     TestBed.configureTestingModule({
@@ -28,7 +19,7 @@ describe('Notifications Test Suite', () => {
       ],
       providers: [
         NotificationService,
-        { provide: NotificationStore },
+        NotificationStore,
         { provide: SETTINGS, useValue: { host: 'localhost:8080', ssl: false } }
       ],
     });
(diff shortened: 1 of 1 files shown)


libs/notification/src/lib/notification/+state/notification.service.spec.ts
```diff
@@ -40,10 +49,19 @@ describe('Notifications Test Suite', () => {
...
   it('Should mark notifications as read', async () => {
     await db.doc('notifications/1').set({ isRead: false });
     await service.readNotification({ id: '1' });
     const doc = await db.doc('notifications/1').ref.get();
     expect(doc.data().isRead).toBeTruthy();
   });
+
+  it('formats notification', () => {
+    const newNotif = service.formatFromFirestore(notif);
+    expect(newNotif).toEqual({ ...notif, format: notif});
```
GrandSchtroumpf, line 65: Your code is exacly like creating a spy on the store. I think the best would be to actually do that as it would be easier & clearer.

xmano: Agree and implementing this with a spy.

libs/notification/src/lib/notification/+state/notification.service.spec.ts
```diff
@@ -40,10 +43,20 @@ describe('Notifications Test Suite', () => {
...
+
   it('Should mark notifications as read', async () => {
     await db.doc('notifications/1').set({ isRead: false });
     await service.readNotification({ id: '1' });
     const doc = await db.doc('notifications/1').ref.get();
     expect(doc.data().isRead).toBeTruthy();
   });
+
+  it('formats notification', () => {
+    service.formatFromFirestore = jest.fn();
```
GrandSchtroumpf, line 58: You're not testing anything here. You need to spy on the store, not the service.

libs/notification/src/lib/notification/+state/notification.service.spec.ts
```diff
@@ -40,10 +43,20 @@ describe('Notifications Test Suite', () => {
...
+    service.formatFromFirestore(notif);
```
GrandSchtroumpf, line 59: to be honest you don't even need a real notif here. An empty object you be enough for the sake of the test : 
```typescript
service.formatFromFirestore({} as any); // You can put "as any" if needed.
```

libs/notification/src/lib/notification/+state/notification.service.spec.ts
xmano: My mistake. Have corrected this.

libs/notification/src/lib/notification/+state/notification.service.spec.ts
xmano: Agreed

libs/notification/src/lib/notification/+state/notification.service.spec.ts
```diff
@@ -25,7 +19,7 @@ describe('Notifications Test Suite', () => {
       ],
       providers: [
         NotificationService,
-        { provide: NotificationStore, useClass: NotificationStoreStub },
+        NotificationStore,
```
GrandSchtroumpf, line 22: In the future, if you need to spy multiple methods you can also do:
```typescript
{ provide: NotificationStore, useValue createSpyObj('NotificationStore', ['formatNotification']) }
```

xmano:
Closes #3284

PR #3341, opened 2020-07-20
Issue 3306 form list

fritzschoff:
ref #3306


libs/ui/src/lib/form/list/form-list.component.html
```diff
@@ -0,0 +1,11 @@
+<ng-container *ngIf="tableView else form">
+  <ng-content select="bf-form-list-table"></ng-content>
+</ng-container>
+<ng-template #form>
+  <ng-template [ngTemplateOutlet]="formView" [ngTemplateOutletContext]="{ $implicit: localForm }">
+  </ng-template>
+  <button mat-stroked-button color="primary" (click)="saveForm()">
+    <mat-icon svgIcon="floppy"></mat-icon>
+    {{ buttonText }}
```
GrandSchtroumpf, line 9: Put a span here.

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -1,4 +1,4 @@
-<ng-container *ngIf="(hasMovies$ | async) else noMovie">
+<!-- <ng-container *ngIf="(hasMovies$ | async) else noMovie">
```
GrandSchtroumpf, line 1: Remove if you're not using.

libs/ui/src/lib/form/table/form-table.component.scss
```diff
@@ -0,0 +1,41 @@
...
+
+table {
+  width: 100%;
+}
+
+mat-paginator {
+  width: 100%;
+  bottom: 0px;
+  position: sticky;
+  z-index: 10;
```
GrandSchtroumpf, line 18: put "1" or "2" maximum if required

libs/ui/src/lib/form/table/form-table.component.scss
```diff
@@ -0,0 +1,41 @@
...
+}
+
+table {
+  width: 100%;
+}
+
+mat-paginator {
+  width: 100%;
+  bottom: 0px;
+  position: sticky;
```
GrandSchtroumpf, line 17: I'm curious why you need that. There is no relative position above. Could you try without position ?

libs/ui/src/lib/form/table/form-table.component.scss
```diff
@@ -0,0 +1,41 @@
...
+  display: flex;
+}
+
+::-webkit-scrollbar {
+  height: 10px;
+}
+
+::-webkit-scrollbar-track {
+  height: 10px;
+  background: linear-gradient(to bottom, transparent 30%, #cfd8ff 30%, #cfd8ff 70%, transparent 70%);
```
GrandSchtroumpf, line 35: I think Stephanie said in the case is to have a flat track.
I think, you should be able to do : 
```scss
background-color: var(--primary);
opacity: 0.5:
```

libs/ui/src/lib/form/table/form-table.component.scss
```diff
@@ -0,0 +1,41 @@
...
+}
+
+::-webkit-scrollbar-thumb {
+  background: #1a49f5;
```
GrandSchtroumpf, line 39: You can use `background-color: var(--primary);`

libs/ui/src/lib/form/table/form-table.component.ts
```diff
@@ -0,0 +1,117 @@
...
+    this.showTable$ = values$.pipe(
+      map(value => !!value.length),
+      tap(value => value ? this.formItem = this.form.at(0) : this.add()),
+      distinctUntilChanged()
+    );
+    // Show Paginator if table size goes beyond page size
+    this.showPaginator$ = values$.pipe(
+      map(value => {
+        /* We need to reconnect the paginator in order to force a rerendering of the pagination */
+        if (value.length >= this.pageSize) {
```
GrandSchtroumpf, line 72: I would have done it in a `tap` :
```typescript
this.showPaginator$ = values$.pipe(
  map(value => value.length >= this.pageSize)
  distinctUntilChanged(),
  tap(isVisible => isVisible ? this.dataSource.paginator = this.paginator : null)
)
```
But I'm curious how it actually work as the paginator shouldn't exist at that point.

libs/ui/src/lib/form/table/form-table.component.ts
```diff
@@ -0,0 +1,117 @@
...
+  add() {
+    this.formItem = this.form.createControl({});
+    this.showSave = true;
+  }
+
+  // Edit existing form, we don't want save button as content is updated in real time
+  edit(index: number) {
+    this.formItem = this.form.at(index);
+    this.showSave = false;
+    this.cdr.detectChanges();
```
GrandSchtroumpf, line 104: markForCheck doesn't work ?

libs/ui/src/lib/form/table/form-table.component.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  // Remove one line in the form
+  remove(index: number) {
+    this.form.removeAt(index);
+  }
+
+  // Push the form into the list
+  save() {
+    this.form.push(this.formItem);
+    this.showSave = false;
```
GrandSchtroumpf, line 115: You might need a `this.cdr.detechChanges()` here no ?

Issue #3329, opened 2020-07-17
e2e watermark restore


Hugo-Hontans:
When we restore database with prepareForTesting, there are lot of users without watermark, I think actually watermark are created by onUpdateUser. We should probably create a small ops script that generate all the watermarks and include it at the end of prepareForTesting.

pldespaigne:
migration 24 is responsible for generating all the watermarks.
- [ ] copy paste & refactor migration 24 into it's own ops script

Issue #3300, opened 2020-07-15
Greenlight existing e2e tests


Jackseed:
- [x] Movie form
- [ ] wishlist
- [x] watermark issue (when we restore database, some users don't have watermark, maybe link to onUpdateUser backend function) https://github.com/blockframes/blockframes/issues/3329
- [ ] screening-invitations

Hugo-Hontans:
I created a specific issue for watermark #3329

PR #3342, opened 2020-07-20
Fix watermark in migration

RemcoSimonides:
Issue #3329 && #3300 
Add function to add fix watermarks

### DIFF
--- a/apps/backend-functions/src/internals/watermark.ts
+++ b/apps/backend-functions/src/internals/watermark.ts
@@ -1,14 +1,13 @@
 import { getWatermark } from '../templates/watermark';
 import { PublicUser } from '../data/types';
 import { admin, getStorageBucketName } from './firebase';
-import { HostedMedia } from '@blockframes/media/+state/media.firestore';
 
 /**
  * - Generate a svg file with the name & email of the user
  * - Store the watermark file in the storage bucket
  * - User's firestore doc is updated by onFileUpload backend function
  */
-export async function upsertWatermark(user: PublicUser): Promise<HostedMedia> {
+export async function upsertWatermark(user: PublicUser): Promise<void> {
 
   if (!user.email) {
     throw new Error(`Cannot generate watermark for user ${user.uid} because 'email' is not provided.`);
@@ -23,12 +22,4 @@ export async function upsertWatermark(user: PublicUser): Promise<HostedMedia> {
     file.createWriteStream({ contentType: 'image/svg+xml' }).end(watermark, () => res());
   });
 
-  const [signedUrl] = await file.getSignedUrl({ action: 'read', expires: '01-01-3000', version: 'v2' });
-
-  const watermarkMedia = {
-    ref,
-    url: signedUrl,
-  }
-
-  return watermarkMedia;
 }
--- a/apps/backend-functions/src/internals/watermark.ts
+++ b/apps/backend-functions/src/internals/watermark.ts
@@ -1,20 +1,14 @@
 import { getWatermark } from '../templates/watermark';
 import { PublicUser } from '../data/types';
-import { admin, db, getStorageBucketName } from './firebase';
+import { admin, getStorageBucketName } from './firebase';
 import { HostedMedia } from '@blockframes/media/+state/media.firestore';
 
 /**
  * - Generate a svg file with the name & email of the user
  * - Store the watermark file in the storage bucket
- * - Update the user's firestore doc to link it to the watermark file
+ * - User's firestore doc is updated by onFileUpload backend function
  */
-export async function upsertWatermark(user: PublicUser) {
-  const watermark = await _upsertWatermark(user);
-  const userRef = db.collection('users').doc(user.uid);
-  return userRef.update({ watermark });
-}
-
-export async function _upsertWatermark(user: PublicUser): Promise<HostedMedia> {
+export async function upsertWatermark(user: PublicUser): Promise<HostedMedia> {
 
   if (!user.email) {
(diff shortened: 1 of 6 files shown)


apps/backend-ops/src/firestoreMigrations/0028.ts
```diff
@@ -178,9 +176,7 @@ const updateUserAvatarAndWaterMark = async (user: OldPublicUser, storage: Storag
   try {
     const newImageRef = await updateImgRef(user, 'avatar', storage);
     user.avatar = newImageRef;
-    const watermark = await _upsertWatermark(user as any); // _upsertWatermark only require uid, email, firstName, lastName, witch are common between the 2 types
-    user.watermark.ref = watermark.ref;
-    user.watermark.urls.original = watermark.url;
+    await upsertWatermark(user as any); // _upsertWatermark only require uid, email, firstName, lastName, witch are common between the 2 types
```
pldespaigne, line 179: please edit the comment as `_upsertWatermark` doesn't exists aymore.
Something like `// upsertWatermark only need firstName, lastName, email witch are also present in OldPublicUser`

apps/backend-ops/src/users.ts
```diff
@@ -145,3 +149,24 @@ export async function createUsers(): Promise<any> {
...
+  try {
+    const users = await getCollection<PublicUser>('users');
+    const { storage } = loadAdminServices();
+    const bucket = storage.bucket(getStorageBucketName());
+
+    const promises = users.map(async user => {
+      const ref = `users/${user.uid}/watermark/${user.uid}.svg`;
+      const to = bucket.file(ref);
+      const exists = await to.exists();
+      if (!exists) {
```
pldespaigne, line 163: `to.exists()` returns a `[boolean]`, so when you do `if (!exists)` you check if the array is undefined or null witch never happen. So you code will never go inside your `if`.

You should refactor into `const [exists] = await to.exists();` intead

apps/backend-functions/src/internals/watermark.ts
```diff
@@ -1,20 +1,14 @@
...
+ * - User's firestore doc is updated by onFileUpload backend function
  */
-export async function upsertWatermark(user: PublicUser) {
-  const watermark = await _upsertWatermark(user);
-  const userRef = db.collection('users').doc(user.uid);
-  return userRef.update({ watermark });
-}
-
-export async function _upsertWatermark(user: PublicUser): Promise<HostedMedia> {
+export async function upsertWatermark(user: PublicUser): Promise<HostedMedia> {
```
pldespaigne, line 11: `upsarteWatermark` only work on the storage, so it should not return anything.
also remove line 26-33

pldespaigne: also refactor the function to wait for `ref` field to be empty in the db.
- check value of ref :
  - ref is empty do nothing go to next step
  - ref is not empty: check if file exists
    - if it exists, delete it and wait for ref to become empty
    - if it doesn't exist (that's a mistake) just empty the ref in the db
- create the new watermark in the storage, do not touch the ref

apps/backend-ops/src/users.ts
```diff
@@ -145,3 +151,21 @@ export async function createUsers(): Promise<any> {
   const usersWithPassword = users.map(user => ({ ...user, password: 'password' }));
   return createAllUsers(usersWithPassword, auth);
 }
+
+export async function generateWatermarks() {
+
+  // activate migration to prevent cloud functions to trigger
```
pldespaigne, line 157: `// activate maintenance`

apps/backend-ops/src/users.ts
```diff
@@ -145,3 +151,21 @@ export async function createUsers(): Promise<any> {
...
+  await startMaintenance();
+
+  const users = await getCollection<PublicUser>('users');
+  const promises = []
+  for (const user of users) {
+    promises.push(db.doc(`users/${user.uid}`).update({ watermark: createHostedMedia() }));
```
pldespaigne, line 163: this is not needed, as it will be erased/overwritten by sync

apps/backend-ops/src/users.ts
```diff
@@ -145,3 +151,21 @@ export async function createUsers(): Promise<any> {
...
+    promises.push(upsertWatermark(user));
+  }
+  await Promise.all(promises);
+
+  // deactivate migration
```
pldespaigne, line 168: maintenance*

apps/backend-functions/src/main.ts
```diff
@@ -272,10 +272,10 @@ export const relayerSend = functions.https
 //--------------------------------
 
 /** Trigger: on every file uploaded to the storage. */
-export const onFileUpload = functions.storage.object().onFinalize(data => linkFile(data))
+export const onFileUpload = functions.storage.object().onFinalize(data => skipInMaintenance(linkFile(data)))
```
pldespaigne, line 275: `skipInMaintenance` expect a function as parameter, but here you gave it the result of `linkFile` wich is a promise. You should refactor into `skipInMaintenance(() => linkFile(data))` instead

apps/backend-functions/src/main.ts
```diff
@@ -272,10 +272,10 @@ export const relayerSend = functions.https
...
 
 //--------------------------------
 //         File delete          //
 //--------------------------------
 
-export const onFileDelete = functions.storage.object().onDelete(data => unlinkFile(data))
+export const onFileDelete = functions.storage.object().onDelete(data => skipInMaintenance(unlinkFile(data)))
```
pldespaigne, line 281: same as above

Issue #3292, opened 2020-07-15
Catalog dashboard layout


Hugo-Hontans:
- [ ] Implement a DashboardComponent at the same level as the dashboard module to replace the layout component (catalog/dashboard/layout). Use the layout-dashboard component in lib/layout
- [ ] Update the e2e test which use the tag `catalog-layout`

xmano:
Fixed by PR #3353

PR #3353, opened 2020-07-22
Refactor(catalog-dashboard): Use a UI dashboard lib component

xmano:
Refactoring Catalog app dashboard #3292


apps/catalog/catalog/src/app/dashboard/dashboard.component.html
```diff
@@ -0,0 +1,23 @@
+﻿<layout-dashboard hideSearchbar>
+  <aside fxLayout="column" fxLayoutAlign="space-between">
+    <mat-nav-list fxFlex>
+      <a mat-list-item routerLink="home" routerLinkActive="active">
+        <mat-icon matListIcon svgIcon="dashboard"></mat-icon>
+        <span matLine><b>Dashboard</b></span>
+      </a>
+      <a mat-list-item routerLink="titles" routerLinkActive="active">
```
Hugo-Hontans, line 8: Normally on current develop the routerLink should be `title` and not `titles`.
Can you verify this route plz ?

xmano: Corrected.

apps/catalog/catalog/src/app/dashboard/dashboard.component.html
```diff
@@ -0,0 +1,23 @@
+﻿<layout-dashboard hideSearchbar>
```
GrandSchtroumpf, line 1: I looked into the code and it seems that there is no input hideSearchbar in the layout-dashboard. You can remove it here an din the festival app too.

xmano: I removed `hideSearchbar` from both catalog & festival.

xmano:
Closes #3292

PR #3358, opened 2020-07-22
Issue 3306 list

### LINKED ISSUE #3306: formList component
some stuff

fritzschoff:
close #3306

### DIFF
--- a/libs/ui/src/lib/form/list/form-list.component.html
+++ b/libs/ui/src/lib/form/list/form-list.component.html
@@ -6,16 +6,14 @@
   <ng-container *ngIf="list.length">
     <mat-list class="surface">
       <mat-list-item *ngFor="let item of list; let index = index">
-        <ng-template [ngTemplateOutlet]="itemRef || text" [ngTemplateOutletContext]="{ $implicit: item }">
+        <ng-template [ngTemplateOutlet]="itemRef || text" [ngTemplateOutletContext]="{ $implicit: item }" matLine>
         </ng-template>
-        <span>
-          <button mat-icon-button (click)="edit(index)">
-            <mat-icon svgIcon="pencil"></mat-icon>
-          </button>
-          <button mat-icon-button (click)="remove(index)">
-            <mat-icon svgIcon="trash"></mat-icon>
-          </button>
-        </span>
+        <button mat-icon-button (click)="edit(index)">
+          <mat-icon svgIcon="pencil"></mat-icon>
+        </button>
+        <button mat-icon-button (click)="remove(index)">
+          <mat-icon svgIcon="trash"></mat-icon>
+        </button>
       </mat-list-item>
     </mat-list>
   </ng-container>
@@ -35,6 +33,6 @@
 <ng-container *ngIf="showSave">
   <button mat-stroked-button color="primary" (click)="save()">
     <mat-icon svgIcon="floppy"></mat-icon>
-    Save
+    <span>Save</span>
   </button>
 </ng-container>
--- a/libs/ui/src/lib/form/list/form-list.component.html
+++ b/libs/ui/src/lib/form/list/form-list.component.html
@@ -1,22 +1,26 @@
 <!---------->
 <!-- LIST -->
 <!---------->
+
 <ng-container *ngIf="list$ | async as list">
   <ng-container *ngIf="list.length">
     <mat-list class="surface">
       <mat-list-item *ngFor="let item of list; let index = index">
-        <ng-template [ngTemplateOutlet]="template || text" [ngTemplateOutletContext]="{ $implicit: item }">
+        <ng-template [ngTemplateOutlet]="itemRef || text" [ngTemplateOutletContext]="{ $implicit: item }">
         </ng-template>
-        <button mat-icon-button (click)="edit(index)">
-          <mat-icon svgIcon="pencil"></mat-icon>
-        </button>
-        <button mat-icon-button (click)="remove(index)">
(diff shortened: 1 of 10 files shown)


libs/ui/src/lib/form/table/form-table.component.html
```diff
@@ -48,6 +48,6 @@
 <ng-container *ngIf="showSave">
   <button mat-stroked-button color="primary" (click)="save()">
     <mat-icon svgIcon="floppy"></mat-icon>
-    <span>Save</span>
```
fritzschoff, line 51: if you put a span, the layout goes nuts

libs/ui/src/lib/form/list/form-list.component.html
```diff
@@ -0,0 +1,40 @@
...
+<!-- LIST -->
+<!---------->
+
+<ng-container *ngIf="list$ | async as list">
+  <ng-container *ngIf="list.length">
+    <mat-list class="surface">
+      <mat-list-item *ngFor="let item of list; let index = index">
+        <ng-template [ngTemplateOutlet]="itemRef || text" [ngTemplateOutletContext]="{ $implicit: item }">
+        </ng-template>
+        <span>
```
GrandSchtroumpf, line 11: Do you really need this span ? I think it would be better to have the two button unwrapped. I recall some weird behavior on responsive when you wrap two buttons at the end of a list-item.

libs/ui/src/lib/form/list/form-list.component.html
```diff
@@ -0,0 +1,40 @@
...
+<!---------->
+<!-- VIEW -->
+<!---------->
+
+<ng-template [ngTemplateOutlet]="formView" [ngTemplateOutletContext]="{ $implicit: formItem }">
+</ng-template>
+<ng-container *ngIf="showSave">
+  <button mat-stroked-button color="primary" (click)="save()">
+    <mat-icon svgIcon="floppy"></mat-icon>
+    Save
```
GrandSchtroumpf, line 38: Use span here to have the 8px margin between text & icon

libs/ui/src/lib/form/list/form-list.component.scss
```diff
@@ -0,0 +1,4 @@
+span  {
```
GrandSchtroumpf, line 1: I'm pretty sure you can remove that

libs/ui/src/lib/form/list/form-list.component.ts
```diff
@@ -0,0 +1,76 @@
...
+  list$: Observable<any[]>;
+  showSave: boolean;
+  formItem: FormEntity<EntityControl<T>, T>;
+
+  constructor(private cdr: ChangeDetectorRef) { }
+
+  ngOnInit() {
+    this.list$ = this.form.valueChanges.pipe(
+      startWith(this.form.value),
+      tap(value => value.length ? this.formItem = this.form.at(0) : this.add()),
```
GrandSchtroumpf, line 48: I don't see the point of that. This will be triggered each time you update one value of a form.
I think it's better to just remove this line.

libs/ui/src/lib/form/list/form-list.component.ts
```diff
@@ -0,0 +1,76 @@
...
+      distinctUntilChanged())
+  }
+
+  // Add a clean form and show save button
+  add() {
+    this.formItem = this.form.createControl({});
```
GrandSchtroumpf, line 54: I think you need to do `this.form.createControl()` without params. Else it'll break with string or number.

libs/ui/src/lib/form/table/form-table.component.html
GrandSchtroumpf: That's weird. Are you styling the span in the css ?

libs/ui/src/lib/form/list/form-list.component.ts
fritzschoff: and then whats happening if there is no form from the beginning OR when the user delets all his forms?

libs/ui/src/lib/form/list/form-list.component.scss
fritzschoff: how do you want to attach the span to the right of the lsit?

libs/ui/src/lib/form/list/form-list.component.ts
GrandSchtroumpf: Like in the table, you don't display the list.

libs/ui/src/lib/form/list/form-list.component.scss
GrandSchtroumpf: When you put "matline" in the template it should push the content to the far right by default.

libs/ui/src/lib/form/list/form-list.component.ts
fritzschoff: yes. but we still need a localForm

fritzschoff: formItem will be undefined

GrandSchtroumpf: You create one onInit & on remove
```typescript
remove(index: number) {
  if (!this.form.length) {
    this.formItem = this.form.createControl()
  }
}
```

libs/ui/src/lib/form/list/form-list.component.ts
fritzschoff: parameter is required

Issue #3083, opened 2020-06-16
Movie migration


Mussuranae:
Here are the future changes for the movie data model and it would be good to do unit test for testing :

**Main Section :**
- [x] productionYear => DELETE
- [x] officialsIds => KEP ONLY IMDB
- [x] stakeholders => MOVE INTO OTHER SECTION
- [x] directors[].filmography => RENAME FROM shortBiography
- [x] directors[].status => CREATE
- [x] directors[].category => CREATE
- [x] workType => RENAME IN contentType
- [x] poster => **it has been done in the `media-backend` branch** (ask @pldespaigne if needed)
- [x] banner => **it has been done in the `media-backend` branch** (ask @pldespaigne if needed)

**Sales Info Section : (deletion of this section)**
- [x] broadcasterCoproducers => DELETE
- [ ] certifications => MOVE INTO ADDITIONAL SECTION
- [x] color => MOVE INTO TECHNICAL SECTION
- [x] format => MOVE INTO TECHNICAL SECTION
- [x] formatQuality => MOVE INTO TECHNICAL SECTION
- [ ] originalRelease => MOVE INTO ADDITIONAL SECTION
- [x] physicalHVRelease => DELETE
- [ ] rating => MOVE INTO ADDITIONAL SECTION
- [ ] scoring => ???
- [x] soundFormat => MOVE INTO TECHNICAL SECTION
- [x] releaseYear => MOVE INTO MAIN

![image](https://user-images.githubusercontent.com/48033264/84780978-8a8c2280-afe6-11ea-9c41-0efa50d8761c.png)

**Production Section : (new section)**
- [ ] producers => MOVE FROM SALE CAST
- [ ] productionCompany => MOVE FROM MAIN (STAKEHOLDERS)
- [ ] coProductionCompany => MOVE FROM MAIN (STAKEHOLDERS)
- [ ] distributors => MOVE FROM MAIN (STAKEHOLDERS)
- [ ] salesAgent => MOVE FROM MAIN (STAKEHOLDERS)

**Technical Section : (new one)**
See Sales Info section

**Additional Section : (new one)**
See Sales Info section and Budget Section

![image](https://user-images.githubusercontent.com/48033264/84781033-9aa40200-afe6-11ea-8058-5099727ff9f1.png)

**Budget Section :**
- [ ] boxOffice => MOVE INTO ADDITIONAL SECTION

**Sales Cast :**
- [x] Rename it to Artistic Team

**Sales Agent Deal Section**
[...]

PR #3361, opened 2020-07-23
Flattened movie

Mussuranae:
RIP Movie object, flattened almost all section.
Related to #3083


apps/backend-ops/src/firestoreMigrations/temp-31.ts
```diff
@@ -0,0 +1,89 @@
...
+    delete movie.salesAgentDeal;
+    delete movie.deliveryIds;
+
+    const newData = {
+      ...movie,
+      // The only section left
+      promotional: promotionalElements,
+
+      // All fields are now on the root of the movie
+      banner: main.banner || null,
```
GrandSchtroumpf, line 46: Prefer "??" instead of "||". because : `0 || null -> null` `0 ?? null -> 0`

GrandSchtroumpf: Also I think it's better to not set null at all & clean the data before the batch with 
```typescript
for (const key in newData) {
  if (newData[key] === undefined) {
    delete newData[key];
  }
}
```
Like that you don't put "null" on unset data.

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -71,59 +55,22 @@ export function createMovie(params: Partial<Movie> = {}): Movie {
...
+    contentType: 'feature_film',
     directors: [],
     genres: [],
-    contentType: 'feature_film',
     originalLanguages: [],
     originCountries: [],
-    status: null,
-    customGenres: [],
     releaseYear: null,
+    storeConfig: createStoreConfig(),
```
GrandSchtroumpf, line 64: I think you need to put it below the ...params with the params.storeConfig inside. Same for the title.

Issue #3162, opened 2020-06-23
Reproduce prod environment on staging


Jackseed:
## Description
Staging has diverged from prod and is no longer a trusted test environment. 
In order to test properly we need to have on staging an up-to-date firestore & firestorage taken from prod. We also need to anonymize prod firestore beforehand by changing:
- Org name (`orgs.denomination.full` & `orgs.denomination.public`)
- Org contact email address (`orgs.email`)
- User name (`user.firstName` & `user.lastName`)
- User email (`user.email`)

Anonymizing prod protects users' data privacy and avoid staging to spam emails to real users. 

In order to be easily readable for us (and thus better testing) we used these conventions when generating the fake data:
- org contact email address was derived from org name (contact@orgName.fake.cascade8.com)
- user email was derived from user name (userName@orgName.fake.cascade8.com)

These conventions should be updated to use real email address owned by cascade8 to avoid our Sendgrid email score to be polluted by the tests. 
E.g. use something like fakeAccounts+orgName-contact@cascade8.com & fakeAccounts+userName-orgName@cascade8.com

The process should be this:
1. export prod db & storage
2. anonymize prod db
3. import anonymized prod db & storage
4. erase and recreate firebase authentication users based on anonymize prod db

Best would be to be able to activate that script whenever we want and/or to activate it on each staging deploy. 

## Current tools
- Functions exist to backup & restore firestore following this url: projectHostingUrl/admin/data/backup & projectHostingUrl/admin/data/restore. [Here](https://www.notion.so/cascade8/Data-import-backup-c3279edb834f4a25b130adaecacef0d4) is an article to set it up on a project.
When restore function is used, authentication is reset to match firestore
- [Notion article](https://www.notion.so/cascade8/Bucket-storage-backup-restore-000b1dbb9df8453d807377e158a2450b) to use gs util to import & export firestorage; Google cloud also [exposes functions](https://console.cloud.google.com/transfer/cloud?project=blockframes&authuser=1) to do the same
- Anonymizing prod was done with a Python script 

## Acceptance criteria
- [ ] Update backup & restore db functions if needed ([ressource](https://dev.to/mikgross/firebase-firestore-migrations-olg) from François
- [ ] Add a storage backup & restore to the db functions
- [ ] Include a script to anonymize db
- [ ] Add it on the deploy staging flow and/or as a callable script

george43g:
"Staging has diverged from prod and is no longer a trusted test environment"

[...]

PR #3369, opened 2020-07-24
chore(prepareForTesting): progress save for anonymization script

george43g:
re #3162 

This PR is a draft, almost complete anonymisation process/workflow for 'prepareForTesting' script which successfully takes DB backup from storage, anonymises it and generates local fixtures for testing.

Things left:
- implement a Queue as large DB will crash it, but has been tested working on small DB.
- More thorough testing needed


.circleci/config.yml
```diff
@@ -261,7 +261,8 @@ jobs:
             npx firebase use staging &&
             cp firestore.rules firestore.rules.regular &&
             cp firestore.maintenance.rules firestore.rules &&
-            npx firebase deploy --only hosting:catalog,hosting:festival,storage,firestore &&
+            npx firebase deploy --only hosting:catalog,hosting:festival,storage &&
+            npx firebase deploy --only firestore --force &&
```
phpgeekfr, line 265: OK this should do the trick. You branch is a bit far from develop, you will need to add `hosting:crm` also

.circleci/config.yml
```diff
@@ -261,7 +261,8 @@ jobs:
             npx firebase use staging &&
             cp firestore.rules firestore.rules.regular &&
             cp firestore.maintenance.rules firestore.rules &&
-            npx firebase deploy --only hosting:catalog,hosting:festival,hosting:crm,storage,firestore &&
+            npx firebase deploy --only hosting:catalog,hosting:festival,hosting:crm,storage &&
+            npx firebase deploy --only firestore --force &&
             npx firebase deploy --only functions --force
```
GrandSchtroumpf, line 266: Looks like we could just run `firebase deploy` at that point no ?

apps/backend-ops/src/admin.ts
```diff
@@ -27,11 +30,18 @@ export function loadAdminServices(): AdminServices {
   if (!admin.apps.length) {
     admin.initializeApp({
       ...firebase,
-      credential: admin.credential.applicationDefault(),
+      credential: admin.credential.cert(
```
GrandSchtroumpf, line 33: It's supposed to be the same isn't it?

apps/backend-ops/src/main.ts
```diff
@@ -1,14 +1,19 @@
+import 'tsconfig-paths/register'; // This is so it can be run via ts-node
```
GrandSchtroumpf, line 1: Best would be to move it into tools. I don't like to have a dependancy in the wild like that. It stays in the repo and after several months nobody remember why it's here. 
Keep it like that for this PR, but let's remove it in the coming days.

apps/backend-ops/src/user-fixture.ts
```diff
@@ -0,0 +1,28 @@
...
+import { loadAdminServices } from './admin';
+
+const { db, auth } = loadAdminServices();
+
+/**
+ * This function reads users from Firestore and generates users fixtures file for repo
+ * @param db firestore instance
+ */
+export async function generateFixturesFile() {
+  const usersQuerySnapshot = await db.collection('users').get();
```
GrandSchtroumpf, line 13: There are utils method in the lib `firebase-utils` to deal with that in a clean way : 
```typescript
import { getCollection } from 'firebase-utils';
...
const users = await getCollection('users');
```

apps/backend-ops/src/user-fixture.ts
```diff
@@ -0,0 +1,28 @@
...
+  const users = usersQuerySnapshot.docs.map((snapshot) => {
+    const user = snapshot.data();
+    user.id = snapshot.id;
```
GrandSchtroumpf, line 16: user has no key "id", it's "uid". And you don't need to add it.

apps/backend-ops/src/user-fixture.ts
```diff
@@ -0,0 +1,28 @@
...
+import { updateUSERS } from './users';
+import { loadAdminServices } from './admin';
+
+const { db, auth } = loadAdminServices();
+
+/**
+ * This function reads users from Firestore and generates users fixtures file for repo
+ * @param db firestore instance
+ */
+export async function generateFixturesFile() {
```
GrandSchtroumpf, line 12: Rename to `generateUserFixtures`

apps/backend-ops/src/user-fixture.ts
```diff
@@ -0,0 +1,28 @@
...
+    user.id = snapshot.id;
+    return user;
+  });
+  const fileOutput = users.map((user) => ({
+    email: user.email,
+    uid: user.uid,
+    password: 'blockframes',
+  }));
+
+  writeFileSync(resolve(process.cwd(), 'tools/users.fixture.json'), JSON.stringify(fileOutput));
```
GrandSchtroumpf, line 25: Use join to keep system specific delimiter

apps/backend-ops/src/user-fixture.ts
```diff
@@ -0,0 +1,28 @@
...
+  updateUSERS(fileOutput);
+  // console.log(fileOutput);
```
GrandSchtroumpf, line 27: Remove if you don't use

apps/backend-ops/src/user-fixture.ts
GrandSchtroumpf: So the function should look like that: 
```typescript
const dest = join(process.cwd(), 'tools/users.fixture.json');
export async function generateUserFixtures() {
  const users = await getCollection('users');
  const output = users.map(user => ({
    email: user.email,
    uid: user.uid,
    password: 'blockframes',
  }));
  await fs.promises.writeFile(dest, JSON.stringify(output));
  return updateUsers();
}
```

apps/backend-ops/src/users.ts
```diff
@@ -12,6 +13,12 @@ import { upsertWatermark } from 'apps/backend-functions/src/internals/watermark'
 import { startMaintenance, endMaintenance, isInMaintenance } from '@blockframes/firebase-utils';
 import { loadDBVersion } from './migrations';
 
+export let USERS = usersFixture;
+
+export function updateUSERS(data: any[]) {
```
GrandSchtroumpf, line 18: I don't really like this method, it creates a huge side effect, quite difficult to manage. At this point it would be better to update the USER object directly in other files (which I don't recommend neither).
Please find a better way to achive your goal without introducing this ind of side effect.

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -0,0 +1,50 @@
+import type * as admin from 'firebase-admin';
+
+/**
+ * deletes all users in Firebase Auth
+ * @param auth the auth instance that you want to clear
```
GrandSchtroumpf, line 5: typo: "clear" the auth

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+export async function deleteAllUsers(auth: admin.auth.Auth) {
+  const allUsers: admin.auth.UserRecord[] = [];
+  let result: any = { pageToken: undefined };
```
GrandSchtroumpf, line 9: Don't use any, prefer an empty object and use const as you don't reassign it: 
```typescript
const result: Partial<{ pageToken: ... }> = {};
```

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -0,0 +1,50 @@
...
+  let timeMsg = 'Loading users took';
+  console.time(timeMsg);
+  do {
+    result = await auth.listUsers(1000, result.pageToken);
+    allUsers.push(...result.users);
+  } while (result.pageToken);
+  console.timeEnd(timeMsg);
+
+  console.log('Starting user deletion...');
+  timeMsg = `Deleting ${allUsers.length} users took`;
```
GrandSchtroumpf, line 21: What the point of reassigning it ? Why not doing it like that directly ?
```typescript
console.log(`Deleting ${allUsers.length} users took`)
```

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+export async function deleteAllUsers(auth: admin.auth.Auth) {
+  const allUsers: admin.auth.UserRecord[] = [];
+  let result: any = { pageToken: undefined };
+
+  console.log('Loading users now...');
+  let timeMsg = 'Loading users took';
+  console.time(timeMsg);
+  do {
+    result = await auth.listUsers(1000, result.pageToken);
```
GrandSchtroumpf, line 15: I'm not sure to understand what pageToken is.

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -0,0 +1,50 @@
...
+    allUsers.push(...result.users);
+  } while (result.pageToken);
+  console.timeEnd(timeMsg);
+
+  console.log('Starting user deletion...');
+  timeMsg = `Deleting ${allUsers.length} users took`;
+  console.time(timeMsg);
+
+  const allUsersUids = allUsers.map(record => record.uid);
+  async function* iterateDeleteUsers(uids: string[]) {
```
GrandSchtroumpf, line 25: i don't understand what this code is doing. Please try to simplify your code to something more readable for any developer :)

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -0,0 +1,50 @@
...
+      const batch = uids.splice(0, 10).map(uid => auth.deleteUser(uid).catch(console.warn))
+      yield Promise.all(batch);
+      await new Promise(resolve => setTimeout(resolve, 100));
+    }
+  }
+  const delUsersIterator = iterateDeleteUsers(allUsersUids);
+  const output = { errors: [], successCount: 0, failureCount: 0 };
+  for await (const deletion of delUsersIterator) {
+    console.log('User deleted...')
+    // // const { errors, failureCount, successCount } = deletion;
```
GrandSchtroumpf, line 36: Remove comments

libs/testing/src/lib/firebase/firebase.ts
```diff
@@ -0,0 +1,27 @@
+import { credential, initializeApp } from 'firebase-admin';
+import { initializeAdminApp, initializeTestApp } from '@firebase/testing';
+
+/**
+ * Get test (emulated) admin with specified user authentication
+ * @param auth object containing uid/displayName and claims
+ */
+export const getEmulatedFirebase = (projectId: string, auth: any) => {
```
GrandSchtroumpf, line 8: What the difference with `initFunctionsTestMock`

libs/testing/src/lib/firebase/firestore.ts
```diff
@@ -0,0 +1,27 @@
+import { resolve } from 'path';
+import { readFileSync } from 'fs';
+import { initializeAdminApp, initializeTestApp, loadFirestoreRules } from '@firebase/testing';
+
+/** Gets an emulated Firestore db object that can circumvent security rules. Must be provided same projectId */
+export function getEmulatedAdminFirestore({ projectId, databaseName }) {
```
GrandSchtroumpf, line 6: It's overkill.

libs/testing/src/lib/firebase/firestore.ts
```diff
@@ -0,0 +1,27 @@
...
+  return initializeAdminApp({ projectId, databaseName }).firestore();
+}
+
+/**
+ * Load and enable Firestore rules file found in firebase.json into emulator
+ * @param projectId a string of project Id
+ */
+export function enableFirestoreRulesFor(projectId: string) {
+  const firebaseConfig = require(resolve(process.cwd(), './firebase.json'));
```
GrandSchtroumpf, line 15: use `join` instead of `resolve`

.circleci/config.yml
george43g: This was actually a fix for an unrelated issue - the purpose was to --force firestore indexes to fix an issue with CI.

apps/backend-ops/src/admin.ts
george43g: It wasn't working for me initially, but you're right yeah it's no longer needed with the `config()` from dotenv.

apps/backend-ops/src/main.ts
george43g: Annoyingly this fix isn't working atm.

apps/backend-ops/src/user-fixture.ts
george43g: `db` is a persistence layer, I don't see much benefit of writing another layer above that. It doesn't seem to save much code, and I don't see how the core code would change to benefit everywhere the function is used. But happy to use it - no worries! :)

george43g: `import { getCollection } from 'firebase-utils';` does not exist in my project?

george43g: and the reason I named it `updateUSERS` instead of camel case is that it literally updates an array variable called `USERS` (this is what it was called originally in the code)

apps/backend-ops/src/user-fixture.ts
george43g: done

apps/backend-ops/src/user-fixture.ts
george43g: Done!

apps/backend-ops/src/user-fixture.ts
george43g: done!

apps/backend-ops/src/users.ts
george43g: This was the simplest way to write the solution while making minimal changes to everything else. I will look into this further, I recall there were some issues around this however.

phpgeekfr:
@george43g can we close?

phpgeekfr:
no response. closing

PR #3380, opened 2020-07-28
fix name participant and reload for invitations

Hugo-Hontans:
- [x] Fix participant names
- [x] Reload after accept invitation

### DIFF
--- a/apps/festival/festival-e2e/src/integration/dashboard/screening-invitations-2688.spec.ts
+++ b/apps/festival/festival-e2e/src/integration/dashboard/screening-invitations-2688.spec.ts
@@ -10,8 +10,6 @@ import {
   PARTICIPANT_1_NAME,
   PARTICIPANT_2_NAME,
   ORG_NAME,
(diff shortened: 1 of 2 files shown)


apps/festival/festival-e2e/src/integration/dashboard/screening-invitations-2688.spec.ts
```diff
@@ -54,7 +52,9 @@ describe('User invites other users to his private screening', () => {
     p2.acceptInvitationScreening();
     // Wait for post request to finish
     // TODO: verify UI has changed instead of wait
-    cy.wait(2000);
+    cy.wait(5000);
```
xmano, line 55: Any other strategy to avoid or wait for correct time? Otherwise we see timeout issues in CI?

Hugo-Hontans: Like we reload the page, we don't need to wait, nice

Issue #3193, opened 2020-06-26
[Component] Selections and Reviews Page


joanofdart:
link to [main issue](https://github.com/blockframes/blockframes/issues/2956)

## Points to consider
- [ ] Data matches their type.
- [ ] Component can be reused
- [ ] Reuse sub-components that correlate to their given purpose.

## Sub-components

**_static-select_**
**_movie-form-input_** Related to #3171
**_mat-slide-toggle_**

## Implement Component

_Festivals and Awards_

- [ ] **Input Name:** Name
  - [ ] **Data Model:** festivalPrizes.prizes.name
      - [ ] **Component:** static-select

- [ ] **Input Name:** Prize
  - [ ] **Data Model:** festivalPrizes.prizes.prize
      - [ ] **Component** movie-form-input

- [ ] **Input Name:** Year
  - [ ] **Data Model:** festivalPrizes.prizes.year
      - [ ] **Component** movie-form-input

- [ ] **Input Name:**  Premiere
  - [ ] **Data Model:** festivalPrizes.prizes.premiere
      - [ ] **Component** mat-slide-toggle

_Festival (Autre)_

- [ ] **Input Name:** Name
  - [ ] **Data Model:** festivaPrizes.others[].name
      - [ ] **Component:** static-select

- [ ] **Input Name:** Prize
  - [ ] **Data Model:** festivaPrizes.others[].prize
      - [ ] **Component** movie-form-input

- [ ] **Input Name:** Year
  - [ ] **Data Model:** festivaPrizes.others[].year
      - [ ] **Component** movie-form-input

- [ ] **Input Name:**  Premiere
  - [ ] **Data Model:** festivaPrizes.others[].premiere
      - [ ] **Component** mat-slide-toggle

_Film Reviews_

- [ ] **Input Name:**  Film Critic's Name
  - [ ] **Data Model:** movieReview.criticName
      - [ ] **Component** mat-form-input

- [ ] **Input Name:**  Rervue / Journal Name
  - [ ] **Data Model:** movieReview.journalName
      - [ ] **Component** mat-form-input

- [ ] **Input Name:**  Quote
  - [ ] **Data Model:** movieReview.criticQuote
      - [ ] **Component** mat-form-input

- [ ] **Input Name:**  
  - [ ] **Data Model:** movieReview.revueLink
      - [ ] **Component** mat-form-input

## Description

- name: movie-form-selection-reviews
- lib/app: libs/movie/src/lib/movie/form/selection-reviews

[...]

PR #3394, opened 2020-07-29
APP-3193 Selection & Awards page.

joanofdart:
Related to #3193

### DIFF
--- a/libs/movie/src/lib/movie/pages/reviews/reviews.component.html
+++ b/libs/movie/src/lib/movie/pages/reviews/reviews.component.html
@@ -1,0 +1,97 @@
+<tunnel-page>
+  <h1 subtitle>Selection & Reviews</h1>
+  <ng-container tunnel-form>
+    <section id="festival">
+      <h3>Festival & Awards</h3>
+      <article *ngFor="let control of prizes.controls; let i = index; let last = last">
+        <div fxLayout fxLayoutGap="24px">
+          <mat-form-field appearance="outline" fxFlex>
+            <mat-label>Name</mat-label>
+            <input test-id="first-name" matInput [formControl]="control.get('name')" required />
+          </mat-form-field>
+
+          <mat-form-field appearance="outline" fxFlex>
+            <mat-label>Award/Selection</mat-label>
+            <input test-id="prize" matInput [formControl]="control.get('prize')" required />
+          </mat-form-field>
+        </div>
+
+        <div fxLayout fxLayoutGap="24px">
+          <mat-form-field appearance="outline" fxFlex>
+            <mat-label>Year</mat-label>
+            <input matInput [formControl]="control.get('year')" required type="number" placeholder="YYYY"/>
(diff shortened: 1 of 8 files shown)


libs/movie/src/lib/movie/pages/reviews/reviews.component.html
```diff
@@ -0,0 +1,97 @@
...
+            <mat-error>Please enter a 4 digit date (e.g. 2020)</mat-error>
+            <mat-error *ngIf="control.hasError('required')">
+              Year is required.
+            </mat-error>
+          </mat-form-field>
+
+          <mat-form-field appearance="outline" fxFlex>
+            <mat-label>Status</mat-label>
+            <mat-select matInput [formControl]="control.get('premiere')">
+              <mat-option *ngFor="let premiere of premiereType" [value]="premiere.slug">
```
fritzschoff, line 32: pls use static select here

libs/movie/src/lib/movie/pages/reviews/reviews.component.scss
```diff
@@ -0,0 +1,18 @@
...
+mat-form-field {
+  width: 100%;
+
+}
+
+h3 {
+  margin-bottom: 24px;
+}
+
+mat-divider {
```
fritzschoff, line 14: it feels like the UI should stick to the design and not creating a new behavior on every page. Can you please check with tehm?

libs/movie/src/lib/movie/pages/reviews/reviews.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Component({
+  selector: 'movie-form-reviews',
+  templateUrl: './reviews.component.html',
+  styleUrls: ['./reviews.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MovieFormReviewscComponent {
+  form = this.shell.form;
+  public premiereType = staticModels.PREMIERE_TYPE;
```
fritzschoff, line 13: you can remove that and use the static select. Input scope is just the string that says use `PREMIERE_TYPE`

libs/movie/src/lib/movie/pages/reviews/reviews.component.scss
joanofdart: this is the same behavior from all the pages we've done so far.

libs/movie/src/lib/movie/pages/reviews/reviews.module.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { MovieFormReviewscComponent } from './reviews.component';
+
+@NgModule({
+  declarations: [MovieFormReviewscComponent],
+  imports: [
+    CommonModule,
+    ReactiveFormsModule,
+
+    // Other Modules
```
fritzschoff, line 27: nice!

PR #3407, opened 2020-07-30
Retire welcome component

### LINKED ISSUE #3008: Remove /auth/welcome page
### Description
This page can be removed as it's no longer used and replace with landing page. 

### Acceptance criteria
- [ ] Remove all the links to `/auth/welcome` and replace it with landing page
- [ ] Modify the e2e tests that use it to lead to landing page instead and adapt the tests so it keeps working
- [ ] WelcomeComponent can be delete as it should be no longer used

xmano:
Closes #3008


apps/catalog/catalog-e2e/src/integration/dashboard/movie-tunnel-end.spec.ts
```diff
@@ -18,7 +19,9 @@ const SYNOPSIS = 'An up-close-and-personal portrait of the fashion icon, Karl La
 const PRODUCTION_YEAR = '2006';
 
 beforeEach(() => {
-  clearDataAndPrepareTest();
+  clearDataAndPrepareTest('/');
+  const p1 = new LandingPage();
+  p1.clickSignup();   
   signInAndNavigateToMain();
```
Hugo-Hontans, line 25: Here you can add 
```
  const p1 = new LandingPage();
  p1.clickSignup();   
```
directly inside signInAndNavigateToMain function I think

libs/e2e/src/lib/utils/functions.ts
```diff
@@ -13,8 +13,7 @@ export function clearDataAndPrepareTest(path: string = '/auth') {
 
 /** Start on AuthWelcomePage, on AuthLoginPage and signin. You have to create a new page depending of the app. */
 export function signIn(user: Partial<User>) {
-  const p1: AuthWelcomePage = new AuthWelcomePage();
-  const p2: AuthLoginPage = p1.clickCallToAction();
+  const p2: AuthLoginPage = new AuthLoginPage();
```
Hugo-Hontans, line 16: change p2 for p1 so plz. Keep the same consistency starting with p1

apps/catalog/catalog-e2e/src/integration/dashboard/movie-tunnel.spec.ts
```diff
@@ -17,7 +18,9 @@ import {
 describe('User can navigate to the movie tunnel pages start and main.', () => {
   // Log in and create a new movie
   it('Login into an existing account, navigate on titles list page, go to movie tunnel start page, go on movie tunnel main page', () => {
-    clearDataAndPrepareTest();
+    clearDataAndPrepareTest('/');
+    const p1 = new LandingPage();
+    p1.clickSignup();    
     signInAndNavigateToMain();
```
Hugo-Hontans, line 24: same

apps/catalog/catalog-e2e/src/integration/dashboard/contract-tunnel.spec.ts
```diff
@@ -24,13 +25,15 @@ const TRIGGERING_EVENT = 'First theatrical release';
 
 beforeEach(() => {
   clearDataAndPrepareTest();
+  cy.visit('/');
```
Hugo-Hontans, line 28: By default inside clearDataAndPrepareTest we use `cy.visit('/auth')` and you can add an other path directly. I let you check inside this function.
And if the function goes in route `/auth` you should be directly on AuthLoginPage so you don't need
```
  const p1 = new LandingPage();
  p1.clickSignup();  
```
I think

libs/e2e/src/lib/utils/functions.ts
xmano: OK

apps/catalog/catalog-e2e/src/integration/dashboard/contract-tunnel.spec.ts
xmano: For flexibility I move this inside `signInAndNavigateToMain`

PR #3409, opened 2020-07-30
Feat 3066 - unit tests setup - part 1

phpgeekfr:
Unit testing part of this PR is still in WIP but it's working. I need to check them with @xmano but this can be made after this pull request.
This PR also add a new lib `firebase-utils`


apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+  cleanNotifications,
+  dayInMillis,
+  numberOfDaysToKeepNotifications,
+  cleanUsers
+} from './db-cleaning';
+import { every } from 'lodash';
+import { AdminAuthMocked } from '../../../libs/testing/src//lib/firebase/adminAuthMocked';
+import { AdminServices } from './admin';
+
+const moviesTestSet = require('../../../libs/testing/src/lib/mocked-data-unit-tests/movies.json'); // @TODO (#3066) commit this file only when fully anonymised
```
GrandSchtroumpf, line 18: use alias `@blockframes/testing`

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+  cleanOrganizations,
+  cleanPermissions,
+  cleanDocsIndex,
+  cleanNotifications,
+  dayInMillis,
+  numberOfDaysToKeepNotifications,
+  cleanUsers
+} from './db-cleaning';
+import { every } from 'lodash';
+import { AdminAuthMocked } from '../../../libs/testing/src//lib/firebase/adminAuthMocked';
```
GrandSchtroumpf, line 15: use alias `@blockframes/testing`

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+
+
+jest.setTimeout(30000);
+
+describe('DB cleaning script', () => {
+  let adminServices: AdminServices;
+
+  beforeAll(async () => {
+    const featList = initFunctionsTestMock();
+    adminServices = {
```
GrandSchtroumpf, line 34: Flatten everything :
```typescript
const app = initFunctionsTestMock();
db = app.firestore();
auth = app.auth();
storage = app.storage();
```

phpgeekfr: I will add a comment in code to find a way for this with @xmano

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+  it('should clean movies from unwanted attributes', async () => {
+    const moviesBefore: FirebaseFirestore.QuerySnapshot<FirebaseFirestore.DocumentData> = await adminServices.db.collection('movies').get();
+    await cleanMovies(moviesBefore);
+    const moviesAfter: FirebaseFirestore.QuerySnapshot<FirebaseFirestore.DocumentData> = await adminServices.db.collection('movies').get();
+    const cleanedMovies = moviesAfter.docs.filter(m => isMovieClean(m)).length;
+    expect(moviesTestSet.length).toEqual(cleanedMovies);
+  });
+  it('should remove documents undefined or not linked to existing document from docsIndex', async () => {
+    const [docsIndexBefore, movies,] = await Promise.all([
+      adminServices.db.collection('docsIndex').get(),
```
GrandSchtroumpf, line 141: I would recommend to create a `collectionRef(path: string)` reusable function so you don't need to repeat yourself

apps/backend-ops/src/db-cleaning.ts
```diff
@@ -163,32 +159,46 @@ async function _cleanInvitation(doc: any, invitation: any) { // @TODO (#3175) w8
...
-        if (invalidOrganization) {
+        let update = false;
+
+        if (invalidOrganization) { // @todo #3066 create a test for this
           delete user.orgId;
-          await userDoc.ref.update(user);
+          update = true;
+        }
+
+        if (user.name !== undefined) {
```
GrandSchtroumpf, line 190: what about null or empty string ?

libs/testing/src/lib/mocked-data-unit-tests/auth.json
```diff
@@ -0,0 +1,26 @@
+[
```
GrandSchtroumpf, line 1: I would recommend to use typescript object instead so you can have type safety and update it after each migration (in the case of collections) : 
```typescript
const auth: FirebaseUser[] = [{
  ...
}]
```

libs/testing/src/lib/scripts/create-mocked-data-db-cleaning.ts
```diff
@@ -0,0 +1,219 @@
...
+    movies: [],
+    orgs: [],
+    permissions: [],
+    notifications: [],
+    invitations: [],
+    events: [],
+    users: [],
+  };
+
+  const data = await fs.readFileSync(file, 'utf8');
```
GrandSchtroumpf, line 160: If it's readFileSync you shouldn't need the await. You can do `await fs.promises.readFile` to use promise

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+  beforeAll(async () => {
+    const featList = initFunctionsTestMock();
+    adminServices = { // @TODO #3066 Flatten everything
+      auth: auth(),
+      db: firestore(),
+      storage: storage(),
+      firebaseConfig: featList.firebaseConfig
+    };
+
+    console.log('loading movies data set...');
```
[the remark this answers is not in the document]
phpgeekfr: in firebase emulator

libs/testing/src/lib/mocked-data-unit-tests/auth.json
phpgeekfr: It will be hard to type since this data can be false on purpose to check db cleaning script

apps/backend-ops/src/db-cleaning.spec.ts
phpgeekfr: added comment

PR #3410, opened 2020-07-30
Storage as a source of truth to the db

RemcoSimonides:
My gift for Hugo

### DIFF
--- a/apps/backend-ops/src/firebaseSetup.ts
+++ b/apps/backend-ops/src/firebaseSetup.ts
@@ -10,6 +10,7 @@ import { migrate } from './migrations';
 import { restore } from './admin';
 import { cleanDeprecatedData } from './db-cleaning';
 import { cleanStorage } from './storage-cleaning';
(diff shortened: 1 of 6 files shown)


apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -40,6 +41,10 @@ export async function prepareForTesting() {
   await generateWatermarks();
   console.info('Watermarks generated!');
 
+  console.info('Syncing firestore with storage');
+  await syncStorage();
+  console.info('Firestore is now synced with storage!');
```
phpgeekfr, line 46: Can you also add this in the "upgrade" section ? Because I think we will have to run this on each upgrade

phpgeekfr: also syncStorage is already in `generateWatermarks` function

phpgeekfr:
@RemcoSimonides can you also comment migration 31 in `index.ts` ?

PR #3411, opened 2020-07-30
Invitation unit test

Hugo-Hontans:
- [x] Create basic unit tests for invitation service

### DIFF
--- a/libs/invitation/src/lib/+state/invitation.service.spec.ts
+++ b/libs/invitation/src/lib/+state/invitation.service.spec.ts
@@ -34,7 +34,7 @@ describe('Invitations Test Suite', () => {
 
   afterEach(() => clearFirestoreData({projectId: 'test'}))
 
-  it('Should check notif service is created', () => {
+  it('Should check invitation service is created', () => {
     expect(service).toBeTruthy();
   })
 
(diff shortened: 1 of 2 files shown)


libs/invitation/src/lib/+state/invitation.service.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it('Should check notif service is created', () => {
+    expect(service).toBeTruthy();
+  })
+
+  it('Formats invitation from firestore', () => {
+    const is = TestBed.inject(InvitationService);
+    is.formatFromFirestore = jest.fn();
+    is.formatFromFirestore({} as any);
+    expect(is.formatFromFirestore).toHaveBeenCalled();
```
xmano, line 45: You can check the function does it work exactly according to the logic needed here since there is no need to mock firestore here.

libs/invitation/src/lib/+state/invitation.service.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('Formats invitation to firestore', () => {
+    const is = TestBed.inject(InvitationService);
+    is.formatToFirestore = jest.fn();
+    is.formatToFirestore({} as any);
+    expect(is.formatToFirestore).toHaveBeenCalled();
```
xmano, line 52: Same as above. You need to check if output of formatToFirestore() is according to spec needed.

Also there are other functions inside the service, so you need to implement at least 1 test per function or more depending on what the function does.

libs/invitation/src/lib/+state/invitation.service.spec.ts
Hugo-Hontans: I tried to create mocked invitation and to use `toHaveReturnedWith()` but get `undefined`. I didn't find a solution for that

libs/invitation/src/lib/+state/invitation.service.spec.ts
Hugo-Hontans: I created the issue #3415 for that

xmano (approved):
Kindly make sure UTs passes with latest rebase to the branch before merge.

PR #3420, opened 2020-08-10
clean models & forms

pldespaigne:


apps/backend-ops/src/firestoreMigrations/0028.ts
```diff
@@ -76,9 +79,9 @@ async function updateNotifications(notifications: FirebaseFirestore.QuerySnapsho
     const notification = doc.data() as NotificationDocument;
 
     if (notification.organization) {
-      notification.organization.logo = createHostedMedia();
+      (notification.organization.logo as any) = createHostedMedia();
```
GrandSchtroumpf, line 82: Is it not better to use a type with a migration id ? Else it's going to be difficult to maintain with a lot of as any...

PR #3423, opened 2020-08-11
form-list/table comp.

fritzschoff:


libs/movie/src/lib/movie/pages/main/main.component.html
```diff
@@ -117,38 +119,53 @@ <h3>General Information</h3>
...
-        </footer>
-      </article>
-      <footer fxLayout="column" fxLayoutAlign="center center">
-        <button mat-mini-fab color="primary" (click)="directors.add()" aria-label="Remove a director form">
-          <mat-icon svgIcon="add"></mat-icon>
-        </button>
-        <span class="mat-caption">Add Director</span>
-      </footer>
+      <bf-form-table [displayedColumns]="displayedColumns" [form]="directors" buttonText="Add new Director">
+        <ng-template colRef="First Name" let-director>{{ director.firstName }} </ng-template>
```
GrandSchtroumpf, line 123: I've got the feeling that you don't need to repeat that all the time.
You could do someting like : 
```typescript
columns = {
  firstName: "First Name",
  lastName: "Last Name",
  ...
}
```
And in the bf-form-table
```html
<ng-container *ngFor="for column of column | keyvalue">
  <ng-container *ngIf="template[column.key] else default">
    <ng-template [ngTemplateOutlet]="template[column.key]" ...>
  </ng-container>
  <ng-template #default>
     {{ item[column.key] }}
  </ng-template>
</ng-container>
```

We can discuss this.

libs/ui/src/lib/form/list/form-list.component.html
```diff
@@ -30,4 +30,10 @@
 
 <ng-template [ngTemplateOutlet]="formView" [ngTemplateOutletContext]="{ $implicit: formItem }">
 </ng-template>
-<bf-line-button [buttonText]="buttonText" (buttonClick)="addControl()"></bf-line-button>
+<div class="line-button">
+  <mat-divider id="line"></mat-divider>
```
GrandSchtroumpf, line 34: Do not use id as there will be any "line" in the same page.

libs/ui/src/lib/form/list/form-list.component.html
```diff
@@ -30,4 +30,10 @@
 
 <ng-template [ngTemplateOutlet]="formView" [ngTemplateOutletContext]="{ $implicit: formItem }">
 </ng-template>
-<bf-line-button [buttonText]="buttonText" (buttonClick)="addControl()"></bf-line-button>
+<div class="line-button">
```
GrandSchtroumpf, line 33: Why not `<footer fxLayout="start baseline" fxLayoutGap="16px">` ?

libs/ui/src/lib/form/list/form-list.component.html
GrandSchtroumpf: If you use fxLayout as suggested by the comment above you need to set `<mat-divider fxFlex></mat-divider>`

libs/ui/src/lib/form/table/form-table.component.html
```diff
@@ -45,4 +46,10 @@
 ​
 <ng-template [ngTemplateOutlet]="formView" [ngTemplateOutletContext]="{ $implicit: formItem }">
 </ng-template>
-<bf-line-button [buttonText]="buttonText" (buttonClick)="addControl()"></bf-line-button>
+<div class="line-button">
```
GrandSchtroumpf, line 49: Same as above

libs/ui/src/lib/form/table/form-table.component.ts
```diff
@@ -53,7 +53,7 @@ export class FormTableComponent<T> implements OnInit, AfterViewInit, OnDestroy {
   activeIndex: number;
   pageSize = 5;
   /* We need to keep track of the current page since it will affect the index that we are working on */
-  pageConfig = { pageIndex: 0, pageSize: this.pageSize };
+  pageConfig = { pageIndex: 0, pageSize: 5 };
```
GrandSchtroumpf, line 56: It's always 5 ?

libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -39,10 +39,16 @@ export class FormList<T, Control extends AbstractControl = any> extends FormArra
...
-      const controls = value.map(createControl)
-      const form = new FormList<T>(controls, validators);
-      form['createControl'] = createControl.bind(form);
-      return form;
+      if (Array.isArray(value)) {
+        const controls = value.map(createControl)
+        const form = new FormList<T>(controls, validators);
+        form['createControl'] = createControl.bind(form);
+        return form;
+      } else {
```
GrandSchtroumpf, line 47: In which case it will not be an array ? Normally this should not happen.

libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -39,10 +39,16 @@ export class FormList<T, Control extends AbstractControl = any> extends FormArra
...
+        const control = createControl(value);
+        const form = new FormList<T>([control], validators);
+        form['createControl'] = createControl.bind(form);
```
GrandSchtroumpf, line 50: You're not returning the form.
If this is truly needed to have a if statement try
```typescript
const controls = Array.isArray(value) ? value.map(createControl) : [createControl(value)];
```

libs/ui/src/lib/form/list/form-list.component.html
fritzschoff: this is a selector to override the default border-top-color and border-top-width. Without an id you can't style the line

libs/ui/src/lib/form/table/form-table.component.ts
fritzschoff: for starters,yes

libs/utils/src/lib/form/forms/list.form.ts
fritzschoff: got an error `value.map` is not a function => line 42. couldn't find where it originates, so that is why I created a fallback

libs/movie/src/lib/movie/pages/main/main.component.html
fritzschoff: columns can't be an object, since we are using it for the table columns, `displayedColumns` .
Also if you don't use the colRef directives anymore you are going to take away your might needed possibilites to alter the displayed value, if you save the slug in the formControl but want to display the label, for instance. 
But we can have a quick call / chat about it

libs/ui/src/lib/form/list/form-list.component.html
fritzschoff: don't want to import FlexLayoutModule, just for that.

libs/ui/src/lib/form/list/form-list.component.html
GrandSchtroumpf: If you do `.line-button mat-divider` it should be good I think.

libs/ui/src/lib/form/list/form-list.component.html
GrandSchtroumpf: It won't increase the bundle size ^^.

fritzschoff: okay theeeeeeeeeeeen I do it :D

PR #3424, opened 2020-08-11
add new static

Mussuranae:
Add 2 new static : one for director's category and one for the credit status

### DIFF
--- a/libs/movie/src/lib/movie/pages/artistic/artistic.component.ts
+++ b/libs/movie/src/lib/movie/pages/artistic/artistic.component.ts
@@ -10,7 +10,7 @@ import { staticModels } from '@blockframes/utils/static-model';
 })
 export class MovieFormArtisticComponent {
   form = this.shell.form;
-  public castStatus = staticModels.MEMBER_STATUS;
+  public memberStatus = staticModels.MEMBER_STATUS;
 
   constructor(private shell: MovieFormShellComponent) {}
 
--- a/libs/movie/src/lib/movie/pages/artistic/artistic.component.ts
+++ b/libs/movie/src/lib/movie/pages/artistic/artistic.component.ts
@@ -10,7 +10,7 @@ import { staticModels } from '@blockframes/utils/static-model';
 })
 export class MovieFormArtisticComponent {
   form = this.shell.form;
-  public castStatus = staticModels.CAST_STATUS;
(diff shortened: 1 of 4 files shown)


libs/utils/src/lib/static-model/staticModels.ts
```diff
@@ -108,6 +108,38 @@ const models = {
       'label': 'Production Manager'
     },
   ] as const,
+  'DIRECTOR_CATEGORY': [
```
phpgeekfr, line 111: DIRECTOR_STATUS ? or TYPES?

libs/movie/src/lib/movie/pages/artistic/artistic.component.ts
```diff
@@ -10,7 +10,7 @@ import { staticModels } from '@blockframes/utils/static-model';
 })
 export class MovieFormArtisticComponent {
   form = this.shell.form;
-  public castStatus = staticModels.CAST_STATUS;
+  public castStatus = staticModels.MEMBER_STATUS;
```
phpgeekfr, line 13: no need to rename `public castStatus` as well ?

libs/utils/src/lib/static-model/staticModels.ts
```diff
@@ -218,16 +250,6 @@ const models = {
       'label': 'Completed'
     }
   ] as const,
-  'CAST_STATUS': [
-    {
-      'slug': 'notconfirmed',
```
phpgeekfr, line 223: do we need a migration for this ?

libs/movie/src/lib/movie/pages/artistic/artistic.component.ts
Mussuranae: Yes I can, thanks !

libs/utils/src/lib/static-model/staticModels.ts
Mussuranae: It will be the same for director/cast/crew, so the new one MEMBER_STATUS

libs/utils/src/lib/static-model/staticModels.ts
Mussuranae: I don't think so, But I will check on prod if we have some people that update this field.

Issue #3194, opened 2020-06-26
[Component] Additional Information Page


joanofdart:
link to [main issue](https://github.com/blockframes/blockframes/issues/2956)

## Points to consider
- [ ] Data matches their type.
- [ ] Component can be reused
- [ ] Reuse sub-components that correlate to their given purpose.

## Sub-components

**_chips-autocomplete_**
**_static-select_**
**_movie-form-input_** => Related to #3170
**_mat-slide-toggle_**
**_mat-datepicker_**

## Implement Component

_Release_

- [ ] **Input Name:** Country
  - [ ] **Data Model:** salesInfo.originalRelease.country
      - [ ] **Component:** chips-autocomplete 

- [ ] **Input Name:** Release media
  - [ ] **Data Model:** salesInfo.originalRelease.media
      - [ ] **Component:** static-select

- [ ] **Input Name:** Date
  - [ ] **Data Model:** salesInfo.originalRelease.date
      - [ ] **Component:** mat-datepicker

_Box Office_

- [ ] **Input Name:** Country
  - [ ] **Data Model:** budget.boxOffice.territory
      - [ ] **Component:** chips-autocomplete

- [ ] **Input Name:** Metrics
  - [ ] **Data Model:** budget.boxOffice.unit
      - [ ] **Component:** static-select

- [ ] **Input Name:** Earnings
  - [ ] **Data Model:** budget.boxOffice.value
      - [ ] **Component:** movie-form-input

_Rating_

- [ ] **Input Name:** Rating
  - [ ] **Data Model:** salesInfo.rating.value
      - [ ] **Component:** movie-form-input

- [ ] **Input Name:** Country
  - [ ] **Data Model:** salesInfo.rating.country
      - [ ] **Component:** chips-autocomplete

_Qualifications_

- [ ] **Input Name:** Qualifications
  - [ ] **Data Model:** salesInfo.certifications
      - [ ] **Component:** mat-slide-toggle

## Description

- name: movie-form-additional-information
- lib/app: libs/movie/src/lib/movie/form/additional-information

## Type
- [ ] **Layout**: _Structure a page with content projection (ex: MovieView)._
- [ ] **Page**: _Loaded by the router. Usually used in an app (ex: CatalogDashboardMovieView)._
- [ ] **Component**: _Structure a piece of data in the page (ex: MovieTable)._
- [ ] **Display**: _Render data in a page. Can use a component-type (ex: MovieMain, MovieBudget)._
[...]

PR #3425, opened 2020-08-11
APP-3194 Additional Info Page done

joanofdart:
Related to #3194

### DIFF
--- a/libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
+++ b/libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
@@ -24,14 +24,14 @@
           <button mat-mini-fab color="warn" (click)="originalRelease.removeAt(i)" aria-label="Remove a Release">
             <mat-icon svgIcon="trash"></mat-icon>
           </button>
-          <span class="mat-caption">Delete a Release</span>
+          <span>Delete a Release</span>
         </footer>
       </article>
       <footer fxLayout="column" fxLayoutAlign="center center">
         <button mat-mini-fab color="primary" (click)="originalRelease.add()" aria-label="Add a Release">
           <mat-icon svgIcon="add"></mat-icon>
         </button>
-        <span class="mat-caption">Add a Release</span>
+        <span>Add a Release</span>
       </footer>
     </section>
 
@@ -89,7 +89,7 @@
           </mat-button-toggle>
         </ng-container>
       </mat-button-toggle-group>
-      <p class="mat-caption">Select all that apply</p>
+      <p>Select all that apply</p>
     </section >
 
 
--- a/libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
+++ b/libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
@@ -1,0 +1,97 @@
+<tunnel-page>
+  <h1 subtitle>Additional Information</h1>
+  <ng-container tunnel-form>
+    <section id="release">
+      <h3>Release</h3>
+      <article *ngFor="let control of originalRelease.controls; let i = index; let last = last">
+        <div fxLayout fxLayoutGap="24px">
+          <chips-autocomplete [form]="control.get('country')" model="TERRITORIES" placeholder="Type and Select a Country" fxFlex></chips-autocomplete>
+
+          <static-select [control]="control.get('media')" scope="MEDIAS" label="Role" label="Select Release Media" fxFlex></static-select>
+        </div>
+
+        <mat-form-field appearance="outline">
+          <mat-label>Release Date</mat-label>
+          <input matInput [formControl]="control.get('date')" [matDatepicker]="datePicker" placeholder="YYYY/MM/DD">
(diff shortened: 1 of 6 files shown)


libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
```diff
@@ -0,0 +1,97 @@
...
+            <mat-icon svgIcon="trash"></mat-icon>
+          </button>
+          <span class="mat-caption">Delete a Release</span>
+        </footer>
+      </article>
+      <footer fxLayout="column" fxLayoutAlign="center center">
+        <button mat-mini-fab color="primary" (click)="originalRelease.add()" aria-label="Add a Release">
+          <mat-icon svgIcon="add"></mat-icon>
+        </button>
+        <span class="mat-caption">Add a Release</span>
```
fritzschoff, line 34: @GrandSchtroumpf this bothers me big time, since a while. The html `caption` doesn't have the styling of the `mat-caption` class. Would be much cleaner html to just use the `caption` html tag and have the styling you want, and not adding an extra class to that!

libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
```diff
@@ -0,0 +1,97 @@
...
+        </div>
+
+        <mat-form-field appearance="outline">
+          <mat-label>Earnings</mat-label>
+          <input test-id="earnings" matInput type="number" [formControl]="control.get('value')" min="0">
+          <mat-error *ngIf="control.get('value').hasError('min')">
+            Only positive values are accepted.
+          </mat-error>
+          <span *ngIf="units[i] === '$'" matPrefix>{{ units[i] }}&nbsp;</span>
+          <span *ngIf="units[i] !== '$'" matSuffix>&nbsp;{{ units[i] }}</span>
```
fritzschoff, line 63: maybe use `; else notDollar` then you don't need to repeat yourself

libs/movie/src/lib/movie/pages/additional-information/additional-information.component.ts
```diff
@@ -0,0 +1,55 @@
+import { Component, ChangeDetectionStrategy, OnInit } from '@angular/core';
```
fritzschoff, line 1: I'm really glad that I got a brave colleague that sorts the import just like me, if you want you can do it in the component as well. 
But I can understand if you think that is just a bit too much of an overhead.

libs/movie/src/lib/movie/pages/additional-information/additional-information.module.ts
```diff
@@ -0,0 +1,64 @@
...
+import { MovieFormStoryModule } from '@blockframes/movie/form/story/story.module';
+import { MovieFormKeywordsModule } from '@blockframes/movie/form/promotional-description/keywords/keywords.module';
+
+// Blockframes UI
+import { TunnelPageModule } from '@blockframes/ui/tunnel';
+import { StaticSelectModule } from '@blockframes/ui/static-autocomplete/select/static-select.module';
+import { ChipsAutocompleteModule } from '@blockframes/ui/static-autocomplete/chips/chips-autocomplete.module';
+
+import { MovieFormAdditionalInformationComponent } from './additional-information.component';
+import { FlexLayoutModule } from '@angular/flex-layout';
```
fritzschoff, line 25: if you want, move it up to the corresponding section

libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
joanofdart: Actually I just copy/pasted this x) so that might be something I need to remove

libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
joanofdart: Still you have to add another block below with the else template.

libs/movie/src/lib/movie/pages/additional-information/additional-information.component.ts
joanofdart: <3 I actually forgot to sort some, let me fix that!

libs/movie/src/lib/movie/pages/additional-information/additional-information.module.ts
joanofdart: Yup, this one was the one I was talking about above lol

libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
joanofdart: Ahh I see what you're talking about. We'll have to modify these on every page then.

libs/movie/src/lib/movie/pages/additional-information/additional-information.component.html
fritzschoff: Sure. It was not about saving a line, it is more like that JavaScript is a high abstract language and with that, you almost can create human readable sentences. But this is really the cherry on top and it is not a big game changer, just wanted to mark it.

PR #3428, opened 2020-08-11
validators updated

Mussuranae:
Rework of the validators

### DIFF
--- a/libs/movie/src/lib/movie/form/main/genres/genres.component.html
+++ b/libs/movie/src/lib/movie/form/main/genres/genres.component.html
@@ -10,11 +10,14 @@
 <article fxLayout="row" fxLayoutAlign="start start" fxLayoutGap="24px">
   <mat-form-field appearance="outline">
     <mat-label>Genres</mat-label>
-    <mat-select>
+    <mat-select required>
       <mat-option [value]="genre.slug" *ngFor="let genre of genres$ | async" (click)="add(genre.slug)">
         {{ genre.label }}
       </mat-option>
     </mat-select>
+    <mat-error *ngIf="!(genres$ | async).length">
+      Please specify your project's genre(s).
+    </mat-error>
   </mat-form-field>
 </article>
 
--- a/libs/movie/src/lib/movie/form/movie.form.ts
+++ b/libs/movie/src/lib/movie/form/movie.form.ts
@@ -93,7 +93,7 @@ function createMovieControls(movie: Partial<Movie>) {
     cast: FormList.factory(entity.cast, el => new CreditForm(el)),
     certifications: new FormControl(entity.certifications),
     color: new FormControl(entity.color),
-    contentType: new FormControl(entity.contentType),
+    contentType: new FormControl(entity.contentType, [Validators.required]),
     crew: FormList.factory(entity.crew, el => new CreditForm(el)),
     customGenres: FormList.factory(entity.customGenres),
     directors: FormList.factory(entity.directors, el => new DirectorForm(el)),
@@ -101,15 +101,15 @@ function createMovieControls(movie: Partial<Movie>) {
     estimatedBudget: new FormControl(entity.estimatedBudget),
     format: new FormControl(entity.format),
     formatQuality: new FormControl(entity.formatQuality),
-    genres: new FormStaticArray(entity.genres, 'GENRES'),
-    internalRef: new FormControl(entity.internalRef),
+    genres: new FormStaticArray(entity.genres, 'GENRES', [Validators.required]),
+    internalRef: new FormControl(entity.internalRef, [Validators.maxLength(30)]),
     keyAssets: new FormControl(entity.keyAssets, [Validators.maxLength(750)]),
     keywords: FormList.factory(entity.keywords),
     languages: MovieVersionInfoForm.factory(entity.languages, createLanguageControl),
-    logline: new FormControl(entity.logline, [Validators.maxLength(180)]),
+    logline:  new FormControl(entity.logline, [Validators.maxLength(350)]),
     originalLanguages: FormList.factory(entity.originalLanguages, el => new FormStaticValue(el, 'LANGUAGES')),
     originalRelease: FormList.factory(entity.originalRelease, el => new OriginalReleaseForm(el)),
-    originCountries: FormList.factory(entity.originCountries, el => new FormStaticValue(el, 'TERRITORIES')),
+    originCountries: FormList.factory(entity.originCountries, el => new FormStaticValue(el, 'TERRITORIES'), [Validators.required]),
     poster: new MoviePromotionalHostedMediaForm(entity.poster),
     prizes: FormList.factory(entity.prizes, el => new MoviePrizeForm(el)),
     producers: FormList.factory(entity.producers, el => new CreditForm(el)),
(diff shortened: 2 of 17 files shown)


libs/movie/src/lib/movie/form/main/genres/genres.component.html
```diff
@@ -10,11 +10,14 @@
 <article fxLayout="row" fxLayoutAlign="start start" fxLayoutGap="24px">
   <mat-form-field appearance="outline">
     <mat-label>Genres</mat-label>
-    <mat-select>
+    <mat-select required>
```
GrandSchtroumpf, line 13: This is a deprecated component it shouldn't be used. We should use static-select instead.

libs/movie/src/lib/movie/form/movie.form.ts
```diff
@@ -93,23 +93,23 @@ function createMovieControls(movie: Partial<Movie>) {
...
     crew: FormList.factory(entity.crew, el => new CreditForm(el)),
     customGenres: FormList.factory(entity.customGenres),
     directors: FormList.factory(entity.directors, el => new DirectorForm(el)),
     // We use FormControl because objet { from, to } is one value (cannot update separately)
     estimatedBudget: new FormControl(entity.estimatedBudget),
     format: new FormControl(entity.format),
     formatQuality: new FormControl(entity.formatQuality),
-    genres: new FormStaticArray(entity.genres, 'GENRES'),
-    internalRef: new FormControl(entity.internalRef),
+    genres: new FormStaticArray(entity.genres, 'GENRES', [Validators.required]),
```
GrandSchtroumpf, line 104: Maybe it would be better to have `[Validators.minLength(1)]` instead.

libs/movie/src/lib/movie/form/movie.form.ts
```diff
@@ -93,23 +93,23 @@ function createMovieControls(movie: Partial<Movie>) {
...
+    internalRef: new FormControl(entity.internalRef, [Validators.maxLength(30)]),
     keyAssets: new FormControl(entity.keyAssets, [Validators.maxLength(750)]),
     keywords: FormList.factory(entity.keywords),
     languages: MovieVersionInfoForm.factory(entity.languages, createLanguageControl),
-    logline: new FormControl(entity.logline, [Validators.maxLength(180)]),
-    originalLanguages: FormList.factory(entity.originalLanguages, el => new FormStaticValue(el, 'LANGUAGES')),
+    logline:  new FormControl(entity.logline, [Validators.maxLength(350)]),
+    originalLanguages: FormList.factory(entity.originalLanguages, el => new FormStaticValue(el, 'LANGUAGES'), [Validators.required]),
```
GrandSchtroumpf, line 110: same as above for every list. It makes more sense I think

libs/movie/src/lib/movie/form/movie.form.ts
```diff
@@ -344,7 +344,7 @@ export type BoxOfficeFormControl = ReturnType<typeof createBoxOfficeFormControl>
 
 export class BoxOfficeForm extends FormEntity<BoxOfficeFormControl> {
   constructor(boxOffice?: Partial<BoxOffice>) {
-    super(createBoxOfficeFormControl(boxOffice))
+    super(createBoxOfficeFormControl(boxOffice), [boxOfficeRequired])
```
GrandSchtroumpf, line 347: Just make territory & value required instead of creating a custom validator.

libs/movie/src/lib/movie/form/movie.form.ts
```diff
@@ -560,6 +567,16 @@ function createRunningTimeFormControl(runningTime?: Partial<Movie['runningTime']
 
 type RunningTimeFormControl = ReturnType<typeof createRunningTimeFormControl>;
 
+const runningTimeRequired: ValidatorFn = (form: RunningTimeForm) => {
+  const time = form.get('time').value;
```
GrandSchtroumpf, line 571: I think it take a lot of code for : 
```typescript
const { time, status } = form.value;
return status === 'confirmed' && !time
  ? { timeRequired: true }
  : null;
```

libs/movie/src/lib/movie/pages/main/main.component.html
```diff
@@ -78,39 +75,62 @@ <h3 type>Poster*</h3>
...
         </mat-form-field>
 
-        <!-- TODO uncomment this when data model is ready -->
-        <!-- <static-select scope="RELEASE_STATUS" [control]="form.get('status')" label="Estimated/Confirmed" fxFlex></static-select> -->
+        <static-select
+        scope="RELEASE_STATUS"
+        [control]="form.get('release').get('status')"
+        label="Estimated/Confirmed *"
+        error="form.get('release').get('status').hasError('required')"
+        errorText="Please specify if the release date is estimated or confirmed."
```
GrandSchtroumpf, line 93: Can you use mat-error instead.

libs/ui/src/lib/static-autocomplete/chips/chips-autocomplete.component.html
```diff
@@ -1,4 +1,7 @@
 <mat-form-field appearance="outline">
+  <mat-label>
+    <ng-content select="mat-label"></ng-content>
```
GrandSchtroumpf, line 3: 👍

libs/ui/src/lib/static-autocomplete/select/static-select.component.ts
```diff
@@ -18,6 +18,8 @@ export class StaticSelectComponent {
   @Input() label?: string;
   @Input() hintText?: string;
   @Input() hintAlign: string;
+  @Input() errorText?: string;
```
GrandSchtroumpf, line 21: Best to remove that before we forgot.

libs/ui/src/lib/static-autocomplete/select/static-select.component.html
```diff
@@ -12,4 +12,7 @@
   <mat-hint *ngIf="hintText" [align]="hintAlign">
     {{ hintText }}
   </mat-hint>
+  <mat-error *ngIf="error">
+    {{ errorText }}
```
GrandSchtroumpf, line 16: Use ng-content select="mat-error"

libs/movie/src/lib/movie/form/main/genres/genres.component.html
Mussuranae: The problem with genre is the other genre. We use it in the new pages so I change it to get the good modification.

libs/movie/src/lib/movie/form/movie.form.ts
Mussuranae: It's what I did at the beginning but I got an error and like this it was working... :/

libs/movie/src/lib/movie/form/main/genres/genres.component.html
GrandSchtroumpf: Indeed.

libs/movie/src/lib/movie/form/movie.form.ts
GrandSchtroumpf: Then you didn't wrote it the right way. I'm pretty sure that if your code is working, my is too.

PR #3430, opened 2020-08-12
created firebase tools lib

phpgeekfr:


libs/firebase-tools/tsconfig.lib.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../../dist/out-tsc",
+    "target": "es2015",
+    "declaration": true,
+    "inlineSources": true,
+    "types": [],
+    "lib": ["dom", "es2018"]
```
GrandSchtroumpf, line 9: I don't think you need dom

libs/firebase-tools/tsconfig.lib.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../../dist/out-tsc",
+    "target": "es2015",
+    "declaration": true,
+    "inlineSources": true,
+    "types": [],
```
GrandSchtroumpf, line 8: Maybe add "node" here

PR #3433, opened 2020-08-12
George/get prod backup

george43g:
related to #3162

This is an addition of a script which downloads latest prod backup


tools/scripts/get-prod-backup.ts
```diff
@@ -0,0 +1,33 @@
...
+  projectId: firebase.projectId,
+  credential: admin.credential.cert(process.env.GOOGLE_APPLICATION_CREDENTIALS),
+});
+
+const storage = admin.storage();
+
+async function getProdBackup() {
+  const [files] = await storage.bucket(backupBucket).getFiles();
+  const sortableFiles = files.map((file) => ({ file, date: new Date(file.metadata?.timeCreated) }));
+  // for (const file of files) {
```
GrandSchtroumpf, line 17: Can you remove this ?

tools/scripts/get-prod-backup.ts
```diff
@@ -0,0 +1,33 @@
...
+  storageBucket: backupBucket,
+  projectId: firebase.projectId,
+  credential: admin.credential.cert(process.env.GOOGLE_APPLICATION_CREDENTIALS),
+});
+
+const storage = admin.storage();
+
+async function getProdBackup() {
+  const [files] = await storage.bucket(backupBucket).getFiles();
+  const sortableFiles = files.map((file) => ({ file, date: new Date(file.metadata?.timeCreated) }));
```
GrandSchtroumpf, line 16: I would have sort with the alphabetic order on name (which is the timestamp). But using timeCreated looks ok for me.

tools/scripts/get-prod-backup.ts
```diff
@@ -0,0 +1,33 @@
...
+  // for (const file of files) {
+  //   const date = new Date(file.metadata?.timeCreated);
+  //   console.log(date);
+  // }
+  let latestFile = sortableFiles.pop();
```
GrandSchtroumpf, line 21: You can use `.sort` :
```typescript
const last = files.sort((a, b) => b.date - a.date).pop(); // I think this is (b - a) here, but not sure.
```

tools/scripts/get-prod-backup.ts
```diff
@@ -0,0 +1,33 @@
...
+  for (const { date, file } of sortableFiles) {
+    if (date > latestFile.date) {
+      latestFile = { date, file };
+    }
+  }
+  console.log(`Latest backup: ${latestFile.date}`);
+  const destination = `${process.cwd()}/${latestFile.file.name}`;
```
GrandSchtroumpf, line 28: use `join` instead because of windows "\".

tools/scripts/get-prod-backup.ts
```diff
@@ -0,0 +1,33 @@
...
+  //   const date = new Date(file.metadata?.timeCreated);
+  //   console.log(date);
+  // }
+  let latestFile = sortableFiles.pop();
+  for (const { date, file } of sortableFiles) {
+    if (date > latestFile.date) {
+      latestFile = { date, file };
+    }
+  }
+  console.log(`Latest backup: ${latestFile.date}`);
```
GrandSchtroumpf, line 27: prefere `console.log('Latest backup: ', latestFile.date)` when possible.

tools/scripts/get-prod-backup.ts
```diff
@@ -0,0 +1,33 @@
...
+  const destination = `${process.cwd()}/${latestFile.file.name}`;
+  await latestFile.file.download({ destination });
```
GrandSchtroumpf, line 29: I would do `return` instead of await here. Just in case we reuse this function.

PR #3435, opened 2020-08-12
Db cleaning ut

xmano:
Fix init code


apps/backend-ops/src/admin.ts
```diff
@@ -24,10 +24,12 @@ export interface AdminServices {
 }
 
 export function loadAdminServices(): AdminServices {
-  admin.initializeApp({
-    ...firebase,
-    credential: admin.credential.applicationDefault(),
-  });
+  if (!admin.apps.length) {
```
phpgeekfr, line 27: I think I already updated it

libs/testing/src/lib/firebase/functions.ts
```diff
@@ -23,11 +23,9 @@ config();
  */
 export function initFunctionsTestMock(offline = true, overrideConfig?: AppOptions): FirebaseTestConfig {
   if (offline) { // ** Connect to emulator
-    const firebaseTest: any = firebaseFunctionsTest();
-
+    let firebaseTest:any = firebaseFunctionsTest();
```
phpgeekfr, line 26: use const or the linter will yell. And add some formatting

xmano: Fixed.

PR #3436, opened 2020-08-12
Feat 3066 debug part 1

phpgeekfr:


apps/backend-ops/src/assets/users.fixture.ts
```diff
@@ -1,10 +1,5 @@
-export interface UserConfig {
-  uid: string;
-  email: string;
-  password: string;
-
-  [key: string]: any;
-}
+import { UserConfig } from "@blockframes/firebase-utils";
+export { UserConfig };
```
xmano, line 2: `export  { UserConfig } from "@blockframes/firebase-utils";   `

wont work?

apps/backend-ops/src/users.ts
```diff
@@ -39,38 +38,6 @@ async function createAllUsers(users: UserConfig[], auth: Auth): Promise<any> {
   return Promise.all(ps);
 }
 
-/**
```
xmano, line 42: I think moving this to utils wont help much in reuse so maybe leave it here?

xmano: Major issue is with `upsertWatermark`

Issue #3175, opened 2020-06-24
Issues regarding production database and storage


Jackseed:
- [x] 1. Delete old notification & invitations (<14 days) except pending invitation to become a member of an organization.
- [x] 2. Delete notification & invitation from doc that doesn't exist anymore.
- [x] 3. Delete the last image ref artefacts

Checkout #3047

GrandSchtroumpf:
Production database and storage are a mess, and a cleaning is needed.

Todo/ To check : 

- [ ] DistributionRight object in movie root instead of sub-collection
- [ ] 89 files with name too long in storage
- [ ] check the `movie` dir (which should now be `movies`)
  this is caused by the `presentation_deck` and other pdf files that are not uploaded to the good directory, we should merge them back into the good `movies` folder
- [ ] watermarks are stored in the root at `watermarks/` _(with an `s`)_ where they should be not in the root but in `users/<ID>/watermark/` _(without an `s`_), so we need to move them
  *Alternatively we can also just delete them all, and re generate them with the updated backend function, it shoudl create them in the good folder.*
- [ ] keep only the used images: some folders can have 2 to 5 images 
- [ ] check if number of user in `Authentication` is the same as in db
- [ ] at the end of the cleaning : delete every folders/files that are not in the good place/doesn't have the good name

phpgeekfr:
- [ ] check if a permission cleaning is needed

phpgeekfr:
- [ ] docsIndex collection with undefined document in it

phpgeekfr:
- [ ] create a migration script for remaining files that are correctly linked in DB but not in the good folder.
- [ ] update front so the resources are saved in the right directory

Issue #3066, opened 2020-06-14
Test & reactivate clean script


GrandSchtroumpf:
- [x] unit-test the script : `apps/backend-ops/src/db-cleaning.ts`
- [x] reactivate it in `apps/backend-ops/src/firebaseSetup.ts`
- [x] unit-test the script : `apps/backend-ops/src/storage-cleaning.ts`
- [x] reactivate it in `apps/backend-ops/src/firebaseSetup.ts`

PR #3437, opened 2020-08-12
Feat 3066 part2 - to the road of good unit tests

phpgeekfr:


apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -13,30 +12,38 @@ import {
...
+import { removeUnexpectedUsers } from './users';
+import { UserConfig } from './assets/users.fixture';
 
 type Snapshot = FirebaseFirestore.QuerySnapshot<FirebaseFirestore.DocumentData>
 jest.setTimeout(30000);
 
-describe('DB cleaning script', () => {
+
+describe('Unit testing examples', () => {
+  it('should instruct you how to test expected errors', async () => {
```
GrandSchtroumpf, line 31: I don't see the point of this test. It's not related to the db cleaning script. It's more a general knowledge of how an Error looks like.

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -13,30 +12,38 @@ import {
...
-    const featList = initFunctionsTestMock();
-    adminServices = { // @TODO #3066 Flatten everything
-      auth: auth(),
-      db: firestore(),
-      storage: storage(),
-      firebaseConfig: featList.firebaseConfig
-    };
-
+    initFunctionsTestMock();
+    adminServices = loadAdminServices();
```
GrandSchtroumpf, line 46: So you decided to carry on with the adminServices ?

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -140,14 +144,143 @@ describe('DB cleaning script', () => {
...
+    // Check if data have been correctly added
+    const usersBefore = await adminServices.db.collection('users').get();
+    expect(usersBefore.docs.length).toEqual(10);
+
+    // Add a new auth user that is not on db
+    const fakeAuthUser = { uid: 'abc123', email: 'test@abc123.com' };
+    adminAuth.users.push(fakeAuthUser);
+
+    const authBefore = await adminAuth.listUsers();
+    expect(authBefore.users.length).toEqual(3);
```
GrandSchtroumpf, line 185: Looks like another test. If this expect has nothing to do with the previous put it in another "it" test.

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -140,14 +144,143 @@ describe('DB cleaning script', () => {
...
+    const user = await adminAuth.getUserByEmail(fakeAuthUser.email);
+    expect(user).toEqual(undefined);
+
+  });
+
+  it('should smoothly delete an user that is not in auth', async () => {
+    const adminAuth = new AdminAuthMocked() as any;
+
+    // User with no orgId
+    const testUser1 = { uid: '1M9DUDBATqayXXaXMYThZGtE9up1', email: 'bdelorme+users-392@cascade8.com' };
```
GrandSchtroumpf, line 203: It would be easier to read if uid are "1" or "A". Something easy to read.
I would recommand to name your email "johndoe@fake.com" or something like that.

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -184,15 +317,43 @@ function isNotificationClean(d: any) {
...
+    const docRef = adminServices.db.collection(collection).doc(d.id || d.uid);
+    if (d.date?._seconds) { d.date = new Date(d.date._seconds * 1000) };
+    await docRef.set(d);
+  }, 50, false)
+}
+
+async function resetDb(adminServices: AdminServices) {
+  const collections = ['movies', 'orgs', 'permissions', 'docsIndex', 'notifications', 'events', 'users', 'invitations'];
+
+  let documents = [];
```
GrandSchtroumpf, line 353: You can use const. It's an array.

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -184,15 +317,43 @@ function isNotificationClean(d: any) {
...
+  for (const collection of collections) {
+    const docs = await adminServices.db.collection(collection).get();
+    documents = documents.concat(docs.docs);
+  }
+  return runChunks(documents, async d => await d.ref.delete(), 50, false);
```
GrandSchtroumpf, line 358: You don't need neither async nor await.

apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -184,15 +317,43 @@ function isNotificationClean(d: any) {
...
+    await docRef.set(d);
+  }, 50, false)
+}
+
+async function resetDb(adminServices: AdminServices) {
+  const collections = ['movies', 'orgs', 'permissions', 'docsIndex', 'notifications', 'events', 'users', 'invitations'];
+
+  let documents = [];
+  for (const collection of collections) {
+    const docs = await adminServices.db.collection(collection).get();
```
GrandSchtroumpf, line 355: Here you are awaiting for the collection after each iteration. Don't you want to use a Promise.all instead ?

apps/backend-ops/src/db-cleaning.ts
```diff
@@ -74,34 +74,32 @@ export async function cleanDeprecatedData(adminServices: AdminServices) {
...
     } else {
-      await _cleanNotification(db, doc, notification);
+      await _cleanNotification(doc, notification);
     }
   });
 
 }
 
-async function _cleanNotification(db: Firestore, doc: any, notification: any) { // @TODO (#3175 #3066) w8 "final" doc structure
+async function _cleanNotification(doc: any, notification: any) { // @TODO (#3175 #3066) w8 "final" doc structure
```
GrandSchtroumpf, line 102: Bettr to change name. For example : 
- cleanAllNotifications
- cleanOneNotification

Also could ou type the arguments ?

apps/backend-ops/src/firestoreMigrations/0029.ts
```diff
@@ -23,21 +22,21 @@ import { OldImgRef } from './old-types';
  */
 export async function upgrade(db: Firestore) {
   try {
-    const users = await getCollection<PublicUser>('users');
+    const users: PublicUser[] = await db.collection('users').get().then(ref => ref.docs.map(d => d.data() as PublicUser));
```
GrandSchtroumpf, line 25: Why removing `getCollection`. I've the feeling that it's the same method.

libs/testing/src/lib/firebase/adminAuthMocked.ts
```diff
@@ -11,19 +11,23 @@ export class AdminAuthMocked {
...
+        pageToken
       }
       resolve(out);
     });
   }
 
   deleteUser(uid: string) {
     return new Promise((resolve) => {
-      resolve(true);
+      const user = this.users.some(u => u.uid === uid);
```
GrandSchtroumpf, line 26: Rename variable to `hasUser` as this is a boolean

GrandSchtroumpf: I think this is not needed to verify this though. You can directly remove the user with a filter no ?

libs/testing/src/lib/mocked-data-unit-tests/auth.json
```diff
@@ -1,26 +1,10 @@
 [
```
GrandSchtroumpf, line 1: Could you name for folder "mock" instead of "mocked-data-unit-tests" ? ^^
Or better "fixtures"

libs/testing/src/lib/scripts/create-mocked-data-db-cleaning.ts
```diff
@@ -142,6 +142,68 @@ function cleanDocument(doc: any, collection: string) {
   return doc;
 }
 
+function cleanOrg(org: any) {
```
GrandSchtroumpf, line 145: What's the goal of this function ? Once you need to clean fixtures it feels that you go too far in the unit test...
Fixtures should be plug&play.

GrandSchtroumpf: Ok this is a createOrg, then why not just using the one we've got already and passing default data ?

apps/backend-functions/src/users.ts
```diff
@@ -1,12 +1,12 @@
 import * as admin from 'firebase-admin';
 import * as functions from 'firebase-functions';
-import { db } from './internals/firebase';
+import { db, getStorageBucketName } from './internals/firebase';
```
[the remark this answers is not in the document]
phpgeekfr: It is used in backend-functions from the beginning. And it just return `bucketStorage` as you said. I had to get rid of it into backend-ops. Do you want me to remove it also for backend-functions @GrandSchtroumpf

apps/backend-ops/src/db-cleaning.spec.ts
phpgeekfr: Yep. I will put it in an "example-test.js"

apps/backend-ops/src/firestoreMigrations/0029.ts
phpgeekfr: Because this method is defined in `backend-functions`

GrandSchtroumpf: I think `getDocument` is in firebase-utils now. Maybe you can move `getCollection` as well ;)

apps/backend-functions/src/users.ts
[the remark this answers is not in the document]
GrandSchtroumpf: I would just not use it. And mabe put a @deprecated above if it doesn't break the build.

apps/backend-ops/src/db-cleaning.spec.ts
phpgeekfr: nop, cannot: 
`documents = documents.concat(docs.docs);`

apps/backend-ops/src/db-cleaning.ts
phpgeekfr: I made some renaming.
concerning the typing see my comment `@TODO (#3175 #3066) w8 "final" doc structure`

libs/testing/src/lib/scripts/create-mocked-data-db-cleaning.ts
phpgeekfr: In fact, I think I will get rid of this entire file ( like how I did it the third describe)

libs/testing/src/lib/mocked-data-unit-tests/auth.json
phpgeekfr: In fact, I think I will get rid of this entire directory ( like how I did it the third describe)

apps/backend-ops/src/firestoreMigrations/0029.ts
phpgeekfr: do you want me to try moving it in the libs @GrandSchtroumpf  ?

apps/backend-ops/src/db-cleaning.spec.ts
phpgeekfr: The two first `expect` are here to check if data is ready/ well formed before making the actual test.

apps/backend-functions/src/users.ts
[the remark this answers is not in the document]
phpgeekfr: OK added a deprecated comment.

apps/backend-ops/src/firestoreMigrations/0029.ts
GrandSchtroumpf: yes, I think it won't be that difficult. You can reexport it like you did with the `getDocument`. It shouldn't be that difficult I think.

PR #3438, opened 2020-08-12
Scroll bar for too wide filters

RemcoSimonides:
Quick UI fix for overflowing text in filters for movies


libs/movie/src/lib/movie/components/title-filter/title-filter.component.scss
```diff
@@ -1,14 +1,23 @@
...
-  overflow: hidden;
+  width: 100%;
+}
+
+section {
+  overflow: auto;
+  white-space: nowrap;
+  padding-bottom: 24px;
 
   .direction {
```
GrandSchtroumpf, line 15: You can remove that alongside the buttons. Same with right & left

Issue #3412, opened 2020-07-30
Debug release 1.8


Jackseed:
## Dashboard homepage
- [x] Update the message when the user has no movie, there are 2 cases:
1. Case when the org has really no movie on the festival app (they could have some on another app). To check that, search for `storeConfig.appAccess.festival: true` inside the movies included into org `movieIds`. If there is none, then display the same message as the screenshot (= don't change a thing):
![image](https://user-images.githubusercontent.com/14964443/88951428-3e472a00-d296-11ea-8036-b942b4ba02f5.png)
2. Case when the org has created draft movies on the festival app, but none are validated. To check that, search for `storeConfig.appAccess.festival: true` inside the movies included into org `movieIds` and then verify `storeConfig.status`, if there is at least 1 film with `appAccess: festival` to `true` but none of them have a `storeConfig.status` to `accepted`, then change the message for:
Title: "You have no title on the marketplace yet."
Subtitle: "As long as you have no title to display, your organization will not appear on the marketplace. To import titles, you can either add several titles at once through bulk import, or just one at a time by filling in the platform's form. N.B.: Only submitted projects will appear on the marketplace, not drafts. "
**adding a line break after each sentence on subtitle**

## Marketplace homepage
- [x] **See comment from RemcoSimonides**. The section "Featured Sales Agents", don't display ALL the sales agents, use the same query used in `/c/o/marketplace/organization`, i.e. sales agents that have at least 1 film published on the marketplace:
- org with `appAccess.festival.dashboard: true`
&& 
- movieIds including at least 1 movie with `storeConfig.appAccess.festival: true` && `storeConfig.status: accepted`
![image](https://user-images.githubusercontent.com/14964443/88957223-d9440200-d29e-11ea-87e6-fcca3f4333d7.png)

- [x] @RemcoSimonides There is a missing background image to add:
![image](https://user-images.githubusercontent.com/14964443/88954305-78b2c600-d29a-11ea-9d5c-cbf0ade44102.png)
vs [Invision](https://projects.invisionapp.com/d/main?origin=v7#/console/19607961/413926229/preview?scrollOffset=4187.77783203125)
![image](https://user-images.githubusercontent.com/14964443/88954286-6cc70400-d29a-11ea-8860-e2c884aa000c.png)
**/!\ Beware Safari, don't forget the fallback** 

### Featured seller component
- [x] Add the right padding/margin between logo and org name + org films:
[...]

PR #3440, opened 2020-08-12
Count only includes accepted movies now

### LINKED ISSUE #3308: Fix accepted movie count of organization
Currently the organization banner component counts the number of movieIds of the organization. This also inlcudes movies which are not (yet) accepted. This number needs to be the total number of accepted movies.

This can be done in two ways;
1 with a pipe:
component.html:
`<ng-container *ngIf="org | orgMovie | async | filterBy:filterMovie">`

component.ts:
`filterMovie = (movie: Movie) => !!movie.status === 'accepted';`

pipe.ts:
```
export FilterByPipe<T> {
  transform<T>(items: T[], filter: (item: T) => boolean = (i) => true) {
    return items.filter(filter);
  }
}
```

Or
~~2 with data aggregation in Firestore
Every time a movie is accepted or an accepted movie is deleted, then an aggregated number is increased or decreased.~~

RemcoSimonides:
Remove movies that have not yet been accepted or not available for festival from movie count

fix #3308 && todo in #3412

### DIFF
--- a/libs/organization/src/lib/organization/components/banner/banner.component.ts
+++ b/libs/organization/src/lib/organization/components/banner/banner.component.ts
@@ -1,5 +1,6 @@
 import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
 import { Organization } from '../../+state';
+import { Movie } from '@blockframes/movie/+state/movie.model';
 
 @Component({
   selector: '[org] org-banner',
@@ -11,4 +12,5 @@ export class OrganizationBannerComponent {
   
   @Input() org: Organization;
 
+  filterMovie = (movie: Movie) => movie.main.storeConfig.status === 'accepted' && movie.main.storeConfig.appAccess.festival;
 }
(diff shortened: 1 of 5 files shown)


libs/organization/src/lib/organization/components/banner/banner.component.ts
```diff
@@ -11,4 +12,5 @@ export class OrganizationBannerComponent {
   
   @Input() org: Organization;
 
+  filterMovie = (movie: Movie) => movie.main.storeConfig.status === 'accepted' && movie.main.storeConfig.appAccess.festival;
```
GrandSchtroumpf, line 15: @Mussuranae is it going to be an issue with your migration ?

libs/utils/src/lib/pipes/filter-by.pipe.ts
```diff
@@ -0,0 +1,15 @@
+import { Pipe, PipeTransform, NgModule } from '@angular/core';
+
+@Pipe({ name: 'filterBy' })
+export class FilterBy implements PipeTransform {
+  transform<T>(items: T[], filter: (item: T) => boolean = (i) => true) {
+    if (!items) return;
```
GrandSchtroumpf, line 6:
```typescript
return Array.isArray(items) ? items.filter(filter) : [];
```
Let that you don't need the "?" in ".?length" above.

PR #3445, opened 2020-08-13
Upload restore-ci to blockframes-ci

GrandSchtroumpf:
create upload-db
update package.json
update which certificates is used
use static names in tmp folder


package.json
```diff
@@ -55,8 +55,9 @@
     "storybook:ui": "ng run ui:storybook",
     "storybook:movie": "ng run movie:storybook",
     "prerender:festival": "ng build festival --prod && ng run festival:server --configuration production && angular-prerender --browser-target festival:build --server-target festival:server",
-    "anonymize-db": "npx ts-node -P tools/tsconfig.tools.json tools/scripts/anon-firestore.ts",
-    "get-latest-prod-db": "npx ts-node -P tools/tsconfig.tools.json tools/scripts/get-prod-backup.ts"
+    "get-latest-prod-db": "npx ts-node -P tools/tsconfig.tools.json tools/scripts/get-prod-backup.ts",
```
george43g, line 58: just an idea, since these commands are related, we could start using a prefix to group them (helps with tab completion too)
```
"db:anonymize"
"db:get-latest-prod"
"db:upload"
```

tools/scripts/get-prod-backup.ts
```diff
@@ -5,10 +5,14 @@ import * as admin from 'firebase-admin';
...
 
+if (!('FIREBASE_PRODUCTION_SERVICE_ACCOUNT' in process.env)) {
+  throw new Error('Key "FIREBASE_PRODUCTION_SERVICE_ACCOUNT" does not exist in .env');
+}
+
 admin.initializeApp({
   storageBucket: backupBucket,
   projectId: firebase.projectId,
-  credential: admin.credential.cert(process.env.GOOGLE_APPLICATION_CREDENTIALS)
+  credential: admin.credential.cert(process.env.FIREBASE_PRODUCTION_SERVICE_ACCOUNT)
```
george43g, line 15: In this case, the key is stored as an actual env variable - plaintext.
This currently works because it receives a relative path to the key file
So this will crash. The variable needs to be `JSON.parse()` and then passed as an object - I'll have to double check how parsed service account keys are used

george43g: because in CI, it's the plaintex while on local dev machine that var was just a rel path

tools/scripts/upload-db.ts
```diff
@@ -0,0 +1,39 @@
...
+config();
+
+if (!('FIREBASE_CI_SERVICE_ACCOUNT' in process.env)) {
+  throw new Error('Key "FIREBASE_CI_SERVICE_ACCOUNT" does not exist in .env');
+}
+
+admin.initializeApp({
+  storageBucket: backupBucket,
+  projectId: firebase.projectId,
+  credential: admin.credential.cert(process.env.FIREBASE_CI_SERVICE_ACCOUNT)
```
george43g, line 15: same here - parse plaintext JSON obj

george43g: OR, I think it get's written to disk at a specified path - but I think there's a way to utilise the parsed object as well

tools/scripts/upload-db.ts
```diff
@@ -0,0 +1,39 @@
...
+
+async function uploadDB() {
+  try {
+    // Ensure parent folder exist
+    if(!existsSync(restore)) {
+      throw new Error(`File ${restore} doesn't exist.`);
+    }
+    const destination = `${new Date().toISOString()}-ANONYMIZED.jsonl`;
+    await storage.bucket(backupBucket).upload(restore, { destination });
+    console.log(`Restore DB has been saved to: ${backupBucket}/${destination}`);
```
george43g, line 28: `console.log(`Restore DB has been saved to: gs://${backupBucket}/${destination}`);` ??

tools/scripts/upload-db.ts
george43g: "Initialize without parameters
The SDK can also be initialized with no parameters. In this case, the SDK uses Google Application Default Credentials and reads options from the FIREBASE_CONFIG environment variable. If the content of the FIREBASE_CONFIG variable begins with a { it will be parsed as a JSON object. Otherwise the SDK assumes that the string is the name of a JSON file containing the options."

tools/scripts/get-prod-backup.ts
george43g: So if you just copy it to "FIREBASE_CONFIG" , and then init without a config it should pick it up, I'm pretty sure - see comment below

PR #3457, opened 2020-08-14
App summary page

joanofdart:
Summary Page :)


libs/movie/src/lib/movie/pages/summary/summary.component.ts
```diff
@@ -31,9 +32,19 @@ export class MovieFormSummaryComponent {
     private routerQuery: RouterQuery
   ) { }
 
-  public getPath(segment: string) {
+  public get genres() {
+    return [this.form.get('genres'), ...this.form.get('customGenres').controls];
```
GrandSchtroumpf, line 36: Is it not `return [this.form.get('genres').controls, ...this.form.get('customGenres').controls]` ?

Issue #3355, opened 2020-07-22
App Bar Search


Hugo-Hontans:
Talking about the App Bar Search inside the top nav bar of catalog/marketplace/title-list and festival/marketplace/title-list components.
This bar appears when we scroll to bottom and replace the classic search bar.
We got issues:
  - When the user does a search without result, he is stucked at the top of the page without the App Bar Search. So he cannot start a new search, he is blocked and must refresh the page.
  - If the result of the search if for example only one movie and the user scroll to top, the App Bar Search disappears and he doesn't understand how start a new search.

This 2 bugs can be solved by linked the 2 search bars with the same value. Like this the user is never lost.
Can be good t do a point on the expected UX with @2amarkina and @Jackseed before.

Jackseed:
Best would be to 'connect' the 2 bars so they have the same text on both. 
Check with @2amarkina the wanted behavior and if too complicated remove the search bar on top

PR #3461, opened 2020-08-17
Added delay to prevent the bug from happening

RemcoSimonides:
Fix to do from issue #3412 

"On top of #3355, top search bar that should appear when you scroll down doesn't appear if you do this flow:
1. go on all films
2. go on a public org page (e.g. `c/o/marketplace/organization/GdFUevWIifjk6VEYgfCT/title`)
3. go back on the page all films. Don't know why but the 2 look linked..."

Actually this bug happens on all pages which use the pageBar directive. This is caused by the values in `entry.rootBounds` to be all 0. After a small delay the `entry.rootBounds` do have the values they should have.

I expected animations to be the cause of this issue, but even when disabling animations this bug still happens...

### DIFF
--- a/libs/ui/src/lib/app-bar.ts
+++ b/libs/ui/src/lib/app-bar.ts
@@ -5,7 +5,6 @@ import { trigger, style, transition, animate, query } from '@angular/animations'
 import { Easing } from '@blockframes/utils/animations/animation-easing';
 import { MatButtonModule } from '@angular/material/button';
 import { MatIconModule } from '@angular/material/icon';
-import { delay } from '@blockframes/utils/helpers';
 
 @Component({
   selector: 'app-bar',
@@ -79,11 +78,11 @@ export class AppContainerDirective {
   constructor(
     @Inject(DOCUMENT) private document: Document,
     ref: ElementRef,
-    private zone: NgZone
+    zone: NgZone
   ) {
     this.container = ref.nativeElement;
 
-    this.zone.runOutsideAngular(() => {
+    zone.runOutsideAngular(() => {
       const heightSize = 80;
       const options = {
         root: this.container,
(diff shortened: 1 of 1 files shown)


libs/ui/src/lib/app-bar.ts
```diff
@@ -93,7 +92,8 @@ export class PageBarDirective implements AfterViewInit, OnDestroy {
     private zone: NgZone,
   ) {}
 
-  ngAfterViewInit() {
+  async ngAfterViewInit() {
+    await delay(500); // prevents sizes in entry.rootBounds to not be 0
```
pldespaigne, line 96: I don't know what was the issue, but I think this is not a really good solution.
Why 500ms ? How can you be sure that after 500ms it will be ok ?

To me it looks like you just prevent the symptoms without treating the underlying problem.

GrandSchtroumpf, line 96: You should never use a delay to solve a problem.

RemcoSimonides: I am aware of this ;) it is the only thing that did something and I think knowing a delay does work will help us to find the source of the issue. Let's look at it together tomorrow, ok?

RemcoSimonides: Yea, this isn't the definite solution ;)

libs/ui/src/lib/app-bar.ts
```diff
@@ -3,9 +3,9 @@ import { PortalModule, TemplatePortal } from '@angular/cdk/portal';
 import { CommonModule, DOCUMENT } from '@angular/common';
 import { trigger, style, transition, animate, query } from '@angular/animations';
 import { Easing } from '@blockframes/utils/animations/animation-easing';
-import { Observable } from 'rxjs';
 import { MatButtonModule } from '@angular/material/button';
 import { MatIconModule } from '@angular/material/icon';
+import { delay } from '@blockframes/utils/helpers';
```
GrandSchtroumpf, line 8: remove it please

libs/ui/src/lib/app-bar.ts
```diff
@@ -72,53 +72,73 @@ export class AppBarComponent {
...
 export class AppContainerDirective {
+  private observer: IntersectionObserver;
   container: HTMLElement;
   @Input('appContainer') appBar: AppBarComponent;
   @Output() toggle = new EventEmitter();
-  constructor(ref: ElementRef) {
+  constructor(
+    @Inject(DOCUMENT) private document: Document,
+    ref: ElementRef,
+    private zone: NgZone
```
GrandSchtroumpf, line 82: You can remove private here

GrandSchtroumpf (approved):

PR #3462, opened 2020-08-17
Replaced webp by png image

RemcoSimonides:
Fix to do in issue #3412 

"The top banner on movie list page looks like missing a fallback, it's blank on Safari"


libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -1,7 +1,7 @@
 header#nav {
   padding: 48px;
   padding-bottom: 0px; // Remove padding for filters
-  background-image: url('/assets/images/banner_header_page.webp');
+  background-image: url('/assets/images/banner_header_page.png');
```
pldespaigne, line 4: this will prevent webp images for every browser.

remove `background-image` from the css, and use `/media/directives/image-reference/background-reference.directive.ts` instead

libs/movie/src/lib/movie/layout/list/title-list.component.scss
```diff
@@ -29,7 +29,7 @@ header.pageBar {
   position: sticky;
   top: 0;
   z-index: 2;
-  background-image: url('/assets/images/banner_header_page.webp');
+  background-image: url('/assets/images/banner_header_page.png');
```
pldespaigne, line 32: same

PR #3466, opened 2020-08-17
feat(fetch-db-from-ci.ts): add script to fetch db from ci and upload …

george43g:
…to local env
- write script
- add calling script to package.json

### DIFF
--- a/tools/scripts/fetch-db-from-ci.ts
+++ b/tools/scripts/fetch-db-from-ci.ts
@@ -1,0 +1,80 @@
+import { join } from 'path';
+import { firebase, backupBucket } from '../../env/env';
+import { backupBucket as backupBucketCI, firebase as firebaseCI } from '../../env/env.ci';
+import * as admin from 'firebase-admin';
+import { config } from 'dotenv';
+import { existsSync, mkdirSync } from 'fs';
+config();
+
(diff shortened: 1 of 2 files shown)


tools/scripts/fetch-db-from-ci.ts
```diff
@@ -0,0 +1,80 @@
...
+    console.log('Bucket name: ', last.bucket.name);
+
+    // Ensure parent folder exist
+    if (!existsSync(folder)) {
+      mkdirSync(folder);
+    }
+
+    // Dowload lastest backup
+    const destination = join(folder, last.name);
+    await last.download({ destination });
```
GrandSchtroumpf, line 63: Wouldn't it be easier to use `last.copy(backupBucket)` directly ? As it's supposed to files across storages.

george43g: Of course - I tried this initially. However there was a permissions issue. So one app needs to have permissions on both buckets - I think copy is for buckets within the same project. This seemed to work while `copy()` did not.

GrandSchtroumpf: You need the permission for both app anyway no ?
Anyway you can merge.

PR #3467, opened 2020-08-18
Ellipses when overflowing text and fixed invitation-action component …

RemcoSimonides:
Fixed a to do in #3412 

Instead of using ellipses for more than 100 characters as described in the todo, I use ellipses for all overflowing text.
image

### DIFF
--- a/libs/invitation/src/lib/components/item/item.component.html
+++ b/libs/invitation/src/lib/components/item/item.component.html
@@ -5,30 +5,30 @@
     alt="Organization logo">
   <div matLine>
     <ng-container [ngSwitch]="invitation.type">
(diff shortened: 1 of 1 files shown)


libs/invitation/src/lib/components/item/item.component.html
```diff
@@ -5,38 +5,38 @@
     alt="Organization logo">
   <div matLine>
     <ng-container [ngSwitch]="invitation.type">
-      <div *ngSwitchCase="'attendEvent'">
+      <span *ngSwitchCase="'attendEvent'">
```
fritzschoff, line 8: can't you just use a ng-container here, then you don't produce a html tag in the DOM?

PR #3483, opened 2020-08-19
Feat 3066 part3

phpgeekfr:


libs/testing/src/lib/firebase/functions.ts
```diff
@@ -45,3 +45,17 @@ export function getTestingProjectId() {
...
+
+////////////
+// DB TOOLS
+////////////
+
+export function populate(collection: string, set: any[]) {
+  const db = admin.firestore();
+  return runChunks(set, async (d) => {
+    const docRef = db.collection(collection).doc(d.id || d.uid);
+    if (d.date?._seconds) { d.date = new Date(d.date._seconds * 1000) };
```
GrandSchtroumpf, line 57: There is a function `reEncodeObject` in "backend-function/backup.ts" to do reencode the whole object (so you don't only support the key "date").

PR #3489, opened 2020-08-19
refactor(test-workflow.ts): move fetch-db-from-ci to firebase-utils a…

george43g:
…nd fix tsconfig


libs/firebase-utils/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './lib/firebase-utils';
 export * from './lib/maintenance';
-export * from './lib/watermark';
\ No newline at end of file
+export * from './lib/watermark';
+export * from './lib/test-workflow';
```
GrandSchtroumpf, line 4: I'm not sure about the name. What about "sync-with-ci" ?

libs/firebase-utils/src/lib/test-workflow.ts
```diff
@@ -0,0 +1,79 @@
...
+      credential: admin.credential.cert(cert),
+    },
+    'CI-app'
+  );
+
+  const app = admin.initializeApp({
+    storageBucket: backupBucket,
+    projectId: firebase.projectId,
+    credential: admin.credential.cert(process.env.GOOGLE_APPLICATION_CREDENTIALS),
+  });
```
GrandSchtroumpf, line 39: Might be a good idea to provide a name here too.

libs/firebase-utils/src/lib/test-workflow.ts
```diff
@@ -0,0 +1,79 @@
...
+  let cert: string | admin.ServiceAccount;
+  try {
+    // If service account is a stringified json object
+    cert = JSON.parse(process.env.FIREBASE_CI_SERVICE_ACCOUNT);
+  } catch (err) {
+    // If service account is a path
+    cert = process.env.FIREBASE_CI_SERVICE_ACCOUNT;
+  }
+
+  const ci = admin.initializeApp(
```
GrandSchtroumpf, line 26: It's not urgent, but I'm concerne about initializing apps in the script directly. I've got the feeling that it'll create issues when we glue everything together or for unit tests.

tools/users.fixture.json
```diff
@@ -0,0 +1 @@
+[{"email":"George -Gramstr-fake@cascade8.com","uid":"Tojd0jMtfgOZ6YzH55xFmJJvUhk2","password":"blockframes"}]
```
GrandSchtroumpf, line 1: Should we gitignore this file ?

GrandSchtroumpf (approved):

libs/firebase-utils/src/index.ts
george43g: i thought this file would export functions to do with test workflow. but if it's just for that one function (which loses some semantics via lack of context) then we can name it something like "copy-db-from-ci"

GrandSchtroumpf: yes, let's keep one file = one function for now.

libs/firebase-utils/src/lib/test-workflow.ts
george43g: It's only doing it this way because it's connecting to two different projects with different sets of credentials. Otherwise we'd have to build the ability to connect to CI into the lib

tools/users.fixture.json
george43g: woops - this was added by accident

george43g: yes

PR #3490, opened 2020-08-19
Get accepted movies without resetting the store

RemcoSimonides:
Fix to do in issue #3412 

"When you go from org public page on marketplace (e.g. /c/o/marketplace/organization/e1VXeusNJK6pb8kmVnUn/title) and then on wishlist only the films from that sales agents + included in your wishlist are visible."

### DIFF
--- a/apps/festival/festival/src/app/marketplace/organization/title/title.component.ts
+++ b/apps/festival/festival/src/app/marketplace/organization/title/title.component.ts
@@ -15,7 +15,7 @@ import { DynamicTitleService } from '@blockframes/utils/dynamic-title/dynamic-ti
 })
 export class TitleComponent implements OnInit {
   @HostBinding('@scaleIn') private animePage;
-  public movies$: Observable<Movie[]>;
+  public titles$: Observable<Movie[]>;
 
   constructor(
     private service: MovieService,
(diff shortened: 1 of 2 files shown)


apps/festival/festival/src/app/marketplace/organization/title/title.component.ts
```diff
@@ -13,30 +13,23 @@ import { DynamicTitleService } from '@blockframes/utils/dynamic-title/dynamic-ti
...
-  ngOnInit(): void {
+  ngOnInit() {
     this.dynTitle.setPageTitle('Sales Agent', 'Line-up');
-    this.sub = this.parent.org$.pipe(
+    this.movies$ = this.parent.org$.pipe(
       map(org => org.movieIds),
-      tap(_ => this.store.reset()),
-      switchMap(movieIds => this.service.syncManyDocs(movieIds))
-    ).subscribe();
+      switchMap(movieIds => this.service.valueChanges(movieIds)),
```
GrandSchtroumpf, line 30: Add a TODO: "update with a query when movie model is updated with the orgs inside".
Basically here we need to get everything and then we filter. Best would be to query it: 
```typescript
ref => ref.where('orgs', 'array-contains', orgId).where('main.storeConfig.status', '==', 'accepted').where(...)
```
But the movie model is not right for that yet.

apps/festival/festival/src/app/marketplace/organization/title/title.component.ts
```diff
@@ -13,30 +13,23 @@ import { DynamicTitleService } from '@blockframes/utils/dynamic-title/dynamic-ti
...
-    private store: MovieStore,
     private parent: ViewComponent,
     private dynTitle: DynamicTitleService,
   ) { }
 
-  ngOnInit(): void {
+  ngOnInit() {
     this.dynTitle.setPageTitle('Sales Agent', 'Line-up');
-    this.sub = this.parent.org$.pipe(
+    this.movies$ = this.parent.org$.pipe(
```
GrandSchtroumpf, line 28: Keep the name "titles$"

GrandSchtroumpf (approved):

PR #3492, opened 2020-08-19
Validators

Mussuranae:
Rework some input that don't appear in red when error


libs/movie/src/lib/movie/pages/main/main.component.ts
```diff
@@ -26,6 +28,23 @@ export class MovieFormMainComponent {
...
+  }
+
+  ngOnDestroy() {
+    this.sub.unsubscribe();
+  }
+
+  focusOut() {
+    return this.form.get('runningTime').get('time').setErrors({required: true});
+  }
+
```
pldespaigne, line 47: this is not really clear to use valueChange on the hole form just for one field and to call `focusOut`

pldespaigne (commented):
I think there is a deeper problem here, because unlike genre & territories, this is just a simple input field, so it should be the regular way of validating.

Also why do you ask me for review if you merge anyway ^^

PR #3507, opened 2020-08-20
George/solve user sync

george43g:
- had to disable some steps in prepareForTesting as they were not working
- get db from CI
- sync users to that db
- remove references to fixture file in backend-ops
- some small fixes to ensure backend-ops can run without compilation


apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -45,13 +51,21 @@ export async function prepareForTesting() {
...
-  await syncStorage();
-  console.info('Firestore is now synced with storage!');
+  // console.info('Syncing firestore with storage');
+  // await syncStorage();
+  // console.info('Firestore is now synced with storage!');
 
   process.exit(0);
 }
 
+function getDbFromTmp(dbBackupPath: string) {
```
GrandSchtroumpf, line 61: Create a function `readJsonl(path: string)` in firebase-utils instead, so we can use it everywhere.

apps/backend-ops/src/users.ts
```diff
@@ -13,6 +12,17 @@ import { startMaintenance, endMaintenance, isInMaintenance } from '@blockframes/
 import { loadDBVersion } from './migrations';
 import { firebase } from '@env';
 export const { storageBucket } = firebase;
+
+export interface UserConfig {
+  uid: string;
+  email: string;
+  password: string;
```
GrandSchtroumpf, line 19: If this is always "blockframes" then type it like that : 
```typescript
export interface UserConfig {
  uid: string;
  email: string;
  password: 'blockframes';
  [key: string]: any;
}
```
Like that you can remove the `USER_FIXTURES_PASSWORD` below.

apps/backend-ops/src/users.ts
```diff
@@ -36,7 +46,13 @@ async function createUserIfItDoesntExists(auth: Auth, userConfig: UserConfig): P
  * @param auth  Firestore Admin Auth object
  */
 async function createAllUsers(users: UserConfig[], auth: Auth): Promise<any> {
-  const ps = users.map(user => createUserIfItDoesntExists(auth, user));
+  // ! Ensure there are no duplicates!
```
GrandSchtroumpf, line 49: Add a todo : "Remove this once anon script differenciate emails with same left part"
create an issue for that and put the id of the issue in the TODO comment.
We dont want to keep this in the code for long.

apps/backend-ops/src/users.ts
```diff
@@ -64,17 +80,34 @@ export async function removeUnexpectedUsers(expectedUsers: UserConfig[], auth: A
     // This is "good enough", but do not reproduce in frontend / backend code.
     for (const user of usersToRemove) {
       console.log('removing user:', user.email, user.uid);
-      await auth.deleteUser(user.uid);
-      await sleep(100);
     }
+    await auth.deleteUsers(usersToRemove.map((user) => user.uid));
+    await sleep(100);
```
GrandSchtroumpf, line 85: What this point of sleep ?

apps/backend-ops/src/users.ts
```diff
@@ -64,17 +80,34 @@ export async function removeUnexpectedUsers(expectedUsers: UserConfig[], auth: A
...
-export async function syncUsers(): Promise<any> {
+export interface JsonlDbRecord {
+  docPath: string;
+  content: { [key: string]: any };
+}
+
+function readUsersFromDb(db: JsonlDbRecord[]): UserConfig[] {
+  return db
+    .filter((doc) => doc.docPath.includes('users/'))
+    .filter((userDoc) => 'email' in userDoc.content)
```
GrandSchtroumpf, line 99: If there are many thousand users you should do that in one loop instead of two. But it's not the case right now so it's fine.

apps/backend-ops/src/users.ts
```diff
@@ -155,7 +188,7 @@ export async function generateWatermarks() {
   const dbVersion = await loadDBVersion(db);
   // activate maintenance to prevent cloud functions to trigger
   let startedMaintenance = false;
-  if (!await isInMaintenance()) {
+  if (!(await isInMaintenance())) {
```
GrandSchtroumpf, line 191: Please disable your auto correction ^^

libs/firebase-utils/src/lib/copy-db-from-ci.ts
```diff
@@ -75,11 +75,13 @@ export async function copyDbFromCi() {
...
     const result = await myBucket.upload(destination, { destination: last.name });
     console.log('File uploaded as', result[0].name);
+    return destination;
   } catch (err) {
     if ('errors' in err) {
       err.errors.forEach((error: { message: any }) => console.error('ERROR:', error.message));
     } else {
       console.log(err);
     }
+    return null;
```
GrandSchtroumpf, line 85: Why returning null instead of leaving to undefined ?

GrandSchtroumpf (changes requested):

apps/backend-ops/src/users.ts
george43g: if there are more than 1000 users, they will be paginated. And you can only delete 1k at a time. So there will be a loop, it just adds a 100ms delay.

apps/backend-ops/src/users.ts
george43g: reminds me of CS days where we actually learned about algorithms lol

apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -25,10 +31,10 @@ export async function prepareForTesting() {
...
   console.info('Database ready for testing!');
 
-  console.info('Cleaning unused db data...');
-  const { db, auth } = loadAdminServices();
-  await cleanDeprecatedData(db, auth);
-  console.info('DB data clean and fresh!');
+  // console.info('Cleaning unused db data...');
+  // const { db, auth } = loadAdminServices();
+  // await cleanDeprecatedData(db, auth);
+  // console.info('DB data clean and fresh!');
```
phpgeekfr, line 37: I found Why you had to comment this. This is because of the anonymization process:

In prod we have: matlmayer@cascade8.com & matlmayer@logicalpictures.com
The ano email will be: matlmayer@fake.com for both ..

apps/backend-ops/src/users.ts
phpgeekfr: I think you should remove it now, and fix the anon script

PR #3509, opened 2020-08-20
Validators

Mussuranae:
Rework running time input

### DIFF
--- a/libs/movie/src/lib/movie/pages/main/main.component.ts
+++ b/libs/movie/src/lib/movie/pages/main/main.component.ts
@@ -29,20 +29,16 @@ export class MovieFormMainComponent implements OnInit, OnDestroy {
   constructor(private shell: MovieFormShellComponent, private route: ActivatedRoute) { }
 
   ngOnInit() {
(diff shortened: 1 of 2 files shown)


libs/movie/src/lib/movie/pages/main/main.component.ts
```diff
@@ -29,10 +29,10 @@ export class MovieFormMainComponent implements OnInit, OnDestroy {
...
 
   ngOnInit() {
-    this.sub = this.form.valueChanges.subscribe(() => {
-      console.log(this.form.valid);
-      if (!this.form.valid) {
-        this.focusOut();
+
+    this.sub = this.form.runningTime.get('status').valueChanges.subscribe(value => {
+      if (value === "confirmed") {
+        return this.form.get('runningTime').get('time').setErrors({required: true});
```
pldespaigne, line 35: this is almost perfect, but what happen if the user set some time, then select "confirmed" ? Will set the required error even if the time was already set ?

Mussuranae: Good catch ! Yes it set the input in error even if there as a time.

PR #3511, opened 2020-08-20
Fix restore timeout and execution errors

phpgeekfr:


libs/firebase-utils/src/lib/maintenance.ts
```diff
@@ -27,7 +27,7 @@ export async function endMaintenance() {
   });
 }
 
-export async function isInMaintenance() {
+export async function isInMaintenance(security = EIGHT_MINUTES_IN_MS) {
```
GrandSchtroumpf, line 30: Add a comment to explain what the security is doing.

apps/backend-ops/src/admin.ts
```diff
@@ -57,11 +59,34 @@ export async function restore(appURL: string) {
...
       } else if (response.statusCode < 200 || response.statusCode > 299) {
-        reject(`invalid status code on restore: ${response.statusCode}. Check your firebase functions logs (admin function)`)
+        /**
+         * @dev this is a hack to handle unattended responses from 
+         * express server (502). The nginx proxy on google sides seems to hang
+         * up after a few seconds. To be sure that the restore is ended,
+         * we wait for 300 seconds.
+         */
+        if (response.statusCode === 502) {
+          let attempts = 10;
```
GrandSchtroumpf, line 73: I'm fine with the while.
Personally I find `for (let i = 0; i < attempts; i++)` more readable.

PR #3535, opened 2020-08-25
Polyfill Intersection Observer

RemcoSimonides:
To do in issue #3412 

Intersection Observer is not supported for Safari version lower than 12.2. Therefore it needs to be polyfilled.

### DIFF
--- a/libs/ui/src/lib/app-bar.ts
+++ b/libs/ui/src/lib/app-bar.ts
@@ -82,13 +82,15 @@ export class AppContainerDirective {
   ) {
     this.container = ref.nativeElement;
 
-    zone.runOutsideAngular(() => {
+    zone.runOutsideAngular(async () => {
       const heightSize = 80;
       const options = {
         root: this.container,
         rootMargin: `-${heightSize}px 0px 0px 0px`,
         threshold: 0
       }
+
+      if (!('IntersectionObserver' in window)) await import('intersection-observer');
       this.observer = new IntersectionObserver(([entry]) => {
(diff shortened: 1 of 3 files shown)


apps/festival/festival/src/polyfills.ts
```diff
@@ -97,4 +97,9 @@ if (!(Promise as any).allSettled) {
           }))
       )
   );
-}
\ No newline at end of file
+}
+
+/***************************************************************************************************
+* Intersection Observer
```
GrandSchtroumpf, line 103: Add comment to explain in which browser / platform it doesn't work.

GrandSchtroumpf (approved):

libs/ui/src/lib/app-bar.ts
```diff
@@ -82,13 +82,15 @@ export class AppContainerDirective {
...
-    zone.runOutsideAngular(() => {
+    zone.runOutsideAngular(async () => {
       const heightSize = 80;
       const options = {
         root: this.container,
         rootMargin: `-${heightSize}px 0px 0px 0px`,
         threshold: 0
       }
+
+      if (!('IntersectionObserver' in window)) await import('intersection-observer'); // IntersectionObserver doesnt work in Safari older than 12.2
```
GrandSchtroumpf, line 93: Put the comment above:
```typescript
// IntersectionObserver doesnt work in Safari older than 12.2
if (!('IntersectionObserver' in window)) {
  await import('intersection-observer');
}
```

GrandSchtroumpf (approved):

PR #3538, opened 2020-08-25
Added movie-form-budget

joanofdart:


libs/movie/src/lib/movie/form/budget/budget.component.scss
```diff
@@ -0,0 +1,51 @@
...
+  mat-divider {
+    border-top-color: var(--primary);
+    border-top-width: 2px;
+    margin-right: 16px;
+    width: 100%;
+  }
+  span {
+    white-space: nowrap;
+  }
+}
```
pldespaigne: please clean the file when you copy/paste

libs/movie/src/lib/movie/form/budget/budget.component.ts
```diff
@@ -0,0 +1,19 @@
...
+import { MovieForm } from '../movie.form';
+
+@Component({
+  selector: 'movie-form-budget',
+  templateUrl: './budget.component.html',
+  styleUrls: ['./budget.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class MovieFormBudgetComponent {
+  @Input() form: MovieForm;
```
pldespaigne, line 15: input should be a `FormControl` so that in the parent component we do `<movie-form-budget [form]="form.estimatedBudget">`

libs/movie/src/lib/movie/form/budget/budget.module.ts
```diff
@@ -0,0 +1,32 @@
...
+// Material
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { MatFormFieldModule } from '@angular/material/form-field';
+import { MatSelectModule } from '@angular/material/select';
+
+@NgModule({
+  imports: [
+    CommonModule,
+    FlexLayoutModule,
+    QueryListFindModule,
```
pldespaigne, line 21: I don't think you need that

libs/movie/src/lib/movie/form/movie.form.ts
```diff
@@ -254,6 +254,10 @@ export class MovieForm extends FormEntity<MovieControl, Movie> {
     return this.get('title');
   }
 
+  get estimatedBudget() {
+    return this.get('estimatedBudget');
+  }
+
```
pldespaigne, line 260: I don't think you need that, Max add getters at the form level

pldespaigne (changes requested):

libs/movie/src/lib/movie/form/budget/budget.component.scss
joanofdart: already did, : P

libs/movie/src/lib/movie/form/budget/budget.component.ts
joanofdart: alright

PR #3544, opened 2020-08-25
George/copy storage from prod to ci

george43g:
- updated config.yml to run backup job every week
- cron is filtered on develop
- also added script.sh and updated dockerfile - however only config.yml is in use
- a google docker image is used to run gsutil, which does fast copy of storage into a dir in `blockframes-ci`
- the service account for blockframes-ci was granted an extra permission to access prod storage.
- backup job has been run as a test: https://app.circleci.com/pipelines/github/blockframes/blockframes/10186/workflows/a09edf24-ed87-48a6-82ab-f464b2801bf7/jobs/20239


.circleci/Dockerfile
```diff
@@ -33,3 +33,9 @@ RUN export JAVA_HOME
 ## Install firebase and the emulator (requires java)
 RUN sudo npm install --global firebase-tools
 RUN firebase setup:emulators:firestore
+
+## Install gcloud and gsutil for bucket backup facilities
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
```
phpgeekfr, line 38: where does this files comes from `usr/share/keyrings/cloud.google.gpg` ( just so I can understand)

.circleci/sync-prod-storage-to-ci.sh
```diff
@@ -0,0 +1,20 @@
...
+export GOOGLE_APPLICATION_CREDENTIALS="creds.json"
+export TERM=xterm-256color
+
+## ! Use a special service account with permossionsOn both? or two service account keys? (if uploading from dir?)
+echo "${FIREBASE_CI_SERVICE_ACCOUNT}" > creds.json
+
+gcloud auth activate-service-account --key-file=creds.json
+gcloud --quiet config set project ${ENV}
+gcloud info
+gsutil -m cp -r gs://blockframes.appspot.com "gs://blockframes-ci-storage-backup/storage-backup-$(date +%d-%m-%Y)/"
```
phpgeekfr, line 20: Is this the manual way ?

.circleci/config.yml
```diff
@@ -385,3 +398,13 @@ workflows:
...
           requires:
             - build-and-test-all
+  copy-storage-from-prod-weekly:
+    triggers:
+      - schedule:
+          cron: '0 0 * * 0'
+          filters:
+            branches:
+              only:
+                - develop
```
phpgeekfr, line 408: I don't understand the only develop. Can you explain a bit?

phpgeekfr (approved):
LGTM. Just got some questions, see below

GrandSchtroumpf:
I'm not familiar with circleCi but on Github Action we would run some command (like `npm install -g ....`) to install the missing dependancies instead of putting it directly in a docker image. It is the "standard" way in circleCI ?

.circleci/Dockerfile
george43g: This is from google's docs on how to install gcloud in a debian/linux environment, non interactive.
This installs the utility using apt-get. For some reason, a package repository needs to be added first. It's all just linuxy stuff. To be honest it would take me a while to understand this command and it's not even tested.

I have achieved everything so far by using a google docker image with gsutil preinstalled. If we also wanted to run firebase-tools, we'd need to update our docker image using this dockerfile. However, I have not published the docker image yet. It would be using the script below.

.circleci/sync-prod-storage-to-ci.sh
george43g: correct, this would run in CI.

.circleci/config.yml
george43g: You must use a branch filter based on these docs: https://circleci.com/docs/2.0/workflows/#specifying-a-valid-schedule

I guess the reason is that CI needs to know which branch to run this on when triggered by `cron` rather than a push to a specific branch. Imagine this - it's triggered on a Sunday night - which branch does it read code from?
This also means, that until this PR is merged into `develop`, the schedule will not execute.

phpgeekfr: OK thanks I understand.
It should go on develop by the end of the day so we will be able to check it

george43g:
> I'm not familiar with circleCi but on Github Action we would run some command (like `npm install -g ....`) to install the missing dependancies instead of putting it directly in a docker image. It is the "standard" way in circleCI ?

Good question. `gsutil` and `gcloud` are not an `npm` lib. They are written in python I believe. You *could* try to run the install steps specified in the dockerfile as part of the workflow, however, this is untested. For example, the current docker image has firebase-tools installed; It also helps with speed and caching.

Here I have simply used a small default image from google that has gsutil installed.

Do you want me to either update our own docker image, or use the existing one and attempt to run the install commands as part of the workflow instead?

PR #3551, opened 2020-08-26
App summary page design completion

joanofdart:
Adding the warning showing to the top of the page.

### DIFF
--- a/libs/movie/src/lib/movie/pages/summary/summary.component.scss
+++ b/libs/movie/src/lib/movie/pages/summary/summary.component.scss
@@ -25,20 +25,12 @@ button.submit {
   margin: auto;
 }
 
-.warning {
+.mat-body {
   margin-bottom: 36px;
-  font-size: 18px;
 
   & span {
     display: block;
   }
 
-  .missing-fields {
-    color: #f6a92c
-  }
-
-  .mandatory-fields {
-    color: #ff3f33;
-  }
 }
 
--- a/libs/movie/src/lib/movie/pages/summary/summary.component.scss
+++ b/libs/movie/src/lib/movie/pages/summary/summary.component.scss
@@ -21,7 +21,24 @@ mat-divider {
   margin-bottom: 36px;
 }
 
-
 button.submit {
   margin: auto;
 }
+
+.warning {
+  margin-bottom: 36px;
+  font-size: 18px;
+
+  & span {
+    display: block;
+  }
+
+  .missing-fields {
+    color: #f6a92c
+  }
+
+  .mandatory-fields {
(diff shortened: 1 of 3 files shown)


libs/movie/src/lib/movie/pages/summary/summary.component.scss
```diff
@@ -21,7 +21,24 @@ mat-divider {
...
 }
 
-
 button.submit {
   margin: auto;
 }
+
+.warning {
+  margin-bottom: 36px;
+  font-size: 18px;
```
Mussuranae, line 30: Instead of putting font size here, you can add the class `mat-body` to get the good font-size

libs/movie/src/lib/movie/pages/summary/summary.component.scss
```diff
@@ -21,7 +21,24 @@ mat-divider {
...
+
+  & span {
+    display: block;
+  }
+
+  .missing-fields {
+    color: #f6a92c
```
Mussuranae, line 37: This one also, you can write here `var(--beware)` 
and in the `styles/theme.scss` file, you can add a variable `--beware: #{mat-color($background, beware)};`

If we want to change color in the theme, it will be a mess do update component if we put color directly in it.

Mussuranae (approved):

libs/movie/src/lib/movie/pages/summary/summary.component.ts
```diff
@@ -29,7 +31,10 @@ export class MovieFormSummaryComponent {
     private query: MovieQuery,
     private snackBar: MatSnackBar,
     private routerQuery: RouterQuery
-  ) { }
+  ) {
+    this.findInvalidControlsRecursive(this.form);
+    this.form.valueChanges.subscribe(() => this.findInvalidControlsRecursive(this.form))
```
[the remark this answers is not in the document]
joanofdart: just noticed that I need to dispose of this. Adding a subscription and the onDestroy LC.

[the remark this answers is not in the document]
pldespaigne: yup, put this in the `ngOnInit` and unsubscribe in the `ngOnDestroy`

pldespaigne (changes requested):

libs/movie/src/lib/movie/pages/summary/summary.component.scss
joanofdart: Nevermind, we already have **mandatory-field**  and **missing-field** classes

pldespaigne (approved):

PR #3555, opened 2020-08-26
More changes to the summary page

joanofdart:

### DIFF
--- a/libs/movie/src/lib/movie/pages/shell/shell.component.ts
+++ b/libs/movie/src/lib/movie/pages/shell/shell.component.ts
@@ -106,7 +106,7 @@ export class MovieFormShellComponent implements TunnelRoot, OnInit, AfterViewIni
     private dialog: MatDialog,
     private mediaService: MediaService,
     private route: ActivatedRoute,
-  ) { }
+  ) {}
 
   ngOnInit() {
     this.exitRoute = `../../../title/${this.query.getActiveId()}`;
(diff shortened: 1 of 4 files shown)


libs/movie/src/lib/movie/pages/summary/summary.component.html
```diff
@@ -88,20 +88,35 @@ <h5>
...
+              </a>
+            </ng-container>
+            <ng-container *ngFor="let country of form.originCountries.controls; let isLast = last" #showListOfCountries>
               <missing-control [control]="country" scope="TERRITORIES" [isLast]="isLast"></missing-control>
             </ng-container>
           </li>
           <li>
-            <strong>Original Language </strong>
-            <ng-container *ngFor="let language of form.originalLanguages.controls; let isLast = last">
+            <strong>Original Language</strong>
```
Mussuranae, line 102: You can add also the plural super-thing here, if there is more than one language.

pldespaigne (approved):

libs/movie/src/lib/movie/pages/shell/shell.component.ts
```diff
@@ -129,6 +130,19 @@ export class MovieFormShellComponent implements TunnelRoot, OnInit, AfterViewIni
     this.sub.unsubscribe();
   }
 
+  private checkIfElementIsReady(id: string) {
+    return new Promise<HTMLElement>((resolve, _) => {
+      const el = this.doc.getElementById(id);
+      if (el) {
+        resolve(el);
+      }
+      new MutationObserver((_, observer) => {
```
GrandSchtroumpf, line 139: as discussed on slack, it feels overkill, also have you check if it works on Safari. It feels like we'll need another polyfill on it. Maybe a simple setTimeout or something like that would have patch it.

joanofdart: I could actually do a simple timeout, but it wouldn't be reliable in my opinion.
Wouldn't it be affected depending on the user's speed?

Lastly, it its supported by safari from version 6 (2011 onwards).

PR #3558, opened 2020-08-26
Fix fallback image for default organization banner

RemcoSimonides:
Fix fallback image for default organization banner


fritzschoff (changes requested):
fix commit tree

RemcoSimonides:
Closed in favor of #3571 to fix commit tree

PR #3565, opened 2020-08-26
update on db-cleaning script - do not remove outdated invitations to events

phpgeekfr:


apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -433,19 +433,28 @@ describe('DB cleaning script', () => {
...
-      {
-        id: 'event-A',
-        // Outdated by 30 seconds
-        end: { _seconds: currentTimestamp - 30 },
-      },
       {
         id: 'event-B',
-        // Not outdated
-        end: { _seconds: currentTimestamp },
+        // Outdated by 30 seconds, but should be kept
```
GrandSchtroumpf, line 456: why ?

apps/backend-ops/src/db-cleaning.ts
```diff
@@ -349,10 +349,7 @@ function isInvitationValid(invitation: InvitationDocument, existingIds: string[]
 
       if (existingIds.includes(invitation.docId)) {
         const event = events.find(e => e.id === invitation.docId);
-        const eventEndTimestamp = event.end.toMillis();
-
-        // Cleaning finished events
-        if (eventEndTimestamp < currentTimestamp) {
+        if(event === undefined){
```
GrandSchtroumpf, line 352: If you don't use the event after prefer :
```typescript
if (event.sum(e => e.id === invitation.docId)) {
  return false
}
```

GrandSchtroumpf (approved):

apps/backend-ops/src/db-cleaning.spec.ts
phpgeekfr: why why ? ^^

GrandSchtroumpf (approved):

PR #3566, opened 2020-08-26
George/restore ci storage backup to local env

george43g:
- wrote script in `firebase-utils` so it can be accessed in `backend-ops` as well as a standalone script in `tools/scripts`.
- To test this script, checkout and run this command: `npx ts-node -P tools/tsconfig.tools.json tools/scripts/restore-storage-from-ci.ts`


libs/firebase-utils/src/lib/restore-storage-from-ci.ts
```diff
@@ -0,0 +1,91 @@
...
+  const ci = admin.initializeApp(
+    {
+      storageBucket: CI_STORAGE_BACKUP,
+      projectId: firebaseCI.projectId,
+      credential: admin.credential.cert(cert),
+    },
+    'CI-app'
+  );
+
+  // const app = admin.initializeApp(
```
GrandSchtroumpf, line 44: remove if you're not using

libs/firebase-utils/src/lib/restore-storage-from-ci.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  try {
+    const ciStorage = ci.storage();
+    const [files, nextQuery, apiResponse] = await ciStorage.bucket(CI_STORAGE_BACKUP).getFiles({
+      autoPaginate: false,
+      delimiter: '/',
+    });
+    const folders = apiResponse.prefixes as string[];
+    const latestFolder = folders
+      .map((folderName) => {
```
GrandSchtroumpf, line 60: I'm not sure to understand why you need to do that.

libs/firebase-utils/src/lib/restore-storage-from-ci.ts
```diff
@@ -0,0 +1,91 @@
...
+      })
+      .sort((a, b) => Number(a.date) - Number(b.date))
+      .pop();
+    const { folderName } = latestFolder;
+    console.log('Latest backup:', folderName);
+
+    console.log('Clearing storage bucket:', firebase.storageBucket);
+    try {
+      process.stdout.write(execSync(`gsutil -m rm -r "gs://${firebase.storageBucket}/*"`));
+    } catch (e) {}
```
GrandSchtroumpf, line 76: If gsutil fails because it's not install on developer's computer, then is it goes in the stdout or in the catch ?
If it goes in the catch you should display a message.

libs/firebase-utils/src/lib/restore-storage-from-ci.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    const cmd = `gsutil -m cp -r gs://${CI_STORAGE_BACKUP}/${folderName}* gs://${firebase.storageBucket}`;
+    console.log('Running command:', cmd);
```
GrandSchtroumpf, line 79: Maybe `"copying storage from blockframe-ci to your storage"`

GrandSchtroumpf (approved):

GrandSchtroumpf:
I think one of the first thing to do after the whole process is done, would be to factorize all the functions to not repeating ourselves in each of these method.

libs/firebase-utils/src/lib/restore-storage-from-ci.ts
george43g: This is copying from the ci storage backups bucket to your local one. The CI stores a history of backups with the date in the folderdame. So we parse it into a date and sort it to find the latest backup.

Issue #3552, opened 2020-08-26
changing user role within org does not work on CRM app 


phpgeekfr:
changing user role within org does not work on CRM app 
![image](https://user-images.githubusercontent.com/1194344/91300229-3a2dff80-e7a3-11ea-81ac-fc8019fa8ce2.png)
![image](https://user-images.githubusercontent.com/1194344/91300259-444ffe00-e7a3-11ea-9afd-b7c959021699.png)

This is because we try to update current permissionDocument (the one that the admin user is connected to and not the permission document belonging to user we want to update)

PR #3570, opened 2020-08-27
Fix wrong orgId used when updating user

RemcoSimonides:
Fix issue #3552

### DIFF
--- a/libs/organization/src/lib/organization/+state/organization.service.ts
+++ b/libs/organization/src/lib/organization/+state/organization.service.ts
@@ -9,7 +9,7 @@ import {
 } from './organization.model';
 import { OrganizationStore, OrganizationState } from './organization.store';
 import { OrganizationQuery } from './organization.query';
-import { CollectionConfig, CollectionService, WriteOptions, Query, queryChanges } from 'akita-ng-fire';
+import { CollectionConfig, CollectionService, WriteOptions } from 'akita-ng-fire';
 import { createPermissions, UserRole } from '../../permissions/+state/permissions.model';
 import { toDate } from '@blockframes/utils/helpers';
 import { AngularFireFunctions } from '@angular/fire/functions';
@@ -17,7 +17,6 @@ import { UserService, OrganizationMember, createOrganizationMember, PublicUser }
 import { PermissionsService, PermissionsQuery } from '@blockframes/permissions/+state';
 import { orgNameToEnsDomain, getProvider } from '@blockframes/ethers/helpers';
 import { network, baseEnsDomain } from '@env';
-import { QueryFn } from '@angular/fire/firestore/interfaces';
 
 @Injectable({ providedIn: 'root' })
 @CollectionConfig({ path: 'orgs' })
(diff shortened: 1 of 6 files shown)


libs/organization/src/lib/permissions/+state/permissions.service.ts
```diff
@@ -80,7 +82,7 @@ export class PermissionsService extends CollectionService<PermissionsState> {
 
   /** Update role of a member of the organization */
   private async _updateMemberRole(uid: string, role: UserRole) {
-    const orgId = this.query.getActiveId();
+    const orgId = (await this.userService.getValue(uid)).orgId
```
phpgeekfr, line 85: There is more to do here, you also need to update   :
```
this.query.hasAlreadyThisRole(uid, role)) 
this.hasLastSuperAdmin(uid, role)) 
```
Because this methods will use the query to check if user  already have this role & if he is last super admin but we want to check against the permission doc of the user we want to update, not the permission doc linking to the current logged in user

phpgeekfr (changes requested):
You missed something ;)

libs/organization/src/lib/permissions/+state/permissions.service.ts
phpgeekfr: My first idea was to move hasAlreadyThisRole isUserSuperAdmin outside of the query

libs/organization/src/lib/permissions/+state/permissions.query.ts
```diff
@@ -46,7 +46,4 @@ export class PermissionsQuery extends QueryEntity<PermissionsState, Permissions>
     return this.getActive().roles[userId] === 'superAdmin';
   }
```
phpgeekfr, line 48: UserRole import should not but usefull now.

libs/organization/src/lib/permissions/+state/permissions.service.ts
```diff
@@ -1,11 +1,12 @@
 import { Injectable } from '@angular/core';
 import { PermissionsQuery } from './permissions.query';
-import { UserRole, createDocPermissions } from './permissions.firestore';
+import { UserRole, createDocPermissions, PermissionsDocument } from './permissions.firestore';
 import { PermissionsState, PermissionsStore } from './permissions.store';
 import { CollectionService, CollectionConfig, AtomicWrite } from 'akita-ng-fire';
 import { firestore } from 'firebase/app';
 import { OrganizationQuery } from '@blockframes/organization/+state/organization.query';
 import { Contract } from '@blockframes/contract/contract/+state/contract.model';
+import { UserService } from '@blockframes/user/+state';
```
phpgeekfr, line 10: `private query: PermissionsQuery,` is not used anymore. You can remove it

phpgeekfr: also, I did not notice but the `removeMember` method in OrganizationService have the same problem. Can you do the same that you did on `updateMemberRole` ? PermissionsQuery import should not be usefull anymore

phpgeekfr (changes requested):
LGTM but I did not noticed that we have the same issue with removeMember method in OrganizationService (see comments).. Can you make the change ?

libs/organization/src/lib/permissions/+state/permissions.query.ts
RemcoSimonides: You have a good eye for detail :)

libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -9,15 +9,14 @@ import {
 } from './organization.model';
 import { OrganizationStore, OrganizationState } from './organization.store';
 import { OrganizationQuery } from './organization.query';
-import { CollectionConfig, CollectionService, WriteOptions, Query, queryChanges } from 'akita-ng-fire';
+import { CollectionConfig, CollectionService, WriteOptions } from 'akita-ng-fire';
 import { createPermissions, UserRole } from '../../permissions/+state/permissions.model';
 import { toDate } from '@blockframes/utils/helpers';
 import { AngularFireFunctions } from '@angular/fire/functions';
 import { UserService, OrganizationMember, createOrganizationMember, PublicUser } from '@blockframes/user/+state';
 import { PermissionsService, PermissionsQuery } from '@blockframes/permissions/+state';
```
phpgeekfr, line 17: PermissionsQuery not used anymore. Your linter can warn you about this

phpgeekfr (approved):
LGTM

PR #3577, opened 2020-08-27
Feat 3188 - storage rules

phpgeekfr:


tools/mandatory-env-vars.json
```diff
@@ -13,5 +13,6 @@
   {
     "key": "FIREBASE_CI_SERVICE_ACCOUNT",
     "msg": "You have missing service account keys. Please see: https://www.notion.so/cascade8/Preparing-your-Environment-e3c184db24d447c496c3241d4f16bc94"
-  }
+  },
+  { "key": "IMGIX_TOKEN", "msg": "API key for the video player is missing." }
```
GrandSchtroumpf, line 17: Imgix !== video player. Also give a link to the right documentation in notion.

apps/backend-functions/src/media.ts
```diff
@@ -1,101 +1,142 @@
...
+    if (filePath === savedRef) {
+      const to = bucket.file(filePath);
+      const [exists] = await from.exists();
+      if (exists) {
+        await from.copy(to);
+        await from.delete();
+      }
+      return true;
+    } else {
+      await from.delete();
```
GrandSchtroumpf, line 32: I'm not sure to understand when this code runs.

apps/backend-functions/src/media.ts
```diff
@@ -1,101 +1,142 @@
...
+      return;
+    }
 
-  if ( filePathElements.length < 4 ) {
-    throw new Error(`Upload Error : File Path ${filePath}
-    is malformed, it should at least contain 2 slash
-    Example: 'collection/id/field/fileName'`);
+    // unlink the firestore
+    // ! this will not work with array in the path like for poster
+    return doc.update({ [fieldToUpdate]: '' });
```
GrandSchtroumpf, line 60: I think you should use `FieldValue` for this kind of update. Not sure though

apps/backend-functions/src/media.ts
```diff
@@ -1,101 +1,142 @@
...
+    if (data.name !== ref) {
+      return;
+    }
 
-  if ( filePathElements.length < 4 ) {
-    throw new Error(`Upload Error : File Path ${filePath}
-    is malformed, it should at least contain 2 slash
-    Example: 'collection/id/field/fileName'`);
+    // unlink the firestore
+    // ! this will not work with array in the path like for poster
```
GrandSchtroumpf, line 59: poster isn't an array anymore I think.

apps/backend-functions/src/media.ts
```diff
@@ -1,101 +1,142 @@
...
+    const params = formatParameters(p);
+    let toSign = `${imgixToken}${data.ref}`;
 
-  const docData = docSnapshot.data()!;
+    if (!!params) {
+      toSign = `${toSign}?${params}`;
+    }
 
-  const fieldToUpdate = filePathElements.join('.');
+    return createHash('md5').update(toSign).digest('hex');
```
GrandSchtroumpf, line 90: Are we always using md5 for the hashes ?

apps/backend-functions/src/media.ts
```diff
@@ -1,101 +1,142 @@
...
- * eventually trigger post-processing (like image resize).
- */
-export async function linkFile(data: functions.storage.ObjectMetadata) {
+export const deleteMedia = async (data: { ref: string }, context: CallableContext): Promise<void> => {
 
-  // get the needed values
-  const { filePath, doc, fieldToUpdate } = await getDocAndPath(data.name);
+  if (!context?.auth) { throw new Error('Permission denied: missing auth context.'); }
+
+  if (!(await isAllowedToDelete(data.ref, context.auth.uid))) { throw new Error('Permission denied. User is not allowed'); }
```
GrandSchtroumpf, line 99: It's a little bit too much on one line : 
```typescript
const canDelete = await isAllowedToDelete(data.ref, contet.auth.uid);
if (canDelete) {
  throw ...
}
```

apps/backend-functions/src/media.ts
```diff
@@ -1,101 +1,142 @@
...
-  const { doc, docData, fieldToUpdate } = await getDocAndPath(data.name);
+function getPathInfo(ref: string) {
+  const refParts = ref.split('/');
 
-  // if firestore wasn't link to this file, we should not unlink it
-  const ref: string = get(docData, fieldToUpdate);
-  if (data.name !== ref) {
-    return;
+  const pathInfo = {} as any;
+  if (['protected', 'public'].includes(refParts[0])) {
```
GrandSchtroumpf, line 134: Maybe you can create a type with 'protected', 'public' & 'tmp' if it makes sense at some point to avoid typos.

apps/backend-functions/src/media.ts
```diff
@@ -1,101 +1,142 @@
...
-  // get the needed values
-  const { doc, docData, fieldToUpdate } = await getDocAndPath(data.name);
+function getPathInfo(ref: string) {
+  const refParts = ref.split('/');
 
-  // if firestore wasn't link to this file, we should not unlink it
-  const ref: string = get(docData, fieldToUpdate);
-  if (data.name !== ref) {
-    return;
+  const pathInfo = {} as any;
```
GrandSchtroumpf, line 133: type it:  `Record<string, string>` ?

apps/backend-ops/src/firestoreMigrations/0035.ts
```diff
@@ -0,0 +1,66 @@
+import { Firestore } from '../admin';
+import { File as GFile } from '@google-cloud/storage';
+import { startMaintenance, endMaintenance, runChunks, getDocAndPath } from '@blockframes/firebase-utils';
+import { firebase } from '@env';
+import { has, get } from 'lodash';
+export const { storageBucket } = firebase;
+const EMPTY_REF = '';
+
+export async function upgrade(_: Firestore, storage: Storage) {
```
GrandSchtroumpf, line 9: Why do you need the first params if you don't use it ?

apps/backend-ops/src/syncStorage.ts
```diff
@@ -1,6 +1,5 @@
-import { loadAdminServices } from "@blockframes/firebase-utils";
+import { loadAdminServices, getDocAndPath } from "@blockframes/firebase-utils";
```
GrandSchtroumpf, line 1: Are you sure you need it ?

libs/firebase-utils/src/lib/watermark.ts
```diff
@@ -19,22 +19,26 @@ export function getWatermark(email: string, firstName: string = '', lastName: st
 /**
  * - Generate a svg file with the name & email of the user
  * - Store the watermark file in the storage bucket
- * - User's firestore doc is updated by onFileUpload backend function
+ * - Update the user document
  */
-export async function upsertWatermark(user: PublicUser, bucketName: string): Promise<any> {
+export async function upsertWatermark(user: PublicUser, bucketName: string, protectedMedia = false): Promise<any> {
```
GrandSchtroumpf, line 24: Better to have the third params as a string literal : 
```typescript
export async function upsertWatermark(user: PublicUser, bucketName: string, privacy: FilePrivacy = 'public')
```
If we need other folders in the future it's better than using a boolean.

libs/firebase-utils/src/lib/watermark.ts
```diff
@@ -19,22 +19,26 @@ export function getWatermark(email: string, firstName: string = '', lastName: st
...
 
   if (!user.email) {
     throw new Error(`Cannot generate watermark for user ${user.uid} because 'email' is not provided.`);
   }
 
   const watermark = getWatermark(user.email, user.firstName, user.lastName);
 
-  const ref = `users/${user.uid}/watermark/${user.uid}.svg`;
+  const ref = `${protectedMedia ? 'protected' : 'public'}/users/${user.uid}/watermark/${user.uid}.svg`;
```
GrandSchtroumpf, line 32: With comment above you can do `${privacy}/users/...`

libs/media/src/lib/+state/media.service.ts
```diff
@@ -3,20 +3,24 @@ import { Injectable, Injector } from "@angular/core";
...
 import { delay, BehaviorStore } from "@blockframes/utils/helpers";
+import { ImageParameters, getImgSize, getImgIxResourceUrl } from "../directives/image-reference/imgix-helpers";
+import { clamp } from '@blockframes/utils/utils';
 
 @Injectable({ providedIn: 'root' })
 export class MediaService {
 
   private _tasks = new BehaviorStore<AngularFireUploadTask[]>([]);
 
+  private breakpointsWidth = [600, 1024, 1440, 1920];
```
GrandSchtroumpf, line 22: just "breakpoints". It's always on width. I think theses variables exists somewhere though. It would be better to import them that hardcode them here.

libs/media/src/lib/+state/media.service.ts
```diff
@@ -27,22 +31,19 @@ export class MediaService {
...
 
   async upload(uploadFiles: UploadData | UploadData[]) {
     const files = Array.isArray(uploadFiles) ? uploadFiles : [uploadFiles];
-    const tasks = files.map(file => this.storage.upload(`${file.path}${file.fileName}`, file.data));
+    /**
+     * @dev Every file goes into tmp dir.
+     * Then a backend functions performs a check on DB document to check if
+     * file have to be moved to correct folder or deleted.
+     */
+    const tasks = files.map(file => this.storage.upload(`tmp/${file.path}${file.fileName}`, file.data));
```
GrandSchtroumpf, line 46: It would be nice if the `file.path` doesn't have the last "/", as it makes the code a little bit harder to read. And a little more prompt to error

libs/media/src/lib/+state/media.service.ts
```diff
@@ -127,4 +122,101 @@ export class MediaService {
...
+  /**
+   * This https callable method will check if current user asking for the media
+   * have the rights to do so. 
+   * @dev protected resources are stored the same way than other resources but in the "protected" directory.
+   * For example 
+   * @param ref (without "/protected")
+   * @param parametersSet ImageParameters[]
+   */
+  private async getProtectedMediaToken(ref: string, parametersSet: ImageParameters[]): Promise<string[]> {
+    ref = !ref.startsWith('/') ? `/${ref}` : ref;
```
GrandSchtroumpf, line 134: If you can stripe out the 'protected" in case of a mistake in the params.

libs/media/src/lib/+state/media.service.ts
```diff
@@ -127,4 +122,101 @@ export class MediaService {
...
+    const f = this.functions.httpsCallable('getMediaToken');
```
GrandSchtroumpf, line 135: I think it's better/cleaner to set that as a params of the service: 
```typescript
class MediaService {
  private getMediaToken = this.unctions.httpsCallable('getMediaToken');
}
```

libs/media/src/lib/+state/media.service.ts
```diff
@@ -127,4 +122,101 @@ export class MediaService {
...
+    return f({ ref, parametersSet }).toPromise();
+  }
+
+  async generateImageSrcset(ref: string, p: ImageParameters): Promise<string> {
+    const refParts = ref.split('/');
+    const protectedMedia = refParts[0] === 'protected';
+
+    if (['public', 'protected'].includes(refParts[0])) {
```
GrandSchtroumpf, line 143: Same as discussed above.

libs/media/src/lib/+state/media.service.ts
```diff
@@ -127,4 +122,101 @@ export class MediaService {
...
+      refParts.shift();
+      ref = refParts.join('/');
+    }
+
+    const sizes = getImgSize(ref);
```
GrandSchtroumpf, line 148: Can you do : 
```typescript
const params = getImgSize(ref).map(size => ({ ...p, w: size } as ImageParameters));
const tokens = protectedMedia
  ? await this.getProtectedMediaToken(ref, params)
  : [];
const urls = params.map((param, index) => {
  if (tokens[index]) { param.s = tokens[index] };
  return `${getImgIxResourceUrl(ref, params)} ${param.w}w`;
})
```

libs/media/src/lib/+state/media.service.ts
```diff
@@ -127,4 +122,101 @@ export class MediaService {
...
+  generateSingleImageUrl(ref: string, p: ImageParameters, w = 0): Promise<string> {
+    const parameters: ImageParameters = { ...p, w };
+    return this.generateImgIxUrl(ref, parameters);
+  }
+
+  /**
+   * 
+   * @param ref string
+   */
+  generatePdfUrl(ref: string): Promise<string> {
```
GrandSchtroumpf, line 179: Do we really need all these methods ? Look like a lot a repetition here.

libs/media/src/lib/+state/media.service.ts
```diff
@@ -127,4 +122,101 @@ export class MediaService {
...
+    return this.generateImgIxUrl(ref);
+  }
+
+  /**
+   * 
+   * @param ref string
+   * @param p any
+   */
+  async generateImgIxUrl(ref: string, p: any = {}): Promise<string> {
```
GrandSchtroumpf, line 188: p shouldn't be any. Can it be a string ? a number ? an array ? ...
Also give a better name.

libs/media/src/lib/+state/media.service.ts
```diff
@@ -127,4 +122,101 @@ export class MediaService {
...
+    const refParts = ref.split('/');
+    const protectedMedia = refParts[0] === 'protected';
+
+    if (['public', 'protected'].includes(refParts[0])) {
+      refParts.shift();
+      ref = refParts.join('/');
+    }
```
GrandSchtroumpf, line 195: Can you do : 
```typescript
const refParts = ref.split('/');
const privacy = refParts.shift();
if (['public', 'protected'].includes(privacy)) {
  ref = refParts;
}
```

libs/media/src/lib/+state/media.service.ts
```diff
@@ -127,4 +122,101 @@ export class MediaService {
...
+
+    if (protectedMedia) {
+      p.s = (await this.getProtectedMediaToken(ref, [p]))[0];
```
GrandSchtroumpf, line 198: What is p ? What it s ? What is the first value returned by `getProtectedMediaToken` ?
This line it pretty unreadable.

apps/backend-functions/src/media.ts
phpgeekfr: This is what imgIx expect yes

libs/media/src/lib/directives/image-reference/imgix-helpers.ts
```diff
@@ -51,50 +53,7 @@ export function formatParameters(parameters: ImageParameters): string {
...
-  );
-
-  const sizeParameters: ImageParameters = { ...parameters, w: imageWidth };
-  const query = formatParameters(sizeParameters);
-
-  return `https://${firebase.projectId}.imgix.net/${ref}?${query}`;
-}
+export function getImgIxResourceUrl(ref: string, parameters: ImageParameters, protectedMediaDir = 'protected') {
+  const query = formatParameters(parameters);
+  return `https://${firebase.projectId}${parameters.s ? `-${protectedMediaDir}` : ''}.imgix.net/${ref}?${query}`;
```
GrandSchtroumpf, line 58: Create intermediary variables to make it more readable.
```typescript
const something = parameters.s ? `${firebase.projectId}-${protectedMediaDir}` : firebase.projectId}:
return `https://${something }.imgix.net/${ref}?${query}`
```

libs/media/src/lib/+state/media.service.ts
```diff
@@ -127,4 +122,101 @@ export class MediaService {
...
+    }
+
+    return getImgIxResourceUrl(ref, p);
+  }
+
+  generateBackgroundImageUrl(ref: string, p: ImageParameters): Promise<string> {
+
+    let clientWidth = 1024;
```
GrandSchtroumpf, line 206: Why is it hardcoded ?

apps/backend-functions/src/media.ts
phpgeekfr: linkFile Is called at each new file on the storage.
We only take care about the `tmp` dir.
If the ref of the file in tmp dir is found on the document in db, the file is moved to the right dir.
If the ref is not found on db, we delete it because it means that this file is lost

libs/media/src/lib/components/cropper/cropper.component.ts
```diff
@@ -138,10 +147,10 @@ export class CropperComponent implements OnInit {
       this.nextStep('show');
 
       // regexp selects part of string after the last . in the string (which is always the file extension) and replaces this by '.webp'
-      const fileName = this.file.name.replace(/(\.[\w\d_-]+)$/i, '.webp');
+      const fileName = sanitizeFileName(this.file.name.replace(/(\.[\w\d_-]+)$/i, `.webp`));
```
GrandSchtroumpf, line 150: I remember we had some issues some sanitizing the filename. Are you sure of you ?
Also use single-quote for 'webp'

libs/media/src/lib/components/cropper/cropper.component.ts
```diff
@@ -171,4 +180,9 @@ export class CropperComponent implements OnInit {
     this.step.next(name);
   }
 
+
+  /** Returns an observable of the download url of an image based on its reference */
```
GrandSchtroumpf, line 184: observable ? Not a Promise as typed ?

libs/media/src/lib/components/upload/file-upload/file-upload.component.ts
```diff
@@ -28,6 +28,7 @@ export class FileUploadComponent implements OnInit {
   /** firestore path */
   @Input() storagePath: string;
   @Input() form: HostedMediaForm;
+  @Input() protected = false;
```
GrandSchtroumpf, line 31: I would prefer `@Input privacy: Privacy` as string literal

libs/media/src/lib/directives/image-reference/background-reference.directive.ts
```diff
@@ -31,10 +32,9 @@ export class BackgroundReferenceDirective implements OnInit, OnDestroy {
       this.ref$.next('');
     } else {
       try {
+        this.mediaService.generateBackgroundImageUrl(image, this.parameters)
```
GrandSchtroumpf, line 35: Change to : 
```typescript
this.mediaService.generateBackgroundImageUrl(image, this.parameters)
 .then(url => this.ref$.next(url))
 .catch(err => this.ref$.next(''));
```

libs/movie/src/lib/movie/components/promotional-links/promotional-links.pipe.ts
```diff
@@ -1,23 +1,34 @@
...
+  constructor(
+    private mediaService: MediaService,
+  ) { }
+
+  async transform(links: string[], movie: Movie): Promise<any[]> {
+    const _links = await Promise.all(links.map(async link => {
+      const ref = (movie.promotional[link] as string);
+      const useImgIx = ['scenario', 'presentation_deck'].includes(link);
+      if (!!ref) {
+        const url = useImgIx ? await this.mediaService.generatePdfUrl(ref) : ref;
```
GrandSchtroumpf, line 21: Can the url change depending on external values than the links & movie ?
Like, if there are some state in the media service that change can it impact the result ?
If yes, change `pure: true` to `pure: false`

libs/utils/src/lib/file-sanitizer.ts
```diff
@@ -1,27 +1,32 @@
...
+  const fileParts = str.split('.');
+  // removes extension
+  fileParts.pop();
+  // replace spaces by "-"
+  const fileNameWithoutExt = fileParts.join('.').split(' ').join('-');
+
+  return `${fileNameWithoutExt.substr(0, 100)}.${getFileExtension(str)}`;
+}
+
+export function getStoragePath(path: string, protectedMedia = false): string {
```
GrandSchtroumpf, line 15: Use a string literal instead of a boolean.

GrandSchtroumpf (changes requested):

apps/backend-ops/src/firestoreMigrations/0035.ts
phpgeekfr: This is how migrations are started.
CF `apps/backend-ops/src/migrations.ts`
```
    for (const migration of migrations) {
      console.info(`applying migration: ${migration.version}`);
      await migration.upgrade(db, storage);
      console.info(`done applying migration: ${migration.version}`);
    }
```

apps/backend-ops/src/syncStorage.ts
phpgeekfr: It is used in this file but the entire file will be removed in https://github.com/blockframes/blockframes/pull/3669

libs/media/src/lib/+state/media.service.ts
phpgeekfr: It is a private method called by methods that already strips it out. But it can add another check if you want

libs/media/src/lib/+state/media.service.ts
phpgeekfr: Not my code but this is the default width for the image that can be changed after :
```
    if (!!window || !!window.innerWidth) {
      clientWidth = clamp(window.innerWidth, this.breakpointsWidth);
    }
```

Do you want to put this in env or something?

libs/media/src/lib/components/cropper/cropper.component.ts
phpgeekfr: The sanitizeFileName method has been changed. It is now only replacing spaces by "-" and make sure that the file length is not more than 100  (max is 255). So It should not create issues. (also updated the quotes for wepb)

apps/backend-functions/src/media.ts
GrandSchtroumpf: Maybe add a comment to explain that we delete() lost files.

libs/media/src/lib/+state/media.service.ts
GrandSchtroumpf: No it's fine. Maybe add a comment "// default width".

libs/media/src/lib/+state/media.service.ts
phpgeekfr: No I can't because I don't want to shift the array if it does not include public or protected. But I made some updates on this method that make it more readable

GrandSchtroumpf: You're not reusing `refParts` later so there is no problem.
- if privacy is 'public' or 'protected' you reassign the ref with the shifted version
- else you keep ref as it is.

GrandSchtroumpf (changes requested):

apps/backend-functions/src/media.ts
phpgeekfr: I don't understand your remark

libs/firebase-utils/src/lib/firebase-utils.ts
```diff
@@ -21,6 +22,71 @@ export function getCollection<T>(path: string): Promise<T[]> {
...
+  const db = admin.firestore();
+
+  if (!filePath) {
+    throw new Error('Upload Error : Undefined File Path');
+  }
+
+  const filePathElements = filePath.split('/');
+
+  if (filePathElements.length < 4) {
+    throw new Error(`Upload Error : File Path ${filePath}
```
GrandSchtroumpf, line 39: I recommend to do : 
```typescript
const error  = `Upload Error : File Path ${filePath} is malformed.`;
const solution = 'It should at least contain 3 slash.';
const example = 'Example: 'public/collection/id/field/fileName';
throw new Error(`${error} ${solution}\n${example}`);
```

libs/media/src/lib/+state/media.service.ts
```diff
@@ -51,13 +54,14 @@ export class MediaService {
...
-   * Delete a file from teh firebase storage.
+   * Delete a file from the firebase storage.
    * @note the function needs the **full** path of the file
    * **this include the file name!**
    * @note usually you can use `HostedMediaFormValue.oldRef` to feed the `path` param
    */
-  async removeFile(path: string) {
-    await this.storage.ref(path).delete().toPromise();
+  async removeFile(ref: string) {
+    const f = this.functions.httpsCallable('deleteMedia');
```
GrandSchtroumpf, line 63: Do the same as for `getMediaToken` by putting it at the top.

GrandSchtroumpf (approved):

PR #3579, opened 2020-08-27
Review tunnel

Mussuranae:
Review the tunnel, fix little bug and display correctly some informations, css also


styles/components/mat-list.scss
```diff
@@ -7,6 +7,11 @@
   $background: map-get($theme, background);
   $foreground: map-get($theme , foreground);
 
+  // use mainly for the sidenav in movie tunnel, I can't override this in the component, it has to be at the style root.
+  .mat-list-base .mat-list-item .mat-list-item-content {
+    padding: 0px;
```
joanofdart, line 12: no need for "px" on "0"

joanofdart (approved):

PR #3586, opened 2020-08-28
feat(prepareForTesting): show warning if needed .env vars are missing

george43g:
I don't know if this solution is way overkill. I could simply do an `if` `then` `console.log` if you prefer.


phpgeekfr (approved):
LGTM !

PR #3593, opened 2020-08-31
Fix image ref being undefined

RemcoSimonides:
Fix for error on sentry: https://sentry.io/organizations/cascade-8/issues/1850806475/?project=1540126&referrer=slack


phpgeekfr (approved):
Should do the trick but I would like to find some `uids` causing this ..

phpgeekfr:
Found it : MHDMpU3oKteg5Y2k4Jx7LDBxL8U2

PR #3595, opened 2020-09-01
refactor(backend-ops): general bug fixes and code cleanup, moved load…

george43g:
…AdminServices

- refactored out initialisation code
- small bug fixes

- this is branched from my other pull request george/prepareForTesting so we should merge this first then that (I wanted to retain my work from there as I've already done some refactoring I needed for this)


apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -7,10 +7,11 @@ import { appUrl } from '@env';
 import { syncUsers, generateWatermarks } from './users';
 import { upgradeAlgoliaMovies, upgradeAlgoliaOrgs, upgradeAlgoliaUsers } from './algolia';
 import { migrate } from './migrations';
-import { restore, loadAdminServices } from './admin';
+import { restore } from './admin';
+import { loadAdminServices } from "@blockframes/firebase-utils";
 import { cleanDeprecatedData } from './db-cleaning';
 import { cleanStorage } from './storage-cleaning';
-import { syncStorage } from './syncStorage';
+// import { syncStorage } from './syncStorage';
```
GrandSchtroumpf, line 14: If you don't need, remove

libs/firebase-utils/src/lib/util.ts
```diff
@@ -1,3 +1,7 @@
+import * as admin from 'firebase-admin';
```
GrandSchtroumpf, line 1: Maybe call the file "utils" with a "s" as there are several tools.

libs/firebase-utils/src/lib/util.ts
```diff
@@ -21,3 +25,58 @@ export function warnMissingVars(): void | never {
...
+
+export interface AdminServices {
+  auth: admin.auth.Auth;
+  db: admin.firestore.Firestore;
+  storage: admin.storage.Storage;
+  firebaseConfig: { projectId: string };
+  ci: admin.app.App;
+}
+
+export let app: admin.app.App;
```
GrandSchtroumpf, line 39: I don't feel comfortable exporting a mutable object. I would recommend not to do it.

libs/firebase-utils/src/lib/util.ts
```diff
@@ -21,3 +25,58 @@ export function warnMissingVars(): void | never {
...
+export let ci: admin.app.App;
+
+export function loadAdminServices(): AdminServices {
+  config();
+  warnMissingVars()
```
GrandSchtroumpf, line 44: You forgot a semi colon ;)

tools/scripts/upload-db.ts
```diff
@@ -1,48 +1,28 @@
...
-  cert = process.env.FIREBASE_CI_SERVICE_ACCOUNT;
-}
-
-admin.initializeApp({
-  storageBucket: backupBucket,
-  projectId: firebase.projectId,
-  credential: admin.credential.cert(cert)
-});
-const storage = admin.storage();
+const ciStorage = loadAdminServices().ci.storage();
```
GrandSchtroumpf, line 7: Wouldn't it better to put it in the function ?

GrandSchtroumpf (changes requested):

libs/firebase-utils/src/lib/util.ts
george43g: Interesting point,
If you read nx documentation on how they recommend people structure and name their libs, it's counter-intuitive but interesting.
So for example, if we had firebase-utils, you'll allow the directory structure to group libs together that share a purpose.
So you could have:
`libs/util (general global js utils)`
`libs/firebase/util (utils for firebase)`
`libs/firebase/testing (test related stuff with firebase)`
`libs/firebase/testing/cypress (hold firebase cypress plugins)`
`libs/firebase/testing/util (hold firebase testing utils)`
`libs/festival-e2e/components (hold angular components/modults)`
In the above example, you would have many folders called `util` but their context determines their purpose.

apps/backend-ops/src/firebaseSetup.ts
george43g: It's commented out in the function as well, so if it's re-introduced so will the import. Are we getting rid of that function all together?

libs/firebase-utils/src/lib/util.ts
george43g: Ahh sorry I didn't see this - this was a remnant from the refactoring feature of vscode

apps/backend-ops/src/firebaseSetup.ts
phpgeekfr: syncStorage will be uncommented once migration 31 is passed and some tests have been done on it. So you can leave it I think

tools/scripts/upload-db.ts
george43g: when ci app is used, we don't know which parts of it are going to be needed.
Ideally you'd have an object like:
```
{ project, app, projectCI, appCI } 
```
We could do similar to clean up the use of `env/**` - seems like a future thing. lmk.

phpgeekfr (changes requested):
When running the tests, I get : 
```
Test suite failed to run

    TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    apps/backend-ops/src/db-cleaning.spec.ts:16:10 - error TS2305: Module '"./admin"' has no exported member 'loadAdminServices'.

    16 import { loadAdminServices } from './admin';
```
npm run firebase:emulator
npx jest apps/backend-ops/src/storage-cleaning.spec.ts 
npx jest apps/backend-ops/src/db-cleaning.spec.ts

george43g:
ahh the tests! Tottally forgot about that!

george43g:
So every time `loadAdminServices` is called it will try to connect to the CI environment as well. The ideal way to write this function is to make it dynamic somehow, ie, CI is only initialised when it is explicitly called. Otherwise this could result in errors down the line.

Issue #3592, opened 2020-08-31
Admin panel - Prepare external support


Jackseed:
## Description
An external company will act as our client support. We will only give them access to admin panel so they don't use firebase directly. 
They should start working early September, we need to update admin panel before that. 

## Features and updates requested
- [x] Org import (/c/o/dashboard/import), contact email shouldn't be necessary

maltmayer:
- [ ] event invitation through CRM : should come from the org organizing the screening, not Archipel Content. 
![image](https://user-images.githubusercontent.com/55785823/91712534-34ac2d00-eb88-11ea-836e-2ebfc48d17bc.png)

maltmayer:
- [ ] "change user role in an org" function does not work

maltmayer:
- [ ] pages take a lot of time to load, would be great to optimize that

- [ ] remember filters/research through navigation (when we go back to previous page, would be great to retrieve the filter used for the search, otherwise we keep researching the same things over and over again)

maltmayer:
- [ ] would be great to have 2 options to create orgs : 1. through import, 2. one by one manually (just like for movies on the dashboard)

Jackseed:
- [ ] Heavy but important: we need to add a delete button (+ confirmation dialog) on users. The delete should delete the user and all corresponding entry in the db (id in org userIds, invitations, notifications,...)

RemcoSimonides:
A combination of Pagination, Filtering and Sorting isn't possible to make. Therefore we need to come up with a different solution to improve performance. Here are some of the options;

1. Remove filtering functionality or focus filtering on only one field (or select a field you want to filter on)
2. Remove sorting functionality (by switching to Algolia)
3. Remove distribution rights count & contracts count from movies, and user role from users

Hopefully soon I'll come up with more solutions... to be continued

RemcoSimonides:
- [ ] Move user from one org to another without invitation

PR #3599, opened 2020-09-01
Organization email no longer required

RemcoSimonides:
Fix first to do in issue #3592

### DIFF
--- a/libs/import/src/lib/components/view-extracted-elements/organizations/view-extracted-organizations.component.ts
+++ b/libs/import/src/lib/components/view-extracted-elements/organizations/view-extracted-organizations.component.ts
@@ -239,6 +239,17 @@ export class ViewExtractedOrganizationsComponent implements OnInit {
     // REQUIRED FIELDS
     //////////////////
 
+    // EMAIL
+    if (!organization.email) {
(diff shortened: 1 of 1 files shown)


libs/import/src/lib/components/view-extracted-elements/organizations/view-extracted-organizations.component.ts
```diff
@@ -239,17 +239,6 @@ export class ViewExtractedOrganizationsComponent implements OnInit {
     // REQUIRED FIELDS
     //////////////////
 
-    // EMAIL
-    if (!organization.email) {
-      importErrors.errors.push({
```
phpgeekfr, line 244: You can also transform it to `warning` if needed

phpgeekfr (approved):
It you tested some import & it worked, LGTM ;)

libs/import/src/lib/components/view-extracted-elements/organizations/view-extracted-organizations.component.ts
RemcoSimonides: I like that idea

Issue #3514, opened 2020-08-21
remove syncUsers patch that checks for dupe user emails in backend-ops


george43g:
make sure to remove this code eventually once no longer needed
`apps/backend-ops/src/users.ts`
`async function createAllUsers`

PR #3604, opened 2020-09-02
George/fix anon sendgrid score

george43g:
- must run npm install to test locally - added `faker` npm package
- random numbers added to prevent collision

email format is     user.email = `dev+${name}-${rand}@cascade8.com`;


tools/scripts/anonymize-db.ts
```diff
@@ -25,9 +26,14 @@ function anonymize({ docPath, content }) {
   return { docPath, content };
 }
 function updateEmail(user) {
-  if (user && user.email) {
-    const [prefix] = user.email.split('@');
-    user.email = `${prefix}@fake.com`;
+  if (user?.email) {
+    // const [prefix] = user.email.split('@');
```
phpgeekfr, line 30: remove if not used

phpgeekfr (approved):
LGTM, but it looks like you can do #3514 in this PR too

PR #3607, opened 2020-09-02
added some more cleaning in migration 31

phpgeekfr:


apps/backend-ops/src/firestoreMigrations/0031.ts
```diff
@@ -206,6 +206,27 @@ async function updateMovies(
       });
     }
 
+    if (movie.salesCast?.cast && movie.salesCast.cast.length) {
```
GrandSchtroumpf, line 209: You can do : `if (movie.salesCast?.cast?.length)`

apps/backend-ops/src/firestoreMigrations/0031.ts
```diff
@@ -218,22 +239,22 @@ async function updateMovies(
...
+          if (o.avatar) {
             o.avatar = '';
           }
           return o;
         });
       }
     });
 
-    if(movie.festivalPrizes?.prizes && movie.festivalPrizes.prizes.length) {
+    if (movie.festivalPrizes?.prizes && movie.festivalPrizes.prizes.length) {
```
GrandSchtroumpf, line 255: Same as above

GrandSchtroumpf (approved):

PR #3615, opened 2020-09-03
Send email from organization

RemcoSimonides:
To do in issue #3592 

"event invitation through CRM : should come from the org organizing the screening, not the org of the connected blockframesAdmin"


GrandSchtroumpf (approved):

RemcoSimonides:
Issue description was misunderstood and therefore this PR is nullified

PR #3616, opened 2020-09-03
fix(anonymize-db.ts): fix major bug with firebase initialisation, kee…

george43g:
…p emails consistent

- consistent emails
- fix initialisation bug with scripts


phpgeekfr (approved):
LGTM

PR #3617, opened 2020-09-03
Improved performance for CRM overview

RemcoSimonides:
to do in issue #3592 

"pages take a lot of time to load, would be great to optimize that. If that means reducing the nb of info on 1 page, than I think a good number would be 50 per page by default (and then we can change if need be)"

### DIFF
--- a/libs/admin/src/lib/admin-panel/pages/movies/movies.component.ts
+++ b/libs/admin/src/lib/admin-panel/pages/movies/movies.component.ts
@@ -1,9 +1,11 @@
 import { Component, OnInit, ChangeDetectorRef, ChangeDetectionStrategy } from '@angular/core';
 import { MovieService } from '@blockframes/movie/+state/movie.service';
-import { getValue, downloadCsvFromJson } from '@blockframes/utils/helpers';
+import { getValue, downloadCsvFromJson, BehaviorStore } from '@blockframes/utils/helpers';
 import { DistributionRightService } from '@blockframes/distribution-rights/+state/distribution-right.service';
 import { ContractService } from '@blockframes/contract/contract/+state/contract.service';
 import { OrganizationService, orgName, Organization } from '@blockframes/organization/+state';
+import { PageEvent } from '@angular/material/paginator';
+import { Sort } from '@angular/material/sort';
 
 @Component({
   selector: 'admin-movies',
@@ -14,31 +16,24 @@ import { OrganizationService, orgName, Organization } from '@blockframes/organiz
 export class MoviesComponent implements OnInit {
   public versionColumns = {
     'id': 'Id',
-    'main.internalRef': 'Internal Ref',
     'main.poster': 'Poster',
     'main.title.original': 'Original title',
     'org': 'Organization',
     'main.storeConfig.status': 'Status',
-    'main.storeConfig.storeType': 'Store type',
-    'distributionRightsInfo': 'Distribution rights',
-    'contractsInfo': 'Contracts',
     'edit': 'Edit',
   };
 
   public initialColumns: string[] = [
     'id',
     'main.poster',
-    'main.internalRef',
     'main.title.original',
     'org',
     'main.storeConfig.status',
-    'main.storeConfig.storeType',
-    'distributionRightsInfo',
-    'contractsInfo',
     'edit',
   ];
   public rows: any[] = [];
   public orgs: Record<string, Organization> = {};
+  public exporting = new BehaviorStore(false);
 
   constructor(
     private movieService: MovieService,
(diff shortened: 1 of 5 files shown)


libs/admin/src/lib/admin-panel/pages/movies/movies.component.ts
```diff
@@ -1,9 +1,11 @@
 import { Component, OnInit, ChangeDetectorRef, ChangeDetectionStrategy } from '@angular/core';
 import { MovieService } from '@blockframes/movie/+state/movie.service';
-import { getValue, downloadCsvFromJson } from '@blockframes/utils/helpers';
+import { getValue, downloadCsvFromJson, BehaviorStore } from '@blockframes/utils/helpers';
```
RemcoSimonides, line 3: 🤔 actually I don't need to use BehaviorStore. A simple Boolean would suffice. Although we could use Behaviorstore to have more statuses like 'download complete'

libs/admin/src/lib/admin-panel/pages/movies/movies.component.ts
```diff
@@ -14,31 +16,24 @@ import { OrganizationService, orgName, Organization } from '@blockframes/organiz
...
   };
 
   public initialColumns: string[] = [
     'id',
     'main.poster',
-    'main.internalRef',
     'main.title.original',
     'org',
     'main.storeConfig.status',
-    'main.storeConfig.storeType',
```
phpgeekfr, line 35: this info is not needed anymore  ?

libs/admin/src/lib/admin-panel/pages/movies/movies.component.ts
```diff
@@ -14,31 +16,24 @@ import { OrganizationService, orgName, Organization } from '@blockframes/organiz
...
-    'main.storeConfig.storeType': 'Store type',
-    'distributionRightsInfo': 'Distribution rights',
-    'contractsInfo': 'Contracts',
     'edit': 'Edit',
   };
 
   public initialColumns: string[] = [
     'id',
     'main.poster',
-    'main.internalRef',
```
phpgeekfr, line 31: this info is not needed anymore  ?

libs/admin/src/lib/admin-panel/pages/users/users.component.ts
```diff
@@ -17,14 +17,9 @@ import { getOrgModuleAccess } from '@blockframes/utils/apps';
 export class UsersComponent implements OnInit {
   public versionColumns = {
     'uid': 'Id',
-    'avatar': 'Avatar',
```
phpgeekfr, line 20: this info is not needed anymore  ?

libs/admin/src/lib/admin-panel/pages/users/users.component.ts
```diff
@@ -97,22 +83,48 @@ export class UsersComponent implements OnInit {
...
+        return {
+          ...u,
+          firstConnexion: this.adminQuery.getFirstConnexion(u.uid),
+          lastConnexion: this.adminQuery.getLastConnexion(u.uid),
+          edit: {
+            id: u.uid,
+            link: `/c/o/admin/panel/user/${u.uid}`,
+          },
+          org: org,
+          orgCountry: org && org.addresses.main.country ? org.addresses.main.country : undefined,
```
phpgeekfr, line 102: I think you can do directly :
```
orgCountry:  org?.addresses?.main.country
```

because if this `org?.addresses?.main.country` is undefined, it will assign orgCountry to undefined

libs/admin/src/lib/admin-panel/pages/users/users.component.ts
```diff
@@ -97,22 +83,48 @@ export class UsersComponent implements OnInit {
...
+      });
+      const data = await Promise.all(promises);
+      const exportedRows = data.map(r => ({
+        'userId': r.uid,
+        'first name': r.firstName ? r.firstName : '--',
+        'last name': r.lastName ? r.lastName : '--',
+        'organization': r.org ? orgName(r.org) : '--',
+        'org id': r.orgId ? r.orgId : '--',
+        'type': r.type ? r.type : '--',
+        'country': r.org && r.org.addresses.main.country ? r.org.addresses.main.country : '--',
```
phpgeekfr, line 115: you can do  :
```
country : r.org?.addresses.main.country || '--'
```

libs/admin/src/lib/admin-panel/pages/users/users.component.ts
```diff
@@ -73,13 +63,9 @@ export class UsersComponent implements OnInit {
           link: `/c/o/admin/panel/user/${u.uid}`,
         },
         org: org,
-        orgCountry: org && org.addresses.main.country ? org.addresses.main.country : undefined,
```
phpgeekfr, line 76: this line is not querying db, why not keeping it ?

libs/admin/src/lib/admin-panel/pages/users/users.component.ts
```diff
@@ -73,13 +63,9 @@ export class UsersComponent implements OnInit {
...
-        userOrgRole: org ? await this.orgService.getMemberRole(org, u.uid) : undefined,
-        type: org ? (getOrgModuleAccess(org).includes('dashboard') ? 'seller' : 'buyer') : undefined
```
phpgeekfr, line 78: this line is not querying db, why not keeping it ? 
just to have less collumns ?

phpgeekfr (approved):
LGTM but look at my comments.
Also, If you can fix the search by org name it would be cool

libs/admin/src/lib/admin-panel/pages/users/users.component.ts
RemcoSimonides: That's right ;) I told Marie these fields do not impact the performance but she preferred the cleaner view :)

PR #3618, opened 2020-09-03
fix(anonymize-db.ts): first draft for consistent user/org anonymisation

george43g:
- fixed bugs
- consistent users
- attempting consistent orgs, introduced types
- looking for feedback and testing


tools/scripts/anonymize-db.ts
```diff
@@ -1,54 +1,115 @@
...
+import { User } from '@blockframes/user/types';
+import { Organization } from '@blockframes/organization/+state';
+import { NotificationDocument } from '@blockframes/notification/types';
 
-const userTable: [string, string][] = []; // uid, newEmail
+const userTable: [string, User][] = []; // uid, newEmail
+const orgTable: [string, Organization][] = []; // uid, newOrg
+
+function isA<T>(doc: any, ...keys: string[]): doc is T {
+  for (const key of keys) if (!(key in doc)) return false;
```
GrandSchtroumpf, line 13: you can do :
```typescript
return keys.every(key => key in doc);
```

tools/scripts/anonymize-db.ts
```diff
@@ -1,54 +1,115 @@
...
 import { resolve } from 'path';
+import { User } from '@blockframes/user/types';
+import { Organization } from '@blockframes/organization/+state';
+import { NotificationDocument } from '@blockframes/notification/types';
 
-const userTable: [string, string][] = []; // uid, newEmail
+const userTable: [string, User][] = []; // uid, newEmail
+const orgTable: [string, Organization][] = []; // uid, newOrg
+
+function isA<T>(doc: any, ...keys: string[]): doc is T {
```
GrandSchtroumpf, line 12: Maybe rename it `hasKeys`. I understand the `isAs` refering to the type guard, but I think `hasKeys` describe more the javascript underlying logic.
Also, I would use a real array instead of spread operator.

tools/scripts/anonymize-db.ts
```diff
@@ -1,54 +1,115 @@
...
+  for (const key of keys) if (!(key in doc)) return false;
+  return true;
+}
+
+function assertType<T>(doc: any, ...keys: string[]): asserts doc is T {
+  for (const key of keys) if (!(key in doc)) throw Error('Wrong document type found!');
+}
+
+// function isOrg(doc: any): doc is Organization {
```
GrandSchtroumpf, line 21: Remove if not used.

tools/scripts/anonymize-db.ts
```diff
@@ -1,54 +1,115 @@
...
+const userTable: [string, User][] = []; // uid, newEmail
+const orgTable: [string, Organization][] = []; // uid, newOrg
+
+function isA<T>(doc: any, ...keys: string[]): doc is T {
+  for (const key of keys) if (!(key in doc)) return false;
+  return true;
+}
+
+function assertType<T>(doc: any, ...keys: string[]): asserts doc is T {
+  for (const key of keys) if (!(key in doc)) throw Error('Wrong document type found!');
```
GrandSchtroumpf, line 18: Same as above : 
```typescript
function assertType<T>(doc: any, ...keys: string[]): asserts doc is T {
  return !isA(doc, keys);
}
```

tools/scripts/anonymize-db.ts
```diff
@@ -1,54 +1,115 @@
...
+//   console.log(keys);
+//   for (const key of keys) if (!doc.hasOwnProperty(key)) return false;
+//   return true;
+// }
+
+function updateUser(user: User) {
+  const foundUserTuple = userTable.find(([uid]) => user.uid === uid);
+  if (foundUserTuple) {
+    // Already set this user previously
+    const [, newUser] = foundUserTuple;
```
GrandSchtroumpf, line 39: Use "_" if you don't need a parameter.

tools/scripts/anonymize-db.ts
```diff
@@ -1,54 +1,115 @@
+// tslint:disable: no-console
 import * as faker from 'faker';
 import { readFileSync, writeFileSync } from 'fs';
 import { resolve } from 'path';
+import { User } from '@blockframes/user/types';
+import { Organization } from '@blockframes/organization/+state';
+import { NotificationDocument } from '@blockframes/notification/types';
 
-const userTable: [string, string][] = []; // uid, newEmail
+const userTable: [string, User][] = []; // uid, newEmail
```
GrandSchtroumpf, line 9: Use records for that, it's more natural for Javascript than matrix : 
```typescript
const newUsers: Record<string, User> = {};
const newOrgs: Record<string, User> = {};
```

tools/scripts/anonymize-db.ts
```diff
@@ -1,54 +1,115 @@
...
+//   for (const key of keys) if (!doc.hasOwnProperty(key)) return false;
+//   return true;
+// }
+
+function updateUser(user: User) {
+  const foundUserTuple = userTable.find(([uid]) => user.uid === uid);
+  if (foundUserTuple) {
+    // Already set this user previously
+    const [, newUser] = foundUserTuple;
+    Object.assign(user, newUser);
```
GrandSchtroumpf, line 40: I find it difficult to read. Object.assign is usually used to do shallow copy, but here you're using it to mutate the user. 
To keep separation of concern I would have done : 
```typescript
function updateUser(user: User) {
  if (!newUsers[user.uid]) {
    newUsers[user.uid] = fakeUser(user);
  }
  return { ...user, ...newUsers[user.uid] };
}

function fakeUser(user: Partial<User> = {}) {
  const { firstName, lastName } = faker.name;
  const email = fakeEmail(firstName);
  return {
    ...user,
    firstName,
    lastName,
    email
  }
}

// Now you can reuse it for updating the org
function fakeEmail(name: string) {
  const random = Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 3);
  return `dev+${name.replace(/\W/g, '').toLowerCase()}-${random}@cascade8.com`;
}
```

tools/scripts/anonymize-db.ts
```diff
@@ -1,54 +1,115 @@
...
+    const email = `dev+${firstName.replace(/\W/g, '').toLowerCase()}-${Math.random()
+      .toString(36)
+      .replace(/[^a-z]+/g, '')
+      .substr(0, 3)}@cascade8.com`;
+    Object.assign(user, { firstName, lastName, email });
+    userTable.push([user.uid, user]);
+  }
+}
+
+function updateOrg(org: Organization) {
```
GrandSchtroumpf, line 54: Same as above.

GrandSchtroumpf (changes requested):

tools/scripts/anonymize-db.ts
george43g: an assertion function must only throw if it fails, they are a return type `void`

GrandSchtroumpf (approved):

tools/scripts/anonymize-db.ts
```diff
@@ -1,59 +1,148 @@
...
+    .toString(36)
+    .replace(/[^a-z]+/g, '')
+    .substr(0, 3)}@cascade8.com`;
+
+function hasKeys<T extends object>(doc: object, ...keys: (keyof T)[]): doc is T {
+  return keys.every((key) => key in doc);
+}
+
+function assertType<T extends object>(doc: any, ...keys: (keyof T)[]): asserts doc is T {
+  if (hasKeys<T>(doc, ...keys)) throw Error('WRONG OBJECT TYPE!');
```
GrandSchtroumpf, line 31: I think it's `if (!hasKeys(...))` isn't it ?

tools/scripts/anonymize-db.ts
```diff
@@ -1,59 +1,148 @@
...
+}
+
+function processOrg<T extends Organization | PublicOrganization>(o: T): T {
+  const companyName = faker.company.companyName();
+  const denomination = { full: companyName, public: companyName };
+  const email = fakeEmail(companyName);
+  return { ...o, denomination, email };
+}
+
+function processInvitation(i: Invitation): Invitation {
```
GrandSchtroumpf, line 48: You can do `function processInvitation(i: Partial<Invitation> = {}): Invitation {`

tools/scripts/anonymize-db.ts
```diff
@@ -1,59 +1,148 @@
...
+
+function processNotification(n: NotificationDocument): NotificationDocument {
+  return {
+    ...n,
+    organization: updateOrg(n.organization),
+    user: updateUser(n.user as PublicUser), // ! TODO: If possible don't use Partial in this type!
+  };
+}
+
+function updateUser(user: User | PublicUser) {
```
GrandSchtroumpf, line 66: You might want to do `updateUser(user: Partial<User> | Partial<PublicUser>)`

tools/scripts/anonymize-db.ts
```diff
@@ -1,59 +1,148 @@
...
+  if (!user) return;
+  if (hasKeys<PublicUser>(user, 'uid') && !hasKeys<User>(user, 'watermark')) { // Is public
+    return createPublicUser( newUsers?.[user.uid] || (newUsers[user.uid] = processUser(user)));
+    // If not in cache, process the user, write to cache, make it a publicUser and return it
+  } else if (hasKeys<User>(user, 'uid')) {
+    return newUsers?.[user.uid] || (newUsers[user.uid] = processUser(user));
```
GrandSchtroumpf, line 72: Quite compact and difficult to read. Many do it into 2 lines.

tools/scripts/anonymize-db.ts
```diff
@@ -1,59 +1,148 @@
...
-  if (docPath.includes('notifications/')) {
-    updateEmail(content.fromUser);
-    updateEmail(content.toUser);
-    updateEmail(content.user);
+  throw Error(`Unable to process org: ${JSON.stringify(org, null, 4)}`);
+}
+
+function anonymizeDocument({ docPath, content: doc }: JsonlDbRecord) {
+  // function pathHasAny(path: string) { return docPath.includes(path) }
+  function pathHasAny(...paths: string[]) {
```
GrandSchtroumpf, line 90: Remove the spread operator here. It makes things difficult to understand.

tools/scripts/anonymize-db.ts
```diff
@@ -1,59 +1,148 @@
...
+    return !paths.every((path) => !docPath.includes(path));
```
GrandSchtroumpf, line 91: Avoid double negation. Using `some` instead of `every` might help.

tools/scripts/anonymize-db.ts
```diff
@@ -1,59 +1,148 @@
...
+    'contracts/',
+    'docsIndex/',
+    'events/',
+    'movies/',
+    'permissions/',
+    'publicContracts/',
+  ];
+  if (!doc || pathHasAny(...ignorePaths)) return { docPath, content: doc };
+
+  try {
```
GrandSchtroumpf, line 105: It's not super clear what happens here. Why not just `docPath.includes('users/')` or `docPath.startWith('users/')` ?

tools/scripts/anonymize-db.ts
```diff
@@ -1,59 +1,148 @@
+// tslint:disable: no-console
```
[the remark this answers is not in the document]
george43g: There is a console.time()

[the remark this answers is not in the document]
GrandSchtroumpf: Ok

tools/scripts/anonymize-db.ts
george43g: This creates an error because the type that's returned is no longer guaranteed to be an invitation, so it would have to return a Partial<Invitation> as well.

tools/scripts/anonymize-db.ts
george43g: same as above

PR #3622, opened 2020-09-04
fix(deploy-functions-config.ts): deploy was not working - added doten…

george43g:
…v.config() to top

fast fix

- fixed functions deploy (config)

### DIFF
--- a/tools/scripts/deploy-function-config.ts
+++ b/tools/scripts/deploy-function-config.ts
@@ -1,28 +1,10 @@
 import 'tsconfig-paths/register';
+import { config } from 'dotenv';
+config();
+
 import * as firebaseTools from 'firebase-tools';
 import { loadSecretsFile } from './lib';
 import { warnMissingVars } from '@blockframes/firebase-utils';
-
-if (!process.env.SENDGRID_API_KEY || !process.env.ALGOLIA_API_KEY) {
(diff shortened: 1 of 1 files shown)


tools/scripts/deploy-function-config.ts
```diff
@@ -41,3 +23,28 @@ function getKeyValFormat(): string[] {
...
+    // Env config values probably doesn't exist in env
+    console.warn(
+      'PLEASE UPDATE TO THE NEW .env FORMAT! Please run "npm run migrate:deploy-secrets"'
+    );
+    loadSecretsFile();
+  }
+  // * Check if we are in CI
+  const FIREBASE_CONFIG: any = {};
+  if (process.env.FIREBASE_CI_TOKEN) FIREBASE_CONFIG.token = process.env.FIREBASE_CI_TOKEN;
+  process.stdout.write('Existing Firebase Functions Config Values:\n');
```
GrandSchtroumpf, line 38: any reason to mix stdout.write & console.log ?

GrandSchtroumpf (approved):

tools/scripts/deploy-function-config.ts
george43g: I don't think the reason for it was a big deal - I'll change it as part of my next PR

PR #3627, opened 2020-09-04
Missing data in crm columns

phpgeekfr:
Since the merge of movie-form, it looks like that bf table component is not handled dot in path :

example ```'storeConfig.status':```


phpgeekfr:
I  think this is also why search by orgName or userName in filters does not work

libs/organization/src/lib/organization/components/member-repertory/member-repertory.component.ts
```diff
@@ -30,10 +30,6 @@ export class MemberRepertoryComponent {
 
   @Input() memberColumnsIndex = ['firstName', 'lastName', 'email', 'position', 'role'];
 
-  get initialColumns() {
-    return this.isSuperAdmin ? [...this.memberColumnsIndex, 'uid'] : this.memberColumnsIndex;
```
RemcoSimonides, line 34: when `isSuperAdmin` the 'uid' field was added. However, you already pass this uid via the `@Input() memberColumnsIndex`. Caused an error where there were two of the same id's for matSort

libs/admin/src/lib/admin-panel/pages/users/users.component.ts
```diff
@@ -76,10 +79,9 @@ export class UsersComponent implements OnInit {
       'firstName',
       'lastName',
       'email',
-      'position',
-      'org.denomination.full',
-      'org.denomination.public'
+      'position'
     ];
+    if (data.org?.denomination?.full) columnsToFilter.push('org.denomination.full', 'org.denomination.public');
```
RemcoSimonides, line 84: Sooo this is ugly... but in the demo data I have many users without an organization which cause an error when trying to filter.

Another option was to edit `getValue(data, c)` to return empty value when path in object is not found

RemcoSimonides:
I've flattened the data sent to bf-table-filter. Without this I couldn't get the matSort to work.
I haven't flattened org on user and movie, so we can't sort on that. Could still do it if it's wanted

libs/organization/src/lib/organization/components/member-repertory/member-repertory.component.ts
phpgeekfr: I reverted this method because the component is used outside of CRM and need 'uid' column. I updated the admin component so we don't have the same id

libs/admin/src/lib/admin-panel/components/guest-table/guest-table.component.ts
```diff
@@ -10,42 +10,44 @@ import { getValue } from '@blockframes/utils/helpers';
...
     'status',
-    'guest.email',
+    'guestEmail',
   ];
 
-  @Input() set invitations(invitations: InvitationDetailed[]) {
+  @Input() set invitations(invitations: any[]) {
     if (invitations) {
       this._invitations = invitations.map(i => {
         const invitation = { ...i } as InvitationDetailed;
```
GrandSchtroumpf, line 47: You can remove that, as the map operator is creating a new instance of the invitation.
Also as a sidenote, it's better to avoid shallow copying as deep keys are need copied.

libs/admin/src/lib/admin-panel/components/guest-table/guest-table.component.ts
```diff
@@ -10,42 +10,44 @@ import { getValue } from '@blockframes/utils/helpers';
...
-        invitation.guest = getGuest(invitation, 'user');
+        invitation.guestEmail = getGuest(invitation, 'user').email;
```
GrandSchtroumpf, line 48: Why do you need to flatten the `guest` object ?

libs/admin/src/lib/admin-panel/pages/users/users.component.ts
```diff
@@ -55,7 +55,10 @@ export class UsersComponent implements OnInit {
     const rows = users.map(async u => {
       const org = u.orgId ? await this.getOrg(u.orgId) : undefined;
       return {
-        ...u,
```
RemcoSimonides, line 58: We could use `...u` again, right?

phpgeekfr: Yes but the new way is also cool because we append only the data we need for the table

libs/admin/src/lib/admin-panel/components/guest-table/guest-table.component.html
```diff
@@ -43,7 +43,7 @@
         <img class="poster" [ref]="poster" asset="empty_poster.webp">
       </ng-container>
       <a [routerLink]="['/c/o/admin/panel/movie', movie.id]">
-        <span>{{ movie.title.international }} ({{ movie.internalRef }}) </span><mat-icon svgIcon="eye"></mat-icon>
+        <span>{{ movie.title.international }} {{ movie.internalRef ? '(' + movie.internalRef + ')': ''}} </span><mat-icon svgIcon="eye"></mat-icon>
```
GrandSchtroumpf, line 46: Does this appends a lot, or is it only here ?
If it appends a lot, would it make sense to create a pipe : `{{ movie | title }}` ?

libs/admin/src/lib/admin-panel/pages/users/users.component.ts
GrandSchtroumpf: Yes

GrandSchtroumpf (approved):

libs/admin/src/lib/admin-panel/components/guest-table/guest-table.component.html
phpgeekfr: I think this is only here

PR #3630, opened 2020-09-07
Added ability to remove user from CRM

RemcoSimonides:
To do in issue #3592 

"we need to add a delete button (+ confirmation dialog) on users. The delete should delete the user and all corresponding entry in the db (id in org userIds, invitations, notifications,...)"

### DIFF
--- a/apps/backend-functions/src/users.ts
+++ b/apps/backend-functions/src/users.ts
@@ -191,6 +191,11 @@ export async function onUserDelete(
 
   // delete events owned by the user?
 
+  // delete media from storage
+  const bucket = admin.storage().bucket(getStorageBucketName());
+  if (!!user.avatar) bucket.file(user.avatar).delete();
+  if (!!user.watermark) bucket.file(user.watermark).delete();
+
 }
 
 export const sendDemoRequest = async (data: RequestDemoInformations): Promise<RequestDemoInformations> => {
--- a/apps/backend-functions/src/users.ts
+++ b/apps/backend-functions/src/users.ts
@@ -188,14 +188,12 @@ export async function onUserDelete(
   notificationsSnap.forEach(notification => db.doc(`notifications/${notification.id}`).delete());
 
   // delete blockframesAdmin doc?
-
-  // delete events owned by the user?
+  db.doc(`blockframesAdmin/${user.uid}`).delete();
 
   // delete media from storage
   const bucket = admin.storage().bucket(getStorageBucketName());
   if (!!user.avatar) bucket.file(user.avatar).delete();
   if (!!user.watermark) bucket.file(user.watermark).delete();
-
 }
 
 export const sendDemoRequest = async (data: RequestDemoInformations): Promise<RequestDemoInformations> => {
--- a/apps/backend-functions/src/users.ts
+++ b/apps/backend-functions/src/users.ts
@@ -187,7 +187,7 @@ export async function onUserDelete(
   const notificationsSnap = await notificationsRef.get();
   notificationsSnap.forEach(notification => db.doc(`notifications/${notification.id}`).delete());
 
-  // delete blockframesAdmin doc?
+  // delete blockframesAdmin doc
   db.doc(`blockframesAdmin/${user.uid}`).delete();
 
   // delete media from storage
--- a/apps/backend-functions/src/users.ts
+++ b/apps/backend-functions/src/users.ts
@@ -3,10 +3,10 @@ import * as functions from 'firebase-functions';
 import { db, getStorageBucketName } from './internals/firebase';
 import { userResetPassword, sendDemoRequestMail, sendContactEmail, accountCreationEmail, userInvite } from './templates/mail';
 import { sendMailFromTemplate, sendMail } from './internals/email';
-import { RequestDemoInformations, PublicUser } from './data/types';
+import { RequestDemoInformations, PublicUser, PermissionsDocument, OrganizationDocument, InvitationDocument } from './data/types';
 import { storeSearchableUser, deleteObject } from './internals/algolia';
 import { algolia } from './environments/environment';
-import { upsertWatermark } from '@blockframes/firebase-utils';
+import { upsertWatermark, getCollection } from '@blockframes/firebase-utils';
 import { getDocument, getFromEmail } from './data/internals';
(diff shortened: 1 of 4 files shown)


RemcoSimonides:
- [ ] Navigate user to previous page after deleting the user?

apps/backend-functions/src/users.ts
```diff
@@ -148,8 +148,49 @@ export async function onUserDelete(
...
+  
+  // remove all notifications related to user
+  const notificationsRef = db.collection(`notifications`).where('toUserId', '==', user.uid);
+  const notificationsSnap = await notificationsRef.get();
+  notificationsSnap.forEach(notification => db.doc(`notifications/${notification.id}`).delete());
+
+  // delete blockframesAdmin doc?
+
+  // delete events owned by the user?
+
```
pldespaigne, line 193: maybe you also need to delete every files in the firebase storage under the path `users/<USER_ID>/` such as avatar & watermark

apps/backend-functions/src/users.ts
```diff
@@ -148,8 +148,54 @@ export async function onUserDelete(
...
+  // remove permissions
+  if (!!user.orgId) {
+    const permissionsRef = db.doc(`permissions/${user.orgId}`);
+    const permissionsDoc = await permissionsRef.get();
+    const permissions = permissionsDoc.data() as PermissionsDocument;
+    const roles = permissions.roles;
+    delete roles[user.uid];
+    permissionsRef.update({roles}); 
+  }
```
phpgeekfr, line 176: You cannot use existing `removeMember` (libs/organization/src/lib/organization/+state/organization.service.ts) for this ?

apps/backend-functions/src/users.ts
```diff
@@ -148,8 +148,54 @@ export async function onUserDelete(
...
+
+  // remove all invitations related to user
+  const invitations = await getCollection<InvitationDocument>(`invitations`)
+  invitations
+    .filter(invitation => invitation.fromUser?.uid === user.uid || invitation.toUser?.uid === user.uid)
+    .map(invitation => invitation.id)
+    .map(id => db.doc(`invitations/${id}`).delete());
```
phpgeekfr, line 183: You can move this to InvitationService

apps/backend-functions/src/users.ts
```diff
@@ -148,8 +148,54 @@ export async function onUserDelete(
...
+  
+  // remove all notifications related to user
+  const notificationsRef = db.collection(`notifications`).where('toUserId', '==', user.uid);
+  const notificationsSnap = await notificationsRef.get();
+  notificationsSnap.forEach(notification => db.doc(`notifications/${notification.id}`).delete());
```
phpgeekfr, line 188: You can move this to NotificationService

apps/backend-functions/src/users.ts
```diff
@@ -148,8 +148,54 @@ export async function onUserDelete(
...
+
+  // delete blockframesAdmin doc?
```
phpgeekfr, line 190: yes

apps/backend-functions/src/users.ts
```diff
@@ -148,8 +148,54 @@ export async function onUserDelete(
...
+
+  // delete events owned by the user?
```
phpgeekfr, line 192: There should not be any events directly linked to user currently (only orgId).

apps/backend-functions/src/users.ts
phpgeekfr: Maybe add a warning before deleting ?  Saying that all invitation will be removed

apps/backend-functions/src/users.ts
phpgeekfr: But (for the future) maybe add a warning before deleting ?  Saying that all event will be removed

libs/admin/src/lib/admin-panel/pages/user/user.component.ts
```diff
@@ -100,4 +105,23 @@ export class UserComponent implements OnInit {
     await this.adminService.sendPasswordResetEmail(this.user.email);
     this.snackBar.open(`Reset password email sent to : ${this.user.email}`, 'close', { duration: 2000 });
   }
+
+  public async deleteUser() {
+    // check if user is not last superAdmin
+    const permissions = await this.permissionService.getValue(this.user.orgId);
+    if (!!this.user.orgId && !this.permissionService.hasLastSuperAdmin(permissions, this.userId, 'member')) {
+      throw new Error('There must be at least one Super Admin in the organization.');
+    }
```
phpgeekfr, line 114: This is already checked in `removeMember` maybe you can update it a bit so it fit your needs

phpgeekfr (approved):
LGTM but look at comments

apps/backend-functions/src/users.ts
RemcoSimonides: It was used as an inspiration ;)

libs/admin/src/lib/admin-panel/pages/user/user.component.ts
```diff
@@ -132,4 +138,25 @@ export class UserComponent implements OnInit {
...
+
+    this.dialog.open(ConfirmComponent, {
+      data: {
+        title: 'Are you sure you want to delete this user?',
+        question: 'this action is irreversable',
```
phpgeekfr, line 152: Typo on irreversable ?

phpgeekfr (approved):

apps/backend-functions/src/users.ts
```diff
@@ -148,8 +148,52 @@ export async function onUserDelete(
...
+    .filter(invitation => invitation.fromUser?.uid === user.uid || invitation.toUser?.uid === user.uid)
+    .map(invitation => invitation.id)
+    .map(id => db.doc(`invitations/${id}`).delete());
+  
+  // remove all notifications related to user
+  const notificationsRef = db.collection(`notifications`).where('toUserId', '==', user.uid);
+  const notificationsSnap = await notificationsRef.get();
+  notificationsSnap.forEach(notification => db.doc(`notifications/${notification.id}`).delete());
+
+  // delete blockframesAdmin doc?
```
phpgeekfr, line 190: you can remove "?"

PR #3636, opened 2020-09-08
George/prepare for testing

george43g:
- rewrote userSync to be much faster/more stable
- helper functions in testing lib for firebase auth
- draft: working on shorter sub-chains of prepareForTesting for devX
- todo: need to add a better error message to anon script fail
- todo: improve algolia stability
- improve overall stability of tool


apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -70,8 +70,22 @@ export async function prepareForTesting() {
   process.exit(0);
 }
 
+export async function prepareDb() {
```
george43g, line 73: This section is still a draft -

apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -70,8 +70,22 @@ export async function prepareForTesting() {
...
+  const { db, auth, storage, getCI } = loadAdminServices();
+
+  const p1 = copyDbFromCi(storage, getCI());
+  const p3 = p1.then(() => restore(appUrl.content));
+  const p4 = p3.then(() => migrate(false));
+  const p5 = p4.then(() => cleanDeprecatedData(db, auth));
+  const p6 = p5.then(generateFixtures)
+  return Promise.all([p1, p2, p3, p4, p5, p6, p8]);
```
GrandSchtroumpf, line 81: You shouldn't use Promise.all on promises that are depending on each other.
A simple `try / catch` with await should do the job : 
```typescript
try {
  await copyDbFromCi(storage, getCI());
  await restore(appUrl.content);
  await migrate(false);
  await cleanDepreactedDate(db, auth);
  await generateFixtures();
} catch (err) {
  throw new Error(err);
}
```

apps/backend-ops/src/users.ts
```diff
@@ -100,8 +101,10 @@ export async function syncUsers(db: JsonlDbRecord[]): Promise<any> {
   const { auth } = loadAdminServices();
 
   const expectedUsers = readUsersFromJsonlFixture(db);
-  await removeUnexpectedUsers(expectedUsers, auth);
-  await createAllUsers(expectedUsers, auth);
+  const deleteResult = await deleteAllUsers(auth);
+  const createResult = await importAllUsers(auth, expectedUsers);
+  console.log(deleteResult);
```
GrandSchtroumpf, line 106: Remove console.log if you don't need.

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -0,0 +1,60 @@
+// tslint:disable: no-console
+import type * as admin from 'firebase-admin';
+
+export async function deleteAllUsers(auth: admin.auth.Auth) {
+  const allUsers: admin.auth.UserRecord[] = [];
+  let result: any = { pageToken: undefined };
```
GrandSchtroumpf, line 6: Don't type any, create a type.

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  console.log('Loading users now...');
+  let timeMsg = 'Loading users took';
+  console.time(timeMsg);
+  do {
```
GrandSchtroumpf, line 11: Could you remove them here ? It would me less difficult to read : 
```typescript
do {
  result = await auth.listUsers(1000, result.pageToken);
  auth.deleteUsers(result.users)
} while (result.pageToken);
```
The current implementation looks a little bit overkill for me.

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -0,0 +1,60 @@
...
+    const { errors, failureCount, successCount } = deletion;
+    output.errors.push(errors);
+    output.successCount += successCount;
+    output.failureCount += failureCount;
+  }
+  console.timeEnd(timeMsg);
+  return output;
+}
+
+export async function importAllUsers(auth: admin.auth.Auth, users: admin.auth.UserImportRecord[]) {
```
GrandSchtroumpf, line 40: `firebase-tools` has a method for that: https://github.com/firebase/firebase-tools#auth-commands
Could you use it ?

george43g: I am using it :)

george43g: In that exact code. It just has a maximum batch size is all.

Once we upgrade to using just the CLI tools, however, might be worth switching to `firebase-tools` commands - however these preserve passwords and hashes for users. Good to backup actual auth on Prod - should make a new issue out of this

george43g:
todo: need to add a better error message to anon script fail - still TODO
todo: improve algolia stability - DONE
improve overall stability of tool - DONE

apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -37,7 +37,7 @@ export async function prepareForTesting() {
   await restoreStorageFromCi(getCI());
   console.info('Storage synced!');
 
-  console.info('Preparing the database...');
+  console.info('Preparing the database...'); // ! Also updates storage
```
GrandSchtroumpf, line 40: What does this comment mean ? If it's a todo add a tag to an issue. If it's supposed to be in the console.info, put it there. Else maybe remove it.

apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -70,25 +70,37 @@ export async function prepareForTesting() {
...
+  await restore(appUrl.content);
+  console.log('Anonymized DB restored. Migrating...');
+  await migrate(false);
+  console.log('DB migration complete. Cleaning up...');
+  await cleanDeprecatedData(db, auth);
+  console.log('Deprecated data removed!');
 }
 
+// /**
+//  * This is an experimental function to see if we can speed this process up
```
GrandSchtroumpf, line 87: If it doesn't speed up the process, remove it.

apps/backend-ops/src/users.ts
```diff
@@ -202,7 +205,7 @@ export async function generateWatermarks() {
     } else {
       await user.ref.update({ watermark: file.name });
     }
-  }, watermarkRowConcurrency);
+  }, process.env?.HEAVY_CHUNK_SIZE && Number(process.env.HEAVY_CHUNK_SIZE));
```
GrandSchtroumpf, line 208: Use env.ts

libs/firebase-utils/src/lib/firebase-utils.ts
```diff
@@ -20,12 +20,12 @@ export function getCollection<T>(path: string): Promise<T[]> {
   return getCollectionRef(path).then(collection => collection.docs.map(doc => doc.data() as T));
 }
 
-export async function runChunks(rows: any[], cb: any, rowsConcurrency = 10, verbose = true) {
-  const chunks = chunk(rows, rowsConcurrency);
+export async function runChunks(rows: any[], cb: any, rowsConcurrency? : number, verbose = true) {
+  const chunks = chunk(rows, rowsConcurrency || ( process.env.CHUNK_SIZE && Number(process.env.CHUNK_SIZE) ) || 10);
```
GrandSchtroumpf, line 24: Same here

libs/firebase-utils/src/lib/util.ts
```diff
@@ -4,6 +4,48 @@ import { firebase as firebaseCI } from 'env/env.ci';
...
+export async function runInBatches<K = any>(batch: K[], cb: (p: K) => Promise<any>, chunkSize = 10) {
+  function* batchGenerator() {
+    const chunks = chunk(batch, chunkSize);
+    while (chunks.length > 0) {
+      console.log(`Operations remaining: ${chunks.length * chunkSize}/${batch.length}`);
+      yield Promise.all(chunks.pop().map(cb))
+    }
+  }
+  const batchIterator = batchGenerator();
+  for await (let i of batchIterator) {
```
GrandSchtroumpf, line 18: I don't get why you do that. Why you need an await or update the i...

libs/firebase-utils/src/lib/util.ts
```diff
@@ -4,6 +4,48 @@ import { firebase as firebaseCI } from 'env/env.ci';
 import { config } from 'dotenv';
 import { readFileSync } from 'fs';
 import requiredVars from 'tools/mandatory-env-vars.json';
+import { chunk } from 'lodash';
+
+export async function runInBatches<K = any>(batch: K[], cb: (p: K) => Promise<any>, chunkSize = 10) {
```
GrandSchtroumpf, line 9: It's quite difficult for me to understand this part. I'm not super fluent in generator function. I usually use for loop instead of while + generator. But maybe your way is more efficient, happy to discuss about that.

phpgeekfr (approved):
LGTM

libs/firebase-utils/src/lib/util.ts
```diff
@@ -4,6 +4,48 @@ import { firebase as firebaseCI } from 'env/env.ci';
...
+  function* batchGenerator() {
```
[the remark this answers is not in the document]
GrandSchtroumpf: It's a generator function : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*
@george43g That's basically my point on the other comment. If it's difficult to read for Bruce & I, it's going to be difficult to read for everybody in the team. Not that generator function is a bad choice, but it requires more knowledge & mental energy to understand what it does.

libs/firebase-utils/src/lib/util.ts
george43g: this is special syntax for async iterators. Basically every iteration of the loop awaits for a promise to resolve.

regarding the i, it's an ugly hack. if we don't use i, we get an error of unused var.

libs/firebase-utils/src/lib/util.ts
george43g: I don't want to spend too much time on this, but basically I didn't end up using this one - but I think it will be useful for future. I changed it to just return the iterator, which should allow for some very neat code down the line.

apps/backend-ops/src/algolia.ts
```diff
@@ -25,14 +25,16 @@ export async function upgradeAlgoliaOrgs() {
   await clearIndex(algolia.indexNameOrganizations, process.env['ALGOLIA_API_KEY']);
 
   const { db } = loadAdminServices();
-  const orgs = await db.collection('orgs').get();
+  const orgsIterator = getCollectionInBatches<OrganizationDocument>(db.collection('orgs'), 'id', 300)
+  for await (const orgs of orgsIterator) {
```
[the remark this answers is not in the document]
george43g: yeah it's an iterator/generator

libs/firebase-utils/src/lib/util.ts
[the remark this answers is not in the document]
george43g: you were right, the purpose was to have a type safe version. I did not want to update yours with types because it actually breaks some things. we can migrate. Also, I changed it anyway to just return the iterator, so you can now both specify the callback function and also what to do with its return value, synchronously.

libs/firebase-utils/src/lib/util.ts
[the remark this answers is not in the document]
george43g: generator function. If it's too esoteric happy to use a loop, it's just a bit neater sometimes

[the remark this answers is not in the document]
GrandSchtroumpf: Let's keep this one. Just for next time prefer loops when possible.

GrandSchtroumpf (approved):

PR #3642, opened 2020-09-09
Added ability to update organization id

RemcoSimonides:
To do in issue #3592 

"Move user from one org to another without invitation"

Notifications are still on since they can't be disabled


phpgeekfr (approved):
LGTM

PR #3645, opened 2020-09-09
[DO NOT MERGE] watermark fallback

pldespaigne:
When a user ask for the screening url, in the backend, we check if he has a watermark, and if not we generate it to prevent error on the screening.

### DIFF
--- a/apps/backend-functions/src/player.ts
+++ b/apps/backend-functions/src/player.ts
@@ -1,13 +1,14 @@
 import { CallableContext } from 'firebase-functions/lib/providers/https';
-import { db, admin } from './internals/firebase';
+import { db, admin, getStorageBucketName } from './internals/firebase';
 import { EventDocument, EventMeta, linkDuration } from '@blockframes/event/+state/event.firestore';
 import { isUserInvitedToScreening } from './internals/invitations/events';
-import { MovieDocument } from './data/types';
+import { MovieDocument, PublicUser } from './data/types';
 import { jwplayerSecret, jwplayerKey } from './environments/environment';
 import { createHash } from 'crypto';
 import { firestore } from 'firebase'
 import { getDocument } from './data/internals';
 import { ErrorResultResponse } from './utils';
+import { upsertWatermark } from '@blockframes/firebase-utils';
 
 // No typing
 const JWPlayerApi = require('jwplatform');
@@ -92,6 +93,44 @@ export const getPrivateVideoUrl = async (
       error: 'NO_INVITATION',
       result: `You have not been invited to see this movie`
     };
+  }
+
+  // watermark fallback : in case the user's watermark doesn't exist we generate it
+  const userRef = db.collection('users').doc(context.auth.uid);
+  const userSnap = await userRef.get();
+  const user = userSnap.data() as PublicUser;
+
+  // if (!user.watermark) { // TODO use this line with the new image data model
+  if (!user.watermark || !user.watermark.ref || !user.watermark.url) {
+
+    upsertWatermark(user, getStorageBucketName());
+
+    // wait for the function to update the user document after watermark creation
+    const success = await new Promise(resolve => {
+
+      const unsubscribe = userRef.onSnapshot(snap => {
+        const userData = snap.data() as PublicUser;
+
+        // if (!!userData.watermark) { // TODO use this line with the new image data model
+        if (!!userData.watermark && !!userData.watermark.ref && !!userData.watermark.url) {
+          unsubscribe();
+          resolve(true);
+        }
+      });
+
+      // timeout after 10s
+      setTimeout(() => {
+        unsubscribe();
+        resolve(false);
+      }, 10000);
+    });
+
+    if (!success) {
+      return {
+        error: 'WATERMARK_CREATION_TIMEOUT',
+        result: `The watermark creation has timeout, please try again later`
+      };
+    }
   }
 
   // we need expiry date in UNIX Timestamp (aka seconds), JS Date give use milliseconds,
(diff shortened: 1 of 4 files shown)


phpgeekfr (approved):
LGTM.

- Look if 10 sec is enough
- Update this PR to make it on develop 
- Also make this PR against master
- Wait for Marie approval before merge !

pldespaigne:
- [x] checked on local 10sec is enough
- [x] checked screening is working with the watermark fallback
- [ ] create PR against master
- [ ] create PR on develop (with new image data model)

apps/backend-functions/src/player.ts
```diff
@@ -94,6 +95,44 @@ export const getPrivateVideoUrl = async (
     };
   }
 
+  // watermark fallback : in case the user's watermark doesn't exist we generate it
+  const userRef = db.collection('users').doc(context.auth.uid);
+  const userSnap = await userRef.get();
+  const user = userSnap.data() as PublicUser;
```
GrandSchtroumpf, line 101: You can use 
```typescript
const user = await getDoc<PlublicUser>(`users/${contet.auth.uid}`)
```

apps/backend-functions/src/player.ts
```diff
@@ -94,6 +95,44 @@ export const getPrivateVideoUrl = async (
...
+
+  // if (!user.watermark) { // TODO use this line with the new image data model
```
GrandSchtroumpf, line 103: Put an issue tag in the todo

apps/backend-functions/src/player.ts
```diff
@@ -94,6 +95,44 @@ export const getPrivateVideoUrl = async (
...
+  if (!user.watermark || !user.watermark.ref || !user.watermark.url) {
+
+    upsertWatermark(user, getStorageBucketName());
```
GrandSchtroumpf, line 106: no `await` here ?

apps/backend-functions/src/player.ts
```diff
@@ -94,6 +95,44 @@ export const getPrivateVideoUrl = async (
...
+
+    // wait for the function to update the user document after watermark creation
+    const success = await new Promise(resolve => {
+
+      const unsubscribe = userRef.onSnapshot(snap => {
+        const userData = snap.data() as PublicUser;
+
+        // if (!!userData.watermark) { // TODO use this line with the new image data model
+        if (!!userData.watermark && !!userData.watermark.ref && !!userData.watermark.url) {
```
GrandSchtroumpf, line 115: `if (userData.watermark?.ref && userData.watermark.url)`
Create a function: 
```typescript
const hasWatermark = (user: PublicUser) => userData.watermark?.ref && userData.watermark.url
```

apps/backend-functions/src/player.ts
```diff
@@ -94,6 +95,44 @@ export const getPrivateVideoUrl = async (
...
+  const userSnap = await userRef.get();
+  const user = userSnap.data() as PublicUser;
+
+  // if (!user.watermark) { // TODO use this line with the new image data model
+  if (!user.watermark || !user.watermark.ref || !user.watermark.url) {
+
+    upsertWatermark(user, getStorageBucketName());
+
+    // wait for the function to update the user document after watermark creation
+    const success = await new Promise(resolve => {
```
GrandSchtroumpf, line 109: I think you want to subscribe before upserting.

libs/event/src/lib/components/player/player.component.ts
```diff
@@ -23,6 +25,7 @@ export class EventPlayerComponent implements AfterViewInit, OnDestroy {
     @Inject(DOCUMENT) private document: Document,
     private eventQuery: EventQuery,
     private authQuery: AuthQuery,
+    private db: AngularFirestore,
```
GrandSchtroumpf, line 28: Inject userService instead. But I think the query should be enough

libs/event/src/lib/components/player/player.component.ts
```diff
@@ -46,20 +49,26 @@ export class EventPlayerComponent implements AfterViewInit, OnDestroy {
...
     const callDeploy = this.functions.httpsCallable('privateVideo');
     const { error, result } = await callDeploy({ eventId: id }).toPromise();
 
     if (!!error) {
-      console.log('ERROR');
+      console.log('ERROR', error);
+      throw new Error(error);
     } else {
 
+      const userRef = this.db.collection('users').doc(this.authQuery.userId);
```
GrandSchtroumpf, line 63: Why not getting the user with the query ?
```typescript
const user = this.authQuery.getActive();
if (!user.watermark?.url) {
  throw new Error('We cannot load video without watermark.');
}
```

libs/event/src/lib/components/player/player.component.ts
```diff
@@ -46,20 +49,26 @@ export class EventPlayerComponent implements AfterViewInit, OnDestroy {
...
+      const userSnap = await userRef.get().toPromise();
+      const userData = userSnap.data() as PublicUser;
+      const watermarkUrl = userData.watermark?.url;
+      if (!watermarkUrl) {
+        console.error('We cannot load video without watermark.');
```
GrandSchtroumpf, line 68: If you throw, no need to log in the console.

apps/backend-functions/src/player.ts
pldespaigne: no because I will need the `userRef` to perform the `onSnapshot` bellow

apps/backend-functions/src/player.ts
pldespaigne: this is already tackled in the `develop` PR

apps/backend-functions/src/player.ts
pldespaigne: no because we `await` for the promise to resolve, if we did it before upserting it will always timeout

libs/event/src/lib/components/player/player.component.ts
pldespaigne: `authQuery` doesn't use `CollectionService` and it's not updated in real time with fresh user data.
At least when I tried it I didn't get the update from the db

pldespaigne:
Wait until Wednesday to merge/deploy in production

PR #3649, opened 2020-09-10
Issue debug movie form

fritzschoff:


libs/movie/src/lib/movie/pipes/has-status.pipe.ts
```diff
@@ -9,6 +9,7 @@ import { MovieForm } from '../form/movie.form';
 })
 export class HasStatusPipe implements PipeTransform {
   transform(form: MovieForm, status: string[]): boolean {
+    console.log(form.get('productionStatus').value)
```
GrandSchtroumpf, line 12: Can you remove ?

GrandSchtroumpf (approved):

libs/movie/src/lib/movie/pipes/has-status.pipe.ts
fritzschoff: yup sure!

PR #3660, opened 2020-09-10
Dashboard view in movie lib

Mussuranae:
Dashboard movie view component in the movie lib

### DIFF
--- a/apps/festival/festival/src/app/dashboard/title/view/view.component.html
+++ b/apps/festival/festival/src/app/dashboard/title/view/view.component.html
@@ -39,17 +39,15 @@
     </article>
   </header>
 
-  <section>
-    <nav mat-tab-nav-bar>
-      <a mat-tab-link
-         *ngFor="let link of navLinks"
-         [routerLink]="link.path"
-         routerLinkActive #rla="routerLinkActive"
-         [active]="rla.isActive">
-        {{link.label}}
-      </a>
-    </nav>
-    <router-outlet></router-outlet>
-  </section>
+  <nav mat-tab-nav-bar>
+    <a mat-tab-link
+        *ngFor="let link of navLinks"
+        [routerLink]="link.path"
+        routerLinkActive #rla="routerLinkActive"
+        [active]="rla.isActive">
+      {{link.label}}
+    </a>
+  </nav>
+  <router-outlet></router-outlet>
 
 </ng-container>
--- a/apps/festival/festival/src/app/dashboard/title/view/view.component.html
+++ b/apps/festival/festival/src/app/dashboard/title/view/view.component.html
@@ -40,57 +40,16 @@
   </header>
 
   <section>
-    <mat-tab-group>
-      <!-- MAIN INFORMATION TAB -->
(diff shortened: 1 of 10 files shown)


libs/movie/src/lib/movie/dashboard/components/view/main/main.component.html
```diff
@@ -0,0 +1,27 @@
+<h3>Main Information</h3>
+<article class="summary">
```
GrandSchtroumpf, line 2: If every article are the class summary, you don't need to specify it, just target `article` in the css.
You might want to do `ngIf` here.

libs/movie/src/lib/movie/dashboard/components/view/main/main.component.html
```diff
@@ -0,0 +1,27 @@
...
+  <movie-summary-main [movie]="movie" [link]="getPath('main')"></movie-summary-main>
```
GrandSchtroumpf, line 3: We should get rid of the `getPath` and create a pipe instead. Or better do it directly in the `missing-control` maybe

libs/movie/src/lib/movie/dashboard/components/view/main/main.component.scss
```diff
@@ -0,0 +1,8 @@
+mat-divider {
+  border-top-width: 2px;
+  margin: 24px 0;
+}
+
+.summary {
```
GrandSchtroumpf, line 6: Use article. 
Better you can do : 
```html
<section fxLayout="column" fxLayoutGap="16px">
  <article *ngIf="">
    <h3>...</h3>
    ....
  </article>
</section>
```

libs/movie/src/lib/movie/dashboard/components/view/main/main.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  styleUrls: ['./main.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class MovieViewMainComponent {
+
+  @Input() movie: MovieForm;
+
+  constructor(private route: ActivatedRoute) {}
+
+  public getPath(segment: string) {
```
GrandSchtroumpf, line 18: I would prefer to move that in the missing-control if it make sense. Or create a pipe that you can reuse in every component.

GrandSchtroumpf (approved):

apps/festival/festival/src/app/dashboard/title/view/view.component.html
```diff
@@ -40,67 +40,16 @@ <h6>{{ getDirectors(movie) }}</h6>
   </header>
 
   <section>
```
GrandSchtroumpf, line 42: You don't need this section

libs/movie/src/lib/movie/dashboard/components/view/main/main.component.ts
GrandSchtroumpf: Remove this method.

GrandSchtroumpf (approved):

PR #3664, opened 2020-09-11
Set appAccess optional chaining

GrandSchtroumpf:
Make sure `appAccess` do not break apps each time we add a new one.

### DIFF
--- a/apps/backend-functions/src/orgs.ts
+++ b/apps/backend-functions/src/orgs.ts
@@ -13,7 +13,7 @@ import { OrganizationDocument, PublicUser, PermissionsDocument } from './data/ty
 import { algolia } from './environments/environment';
 import { NotificationType } from '@blockframes/notification/types';
 import { triggerNotifications, createNotification } from './notification';
-import { app, Module, getAppName } from '@blockframes/utils/apps';
+import { app, module, getAppName } from '@blockframes/utils/apps';
 import { getAdminIds, getAppUrl, getOrgAppKey, getDocument, createPublicOrganizationDocument, createPublicUserDocument, getFromEmail } from './data/internals';
 import { ErrorResultResponse } from './utils';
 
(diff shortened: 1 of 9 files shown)


apps/backend-functions/src/orgs.ts
```diff
@@ -70,17 +70,7 @@ async function notifyOnOrgMemberChanges(before: OrganizationDocument, after: Org
...
-      const accessChanged = (module: Module) => {
-        return after.appAccess[a][module] === true && (!before.appAccess[a] || before.appAccess[a][module] === false);
-      }
-      if (accessChanged('dashboard') || accessChanged('marketplace')) {
-        appAccessChanged = true;
-      }
-    }
-
-    return appAccessChanged;
+    return app.some(a => module.some(m => after.appAccess[a]?.[m] !== before.appAccess[a]?.[m]));
```
phpgeekfr, line 73: If I read the previous method well, `appAccessChanged` was set to true only if some app access changed from false to true. 
Your method will also set `appAccessChanged` to true if  app access changed from true to false

firebase.json
```diff
@@ -41,6 +41,15 @@
...
     },
+    {
+      "target": "financiers",
+      "public": "dist/apps/festival",
+      "ignore": ["firebase.json", "**/.*", "**/node_modules/**"],
+      "rewrites": [{
+        "source": "**",
+        "destination": "/index.html"
+      }]
+    },
```
phpgeekfr, line 52: This is the missing config that was preventing us to deploy target ?

phpgeekfr (approved):

pldespaigne (approved):

apps/backend-functions/src/orgs.ts
phpgeekfr: This will lead to unwantted email

firebase.json
GrandSchtroumpf: I think. This part shouldn't have been in the PR, I didn't test it yet. Anyway, we need it.

apps/backend-functions/src/orgs.ts
GrandSchtroumpf: Yes, but the method is `AppAccessChanged` which implies in both ways. I wanted you to double check that.
I can move back to only false -> true if this is the expected behaviour

phpgeekfr: yes this is the expected behaviour. We only send email to orgAdmins when he is granted a new app access. Not when it is removed

PR #3667, opened 2020-09-11
Issue form list

fritzschoff:


libs/movie/src/lib/movie/form/additional-information/additional-information.component.html
```diff
@@ -121,7 +119,7 @@ <h3>Rating</h3>
     <h3>Qualifications</h3>
     <section id="qualifications">
       <article>
-        <mat-button-toggle-group [formControl]="qualifications" multiple name="certifications"
+        <mat-button-toggle-group [formControl]="form.qualifications" multiple name="certifications"
```
Mussuranae, line 122: In the data model it's `certifications` so maybe you have to change `form.qualifications` to `form.certifications`

Mussuranae (approved):

libs/utils/src/lib/form/validators/validators.ts
```diff
@@ -10,7 +10,7 @@ import {
 import { LANGUAGES_SLUG } from '../../static-model/types';
 import { getLabelBySlug, isInSlug, Scope } from '../../static-model/staticModels';
 
-export const urlValidators = Validators.pattern('^(?:http(s)?:\/\/)([a-zA-Z0-9]+\.[^\s]{2,})+$');
+export const urlValidators = Validators.pattern('(https?://)?([\\da-z.-]+)\\.([a-z.]{2,6})[/\\w .-]*/?');
```
xmano, line 13: Is there no need to escape the forward slashes here? You may want to check that.

PR #3669, opened 2020-09-11
cleaned sync storage

phpgeekfr:


GrandSchtroumpf (approved):
Why did you remove syncStorage ?

phpgeekfr:
I removed it because this method was syncing content from storage to db. To have secure rules, the sync is now from db to storage.

PR #3675, opened 2020-09-14
George/fix functions config deploy

george43g:
turns out the bug is actually with firebase-tools
the two solutions were to downgrade the project's firebase-tools dependency to a lower version
or to just use execSync() to run the config set via a command

The pros/cons weighed up, and using the command, although less elegant, should be more robust and allow us to use the latest version - it seems the module API is less reliable than the CLI


tools/scripts/firebase-tools.d.ts
```diff
@@ -0,0 +1,33 @@
+declare module 'firebase-tools' {
```
GrandSchtroumpf, line 1: If firebase-tools is only used in the "tools" folder it's ok. Else put it inside the firebase-utils lib.
You might also want to publish a @typs/firebase-tools lib at some point.

GrandSchtroumpf (approved):

PR #3685, opened 2020-09-14
George/prepare for testing - static users and more commands

george43g:
- a new fixture is generated - static-users.json
- to determine which users to fetch, we read tools/static-users.json
- a few small bug fixes
- we have additional shorter commands for devs, "prepareStorage" "prepareDb" "syncUsers"

### DIFF
--- a/apps/backend-ops/src/firebaseSetup.ts
+++ b/apps/backend-ops/src/firebaseSetup.ts
@@ -78,6 +78,22 @@ export async function prepareDb() {
   console.log('Deprecated data removed!');
 }
 
+export async function prepareStorage() {
+  const { storage, getCI } = loadAdminServices();
+
+  console.info('Syncing storage with production backup stored in blockframes-ci...');
+  await restoreStorageFromCi(getCI());
+  console.info('Storage synced!');
+
+  console.info('Preparing database & storage by running migrations...');
+  await migrate(false); // run the migration, do not trigger a backup before, since we already have it!
+  console.info('Migrations complete!');
+
+  console.info('Cleaning unused storage data...');
+  await cleanStorage(storage.bucket(storageBucket));
+  console.info('Storage data clean and fresh!');
+}
+
 export async function restoreShortcut() {
   return restore(appUrl.content);
 }
--- a/apps/backend-ops/src/firebaseSetup.ts
+++ b/apps/backend-ops/src/firebaseSetup.ts
@@ -11,13 +11,12 @@ import { restore } from './admin';
 import { loadAdminServices } from "@blockframes/firebase-utils";
 import { cleanDeprecatedData } from './db-cleaning';
 import { cleanStorage } from './storage-cleaning';
-import { copyDbFromCi, readJsonlFile, restoreStorageFromCi, warnMissingVars} from '@blockframes/firebase-utils';
+import { copyDbFromCi, readJsonlFile, restoreStorageFromCi } from '@blockframes/firebase-utils';
 import { firebase } from '@env';
 import { generateFixtures } from './generate-fixtures';
 export const { storageBucket } = firebase;
 
 export async function prepareForTesting() {
-  warnMissingVars()
   const { db, auth, storage, getCI } = loadAdminServices();
   console.log('Fetching anonymized DB from blockframes-ci and uploading to local storage bucket...');
   const dbBackupPath = await copyDbFromCi(storage, getCI());
@@ -62,7 +61,6 @@ export async function prepareForTesting() {
   await generateWatermarks();
   console.info('Watermarks generated!');
 
-  process.exit(0);
 }
 
 export async function prepareDb() {
@@ -92,6 +90,10 @@ export async function prepareStorage() {
   console.info('Cleaning unused storage data...');
   await cleanStorage(storage.bucket(storageBucket));
   console.info('Storage data clean and fresh!');
+
+  console.info('Generating watermarks...');
+  await generateWatermarks();
+  console.info('Watermarks generated!');
 }
 
 export async function restoreShortcut() {
@@ -122,5 +124,4 @@ export async function upgrade() {
   await generateWatermarks();
   console.info('Watermarks generated!');
 
-  process.exit(0);
 }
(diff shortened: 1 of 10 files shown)


apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -78,6 +76,26 @@ export async function prepareDb() {
...
+
+  console.info('Cleaning unused storage data...');
+  await cleanStorage(storage.bucket(storageBucket));
+  console.info('Storage data clean and fresh!');
+
+  console.info('Generating watermarks...');
+  await generateWatermarks();
+  console.info('Watermarks generated!');
+}
+
```
phpgeekfr, line 98: There is some lines duplicated between prepareForTesting prepareStorage prepareDb, can you factorize a bit?

libs/firebase-utils/src/lib/util.ts
```diff
@@ -5,18 +5,16 @@ import { config } from 'dotenv';
 import { readFileSync } from 'fs';
 import requiredVars from 'tools/mandatory-env-vars.json';
 import { chunk } from 'lodash';
+import * as env from '@env'
 
-export function batchIteratorFactory<K = any>(batch: K[], cb: (p: K) => Promise<any>, chunkSize = 10) {
-  function* batchGenerator() {
+export function* batchGenerator<K = any>(batch: K[], cb: (p: K) => Promise<any>, chunkSize = env?.['chunkSize'] || 10) {
```
phpgeekfr, line 10: can you explain the use  of `function*` in comment for other devs ? Or not use this if not mandatory?

phpgeekfr (approved):

apps/backend-ops/src/generate-fixtures.ts
```diff
@@ -78,3 +81,22 @@ async function generateOrgs() {
...
+  console.log('Generating static user fixtures...');
+
+  for (const userType in staticUsers) {
+    try {
+      const { email, uid } = await getDocument<User>(`users/${staticUsers[userType]}`);
+      console.log(`User type: ${userType} found! UID: ${uid}`)
+      staticUsers[userType] = { uid, email, password: USER_FIXTURES_PASSWORD }
+    } catch (e) {
+      console.log(`Failed to find: ${userType} : ${staticUsers[userType]}`)
+      return;
```
GrandSchtroumpf, line 95: Are you blocking everything if one user failed ?

apps/backend-ops/src/users.ts
```diff
@@ -96,14 +97,24 @@ function readUsersFromJsonlFixture(db: JsonlDbRecord[]): UserConfig[] {
     }));
 }
 
-export async function syncUsers(db: JsonlDbRecord[]): Promise<any> {
-  await startMaintenance();
-  const { auth } = loadAdminServices();
+async function getUsersFromDb(db:FirebaseFirestore.Firestore ) {
+  const usersIterator = getCollectionInBatches<User>(db.collection('users'), 'uid', 300);
+  const output: UserConfig[] = [];
+  for await (const users of usersIterator)
```
GrandSchtroumpf, line 103: You should be able to do : 
```typescript
const password = USER_FIXTURES_PASSWORD;
const userMatrix = getCollectionInBatches<User>(db.collection('users'), 'uid', 300);
return userMatrix.map(users => {
  return users.map(({ uid, email }) => ({ uid, email, password })
}).flat();
```

tools/static-users.json
```diff
@@ -0,0 +1,7 @@
+{
```
GrandSchtroumpf, line 1: I think this should be gitignored

GrandSchtroumpf (approved):

tools/static-users.json
```diff
@@ -0,0 +1,7 @@
...
+  "marketConcierge": "gLSyPoYiJSPQ6d50I0HXo8xtTYI2",
+  "marketingManager": "B8UsXliuxwY6ztjtLuh6f7UD1GV2",
+  "salesCoordinator": "aulzag1uZaQaWwEX7QDQSN9SXuc2",
+  "partner": "Coow7hjqiRXwv6e7HNSzZjmMBGo1",
+  "buyer": "1M9DUDBATqayXXaXMYThZGtE9up1"
+}
```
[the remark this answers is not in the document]
george43g: yep, so they can add the users they need and commit changes in order to generate the fixture that has the email/password for that user

libs/firebase-utils/src/lib/watermark.ts
```diff
@@ -21,7 +20,7 @@ export function getWatermark(email: string, firstName: string = '', lastName: st
  * - Store the watermark file in the storage bucket
  * - User's firestore doc is updated by onFileUpload backend function
  */
-export async function upsertWatermark(user: PublicUser, bucketName: string): Promise<any> {
+export async function upsertWatermark(user: PublicUser, bucketName: string, storage?: admin.storage.Storage): Promise<any> {
```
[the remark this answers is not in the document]
george43g: It only becomes an issue when you're dealing with multiple apps at once (let's say CI env and local env) and you want to ensure that the correct app is being accessed. If the order of initialisation changed, and you used this method, it's possible to accidentally run commands on the wrong env, which could be bad.

apps/backend-ops/src/firebaseSetup.ts
george43g: Eventually yes, but the issue is both commands run `migrate()` so if we used those commands within prepareForTesting it would be run twice. Possible but in future.

tools/static-users.json
george43g: That's the one where devs choose the ones they want, then the fixture is generated also containing the email and password. Mainly a convenience thing for devs who want to manually log in

apps/backend-ops/src/users.ts
george43g: This would return a two-dimensional array. This function is a generator and the batch results need to be joined together into an array. I wasn't really able to implement any of this. the `map` function does not exist on userMatrix, and it's not a matrix. It's essentially an array of promises coming in batches.

PR #3704, opened 2020-09-15
feat(backend-ops): disable maintenance mode with --skipMaintenance flag

george43g:
- added guard clause to maintenance methods if maintenance is disabled
- backend-ops which is what CI runs when making a release, will check for this flag
- even without the flag, maintenance mode can be disabled with an environment variable
- `BLOCKFRAMES_MAINTENANCE_DISABLED`
- Pros: freedom to disable it from environment, in CI setup, in terminal, in .env, and also via flag
- cons: maybe add a `console.warn` in `startMaintenance` if it's disabled

### DIFF
--- a/apps/backend-ops/src/main.ts
+++ b/apps/backend-ops/src/main.ts
@@ -12,7 +12,7 @@ import { clearUsers, createUsers, printUsers, generateWatermarks, syncUsers } fr
 import { generateFixtures } from './generate-fixtures';
 
 const args = process.argv.slice(2);
-const [cmd, ...rest] = args;
+const [cmd, ...flags] = args;
 
 async function runCommand() {
   if (cmd === 'prepareForTesting') {
@@ -49,6 +49,15 @@ async function runCommand() {
     showHelp();
     return Promise.reject('Command not recognised');
   }
+}
+
+function flagExists(compare: string) {
+  return flags.find(flag => flag === compare)
+}
+
+if (flagExists('--skipMaintenance') || flagExists('skipMaintenance')) {
+  process.env.BLOCKFRAMES_MAINTENANCE_DISABLED = 'CLI flag';
+  console.warn('WARNING! BLOCKFRAMES_MAINTENANCE_DISABLED is set to true, meaning maintenance mode is disabled!')
 }
 
 const consoleMsg = `Time running command "${cmd}"`;
--- a/apps/backend-ops/src/main.ts
+++ b/apps/backend-ops/src/main.ts
@@ -51,12 +51,12 @@ async function runCommand() {
   }
 }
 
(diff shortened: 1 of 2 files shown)


apps/backend-ops/src/main.ts
```diff
@@ -51,6 +51,15 @@ async function runCommand() {
   }
 }
 
+function flagExists(compare: string) {
```
GrandSchtroumpf, line 54: just a naming convention in the repo : 
- If it's a boolean: use `has***` or `is***`. Then use `[].some()` or `[].every()`
- If it return a value: use `get***` or `query***`. Then user `[].find()` or `[].filter()`

apps/backend-ops/src/main.ts
```diff
@@ -51,6 +51,15 @@ async function runCommand() {
...
+  return flags.find(flag => flag === compare)
+}
+
+if (flagExists('--skipMaintenance') || flagExists('skipMaintenance')) {
```
GrandSchtroumpf, line 58: I think you can put the `--` condition in the `flagExists` function directly (if `flagExists` is supposed to return a boolean).

apps/backend-ops/src/main.ts
```diff
@@ -51,6 +51,15 @@ async function runCommand() {
...
+  process.env.BLOCKFRAMES_MAINTENANCE_DISABLED = 'CLI flag';
```
GrandSchtroumpf, line 59: I don't understand why you need to do that.

GrandSchtroumpf (approved):

apps/backend-ops/src/main.ts
george43g: any value is true, but it must be a string.  I thought maybe store more information as to why it's disabled for diagnostics. But can also just write = "true"

apps/backend-ops/src/main.ts
george43g: It needs to work with and without -- in order for `npm run backend-ops skipMaintenance` to work. You cant use -- when using npm run because it thinks it's a flag for npm.

george43g: but if for some reason in future we want to use this function and check for a flag and differentiate between a word with and without --, then it wouldn't work

apps/backend-ops/src/main.ts
GrandSchtroumpf: this would be better indeed: 
```typescript
process.env.BLOCKFRAMES_MAINTENANCE_DISABLED = true
```

apps/backend-ops/src/main.ts
GrandSchtroumpf: ok

PR #3709, opened 2020-09-16
Issue movie form debug

fritzschoff:


libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -71,6 +71,7 @@ interface MovieRaw<D> {
   originCountries: TerritoriesSlug[], //! required
   poster?: string;
   prizes?: Prize[],
+  customPrize?: Prize[],
```
GrandSchtroumpf, line 74: Did you check with @Mussuranae or @phpgeekfr for the model ?

libs/movie/src/lib/movie/form/additional-information/additional-information.component.html
```diff
@@ -15,8 +15,8 @@ <h3>Release</h3>
               <div fxLayout="row" fxLayoutGap="24px">
                 <form-country test-id="release-country" [form]="releaseForm.get('country')" fxFlex></form-country>
 
-                <static-select test-id="release-media" type="model" [control]="releaseForm.get('media')" scope="MEDIAS"
-                  fxFlex>
+                <static-select test-id="release-media" type="model" [control]="releaseForm.get('media')" scope="MEDIAS" fxFlex
+                  [withoutValues]="['pay-tv', 'free-tv', 'pay-per-view', 'est', 'n-vod','a-vod', 'f-vod', 's-vod', 'planes', 'boats']">
```
GrandSchtroumpf, line 19: looks like you're removing everything. Can you explain why you need to do that ? It looks like a workaround more than a clean way of dealing with a problem.

libs/movie/src/lib/movie/form/available-materials/available-materials.component.html
```diff
@@ -34,21 +34,23 @@ <h5>{{ language.key | titlecase }}</h5>
 
     <!-- Buttons -->
 
-    <footer *ngIf="!showButtons">
+    <!-- Can't use footer html node because tunnel layout component as some styles set on the footer -->
```
GrandSchtroumpf, line 37: Put a class on the footer of the tunnel layout instead. When a component as ViewEncaspulation.None it's not a good idea to style the tag directly.

libs/movie/src/lib/movie/form/reviews/reviews.component.html
```diff
@@ -60,6 +58,62 @@ <h6 class="mat-caption">Premiere</h6>
...
+                <mat-error *ngIf="customPrizeForm.get('prize').hasError('maxlength')">
+                  Awards and selections should not exceed 200 characters.
+                </mat-error>
+                <mat-hint>Please fill in your project's award(s) and/or selection(s) for this festival.</mat-hint>
+              </mat-form-field>
+
+              <mat-form-field appearance="outline" fxFlex>
+                <mat-label>Year</mat-label>
+                <input matInput test-id="prize-year" formControlName="year" type="number" placeholder="YYYY" />
+                <mat-error>Please enter a 4 digit date (e.g. 2020)</mat-error>
```
GrandSchtroumpf, line 88: You don't need a `ngIf hasError` here ?

libs/movie/src/lib/movie/form/reviews/reviews.component.html
```diff
@@ -60,6 +58,62 @@ <h6 class="mat-caption">Premiere</h6>
...
+                <mat-error *ngIf="customPrizeForm.hasError('required')">
+                  Year is required.
+                </mat-error>
+              </mat-form-field>
+
+            </div>
+
+            <h6 class="mat-caption">Premiere</h6>
+            <div>
```
GrandSchtroumpf, line 97: I'm pretty sure you don't need that div.

libs/movie/src/lib/movie/form/shell/shell.component.ts
```diff
@@ -157,7 +157,7 @@ export class MovieFormShellComponent implements TunnelRoot, OnInit, AfterViewIni
   }
 
   confirmExit() {
-    if (this.form.pristine) {
+    if (this.form.status !== 'INVALID') {
```
GrandSchtroumpf, line 160: Hmmm, I think pristine is better, you should just mark the form as pristine after save.

libs/movie/src/lib/movie/form/summary/review/review.component.html
```diff
@@ -35,6 +35,43 @@ <h6>
...
 </article>
 
+<!-- CUSTOM FESTIVAL -->
+<article>
+  <h6>
+    Custom Festival{ movie.customPrize.controls.length, plural, =0 {} =1 {} other {s}}
+    <a mat-icon-button [routerLink]="link" fragment="festival"><mat-icon svgIcon="pencil" color="primary"></mat-icon></a>
+  </h6>
+  <ul>
+    <ng-container *ngIf="!movie.customPrize.controls.length else displayCustomFestival">
```
GrandSchtroumpf, line 45: Can you check with @2amarkina, but I don't think we need to display "missing" when there is no customPrize.

libs/ui/src/lib/static-autocomplete/select/static-select.component.html
```diff
@@ -1,10 +1,10 @@
-<ng-container *ngIf="type === 'model'; else constant">
+<div *ngIf="type === 'model'">
```
GrandSchtroumpf, line 1: Why you needed a div ? two ng-container didn't work ?

libs/ui/src/lib/form/country/country.component.html
```diff
@@ -4,9 +4,12 @@
   <mat-error *ngIf="form.hasError('invalidValue')">
     Please select one of the country suggested.
   </mat-error>
-  <mat-hint>You can select "World" for worldwide box office.</mat-hint>
+  <mat-hint>
+    <ng-content select="mat-hint"></ng-content>
```
GrandSchtroumpf, line 8: Is it working ? back in the day, we could do that.

libs/ui/src/lib/static-autocomplete/select/static-select.component.html
```diff
@@ -19,25 +19,21 @@
...
+      <ng-container *ngFor="let value of staticValue">
+        <mat-option [value]="value">
+          {{ value | toLabel: _scope }}
         </mat-option>
       </ng-container>
     </mat-select>
-    <mat-hint>
-      <ng-content select="mat-hint"></ng-content>
-    </mat-hint>
+    <mat-hint>{{hint}}</mat-hint>
```
GrandSchtroumpf, line 34: stay consistent with the country.component (see comment above)

libs/ui/src/lib/static-autocomplete/select/static-select.component.ts
```diff
@@ -12,28 +12,33 @@ import { boolean } from '@blockframes/utils/decorators/decorators';
...
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class StaticSelectComponent implements OnInit {
+export class StaticSelectComponent {
 
-  public _scope: string[];
-  @Input() scope;
   @Input() type: 'constant' | 'model';
+
+  public staticValue;
```
GrandSchtroumpf, line 19: Call it "options" and type it

libs/ui/src/lib/static-autocomplete/select/static-select.component.ts
```diff
@@ -12,28 +12,33 @@ import { boolean } from '@blockframes/utils/decorators/decorators';
...
+  public _scope
```
GrandSchtroumpf, line 20: type it

libs/ui/src/lib/static-autocomplete/select/static-select.component.ts
```diff
@@ -12,28 +12,33 @@ import { boolean } from '@blockframes/utils/decorators/decorators';
...
   @Input() control: FormControl;
+  @Input() label: string;
+  @Input() hint: string;
   @Input() mode: 'legacy' | 'standard' | 'fill' | 'outline' = 'outline';
   @Input() placeholder: string;
   @Input() @boolean required: boolean;
   @Input() set withoutValues(toFilterValue: any[]) {
     if (this.type === 'constant') {
-      this._scope = Object.keys(staticConsts[this.scope]).filter(scopeValue => toFilterValue.includes(scopeValue))
+      this.staticValue = Object.keys(staticConsts[this._scope]).filter(scopeValue => !toFilterValue.includes(scopeValue));
```
GrandSchtroumpf, line 37: I haven't analyzed every cases, but having two inputs that depends on each other is usually a source of bugs.

GrandSchtroumpf (changes requested):

libs/movie/src/lib/movie/form/additional-information/additional-information.component.html
fritzschoff: request from @Jackseed that only needs a part of this list.

libs/ui/src/lib/form/country/country.component.html
fritzschoff: yes you need to wrap it. I guess we are using it in any static select

libs/ui/src/lib/static-autocomplete/select/static-select.component.html
fritzschoff: yes this is a super edge case. Since you are here in the second ng if case. Meaning there is already an ng-content select above that. Meaning you will project the content there and not here. So this is why I made this ugly workaorund. When we finally merging the two static.ts files, we don't need that anymore.

libs/ui/src/lib/static-autocomplete/select/static-select.component.ts
fritzschoff: true, will look into it

libs/movie/src/lib/movie/form/additional-information/additional-information.component.html
GrandSchtroumpf: I think it only applies if the movie.status is "released" no ?

libs/movie/src/lib/movie/+state/movie.firestore.ts
fritzschoff: yes

libs/movie/src/lib/movie/form/reviews/reviews.component.html
fritzschoff: you need one, otherwise the last mat toggle button will get attched to the end

libs/movie/src/lib/movie/form/shell/shell.component.ts
fritzschoff: unfortunatly pristine doesn't get changes for the global form at all. Even if you touch some inputs

GrandSchtroumpf: strange. What about touched ?

libs/ui/src/lib/static-autocomplete/select/static-select.component.ts
```diff
@@ -12,28 +12,33 @@ import { boolean } from '@blockframes/utils/decorators/decorators';
...
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class StaticSelectComponent implements OnInit {
+export class StaticSelectComponent {
 
-  public _scope: string[];
-  @Input() scope;
   @Input() type: 'constant' | 'model';
+
+  public staticValue: any[] = [];
```
GrandSchtroumpf, line 19: It can be something else than string or number ?

libs/utils/src/lib/pipes/isAbove.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform, NgModule } from '@angular/core';
+
+@Pipe({
+  name: 'isAbove'
+})
+export class IsAbovePipe implements PipeTransform {
+  transform(count: number, min: number): boolean {
+    if (min === 0 || !min) min = 0;
```
GrandSchtroumpf, line 8: You can initialize it to 0 in the transform param (better). Else : 
```typescript
min = min ?? 0;
return count > min;
```

GrandSchtroumpf (approved):

PR #3711, opened 2020-09-16
George/cleanup - move firestoreMigrations to firebase-utils

george43g:
This is a big change so I thought I would make it's own PR for it

interestingly, to allow git to be able to track file history when moving a file, a good practice is to dedicate one commit to just moving the files, without making further changes (otherwise git will lose track). I separated it out, more info here: https://stackoverflow.com/questions/2314652/is-it-possible-to-move-rename-files-in-git-and-maintain-their-history


phpgeekfr (approved):

libs/movie/src/lib/distribution-rights/+state/distribution-right.model.ts
```diff
@@ -56,10 +56,9 @@ export interface DistributionRightWithMovieId {
   movieId: string;
 }
 
-export function createDistributionRightWithMovieId(params: Partial<DistributionRightWithMovieId> = {}): DistributionRightWithMovieId {
+export function createDistributionRightWithMovieId(params: Optional<DistributionRightWithMovieId, 'right'>): DistributionRightWithMovieId {
```
GrandSchtroumpf, line 59: I'm not sure to understand why you need to omit right.
It should look like that : 
```typescript
export function createDistributionRightWithMovieId(params: Optional<DistributionRightWithMovieId> = {}): DistributionRightWithMovieId {
  return {
    ...params,
    right: createDistributionRight(params?.right),
  }
}
```

GrandSchtroumpf (approved):

libs/movie/src/lib/distribution-rights/+state/distribution-right.model.ts
george43g: `'right' is specified more than once, so this usage will be overwritten.ts(2783)`
That's if we don't make right optional.
If we make the whole thing partial, then 
```
Type '{ right: DistributionRight; movieId?: string | undefined; }' is not assignable to type 'DistributionRightWithMovieId'.
  Types of property 'movieId' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.ts(2322)
```

I can propose fixes for all of them and we'll discuss on the PR.

apps/backend-functions/src/media.ts
```diff
@@ -135,74 +135,74 @@ function getPathInfo(ref: string) {
 
 
 export async function cleanUserMedias(before: PublicUser, after?: PublicUser): Promise<void> {
-  const mediasToDelete = [];
+  const mediaToDelete: string[] = [];
```
phpgeekfr, line 138: why did you remove "s"? It is an array of medias

apps/backend-functions/tsconfig.json
```diff
@@ -3,10 +3,8 @@
   "compilerOptions": {
     "types": ["node", "jest", "express"],
     "module": "commonjs",
-    "noImplicitReturns": true,
-    "noUnusedLocals": true,
```
phpgeekfr, line 7: we don't need this anymore?

apps/backend-functions/src/media.ts
george43g: just seemed like wrong english, 'media' is already plural

apps/backend-functions/tsconfig.json
george43g: may need, but for this PR had to turn off due to the number of errors caused preventing merging. we can switch back on in another PR if need be

PR #3714, opened 2020-09-16
fix(syncUsers): implement throttling and urgent bug fix

george43g:
urgent bugfix, better throttling for syncUsers


libs/firebase-utils/src/lib/util.ts
```diff
@@ -30,6 +32,21 @@ export async function* getCollectionInBatches<K>(ref: admin.firestore.Collection
   }
 }
 
+export async function runInBatches<T, K>(batch: K[], cb: (i: K) => Promise<T>, chunkSize = env?.['chunkSize'] || 10) {
```
phpgeekfr, line 35: again ? You said that using RunChunk was enough .

libs/firebase-utils/src/lib/util.ts
```diff
@@ -4,6 +4,8 @@ import { firebase as firebaseCI } from 'env/env.ci';
 import { config } from 'dotenv';
 import { readFileSync } from 'fs';
 import requiredVars from 'tools/mandatory-env-vars.json';
+import { chunk } from 'lodash';
+import * as env from '@env'
```
phpgeekfr, line 8: missing `;`

libs/testing/src/lib/firebase/auth.ts
```diff
@@ -18,19 +19,7 @@ export async function deleteAllUsers(auth: admin.auth.Auth) {
...
-      yield Promise.all(batch.map(userRecord => auth.createUser(userRecord)))
-    }
-  }
-  const importUsersIterator = iterateImportUsers(users);
-  let successCount = 0;
-  for await (const deletion of importUsersIterator) {
-    await sleep(1000)
-    successCount += deletion.length;
-  }
+  const output = await runInBatches(users, userRecord => auth.createUser(userRecord), (env?.['chunkSize'] || 10) * 10)
```
phpgeekfr, line 22: missing `;`

phpgeekfr (changes requested):
Can you use runChunks ? @GrandSchtroumpf and I don't want to have two methods for the same usage

phpgeekfr (approved):

PR #3715, opened 2020-09-16
feat(backend-ops): only run maintenance mode when required RC

george43g:
fast and simple solution to disable maintenance mode in firestore when not required. 
PR for the purposes of discussing this method vs alternatives


.circleci/config.yml
```diff
@@ -245,14 +245,16 @@ jobs:
...
           command: |
             npx firebase use staging &&
+            ( npm run backend-ops isMaintenanceRequired || cp firestore.rules firestore.maintenance.rules ) &&
             cp firestore.rules firestore.rules.regular &&
             cp firestore.maintenance.rules firestore.rules &&
             npx firebase deploy --only hosting:catalog,hosting:festival,hosting:crm,storage &&
             npx firebase deploy --only firestore --force && 
             npx firebase deploy --only functions --force
-      - run: node dist/apps/backend-ops/main.js upgrade
+      - run: npm run backend-ops isMaintenanceRequired && npm run backend-ops upgrade || npm run backend-ops upgrade skipMaintenance
```
phpgeekfr, line 257: This is dangerous to me because not only "migrate" in the upgrade process is needing maintenance mode, cleanDeprecatedData & cleanStorage also do. So if skipMaintenance is set to true, maybe the whole "upgrade" should  be skipped (so we update only front or backend code)

.circleci/config.yml
```diff
@@ -245,14 +245,16 @@ jobs:
       - slack/notify:
           message: '🛵 Deploying to staging.. (https://staging.blockframes.io/)'
       - run:
+          # ! npm run backend-ops isMaintenanceRequired && echo 'YES IT IS' || echo 'NO ITS NOT'
```
phpgeekfr, line 248: required comment ?

.circleci/config.yml
```diff
@@ -245,14 +245,16 @@ jobs:
...
           command: |
             npx firebase use staging &&
+            ( npm run backend-ops isMaintenanceRequired || cp firestore.rules firestore.maintenance.rules ) &&
```
phpgeekfr, line 251: can you explain how this is working ? I don't understand this syntax.  What is the point of copying  firestore.rules into firestore.maintenance.rules if two lines after we have `cp firestore.maintenance.rules firestore.rules`  ?  Can't you just skip firestore.maintenance.rules use if maintenance is not required?

phpgeekfr (approved):
LGTM

Issue #3674, opened 2020-09-14
Debug issue release 1.9 - summary


Jackseed:
## Cropper
- [x] When I hover an already uploaded image with another image, it deletes the previous one 
![image](https://user-images.githubusercontent.com/14964443/92749715-aa817700-f386-11ea-8e15-52b6cb13a525.png)
- [x] The search bar on the film page is confusing, it looks like an image that didn't load. Should have a placeholder, and generally, the same design as the original search bar
![image](https://user-images.githubusercontent.com/55785823/91325556-a1f44280-e7c3-11ea-8948-1357b9789d9b.png)

## Invitations
- [x] Accept all butons turns all invitations to accepted, even the declined one, it should only affect the pending ones
- [x] request to access a screening from buyer to seller has a wrong link id, it uses the invitation id instead of the event id
![image](https://user-images.githubusercontent.com/14964443/92743478-dc8fda80-f380-11ea-84d9-aa358f7bc7bb.png)

## Org list page `/c/o/marketplace/organization`
- [x] https://staging.archipelmarket.com/c/o/marketplace/organization
There is a glitch when you click on the movie icon on the org card

## Marketplace homepage
- [x] `/c/o/marketplace/home` Change the New films carousel on marketplace homepage to really display new films, i.e. sort films by creation dates and display them from the newest to oldest
- [x] Same sort by should be use for films in `c/o/marketplace/organization/orgId/title` > newest to oldest

- [x] Carousels with "In Production" films & "In development" are not displayed on prod but they should. Currently only completed films
![image](https://user-images.githubusercontent.com/14964443/85882734-1b7dad80-b7e0-11ea-8926-2de044ebf30a.png)

## Reset password
- [x] Test reset password function, it's not reliable, lot of complaints about it. Works for me unfortunately so can't give much more details...

## Dashboard event form
- [x] On invitations, email adresses starting with an "s" are not working, the "s" is deleted - I can't reproduce this one. 's' is not removed.
- [x] We shouldn't be able to send an invitation to someone who was already invited.
![image](https://user-images.githubusercontent.com/55785823/82893137-db14e080-9f50-11ea-98a3-8fe915cdde56.png)
- [x] When we change start date of an event, end date should adapt: event should stay the same duration. E.g. Start date 09/15/20 10:00 am / End date  09/15/20 11:00 am > start date changed to 09/16/20 11:00 am, end date should be automatically updated to 09/16/20 noon

## Titles page
[...]

PR #3717, opened 2020-09-17
fix filterForm not triggering initially

RemcoSimonides:
To do on issue #3674 

Movie titles page loading infinitely


fritzschoff (approved):
nice code

PR #3720, opened 2020-09-17
Fix wrong state on leave hover

RemcoSimonides:
To do in issue #3674 

"When I hover an already uploaded image with another image, it deletes the previous one"

### DIFF
--- a/libs/media/src/lib/components/cropper/cropper.component.ts
+++ b/libs/media/src/lib/components/cropper/cropper.component.ts
@@ -111,7 +111,7 @@ export class CropperComponent implements OnInit {
   @HostListener('dragleave', ['$event'])
   onDragLeave($event: DragEvent) {
     $event.preventDefault();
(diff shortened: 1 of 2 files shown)


libs/media/src/lib/components/cropper/cropper.component.ts
```diff
@@ -111,7 +111,11 @@ export class CropperComponent implements OnInit {
   @HostListener('dragleave', ['$event'])
   onDragLeave($event: DragEvent) {
     $event.preventDefault();
-    this.nextStep('drop');
+    if (!!this.form.blobOrFile.value || (this.form.ref?.value && !this.form.delete)) {
```
RemcoSimonides, line 114: !! @ this.form.ref?.value

pldespaigne: auto-review ^^ ☝️

pldespaigne (approved):
don't forget to remove the `delete` flag

PR #3729, opened 2020-09-18
Fix attendEvent permission and link on invitation

RemcoSimonides:
To do in issue #3674 

"Request to access a screening from buyer to seller has a wrong link id, it uses the invitation id instead of the event id"

Also, users that are not an admin didn't have access to requests to attend the event. Now they do.

### DIFF
--- a/libs/invitation/src/lib/guard/invitations.guard.ts
+++ b/libs/invitation/src/lib/guard/invitations.guard.ts
@@ -48,6 +48,9 @@ export class InvitationGuard extends CollectionGuard<InvitationState> {
           return combineLatest([
             this.service.syncCollection(ref => ref.where('fromUser.uid', '==', user.uid)),
             this.service.syncCollection(ref => ref.where('toUser.uid', '==', user.uid)),
+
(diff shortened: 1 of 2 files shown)


libs/invitation/src/lib/guard/invitations.guard.ts
```diff
@@ -48,6 +48,9 @@ export class InvitationGuard extends CollectionGuard<InvitationState> {
           return combineLatest([
             this.service.syncCollection(ref => ref.where('fromUser.uid', '==', user.uid)),
             this.service.syncCollection(ref => ref.where('toUser.uid', '==', user.uid)),
+
+            this.service.syncCollection(ref => ref.where('type', '==', 'attendEvent').where('fromOrg.id', '==', orgId)),
+            this.service.syncCollection(ref => ref.where('type', '==', 'attendEvent').where('toOrg.id', '==', orgId)),
```
fritzschoff, line 53: why complicated when easy is possible!

fritzschoff (approved):

PR #3730, opened 2020-09-18
artistic, production & additional

GrandSchtroumpf:


libs/movie/src/lib/movie/marketplace/additional/additional.component.html
```diff
@@ -1,43 +1,88 @@
...
+    <h2>Release</h2>
+    <article *ngFor="let release of movie.originalRelease">
+      <h5>{{ release.country | translateSlug : 'TERRITORIES'}}</h5>
+      <p>{{ release.media }} {{ release.date | date: 'shortDate' }}</p>
+    </article>
+  </section>
+
+  <mat-divider></mat-divider>
+
+  <section *ngIf="movie | hasKeys: 'estimatedBudget'">
```
Mussuranae, line 14: if there is no estimatedBudget, there will be 2 dividers one after the other ?

libs/movie/src/lib/movie/marketplace/production/production.component.html
```diff
@@ -0,0 +1,53 @@
+<ng-container *ngIf="(movie$ | async) as movie" test-id="content" >
+  <h1 *ngIf="movie | hasKeys: keys">Production Details</h1>
```
Mussuranae, line 2: I don't understand why the hasKey pipe is on a title. Souldn't it be on a section or something like that ?

Mussuranae (approved):

PR #3733, opened 2020-09-18
Update UI for search bar in pageBar

RemcoSimonides:
To do in issue #3674 
"The search bar on the film page is confusing, it looks like an image that didn't load. Should have a placeholder, and generally, the same design as the original search bar"

- [x] change on catalog too

See result below:
image

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/title/list/list.component.html
+++ b/apps/catalog/catalog/src/app/marketplace/title/list/list.component.html
@@ -2,8 +2,9 @@
 
   <!-- App Bar Search -->
   <ng-template titleAppBarSearch>
-    <form class="app-bar" [formGroup]="filterForm">
-      <input type="text" formControlName="query" [value]="filterForm.query">
+    <form class="app-bar" [formGroup]="filterForm" fxLayout fxLayoutAlign="center center">
+      <mat-icon svgIcon="magnifying_glasses"></mat-icon>
+      <input type="text" formControlName="query" [value]="filterForm.query" placeholder="Search for titles, directors, keywords, ...">
     </form>
   </ng-template>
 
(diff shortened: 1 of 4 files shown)


apps/catalog/catalog/src/app/marketplace/title/list/list.component.html
```diff
@@ -2,8 +2,9 @@
 
   <!-- App Bar Search -->
   <ng-template titleAppBarSearch>
-    <form class="app-bar" [formGroup]="filterForm">
-      <input type="text" formControlName="query" [value]="filterForm.query">
+    <form class="app-bar" [formGroup]="filterForm" fxLayout fxLayoutAlign="center center">
+      <mat-icon svgIcon="magnifying_glasses"></mat-icon>
+      <input type="text" formControlName="query" [value]="filterForm.query" placeholder="Search for titles, directors, keywords, ...">
```
fritzschoff, line 7: don't use form here, if you hit enter when you type something in the input, you reload the page

apps/festival/festival/src/app/marketplace/title/list/list.component.html
```diff
@@ -2,8 +2,9 @@
 
   <!-- App Bar Search -->
   <ng-template titleAppBarSearch>
-    <form class="app-bar" [formGroup]="filterForm">
-      <input type="text" formControlName="query" [value]="filterForm.query">
+    <form class="app-bar" [formGroup]="filterForm" fxLayout fxLayoutAlign="center center">
+      <mat-icon svgIcon="magnifying_glasses"></mat-icon>
```
fritzschoff, line 6: same here

fritzschoff (approved):

apps/catalog/catalog/src/app/marketplace/title/list/list.component.html
RemcoSimonides: Did you try it? If I hit enter, it doesn't refresh the page 🤔

fritzschoff: yes, if we use form in the tunnel for instance, we refresh. It is the default behavior of the form html tag when you focus in. But if not the case, all good here

PR #3736, opened 2020-09-18
Polishing on marketplace

GrandSchtroumpf:


libs/movie/src/lib/movie/marketplace/main/main.component.ts
```diff
@@ -1,17 +1,18 @@
-import { Component, ChangeDetectionStrategy, OnInit } from '@angular/core';
+import { Component, ChangeDetectionStrategy, OnInit, HostBinding } from '@angular/core';
```
Mussuranae, line 1: Do you really use the hostBinding ?

Mussuranae (approved):

PR #3739, opened 2020-09-21
Filter out already invited emails

RemcoSimonides:
We shouldn't be able to send an invitation to someone who was already invited.

- [x] Add email Facet to prod_users and staging_users on Algolia too

### DIFF
--- a/libs/invitation/src/lib/form/user/user.component.html
+++ b/libs/invitation/src/lib/form/user/user.component.html
@@ -6,7 +6,7 @@
     label="Invite guests"
     [separators]="separators"
     placeholder="Enter guest email"
-    [filters]="_invitations | mapDeepKey:'toUser.email'"
+    [filters]="invitations | deepKey:'toUser.email'"
     unique>
     <ng-template let-user>
       <div class="profile-item" fxLayout fxLayoutAlign="strt center" fxLayoutGap="8px">
--- a/libs/invitation/src/lib/form/user/user.component.html
+++ b/libs/invitation/src/lib/form/user/user.component.html
@@ -1,6 +1,13 @@
 <div class="search" fxLayout fxLayout="space-between center" fxLayoutGap="16px">
-  <algolia-chips-autocomplete fxFlex index="user" displayWithPath="email" [form]="form" label="Invite guests"
-    [separators]="separators" placeholder="Enter guest email">
+  <algolia-chips-autocomplete fxFlex 
+    index="user" 
+    displayWithPath="email"
+    [form]="form"
+    label="Invite guests"
+    [separators]="separators"
+    placeholder="Enter guest email"
+    [filters]="filter"
+    unique>
     <ng-template let-user>
       <div class="profile-item" fxLayout fxLayoutAlign="strt center" fxLayoutGap="8px">
         <img [ref]="user.avatar" asset="profil_user.webp" alt="Profile">
--- a/libs/invitation/src/lib/form/user/user.component.html
+++ b/libs/invitation/src/lib/form/user/user.component.html
@@ -6,7 +6,7 @@
     label="Invite guests"
     [separators]="separators"
     placeholder="Enter guest email"
-    [filters]="filter"
+    [filters]="_invitations | mapDeepKey:'toUser.email'"
     unique>
     <ng-template let-user>
       <div class="profile-item" fxLayout fxLayoutAlign="strt center" fxLayoutGap="8px">
--- a/libs/invitation/src/lib/form/user/user.component.ts
+++ b/libs/invitation/src/lib/form/user/user.component.ts
@@ -16,13 +16,8 @@ import { Validators } from '@angular/forms';
 })
 export class UserComponent {
   @Input() docId: string;
+  @Input() invitations: Invitation[] = [];
 
-  @Input() set invitations(inv) {
-    if (!inv) return;
-    this._invitations = inv;
-  }
-  _invitations: Invitation[] = [];
-  
   @Input() ownerId: string;
   separators = [ENTER, COMMA, SEMICOLON, SPACE];
   form = createAlgoliaUserForm(Validators.maxLength(50));
--- a/libs/invitation/src/lib/form/user/user.component.ts
+++ b/libs/invitation/src/lib/form/user/user.component.ts
@@ -2,7 +2,7 @@ import { Component, ChangeDetectionStrategy, Input } from '@angular/core';
 import { BehaviorSubject } from 'rxjs';
 import { createAlgoliaUserForm } from '@blockframes/utils/algolia';
 import { scaleIn } from '@blockframes/utils/animations/fade';
-import { InvitationService } from '@blockframes/invitation/+state';
+import { Invitation, InvitationService } from '@blockframes/invitation/+state';
 import { OrganizationService } from '@blockframes/organization/+state';
 import { ENTER, COMMA, SEMICOLON, SPACE } from '@angular/cdk/keycodes';
 import { Validators } from '@angular/forms';
@@ -16,10 +16,15 @@ import { Validators } from '@angular/forms';
 })
 export class UserComponent {
   @Input() docId: string;
+  @Input() set invitations(invitations: Invitation[]) {
+    if (!invitations) return;
+    this.filter = invitations.map(invitation => `email:-${invitation.toUser?.email}`);
+  };
   @Input() ownerId: string;
   separators = [ENTER, COMMA, SEMICOLON, SPACE];
   form = createAlgoliaUserForm(Validators.maxLength(50));
   sending = new BehaviorSubject(false);
+  filter: string[] = []
   constructor(
     private service: InvitationService,
     private orgService: OrganizationService,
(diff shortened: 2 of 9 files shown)


libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
```diff
@@ -74,6 +75,16 @@ export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
 
   @Input() filters: (string | string[])[] = [];
 
+  /**  
+   * Name of attribute which values shouldn't be used before.
+   * Using an attribute that hasnt been used before? make sure to add it to Facets on Algolia */
+  @Input()
```
GrandSchtroumpf, line 81: We've got a decorator to perform that : 
```typescript
@Input() @boolean unique;
```
Use it instead.

libs/invitation/src/lib/form/user/user.component.ts
```diff
@@ -16,10 +16,15 @@ import { Validators } from '@angular/forms';
 })
 export class UserComponent {
   @Input() docId: string;
+  @Input() set invitations(invitations: Invitation[]) {
```
GrandSchtroumpf, line 19: If possible the algolia logic should remain in the autocomplete algolia component

libs/invitation/src/lib/form/user/user.component.ts
```diff
@@ -16,6 +16,13 @@ import { Validators } from '@angular/forms';
 })
 export class UserComponent {
   @Input() docId: string;
+
+  @Input() set invitations(inv) {
```
RemcoSimonides, line 20: I don't like what I have to do here. Invitations come in this component a little later. Is there a way to make this prettier?

libs/invitation/src/lib/form/user/user.component.html
```diff
@@ -1,6 +1,13 @@
...
-  <algolia-chips-autocomplete fxFlex index="user" displayWithPath="email" [form]="form" label="Invite guests"
-    [separators]="separators" placeholder="Enter guest email">
+  <algolia-chips-autocomplete fxFlex 
+    index="user" 
+    displayWithPath="email"
+    [form]="form"
+    label="Invite guests"
+    [separators]="separators"
+    placeholder="Enter guest email"
+    [filters]="_invitations | mapDeepKey:'toUser.email'"
```
RemcoSimonides, line 9: As invitations is an array, I created a mapDeepKey pipe to handle arrays.

libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
```diff
@@ -72,7 +73,21 @@ export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
 
   @Input() separators = [ENTER, COMMA, SEMICOLON];
 
-  @Input() filters: (string | string[])[] = [];
+  @Input() set filters(filters: (string | string[])[]) {
+    filters.map(filter => {
+      if (!this.unique || filter.includes(':')) {
+        this._filters.push(filter);
+      } else {
```
RemcoSimonides, line 80: Ahh didn't test this with a `string[][]` yet.

libs/invitation/src/lib/form/user/user.component.ts
GrandSchtroumpf: Don't use a setter. Use the normal @Input(). And verify in the pipe if the value exist.

libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
GrandSchtroumpf: I don't understand what you're trying to achieve here...

libs/utils/src/lib/pipes/deep-key.pipe.ts
```diff
@@ -8,8 +8,15 @@ export class DeepKeyPipe implements PipeTransform {
   }
 }
 
+@Pipe({ name: 'mapDeepKey' })
+export class MapDeepKeyPipe implements PipeTransform {
+  transform(value: Object[], deepKey: string) {
+    return value.map(obj => deepKey.split('.').reduce((result, key) => result[key], obj));
```
GrandSchtroumpf, line 14: Update the deepKey pipe to support array instead of creating a new pipe

GrandSchtroumpf (changes requested):

libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
```diff
@@ -130,4 +154,8 @@ export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
     this.input.nativeElement.value = value;
     this.form.removeAt(index);
   }
+
+  private getFilter(filter: string): string {
```
GrandSchtroumpf, line 158: Personally I would have done something like : 
```typescript
@Input() filters: string[];

private getFilter() {
  const format = filter => this.unique && !filter.includes(':') ? `${this.displayWithPath}:-${filter}` : filter;
  return this.filter.map(format);
}
```
It would be much cleaner in the end.

GrandSchtroumpf (changes requested):

libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
RemcoSimonides: Looks good but doesn't work. `@Input() filters` first receives null object because a the passed down `invitations | deepkey:'toUsers.email'` is first null and then gets the value. So I can't call `getFilter()` the way you'd make in ngOnInit(). I am curious if you have a solution for this :)

If you don't have any; check my last commit in which I remove support for the matrix ;)

GrandSchtroumpf: Just had a default value : 
```typescript
const format = filter => this.unique && !filter.includes(':') ? `${this.displayWithPath}:-${filter}` : filter;
return this.filter?.map(format) || [];
```

libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
```diff
@@ -112,8 +118,16 @@ export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
   add(value: any) {
     if (!!value) {
       if (typeof value === 'string') {
-        splitValue(value, this.separators).forEach(v => this.form.add(v))
+        splitValue(value, this.separators).forEach(v => {
```
GrandSchtroumpf, line 121: You can factorize that : 
```typescript
const values = typeof value === 'string' ? splitValue(value, this.separators) : [value];
for (const v of values) {
  if (this.unique && !!v) this.addedFilters.push(v);
  this.form.add(v);
}
```
Or something like that.

libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
```diff
@@ -110,15 +117,14 @@ export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
...
-      this.input.nativeElement.value = '';
-      this.searchCtrl.setValue(null);
+    const values = typeof value === 'string' ? splitValue(value, this.separators) : [value];
+    for (const v of values) {
+      if (this.unique && !!v) this.addedFilters.push(v[this.displayWithPath]);
+      this.form.add(v);
     }
+
+    this.input.nativeElement.value = '';
+    this.searchCtrl.setValue(null);
```
GrandSchtroumpf, line 127: Normally you could just do that `this.searchCtrl.reset()` without the need to set the value on the `input.nativeElement`. I think.

libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
```diff
@@ -110,15 +117,14 @@ export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
...
-      if (typeof value === 'string') {
-        splitValue(value, this.separators).forEach(v => this.form.add(v))
-      } else {
-        this.form.add(value);
-      }
-      this.input.nativeElement.value = '';
-      this.searchCtrl.setValue(null);
+    const values = typeof value === 'string' ? splitValue(value, this.separators) : [value];
+    for (const v of values) {
+      if (this.unique && !!v) this.addedFilters.push(v[this.displayWithPath]);
```
GrandSchtroumpf, line 122: what if there is no `displayWithPath` or `v[this.displayWithPath]` doesn't exist ?

GrandSchtroumpf (approved):

libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
RemcoSimonides: Sadly this.searchCtrl doesn't update the input field by default :/

GrandSchtroumpf (approved):

PR #3751, opened 2020-09-22
Move end date with changing start date

RemcoSimonides:
To do in issue #3674 

"When we change start date of an event, end date should adapt: event should stay the same duration. E.g. Start date 09/15/20 10:00 am / End date 09/15/20 11:00 am > start date changed to 09/16/20 11:00 am, end date should be automatically updated to 09/16/20 noon"

### DIFF
--- a/libs/event/src/lib/layout/edit/edit.component.ts
+++ b/libs/event/src/lib/layout/edit/edit.component.ts
@@ -1,11 +1,11 @@
-import { Component, ChangeDetectionStrategy, Input, OnInit, OnDestroy } from '@angular/core';
+import { Component, ChangeDetectionStrategy, Input, OnInit } from '@angular/core';
 import { Router, ActivatedRoute } from '@angular/router';
 import { EventForm } from '../../form/event.form';
 import { EventService } from '../../+state/event.service';
 import { MEETING_MAX_INVITATIONS_NUMBER } from '../../+state/event.firestore';
 import { Invitation }  from '@blockframes/invitation/+state/invitation.model';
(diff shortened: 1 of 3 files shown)


libs/event/src/lib/layout/edit/edit.component.ts
```diff
@@ -57,4 +62,16 @@ export class EventEditComponent implements OnInit {
     this.service.remove(this.form.value.id);
   }
 
+  onEventChange(key: 'start' | 'end') {
```
GrandSchtroumpf, line 65: After seeing the final code I think having two method `onStartChange` & `onEndChange` would be easier to understand.

libs/event/src/lib/layout/edit/edit.component.ts
```diff
@@ -57,4 +62,16 @@ export class EventEditComponent implements OnInit {
...
+    const { start, end } = this.form.value;
+    if (end.getTime() - start.getTime() <= 0) {
+      const keyToUpdate = key === 'start' ? 'end' : 'start';
+      const time = this.form.value[key].getTime();
+      const date = new Date(key === 'start' ? time + this.duration : time - this.duration);
+      this.form.get(keyToUpdate).setValue(date);
+    } else {
+      // Don't reuse the start & end here because value might have changed
+      this.duration = end.getTime() - start.getTime();
```
GrandSchtroumpf, line 74: Looks like you're reusing end & start.

GrandSchtroumpf (approved):

Issue #3741, opened 2020-09-21
Movie form debug / improvement 2/2


Jackseed:
## General
- [ ] When leaving the form, user should be redirected to dashboard movie view `/c/o/dashboard/title/o4y6fD3qJDjQIPCsyq8k/main`
- [ ] I've clicked on the edit button on storyline elements, it leads to the right page but there is no "Back" & "Next" buttons
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/843c182b-fc61-4eb2-9782-38268432caef)

## Main
- [ ] Homogenize image ratio for poster > "3:4" instead of 3/4
- [ ] 4 digit validator doesn't turn green if I:
1. Enter a 6 digit number
2. click out of the input
3. click back in the input and edit my number to 4 digits
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/c3f046a9-0ad2-41ca-b094-80a6561427d7)
- [ ] Other genre placeholder cut before the end
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/6e917262-2f3c-4ac1-94d0-e7ce176f1fe4)

- [ ] Director table, check with @stephanie-cascade8 but I guess that columns should have a minimum width & some padding / margin so they are not sticked to one another
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/6f21bb0f-342b-4d03-ad7b-69a18ce6ceb1)
vs
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/c8a7ead9-f89d-418d-bac1-1f96b5dd1b37)
- [ ] Table should also have horizontal scroll for too long a table
- [ ] Null value on numbers shouldn't be displayed - possible not show any number when null?

## Production information
- [ ] Got the error `TypeError: control.registerOnChange is not a function` when entering the empty page
- [ ] Prod company list misses a divider
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/a57605c6-3cb6-4be4-83f3-e5eaffb4bef4)
vs 
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/ae30fa77-99a7-461a-bc6e-cc572d774ad8)
[...]

PR #3753, opened 2020-09-22
worked on debugging movie form

fritzschoff:
resovles #3741


libs/ui/src/lib/form/display-name/display-name.component.html
```diff
@@ -1,10 +1,10 @@
 <form [formGroup]="form" fxLayout="row" fxLayoutAlign="space-between center" fxLayoutGap="24px">
   <mat-form-field appearance="outline">
-    <mat-label>First Name</mat-label>
+    <mat-label>{{ labelSuffix }}&nbsp;First Name</mat-label>
```
GrandSchtroumpf, line 3: Why not just putting a normal space ?

libs/ui/src/lib/tunnel/layout/layout.component.ts
```diff
@@ -14,9 +14,10 @@ import { Router, NavigationEnd } from '@angular/router';
  * @param arithmeticOperator plus or minus
  */
 function getPage(steps: TunnelStep[], url: string, arithmeticOperator: number): string {
+
   const allRoutes = steps.map(({ routes }) => routes.map(r => r.path));
   const allPath = allRoutes.flat();
-  const current = url.split('/').pop();
+  const current = url.includes('#') ? url.split('#')[0].split('/').pop() : url.split('/').pop();
```
GrandSchtroumpf, line 20: Maybe create a small function with a good naming that do it. Here I don't understand what it does, and why.

GrandSchtroumpf (approved):

libs/ui/src/lib/form/display-name/display-name.component.html
fritzschoff: not respected in the mat-label component

libs/ui/src/lib/tunnel/layout/layout.component.ts
fritzschoff: okay, I will. If you come from summary and got a fragment in your url, the next and prev button won't show up in the tunnel layout

PR #3761, opened 2020-09-22
Feat(e2e): Create a utility function to auto-fill forms.

xmano:
Feature to fill forms. WIP.
Testing with Org create and use it for movies later.

Code works with Org creation use-case.


libs/e2e/src/lib/utils/functions.ts
```diff
@@ -56,3 +56,63 @@ export function getTomorrowDay(date: Date) {
...
+}
+
+interface FormData {
+  name: string;
+  id: string;
+  testId: string;
+  value: string;
+}
+
+export function setForm(selector: string, formOpt: FormOptions) {
```
phpgeekfr, line 73: Can you add a comment to explain how it work ?

phpgeekfr (approved):

libs/e2e/src/lib/utils/functions.ts
xmano: Added and tested in headless mode as well.  (52s)

image

PR #3763, opened 2020-09-22
New sales pitch page

Mussuranae:
Sales Pitch page done

### DIFF
--- a/libs/movie/src/lib/movie/+state/movie.firestore.ts
+++ b/libs/movie/src/lib/movie/+state/movie.firestore.ts
@@ -120,7 +120,7 @@ export interface MoviePromotionalElements {
   notes: string,
   presentation_deck: string,
   promo_reel_link: string,
-  sales_pitch: MovieSalesPitch,
+  salesPitch: MovieSalesPitch,
   scenario: string,
   screener_link: string,
   still_photo: Record<string, string>,
@@ -274,7 +274,7 @@ export interface MovieExpectedPremiere extends MovieExpectedPremiereRaw<Date> {}
 export interface MovieSalesPitch {
   description: string,
   link: string
-  pitchFile: string, // hosted media
+  file: string, // hosted media
 }
 
 export interface MovieGoalsAudience {
--- a/libs/movie/src/lib/movie/+state/movie.firestore.ts
+++ b/libs/movie/src/lib/movie/+state/movie.firestore.ts
@@ -120,7 +120,7 @@ export interface MoviePromotionalElements {
   notes: string,
   presentation_deck: string,
   promo_reel_link: string,
-  // sales_pitch: MovieSalesPitch,
+  sales_pitch: MovieSalesPitch,
   scenario: string,
   screener_link: string,
   still_photo: Record<string, string>,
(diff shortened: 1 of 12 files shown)


libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -120,7 +120,7 @@ export interface MoviePromotionalElements {
   notes: string,
   presentation_deck: string,
   promo_reel_link: string,
-  // sales_pitch: MovieSalesPitch,
+  sales_pitch: MovieSalesPitch,
```
GrandSchtroumpf, line 123: use camelCase for new variables. Except if there is a convention of naming I'm not aware of.

libs/movie/src/lib/movie/+state/movie.firestore.ts
```diff
@@ -274,7 +274,7 @@ export interface MovieExpectedPremiere extends MovieExpectedPremiereRaw<Date> {}
 export interface MovieSalesPitch {
   description: string,
   link: string
-  salesPitch: string, // hosted media
+  pitchFile: string, // hosted media
```
GrandSchtroumpf, line 277: why not just "file" ? we're already in the salesPitch. What is supposed to be this file ?

libs/movie/src/lib/movie/form/promotional-elements/sales-pitch/sales-pitch.component.html
```diff
@@ -0,0 +1,62 @@
+<tunnel-page>
+  <h4 title>Media</h4>
+  <h1 subtitle>Sales Pitch</h1>
+  <ng-container tunnel-form>
+
+    <section id="sales-pitch">
+      <h3>Sales Pitch</h3>
+      <article fxLayout="column">
```
GrandSchtroumpf, line 8: This article is useless. Either you wrap each form + p in an article. Or you remove the article

libs/movie/src/lib/movie/form/promotional-elements/sales-pitch/sales-pitch.component.html
```diff
@@ -0,0 +1,62 @@
...
+
+    <section id="goals">
+      <h3>Audiences & Goals</h3>
+      <bf-form-list [form]="form.goals" buttonText="Add a Goal and Target">
+        <ng-template itemRef let-goals>
+          {{ goals.target }} - {{ goals.goal }}
+        </ng-template>
+
+        <ng-template formView let-goalsForm>
+          <article fxLayout="row" fxLayoutGap="24px" [formGroup]="goalsForm">
```
GrandSchtroumpf, line 45: use a form instead of an article maybe

GrandSchtroumpf (approved):

libs/movie/src/lib/movie/+state/movie.firestore.ts
Mussuranae: the sales pitch document who presents the project

libs/movie/src/lib/movie/+state/movie.firestore.ts
Mussuranae: No it was only because all promotional were like that, it's the only on the fields with documents, I will change it.

Issue #3708, opened 2020-09-16
E2E: Cypress test runner does not detect network changes correctly


xmano:
To improve performance of testing need to have a way to detect route changes correctly in the apps.

Needs investigation.

1. Network requests when done by nx serve with normal browser is faster than cypress network requests in automation tests
 See:
https://github.com/cypress-io/cypress/issues/6388

2. Test runner is unable to detect changes in location (routes) and update.

![image](https://user-images.githubusercontent.com/8731418/93333572-207f5580-f841-11ea-94bb-355da4fb7e57.png)

Even after timeout of 25sec element is not found, although the results are returned for API request.

xmano:
Really annoying aspect is the inability to detect the right event / way until page loads. Giving a timeout in fact really delays the test!

`  /** Connection returns to the home page of the application in which you are. You have to create it. */
  public clickSignIn() {
    cy.server()
    cy.route('POST', '/identitytoolkit/**').as('postLogin')
    cy.get('auth-signin-form button[type=submit]', {timeout: 3000}).click();
    cy.wait('@postLogin').then((xhr) => {
      //cy.get('catalog-home', {timeout: 60000});
    })
    cy.location('pathname', {timeout: 90000})
      .should('include', 'dashboard/home');
  }
`

There is an event but we cannot call other cy commands within it.. argh!

`
  await cy.on('url:changed', (url) => {
    console.log(url);
    urlNew = url;
    //cy.wrap(urlNew).as('urlNew');
    // cy.get('catalog-home', {timeout: 60000});
    // const p3 = new HomePage();
  
    // // Navigate to movie-tunnel-main
    // const p4: TitlesListPage = TitlesListPage.navigateToPage();
    // const p5: StartTunnelPage = p4.clickAdd();
    // const p6: TunnelMainPage = p5.clickBegin();
  })
`

xmano:
For network requests following references might help:

https://github.com/cypress-io/cypress/issues/2374#issuecomment-644064939

https://github.com/firebase/firebase-js-sdk/issues/1674#issuecomment-587710516

Blacklisting sites can help:

https://www.jondjones.com/frontend/javascript/testing/how-to-stop-your-cypress-tests-timing-out/

xmano:
Related with XHR and UI Update.. Cypress cannot detect the UI element:

![image](https://user-images.githubusercontent.com/8731418/93503921-5f430780-f936-11ea-8879-b6b87d8579ab.png)

[...]

PR #3785, opened 2020-09-23
suggestions for organization-creation.spec.ts

phpgeekfr:
suggestions for organization-creation.spec.ts

Sans titre


xmano (commented):
While we can put a delay at places and make the test pass, it is just a workaround.

cy.wait() is considered anti-pattern and its use must be restricted to places where navigating out will cause actions to be cancelled or incorrect UI operation.  

The real issue still persists when I checked with original deploy e2e test logs it gets completed in 34s and your test is taking about 79s. 

For this reason I have opened the issue:
https://github.com/blockframes/blockframes/issues/3708

Lets discuss the correct way forward.

xmano:
I will submit a PR which eliminates waits and yet it is reliable. I think it could be the answer to other E2E problem. Please have a look at that.

PR #3799, opened 2020-09-25
Feat 2586 - upload heavy files on storage

phpgeekfr:
Feat #2586 - upload heavy files on storage


apps/backend-functions/src/player.ts
```diff
@@ -81,14 +78,24 @@ export const getPrivateVideoUrl = async (
     };
   }
 
-  if (!movie.hostedVideo) {
+  if (!data.jwPlayerId) {
+    // no jwPlayerId in request, we assume user want to see the main video (screener)
+    if (!movie.promotional.videos?.screener?.jwPlayerId) {
+      return {
+        error: 'NO_VIDEO',
+        result: 'The creator of the movie hasn\'t uploaded any video for you to watch'
```
pldespaigne, line 86: could be cool to add the `eventId` or `movieId` so that we have hint of what's going wrong in the error message, so we can have a look later in the db.

apps/backend-functions/src/player.ts
```diff
@@ -81,14 +78,24 @@ export const getPrivateVideoUrl = async (
...
+      };
+    } else {
+      data.jwPlayerId = movie.promotional.videos?.screener?.jwPlayerId;
+    }
+  } else if (!isJwplayerIdBelongingToMovie(data.jwPlayerId, movie)) {
     return {
       error: 'NO_VIDEO',
-      result: `The creator of the movie hasn't uploaded any video for you to watch`
+      result: 'The video you are looking for doest not exists'
```
pldespaigne, line 94: same

pldespaigne (approved):

PR #3811, opened 2020-09-28
Fix invite members for organization

RemcoSimonides:
It wasn't possible to invite a member to an organization anymore. Old code is also refactoring to use the new bf-form-list component instead.

### DIFF
--- a/libs/organization/src/lib/organization/components/member-add/member-add.component.scss
+++ b/libs/organization/src/lib/organization/components/member-add/member-add.component.scss
@@ -1,8 +1,9 @@
 @import '~@angular/flex-layout/mq';
 
 .email {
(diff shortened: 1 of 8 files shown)


libs/organization/src/lib/organization/components/member-add/member-add.component.scss
```diff
@@ -1,8 +1,9 @@
 @import '~@angular/flex-layout/mq';
 
 .email {
```
fritzschoff, line 3: remove class and just do 
```
bf-form-list {
mat-form-field{
width:100%
 }
}
``` 
classes like theses get easily lost and out of consistency.

fritzschoff (approved):

PR #3814, opened 2020-09-28
Audience & goals changes

Mussuranae:
Update form and summary component to match new design on sales pitch page

### DIFF
--- a/libs/movie/src/lib/movie/+state/movie.model.ts
+++ b/libs/movie/src/lib/movie/+state/movie.model.ts
@@ -93,9 +93,9 @@ export function createMovie(params: Partial<Movie> = {}): Movie {
 
     ...params,
     banner: params.banner ?? '',
+    audience: createAudienceGoals(params.audience),
     estimatedBudget: createRange<number>(params.estimatedBudget),
     expectedPremiere: createExpectedPremiere(params.expectedPremiere),
-    goals: createAudienceGoals(params.goals),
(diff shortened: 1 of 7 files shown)


libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -312,8 +312,8 @@ export function createExpectedPremiere(params: Partial<MovieExpectedPremiere> =
 
 export function createAudienceGoals(params: Partial<MovieGoalsAudience> = {}): MovieGoalsAudience {
   return {
-    target: '',
-    goal: '',
+    target: [],
```
GrandSchtroumpf, line 315: Maybe targetS with a "s". Same for goalS

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -96,6 +95,7 @@ export function createMovie(params: Partial<Movie> = {}): Movie {
     banner: params.banner ?? '',
     estimatedBudget: createRange<number>(params.estimatedBudget),
     expectedPremiere: createExpectedPremiere(params.expectedPremiere),
+    goals: createAudienceGoals(params.goals),
```
GrandSchtroumpf, line 98: what about "audience" ? Now that it's not an array anymore having "goals" doesn't make so much sense to me...
If it requires a migration don't bother.

GrandSchtroumpf (approved):

PR #3827, opened 2020-09-30
prevent ref.split('/')[0] to equal ''

phpgeekfr:
prevent ref.split('/')[0] to equal ''


apps/backend-functions/src/media.ts
```diff
@@ -120,6 +120,7 @@ async function isAllowedToAccessMedia(ref: string, uid: string): Promise<boolean
 }
 
 function getPathInfo(ref: string) {
+  ref = ref.startsWith('/') ? ref.slice(1) : ref;
   const refParts = ref.split('/');
```
GrandSchtroumpf, line 124: Maybe `ref.split('/').filter(v => !!v)` would be more explicit

GrandSchtroumpf (approved):

PR #3839, opened 2020-10-01
build(backend-functions): create eu-based function resource locations

george43g:
- identify correct string for best function location
- create additional resource locations in one line
- update type imports in files that rely on `functions`

### DIFF
--- a/apps/backend-functions/src/internals/firebase.ts
+++ b/apps/backend-functions/src/internals/firebase.ts
@@ -1,7 +1,7 @@
 import * as firebaseFunctions from 'firebase-functions';
 import * as admin from 'firebase-admin';
 
-export const functions = firebaseFunctions.region('us-central1', 'europe-west1')
+export const functions = firebaseFunctions.region('europe-west1')
 
 export type DocumentReference = admin.firestore.DocumentReference;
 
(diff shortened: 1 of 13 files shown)


apps/backend-functions/src/internals/firebase.ts
```diff
@@ -16,7 +18,7 @@ export const serverTimestamp = admin.firestore.FieldValue.serverTimestamp;
 export const getBackupBucketName = (): string => backupBucket;
 export const getStorageBucketName = (): string => storageBucket;
 
-export { admin, functions };
+export { admin };
```
GrandSchtroumpf, line 21: An idea why we reexport admin ?

apps/backend-functions/src/internals/firebase.ts
```diff
@@ -1,12 +1,14 @@
-import * as functions from 'firebase-functions';
+import * as firebaseFunctions from 'firebase-functions';
```
GrandSchtroumpf, line 1: I would do : 
```typescript
import { region, config } from 'firebase-functions';

export const function = region(..);

File: apps/backend-functions/src/main.ts [hunk 3]
```diff
@@ -33,7 +34,7 @@ import { skipInMaintenance } from '@blockframes/firebase-utils';
 const heavyConfig = {
```
GrandSchtroumpf, line 34: Couldn't you type it as : `const heavyConfig: RuntimeOptions = { ... }` ?
It's a safer typinig.

GrandSchtroumpf (approved):

GrandSchtroumpf (approved):

PR #3841, opened 2020-10-01
Statics data

Mussuranae:
Delete cast_role and move member_status into const

### DIFF
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0038.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0038.ts
@@ -1,17 +1,28 @@
 import { Firestore } from '@blockframes/firebase-utils';
-import { createHostedVideo } from '@blockframes/movie/+state/movie.model';
+import { createHostedVideo, Credit } from '@blockframes/movie/+state/movie.model';
 import { MovieDocument } from 'apps/backend-functions/src/data/types';
 
-// Replace the old value for unitBox in box office
+
+/**
+ * Replace the old value for unitBox in box office
+ * Update the status for cast, crew and director
+ * Update the screener with video uploaded
+*/
 export async function upgrade(db: Firestore) {
   const movies = await db.collection('movies').get();
   const batch = db.batch();
 
-  movies.docs.map(movieDoc => {
+  const updates = movies.docs.map(movieDoc => {
     const data = movieDoc.data();
 
     const jwPlayerId = data.hostedVideo;
     delete data.hostedVideo;
+
+    const newCast = data.cast.map(person => {
+      delete person.role;
+      person.status = !!person.status ? updateMemberStatus(person.status) : '';
+      return person;
+    })
 
     const newData = {
       ...data,
@@ -29,7 +40,10 @@ export async function upgrade(db: Firestore) {
         } else {
           return { ...box }
         }
-      })
+      }),
+      cast: data.cast.length ? newCast : data.cast,
+      crew: updateMember(data.crew),
+      directors: updateMember(data.directors),
     } as MovieDocument;
 
     if (jwPlayerId) {
@@ -43,5 +57,28 @@ export async function upgrade(db: Firestore) {
   })
 
   console.log('Movie updated.')
+  await Promise.all(updates);
   await batch.commit();
+}
+
+
+
+// Update the status of directors and crew
+function updateMember(member: Credit[]) {
+  if (member.length) {
+    member.forEach(person => {
+      person.status = !!person.status ? updateMemberStatus(person.status) : '';
+    })
+  }
+  return member;
+}
+
+// Correct typo from the old static model
+function updateMemberStatus(status: string) {
+  if (status === 'loosely-attached') {
+    return 'looselyAttached';
+  }
+  else {
+    return status
+  };
 }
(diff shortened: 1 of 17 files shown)


libs/import/src/lib/components/view-extracted-elements/movies/view-extracted-movies.component.ts
```diff
@@ -417,7 +417,7 @@ export class ViewExtractedMoviesComponent implements OnInit {
 
         // CREDITS (Principal Cast)
         if (spreadSheetRow[SpreadSheetMovie.cast]) {
-          movie.cast = formatCredits(spreadSheetRow[SpreadSheetMovie.cast], this.separator, '\\s+', 'CAST_ROLES');
+          movie.cast = formatCredits(spreadSheetRow[SpreadSheetMovie.cast], this.separator, '\\s+');
```
phpgeekfr, line 420: I'm not sure for this one.  If CAST_ROLES is removed, it will try to parse filmography instead

libs/utils/src/lib/spreadsheet/format.ts
```diff
@@ -28,9 +28,6 @@ export function formatCredit(str: string, separator: string = '\\s+', thirdItemT
         case 'PRODUCER_ROLES':
           role = getCodeIfExists(thirdItemType, roleName as ExtractCode<'PRODUCER_ROLES'>);
           break;
```
phpgeekfr, line 30: I think we need to keep the `case 'CAST_ROLES':` can you keep it and make it work with the new constant ?

libs/firebase-utils/src/lib/firestoreMigrations/0038.ts
```diff
@@ -1,18 +1,29 @@
...
+ * Replace the old value for unitBox in box office
+ * Update the status for cast, crew and director
+ * Update the screener with video uploaded
+*/
 export async function upgrade(db: Firestore) {
   const movies = await db.collection('movies').get();
   const batch = db.batch();
 
-  movies.docs.map(movieDoc => {
+  const updates = movies.docs.map(movieDoc => {
```
phpgeekfr, line 15: Why did you added `const updates = ` and ` await Promise.all(updates);` ?  I think migration was working without before and 0037 does not have it.

libs/firebase-utils/src/lib/firestoreMigrations/0038.ts
```diff
@@ -29,7 +40,10 @@ export async function upgrade(db: Firestore) {
         } else {
           return { ...box }
         }
-      })
+      }),
+      cast: data.cast.length ? newCast : data.cast,
+      crew: updateMember(data.crew),
```
phpgeekfr, line 45: rename to updateMembers since  it expect a credit[]

libs/firebase-utils/src/lib/firestoreMigrations/0038.ts
```diff
@@ -43,5 +57,28 @@ export async function upgrade(db: Firestore) {
...
 
   console.log('Movie updated.')
+  await Promise.all(updates);
   await batch.commit();
 }
+
+
+
+// Update the status of directors and crew
+function updateMember(member: Credit[]) {
```
phpgeekfr, line 67: member => members  ( this is an array)

libs/firebase-utils/src/lib/firestoreMigrations/0038.ts
```diff
@@ -43,5 +57,28 @@ export async function upgrade(db: Firestore) {
...
+  if (member.length) {
+    member.forEach(person => {
```
phpgeekfr, line 69: It looks like you are modifying member array by reference, did you checked that member array was updated as intended? I'm asking this because this is not the method we usually chose to work with, we mostly use map

phpgeekfr: You could maybe simply do  
```
function updateMembers(members: Credit[]) : Credit[] {
 return members.map(person => {
   person.status = !!person.status ? updateMemberStatus(person.status) : '';
   return person
  })
}
```

libs/firebase-utils/src/lib/firestoreMigrations/0038.ts
```diff
@@ -43,5 +57,28 @@ export async function upgrade(db: Firestore) {
...
+    })
+  }
+  return member;
+}
+
+// Correct typo from the old static model
+function updateMemberStatus(status: string) {
+  if (status === 'loosely-attached') {
+    return 'looselyAttached';
+  }
```
phpgeekfr, line 80: bad indent. 

you can do 
```
function updateMemberStatus(status: string) {
  return status === 'loosely-attached' ? 'looselyAttached': status
}
```

libs/firebase-utils/src/lib/firestoreMigrations/0038.ts
Mussuranae: It was working also with forEach, I encountered several month ago issue with the map and batch, so I tried at this moment the forEach. I rework it with map.

libs/utils/src/lib/spreadsheet/format.ts
Mussuranae: Aïe, my PR was about deleting this :/ Do you think we can create a new format function for cast ?

phpgeekfr (approved):

PR #3847, opened 2020-10-02
added participant component

OmarSebouai:
Participant Component - Issue#3758

This is the participant component. As seen in the last meeting, we are providing the participant panel in standalone.

Usage: 
```
<event-participants-nav [participants]="arrayOfParticipantConnected$ | async">
  <event-video
    [arrayOfParticipantConnected$]="arrayOfParticipantConnected$"
    [$participantConnectedDataSource]="$participantConnectedDataSource"
    [localParticipantConnected$]="localParticipantConnected$"
    [localPreviewTracks$]="localPreviewTracks$"
    [dominantParticipantForBuyer$]="dominantParticipantForBuyer$"
    [event]="event"
  ></event-video>
</event-participants-nav>
```

image


libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,40 @@
...
+        <button mat-icon-button (click)="drawer.close()">
+          <mat-icon svgIcon="cross"></mat-icon>
+        </button>
+      </div>
+      <div class="content">
+        <div *ngFor="let participant of participants" class="participants">
+          <div class="participant">
+            <div class="participant-info">
+              <div class="profile">
+                <span>{{getInitialFromParticipant(participant)}}</span>
```
GrandSchtroumpf, line 18: Never use method in for BoundText or inputs. It's trigger at every change detection. Use pure pipes instead that use memorization.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,40 @@
...
+              </div>
+              <div>
+                <caption class="mat-caption name">{{participant.firstName}} {{participant.lastName}}</caption>
```
GrandSchtroumpf, line 21: Caption is used to describe something like an image, a graph, a table... It doesn't seems to be the case here. Use material component or correct semantic like p, span, .. or headings.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,40 @@
...
+                <caption class="mat-caption">{{participant.organizationName}}</caption>
+              </div>
+            </div>
+<!--            <mat-icon svgIcon="close" color="warn" class="close"></mat-icon>-->
```
GrandSchtroumpf, line 25: Remove comment if you don't use.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,40 @@
...
+          </div>
+          <hr>
```
GrandSchtroumpf, line 27: use material component divider.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,40 @@
...
+        </div>
+      </div>
+      <div class="action">
```
GrandSchtroumpf, line 30: Use material component or correct semantic (footer, ...)

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,40 @@
...
+        <mat-icon svgIcon="add_member"></mat-icon>
+        <caption class="mat-caption">Invite participants</caption>
+      </div>
+    </div>
+  </mat-drawer>
+  <button mat-icon-button class="menu-button" (click)="drawer.toggle()" *ngIf="!drawer.opened">
```
GrandSchtroumpf, line 36: Use the right method. If the button is display only when drawer is closed then use drawer.open().

GrandSchtroumpf: Also extract the `*ngIf` logic in a parent `<ng-container *ngIf="!drawer.opened">` for better separation of concern.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,40 @@
...
+    <mat-icon svgIcon="group"></mat-icon>
+  </button>
+  <ng-content></ng-content>
```
GrandSchtroumpf, line 39: Wrap this into a `<mat-sidenav-content>`

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+  height: 100%;
+  width: 100%;
+  background-color: black;
+}
+
+.menu {
+    height: 100%;
+  pointer-events: none;
+  .sidenav {
+    width: 30%;
```
GrandSchtroumpf, line 13: Isn't it too big on large screen ?

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
+@import 'node_modules/@angular/flex-layout/mq';
```
GrandSchtroumpf, line 1: You're not using it. Remove it.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+}
+
+.menu {
+    height: 100%;
+  pointer-events: none;
+  .sidenav {
+    width: 30%;
+    min-width: 250px;
+    border: none;
+    background-color: #33305B;
```
GrandSchtroumpf, line 16: We should color based on the current theme. Check with Stéphanie if we can apply a color from the theme here.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+
+:host {
+  height: 100%;
+  width: 100%;
+  background-color: black;
+}
+
+.menu {
+    height: 100%;
+  pointer-events: none;
```
GrandSchtroumpf, line 11: why do you need to remove the point-events ? It'll break swipe effect on tablet...

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+  .sidenav {
```
GrandSchtroumpf, line 12: Use tag name instead of class for the selector.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+    width: 30%;
+    min-width: 250px;
+    border: none;
+    background-color: #33305B;
+    pointer-events: all;
+    div {
+      border: none;
+    }
+    .sidenav-content {
+      display: flex;
```
GrandSchtroumpf, line 22: use the `FlexLayoutModule` in the html : 
```html
<nav fxLayout="column">
```

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+.menu {
+    height: 100%;
+  pointer-events: none;
+  .sidenav {
+    width: 30%;
+    min-width: 250px;
+    border: none;
+    background-color: #33305B;
+    pointer-events: all;
+    div {
```
GrandSchtroumpf, line 18: Only apply style on element that needs it. Here you're removing the border of everything even the element doesn't have a border already.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+      border: none;
+    }
+    .sidenav-content {
+      display: flex;
+      flex-direction: column;
+      height: 100%;
+      .mat-icon {
+        margin-right: 5px;
```
GrandSchtroumpf, line 26: There is no tag with the class "mat-icon" in the html. Use tag selector when possible.
use multiple of 8 for the margin & padding.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+        margin-left: 5px;
+      }
+      .title {
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+        background-color: white;
```
GrandSchtroumpf, line 33: Use theme colors. 
By using material component (when possible) you won't have to bother about that because they are already wired to the theme.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+    }
+    .sidenav-content {
+      display: flex;
+      flex-direction: column;
+      height: 100%;
+      .mat-icon {
+        margin-right: 5px;
+        margin-left: 5px;
+      }
+      .title {
```
GrandSchtroumpf, line 29: Use flexLayout : 
```html
<header fxLayout fxLayoutAlign="space-between center">
```
Same each time you use flex or grid

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+        justify-content: space-between;
+        align-items: center;
+        background-color: white;
+        .title-info {
+          height: 10%;
+          min-height: 30px;
+          display: flex;
+          align-items: center;
+        }
+        .mat-icon {
```
GrandSchtroumpf, line 40: Don't change the size of mat-icon. We keep the same everywhere in the app.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+        margin-right: 5px;
+        margin-left: 5px;
+      }
+      .title {
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+        background-color: white;
+        .title-info {
+          height: 10%;
```
GrandSchtroumpf, line 35: I'm not comfortable using a % on height. I think if you use the right material component you won't need that anymore.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+          align-items: center;
+        }
+        .mat-icon {
+          width: 16px;
+        }
+      }
+
+
+      .content {
+        height: 80%;
```
GrandSchtroumpf, line 47: Why not taking the rest of the space instead of 80% ? 
If so use `flex: 1` in css or better `<tag fxFlex>`

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+        overflow: scroll;
```
GrandSchtroumpf, line 48: Why are you forcing the scroll ? And not displaying it.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+        &::-webkit-scrollbar {
+          display: none;
+        }
+        .participants {
+          margin: 10px;
```
GrandSchtroumpf, line 53: Use 16px (2 * 8)

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+            display: flex;
+            align-items: center;
+            justify-content: space-between;
+            .participant-info {
+              display: flex;
+              align-items: center;
+              justify-content: space-between;
+            }
+            .profile {
+              background-color: white;
```
GrandSchtroumpf, line 64: Use theme color

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+          .participant {
+            display: flex;
+            align-items: center;
+            justify-content: space-between;
+            .participant-info {
+              display: flex;
+              align-items: center;
+              justify-content: space-between;
+            }
+            .profile {
```
GrandSchtroumpf, line 63: There are component for that in the material library

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,40 @@
...
+        </div>
+        <button mat-icon-button (click)="drawer.close()">
+          <mat-icon svgIcon="cross"></mat-icon>
+        </button>
+      </div>
+      <div class="content">
+        <div *ngFor="let participant of participants" class="participants">
+          <div class="participant">
+            <div class="participant-info">
+              <div class="profile">
```
GrandSchtroumpf, line 17: We've got the avatar of the participant. It's better to use it. Check the `ImageReferenceModule` from `@blockframes/media` to display images. You can use a `*ngIf="participant.avatar; else initials"` to fallback to the initials.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+              width: 50px;
+              height: 50px;
+              border-radius: 50%;
+              margin-right: 10px;
+              text-align: center;
+              display: flex;
+              align-items: center;
+              justify-content: center;
+
+              span {
```
GrandSchtroumpf, line 74: I don't understand the benefit of this style if the container already apply center center. Only apply style that is required.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+                vertical-align: middle;
+                display: inline-flex;
+              }
+            }
+            .name {
+              color: white;
```
GrandSchtroumpf, line 80: Use theme colors

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+            }
+            .mat-caption {
```
GrandSchtroumpf, line 82: All the mat-caption are wrapping text node that are not affected by flex. Why using it ?

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+              display: flex;
+            }
+            .close {
```
GrandSchtroumpf, line 85: You commented out this tag. Remove unnecessary code.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+              }
+            }
+          }
+        }
+
+      }
+
+      .action {
+        opacity: 0.5;
+        background-color: #5C597B;
```
GrandSchtroumpf, line 99: Use the attribute "color" of the `mat-icon-button` directive

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+                cursor: pointer;
+              }
+            }
+          }
+        }
+
+      }
+
+      .action {
+        opacity: 0.5;
```
GrandSchtroumpf, line 98: Don't reduce opacity it reduce contrast which makes it difficult for color-blind people to find the action button.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
GrandSchtroumpf: Looks like an action button but there is not (click) event. Either semantic is wrong or feature is missing.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+      }
+      hr {
+        opacity: 0.5;
+      }
+    }
+  }
+
+  .menu-button {
+    position: absolute;
+    right: 0;
```
GrandSchtroumpf, line 120: top: 8px; right: 8px;

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,123 @@
...
+    pointer-events: all;
```
GrandSchtroumpf, line 121: Remove this part if you remove pointer event: none on container.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.ts
```diff
@@ -0,0 +1,29 @@
+import {Component, OnInit, ChangeDetectionStrategy, Input} from '@angular/core';
+
+import { Participant as IParticipantMeeting } from 'twilio-video';
```
GrandSchtroumpf, line 3: why creating an alias ?

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  ) { }
+
+  ngOnInit(): void {
+  }
+
+  /**
+   *
+   * @param participant
+   */
+  getInitialFromParticipant(participant){
```
GrandSchtroumpf, line 26: type your code and use a pipe instead.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ParticipantsNavComponent implements OnInit {
+
+  @Input() participants: IParticipantMeeting[];
+
+  constructor(
+  ) { }
+
+  ngOnInit(): void {
```
GrandSchtroumpf, line 19: Remove onInit and constructor if you don't use

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Component({
+  selector: 'event-participants-nav',
+  templateUrl: './participants-nav.component.html',
+  styleUrls: ['./participants-nav.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ParticipantsNavComponent implements OnInit {
+
+  @Input() participants: IParticipantMeeting[];
```
GrandSchtroumpf, line 14: We should get the user from Firestore instead to have access to the avatar.

GrandSchtroumpf (changes requested):
Please use material components, use tag for the selectors and don't apply unnecessary css.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -1,123 +1,67 @@
...
-          display: flex;
-          align-items: center;
+    header {
+      position: sticky;
+      z-index: 1;
+      top: 0;
+      background-color: var(--whiteBackground);
+      span {
+        span {
+          font-size: 0.7em;
```
GrandSchtroumpf, line 12: use angular material typography instead of overriding the size. I think a class="mat-caption" would do the job here : 
https://material.angular.io/guide/typography

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -1,123 +1,67 @@
...
         }
-        .mat-icon {
-          width: 16px;
+        mat-icon {
```
GrandSchtroumpf, line 14: Better set a padding on header : 
```html
header {
  padding: 0 8px;
}
```

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -1,123 +1,67 @@
...
       }
-
-
-      .content {
-        height: 80%;
-        overflow: scroll;
+    }
+    mat-list {
+      mat-list-item {
+        height: unset;
```
GrandSchtroumpf, line 21: Then don't set it ;)

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -1,123 +1,67 @@
...
+            font-weight: bold;
+            text-overflow: ellipsis;
+          }
+          .profile-organization {
+            font-size: 0.8em;
+          }
+
+          .avatar-text {
+            background-color: var(--whiteBackground);
+            width: 7vh;
```
GrandSchtroumpf, line 40: The matListAvatar set the height & width already : 
https://github.com/angular/components/blob/master/src/material/list/list.scss#L10

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -1,123 +1,67 @@
...
+            height: 7vh;
+            border-radius: 50%;
```
GrandSchtroumpf, line 42: This should be managed by the matListAvatar directive :
https://github.com/angular/components/blob/master/src/material/list/list.scss#L145

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -1,40 +1,35 @@
...
+        <mat-icon svgIcon="group"></mat-icon>
+        <span>List of participants</span>
+      </span>
+      <button mat-icon-button (click)="sidenav.close()">
+        <mat-icon svgIcon="cross"></mat-icon>
+      </button>
+    </header>
+    <mat-list role="list" fxLayout="column" fxLayoutAlign="space-around start">
+      <ng-container *ngFor="let participant of participants">
+        <mat-list-item>
```
GrandSchtroumpf, line 14: Use mat-list-item way of dealing with text & avatar: 
https://material.angular.io/components/list/overview#lists-with-avatars

```html
<mat-list-item>
  <span matListAvatar>...</span>
  <h3 matLine>{{participant.festivalData | displayName}}</h3>
  <span matLine>{{ participant.festivalData.organizationName }}</span>
</mat-list-item>
```
Here I use the displayName pipe that you can find in "@blockframes/utils/pipes" :
In the module:
```typescript
import { DisplayNameModule } from '@blockframes/utils/pipes';
```

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
OmarSebouai: It seems like there was a fixed set on the original mat-list-item (56px)

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
OmarSebouai: Will have a closer look into this and make a quick feedback. We didn't plan on implementing avatars at first but at this point, we're already into it =)

GrandSchtroumpf: You don't need to implement the avatar img right now. We can merge this PR without.
But by using the matListAvatar directive you use the theme & style of material (and blockframes). So it won't break the UI feeling.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
GrandSchtroumpf: Then let's keep the mat-list-item height. By keeping default css values as much as possible it makes it much easier to keep a UI harmony across all apps, and update it everywhere at once.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -7,53 +7,35 @@ mat-sidenav-container {
...
-            text-overflow: ellipsis;
-          }
-          .profile-organization {
-            font-size: 0.8em;
-          }
+        h3 {
+          font-size: 1em;
+          font-weight: bold;
+          text-overflow: ellipsis;
+          text-align: left;
```
GrandSchtroumpf, line 25: Is it really required ?

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -7,53 +7,35 @@ mat-sidenav-container {
       z-index: 1;
       top: 0;
       background-color: var(--whiteBackground);
```
GrandSchtroumpf, line 9: sorry I missed this one. whiteBackground might be a little aggressive on dark mode. You can use `var(--background-card)`. It should be white on light mode & black on dark mode.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.ts
```diff
@@ -11,5 +11,5 @@ import {IParticipantMeeting} from "@blockframes/event/components/meeting/+state/
 })
 export class ParticipantsNavComponent {
 
-  @Input() participants: IParticipantMeeting[];
+  @Input() participants$: Observable<IParticipantMeeting[]>;
```
GrandSchtroumpf, line 14: Never pass an observable as input you'll loose all the benefice of ChangeDetectionStrategy.OnPush. 
The async pipe should be given by the parent: 
```html
<event-meeting-participants-nav [participants]="participants | async">
```

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -7,53 +7,35 @@ mat-sidenav-container {
...
-          .profile-name {
-            font-size: 1em;
-            font-weight: bold;
-            text-overflow: ellipsis;
-          }
-          .profile-organization {
-            font-size: 0.8em;
-          }
+        h3 {
+          font-size: 1em;
```
GrandSchtroumpf, line 22: I think that you can remove this size. The h3 + matLine should do the job of making the first line bigger than the second.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.ts
OmarSebouai: Oh ok!

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.ts
```diff
@@ -0,0 +1,13 @@
+import {ChangeDetectionStrategy, Component, Input} from '@angular/core';
+import {IParticipantMeeting} from "@blockframes/event/components/meeting/+state/meeting.interface";
+
+
+@Component({
+  selector: 'event-meeting-participants-nav',
+  templateUrl: './participants-nav.component.html',
+  styleUrls: ['./participants-nav.component.scss']
```
GrandSchtroumpf, line 8: ah ! I missed this one : Use `changeDetection: ChangeDetectionStrategy.OnPush` to only run the change detection when the input change.

GrandSchtroumpf (approved):
Looks pretty good 👍. I left you one comment on the changeDetection that I missed. 
@pldespaigne will test the branch to see if everything work well, it would be cool if you can provide him a set of participant fixtures so he can test the component easily.
Sorry it took so much time for this one. Now it should go faster for the next ones.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.ts
OmarSebouai: True! Thanks for all your feedback, we'll make sure we take everything into account for the next push!

Btw, I added in the original comment the code we use to wrap the participant component and add participants.

OmarSebouai: It's weird. Somehow, the change isn't applied with the `changeDetection: ChangeDetectionStrategy.OnPush` line. But it does work when we remove it.

GrandSchtroumpf: Which changes are not applied ? `ChangeDetectionStrategy.OnPush` should only trigger the changes when the input is updated. This can happen when you pass an array for example as an input and you mutate it with `paraticipant.push()` in the parent component instead making an immutable change.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,31 @@
+<mat-sidenav-container>
+  <mat-sidenav #sidenav [position]="'end'" [mode]="'over'" [autoFocus]="true">
```
pldespaigne, line 2: you don't need to use brackets for constant primitives like strings, moreover the sidenav should be in `side` mode so that it doesn't close automatically when the user click outside

`<mat-sidenav #sidenav position="end" mode="over">`

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.html
```diff
@@ -0,0 +1,31 @@
...
+        <span>List of participants</span>
+      </span>
+      <button mat-icon-button (click)="sidenav.close()">
+        <mat-icon svgIcon="cross"></mat-icon>
+      </button>
+    </header>
+    <mat-list role="list" fxLayout="column" fxLayoutAlign="space-around start">
+      <ng-container *ngFor="let participant of participants">
+        <mat-list-item>
+          <span matListAvatar class="avatar-text ">{{participant | eventInitialParticipant}}</span>
```
pldespaigne, line 15: You haven't committed the `eventInitialParticipant` pipe, I was unable to review it and to test it's behavior

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.scss
```diff
@@ -0,0 +1,46 @@
+mat-sidenav-container {
```
pldespaigne, line 1: Please add `height: 100%` to the `mat-sidenav-container` so that your component will take the space of it's parent.
I have tested it and without `height: 100%` and your component is not visible.

libs/event/src/lib/components/meeting/participant-nav/participants-nav.component.ts
```diff
@@ -0,0 +1,13 @@
+import {ChangeDetectionStrategy, Component, Input} from '@angular/core';
+import {IParticipantMeeting} from "@blockframes/event/components/meeting/+state/meeting.interface";
+
+
+@Component({
+  selector: 'event-meeting-participants-nav',
```
pldespaigne, line 6: if the `participants` input is mandatory you can force it in the selector like that:

`selector: '[participants] event-meeting-participants-nav',`

package.json
```diff
@@ -122,6 +122,8 @@
     "rxjs": "~6.5.5",
     "saturn-datepicker": "^8.0.2",
     "tmp": "0.1.0",
+    "twilio": "3.49.2",
+    "twilio-video": "^2.7.2",
```
pldespaigne, line 126: please also add typing:

`npm i @types/twilio @types/twilio-video --save-dev`

pldespaigne (changes requested):
Your component is not part of any module, so right now it cannot be mounted in the app.
Please add a `participants-nav.module.ts` module in the same folder as the component.

This way it became self contained and can be mounted anywhere.

pldespaigne (approved):
SGTM 👍

PR #3859, opened 2020-10-05
Show multi day even at every day it is screening

RemcoSimonides:
To do issue #3674 

"Screening lists (/c/o/marketplace/event + /c/o/marketplace/organization/orgId/screening + /c/o/marketplace/title/movieId/screenings) display events based on their starting date but for multiple day screenings they should appear on every day of the screening.
For example: an event from 10/15 to 10/17 is currently shown only for the 15th while it should be displayed on 15th, 16th and 17th"

-- And --

"Images section title shouldn't be displayed if there is no stills on the film"

### DIFF
--- a/libs/utils/src/lib/pipes/filter-by-date.pipe.ts
+++ b/libs/utils/src/lib/pipes/filter-by-date.pipe.ts
@@ -48,8 +48,9 @@ export class FilterByDatePipe implements PipeTransform {
    * @param value A list to order by date
    * @param timeFrame 
    * @param key The key where to find the date value
+   * @param keyFinish The key where to find the end date value. If used, date found at key is used as starting date.
    */
-  transform(value: any[], timeFrame: TimeFrame, key: string = 'date') {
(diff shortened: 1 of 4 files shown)


libs/utils/src/lib/pipes/filter-by-date.pipe.ts
```diff
@@ -48,16 +48,22 @@ export class FilterByDatePipe implements PipeTransform {
...
     if (!Array.isArray(value)) {
       return value;
     }
     const { from, to, type } = timeFrame;
     const now = startOfDay(Date.now());
     const fromDate = add({ [type]: from }, now);
     const toDate = add({ [type]: to }, now);
-    return value.filter(v => v[key] >= fromDate && v[key] < toDate);
+    return value.filter(v => {
+      if (!!keyFinish) {
```
fritzschoff, line 62: no need to cast it into a boolean. optional parameters are undefined so you could just do `if (keyFinish)`

fritzschoff (approved):
lgtm

PR #3863, opened 2020-10-06
Fix dashboard title list

GrandSchtroumpf:
fix deep key pipe
update dashboard title list on festival
update dashboard title list on financiers
use mat-button instead of mat-stroked-button

### DIFF
--- a/libs/movie/src/lib/movie/pipes/analytics.pipe.ts
+++ b/libs/movie/src/lib/movie/pipes/analytics.pipe.ts
@@ -1,0 +1,23 @@
+import { NgModule, Pipe, PipeTransform } from "@angular/core";
+import { MovieAnalytics, MovieQuery } from "../+state";
+import { map } from "rxjs/operators";
+import { Observable } from "rxjs";
+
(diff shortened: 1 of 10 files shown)


libs/movie/src/lib/movie/pipes/analytics.pipe.ts
```diff
@@ -0,0 +1,23 @@
...
+import { MovieAnalytics, MovieQuery } from "../+state";
+import { map } from "rxjs/operators";
+import { Observable } from "rxjs";
+
+@Pipe({ name: 'getViews' })
+export class GetViewsPipe implements PipeTransform {
+  constructor(private query: MovieQuery) {}
+  transform(movieId: string): Observable<number> {
+    const getViews = (analytics?: MovieAnalytics) => {
+      return analytics?.movieViews.current.reduce((sum, event) => sum + event.hits, 0) || 0;
```
RemcoSimonides, line 11: I think you could use `??` instead of `||` to also catch null values

RemcoSimonides (approved):

PR #3876, opened 2020-10-06
update docker image used in CI

george43g:
- docker image temporarily updated until we get the blockframes org access

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ orbs:
 executors:
   blockframes-ci:
     docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -65,7 +65,7 @@ orbs:
 executors:
   blockframes-ci:
     docker:
-      - image: blockframes/ci:latest
+      - image: george43g/blockframes-ci:gcloud
```
phpgeekfr, line 68: Can you add a comment with issue saying that this is temp ?

phpgeekfr (approved):

PR #3882, opened 2020-10-06
fixed some release issues

fritzschoff:
ref #3674

### DIFF
--- a/apps/festival/festival/src/app/dashboard/title/list/list.component.ts
+++ b/apps/festival/festival/src/app/dashboard/title/list/list.component.ts
@@ -49,10 +49,10 @@ export class ListComponent implements OnInit, OnDestroy {
     this.titles$ = this.orgQuery.selectActive().pipe(
       switchMap(org => this.service.valueChanges(org.movieIds)),
       map(movies => movies.filter(movie => !!movie)),
-      map(movies => movies.filter(movie => movie.storeConfig.appAccess.financiers)),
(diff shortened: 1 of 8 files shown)


apps/festival/festival/src/app/dashboard/title/list/list.component.ts
```diff
@@ -49,10 +49,10 @@ export class ListComponent implements OnInit, OnDestroy {
     this.titles$ = this.orgQuery.selectActive().pipe(
       switchMap(org => this.service.valueChanges(org.movieIds)),
       map(movies => movies.filter(movie => !!movie)),
-      map(movies => movies.filter(movie => movie.storeConfig.appAccess.financiers)),
+      map(movies => movies.filter(movie => movie.storeConfig.appAccess.festival)),
       filter(movies => !!movies.length),
       tap(movies => {
-        !!movies.length ?
+        movies.length ?
```
RemcoSimonides, line 55: Bruce also fixed this one partially; https://github.com/blockframes/blockframes/pull/3877/files
But seems like you went one step further by removing `!!`. why remove !! btw?

RemcoSimonides (approved):

apps/festival/festival/src/app/dashboard/title/list/list.component.ts
fritzschoff: cause if you are using the ternary operator it will cast the value anyway into a boolean, no need for double cast. Also 0 is falsy and +1 is truthy

PR #3893, opened 2020-10-06
code is ahead db in production

phpgeekfr:
code is ahead db in production: new uploads are stored in main.banner.media for example but the expected for migration 31 was promotionnalElements.banner.media.


libs/firebase-utils/src/lib/firestoreMigrations/0031.ts
```diff
@@ -296,6 +296,9 @@ const changeResourceDirectory = async (
     } else if (ref.includes(`movies/${docId}/promotionalElements.banner.media`)) {
       oldRef = ref;
       newRef = `movies/${docId}/banner/${ref.split('/').pop()}`;
+    } else if (ref.includes(`movies/${docId}/main.banner.media`)) { // Code is ahead DB
```
GrandSchtroumpf, line 299: I'm not sure to get the comment. But it'll be deprecated after release, so better removing it (the comment).

GrandSchtroumpf (approved):

PR #3909, opened 2020-10-07
Statics data

Mussuranae:
Change model into constant :
- MOVIE_CURRENCIES
- PREMIERE_TYPE
- RELEASE_STATUS
- SCREENING_STATUS
- STORE_TYPE


libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -437,7 +438,7 @@ export class ViewExtractedContractsComponent implements OnInit {
       // Check if priceParts have at least two parts (amount and currency)
       if (priceParts.length >= 2) {
         const amount = parseInt(priceParts[0], 10);
-        const currency = getCodeIfExists('MOVIE_CURRENCIES', priceParts[1]);
+        const currency = getKeyFromValue('movieCurrencies', priceParts[1]) as MovieCurrencies;
```
phpgeekfr, line 441: getCodeIfExists was working with key or value as input. getKeyFromValue seems to be only checking value.

getCodeIfExists was built this was so we can find code from key **or** value allowing users to enter either the key or the corresponding value in the Excel sheet

libs/movie/src/lib/movie/form/main/main.component.html
```diff
@@ -93,7 +93,7 @@ <h3>General Information</h3>
         </mat-form-field>
 
         <ng-container *ngIf="form | hasStatus: ['development', 'shooting', 'post_production']">
-          <static-select scope="RELEASE_STATUS" type="model" [control]="form.release.get('status')" required fxFlex>
+          <static-select scope="screeningStatus" type="constant" [control]="form.release.get('status')" [withoutValues]="['tobedetermined']" required fxFlex>
```
phpgeekfr, line 96: It was RELEASE_STATUS but you changed it to screeningStatus

phpgeekfr: they are now the same ones?

libs/import/src/lib/components/view-extracted-elements/movies/view-extracted-movies.component.ts
```diff
@@ -255,7 +263,7 @@ export class ViewExtractedMoviesComponent implements OnInit {
```
phpgeekfr, line 263: did you made some tests to check that everything is ok ?

libs/utils/src/lib/common-interfaces/price.ts
```diff
@@ -134,7 +134,7 @@ export interface PaymentDocument extends PaymentRaw<Timestamp> { }
  * A factory function that creates Price
  */
 export function createPrice(price: Partial<Price> = {}): Price {
-  const defaultCurrency = getCodeIfExists('MOVIE_CURRENCIES', 'EUR');
+  const defaultCurrency = getKeyFromValue('movieCurrencies', 'EUR') as MovieCurrencies;
```
phpgeekfr, line 137: If possible, it would be cool to have `getKeyFromValue<MovieCurrencies>('movieCurrencies', 'EUR')` instead.

libs/utils/src/lib/form/forms/static-value.form.ts
```diff
@@ -18,3 +19,12 @@ export class FormStaticArray<S extends Scope> extends FormControl {
     super(value, [isSlugArrayValidator(scope), ...validators]);
   }
 }
+
+// Form for static constant until we delete every static model
```
phpgeekfr, line 23: maybe add an issue number to this comment

phpgeekfr (approved):
LTGM

libs/movie/src/lib/movie/form/main/main.component.html
Mussuranae: They are the same since the beginning, the only difference is that screeningStatus has one value more, but with the `[withoutValues]` attribute, I can easily skip it. So I though it was useless to have 2 statics that are almost the same.

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
Mussuranae: The fact here is the `getCodeIfExists` doesn't work with constants, I will try to reproduce it with constant.

Mussuranae: I used the `getKeyIfExists` function that was already inside the `helpers` file. Looks like the function I wanted already existed.

libs/utils/src/lib/common-interfaces/price.ts
Mussuranae: Deleted in favor of `getKeyIfExists`

libs/utils/src/lib/static-model/staticConsts.ts
```diff
@@ -116,18 +116,77 @@ const constants = {
...
+    boats: 'Boats',
+    hotels: 'Hotels',
+    // tVodStreaming: 'T-VOD, ER, DTR, Streaming',
+    // allRights: 'All rights',
+    // trains: 'Trains',
+    // remake: 'Remake',
+    // bookAdaptation: 'Book Adaptation',
+    // musicPublishing: 'Music Publishing',
+    // merchandising: 'Merchandising',
+    // ancillary: 'Ancillary'
```
phpgeekfr, line 171: can you remove this ? Or you will get it back ?

phpgeekfr (approved):

libs/utils/src/lib/static-model/staticConsts.ts
Mussuranae: I will ask to Vincent if we can remove it yes, and if not, we can filter in the component if needed

PR #3915, opened 2020-10-08
maintenance mode fix

phpgeekfr:


.circleci/config.yml
```diff
@@ -178,8 +178,8 @@ jobs:
           cp firestore.rules firestore.rules.regular &&
           cp firestore.maintenance.rules firestore.rules &&
           npx firebase deploy --only hosting:catalog,hosting:festival,hosting:crm,storage &&
-          npx firebase deploy --only firestore --force &&
-          npx firebase deploy --only functions --force
+          npx firebase deploy --only functions --force &&
```
phpgeekfr, line 181: switched firestore and functions because `npx firebase deploy --only firestore --force` will push maintenance.rules and we don't want maintenance.rules to be active before `upgrade` enables maintenance mode.

firestore.maintenance.rules
```diff
@@ -1,6 +1,20 @@
...
+
 service cloud.firestore {
   match /databases/{database}/documents {
-    allow read, write: if false;
+    match /{document=**} {
+      allow read, write: if false;
+    }
+
+    match /_META/_MAINTENANCE {
+      allow read: if true;
```
phpgeekfr, line 16: Allow users to read maintenance doc

libs/ui/src/lib/maintenance/maintenance.service.ts
```diff
@@ -1,17 +1,17 @@
...
+import { _isInMaintenance, IMaintenanceDoc, META_COLLECTION_NAME, MAINTENANCE_DOCUMENT_NAME } from '@blockframes/utils/maintenance';
 
 @Injectable({ providedIn: 'root' })
 export class MaintenanceService {
-  
-  // if document doesn't exist, we are not in the maintenance
-  isInMaintenance$ = this.db.doc<IMaintenanceDoc>(maintenancePath).valueChanges().pipe(
-    map(maintenanceDoc => maintenanceDoc ? isInMaintenance(maintenanceDoc) : false)
+
+  // if document doesn't exist, it means that there is something not normal, we force maintenance mode to true.
```
phpgeekfr, line 9: changed the behavior because we should always have a maintenance doc. If not => maintenanceMode is on because there is an error with DB

libs/firebase-utils/src/lib/maintenance.ts
```diff
@@ -37,29 +34,22 @@ export async function endMaintenance() {
...
-  const now = admin.firestore.Timestamp.now();
+export async function isInMaintenance(delay = EIGHT_MINUTES_IN_MS): Promise<boolean> {
+  try {
+    const ref = maintenanceRef();
+    const doc = await ref.get();
 
-  if (endedAt) {
-    return endedAt.toMillis() + delay > now.toMillis();
-  }
+    // if document doesn't exist, it means that there is something not normal,
```
phpgeekfr, line 42: changed the behavior because we should always have a maintenance doc. If not => maintenanceMode is on because there is an error with DB

PR #3916, opened 2020-10-08
Add summary to the campaign tunnel

GrandSchtroumpf:


libs/organization/campaign/src/lib/+state/campaign.model.ts
```diff
@@ -34,7 +34,8 @@ export function createPerk(params: Partial<Perk> = {}): Perk {
     title: '',
     minPledge: 0,
     amount: {
-      total: 0
+      total: 0,
```
fritzschoff, line 37: sure about the default value? should be the same issue with box office

fritzschoff: just make it optional

fritzschoff (approved):

libs/organization/campaign/src/lib/+state/campaign.model.ts
GrandSchtroumpf: It won't be the same issue because it's inside an array. So by default perks is an empty array. Wihle boxOffice is an object, so all keys always exist.

PR #3921, opened 2020-10-09
Add cookie settings to profile page

RemcoSimonides:
To do issue #3935

https://projects.invisionapp.com/d/main#/console/19607961/420299224/preview

### DIFF
--- a/libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.html
+++ b/libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.html
@@ -1,0 +1,11 @@
+<mat-card>
+  <mat-card-content>
+    <cookie-form [form]="form"></cookie-form>
+  </mat-card-content>
+</mat-card>
+<footer fxLayout fxLayoutAlign="center center">
+  <button mat-flat-button color="accent" (click)="update()">
+    <mat-icon svgIcon="floppy"></mat-icon>
+    <span>Update</span>
+  </button>
+</footer>
--- a/libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.ts
+++ b/libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.ts
@@ -1,0 +1,29 @@
+import { Component } from '@angular/core';
+import { CookiesConsentForm } from '@blockframes/utils/gdpr-cookie/cookie-form/cookie.form';
+import { GDPRService } from '@blockframes/utils/gdpr-cookie/gdpr-service/gdpr.service';
+
+@Component({
+  selector: 'profile-cookie',
+  templateUrl: 'profile-cookie.component.html',
+  styleUrls: ['profile-cookie.component.scss']
+})
+
+export class ProfileCookieComponent {
+
+  public form: CookiesConsentForm;
+  public hasAccepted: boolean;
+
+  constructor(private gdpr: GDPRService) {
+    this.form = new CookiesConsentForm(gdpr.cookieConsent)
+  }
+  
+  update() {
+    const settings = this.form.value;
+
+    if (settings) {
+      this.gdpr.enableAnalytics(settings.googleAnalytics);
+      this.gdpr.enableIntercom(settings.intercom);
+      this.gdpr.enableYandex(settings.yandex);
+    }
+  }
+}
--- a/libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.ts
+++ b/libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.ts
@@ -3,7 +3,7 @@ import { CookiesConsentForm } from '@blockframes/utils/gdpr-cookie/cookie-form/c
 import { GDPRService } from '@blockframes/utils/gdpr-cookie/gdpr-service/gdpr.service';
 
 @Component({
-  selector: 'profile-cookie',
+  selector: 'user-profile-cookie',
   templateUrl: 'profile-cookie.component.html',
   styleUrls: ['profile-cookie.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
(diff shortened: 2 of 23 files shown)


libs/utils/src/lib/gdpr-cookie/cookie-dialog/cookie-dialog.component.scss
```diff
@@ -0,0 +1,8 @@
+mat-divider {
+  margin: 16px 0;
+  position: unset !important;
```
RemcoSimonides, line 3: 💀

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.scss
```diff
@@ -1,15 +1,7 @@
...
-      position: inherit;
-      width: auto;
-    }
-  }
+mat-radio-button {
+  margin: 16px;
 }
+mat-divider {
+  margin: 16px 0;
+  position: unset !important;
```
RemcoSimonides, line 6: 💀

libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.ts
```diff
@@ -0,0 +1,29 @@
+import { Component } from '@angular/core';
+import { CookiesConsentForm } from '@blockframes/utils/gdpr-cookie/cookie-form/cookie.form';
```
RemcoSimonides, line 2: CookiesConsentForm is still in this separate file; but I think this is best because it's now used in two components; cookie-dialog.component.ts and profile-cookie.component.ts.
What do you say @GrandSchtroumpf , still leave it here?

libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.html
```diff
@@ -0,0 +1,11 @@
+<mat-card>
```
GrandSchtroumpf, line 1: It's kind of weird to pu the cookies form in a card just la that. Why do you need a card ?

libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.html
```diff
@@ -0,0 +1,11 @@
...
+  <mat-card-content>
+    <cookie-form [form]="form"></cookie-form>
+  </mat-card-content>
+</mat-card>
+<footer fxLayout fxLayoutAlign="center center">
```
GrandSchtroumpf, line 6: Why the footer is not in the card ?

libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.ts
GrandSchtroumpf: ok

libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.ts
```diff
@@ -0,0 +1,29 @@
...
+import { GDPRService } from '@blockframes/utils/gdpr-cookie/gdpr-service/gdpr.service';
+
+@Component({
+  selector: 'profile-cookie',
+  templateUrl: 'profile-cookie.component.html',
+  styleUrls: ['profile-cookie.component.scss']
```
GrandSchtroumpf, line 8: add changeDetection

libs/user/src/lib/auth/pages/profile-view/profile-view.component.html
```diff
@@ -14,14 +14,19 @@ <h1>{{ user.firstName }} {{ user.lastName }}</h1>
         <span>Previous page</span>
       </a>
   </header>
+  <section>
```
GrandSchtroumpf, line 17: I think you don't need this section

libs/utils/src/lib/gdpr-cookie/cookie-dialog/cookie-dialog.component.html
```diff
@@ -0,0 +1,10 @@
+<mat-card>
```
GrandSchtroumpf, line 1: Same. I've got the feeling you don't need the mat-card

libs/utils/src/lib/gdpr-cookie/cookie-dialog/cookie-dialog.component.scss
GrandSchtroumpf: If you remove the mat-card you won't need that anymore.

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.scss
GrandSchtroumpf: I really don't understand why you need that here.

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.ts
```diff
@@ -1,27 +1,35 @@
...
-  constructor(
-    private dialog: MatDialog,
-    private dialogRef: MatDialogRef<any>,
-  ) { 
+  public masterToggleStatus: 'accept' | 'reject' | 'other'
+  private sub: Subscription;
 
+  ngOnInit() {
+    this.sub = this.form.valueChanges.pipe(startWith(this.form.value)).subscribe(data => {
+      const allAccepted = !Object.keys(data).some(key => !data[key]);
```
GrandSchtroumpf, line 25: Prefer using "every" instead of using negative operators: 
```typescript
const allAccepted = Object.keys(data).every(key => data[key]));
```

GrandSchtroumpf (changes requested):

libs/user/src/lib/auth/pages/profile-cookie/profile-cookie.component.html
RemcoSimonides: Designs
Here the button is outside the card; https://projects.invisionapp.com/d/main#/console/19607961/420299224/preview
And here it is within the card; https://projects.invisionapp.com/d/main#/console/19607961/418991527/preview

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie-form.component.ts
RemcoSimonides: Nice! I was searching for this :D

libs/user/src/lib/auth/pages/profile-view/profile-view.component.html
RemcoSimonides: I do; has to do with the padding added. It's theoretically possible to remove `section`, but then the scss would get quite a lot more complicated.

GrandSchtroumpf (approved):

PR #3922, opened 2020-10-09
fix cropper issues

phpgeekfr:


libs/media/src/lib/components/cropper/cropper.component.ts
```diff
@@ -77,11 +78,14 @@ export class CropperComponent implements OnInit {
   /** Disable fileUploader & delete buttons in 'show' step */
   @Input() useFileUploader?= true;
   @Input() useDelete?= true;
-  @Input() filePrivacy : Privacy = 'public';
+  @Input() filePrivacy: Privacy = 'public';
+  @Input() public types: string[] = ['image/jpeg', 'image/png'];
```
GrandSchtroumpf, line 82: remove public. We never use it

GrandSchtroumpf (approved):

Issue #3756, opened 2020-09-22
Meeting - US - Seller joins a Meeting


OmarSebouai:
# Statement

As a Seller, I want to join a Visio Call Meeting so that I can see the other Buyers.

# ToDo

* [ ] Request Twilio access token and join room
* [ ] Tields view is the default view
* [ ] Get all participants name and organization
* [ ] Add mock UI if user camera isn't activated (with Name + Firstname initials)
* [ ] Show translucent frame with Name/Lastname under participants
* [ ] Add component to dashboard

# Out of Scope
* Add Video Settings (link video settings ticket here)
* Add View switch button (link View Switch ticket here)
* Add Participants button (link Participants ticket here)

# DoD
- Use Angular Material for components
- Use Cascade8 icons (available in code assets/icons)
- Unit Test using Jest
- Inline Documentation
- Feature documention in Notion
- Akita NgFire
- Deployment Test OK
- Unit Test using Jest
- Exception management and log using Sentry
- Inline Documentation
- Strong type, add security in rules, always callback
- Code follow Cascade8 Guidelines (see exemple)
- MR done

link to epic #3608

Issue #3754, opened 2020-09-22
Meeting - TS - Get Twilio Access Token


OmarSebouai:
# ToDo

* [x] Edit .env template (sample.env)
* [x] Add Twilio SID, ApiKey and ApiSecret to environment
* [x] Create twilio.ts and add getAccessToken function
* [x] Verify if user belong to meeting before generating Twilio Token
* [x] Add twilio functions in main.ts
* [ ] Add Unit Test using Jist

# DoD
- Use Angular Material for components
- Use Cascade8 icons (available in code assets/icons)
- Unit Test using Jest
- Inline Documentation
- Feature documention in Notion
- Akita NgFire
- Deployment Test OK
- Unit Test using Jest
- Exception management and log using Sentry
- Inline Documentation
- Strong type, add security in rules, always callback
- Code follow Cascade8 Guidelines (see exemple)
- MR done

link to epic #3608

PR #3941, opened 2020-10-12
Issue3756 seller join meeting

OmarSebouai:
Issue #3756 

This pull request contains some work that are present in the Issue #3754 to allow you to join a Twilio Room (otherwise, this won't be testable).

The BUYER view isn't available for this pull request. If you want to test it, you can simple remove the *ngIf="isSeller" line in:
`/libs/event/src/lib/component/meeting/video/video-component.html line 1`

This way, everyone will have a Seller View and be able to see each other.


libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+   * Get all participant already in the room.
+   * @param participants - All participants connected in the room
+   */
+  getParticipantOfParticipantsMapAlreadyInRoom(participants: Map<string, RemoteParticipant>) {
+    return Array.from(participants).map((
+      participant: any
+    ) => {
+      //participant[0] is theprivate key
+      return participant[1];
+    });
```
pldespaigne, line 230: same as above

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+
+  /**
+   * Function to begin the connection to twilio
+   * First we get the access token with de cloud function
+   * Second we connect to the room with the access token
+   * @param event
+   */
+  doConnectToMeetingService(event) {
+    this.eventService.getTwilioAccessToken(event.id)
+      .then(async (value: ErrorResultResponse) => {
```
pldespaigne, line 242: use `async/await` instead

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+  getIfVideoIsAvailable() {
+    return navigator.mediaDevices.getUserMedia({video: true})
+      .then( () => {
+        this.doSetupLocalVideoAndAudio('video', true);
+        return true;
+      })
+      .catch( () => {
+        this.doSetupLocalVideoAndAudio('video', false);
+        return false;
+      })
```
pldespaigne, line 182: use `async/await` instead

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+
+    const connectOptions = options;
+    if (this.previewTracks) {
+      connectOptions.tracks = this.previewTracks;
+      connectOptions.enableDominantSpeaker = true;
+    }
+
+    connect(accessToken, connectOptions).then((r: Room) => this.roomJoined(r, event), (error) => {
+      console.log('error : ', error)
+    });
```
pldespaigne, line 288: use `async/await` instead

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+   * @param room - room twilio where we are connected
+   * @param event - event when we com from / cascade8
+   */
+  async roomJoined(room: Room, event: Event) {
+    //save activeRoom
+    this.activeRoom = room;
+
+    if (!!room.participants) {
+      const tracksOfParticipants = this.getParticipantOfParticipantsMapAlreadyInRoom(room.participants);
+      if (!!tracksOfParticipants && tracksOfParticipants.length > 0) {
```
pldespaigne, line 303: Remove this `if` as `getParticipantOfParticipantsMapAlreadyInRoom` should always return an array.
Moreover if the array is empty the `for` loop bellow will not run.

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+        for (const indexParticipant in tracksOfParticipants) {
+          const remoteMeetingParticipant = await this.createIParticipantMeeting(tracksOfParticipants[indexParticipant], event);
+          this.addParticipantToConnectedParticipant(remoteMeetingParticipant);
+        }
```
pldespaigne, line 307: The `for` loop can be factorized into a `.forEach`. also you don't need to `await` between each iteration.
Instead you can add promises to an array and do `await Promise.all()` after the loop to `await` concurently.

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+      }
+    }
+
+    const localMeetingParticipant = await this.createIParticipantMeeting(room.localParticipant, event, true );
```
pldespaigne, line 311: you don't need this `const`

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+
+    this.addParticipantToConnectedParticipant(localMeetingParticipant);
+    this.localParticipant = localMeetingParticipant;
+
+    await this.setUpRoomEvent(room, event);
```
pldespaigne, line 316: you don't need to `await` here

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+    const isDominantSpeaker = twilioParticipant.identity === event.organizedBy.uid
+
+    return {
+      identity: twilioParticipant.identity,
+      twilioData: twilioParticipant,
+      festivalData: {
+        firstName: remoteUser.firstName,
+        lastName: remoteUser.lastName,
+        avatar: remoteUser.avatar,
+        organizationName: 'To Be Implemented',
```
pldespaigne, line 339: Don't leave this in the code, an empty string is better in case we forget about it and it goes to production.

But in this case you can do something like
```Typescript
const remoteOrg = await this.orgService.getValue(remoteUser.orgId);

...
organizationName: remoteOrg.denomination.full,
```

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+
+  /**
+   * Make a IParticipantMeeting
+   * @param twilioParticipant
+   * @param event
+   * @param isLocalSpeaker
+   * @private
+   */
+  private async createIParticipantMeeting(twilioParticipant: Participant, event: Event, isLocalSpeaker = false) {
+    const remoteUser = await this.userService.getUser(twilioParticipant.identity)
```
pldespaigne, line 328: you forget the `;`

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+      twilioData: twilioParticipant,
+      festivalData: {
+        firstName: remoteUser.firstName,
+        lastName: remoteUser.lastName,
+        avatar: remoteUser.avatar,
+        organizationName: 'To Be Implemented',
+      },
+      isDominantSpeaker: isDominantSpeaker,
+      isLocalSpeaker: isLocalSpeaker,
+    } as IParticipantMeeting;
```
pldespaigne, line 343: put the return type in the function prototype

libs/event/src/lib/components/meeting/avatar-initial/avatar-initial.component.scss
```diff
@@ -0,0 +1,9 @@
+span {
+  width: 10vh;
```
GrandSchtroumpf, line 2: Use px. Or better use material typography

libs/event/src/lib/components/meeting/avatar-initial/avatar-initial.component.scss
```diff
@@ -0,0 +1,9 @@
...
+  height: 10vh;
+  border-radius: 50%;
```
GrandSchtroumpf, line 4: A rounded text ?

libs/event/src/lib/components/meeting/container-video/container-video.component.ts
```diff
@@ -0,0 +1,59 @@
...
+   */
+  doSetupLocalVideoAndAudio({kind, boolToChange}) {
+    this.meetingService.doSetupLocalVideoAndAudio(kind, boolToChange)
+  }
+
+  /**
+   * when ngDestroy we disconnect the local participant;
+   */
+  ngOnDestroy() {
+    console.log('ngOnDestroy')
```
GrandSchtroumpf, line 56: Please remove console.log

libs/event/src/lib/components/meeting/video/video.component.scss
```diff
@@ -0,0 +1,25 @@
+:host {
+  display: flex;
+  height: 100%;
+  padding: 32px 32px;
+  background-color: var(--background);
+  position: relative;
```
GrandSchtroumpf, line 6: Do you have absolute children ? If not remove this.

libs/event/src/lib/components/meeting/video/video.component.scss
```diff
@@ -0,0 +1,25 @@
...
+section {
+  width: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+mat-grid-list {
+  flex-grow: 1;
+  mat-grid-tile {
```
GrandSchtroumpf, line 18: Remove dead code.

libs/event/src/lib/components/meeting/avatar-initial/avatar-initial.component.scss
OmarSebouai: We wanted to have a circle with text inside (to mock avatar). Instead, we will be using a section with vw size to make the circle, and matcaption for the text.

libs/event/src/lib/components/meeting/avatar-initial/avatar-initial.component.scss
OmarSebouai: See above

libs/event/src/lib/components/meeting/avatar-initial/avatar-initial.component.scss
GrandSchtroumpf: Would it work if you use the class `.mat-list-avatar` on the span ? Else use px & breakpoints if it breaks on tablet.

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+   * Make a IParticipantMeeting
+   * @param twilioParticipant
+   * @param event
+   * @param isLocalSpeaker
+   * @private
+   */
+  private async createIParticipantMeeting(twilioParticipant: Participant, event: Event, isLocalSpeaker = false) {
+    const remoteUser = await this.userService.getUser(twilioParticipant.identity)
+
+    const isDominantSpeaker = twilioParticipant.identity === event.organizedBy.uid
```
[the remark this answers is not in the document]
OmarSebouai: see above

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,453 @@
...
+   */
+  getLocalVideoMicStatus(): Observable<IStatusVideoMic> {
+    return this.localVideoMicStatus$;
+  }
+
+  getActiveUser() {
+    return this.query.user;
+  }
+
+  getIfIsReelOwner(event: Event) {
```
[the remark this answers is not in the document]
pldespaigne: please use `event.isOwner` anyway is it will be the single source of truth, we are in the process of fixing it

libs/event/src/lib/components/meeting/container-video/container-video.component.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  //Input event meeting
+  @Input() event: Event;
+
+  //All Remote Participants in the room Twilio (all participant connected without local)
+  remoteParticipants$: Observable<IParticipantMeeting[]>;
+
+  user: User;
+
+  isSeller: boolean;
```
pldespaigne, line 30: You only use `isSeller` to pass it down to the `event-meeting-video` component.
So why don't you use `event.isOwner` directly in the `event-meeting-video` component ?

libs/event/src/lib/components/meeting/container-video/container-video.component.ts
```diff
@@ -0,0 +1,70 @@
...
+
+@Component({
+  selector: '[event] event-meeting-container-video',
+  templateUrl: './container-video.component.html',
+  styleUrls: ['./container-video.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ContainerVideoComponent implements OnInit, OnDestroy {
+
+  //Input event meeting
```
pldespaigne, line 22: useless comment

libs/event/src/lib/components/meeting/container-video/container-video.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  @Input() event: Event;
+
+  //All Remote Participants in the room Twilio (all participant connected without local)
```
pldespaigne, line 25: prefer ts-doc sytax `/** ... */`

libs/event/src/lib/components/meeting/container-video/container-video.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  remoteParticipants$: Observable<IParticipantMeeting[]>;
+
+  user: User;
+
+  isSeller: boolean;
+
+  constructor(private meetingService: MeetingService,
+              private query: AuthQuery) {
+  }
```
pldespaigne, line 34: indentation

libs/event/src/lib/components/meeting/container-video/container-video.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Event come from child when audio or video is deactivated or activated
+   *
+   * @param identity
+   * @param kind
+   * @param boolToChange
+   */
+  setupVideoAudio({identity, kind, boolToChange}) {
+    this.meetingService.setupVideoAudio(identity, kind, boolToChange)
+  }
```
pldespaigne, line 62: why do you need to emit event to the parent, only to call the service ? could you not call the service directly in the `event-meeting-video` component ?

libs/event/src/lib/components/meeting/participant/participant.abstract.ts
```diff
@@ -0,0 +1,59 @@
...
+      return trackPublication.track;
+    });
+    this.attachTracks(tracks, container);
+  }
+
+  /**
+   *  Detach the Tracks from the DOM.
+   * @param tracks - track to detach of the DOM
+   */
+  detachTracks(tracks) {
```
pldespaigne, line 37: type

libs/event/src/lib/components/meeting/participant/participant.abstract.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+    });
+  }
+
+  /**
+   * Attach the Participant's Tracks to the DOM.
+   * @param participant - participant to attach in the container
+   * @param container - container of participant in DOM
+   */
+  attachParticipantTracks(participant: Participant, container) {
```
pldespaigne, line 24: type

libs/event/src/lib/components/meeting/participant/participant.abstract.ts
```diff
@@ -0,0 +1,59 @@
...
+import {Participant} from 'twilio-video';
+
+export abstract class AbstractParticipant {
+
+  /**
+   * Attach the Tracks to the DOM.
+   * @param tracks - track to attach in the container
+   * @param container - container from DOM to attach the track
+   */
+  attachTracks(tracks, container) {
```
pldespaigne, line 11: type

libs/event/src/lib/components/meeting/participant/remote/remote.component.ts
```diff
@@ -0,0 +1,85 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class RemoteComponent extends AbstractParticipant implements AfterViewInit {
+
+  @Input() participant: IParticipantMeeting
+  @Input() twilioData: Participant
+
+  @Output() eventSetupVideoAudio = new EventEmitter();
+
+  constructor(private renderer: Renderer2, private elm: ElementRef) {
```
pldespaigne, line 33: `renderer` is unused

libs/event/src/lib/components/meeting/video/video.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  templateUrl: './video.component.html',
+  styleUrls: ['./video.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class VideoComponent {
+
+  @Input() remoteParticipants: IParticipantMeeting[];
+  @Input() event: Event;
+  @Input() getTwilioParticipant: (uid: string) => Participant;
+  @Input() isSeller: boolean;
```
pldespaigne, line 22: delete this and use `event.isOwner` directly in the html

Don't resolve comment if they are not fixed, either discuss it or fix it, but don't do this, it's like you don't care of the review

image

pldespaigne (changes requested):

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  isVideoAvailable() {
+    return navigator.mediaDevices.getUserMedia({video: true})
+      .then(() => {
+        return true;
+      })
+      .catch(() => {
+        return false;
+      })
+  }
```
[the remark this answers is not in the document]
OmarSebouai: Indeed, after applying the change, we still had issues with this. I think we took this sample directly from the twilio demo. Will look further into it!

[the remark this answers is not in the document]
OmarSebouai: Hey PL, here is the documentation regarding this function from Mozilla: https://developer.mozilla.org/fr/docs/Web/API/MediaDevices/getUserMedia

We can replace it like that if you agree:
```
  async isAudioAvailable(): Promise<boolean> {
    try {
      await navigator.mediaDevices.getUserMedia({audio: true});
      return true;
    } catch (err) {
      return false;
    }
  }
```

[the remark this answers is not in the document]
pldespaigne: Hey 👋 , this solution ☝️ looks perfect to me

libs/event/src/lib/components/meeting/participant/remote/remote.component.ts
```diff
@@ -0,0 +1,98 @@
...
+  templateUrl: './remote.component.html',
+  styleUrls: ['./remote.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class RemoteComponent implements AfterViewInit, OnDestroy {
+
+  @Input() participant: IParticipantMeeting;
+  @Input() twilioData: Participant;
+
+  @ViewChild('remoteVideo') containerRemoteVideo: ElementRef;
```
pldespaigne, line 40: use an `*ngFor` in the template instead of this

libs/event/src/lib/components/meeting/video/video.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  selector: '[remoteParticipants] [event] [getTwilioParticipant] event-meeting-video',
+  templateUrl: './video.component.html',
+  styleUrls: ['./video.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class VideoComponent {
+
+  @Input() remoteParticipants: IParticipantMeeting[];
+  @Input() event: Event;
+  @Input() getTwilioParticipant: (uid: string) => Participant;
```
pldespaigne, line 21: never pass functions as input

pldespaigne (changes requested):

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,382 @@
...
+
+  /**
+   * Deactive local track of active Room
+   * @param activeRoom: Room (twilio-video Object)
+   */
+  deactivateLocalTracks(activeRoom: Room): void {
+    if (!!activeRoom) {
+      const arrayOfLocalTrack = [];
+      activeRoom.localParticipant.tracks.forEach((track: (LocalAudioTrackPublication | LocalVideoTrackPublication)) => {
+        arrayOfLocalTrack.push(track.track);
```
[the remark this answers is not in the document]
OmarSebouai: We didn't understand this comment. Can we have more info?

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,322 @@
...
+
+  private twilioParticipants: Map<String, Participant> = new Map<String, Participant>();
+
+  activeRoom: Room;
+
+  previewTracks: (LocalAudioTrack | LocalVideoTrack | LocalDataTrack)[];
+
+  accessToken: string;
+
+  localParticipant: IParticipantMeeting;
```
pldespaigne, line 55: Right now this variable is not used (you set it's value in `roomJoined`, but you never use it afterward). So either you delete it, or you use it but in this case you can ensure that the local participant is not stored in `connectedParticipants$` this way you will not need to filter on it.

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,322 @@
...
+    };
+  }
+
+
+  /**
+   * SetUp all event of the Room went we are connected
+   * @param room - Room connected
+   * @param event
+   */
+  setUpRoomEvent(room: Room, event: Event): void {
```
pldespaigne, line 260: - Question : Do you need the `room` param ? Could not use `this.activeRoom` ?
- You can also `import { EventEmitter } from 'events';` and then do `(this.activeRoom as EventEmitter).on`, this way the linter error is gone

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,322 @@
...
+      this.connectedParticipants$.next([]);
+    }
+  }
+
+
+  /**
+   * Deactive local track of active Room
+   * @param activeRoom: Room (twilio-video Object)
+   */
+  deactivateLocalTracks(activeRoom: Room): void {
```
pldespaigne, line 310: Could you also use `this.activeRoom` instead ?

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,322 @@
...
+        map((participants) => participants.filter((participant) => !participant.isLocalSpeaker)
+        )
+      );
+  }
+
+  /**
+   * function to remove a specific participant from the array of participant connected (participantConnected$)
+   * @param participant: IParticipantMeeting : Participant to remove
+   */
+  removeParticipant(participant: IParticipantMeeting | Participant): void {
```
pldespaigne, line 98: All you need as param is `identity: string`

libs/event/src/lib/components/meeting/+state/meeting.service.ts
```diff
@@ -0,0 +1,322 @@
...
+
+    if (!!room.participants) {
+      const participants: Participant[] = Array.from(room.participants.values());
+      const tracks: (Promise<void | Participant>)[] = [];
+      participants.forEach((participant) => {
+        tracks.push(
+          this.createIParticipantMeeting(participant.identity, event)
+            .then((remoteParticipant) => {
+              this.addParticipant(remoteParticipant, participant)
+            })
```
pldespaigne, line 204: Please use `async/await`, also I noticed that you always call `createIParticipantMeeting` just before `addParticipant`, so I think it's better to call `createIParticipantMeeting` from inside of `addParticipant`

pldespaigne (changes requested):

pldespaigne:
Closing & Deleting : We are done implementing the core of the feature, we don't need this PR as example anymore

Issue #3852, opened 2020-10-02
Allow IMG_TOKEN and SENDGRID_API_KEY env vars in circleCI to be updated regarding the environment: production, staging, demo ...


phpgeekfr:
Allow IMG_TOKEN and SENDGRID_API_KEY env vars in circleCI to be updated regarding the environment: production, staging, demo ...

PR #3942, opened 2020-10-12
demo1 - CI, env, docker, gcloud, gsutil, angular.json fixes/upgrades

george43g:
- resolve issue #3852 (you can simply prefix with env name `demo1_SENDGRID`, if it doesn't find its own env it will default to base key)
- angular.json/package.json - added commands and workflows that are backwards compatible, making build/deploy easier and more readable in config
- circleci - updated demo1 as a proposal for a new way of writing config - simplified, commands can also be run locally
- further improvement possible (env and project selection, more commands to factor circle config possible)
- permission and setup of gcloud/gsutil and service accounts
- diagnose and fix docker container to successfully run gsutil
- store configs of multiple envs at once more easily
- fix env selection and release without maintenance process

Deploying one-by-one might actually slow the process down. Can still easily switch back to build:all and deploy:all, but we should start maintenance before the deploy, not just during the upgrade (todo)


angular.json
```diff
@@ -360,6 +376,71 @@
...
+                "command": "npx cross-env firebase use $ENV"
+              },
+              {
+                "command": "npx cross-env npm run deploy:functions:config $ENV"
+              },
+              {
+                "command": "npx cross-env IS_NX=true firebase deploy --force --only functions"
+              }
+            ],
+            "parallel": false,
```
GrandSchtroumpf, line 401: This is the default you don't need to specify

angular.json
```diff
@@ -360,6 +376,71 @@
             "jestConfig": "apps/backend-functions/jest.config.js",
             "passWithNoTests": true
           }
+        },
+        "deploy": {
+          "builder": "@nrwl/workspace:run-commands",
+          "options": {
+            "envFile": ".env",
+            "commands": [
```
GrandSchtroumpf, line 384: You can do : 
```
"commands": [
  "npx nx lint backend-functions",
  "npm run build:backend-functions",
  ...
]
```
I find it cleaner

angular.json
```diff
@@ -360,6 +376,71 @@
...
+                  "command": "npx cross-env ENV=demo1 npm run build:backend-functions"
+                },
+                {
+                  "command": "npx firebase use demo1"
+                },
+                {
+                  "command": "npm run deploy:functions:config demo1"
+                },
+                {
+                  "command": "npx cross-env IS_NX=true firebase deploy --force --only functions"
```
GrandSchtroumpf, line 420: What is `IS_NX` for ?
Why do you need to force ?

angular.json
```diff
@@ -651,6 +732,35 @@
...
+              "optimization": true,
+              "outputHashing": "all",
+              "sourceMap": true,
+              "extractCss": true,
+              "namedChunks": false,
+              "aot": true,
+              "extractLicenses": true,
+              "vendorChunk": false,
+              "buildOptimizer": true,
+              "budgets": [
```
GrandSchtroumpf, line 752: We don't need the budgets for demo

apps/backend-ops/src/tools.ts
```diff
@@ -19,3 +21,20 @@ export function disableMaintenanceMode() {
   process.env.BLOCKFRAMES_MAINTENANCE_DISABLED = 'true';
   console.warn('Maintenance mode is disabled!')
 }
+
+export async function displayCredentials() {
+  let GAP: any;
```
GrandSchtroumpf, line 26: type it with the right type

package.json
```diff
@@ -8,28 +8,35 @@
...
-    "build:backend-ops": "cross-env ng build backend-ops --configuration=\"$ENV\"",
-    "build:crm": "npx ng build crm --configuration=\"$ENV\"",
+    "start": "nx serve --hmr --disable-host-check",
+    "build:stats": "nx build --prod --stats-json",
+    "build:catalog": "cross-env nx build catalog --configuration=\"$ENV\"",
+    "build:festival": "cross-env nx build festival --configuration=\"$ENV\"",
+    "build:financiers": "cross-env nx build financiers --configuration=\"$ENV\"",
+    "build:backend-functions": "cross-env nx build backend-functions --configuration=\"$ENV\"",
+    "build:backend-ops": "cross-env nx build backend-ops --configuration=\"$ENV\"",
+    "build:crm": "cross-env nx build crm --configuration=\"$ENV\"",
```
GrandSchtroumpf, line 18: Could you change `--configuration=\"$ENV\""` to `--configuration=$ENV` ?

tools/scripts/deploy-function-config.ts
```diff
@@ -8,22 +8,51 @@ import { warnMissingVars } from '@blockframes/firebase-utils';
...
 
+const args = process.argv.slice(2);
+const [arg, ...flags] = args;
+
+if (arg) console.log('The following args were detected:', args)
+
+/**
+ * This tuple array maps field names to the environment variable key to set them to
+ */
+const functionsConfigMap: [string, string][] = [
```
GrandSchtroumpf, line 19: Can you use a record ?
```typescript
const config = {
  'sendgrid.api_key': 'SENDGRID_API_KEY',
  'relayer.mnemonic': 'ETHEREUM_MNEMONIC',
  ...
}
```
Looks cleaner like that

tools/scripts/deploy-function-config.ts
```diff
@@ -8,22 +8,51 @@ import { warnMissingVars } from '@blockframes/firebase-utils';
...
+  }
+
+  /**
+   * This generates the statement line used to set function config values in Firebase Functions
+   */
+  function getSettingStatement([fieldPath, envKey]) {
+    return `${fieldPath}="${process.env[getKeyName(envKey)]}"`
+  }
+
+  return functionsConfigMap.map(getSettingStatement)
```
GrandSchtroumpf, line 55: If you change config into a record, it would look like that
```typescript
return Object.entries(config).map(getSettingStatement)
```

tools/scripts/functions-pre-deploy.ts
```diff
@@ -0,0 +1,18 @@
+import { execSync } from 'child_process';
+
+function logBuffer(b: Buffer) {
+  process.stdout.write(b);
+}
+
+if (!process.env['IS_NX']) {
```
GrandSchtroumpf, line 7: Is it for retrocompatibility ? If yes add a comment and a todo to remove it after a specific date (since we don't have versioning system yet)

.circleci/config.yml
```diff
@@ -262,30 +269,33 @@ jobs:
...
+      - ENV: demo1
+      - PROJECT_ID: blockframes-demo-1
       - NODE_OPTIONS: '--max_old_space_size=7168'
       - GOOGLE_APPLICATION_CREDENTIALS: creds.json
     steps:
       - checkout_project
       - run: echo "${FIREBASE_STAGING_SERVICE_ACCOUNT}" > creds.json
-      - run: cp env/demo/env.demo1.ts env/env.ts # required for the typechecking & file rewrite to work
-      - run: cp env/demo/env.demo1.ts env/env.staging.ts # prevent rewrites
-      - run: npm run build:all
```
phpgeekfr, line 273: where is the build done now ?

angular.json
george43g: I should rename this to "SKIP_PREDEPLOY_HOOKS"

This is to allow devs to be able to run firebase deploy --only functions directly without having to build first (it will still run redeploy hooks in that situation) for backwards compatibility

angular.json
george43g: this is not true

angular.json
george43g: this is so much better yeah, didn't know they allowed this

tools/scripts/functions-pre-deploy.ts
george43g: Sort of, except we won't remove it to give devs the ability to use `firebase deploy --only functions: funcName` without having to build firts

.circleci/config.yml
george43g: it's done as part of the deploy steps. However, it might be faster to build all at once in parallel and then just use `firebase deploy --force`

.circleci/config.yml
```diff
@@ -234,65 +271,32 @@ jobs:
...
-      - NODE_OPTIONS: '--max_old_space_size=7168'
-      - GOOGLE_APPLICATION_CREDENTIALS: creds.json
+        PROJECT_ID: blockframes-staging
+        NG_CONFIG: staging
+        SERVICE_ACCOUNT: FIREBASE_STAGING_SERVICE_ACCOUNT
+        ENV_TS: 'env/env.staging.ts'
     steps:
       - checkout_project
-      - run: echo "${FIREBASE_STAGING_SERVICE_ACCOUNT}" > creds.json
+      - prepare_gcloud
```
phpgeekfr, line 280: is gcloud used in the `deploy-to-staging` step ?

tools/scripts/deploy-function-config.ts
```diff
@@ -7,26 +7,57 @@ import { loadSecretsFile, absSecretPath, absTemplatePath } from './lib';
...
+  'twilio.account.sid': 'TWILIO_ACCOUNT_SID',
+  'twilio.api.key.secret': 'TWILIO_API_KEY_SECRET',
+  'twilio.api.key.sid': 'TWILIO_API_KEY_SID'
+ }
 
 /**
  * This is temporary because key names are hardcoded.
  * Future versions will not hardcode this.
  * But need to figure out how to indicate nested objects (more underscores?)
  */
```
phpgeekfr, line 38: Comment should be updated  and moved to `functionsConfigMap` maybe

phpgeekfr (approved):
LGTM but look at comments

phpgeekfr:
`.circleci/deploy_demos.sh` is not used anymore, can you remove it ?

phpgeekfr:
add `prepare_gcloud` to nightly-e2e step

.circleci/config.yml
george43g: I should rename this - this step is meant to be "preparing anything to do with google cloud"

PR #3955, opened 2020-10-13
fixed bug

fritzschoff:


apps/catalog/catalog/src/app/marketplace/title/list/list.component.html
```diff
@@ -3,7 +3,7 @@
   <!-- App Bar Search -->
   <list-page-app-bar>
     <form class="app-bar" [formGroup]="searchForm">
-      <input type="text" formControlName="query" [value]="searchForm.query">
+      <input type="text" formControlName="query" [value]="searchForm.query.value">
```
GrandSchtroumpf, line 6: You shouldn't need the value at all actually. Because you already bind it with formControlName

GrandSchtroumpf: Same goes everywhere below.

GrandSchtroumpf (approved):

Issue #3946, opened 2020-10-13
Rework Media list select


pldespaigne:
- [x] change file image depending on whether it's a pdf/image/video/unknown
- [x] use the horizontal list/carousel component
- [x] make it beautiful 

We don't have the design yet, but it should look a bit like that
![image](https://user-images.githubusercontent.com/5858657/95842396-c93caa00-0d46-11eb-93cc-b664ae7cf7f4.png)

link to epic #3679

PR #3959, opened 2020-10-13
Rework media list select

RemcoSimonides:
issue #3946


libs/media/src/lib/pipes/file-type-image.pipe.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }
+}
+
+@NgModule({
+  declarations: [FileTypeImagePipe],
+  imports: [CommonModule],
+  exports: [FileTypeImagePipe]
+})
+export class FileTypeImagePipeModule { }
```
pldespaigne, line 34: We have almost the same code in `libs\utils\src\lib\pipes\fileName.pipe.ts` can be cool to factorize it somehow

pldespaigne (changes requested):

pldespaigne (approved):

PR #3960, opened 2020-10-13
Campaign start marketplace

GrandSchtroumpf:


libs/organization/campaign/src/lib/marketplace/marketplace.component.html
```diff
@@ -0,0 +1,28 @@
...
+    <h2>Investment Proposal</h2>
+    <article fxLayout *ngIf="campaign.cap">
+      <h5>Total investment needed</h5>
+      <p>{{ campaign.cap | currency:'USD' }}</p>
+    </article>
+    <article fxLayout *ngIf="campaign.minPledge">
+      <h5>Minimum investment amount</h5>
+      <p>{{ campaign.minPledge | currency:'USD' }}</p>
+    </article>
+    <mat-divider></mat-divider>
```
RemcoSimonides, line 14: Outside section

RemcoSimonides (approved):

libs/organization/campaign/src/lib/marketplace/marketplace.component.html
GrandSchtroumpf: No if I put it outside the section it will always be visible. So I might stack divider if most of the sections are not visible.

PR #3963, opened 2020-10-13
Add progress for campaign

GrandSchtroumpf:


libs/organization/campaign/src/lib/components/progress/progress.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ProgressComponent {
+  received: number = 0;
+  progress: number = 0;
+  steps: number[] = [];
+
+  @Input()
+  set campaign(campaign: Campaign) {
+    if (campaign) {
```
fritzschoff, line 17: why if?

GrandSchtroumpf: The first value of an @Input() with a | async is usually null.

PR #3979, opened 2020-10-15
use setAllValue in the form instead of patchAllValue

GrandSchtroumpf:


libs/movie/src/lib/movie/form/movie.form.ts
```diff
@@ -59,6 +59,7 @@ import { createCredit, Stakeholder, createStakeholder, Director } from '@blockfr
 import { createMovieAppAccess } from '@blockframes/utils/apps';
 import { toDate } from '@blockframes/utils/helpers';
 import { Languages } from '@blockframes/utils/static-model';
+import { keys } from 'lodash';
```
fritzschoff, line 62: back in the good old days, I heard a french guy say: `you shall not use lodash in frontend code`... he almost got the same beard as you...

fritzschoff: btw, am I blind or do you don't use it?

libs/movie/src/lib/movie/form/shell/shell.component.ts
```diff
@@ -167,11 +165,11 @@ export class MovieFormShellComponent implements TunnelRoot, OnInit, AfterViewIni
         inline: 'start',
       });
     });
-    this.sub.push(routerSub);
+    this.sub.add(routerSub);
```
fritzschoff, line 168: whoop whoop

fritzschoff (approved):

libs/movie/src/lib/movie/form/movie.form.ts
GrandSchtroumpf: WTF it is doing here...

PR #3983, opened 2020-10-15
Issue renaming

fritzschoff:


libs/contract/src/lib/contract/form/party/party.component.ts
```diff
@@ -1,6 +1,6 @@
 import { algolia } from '@env';
 import { coerceBooleanProperty } from '@angular/cdk/coercion';
-import { staticModels, staticConsts } from '@blockframes/utils/static-model';
+import { staticModel } from '@blockframes/utils/static-model';
```
GrandSchtroumpf, line 3: Could you do `import { legalRoles, subLicensorRoles } ` ?

GrandSchtroumpf (approved):

Issue #1438, opened 2019-12-18
Add a deleted mark to movies when they are deleted instead of deleting them permanently


Hugo-Hontans:
### Description
We won't really delete movies but users will be able to:
- Archive movie (new movie status)
- Once archive, delete it which in fact will remove that org from this movie, but movie will stay in db

### UX
[Here are the wireframes](https://www.figma.com/file/WSbqoyv3Mzkip6yQJGrQR6/%5B-Dashboard-Film-Page-Header-Update-%5D?node-id=108%3A0)

Description of two new statuses:
Archived - titles that are not “in work” anymore, are not displayed in “All titles”, exist as a sort of "trash bin"
Deleted - “deleted” for user (do not appear on app), but actually means for us that a title still exists in the database, but is not displayed in this particular app

Jackseed:
We already store some metadata on `_meta`, we could use that

phpgeekfr:
Update write also the movie guards

Jackseed:
All the guards has to change also so the deleted movies are not loaded

Jackseed:
This is a button to add on the admin panel

Jackseed:
This is a good idea but it has to be developed on the UX side: for the user the mark "deleted" would be more something like "archive" so then they could be re-created if needed. 
It means that we should think of a place where the user can find back the deleted objects. 

There is also the pending question of what do we do with the "satellite" objects that depend on the one we just deleted (for example events linked to movies). 
We could keep it but then if there is a link to the archived movie where does it redirect? Or do we delete it but then we should warn the user of this. Also it could be archived too but then the db would become a great mess. 

@Mussuranae started a list of the satellite objects and what are the delete consequences: https://docs.google.com/spreadsheets/d/1tnAAKEnoUOwpn1yaDGPjmoFcmO34rOCEhjcLji92hYk/edit#gid=0. 

@2amarkina this is not an urgent subject but I add it on the design board so you can work on it at a time before it's implemented.

Jackseed:
Some comments made during sprint 48 retro:

we could use same tabs from Content on top of the title table
"archived" would be a new title status
users should be able to delete the title once archived
flows to design: archive, unarchive and delete a title
could use the same validation modale used in the crm (delete a movie or org)

jdrns:
[...]

PR #3994, opened 2020-10-16
fix #3888

### LINKED ISSUE #3888: Add createdAt on movies
On the meta object we currently store `createdBy` but not `createdAt`.
This should be added.

phpgeekfr:
fix #3888

### DIFF
--- a/libs/movie/src/lib/movie/+state/movie.service.ts
+++ b/libs/movie/src/lib/movie/+state/movie.service.ts
@@ -1,9 +1,16 @@
 import { Injectable } from '@angular/core';
 import { CollectionConfig, CollectionService, WriteOptions } from 'akita-ng-fire';
 import { switchMap, filter, tap, map } from 'rxjs/operators';
-import { createMovie, Movie, MovieAnalytics, SyncMovieAnalyticsOptions, createStoreConfig } from './movie.model';
+import {
+  createMovie,
+  Movie,
+  MovieAnalytics,
+  SyncMovieAnalyticsOptions,
+  createStoreConfig,
+  createDocumentMeta
+} from './movie.model';
 import { MovieState, MovieStore } from './movie.store';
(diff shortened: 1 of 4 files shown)


libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -61,6 +68,9 @@ export class MovieService extends CollectionService<MovieState> {
     const userId = movie._meta?.createdBy ? movie._meta.createdBy : this.authQuery.userId;
     const user = await this.userService.getUser(userId);
 
+    const movieRef = this.db.doc(`movies/${movie.id}`).ref;
```
GrandSchtroumpf, line 71: You can do `const ref = this.getRef(movie.id)`

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -69,14 +79,20 @@ export class MovieService extends CollectionService<MovieState> {
...
+      '_meta.updatedAt': new Date()
+    });
   }
 
   /** Update deletedBy (_meta field of movie) with the current user and remove the movie. */
   public async remove(movieId: string) {
     const userId = this.authQuery.userId;
     // We need to update the _meta field before remove to get the userId in the backend function: onMovieDeleteEvent
-    await this.db.doc(`movies/${movieId}`).update({ '_meta.deletedBy': userId });
+    await this.db.doc(`movies/${movieId}`).update({
```
GrandSchtroumpf, line 92: I don't see the point of updating a document before deleting it.

GrandSchtroumpf (approved):

libs/movie/src/lib/movie/+state/movie.service.ts
phpgeekfr: Because the issue #1438 is going to change the behavior (adding a deleted flag instead of true delete)

Issue #3819, opened 2020-09-29
Media Financiers


GrandSchtroumpf:
# Media Financiers

[Figma](https://www.figma.com/file/x1s9SZZT2mts6kmBuIk9uU/MVP-Media-Financiers)
[Zeplin](https://app.zeplin.io/project/5f3100d317a0041be2ad99c4/dashboard?sid=5f5b4eef8f5db372eebcfc0d)

## Utils
- [x] Update static model to key/value ref #3848  
- [x] Remove default values in the movie
- [x] Update boxoffice unit with a shared unit static model.

## Home
- [X] Add a <ng-content></ng-content> for the action on the movie-card
- [x] Add "cinema investing guide" in assets
- [x] Don't do the query on directors
- [x] Reorganize the home page

## Movie
- [x] Movie Form: Fixing UI bugs
- [x] Marketplace View:
 - (#3820) Add Sale Pitch
 - Financial Info
 - Shooting information
- [x] Movie list: hide "image" at the bottom when no still photos
- [x] Movie list: only get movies from financiers (with an active campaign)
- [x] Movie list: add filter on social goal ref #3989 

## Organization
- [x] Org list: add filters on countries ref #3836 
- [x] Org list: only display orgs from financiers ref #3836 
- [x] Org view > projects: Only display movies from financiers ref #3836 

## Campaign (previously called "project")
- [x] (#3804) Data model
- [x] Rules on campaign
- [x] Movie Dashboard list: Partner tab
- [ ] Movie List: add minimum investment filter
- [x] Movie Marketplace: tab partner + privilege card
- [x] Movie Marketplace: funding progress

## Waterfall (WIP)
- (#3805)

PR #3995, opened 2020-10-16
Issue list campaigne

fritzschoff:
ref #3819


libs/ui/src/lib/static-autocomplete/check-boxes/check-boxes.component.ts
```diff
@@ -35,6 +40,11 @@ export class StaticCheckBoxesComponent implements OnInit {
 
   ngOnInit() {
     this.items = staticModel[this.scope];
+    this.sub = this.form.valueChanges.subscribe(value => {
```
GrandSchtroumpf, line 43: Use filter here : 
```typescript
this.form.valueChanges.pipe(
  filter(value => !value.length) // Only trigger when value is empty
).subscribe(_ => this.checkboxes.forEach(box => box.checked = false))
```

GrandSchtroumpf (approved):

PR #3996, opened 2020-10-16
Production filter

Mussuranae:
Production filter with color and checkbox are now cleared also when the form is cleared


apps/festival/festival/src/app/marketplace/title/list/list.component.ts
```diff
@@ -86,7 +86,6 @@ export class ListComponent implements OnInit, OnDestroy {
   clear() {
     const initial = createMovieSearch({ appAccess: ['festival'], storeConfig: ['accepted'] });
     this.filterForm.reset(initial);
```
phpgeekfr, line 88: remove `private cdr: ChangeDetectorRef,`

phpgeekfr (approved):

PR #3999, opened 2020-10-19
UT: Add unit test for media service.

xmano:
Added unit test for media service 

Closing the earlier PR #3128

`npx jest media.service.spec.ts`

image

### DIFF
--- a/libs/media/src/lib/+state/media.service.spec.ts
+++ b/libs/media/src/lib/+state/media.service.spec.ts
@@ -33,7 +33,7 @@ describe('Media Service Test Suite', () => {
         })},
         { provide: Overlay, useFactory: () => ({
             create: () => ({attach: _ => ({}), detach: _ => ({})}),
-            position: () => ({ global: _ => ({ bottom: _ => ({ left: _ => true }) }) })
+            position: () => ({ global: () => ({ bottom: () => ({ left: () => true }) }) })
         })},
         { provide: SETTINGS, useValue: { host: 'localhost:8080', ssl: false } }
       ],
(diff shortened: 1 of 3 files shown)


libs/media/src/lib/+state/media.service.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        })},
+        { provide: SETTINGS, useValue: { host: 'localhost:8080', ssl: false } }
+      ],
+    });
+    db = TestBed.inject(AngularFirestore);
+    storage = TestBed.inject(AngularFireStorage);
+    service = TestBed.inject(MediaService);
+
+    await loadFirestoreRules({
+      projectId: "test",
```
fritzschoff, line 47: if possible can you use single quotes? Just out of consistency of this repo.

libs/media/src/lib/+state/media.service.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      rules: readFileSync('./firestore.test.rules', "utf8")
```
fritzschoff, line 48: same here if possible

fritzschoff (approved):
LGTM!

libs/media/src/lib/+state/media.service.spec.ts
xmano: Updated code .. removed loading of rules. So this  & above goes away

libs/media/src/lib/+state/media.service.spec.ts
xmano: code removed.

phpgeekfr:
Yeah good job :)

PR #4003, opened 2020-10-20
Fix form & add fundings

GrandSchtroumpf:


libs/organization/campaign/src/lib/form/campaign.shell.config.ts
```diff
@@ -25,7 +26,13 @@ export class CampaignShellConfig implements FormShellConfig<CampaignControls, Ca
 
   async onSave(options: FormSaveOptions): Promise<any> {
     const id: string = this.route.getParams('movieId');
-    await this.service.save(id, this.form.value);
+
+    // TODO (DON'T LET ME MERGE): extract media from form value
```
fritzschoff, line 30: okay, I won't

fritzschoff (changes requested):

libs/movie/src/lib/movie/form/main/main.component.html
```diff
@@ -174,7 +174,7 @@ <h3>General Information</h3>
     <mat-divider></mat-divider>
 
     <section id="directors">
-      <h3>Directors</h3>
+      <h3>Director(s)*</h3>
```
fritzschoff, line 177: Either just use the plural or use the plural pipe and check for the array length. The parenthesis is really not a good solution

libs/organization/campaign/src/lib/form/fundings/fundings.component.scss
```diff
@@ -0,0 +1,3 @@
+section {
+  margin-bottom: 32px;
```
fritzschoff, line 2: why do you need that here? In the form we are not setting this all the time. It makes the design in consistent

fritzschoff (approved):

libs/movie/src/lib/movie/form/main/main.component.html
GrandSchtroumpf: This is the design from Anastasiia

fritzschoff: and also from Mathilde but we said we are using either the plural pipe here or using the plural right away!

libs/organization/campaign/src/lib/form/fundings/fundings.component.scss
GrandSchtroumpf: The design is already not entirely consistent. The section were glued to each other. We'll have a full checkup around December with better design guidelines once everything is done

Issue #3943, opened 2020-10-13
Integrate Org file Upload in the app


pldespaigne:
- [x] Add a new page called "My Files" to festival dashboard
- [x] Add this page to the side menu on the left
- [x] Remove the "Organization Resources" from the profile menu drop down in the top right corner
- [x] Refactor the "Organization Resources" so that it behave like that:

- [x] Add support for movies in file explorer
- [x] Make `ng-content` of cropper and file-upload dynamic

![image](https://user-images.githubusercontent.com/27687382/96845526-13154680-1451-11eb-9b4c-12f02faee6bc.png)

link to epic #3678

PR #4012, opened 2020-10-21
File explorer support org logo

RemcoSimonides:
issue #3943 

- [x] support org logo


libs/media/src/lib/components/dialog/image/image.component.ts
```diff
@@ -0,0 +1,25 @@
...
+import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { HostedMediaForm } from '@blockframes/media/form/media.form';
+import { MediaRatioType } from '../../cropper/cropper.component';
+
+@Component({
+  selector: 'image-dialog',
+  templateUrl: 'image.component.html'
+})
+
+export class ImageDialogComponent {
```
RemcoSimonides, line 11: I've made a separate dialog component for images. It could also be merged together in one dialog since there is only a difference in the html and the injected data.

libs/media/src/lib/components/file-explorer/file-explorer.component.html
```diff
@@ -9,27 +9,27 @@
...
           <span>Upload file</span>
         </button>
       </div>
 
-      <ng-container *ngIf="org | deepKey:'documents.notes' as notes">
-        <bf-table-filter [source]="notes" [columns]="columns" [initialColumns]="initialColumns">
-          <ng-template colRef="ref" let-ref>
-            <img [asset]="ref | fileTypeImage">
+      <ng-container *ngIf="org | deepKey:activeDirectory.path as directory">
+        <bf-table-filter [source]="directory | toArray" [columns]="columns" [initialColumns]="initialColumns">
```
RemcoSimonides, line 28: The table only accepts an array as a source. However, single files are directly on an object and not in an array (logo, poster, banner) - That's why I cast the object to Array. Buuuttt maybe I shouldn't use `<bf-table-filter>` in this case

libs/media/src/lib/components/file-explorer/file-explorer.component.html
```diff
@@ -9,27 +9,27 @@
...
+          <ng-template colRef="ref" let-item="item">
+            <img [asset]="item[activeDirectory.fileRefField] | fileTypeImage">
           </ng-template>
-          <ng-template colRef="title" let-name>{{ name }}</ng-template>
+          <ng-template colRef="name" let-item="item">{{ item[activeDirectory.docNameField] | fileName }}</ng-template>
```
RemcoSimonides, line 32: Documents for Organization consist of `{ title: string, ref: string }` instead of `{ ref: string }` for all other files. Documents show the title as the name of the document, while all other files just have `ref | fileName` as the name of the file.

Using the `fileName` pipe for `title` in Documents for Organization can bug if they use '/' in the name of the title

libs/media/src/lib/components/file-explorer/file-explorer.component.ts
```diff
@@ -49,20 +55,31 @@ export class FileExplorerComponent {
...
         {
-          icon: 'folder',
           name: 'Documents',
-          path: 'documents.path',
+          type: 'file',
+          multiple: true,
+          path: 'documents.notes',
+          docNameField: 'title',
+          fileRefField: 'ref',
+          storagePath: `orgs/${org.id}/documents/notes/`,
```
RemcoSimonides, line 64: `storagePath` could be derived from `path` too

libs/media/src/lib/components/file-explorer/file-explorer.component.ts
```diff
@@ -85,48 +102,86 @@ export class FileExplorerComponent {
...
         title: 'Are you sure you want to delete this file?',
         question: ' ',
         buttonName: 'Yes',
         onConfirm: async () => {
-          const notes = org.documents.notes.filter(n => n.title !== note.title);
-          await this.organizationService.update(this.orgId, { documents: { notes: notes } });
+          const org = await this.organizationService.getValue(this.orgId);
+          const orgForm = new OrganizationForm(org);
+
+          if (this.activeDirectory.multiple) {
```
RemcoSimonides, line 115: This logic definitely needs some rework. It works now specifically for documents and logo; will never work for movies yet.

libs/utils/src/lib/pipes/deep-key.pipe.ts
```diff
@@ -6,7 +6,7 @@ export class DeepKeyPipe implements PipeTransform {
   transform(value: Object, deepKey: string) {
 
     try {
-      if (!value || !deepKey) return;
+      if (!value || !deepKey) return value;
```
RemcoSimonides, line 9: - [ ] Needs proper testing still!
I pass in an empty deepKey for logo since it's directly on the main object (`org`)

libs/media/src/lib/components/dialog/image/image.component.html
```diff
@@ -0,0 +1,16 @@
+<h1>Edit Image</h1>
+<mat-divider></mat-divider>
+<form [formGroup]="data.form" fxLayout="column" fxLayoutGap="24px">
+  <drop-cropper [ratio]="data.ratio" [storagePath]="data.storagePath" [form]="data.form">
+    <h3 type>Logo</h3>
```
RemcoSimonides, line 5: The ng-content of the `<drop-cropper>` and `<file-upload>` component can be different for all files.... Especially for images since they all have different ratio's. Should we define the html for this already in the directories object in the `file explorer component`?

libs/media/src/lib/components/dialog/image/image.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, Inject } from '@angular/core';
+import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { HostedMediaForm } from '@blockframes/media/form/media.form';
+import { MediaRatioType } from '../../cropper/cropper.component';
+
+@Component({
+  selector: 'image-dialog',
+  templateUrl: 'image.component.html'
+})
+export class ImageDialogComponent {
```
RemcoSimonides, line 10: ImageDialogComponent isn't used anymore after this PR. But it will be used again when support for still_photos is added ;)

libs/media/src/lib/components/file-explorer/file-explorer.component.ts
```diff
@@ -73,60 +96,119 @@ export class FileExplorerComponent {
...
+            }
+          });
+
+          if (index > -1) {
+            formList.removeAt(index);
+            const { documentToUpdate } = extractMediaFromDocumentBeforeUpdate(orgForm);
+            await this.organizationService.update(this.orgId, documentToUpdate);
+          } else {
+            console.warn('Control not found');
+          }
```
pldespaigne, line 145: don't leave this please, either throw, or do nothing, or add a todo comment to remove it later

libs/media/src/lib/components/dialog/file/file.component.ts
```diff
@@ -0,0 +1,34 @@
...
+})
+
+export class FileDialogComponent {
+
+  constructor(
+    private dialogRef: MatDialogRef<FileDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public data: { form: HostedMediaWithMetadataForm, privacy: Privacy, storagePath: string }
+  ) {
+    data.form.get('ref').get('ref').setValidators(Validators.required);
+    data.form.get('title').setValidators(Validators.required);
```
pldespaigne, line 23: I prefer that you put this in the `ngOnInit`

pldespaigne (approved):

PR #4021, opened 2020-10-21
hotfix

george43g:
update sample.env


sample.env
```diff
@@ -5,15 +5,24 @@
...
 TWILIO_ACCOUNT_SID=""
 TWILIO_API_KEY_SECRET=""
 TWILIO_API_KEY_SID=""
+
+## Service accounts
+FIREBASE_CI_SERVICE_ACCOUNT=""
+GOOGLE_APPLICATION_CREDENTIALS=""
+
+## Set custom build and deploy config for ng/nx
+# NG_CONFIG=""
```
phpgeekfr, line 24: what is default ?

sample.env
```diff
@@ -5,15 +5,24 @@
...
+# PROJECT_ID="blockframes-yourName"
```
phpgeekfr, line 25: same

phpgeekfr (approved):

sample.env
george43g: there is no default. default is blank

sample.env
george43g: no that's just an example. again, no default. whatever firebase is currently `use` isThe default

sample.env
phpgeekfr: Can you give an example value ? And why it should be used ?

sample.env
phpgeekfr: OK this is to override `firebase use` . Can you explain this in comment and explain in which context it should be used?

Add a link to notion page if there is an existing one for this

sample.env
george43g: example value:
`production` to compile for production 
`staging` used by demo environments


... come to think of it, the `PROJECT_ID` could be extracted from the config, as each one corresponds to a projectId , for example - 

NG_CONFIG=production
PROJECT_ID=blockframes

george43g: these are mainly used by CI, but if a user wanted to work on demo1, staging, for example, they could set these and run nx deploy XXX

but they can also run nx deploy xxx --configuration=demo1 if they want to. atm both options work, mainly used by us

PR #4030, opened 2020-10-22
fixed bug

fritzschoff:
fixed form bug


libs/ui/src/lib/form/list/form-list.component.ts
```diff
@@ -48,9 +48,13 @@ export class FormListComponent<T> implements OnInit {
...
       startWith(this.form.value),
-      distinctUntilChanged())
+      distinctUntilChanged(),)
 
-      this.add();
+    this.add();
+  }
+
+  ngOnDestroy() {
+    if (this.form?.valid) this.save()
```
GrandSchtroumpf, line 57: Don't check the validity of the form, check pristine of the `formItem` instead.

libs/ui/src/lib/form/table/form-table.component.ts
```diff
@@ -92,7 +92,8 @@ export class FormTableComponent<T> implements OnInit, AfterViewInit, OnDestroy {
   }
 
   ngOnDestroy() {
-    this.sub.unsubscribe();
+    if (this.form?.valid) this.save()
```
GrandSchtroumpf, line 95: Same as above

GrandSchtroumpf (changes requested):

GrandSchtroumpf (approved):

PR #4038, opened 2020-10-22
Update(e2e): Refactor Movie Tunnel test to avoid slow navigation.

### LINKED ISSUE #4000: E2E : Cypress XHR response very slow during movie creation
In movie-tunnel test, at the step when a new movie is created, there is extreme slowness noticed.

It is suspected that multiple XHR at this point are causing the issue.

cy.wait() is not a good workaround here. We need better way to avoid this slow response issue.

A possible solution could be to disable GA since they are not directly related for the UI test.

xmano:
Refactor to fix #4000 

This is not complete fix but get test to work reliably.


apps/festival/festival-e2e/src/integration/dashboard/movie-tunnel.spec.ts
```diff
@@ -130,93 +129,132 @@ const Movie = {
...
 
-    cy.log('[Summary - Main fields]');
-    cy.get('#main-information [test-id]').each(el => {
-      console.log(el);
-      const key = el.attr('test-id');
-      cy.wrap(el).contains(Movie.mainInfo[key])      
+    MovieFormSummary.forEach(section => {
+      cy.log(`=>[${section.title}]`);
+      cy.get(section.selector).each(el => {
+        console.log(el);
```
phpgeekfr, line 253: console.log are needed ? or just for debug ?

apps/festival/festival-e2e/src/integration/dashboard/movie-tunnel.spec.ts
```diff
@@ -1,11 +1,10 @@
 /// <reference types="cypress" />
 
-import { clearDataAndPrepareTest, setForm } from '@blockframes/e2e/utils/functions';
+import { clearDataAndPrepareTest, setForm, testPromise } from '@blockframes/e2e/utils/functions';
```
phpgeekfr, line 3: not used import `testPromise`

phpgeekfr (approved):

apps/festival/festival-e2e/src/integration/dashboard/movie-tunnel.spec.ts
xmano: For debug.. but it dont matter for production.

apps/festival/festival-e2e/src/integration/dashboard/movie-tunnel.spec.ts
xmano: removed.

PR #4054, opened 2020-10-23
Security pop up

Mussuranae:
Add a security pop up when user wants to delete an event (to ask confirmation of the deletion).
Also add this on the screening page, when the screening is running and the user wants to leave the page.


libs/event/src/lib/guard/session.guard.ts
```diff
@@ -28,4 +34,20 @@ export class SessionGuard implements CanActivate {
     }
     return true;
   }
+
+  canDeactivate(
```
GrandSchtroumpf, line 38: remove parameters if you don't use them.

phpgeekfr (approved):

GrandSchtroumpf (approved):

PR #4056, opened 2020-10-23
UnitTest: Twilio unit-test passes with local emulator.  Fix #3824

### LINKED ISSUE #3824: Initialization of app crash when doing Unit tests on backend-functions dir.
Initialization of app crash when doing Unit tests on backend-functions dir.
This is certainly due to this line in "apps/backend-functions/src/internals/firebase.ts" :
```
admin.initializeApp(functions.config().firebase);
```
Which is being called before `initFunctionsTestMock();` inside beforeAll of tests.

xmano:
Setup Needed:

1. Twilio account and keys to be setup and in .env 
2. Deploy the cloud functions
    firebase deploy --only functions:getAccessToken
3. Configs should be in .runtimeconfig.json
4. Credentials in tools\credentials\blockframes-xxx-yyy.json (?)
5. Local emulators should be started
 
Run with npx jest  apps/backend-functions/src/twilio.spec.ts

image

TODO:
1. Explore wrapping the cloud functions.
2. Check credentials are needed
3. npm run test  should pass.
4. Create a notion page for documenting the setup and code example


GrandSchtroumpf (approved):

pldespaigne (approved):

apps/backend-functions/src/internals/invitations/meetings.ts
```diff
@@ -0,0 +1,23 @@
+import * as admin from 'firebase-admin';
+
+/**
+ * Return true if userId has accepted invitation for eventId
+ * @param userId : string - Id of user
+ * @param eventId : string - Id of doc (id of event in Invitation)
+ */
+export async function hasUserAcceptedEvent(userId: string, eventId: string) {
+  const db = admin.firestore();
```
george43g, line 9: the db is already initialized in internals/firebase

also, I'm not sure if this is the right place for this file

xmano: That is the reason we move it out of internals to avoid initialising it first. I have ensured that proper init happens in `initFunctionsTestMock()` in beforeAll().

PR #4073, opened 2020-10-26
added preview modal

fritzschoff:


libs/ui/src/lib/preview-modal/dialog-preview/dialog-preview.component.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  public src: string
+
+  constructor(
+    @Inject(MAT_DIALOG_DATA) public data,
+    private mediaService: MediaService
+  ) { }
+
+  ngOnInit() {
+    if(this.data.ref?.changingThisBreaksApplicationSecurity?.includes('blob')) {
```
RemcoSimonides, line 28: `changingThisBreaksApplicationSecurity` what's that?

RemcoSimonides: Looks like you send the whole Form to this component. But maybe you can send only blob so you don't need to check if blob exists in this strange way?

libs/ui/src/lib/preview-modal/preview.component.html
```diff
@@ -0,0 +1,5 @@
+<button mat-icon-button (click)="openModal()" [disabled]="!this.form.ref.value" matTooltip="Click to see preview"
```
RemcoSimonides, line 1: Check for `ref.ref.value` or `ref.blob.value`.

libs/ui/src/lib/preview-modal/dialog-preview/dialog-preview.component.ts
```diff
@@ -0,0 +1,34 @@
...
+      this.src = this.data.ref
+    } else {
+      this.mediaService.generateImgIxUrl(this.data.ref, this.parameters).then(src => this.src = src)
```
RemcoSimonides, line 31: make `ngOnInit` async and then use `this.src = await this.mediaService.generate....`.

libs/ui/src/lib/preview-modal/preview.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class PreviewModalComponent {
+
+  @Input() form: HostedMediaForm
+
+  constructor(private dialog: MatDialog, private sanitizer: DomSanitizer) { }
+
+  openModal() {
+    let ref: any;
```
RemcoSimonides, line 20: In which case is ref not a `string`?

RemcoSimonides (changes requested):

libs/ui/src/lib/preview-modal/dialog-preview/dialog-preview.component.ts
fritzschoff: i need the whole form if the blob is not present I need the ref. I could send down only the form value but this gives me more flexibility and since there is no performance issue, I don't see any issue here.

RemcoSimonides (approved):

PR #4082, opened 2020-10-27
bug fix e2e movie form

phpgeekfr:


libs/movie/src/lib/movie/form/shell/shell.component.ts
```diff
@@ -105,7 +105,7 @@ function getSteps(status: ProductionStatus, appSteps: TunnelStep[] = []): Tunnel
   return steps.map(step => {
     return {
       ...step,
-      routes: step.routes.filter(route => !route?.shouldDisplay)
+      routes: step.routes.filter(route => route.shouldDisplay === undefined || route.shouldDisplay === true)
```
xmano, line 108: Are we changing the filter condition here?

`route.shouldDisplay === true`

is opposite of what was there earlier.

xmano (commented):
I will take care of TODO in e2e test.

Please check the filter condition. If logic is ok then go ahead. Otherwise LG.

xmano (approved):
LGTM

PR #4089, opened 2020-10-28
SendGrid Service + added send mail with tpl from front

phpgeekfr:


libs/utils/src/lib/emails/sendgrid.service.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from "@angular/core";
+import { AngularFireFunctions } from '@angular/fire/functions';
+import { ErrorResultResponse } from '@blockframes/utils/utils';
+import { EmailRequest, EmailTemplateRequest } from '@blockframes/utils/emails/utils';
+import { callFunction } from 'akita-ng-fire';
+
+@Injectable({ providedIn: 'root' })
+export class SendGridService {
```
GrandSchtroumpf, line 8: I would have done like that : 
```typescript
export class Sendgrid {
  sendTest: (request: EmailRequest, from?: string) => Promise<ErrorResultResponse> = callFunction.bind(this.functions, 'sendTestMail');
  sendTemplate: (request: EmailTemplateRequest, from?: string) => Promise<void> = callFunction.bind(this.functions, 'sendTemplateMail');
}
```
We are using "sendgrid" service we know that we send a mail, no need to specify.

GrandSchtroumpf (changes requested):

libs/utils/src/lib/emails/sendgrid.service.ts
```diff
@@ -0,0 +1,27 @@
...
+import { callFunction } from 'akita-ng-fire';
+import { App } from '../apps';
+
+@Injectable({ providedIn: 'root' })
+export class SendgridService {
+
+  /**
+   * Allowed only to Blockframes Admins
+   * @param request 
+   * @param from
```
GrandSchtroumpf, line 14: There is no from neither request here..

libs/utils/src/lib/emails/sendgrid.service.ts
```diff
@@ -0,0 +1,27 @@
...
+   * Allowed only to Blockframes Admins
+   * @param request 
+   * @param from 
+   */
+  sendAsAdmin: (data: EmailParameters) => Promise<ErrorResultResponse> = callFunction.bind(this, this.functions, 'sendMailAsAdmin');
+
+  /**
+   * Allowed to every user, but they must use a template Id and cannot choose "from".
+   * Template ids can be allowed to use in apps/backend-functions/src/internals/email.ts isAllowedToUseTemplate
+   * @param request
```
GrandSchtroumpf, line 21: Same as above, I only see data.

libs/utils/src/lib/emails/sendgrid.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class SendgridService {
+
+  /**
+   * Allowed only to Blockframes Admins
+   * @param request 
+   * @param from 
+   */
+  sendAsAdmin: (data: EmailParameters) => Promise<ErrorResultResponse> = callFunction.bind(this, this.functions, 'sendMailAsAdmin');
```
GrandSchtroumpf, line 16: Maybe create a reusable method : 
```typescript
sendAsAdmin: (data: EmailParameters) => Promise<ErrorResultResponse> = this.callable('sendMailAsAdmin');

callable(method: string) {
  return callFunction.bind(this, this.functions, method);
}
```

GrandSchtroumpf (approved):

PR #4094, opened 2020-10-28
George/refactor

george43g:
- hotfix


libs/firebase-utils/src/lib/firestore/export.ts
```diff
@@ -69,5 +69,7 @@ export async function exportFirestoreToBucket(db: FirebaseFirestore.Firestore, b
     await Promise.all(promises);
   }
   console.log('Firestore export done, closing our stream');
-  return new Promise(stream.end);
+  return new Promise((res) => {
+    stream.end(res);
+  });
```
phpgeekfr, line 74: I don't understand why you changed this.

Why not :
```
  return new Promise(res => stream.end(res));

OR 
  return new Promise(stream.end);
```
?

george43g: believe it or not, it fixes a bug

phpgeekfr: lol
ok

PR #4100, opened 2020-10-28
Issue spliting indices

fritzschoff:


apps/backend-functions/src/movie.ts
```diff
@@ -62,8 +62,17 @@ export async function onMovieDelete(
...
 
+  const movieAppAccess = Object.keys(movie.storeConfig.appAccess).filter(access => movie.storeConfig.appAccess[access]);
+
+  const promises = [];
+
   // Update algolia's index
-  await deleteObject(algolia.indexNameMovies, context.params.movieId);
+  movieAppAccess.forEach(appName => {
+    const promise = deleteObject(algolia.indexNameMovies[appName], context.params.movieId)
+    promises.push(promise);
```
fritzschoff, line 72: delete the indices that the movie was stored.
Only bottle would be if you change the appAccess, but I don't think this is possible? Maybe CRM?

apps/backend-ops/src/algolia.ts
```diff
@@ -44,74 +45,65 @@ export async function upgradeAlgoliaOrgs() {
...
+               throw new Error(`Movie ${movie.id} is part of several orgs (${querySnap.docs.map(doc => doc.id).join(', ')})`);
+             } */
+
+          const org = (querySnap.docs[0].data() as OrganizationDocument);
+          const orgName = org.denomination.public || org.denomination.full;
+
+          if (appConfig === 'financiers') {
+            const campaignSnap = await db.collection('campaigns').where('id', '==', movie.id).get();
+            const campaign = (campaignSnap.docs[0].data() as Campaign);
+            if (campaign?.id) {
```
fritzschoff, line 90: of there is a campaign write the minpledge to the movie object and pass it down

apps/backend-ops/src/algolia.ts
```diff
@@ -147,3 +139,38 @@ export async function upgradeAlgoliaUsers() {
...
+      'languages.dubbed',
+      'languages.subtitle',
+      'languages.caption',
+      'originCountries',
+      'status',
+      'storeConfig',
+      'storeType'
+    ],
+  };
+  switch (appConfig) {
```
fritzschoff, line 170: configs are different for every app

apps/backend-functions/src/movie.ts
phpgeekfr: Yes it can be changed in CRM. But onMovieUpdate should change this on algolia when appAcess are changed

phpgeekfr: You can check if appAcess changed in onMovieUpdate and deleteObject if needed

libs/utils/src/lib/algolia.ts
```diff
@@ -102,4 +102,49 @@ export interface AlgoliaMovie {
   objectID: string
 }
 
+// TODO extract that (along with other potential common features) into an algolia file
```
phpgeekfr, line 105: add an issue number please

fritzschoff: gonna remove this issue since we re now in a specific algolia file in utils lib

apps/backend-functions/src/movie.ts
```diff
@@ -114,7 +123,13 @@ export async function onMovieUpdate(
   const creatorOrg = await getDocument<OrganizationDocument>(`orgs/${creator!.orgId}`);
 
   if (creatorOrg.denomination?.full) {
-    await storeSearchableMovie(after, orgName(creatorOrg));
+    for (const app in after.storeConfig.appAccess) {
+      if (after.storeConfig.appAccess[app]) {
+        await storeSearchableMovie(after, orgName(creatorOrg));
```
phpgeekfr, line 128: I think this line will be executed many times, but it should be only executed once no ?

phpgeekfr (approved):

apps/backend-ops/src/algolia.ts
GrandSchtroumpf: I would do it like that : 
```typescript
const baseConfig = {
  searchableAttributes: [ .. ],
  attributesForFaceting: [ .. ]
}
function movieConfig(app: App) {
  switch (app) {
    case 'financiers': return {
      searchableAttributes: movieConfig.searchableAttributes,
      attributesForFaceting [
        ...movieConfig.attributesForFaceting,
        'socialGoals',
        'filterOnly(minPledge)'
      }
    }
  }
}
```
I think it's easier to read personnaly.

libs/ui/src/lib/layout/dashboard/dashboard.component.ts
```diff
@@ -33,7 +34,7 @@ export class DashboardComponent implements AfterViewInit, OnDestroy {
 
   public ltMd$ = this.breakpointsService.ltMd;
 
-  public movieIndex = algolia.indexNameMovies;
+  public movieIndex;
```
GrandSchtroumpf, line 37: type it please.

apps/backend-ops/src/algolia.ts
```diff
@@ -44,74 +45,65 @@ export async function upgradeAlgoliaOrgs() {
...
+  } else {
+
+    // reset config, clear index and fill it up from the db (which is the only source of truth)
+    const config = movieConfig(appConfig)
+
+    await setIndexConfiguration(algolia.indexNameMovies[appConfig], config, process.env['ALGOLIA_API_KEY']);
+    await clearIndex(algolia.indexNameMovies[appConfig], process.env['ALGOLIA_API_KEY']);
+
+    const { db } = loadAdminServices();
+    const moviesIterator = getCollectionInBatches<MovieDocument>(db.collection('movies'), 'id', 300)
```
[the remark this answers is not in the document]
fritzschoff: no idea. I just copy and pasted

apps/backend-ops/src/algolia.ts
```diff
@@ -44,74 +45,65 @@ export async function upgradeAlgoliaOrgs() {
...
+             } */
+
+          const org = (querySnap.docs[0].data() as OrganizationDocument);
+          const orgName = org.denomination.public || org.denomination.full;
+
+          if (appConfig === 'financiers') {
+            const campaignSnap = await db.collection('campaigns').where('id', '==', movie.id).get();
+            const campaign = (campaignSnap.docs[0].data() as Campaign);
+            if (campaign?.id) {
+              movie['minPledge'] = campaign.minPledge;
```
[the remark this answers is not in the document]
fritzschoff: for the campaign, yes. And depending on which app we update the index according to the config

PR #4105, opened 2020-10-28
Event guard dashboard

### LINKED ISSUE #3930: Dashboards are open to every users
Dashboard pages are not secured, you can access it as long as you have the url and an app account. Even to critical pages like event edit pages. 

These pages should be restricted to org members only

Mussuranae:
Add a new guard to prevent user from other org to go on the event form page of event of other org
Fix #3930

### DIFF
--- a/libs/event/src/lib/guard/event-organization.guard.ts
+++ b/libs/event/src/lib/guard/event-organization.guard.ts
@@ -1,0 +1,26 @@
+import { Injectable } from '@angular/core';
+import { ActivatedRouteSnapshot, CanActivate, Router } from '@angular/router';
+import { AuthQuery } from '@blockframes/auth/+state';
+import { OrganizationQuery } from '@blockframes/organization/+state';
+import { EventService } from '../+state';
+
+@Injectable({ providedIn: 'root' })
+export class EventOrganizationGuard implements CanActivate {
+  constructor(
+    private service: EventService,
+    private router: Router,
+    private authQuery: AuthQuery,
+    private orgQuery: OrganizationQuery
+  ) { }
+
+  async canActivate(next: ActivatedRouteSnapshot) {
+    const eventId: string = next.params['eventId'];
+    const event = await this.service.getValue(eventId);
+    console.log(event)
+    console.log(event.ownerId)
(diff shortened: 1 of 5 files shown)


libs/movie/src/lib/movie/guards/movie-tunnel.guard.ts
```diff
@@ -18,7 +18,8 @@ export class MovieTunnelGuard implements CanActivate {
     }
     // When creating a move we need to make sure the org is updated before checking
     org = await this.orgService.getValue(org.id);
-    const redirect = this.router.parseUrl(`/c/o/dashboard/title/${movieId}`);
+    const redirect = this.router.parseUrl(`/c/o/dashboard/title`);
+    console.log(redirect)
```
phpgeekfr, line 22: can you remove console.log?

libs/event/src/lib/guard/event-organization.guard.ts
```diff
@@ -0,0 +1,26 @@
...
+    private router: Router,
+    private authQuery: AuthQuery,
+    private orgQuery: OrganizationQuery
+  ) { }
+
+  async canActivate(next: ActivatedRouteSnapshot) {
+    const eventId: string = next.params['eventId'];
+    const event = await this.service.getValue(eventId);
+    console.log(event)
+    console.log(event.ownerId)
```
phpgeekfr, line 20: can you remove console.log?

phpgeekfr (approved):
LGTM, but make the changes on every apps

libs/event/src/lib/guard/event-organization.guard.ts
```diff
@@ -0,0 +1,26 @@
...
+    const orgId = this.orgQuery.getActiveId();
+    const uid = this.authQuery.userId;
+    const redirect = this.router.parseUrl(`/c/o/dashboard/event`);
```
GrandSchtroumpf, line 23: Isn't it "events" with a "s" ?

Mussuranae: Nope, I checked it in case you were right, but it's "event".

PR #4110, opened 2020-10-29
Fix(E2E): Add search for org name by inputting the string.

xmano:
Fix failing tests with adding a function to input org name and select it.

image

One test is still failing. It may be due to org name not present in notification. Will see if it can be changed by app code or might be a TODO pending debug.

image


xmano:
Maybe I change to Event Name instead of Org Name.. i think this is a better test strategy.

phpgeekfr (approved):

xmano:
Refactored the tests and now all tests are passing in local:

image

PR #4111, opened 2020-10-29
MF Cleanup 5

GrandSchtroumpf:


libs/organization/campaign/src/lib/form/form.ts
```diff
@@ -141,6 +141,10 @@ export class CampaignForm extends FormEntity<CampaignControls, Campaign> {
     super(controls, [compareMinPledge, compareReceived]);
   }
 
+  get currency() {
```
fritzschoff, line 144: could throw a lot of bugs since these getter functions are in our repo always AbstractControls and never value. maybe change the name

fritzschoff (approved):

PR #4112, opened 2020-10-29
fix #4004

### LINKED ISSUE #4004: List of known bug in the Media lib
Please **add your name** next to the bug, so I can contact you if I have trouble reproducing.

## Uploads
- [ ] Upload pdf: upload a file -> Save -> refresh: you don't see the file (@GrandSchtroumpf): issue only visible for the campaign form.

phpgeekfr:


libs/media/src/lib/components/upload/file-upload/file-upload.component.ts
```diff
@@ -27,16 +31,29 @@ export class FileUploadComponent implements OnInit {
...
+
+    // show current file when component loads
     if (!!this.form.blobOrFile.value) {
       this.selected(this.form.blobOrFile.value);
     } else if (!!this.form.oldRef?.value) {
       this.state = 'file';
     }
+
+    // update component when oldRef changes
+    this.sub = this.form.get('oldRef').valueChanges.subscribe(oldRef => {
```
GrandSchtroumpf, line 51: so I would have wrote it like that : 
```typescript
state = new BehaviorSubject();
ngOnInit() {
  const { blobOrFile, oldRef } = this.form.value;
  if (blobOrFile) this.selected(blobOrFile);

  this.sub = this.form.get('oldRef').valueChanges.pipe(
    startWith(oldRef),
    filter(ref => !!ref),
  ).subscribe(_ => this.state.next('file'))
}
```
Using a behavior subject will avoid the need of marking the changedector manually.
But it's up to you :)

GrandSchtroumpf (approved):

PR #4118, opened 2020-10-30
Feat 4063 fix safari bug and campaign form

phpgeekfr:


fritzschoff (approved):

Mussuranae (approved):

libs/movie/src/lib/movie/form/shell/shell.component.ts
```diff
@@ -172,8 +172,10 @@ export class MovieFormShellComponent implements TunnelRoot, OnInit, AfterViewIni
 
   ngOnDestroy() {
     this.sub?.unsubscribe();
-    this.getForm('movie').reset()
-    this.getForm('campaign').reset()
+    this.getForm('movie').reset();
+    if (this.configs[name]) {
```
GrandSchtroumpf, line 176: Where is "name" defined ?

libs/movie/src/lib/movie/dashboard/shell/shell.component.html
```diff
@@ -1,5 +1,5 @@
 <section fxLayout="column">
-  <header fxLayout="row" fxLayout.lt-md="column" fxFlex>
+  <header fxLayout="row" fxLayout.lt-md="column" fxFlex="100%">
```
GrandSchtroumpf, line 2: What is the issue ?

libs/movie/src/lib/movie/form/shell/shell.component.ts
phpgeekfr: typo, fixed

libs/movie/src/lib/movie/dashboard/shell/shell.component.html
phpgeekfr: image

libs/movie/src/lib/movie/form/shell/shell.component.ts
```diff
@@ -173,7 +173,7 @@ export class MovieFormShellComponent implements TunnelRoot, OnInit, AfterViewIni
   ngOnDestroy() {
     this.sub?.unsubscribe();
     this.getForm('movie').reset();
-    if (this.configs[name]) {
+    if (this.configs['campaign']) {
```
GrandSchtroumpf, line 176: Prefer `this.getForm('campaign')?.reset()` then. But I don't understand why you need that.

phpgeekfr: I had to add `?` also to 
``` getForm<K extends keyof ShellConfig>(name: K): ShellConfig[K]['form'] {
    return this.configs[name]?.form;
  }
```
but ok

libs/movie/src/lib/movie/dashboard/shell/shell.component.html
phpgeekfr: @GrandSchtroumpf Waiting for your test before removing this

Issue #3383, opened 2020-07-28
[Unit test] Test existing firestore rules


Jackseed:
Main issue to cut into smaller pieces

xmano:
[ ] Create the setup for unit-test Firestore Security Rules
[ ] Identify the main areas of security concerns (secure areas, private info)
[ ] Write example tests for each of the areas

xmano:
Got basic test to work with new recommended library `@firebase/rules-unit-testing`

![image](https://user-images.githubusercontent.com/8731418/98338045-8d3bf280-202f-11eb-92ae-9a2663604b45.png)

xmano:
Individual collections to test:

[ ] - Campaign : #4187

xmano:
Merging current changes with todos:

![image](https://user-images.githubusercontent.com/8731418/98652641-f5f1da80-2361-11eb-8216-1a14405787c6.png)

xmano:
Closing this issue and tracking the tests using separate issue:
https://github.com/blockframes/blockframes/issues/4373

PR #4122, opened 2020-10-30
Add unit test to check security rules. WIP

xmano:
Feat Issue #3383


apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  });
+
+  const db = app.firestore();
+  return db;
+}
+
+/**
+ * Helper function to setup Firestore DB Data
+ */
+async function setFirestoreDB(projectId: string,
```
GrandSchtroumpf, line 21: 1. Remove "DB": `setFirestore`. We know that firestore is a DB.
2. Type your params
3. format either on one line or one line per params
4. rename `strRulesPath` to `rulesPaths` (if it's an array) or `rulesPath` if it's only one string
5. What is "data" ? The name doesn't help to understand what it is. So either you give abetter name, or you add a line in the comment
6. Don't set the default value to "null", just make it optional, and it'll be undefined (except in the rare case where null works and not undefined).

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+async function setFirestoreDB(projectId: string, 
+              strRulesPath: string[], db: any = null, data: any = null) {
+  // Apply the firestore rules to the project
+  await loadFirestoreRules({
+    projectId,
+    rules: fs.readFileSync(strRulesPath[0], "utf8")
+  });
+
+  // Write data to firestore app
+  if (data) {
```
GrandSchtroumpf, line 30: What kind of format do you expect ? Something like that ?
```
type FirestoreData = {
  [idPath: string]: Object
}
```
Where the data comes from ? I feel like we should use the "--import" property of the emulator to load binary data directly.

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    for (const key in data) {
+      const ref = db.doc(key);
+      await ref.set(data[key]);
```
GrandSchtroumpf, line 33: Any reason to await between each iteration ? I feel like this could be done in concurrency :
```typescript
const promises = Obect.entries(data).map(([path, doc]) => db.doc(path).set(doc));
await Promise.all(promises);
```

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+  if (strRulesPath.length === 2) {
```
GrandSchtroumpf, line 37: Using a array to manage this makes it very difficult to understand what each field for.
Only use arrays when each entry has the same role.

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    return {
+      pass,
+      message: () =>
+        `Expected Op to be allowed, but was denied ${err ? '\nErr: err': ''}`
+    };
+  }
+});
+
+expect.extend({
```
GrandSchtroumpf, line 74: Why here you're using expect, and the other one "fexpect" ?

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    });    
+  }
+}
+
+/**
+ * TODO : 4166
+ * Refactor this jest extension into separate lib and
+ * using proper TS code.
+ */
+const fexpect: any = Object.assign(expect);
```
GrandSchtroumpf, line 51: Why you're doing that ? Can you point out a reference to the right Jest documentation ?

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+
+fexpect.extend({
+  async toAllow(testPromise) {
```
GrandSchtroumpf, line 54: type params
Rewrite like that : 
```typescript
toAllow(test: Promise<????>) {
  return assertSucceeds(test)
    .then(_ => { pass: true, message: '' })
    .catch(err => { pass: false, message: `Expected Op to be allowed, but was denied:${err}` });
}
```

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+    };
+  }
+});
+
+expect.extend({
+  async toDeny(testPromise) {
+    let err = '';
+    let pass = false;
+    try {
+      await assertFails(testPromise);
```
GrandSchtroumpf, line 79: Same as above

apps/backend-ops/src/firestore-rules.spec.ts
GrandSchtroumpf: But I really don't see the point of extending expect for this simple operation. You're loosing type, making the code more complex and difficult to understand for someone else. I would recommend not to do that.

GrandSchtroumpf (changes requested):

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
+﻿import { apps, assertFails, assertSucceeds, initializeTestApp, 
+        loadFirestoreRules } from '@firebase/rules-unit-testing';
+import { testFixture } from './fixtures/data';
+import fs from 'fs';
+
+const setupFirestore = (projectId: string, auth = null) => {
+  process.env.GCLOUD_PROJECT = projectId;
+  process.env.FIRESTORE_EMULATOR_HOST = 'localhost:8080';
```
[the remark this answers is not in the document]
xmano: it was give warning. after adding these it went away.

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  const app = initializeTestApp({
+    projectId,
+    auth
+  });
+
+  const db = app.firestore();
```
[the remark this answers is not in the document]
xmano: ok

apps/backend-ops/src/firestore-rules.spec.ts
xmano: Good catch. Will check again if toDeny() worked then we can eliminate it totally.

xmano: Changed to:
` fexpect.extend({
  async toDeny(testPromise) {
`
for now.

Will cleanup further if expect.extend does not give issues.

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+  let db: any;
+
+  beforeAll(async () => {
+    db  = setupFirestore(projectId, {uid: 'uid-c8'});
+    await setFirestoreDB(projectId, 
+          ['firestore.test.rules', 'firestore.rules'], 
+          db, testFixture);
+  });
+
+  afterAll(async () => {
```
GrandSchtroumpf, line 110: Two errors here : 
- You're using async but not await
- You don't return the Promise.all, so you don't await.

So either : 
```typescript
afterAll(() => Promise.all(apps().map(app => app.delete()));
```
Or : 
```typescript
afterAll(async () => {
  await Promise.all(apps().map(app => app.delete()));
});
```

apps/backend-ops/src/firestore-rules.spec.ts
xmano: 1. changed to setFirestore
2. db : if u give type then it causes a type error due to some missing types
3. formatted to a 80-col.. let me know
4. done
5. changed to dataDB
6. done

apps/backend-ops/src/firestore-rules.spec.ts
xmano: Its json data that app expects.. aka fixture data. I am using step suggested in google codelab and for easy setup.

apps/backend-ops/src/firestore-rules.spec.ts
xmano: changed to rulePath 
we now call the setFirestore() twice.

apps/backend-ops/src/firestore-rules.spec.ts
xmano: This is to extend the expect method of Jest in order to get better clarity on the failure message.

Removed this and basic asserts provided by library

Context should make it clear.

apps/backend-ops/src/firestore-rules.spec.ts
xmano: removed this..

apps/backend-ops/src/firestore-rules.spec.ts
xmano: Fixed

apps/backend-ops/src/firestore-rules.spec.ts
xmano: removed

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,135 @@
+﻿import { apps, assertFails, assertSucceeds, initializeTestApp, 
+        loadFirestoreRules } from '@firebase/rules-unit-testing';
+import { testFixture } from './fixtures/data';
+import fs from 'fs';
+
+const setupFirestore = (projectId: string, auth = null) => {
```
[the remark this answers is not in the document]
xmano: ok

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    auth
+  });
+
+  return app.firestore();
+}
+
+/**
+ * Helper function to setup Firestore DB Data
+ */
+async function setFirestore(projectId: string, rulePath: string, db?: any, dataDB?: Object) {
```
GrandSchtroumpf, line 21: I saw your comments on the type, but I cannot let db be untyped, we need to find a solution. Maybe by forcing the type of the return value of `initFirestoreApp` if needed.

---

Please divide into two methods : 
```typescript
function setRules(projectId: string, rulePath: string) {
  return loadFirestoreRules({
    projectId,
    rules: fs.readFileSync(rulePath, "utf8")
  });
}
```
And :
```typescript
function setFirestore(db: Firestore, data: Record<string, Object>) {
  const promises = Object.entries(data).map(([key, doc]) => db.doc(key).set(doc));
  return Promise.all(promises);
}
```

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+}
+
+describe('Blockframe Admin', () => {
+  const projectId = `rules-spec-${Date.now()}`;
+  let db: any;
+
+  beforeAll(async () => {
+    db  = initFirestoreApp(projectId, {uid: 'uid-c8'});
```
GrandSchtroumpf, line 42: Wouldn't it work like that ? 
```typescript
db = initFirestoreApp(projectId, {uid: 'uid-c8'});
await setFirestore(db, fixture);
await setRules(projectId, 'firestore.rules');
```

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    for (const key in dataDB) {
+      const ref = db.doc(key);
+      await ref.set(dataDB[key]);
+    }
+  }
+}
+
+describe('Blockframe Admin', () => {
+  const projectId = `rules-spec-${Date.now()}`;
+  let db: any;
```
GrandSchtroumpf, line 39: type db

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+  test("should allow blockframe admin to read", async () => {
+    const usersRef = db.collection("users");
+    await assertSucceeds(usersRef.get());
+  });
+
+});
+
+describe('General User', () => {
+  const projectId = `rules-spec-${Date.now()}`;
+  let db: any;
```
GrandSchtroumpf, line 58: type db

apps/backend-ops/src/firestore-rules.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  beforeAll(async () => {
+    db  = initFirestoreApp(projectId, {uid: 'uid-user2'});
```
GrandSchtroumpf, line 61: same as above

GrandSchtroumpf (changes requested):

apps/backend-ops/src/firestore-rules.spec.ts
xmano: app.firestore() has a return type of firebase.firestore.Firestore

if we use this type:

let db: firebase.firestore.Firestore;

then in the spec file we get a type mismatch:

```
let db: firebase.firestore.Firestore
Type 'import("d:/C8/bf-dev/node_modules/@firebase/rules-unit-testing/node_modules/firebase/index").default.firestore.Firestore' is not assignable to type 'firebase.firestore.Firestore'.
  Types of property 'runTransaction' are incompatible.
    Type '<T>(updateFunction: (transaction: import("d:/C8/bf-dev/node_modules/@firebase/rules-unit-testing/node_modules/firebase/index").default.firestore.Transaction) => Promise<T>) => Promise<...>' is not assignable to type '<T>(updateFunction: (transaction: firebase.firestore.Transaction) => Promise<T>) => Promise<T>'.
      Types of parameters 'updateFunction' and 'updateFunction' are incompatible.
        Types of parameters 'transaction' and 'transaction' are incompatible.
          Type 'import("d:/C8/bf-dev/node_modules/@firebase/rules-unit-testing/node_modules/firebase/index").default.firestore.Transaction' is not assignable to type 'firebase.firestore.Transaction'.
            Types of property 'set' are incompatible.
              Type '{ <T>(documentRef: import("d:/C8/bf-dev/node_modules/@firebase/rules-unit-testing/node_modules/firebase/index").default.firestore.DocumentReference<T>, data: Partial<T>, options: import("d:/C8/bf-dev/node_modules/@firebase/rules-unit-testing/node_modules/firebase/index").default.firestore.SetOptions): import("d:/C8/...' is not assignable to type '{ <T>(documentRef: firebase.firestore.DocumentReference<T>, data: Partial<T>, options: firebase.firestore.SetOptions): firebase.firestore.Transaction; <T>(documentRef: firebase.firestore.DocumentReference<...>, data: T): firebase.firestore.Transaction; }'.
                Types of parameters 'documentRef' and 'documentRef' are incompatible.
                  Type 'firebase.firestore.DocumentReference<any>' is not assignable to type 'import("d:/C8/bf-dev/node_modules/@firebase/rules-unit-testing/node_modules/firebase/index").default.firestore.DocumentReference<any>'.ts(2322)
```

image

apps/backend-ops/src/firestore-rules.spec.ts
xmano: will apply the changes required above to this line as well.

apps/backend-ops/src/firestore-rules.spec.ts
xmano: type Firestore = ReturnType<typeof initFirestoreApp>;

apps/backend-ops/src/firestore-rules.spec.ts
xmano: type Firestore = ReturnType<typeof initFirestoreApp>;

apps/backend-ops/src/firestore-rules.spec.ts
xmano: type Firestore = ReturnType<typeof initFirestoreApp>;

apps/backend-ops/src/firestore-rules.spec.ts
xmano: Made changes to fix this.

apps/backend-ops/src/firestore-rules.spec.ts
xmano: Made these changes.

GrandSchtroumpf (approved):

PR #4131, opened 2020-11-03
video player now takes in reference as argument

RemcoSimonides:
Video player now takes in a file reference instead of an eventId. Making it possible to use the media-player component for a video file in any component instead of event only

- [ ] move component to media lib


apps/backend-functions/src/player.ts
```diff
@@ -21,160 +21,161 @@ const JWPlayerApi = require('jwplatform');
...
-    const unauthorized: ErrorResultResponse = {
-      error: 'UNAUTHORIZED',
-      result: `You are not authorized to get the information of this video`
-    };
+        if (!('titleId' in event.meta)) {
+          throw new Error(`Event ${data.eventId} is a screening but doesn't have a 'titleId' !`);
+        }
 
-    if (!isUserInvitedToEvent(context.auth.uid, data.eventId)) {
+        if (event.isPrivate && !await isUserInvitedToEvent(uid, event.meta.titleId)) {
```
RemcoSimonides, line 119: I added `await` in front of `isUserInvitedToEvent` because it returns a Promise. I didn't think not awaiting would work but maybe I am wrong?

pldespaigne: 100% correct

apps/backend-functions/src/player.ts
```diff
@@ -21,160 +21,161 @@ const JWPlayerApi = require('jwplatform');
...
+              const creatorId = (docData as MovieDocument)._meta!.createdBy;
+              const { orgId } = await getDocument<User>(`users/${creatorId}`);
+              const org = await getDocument<OrganizationDocument>(`orgs/${orgId}`);
+              const isAuthorizedInMovie = org.userIds.some(userId => userId === uid);
+              if (!isAuthorizedInMovie) {
+                return unauthorized;
+              }
+              break;
+            default:
+              return unauthorized;
```
pldespaigne, line 159: If the user is not invited to the meeting, either he is an org member and we already authorized him at the beginning or he is not an org member and we should return "unauthorized".

But here the check for `case: 'orgs'` & `case: 'movies'` will never be executed

pldespaigne (changes requested):

pldespaigne (approved):

apps/backend-functions/src/player.ts
```diff
@@ -21,160 +20,142 @@ const JWPlayerApi = require('jwplatform');
...
+      result: 'The reference points to an unknown document'
+    }
   }
 
-  let jwPlayerId: string;
+  let access = security === 'public';
 
-  // CHECK FOR A SCREENING
-  if (event.type === 'screening') {
+  if (!access && security === 'protected') {
```
pldespaigne, line 62: those two condition are the same now

Issue #4019, opened 2020-10-21
Patch release 1.9 bug list - summary bis


Mussuranae:
- [x] When country of an organization is too long, it changes the size of the organization-card on the organization list page
![image (3).png](https://images.zenhubusercontent.com/5d51508f6f935933f7158b07/2fb3d147-ed03-4ec9-8cdb-3c404b0430a7)

- [x] Not really a bug, but on the wishlist, we don't distinguish clearly the cell when some text are very long (here the country)
![image (4).png](https://images.zenhubusercontent.com/5d51508f6f935933f7158b07/b7798c45-50e2-4f2d-9a32-7f928376692f)

- [x] Here the margin isn't right, we need the same space between header and text and text and blue bar on the bottom (like 24px or maybe 36px), on organization page and movie page
![image](https://user-images.githubusercontent.com/48033264/96723312-b5302280-13ae-11eb-8678-7595c75f3bef.png)
![image](https://user-images.githubusercontent.com/48033264/96723349-c24d1180-13ae-11eb-855a-2ac23682fccf.png)

- [x] Here I think it's the slug that is displayed, not the label of the static data (movie page main tab)
![image](https://user-images.githubusercontent.com/48033264/96724005-9a11e280-13af-11eb-8714-5286db7171ec.png)

- [x] don't display 'confirmed' status if we haven't value for the running time (movie page main tab)
![image](https://user-images.githubusercontent.com/48033264/96724157-d04f6200-13af-11eb-9d80-6e79b1ab42dd.png)

- [x] No value displayed for language version
![image](https://user-images.githubusercontent.com/48033264/96724339-0391f100-13b0-11eb-972b-04240b38cb1c.png)

- [x] On movie list, the running time in the movie card is displayed differently depending on ????
![image](https://user-images.githubusercontent.com/48033264/96724769-7c914880-13b0-11eb-89ef-70e3314f4d3b.png)
![image](https://user-images.githubusercontent.com/48033264/96724824-8a46ce00-13b0-11eb-886b-96fff39a4ef4.png)

- [x] Same thing with the movie-header
![image (5)](https://user-images.githubusercontent.com/48033264/97014985-50f69580-154b-11eb-8c37-130fce954f08.png)

- [x] On event page, we have the back-arrow two time on the event header and the movie header
![image](https://user-images.githubusercontent.com/48033264/96726782-a0558e00-13b2-11eb-9b0c-89c04b335009.png)

[...]

PR #4139, opened 2020-11-03
Fixing bugs

Mussuranae:
Lot of little bug of the #4019 issue
- fallback image for preview component
- pop up when you leave a page of an event with the good type of event in the message
- display running time correctly everywhere
- css


libs/ui/src/lib/preview-modal/preview.component.ts
```diff
@@ -12,23 +11,25 @@ import { DialogPreviewComponent } from './dialog-preview/dialog-preview.componen
...
 
+  @Input() link: string;
   @Input() form: HostedMediaForm
+  @ViewChild('dialogTemplate') dialogTemplate: TemplateRef<any>
 
   constructor(private dialog: MatDialog, private sanitizer: DomSanitizer) { }
 
-  openModal() {
+  async openModal() {
     let ref: string | SafeUrl;
```
GrandSchtroumpf, line 21: Is there no pipe for that ? I would have created a pipe on the HostMediaForm to deal with that. Like that you can do : 
```html
<img [ref]="form | getMediaUrl" /> 
```

GrandSchtroumpf (approved):

libs/ui/src/lib/preview-modal/preview.component.ts
Mussuranae: The problem is the type is defined in the function depending on the data we get.

PR #4141, opened 2020-11-03
creating indices for org 

fritzschoff:


libs/firebase-utils/src/lib/algolia.ts
```diff
@@ -63,7 +63,17 @@ export function storeSearchableOrg(org: OrganizationDocument, adminKey?: string)
     country: org.addresses.main.country
   };
 
-  return indexBuilder(algolia.indexNameOrganizations, adminKey).saveObject(orgRecord);
+  /* If a org doesn't have access to the app dashboard or marketplace, there is no need to create or update the index */
+  const orgAppAccess = app.filter(a => {
```
GrandSchtroumpf, line 67: By iterating over each module it would be more future-proof
```typescript
const orgAppAccess = app.filter(a => modules.some(m => org.appAccess[a][m]));
```

GrandSchtroumpf (approved):

PR #4145, opened 2020-11-04
Add member form

GrandSchtroumpf:
Use mat-chips instead of form-list


libs/organization/src/lib/organization/pipes/org-address.pipe.ts
```diff
@@ -2,7 +2,7 @@ import { Pipe, PipeTransform, NgModule } from '@angular/core';
 import { Organization, AddressSet } from '../+state';
 import { territories } from '@blockframes/utils/static-model';
 
-@Pipe({ name: 'orgAddress', pure: true })
```
fritzschoff, line 5: image

fritzschoff (approved):

PR #4150, opened 2020-11-04
Issue on org update

fritzschoff:


libs/firebase-utils/src/lib/algolia.ts
```diff
@@ -179,3 +185,13 @@ export function storeSearchableUser(user: PublicUser, adminKey?: string): Promis
     return new Promise(res => res(true));
   }
 }
+
+export function findOrgAppAccess(org: OrganizationDocument) {
+  return app.filter(a => modules.some(m => {
+    try {
+      return org.appAccess[a][m]
```
fritzschoff, line 192: need to do it in a try and catch block cause can't do add a `?` when you use the bracket operator.

apps/backend-functions/src/orgs.ts
```diff
@@ -175,7 +173,19 @@ export async function onOrganizationUpdate(change: Change<FirebaseFirestore.Docu
   // }
 
   // Update algolia's index
-  await storeSearchableOrg(after)
+  if (after.status !== before.status && after.status === 'accepted') {
+    // add flag to org: "isAccepted"
+    after['isAccepted'] = true;
+  }
+  /* If an org gets his accepted status removed, we want to remove it also from all the indices on algolia */
+  else if (before.status === 'accepted' && after.status === 'pending') {
```
GrandSchtroumpf, line 181: I don't see how this could happen...

env/demo/env.demo1.ts
```diff
@@ -52,7 +52,12 @@ export const firebase = {
 export const algolia = {
   appId: '8E9YO1I9HB',
   searchKey: '4a2990a293c0ee0bfde9ebd66270a47f',
-  indexNameOrganizations: 'demo1_orgs',
+  indexNameOrganizations: {
+    festival: 'demo1_festival_org',
+    financiers: 'demo1_financiers_org',
+    catalog: 'demo1_catalog_org',
+    all: 'demo1_orgs'
```
GrandSchtroumpf, line 59: Why do you need that ?

libs/firebase-utils/src/lib/algolia.ts
```diff
@@ -45,11 +45,12 @@ export function setIndexConfiguration(indexName: string, config: IndexSettings,
 
   return indexBuilder(indexName, adminKey).setSettings(config);
 }
+
 // ------------------------------------
 //           ORGANIZATIONS
 // ------------------------------------
 
-export function storeSearchableOrg(org: OrganizationDocument, adminKey?: string): Promise<any> {
+export function storeSearchableOrg(org: OrganizationDocument, updateAllIndex?: boolean, adminKey?: string): Promise<any> {
```
GrandSchtroumpf, line 53: Prefer an options object, it's much easier to read : 
```typescript
export function storeSearchableOrg(org: OrganizationDocument, options: SearchOption = {})
```
Like that you get in the code : 
```typescript
storeSeachableOrg(org, { allIndexes: true })
```

libs/firebase-utils/src/lib/algolia.ts
```diff
@@ -58,18 +59,23 @@ export function storeSearchableOrg(org: OrganizationDocument, adminKey?: string)
   const orgRecord: AlgoliaRecordOrganization = {
     objectID: org.id,
     name: orgName(org),
-    appAccess: getOrgAppAccess(org),
     appModule: getOrgModuleAccess(org),
-    country: org.addresses.main.country
+    country: org.addresses.main.country,
+    isAccepted: org['isAccepted']
```
GrandSchtroumpf, line 64: You don't need the before / after check before. You can do everything here actually...

libs/firebase-utils/src/lib/algolia.ts
```diff
@@ -58,18 +59,23 @@ export function storeSearchableOrg(org: OrganizationDocument, adminKey?: string)
...
   };
 
-  /* If a org doesn't have access to the app dashboard or marketplace, there is no need to create or update the index */
-  const orgAppAccess = app.filter(a => modules.some(m => org.appAccess[a][m]));
+  /* We want to update the all org index when a org is created for cross app features */
+  if (updateAllIndex) {
+    return Promise.all([indexBuilder(algolia.indexNameOrganizations.all, adminKey).saveObject(orgRecord)]);
```
GrandSchtroumpf, line 69: Looks like there is only one item in your array, you don't need the Promise.all.

libs/firebase-utils/src/lib/algolia.ts
GrandSchtroumpf: Yes you can. Like that:
```typescript
app.filter(a => modules.some(m => org.appAccess[a]?.[m]))
```

libs/organization/src/lib/organization/forms/search.form.ts
```diff
@@ -20,7 +20,7 @@ export function createOrganizationSearch(search: Partial<OrganizationSearch> = {
     page: 0,
     hitsPerPage: 8,
     ...search,
-    appAccess: search.appAccess,
+    isAccepted: true,
```
GrandSchtroumpf, line 23: You're forcing the accepted to true here. Are you sure ? Shouldn't it be before the ...search ?

env/demo/env.demo1.ts
fritzschoff: the all index holds every org for cross app login. If you think of the new connection flow, if a user wants to join a org we need an index that holds all org.

libs/firebase-utils/src/lib/algolia.ts
```diff
@@ -45,11 +49,12 @@ export function setIndexConfiguration(indexName: string, config: IndexSettings,
 
   return indexBuilder(indexName, adminKey).setSettings(config);
 }
+
 // ------------------------------------
 //           ORGANIZATIONS
 // ------------------------------------
 
-export function storeSearchableOrg(org: OrganizationDocument, adminKey?: string): Promise<any> {
+export function storeSearchableOrg(org: OrganizationDocument, options: SearchOptions = {}, adminKey?: string): Promise<any> {
```
GrandSchtroumpf, line 57: Shouldn't adminKey be part of the options ?

apps/backend-functions/src/orgs.ts
RemcoSimonides: Sorry for jumping in here but Marie can do this. She recently put an accepted organization back to pending because an organization was shown up in the app which didn't have any accepted movies (which is another issue with querying organization with only accepted movies)

GrandSchtroumpf (approved):

libs/firebase-utils/src/lib/algolia.ts
fritzschoff: if no all index exist, I going to remove the search options

Issue #4101, opened 2020-10-28
MF review 


Jackseed:
- [x] "Explore the marketplace" > "Discover projects"

### Homepage
- [x] Add a "Featured films" category that just takes all the movies accepted on media financiers with a random order. It should use the poster view and be displayed between "New projects" and "In development"
![image](https://user-images.githubusercontent.com/14964443/97696037-45ecb980-1aa5-11eb-8fbc-c90c564442f1.png)

### Footer
- [x] Contact page not working

- [ ] Change Legal Terms & Privacy Policy according to @lladreyt ' proposition

### Movie list
- [x] Animation on movie card click is awesome but goes beyond the card itself, @stephanie-cascade8 should confirm but imo, it should be contained by the card

### Movie view 
- [x] Add the pdf guide for download
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/5d9ed835-196a-4c17-b202-a926887238f1)

- [ ] ~On hover, the number goes transparent so we can't read it well~ (This is in Apex, I won't do)
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/5e927f7e-e3a7-46ee-8706-196b3129250f)

- [x] Scenario should be changed for "Script"
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/6e9f11ee-65ea-4e38-ab6d-9be96b509d72)

- [x] Return on Investment, change the text for (1st sentence is changed):
Your Return on investment is your share of the total amount of proceeds resulting from the distribution of the film. It may be calculated in % by dividing your share of revenue by the initial investment. If the result is superior to 100%, this means your investment was profitable.

- [x] Change CTA button text for "Open discussion" + add a line of text before "Meet the fund to have further information" 
![image](https://user-images.githubusercontent.com/14964443/97697119-d8418d00-1aa6-11eb-947d-0db4c2178e28.png)

- [x] Invest modale:
- "I'm all set. Ready to invest" is on one line
- Align left the text "Please precise the amount you would like to invest.
If you are not ready to provide us with uch kind of information, just leave a message." and put it after "Subject"
- Add a placeholder to Message: "Questions about cast, financing, investment conditions..."
![image](https://user-images.githubusercontent.com/14964443/97697251-08892b80-1aa7-11eb-8503-7df6f84de483.png)

- [x] Shooting dates are not displayed though working on form & summary
[...]

PR #4154, opened 2020-11-04
Cleanup MF again

GrandSchtroumpf:
Solve most of the issues from #4101

### DIFF
--- a/apps/financiers/financiers/src/app/marketplace/title/view/view.component.html
+++ b/apps/financiers/financiers/src/app/marketplace/title/view/view.component.html
@@ -48,7 +48,7 @@
         </mat-form-field>
         <p class="mat-caption">
           Please precise the amount you would like to invest.<br/>
-          If you are not ready to provide us with uch kind of information, just leave a message.
+          If you are not ready to provide us with such kind of information, just leave a message.
         </p>
         <div fxLayout fxLayoutGap="24px">
           <mat-form-field fxFlex appearance="outline">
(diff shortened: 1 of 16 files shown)


apps/financiers/financiers/src/app/marketplace/title/view/view.component.html
```diff
@@ -51,6 +46,10 @@ <h2>How much would you like to invest in this project ?</h2>
           </mat-select>
           <mat-error>Please select a subject</mat-error>
         </mat-form-field>
+        <p class="mat-caption">
+          Please precise the amount you would like to invest.<br/>
+          If you are not ready to provide us with uch kind of information, just leave a message.
```
fritzschoff, line 51: *such

libs/movie/src/lib/movie/components/title-features/title-features.component.scss
```diff
@@ -1,3 +1,9 @@
+p {
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  min-width: 0;
```
fritzschoff, line 5: what does that trick?

fritzschoff (approved):

libs/movie/src/lib/movie/components/title-features/title-features.component.scss
GrandSchtroumpf: I think it's something like, if the display is inline you need to specify at least a min-width for the text-overflow to work.

PR #4165, opened 2020-11-05
Plan A check how many users have attended

RemcoSimonides:


apps/festival/festival/src/app/marketplace/event/session/session.component.ts
```diff
@@ -40,6 +46,28 @@ export class SessionComponent implements OnInit {
...
+        const invitations = this.invitationQuery.getAll().filter(invitation => invitation.docId === event.id)
+        for (let invitation of invitations) {
+          // if status is accepted
+          // and the user is either fromUser and requested to attend, or toUser and invited
+          if (invitation.status === 'accepted' && 
+            (invitation.mode === 'request' && invitation.fromUser?.uid === this.authQuery.userId) ||
+            (invitation.mode === 'invitation' && invitation.toUser?.uid === this.authQuery.userId)) {
+            await this.invitationService.update(invitation.id, { status: 'attended' });
+          }
+        }
```
pldespaigne, line 69: it can be cool to move this logic in the `invitationQuery` this way you could do something :
```TypeScript
const invitation = await this.invitationQuery.getUserInvitationToEvent(uid, eventId);
await this.invitationService.update(invitation.id, { status: 'attended' });
```

RemcoSimonides:
Hee @pldespaigne, can we close this one? I think you've merged the meeting is over page, right?

PR #4167, opened 2020-11-06
update script

fritzschoff:


apps/backend-ops/src/algolia.ts
```diff
@@ -41,6 +41,17 @@ export async function upgradeAlgoliaOrgs(appConfig?: App) {
     const { db } = loadAdminServices();
     const orgsIterator = getCollectionInBatches<OrganizationDocument>(db.collection('orgs'), 'id', 300)
     for await (const orgs of orgsIterator) {
+
+      for await (const org of orgs) {
```
GrandSchtroumpf, line 45: Why are you awaiting of the orgs ? It's not an iterator isn't it ?
I think `for (const org of orgs)` would be enough

apps/backend-ops/src/algolia.ts
```diff
@@ -41,6 +41,17 @@ export async function upgradeAlgoliaOrgs(appConfig?: App) {
...
+        if (org.movieIds.length) {
+          const movies = await Promise.all(org.movieIds.map(id => getDocument<MovieDocument>(`movies/${id}`)));
```
GrandSchtroumpf, line 47: Could you create a reusable method that could be used in many places. You can put it in "firebase-utils" for example

GrandSchtroumpf (changes requested):

libs/firebase-utils/src/lib/util.ts
```diff
@@ -125,3 +128,9 @@ function getKeyFile(keyFile: string): admin.ServiceAccount {
     return eval('require')(keyFile) as admin.ServiceAccount;
   }
 }
+
+export async function hasAcceptedMovies(org: OrganizationDocument) {
+  const movies = await Promise.all(org.movieIds.map(id => getDocument<MovieDocument>(`movies/${id}`)));
+  const acceptedMovies = movies.filter(movie => movie.storeConfig.status === 'accepted')
```
GrandSchtroumpf, line 134: You could do `return movie.some(movie => movie.storeConfig.status === 'accepted')`.
It's better because it stop at the first element.

GrandSchtroumpf (approved):
👍

libs/firebase-utils/src/lib/util.ts
fritzschoff: sounds good!

PR #4202, opened 2020-11-10
Runningtime data

### LINKED ISSUE #4124: Running time status and time are difficult to display depending on each other
At this moment the `runningTime.time` can be type of number or string, so we can have different kind of data in the DB.
But it's difficult to homogenize the way we dislay the data because of that and also, it complicates the way to display the status data.

For example, if the type of the time is equal to string, we can have some display like this one

And actually, I have 4 differents values for the time in my DB : `TBC, TBD, -` or a number (of course it can be null or empty string  also).

I think we have to rework a little the data and let the time only be a number and let the status changed depending on the time exists or not.

!! Check the number of the issue in the code !!

Mussuranae:
Rework running time to set time as a number only
Fix #4124

### DIFF
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0041.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0041.ts
@@ -23,7 +23,8 @@ export async function upgrade(db: Firestore) {
 function updateRunningTime(runningTime) {
   if (typeof runningTime.time === 'number') {
     runningTime.time = runningTime.time
-  } else delete runningTime.time;
+  }
+  else delete runningTime.time;
 
   return runningTime;
 }
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0041.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0041.ts
@@ -1,0 +1,24 @@
+import { Firestore } from '@blockframes/firebase-utils';
+
+
+export async function upgrade(db: Firestore) {
+  const movies = await db.collection('movies').get();
+  const batch = db.batch();
+
+  movies.docs.map(movieDoc => {
+    const movie = movieDoc.data();
+
+    const newData = {
+      ...movie,
+      runningTime: {
+        ...movie.runningTime,
+        time: (typeof movie.runningTime.time === 'number') ? movie.runningTime.time : '',
+      }
+    };
+
+    return batch.set(movieDoc.ref, newData);
+  })
+
+  console.log('Running time updated on movies');
+  await batch.commit();
+}
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0041.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0041.ts
@@ -10,10 +10,7 @@ export async function upgrade(db: Firestore) {
 
     const newData = {
       ...movie,
-      runningTime: {
-        ...movie.runningTime,
-        time: (typeof movie.runningTime.time === 'number') ? movie.runningTime.time : '',
-      }
+      runningTime: updateRunningTime(movie.runningTime)
     };
 
     return batch.set(movieDoc.ref, newData);
@@ -21,4 +18,12 @@ export async function upgrade(db: Firestore) {
 
   console.log('Running time updated on movies');
   await batch.commit();
+}
+
+function updateRunningTime(runningTime) {
+  if (typeof runningTime.time === 'number') {
+    runningTime.time = runningTime.time
+  } else delete runningTime.time;
+
+  return runningTime;
 }
--- a/libs/movie/src/lib/movie/+state/movie.model.ts
+++ b/libs/movie/src/lib/movie/+state/movie.model.ts
@@ -209,7 +209,6 @@ export function createReleaseYear(release: Partial<MovieRelease> = {}): MovieRel
 
 export function createRunningTime(runningTime: Partial<MovieRunningTime> = {}): MovieRunningTime {
   return {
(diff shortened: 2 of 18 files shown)


libs/movie/src/lib/movie/pipes/running-time.pipe.ts
```diff
@@ -7,7 +7,7 @@ import { Pipe, PipeTransform, NgModule } from '@angular/core';
 })
 export class RunningTimePipe implements PipeTransform {
   transform(value: string | number): string {
-    return (typeof value === 'number') ? `${value}min` : 'TBC';
+    return (typeof value === 'number') ? `${value} min` : '';
```
GrandSchtroumpf, line 10: This should deal with the whole runningTime. Also extract it in a function to be reusable : 
```typesript
function formatRunningTime(value: MovieRunningTime) {
  if (value.time && value.status) return ...;
  if (value.time && !value.status) return ...;
  if (!value.time && value.status) return ...;
  return '';
}
....
transform(value: MovieRunningTime): string {
  return formatRunningTime(value);
}
```

GrandSchtroumpf (changes requested):

libs/movie/src/lib/movie/pipes/running-time.pipe.ts
Mussuranae: I can't display the status because the design is a little different depending on which platform the user is (marketplace or dashboard)

libs/firebase-utils/src/lib/firestoreMigrations/0041.ts
```diff
@@ -0,0 +1,29 @@
...
+  })
+
+  console.log('Running time updated on movies');
+  await batch.commit();
+}
+
+function updateRunningTime(runningTime) {
+  if (typeof runningTime.time === 'number') {
+    runningTime.time = runningTime.time
+  } else delete runningTime.time;
```
GrandSchtroumpf, line 26: Add a new line

libs/movie/src/lib/movie/marketplace/main/main.component.html
```diff
@@ -69,18 +69,11 @@ <h5>Genres</h5>
...
-          && movie.runningTime.time !== 'TBC'
-          && movie.runningTime.time !== 'TBD'
-          && movie.runningTime.time !== '-';
-          else runningTimeStatus">Confirmed</ng-container>
-          <ng-template #runningTimeStatus>
-            {{ movie.runningTime.status | toLabel: 'screeningStatus' }}
-          </ng-template>
+          <ng-container *ngIf="movie.runningTime.time">(</ng-container>
+          <ng-container>{{ movie.runningTime.status | toLabel: 'screeningStatus' }}</ng-container>
+          <ng-container *ngIf="movie.runningTime.time">)</ng-container>
```
GrandSchtroumpf, line 76: No, I don't want to see that again.

GrandSchtroumpf (changes requested):

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -209,7 +209,6 @@ export function createReleaseYear(release: Partial<MovieRelease> = {}): MovieRel
 
 export function createRunningTime(runningTime: Partial<MovieRunningTime> = {}): MovieRunningTime {
   return {
-    time: null,
     status: '',
```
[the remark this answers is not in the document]
Mussuranae: '' is the default value, there is no mandatory status for the moment, it if on discussion (or at least Anastasiia think about it).

libs/movie/src/lib/movie/pipes/running-time.pipe.ts
```diff
@@ -1,13 +1,49 @@
 import { Pipe, PipeTransform, NgModule } from '@angular/core';
+import { MovieRunningTime } from '../+state/movie.firestore';
+import { screeningStatus } from '@blockframes/utils/static-model/static-model';
 
+/**
+ * Format the running time to show 'time min' and the status if needed
+ * @param runningTime Object runningTime
+ * @param isStatusNeeded For some component, like movie-card, we don't want the status in parenthesis
+ */
+export function formatRunningTime(runningTime?: MovieRunningTime, isStatusNeeded: boolean = true) {
```
GrandSchtroumpf, line 10:
```typescript
export function formatRunningTime(runningTime?: MovieRunningTime, showStatus: boolean = true) {
  if (!runningTime) return '';
  const { time, status } = runningTime;
  if (showStatus) {
    if (time && status) return `${time} min (${screeningStatus[status]})`;
    if (!status) return `${time} min`;
    if (!time) return screeningStatus[status];
  }
  if (time) {
    return status === "estimated" ? `≈ ${time} min` : `${time} min`;
  }
}
```

libs/ui/src/lib/marketplace/pages/wishlist/wishlist.component.html
```diff
@@ -36,7 +36,7 @@ <h1>Wishlist</h1>
       <ng-container matColumnDef="runningTime">
         <th mat-header-cell *matHeaderCellDef>Running time</th>
         <td mat-cell *matCellDef="let element">
-          <span>{{ element.runningTime.time | duration}}</span>
+          <span>{{ element.runningTime.time | getRunningTime }}</span>
```
GrandSchtroumpf, line 39: Remove .time : `<span>{{ element.runningTime | getRunningTime }}</span>`

GrandSchtroumpf (approved):

PR #4205, opened 2020-11-10
fix bugs & detect completed import

george43g:
- fix random bugs
- better logs
- detect completed firestore import


phpgeekfr (approved):
LGTM

PR #4212, opened 2020-11-10
fix bugs & detect error import

george43g:
- error out if import not successful
- refactor how the latest file is detected
- stop storing anonDb history
- code improvements


apps/backend-ops/src/admin.ts
```diff
@@ -3,17 +3,17 @@
...
 
 /**
  * Trigger a firestore database restore operation for the given project
  */
-export async function restore() {
+export async function restore(file?: string) {
   const { db } = loadAdminServices();
   const pubsub = new PubSub({ projectId: firebase.projectId });
-  await pubsub.topic('firestore').publish(Buffer.from('import'));
+  await pubsub.topic('firestore').publish(Buffer.from('import'), { ...file && { file } });
```
phpgeekfr, line 16: I don't understand this `, { ...file && { file } }` can you explain?

libs/firebase-utils/src/lib/util.ts
```diff
@@ -118,7 +118,7 @@ export function loadAdminServices(): AdminServices {
 
 export const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
 
-function getKeyFile(keyFile: string): admin.ServiceAccount {
+export function getKeyFile(keyFile: string): admin.ServiceAccount {
```
phpgeekfr, line 121: Can you rename to something more relevant ? Like getServiceAccount or getCredentials ?

apps/backend-ops/src/admin.ts
phpgeekfr: ok nvm. I was not familiar with this

libs/firebase-utils/src/lib/copy-db-from-ci.ts
```diff
@@ -5,25 +5,19 @@ import * as admin from 'firebase-admin';
 import { existsSync, mkdirSync } from 'fs';
 import { catchErrors } from './util';
 
+export const latestAnonDbFilename = 'LATEST-ANONYMIZED.jsonl'
+
 export async function copyDbFromCi(storage: admin.storage.Storage, ci: admin.app.App) {
```
phpgeekfr, line 10: can you rename to copyAnonDbFromCi ? Since this method only works for anon db

libs/firebase-utils/src/lib/firestore/import.ts
```diff
@@ -2,33 +2,42 @@ import readline from 'readline';
...
+    console.log('Clearing the database');
+    await clear(db);
+    await importFirestoreFromGFile(restoreFile, db);
+    await setImportRunning(false);
+    await endMaintenance();
+  } else {
+    await triggerImportError();
+    await endMaintenance();
+    throw new Error('Nothing to restore');
+  }
```
phpgeekfr, line 39: In case of "else" you `endMaintenance` but it is not started. It is started in ` if (fileExists) {` only

tools/scripts/upload-db.ts
```diff
@@ -1,5 +1,5 @@
 import 'tsconfig-paths/register';
```
phpgeekfr, line 1: can you rename this file and `"db:upload-ci": "ts-node -P tools/tsconfig.tools.json tools/scripts/upload-db.ts"` in package.json to upload-anondb-to-ci or something like that ? Since it's dedicated to anonymized db it will be clearer

phpgeekfr (approved):
LGTM but look at comments before merging

PR #4213, opened 2020-11-10
Fix bugs in video upload

RemcoSimonides:
- [x] Remove UI issue with file-upload
- [x] Delete reference from db in case an upload fails
- [x] Delete reference from db from array (still photo / otherVideos) - not sure which reference to delete


libs/media/src/lib/components/upload/widget/upload-widget.component.ts
```diff
@@ -27,8 +40,52 @@ export class UploadWidgetComponent {
...
+    const fieldToUpdate = filePathElements.join('.');
+
+    const currentMediaValue = getDeepValue(docData, fieldToUpdate);
+
+    if (Array.isArray(currentMediaValue)) {
+      // find the right value
+      // Actually quite dangerous for otherVideos since we only know ref and not title
+      // If the same ref is used for a different video, then that one might be deleted
+      // It would be nice to have some sort of ID; this would also solve similar 'bug' in file-explorer component
+    } else {
```
RemcoSimonides, line 83: Read comment - describes an issue I have with array (still photos / otherVideos)

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
```diff
@@ -16,8 +17,20 @@ import { slideUp } from '@blockframes/utils/animations/fade';
 export class UploadWidgetComponent {
   @HostBinding('@slideUp') animation = true;
 
+  @HostListener('window:beforeunload', [ '$event' ])
```
RemcoSimonides, line 20: Is this allowed? Function gets called when the page is destroyed and it does work.

I am afraid ngOnDestroy won't be called for this component since it stays open as long as there are tasks; and there will always be tasks if paused/cancelled. Only when the tasks are removed, the component closes. Deleting reference is therefore also called on `removeTask()` function

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
```diff
@@ -27,8 +40,52 @@ export class UploadWidgetComponent {
...
+    }
+
+    // remove the file name
+    filePathElements.pop();
+
+    const docSnapshot = await this.db.collection(collection).doc(docId).get().toPromise();
+    const docData = docSnapshot.data();
+    const fieldToUpdate = filePathElements.join('.');
+
+    const currentMediaValue = getDeepValue(docData, fieldToUpdate);
```
phpgeekfr, line 76: I think that you can use `getDocAndPath` method here

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
phpgeekfr: I'm not the good person to answer this. @GrandSchtroumpf  ?

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
```diff
@@ -16,8 +17,20 @@ import { slideUp } from '@blockframes/utils/animations/fade';
 export class UploadWidgetComponent {
   @HostBinding('@slideUp') animation = true;
 
+  @HostListener('window:beforeunload', [ '$event' ])
+  beforeUnloadHandler() {
+    // check if there are tasks paused or errored; if so, then remove the reference
+    for (const task of this.tasks.value) {
+      const state = task.task.snapshot.state as 'success' | 'paused' | 'running' | 'canceled'
+      if (state === 'paused' || state === 'canceled') {
```
phpgeekfr, line 25: we remove ref if state is paused ? What happens if I un-pause ?

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
phpgeekfr: For this, I think we can say that we will not use same ref for different video ( I guess).
Here is how it's done in `apps/backend-functions/src/media.ts`

```
    if (Array.isArray(savedRef)) {
      savedRef = savedRef.map(e => e.ref || e).find(ref => ref === filePath) || '';
    } else if (!!savedRef.ref) {
      savedRef = savedRef.ref;
    }
```

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
GrandSchtroumpf: If the page reload you shouldn't care about anything, the browser will kill every process for you.

pldespaigne: SGTM, from what I recall we are using similar `beforeunload` elsewhere in the code

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
pldespaigne: I think it's ok because if you have paused upload when you leave the page (`beforeunload`) those will never complete and so we should clean them from the db

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
```diff
@@ -27,8 +40,52 @@ export class UploadWidgetComponent {
...
   }
 
+  /**
+   * Removes the already-set References in the db 
+   * This function has similarities to function getDocAndPath in @blockframes/firebase-utils
+   */
+  async removeReference(task: AngularFireUploadTask) {
+
+    const path = task.task.snapshot.ref.fullPath;
+    const filePathElements = path.split('/');
```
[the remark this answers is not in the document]
GrandSchtroumpf: You'll need to create a function to get the segments:
```typescript
function storageSegment(path: string) {
  const segments = path.split('/');
  const tmp = segments.shift();
  const privacy = segments.shift();
  const filename = segments.pop();
  const field = segments.pop();
  const docPath = segment;
  return { tmp, privacy, docPath, field, filename };
}
```

libs/media/src/lib/+state/media.service.ts
```diff
@@ -74,7 +76,7 @@ export class MediaService {
     if (!this.overlayRef) {
       this.overlayRef = this.overlay.create(this.overlayOptions);
       const instance = new ComponentPortal(UploadWidgetComponent);
-      instance.injector = Injector.create({ providers: [{ provide: 'tasks', useValue: this._tasks }] });
+      instance.injector = Injector.create({ providers: [{ provide: 'tasks', useValue: this._tasks }, { provide: 'db', useValue: this.db }] });
```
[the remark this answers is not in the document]
RemcoSimonides: Doing that gave an error. This solved it.

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
GrandSchtroumpf: For me you should either use `beforeunload` to tell the user that the uploads are not done and if he close the page they will break. Else you can remove this code.

RemcoSimonides: We don't
image

libs/movie/src/lib/movie/form/media-videos/media-videos.component.html
```diff
@@ -34,10 +34,10 @@ <h3>Other Videos</h3>
...
         <mat-label>Video title (optional)</mat-label>
-        <input matInput [formControl]="currentOtherVideo.title" >
+        <input matInput [formControl]="activeVideo.title" >
       </mat-form-field>
       <file-upload
-        [form]="currentOtherVideo.ref"
+        [form]="activeVideo.ref"
         [storagePath]="getPath('otherVideos')"
         allowedFileType="video"
         filePrivacy="public">
```
GrandSchtroumpf, line 43: Isn't it "public" the default ?

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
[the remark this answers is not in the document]
RemcoSimonides: This function was similar as getDocAndPath. I've merged the two functions into one; the older one (which was only for back-end) also throws errors if the path is invalid.

libs/utils/src/lib/pipes/deep-key.pipe.ts
```diff
@@ -1,6 +1,17 @@
...
+/**
+ * Helper function to dynamically access object value pointed by the `path` param, like the rxjs pluck function
+ * @param object usually the result object from Algolia
+ * @param path string representing the path to the value, usually `this.pathToValue` or `this.displayWithPath`
+ * @example
+ * const object = { main: { nested: { name: 'Joe' } } };
+ * const path = 'main.nested.name';
+ * this.resolve(result); // 'Joe'
+ */
+export const getDeepValue = (object: Object, path: string) => path.split('.').reduce((result, key) => result?.[key], object);
```
RemcoSimonides, line 13: This pipe now also exports the main function (as it was in the `value-by-path.pipe.ts`. I could also move this function to `utils/helper.ts` for example?

libs/utils/src/lib/file-sanitizer.ts
```diff
@@ -2,6 +2,56 @@ export const privacies = ['public', 'protected'] as const;
...
+    isInTmpDir = true;
+  }
+
+  let security: string;
+  // remove "protected/"" or "public/"
+  if (privacies.includes(filePathElements[0] as any)) {
+    security = filePathElements.shift();
+  }
+
+  const collection = filePathElements.shift();
```
GrandSchtroumpf, line 33: This only works if you're in a collection, not sub-collection

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
```diff
@@ -27,8 +42,45 @@ export class UploadWidgetComponent {
...
+  /**
+   * Removes the already-set References in the db 
+   * This function has similarities to function getDocAndPath in @blockframes/firebase-utils
+   */
+  async removeReference(task: AngularFireUploadTask) {
+
+    const path = task.task.snapshot.ref.fullPath;
+    const { collection, docId, fieldToUpdate } = dissectFilePath(path)
+
+    const docSnapshot = await this.db.collection(collection).doc(docId).get().toPromise();
```
GrandSchtroumpf, line 59:
```typescript
const { docPath, field } = dissectFilePath(path);
const snapshot = await this.db.doc(docPath).get().toPromise();
```

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
```diff
@@ -27,8 +42,45 @@ export class UploadWidgetComponent {
...
+    const docData = docSnapshot.data();
+
+    const currentMediaValue = getDeepValue(docData, fieldToUpdate);
```
GrandSchtroumpf, line 62: name it `const media = getDeepValue(docData, fieldToUpdate);` and type it.

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
```diff
@@ -27,8 +42,45 @@ export class UploadWidgetComponent {
...
+
+    if (Array.isArray(currentMediaValue)) {
```
GrandSchtroumpf, line 64:
```typescript
if (Array.isArray(media)) {
  const update = media.filter(ref => typeof ref === 'string' ? ref !== path : ref.ref !== path);
  return snapshot.update({ [field]: update });
} else {
  return snapshot.update({ [field]: null });
}
```

GrandSchtroumpf: OR : `snapshot.get(field).update(null)`
https://firebase.google.com/docs/reference/js/firebase.firestore.FieldValue

libs/utils/src/lib/file-sanitizer.ts
```diff
@@ -2,6 +2,56 @@ export const privacies = ['public', 'protected'] as const;
...
+    const solution = 'It should at least contain 3 slash.';
+    const example = 'Example: public/collection/id/field/fileName';
+    throw new Error(`${error} ${solution}\n${example}`);
+  }
+
+  // remove tmp/
+  let isInTmpDir = false;
+  if (filePathElements[0] === tempUploadDir) {
+    filePathElements.shift();
+    filePath = filePathElements.join('/');
```
GrandSchtroumpf, line 23: remove it

libs/utils/src/lib/file-sanitizer.ts
```diff
@@ -2,6 +2,56 @@ export const privacies = ['public', 'protected'] as const;
...
+
+  if (filePathElements.length < 4) {
+    const error = `Upload Error : File Path ${filePath} is malformed.`;
+    const solution = 'It should at least contain 3 slash.';
+    const example = 'Example: public/collection/id/field/fileName';
+    throw new Error(`${error} ${solution}\n${example}`);
+  }
+
+  // remove tmp/
+  let isInTmpDir = false;
```
GrandSchtroumpf, line 20:
```typescript
const isTmp = segments[0] === 'tmp';
if (isTmp) segment.shift();
```

libs/utils/src/lib/file-sanitizer.ts
```diff
@@ -2,6 +2,56 @@ export const privacies = ['public', 'protected'] as const;
...
+  if (filePathElements[0] === tempUploadDir) {
+    filePathElements.shift();
+    filePath = filePathElements.join('/');
+    isInTmpDir = true;
+  }
+
+  let security: string;
```
GrandSchtroumpf, line 27:
```typescript
const privacy = segments.shift();
```

libs/utils/src/lib/file-sanitizer.ts
```diff
@@ -2,6 +2,56 @@ export const privacies = ['public', 'protected'] as const;
...
+  // remove "protected/"" or "public/"
+  if (privacies.includes(filePathElements[0] as any)) {
+    security = filePathElements.shift();
+  }
+
+  const collection = filePathElements.shift();
+  const docId = filePathElements.shift();
+
+  if (!docId || !collection) {
```
GrandSchtroumpf, line 36:
```typescript
if (docPath.split('/').length % 2 !== 0) {
  throw new Error(`Doc path "${docPath}" is not a document.`);
}
```

GrandSchtroumpf (changes requested):

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
RemcoSimonides: `snapshot.get()` returns the data of the field. It's not possible to call update on the data.

libs/utils/src/lib/file-sanitizer.ts
RemcoSimonides: This isn't going to work because `field` is sometimes separated by a `/`. For example this path: `/public/movies/36JVeZckRc4c8rJim78J/promotional.videos/screener/abc.mp4`

It would be better to have `/promotional.videos.screener/` instead of `/promotional.videos/screener/` and be able to support subcollections; but in order to fix that we need to write a migration etc..

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
RemcoSimonides: `media.filter()` isn't callable because the types weren't matching. `media.findIndex()` does work.

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
```diff
@@ -27,8 +44,34 @@ export class UploadWidgetComponent {
...
+
+    if (Array.isArray(media)) {
+      // Still Photos (string[]), Documents (HostedMediaWithMetadata[]), Notes & Statements (MovieNote) or OtherVideos (HostedVideo[])
+      const index = media.findIndex(ref => typeof ref === 'string' ? ref === path : ref.ref === path);
+      media.splice(index, 1);
+      return snapshot.ref.update({ [field]: media });
+    } else {
+      // Logo, profile image, and more (string), Screener (HostedVideo)
+      // Soon we will have files of type HostedMediaWithMetadata here too
+      // Chose to only update the reference field and not the whole object to not lose potential useful data (e.g. jwPlayerId)
```
RemcoSimonides, line 72: Hee @pldespaigne , Should we remove the whole media object if an upload fails, or should I only empty the ref?

libs/utils/src/lib/file-sanitizer.ts
RemcoSimonides: docPath isn't going to work because `field` is sometimes separated by a `/`. For example this path: `/public/movies/36JVeZckRc4c8rJim78J/promotional.videos/screener/abc.mp4`

It would be better to have `/promotional.videos.screener/` instead of `/promotional.videos/screener/` and be able to support subcollections; but in order to fix that we need to write a migration etc.. (noted in issue #4241)

GrandSchtroumpf (approved):

libs/media/src/lib/components/upload/widget/upload-widget.component.ts
pldespaigne: Looks like a UX question but I would say no since you don't want your 3 paragraphs description to be deleted because of a bad connection ^^

PR #4215, opened 2020-11-11
Fix(e2e): Change selector for reaching login/sign-up buttons.

xmano:
Fixes org create e2e failure


xmano:
It should fix even account creation issue as well.

apps/catalog/catalog-e2e/src/support/pages/marketplace/NavbarPage.ts
```diff
@@ -21,16 +21,16 @@ export default abstract class NavbarPage {
   }
 
   public clickWishlist() {
-    cy.get('main', {timeout: TO.PAGE_ELEMENT}).scrollTo('top');
+    //cy.get('main', {timeout: TO.PAGE_ELEMENT}).scrollTo('top');
```
phpgeekfr, line 24: remove comments if you don't intend to use them in future

phpgeekfr (approved):

apps/catalog/catalog-e2e/src/support/pages/marketplace/NavbarPage.ts
xmano: Ok.

PR #4228, opened 2020-11-11
Fix initialising index and removed app filters

RemcoSimonides:
Algolia Index was initialised incorrectly and removed app filters (there are now separate indices for each app)


libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
```diff
@@ -29,7 +29,6 @@ function splitValue(value: string, keycodes: number[]) {
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
-  public indexName: string;
```
RemcoSimonides, line 32: Wasn't used

fritzschoff (approved):

PR #4244, opened 2020-11-12
fix #4181 Events' organiser have wrong invitation status on their events

### LINKED ISSUE #4181: Events' organiser have wrong invitation status on their events
All the users from an organisation that created an event see the invitation status of the last invitation to this event on the event list instead of their own. 
For example here I created an event with a user A and do not have an invitation for it. With a different user B from another org I requested the access to that event. 
I re-connect with A and see the invitation status of B as my own. If I refuse that invitation, it will be declined even if it's B invitation that has been declined

phpgeekfr:


libs/event/src/lib/components/screening-item/screening-item.component.ts
```diff
@@ -16,16 +17,25 @@ export class ScreeningItemComponent {
...
+  public uid = this.authQuery.userId;
 
   @Input() set event(screening: ScreeningEvent) {
     this.screening = screening;
     this.movie = screening.movie;
     this.poster = screening.movie?.poster;
     this.invitation$ = this.invitationQuery.selectAll().pipe(
-      map(invits => invits.find(e => e.docId === screening.id))
+      map(invits => invits.find(e => {
+        return e.docId === screening.id && (
```
GrandSchtroumpf, line 28: Maybe create a reusable function for that in the invitation.model : 
```typescript
function isInvitationOwner(invitation: Invitation, id: string) {
  return ....;
}
...

map(invits => invits.find(invit =>isInvitationOwner(invit, this.authQuery.userId)))
```

GrandSchtroumpf (approved):

libs/event/src/lib/components/screening-item/screening-item.component.ts
phpgeekfr: maybe not `isInvitationOwner` but `isInvitationRecipient`

GrandSchtroumpf (approved):

Issue #3928, opened 2020-10-12
Change admin email recipient for support@appName


Jackseed:
Admin emails should be redirected to their dedicated support@appName adresses.
It includes:
- [x]  org creation 
- [x] app access
- [x] demo request on landing page

Jackseed:
@phpgeekfr also need to change the email recipient for the contact for demo request on landing page

phpgeekfr:
@Jackseed  does 
- support@archipelmarket.com
- support@archipelcontent.com
- support@mediafinanciers.com

already exists ?

Jackseed:
@phpgeekfr now they do :)

phpgeekfr:
New env vars to put on CI :
```
#prod
blockframes_CASCADE8_ADMIN_CATALOG => support@archipelcontent.com
blockframes_CASCADE8_ADMIN_FESTIVAL => support@archipelmarket.com
blockframes_CASCADE8_ADMIN_FINANCIERS => support@mediafinanciers.com

#staging
blockframesStaging_CASCADE8_ADMIN_CATALOG => dev+staging_catalog@blockframes.io
blockframesStaging_CASCADE8_ADMIN_FESTIVAL => dev+staging_festival@blockframes.io
blockframesStaging_CASCADE8_ADMIN_FINANCIERS => dev+staging_financiers@blockframes.io

```

For other envs, `projectId_CASCADE8_ADMIN` will be used as fallback for `projectId_CASCADE8_ADMIN_appName`

PR #4252, opened 2020-11-13
Feat #3928 & #2826

phpgeekfr:
- Change admin email recipient for support@appName #3928
- Set the from part for the mail sent in `onUserCreate` backend trigger #2826


GrandSchtroumpf (approved):
maybe change "admin" to "support"

Issue #4254, opened 2020-11-16
Nightly E2E fails due to migration failure


xmano:
As seen in the logs:
https://app.circleci.com/pipelines/github/blockframes/blockframes/12640/workflows/b4e2db76-0686-4ba3-980c-c3a45315dcfd/jobs/22974

```
applying migration: 42
TypeError: orgs.map(...).flat is not a function
    at Object.<anonymous> (/home/circleci/repo/libs/firebase-utils/src/lib/firestoreMigrations/0042.ts:9:52)
    at step (/home/circleci/repo/libs/firebase-utils/src/lib/firestoreMigrations/0042.ts:33:23)
    at Object.next (/home/circleci/repo/libs/firebase-utils/src/lib/firestoreMigrations/0042.ts:14:53)
    at fulfilled (/home/circleci/repo/libs/firebase-utils/src/lib/firestoreMigrations/0042.ts:5:58)
    at process._tickCallback (internal/process/next_tick.js:68:7)
the migration failed, revert'ing!
```

This is preventing nightly E2E to run. I am raising another issue to make the process immune to such failure.

For the migration issue need 2 things:

1. Fix to handle org docs not in proper structure .
2. A unit-test that can check these conditions so any future code changes does not skip it.

PR #4258, opened 2020-11-16
Issue 4254

fritzschoff:
ref #4254

### DIFF
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0042.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0042.ts
@@ -13,7 +13,7 @@ export async function upgrade(db: Firestore) {
         const movieDoc = moviesCol.docs.find(doc => doc.ref.id === id);
         const movie = movieDoc?.data();
         /* Check if the movie object holds an orgIds array already, if not, add it. */
-        if (movie?.['orgIds']) {
+        if (movie?.['orgIds'] && !movie?.['orgIds'].includes(id)) {
             movie['orgIds'].push(id);
         } else {
             Object.assign(movie, { orgIds: [id] })
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0042.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0042.ts
@@ -5,7 +5,7 @@ export async function upgrade(db: Firestore) {
     const orgsCol = await db.collection('orgs').get();
     const moviesCol = await db.collection('movies').get();
     const batch = db.batch();
-    const orgs = orgsCol.docs.map(doc => doc.data())
+    const orgs = orgsCol.docs.map(doc => doc.data()).filter(org => org?.id)
     const movieIds = orgs.map(org => org.movieIds).flat().filter(id => id);
     movieIds.map(id => {
         const movieDoc = moviesCol.docs.find(doc => doc.ref.id === id);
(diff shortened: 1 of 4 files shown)


libs/firebase-utils/src/lib/firestoreMigrations/0043.ts
```diff
@@ -5,12 +5,11 @@ export async function upgrade(db: Firestore) {
     const orgsCol = await db.collection('orgs').get();
     const batch = db.batch();
 
-    runChunks(orgsCol.docs, orgDoc => {
+    runChunks(orgsCol.docs, async orgDoc => {
```
xmano, line 8: Dont see await being used inside. Is async needed here?

libs/firebase-utils/src/lib/firestoreMigrations/0043.ts
```diff
@@ -3,14 +3,12 @@ import { runChunks } from '../firebase-utils';
...
         const org = orgDoc.data();
         delete org?.movieIds
-        batch.set(orgDoc.ref, org)
+        await orgDoc.ref.set(org)
     })
 
     console.log('deleted movieIds from orgs collection');
-    await batch.commit();
```
xmano, line 15: I am not very sure about the impact of this. I would request review from either @phpgeekfr  or @Mussuranae here to avoid potential pitfalls here.

xmano (commented):
For batching I would request backend guys to provide their review.

libs/firebase-utils/src/lib/firestoreMigrations/0043.ts
Mussuranae, line 15: Why don't use the batch ? I though it was a better way to write on the DB, also don't you have to `await Promise.all()` to apply your changes ?

libs/firebase-utils/src/lib/firestoreMigrations/0042.ts
```diff
@@ -5,7 +5,7 @@ export async function upgrade(db: Firestore) {
     const orgsCol = await db.collection('orgs').get();
     const moviesCol = await db.collection('movies').get();
     const batch = db.batch();
-    const orgs = orgsCol.docs.map(doc => doc.data())
+    const orgs = orgsCol.docs.map(doc => doc.data()).filter(org => org?.id)
     const movieIds = orgs.map(org => org.movieIds).flat().filter(id => id);
     movieIds.map(id => {
         const movieDoc = moviesCol.docs.find(doc => doc.ref.id === id);
```
Mussuranae, line 11: There is something I don't understand here (lines below, but can't add comment on those lines), a movie can't belong to several organization right ? so why put the organization id in an array ?

Mussuranae (changes requested):
Do you update also the movie firestore ? don't forget to tell Bruce or I when you change some data, we have the draw.oi to update

libs/firebase-utils/src/lib/firestoreMigrations/0043.ts
fritzschoff: I can't use batch because there is a limit of 300 commits in one batch and we have way more orgs than that

libs/firebase-utils/src/lib/firestoreMigrations/0042.ts
fritzschoff: it may can belong to several orgs in the future. There is no reason to limit yourself now to store it in an array. Looking into the future of orgs and operators, they might share some movies.

Mussuranae (approved):

PR #4259, opened 2020-11-16
Change video list to bf form list

RemcoSimonides:
- [x] Other video list now uses bf-form-list component
- [x] Added input to edit the position of the list. Default is top - as it used to be like this
- [x] Fixed a bug where wrong item was deleted from the formlist
- [x] File upload component has issues recognizing when to update the UI. This issue was fixed in the media-videos component, however the same changes cannot be done (easily) in the bf-form-list component due to changing Control Types.
- [x] Remove old code from the media-videos component

### DIFF
--- a/libs/ui/src/lib/form/list/form-list.component.html
+++ b/libs/ui/src/lib/form/list/form-list.component.html
@@ -1,4 +1,4 @@
-<section [fxLayout]="_listPosition">
+<section [fxLayout]="layout[listPosition]">
   <!---------->
   <!-- LIST -->
   <!---------->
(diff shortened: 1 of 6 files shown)


libs/ui/src/lib/form/list/form-list.component.html
```diff
@@ -1,52 +1,54 @@
-<!---------->
-<!-- LIST -->
-<!---------->
+<section [fxLayout]="layout[listPosition]">
```
GrandSchtroumpf, line 1: Best would be to do it with the @HostBinding. But let's learn that another day ;)

GrandSchtroumpf (approved):

PR #4273, opened 2020-11-18
Reverse list and automatically add new form

RemcoSimonides:
- [x] Fix for other videos having to manually add another one and therefore the drop component disappeared. Now the component is always shown
- [x] Since the list is below the form, it's better for the UX to add the items to the top of the list instead of the bottom (especially if the lists gets long)

### DIFF
--- a/libs/movie/src/lib/movie/form/media-videos/media-videos.component.html
+++ b/libs/movie/src/lib/movie/form/media-videos/media-videos.component.html
@@ -32,7 +32,7 @@
     <section>
       <p>Upload other videos such as trailer, teaser, promotional clip, etc...</p>
 
-      <bf-form-list [form]="videoList" buttonText="Add a video" listPosition="bottom">
+      <bf-form-list [form]="videoList" buttonText="Add a video" listPosition="bottom" autoAdd reverseList>
         <ng-template itemRef let-video>
           {{ !!video.ref.fileName ? video.ref.fileName : (video.ref.ref | fileName) }}
         </ng-template>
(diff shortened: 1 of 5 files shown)


libs/ui/src/lib/form/list/form-list.component.html
```diff
@@ -3,7 +3,7 @@
   <!-- LIST -->
   <!---------->
 
-  <ng-container *ngIf="list$ | async as list">
+  <ng-container *ngIf="list$ | async | reverse:reverseList as list">
```
RemcoSimonides, line 6: Is there another way to make this pipe only work when `reverseList` is true?

libs/movie/src/lib/movie/form/media-videos/media-videos.component.html
```diff
@@ -32,7 +32,7 @@ <h3>Other Videos</h3>
     <section>
       <p>Upload other videos such as trailer, teaser, promotional clip, etc...</p>
 
-      <bf-form-list [form]="videoList" buttonText="Add a video" listPosition="bottom">
+      <bf-form-list [form]="videoList" buttonText="Add a video" listPosition="bottom" autoAdd reverseList>
```
GrandSchtroumpf, line 35: Remove buttonText maybe

libs/ui/src/lib/form/list/form-list.component.html
GrandSchtroumpf: Reverse on the list observable : 
```typescript
list$ = combineLatest([
  this.form.valueChanges.pipe(startWith(this.form.value), distinctUntilChange()),
  reverseList$
]).pipe(
  map(([list, reverse]) => reverse ? list.reverse() : list)
)
```
With a Behavior subject on the reverse :
```typescript
reverseList$ = new BehaviorSubject(false);
@Input() @boolean set reverseList(shouldReverse: boolean) {
  this.reverseList$.next(shouldReverse);
}
```

GrandSchtroumpf (approved):

PR #4277, opened 2020-11-18
Issue algolia service

fritzschoff:


libs/utils/src/lib/algolia/helper.utils.ts
```diff
@@ -0,0 +1,29 @@
...
+    const factory = user => new FormControl(createAlogliaUser(user));
+    return FormList.factory<AlgoliaRecordUser, FormControl>([], factory, validators)
+}
+
+
+export function parseFilters(filters: MovieIndexFilters, operator: 'OR ' | 'AND ' = 'OR '): string {
+    const filter = []
+    for (const key of Object.keys(filters)) {
+        if (typeof filters[key] === 'object') {
+            parseFilters(filters[key], operator)
```
GrandSchtroumpf, line 23: This doesn't do anything.
This one should work: 
```typescript
function parseFilters(filters: MovieIndexFilters, operator: ' AND ' | ' OR ' = ' OR ', prefix?: string) {
  const filter = []
  for (const [key, value] of Object.entries(filters)) {
    const path = [prefix, key].filter(v => !!v).join('.');
    if (value === null || value === undefined) {
      filter.push(undefined);
    } else if (typeof value === 'object') {
      const deepFilter = parseFilters(value, operator, path);
      filter.push(deepFilter);
    } else {
      filter.push(`${path} >= ${value ?? 0}`)
    }
  }
  return filter.filter(v => !!v).join(operator);
}
```

GrandSchtroumpf (changes requested):

libs/utils/src/lib/algolia/algolia.service.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    private appName: App;
+
+    constructor(private routerQuery: RouterQuery) {
+        this.appName = getCurrentApp(this.routerQuery);
+    }
+
+    getIndex(name: 'movie' | 'org' | 'user') {
+        if (name === 'user') {
+            return algoliasearch(algolia.appId, algolia.searchKey).initIndex(algoliaIndex[name]);
```
GrandSchtroumpf, line 20: Here you initialize the index each time you call the method. I think it's an mistake. Use memorization as I told you in the previous comments

libs/utils/src/lib/algolia/algolia.service.ts
```diff
@@ -0,0 +1,35 @@
...
+        } else {
+            return algoliasearch(algolia.appId, algolia.searchKey).initIndex(algoliaIndex[name][this.appName]);
+        }
+    }
+
+    query<K extends keyof AlgoliaQueries>(name: K, config: AlgoliaQueries[K]) {
```
GrandSchtroumpf, line 26: What about the return type ?

GrandSchtroumpf (approved):

libs/utils/src/lib/algolia/algolia.service.ts
fritzschoff: will throw a type error since algolia typed the function. Tried to cast it with `as` but it didn't let me.

fritzschoff: Will check again if I can override it

PR #4278, opened 2020-11-18
run backup/restore locally and upgrade use command

george43g:
- experiment: run backup/restore locally rather than in cloud function
- upgraded 'use' command - experimental


phpgeekfr (commented):
`importRunning` can be removed I think.
along with `importComplete`

phpgeekfr (approved):

Issue #3965, opened 2020-10-14
Update the stats table for events by adding: Org name (label: Company Name), org activity (label: Company Activity) & org country (label Country)


phpgeekfr:
- [x] Update the stats table for events by adding: Org name (label: Company Name), org activity (label: Company Activity)
![image](https://user-images.githubusercontent.com/14964443/94548896-af349f00-0251-11eb-9fa2-ae38352eddbc.png)

Edit from Vincent: we also need to add the org country (label: Country) to the table by taking the org address country. 

_Originally posted by @phpgeekfr in https://github.com/blockframes/blockframes/issue_comments/707651149_

Jackseed:
I created a new issue that is more broader than this one: #4257. 
It could be done in 2 times but I prefer to reference this in case you prefer to do all at once

phpgeekfr:
#4257 will be done after since the two tables (invitations & attendees) are not fetching data from the same source

phpgeekfr:
@Jackseed does this fit your needs ? 

![image](https://user-images.githubusercontent.com/1194344/99667509-d0a15280-2a6c-11eb-8c80-3fe20ffb4f4d.png)

Jackseed:
@phpgeekfr it's perfect! Thanks :)

PR #4280, opened 2020-11-18
#3965 Update the stats table for events

phpgeekfr:


apps/backend-functions/src/bigQuery.ts
```diff
@@ -142,18 +143,25 @@ const findByUserId = (users: PublicUser[], userId: string) => {
...
 
-const createEventAnalytics = (result: any, user: PublicUser | undefined): EventAnalytics => {
+const createEventAnalytics = (result: any, user: PublicUser | undefined, org: OrganizationDocument | undefined): EventAnalytics => {
   return {
     ...result,
     email: user?.email,
     firstName: user?.firstName,
-    lastName: user?.lastName
+    lastName: user?.lastName,
+    orgName: org ? orgName(org) : '',
```
GrandSchtroumpf, line 162: maybe orgName could return directly an empty string if there is no org?

apps/backend-functions/src/bigQuery.ts
```diff
@@ -188,16 +196,24 @@ export const requestEventAnalytics = async (
       return getDocument<PublicUser>(`users/${row.userId}`);
     });
     const eventsUsers = await Promise.all(eventsUsersPromises);
-    const eventsUsersNotInOrg = eventsUsers.filter(u => !org.userIds.includes(u.uid));
+    const eventsUsersNotInOrg = eventsUsers.filter(u => !!u && !org.userIds.includes(u.uid));
     const userIdsNotInOrg = eventsUsersNotInOrg.map(u => u.uid);
     // Clean rows without users of the same org
     rows = rows.filter(row => userIdsNotInOrg.includes(row.userId));
 
+    const orgsPromises = eventsUsersNotInOrg.map(u => {
```
GrandSchtroumpf, line 204: Use a filter instead : 
```typescript
const promises = eventsUsersNotInOrg.filter(u => !!u.orgId).map(u => getDocument(`orgs/${u.orgId}`))
```

libs/invitation/src/lib/components/guest-item/guest-item.component.html
```diff
@@ -12,7 +12,10 @@
     <span *ngIf="user.firstName" matLine>{{ user | displayUser | async }}</span>
     <span *ngIf="!user.firstName" matLine>{{ user.email }}</span>
     <span matLine>
-      <small class="mat-caption">{{ invitation.date | date: 'short' }}</small>
+      <small class="mat-caption">
+        {{ invitation.date | date: 'short' }} 
+        ({{ invitation.mode === 'request' ? 'Buyer request' : 'Sales agent invitation' }})
```
GrandSchtroumpf, line 17: this sounds really app specific. Sale agent is only for festival. I don't know how to solve that, but we should be careful with this kind of additional data.

GrandSchtroumpf (approved):

Issue #3933, opened 2020-10-12
CRM App - Delete an org


Jackseed:
Add a button on the org page to delete an org (and all objects related to it). Before deleting, the confirmation modale should be used.
Related objects are:
- users included in userIds
- permissions
- notifications
- invitations
- movies included in movieIds

Bonus: if it's not too difficult, would be good to add it on the org list also with the ability to select multiple orgs to delete at once. Here again the confirmation modale should be used

phpgeekfr:
Display a popup before deleting that shows what is going to be deleted and add an input where user must type "DELETE" to make the delete button working + do the same for user deletion on crm app

PR #4287, opened 2020-11-19
Delete Dialog

Mussuranae:
Create a delete dialog with a form to confirm deletion of movie/organization/user in the CRM app.
Made with Samir.

Related to #3933


libs/admin/src/lib/admin-panel/components/delete-dialog/delete-dialog.component.ts
```diff
@@ -0,0 +1,42 @@
...
+      deletion?: string,
+      onConfirm?: () => void
+    },
+    public dialogRef: MatDialogRef<DeleteDialogComponent>,
+    private snackbar: MatSnackBar
+  ) {}
+
+  public confirm() {
+    if (this.deleteConfirm.value !== 'DELETE' || this.deleteConfirm.invalid) {
+      this.snackbar.open(`You can\'t delete this ${this.data.entity}.`, 'close', { duration: 5000});
```
phpgeekfr, line 30: A better message would be: you did not entered the expected value. with "you can't delete this .." we don't know if this is because DELETE was not entered correctly or if deletion crashed

libs/admin/src/lib/admin-panel/components/delete-dialog/delete-dialog.component.ts
```diff
@@ -0,0 +1,42 @@
...
+      this.dialogRef.close(false);
+    }
+    if (this.deleteConfirm.value === 'DELETE' && this.deleteConfirm.valid) {
```
phpgeekfr, line 33: I think you can do "else" directly instead of ` if (this.deleteConfirm.value === 'DELETE' && this.deleteConfirm.valid) {`

libs/admin/src/lib/admin-panel/pages/user/user.component.ts
```diff
@@ -146,11 +146,10 @@ export class UserComponent implements OnInit {
...
     }
 
-    this.dialog.open(ConfirmComponent, {
+    this.dialog.open(DeleteDialogComponent, {
       data: {
-        title: 'Are you sure you want to delete this user?',
-        question: 'this action is irreversible',
-        buttonName: 'Yes',
+        entity: 'user',
+        deletion: 'This user will be deleted from this organization',
```
phpgeekfr, line 152: which org ? can you update text ? user will be deleted from the app

phpgeekfr (approved):

PR #4292, opened 2020-11-19
clean up maintenance mode activation

george43g:
- clean up maintenance mode
- minor code cleanup/refactor

### DIFF
--- a/apps/backend-ops/src/users.ts
+++ b/apps/backend-ops/src/users.ts
@@ -7,7 +7,6 @@ import { differenceBy } from 'lodash';
 import { loadAdminServices, getCollectionInBatches, sleep } from '@blockframes/firebase-utils';
 import readline from 'readline';
 import { Auth, UserRecord, upsertWatermark, runChunks, JsonlDbRecord } from '@blockframes/firebase-utils';
-import { startMaintenance, endMaintenance, isInMaintenance } from '@blockframes/firebase-utils';
 import { deleteAllUsers, importAllUsers } from '@blockframes/testing/firebase';
 import * as env from '@env';
 import { User } from '@blockframes/user/types';
@@ -111,12 +110,10 @@ async function getUsersFromDb(db:FirebaseFirestore.Firestore ) {
  */
 export async function syncUsers(jsonl?: JsonlDbRecord[]): Promise<any> {
   const { auth, db } = loadAdminServices();
-  await startMaintenance();
   const expectedUsers = jsonl ? readUsersFromJsonlFixture(jsonl) : await getUsersFromDb(db);
   await deleteAllUsers(auth);
   const createResult = await importAllUsers(auth, expectedUsers);
   console.log(createResult);
-  await endMaintenance();
 }
 
 export async function printUsers(): Promise<any> {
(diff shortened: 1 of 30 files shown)


apps/backend-ops/src/users.ts
```diff
@@ -192,20 +189,15 @@ export async function createUsers(): Promise<any> {
...
 
-  const users = await db.collection('users').get();
+  const usersBatch = getCollectionInBatches<User>(db.collection('users'), 'uid')
+  for await (const users of usersBatch) {
+    await runChunks(
+      users,
+      (user: User) => upsertWatermark(user, storageBucket, storage),
+      env?.['heavyChunkSize'] ?? 10
+    );
+  }
```
phpgeekfr, line 200: what is the purpose of this change?

libs/firebase-utils/src/lib/firebase-utils.ts
```diff
@@ -53,14 +53,21 @@ export async function getDocAndPath(fullPath: string | undefined) {
...
+
+export async function runChunks<K, T extends (arg: K) => Promise<any>>(rows: K[], cb: T, rowsConcurrency?: number, verbose = true) {
   const chunks = chunk(rows, rowsConcurrency || env?.['chunkSize'] || 10);
+  const output: AsyncReturnType<T>[] = [];
   for (let i = 0; i < chunks.length; i++) {
     const c = chunks[i];
     if (verbose) { console.log(`Processing chunk ${i + 1}/${chunks.length}`); }
     const promises = c.map(cb);
-    await Promise.all(promises);
+    output.concat(await Promise.all(promises))
```
phpgeekfr, line 68: missing `;` 
so runChunks is now typed ?

phpgeekfr (approved):

apps/backend-ops/src/users.ts
phpgeekfr: With this, we have a new behavior in logs:
```
ssing chunk 84/93
Processing chunk 85/93
Processing chunk 86/93
Processing chunk 87/93
Processing chunk 88/93
Processing chunk 89/93
Processing chunk 90/93
Processing chunk 91/93
Processing chunk 92/93
Processing chunk 93/93
Processing chunk 1/93
Processing chunk 2/93
Processing chunk 3/93
Processing chunk 4/93
Processing chunk 5/93
Processing chunk 6/93
Processing chunk 7/93
Processing chunk 8/93
Processing chunk 9/93
Processing chunk 10/93
Processing chunk 11/93
```
because of batches,  count start from 1 again at each batch


(not an issue but just to let you know)

libs/firebase-utils/src/lib/firebase-utils.ts
george43g: yes!

phpgeekfr: GJ

apps/backend-ops/src/users.ts
george43g: I did this to future-proof it. Once it gets too big, we cant rely on a single collection().get() to get the entire collection, as previously we've seen that as the response gets really big, things go missing. This allows us to batch both the reads and the writes.

phpgeekfr:
In fact, can you add setMaintenance in preparForTEsting also ? It will prevent ttriggers to be trigerred and will also prevent unwanted emails

phpgeekfr:
> In fact, can you add setMaintenance in preparForTEsting also ? It will prevent ttriggers to be trigerred and will also prevent unwanted emails

and this will also be usefull when we run prepareForTesting on staging for example

Issue #4183, opened 2020-11-09
Debug screening banner on movie view


Jackseed:
A new banner has been implemented on marketplace movie view
![banner.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/b887f059-2a76-4385-ab8c-1ce9c7bc1947)
It needs a few debug:
1. Update the names to be the same as their number: "Third session" / "Fourth session" ...
2. The select should list all the sessions available
3. Time should display minutes and timezone (e.g. "5:00AM - 5:30AM (GMT+1)") 
4. The time and date should be centered
5. This banner has to be implemented in all the movie view tabs for the festival app only

Jackseed:
Feedback from prod: this banner enables user to have multiple invitations on the same event (either by previous invitation or by requesting access through the event page/list). 
This is a non wanted behaviour as there could be one accepted and one refused and it's confusing for the sales agent to receive a request from a user already invited. 

Could we limit user invitation for a screening to one only? Or disable the "Ask for invitation" button if the user already has an invitation on the event selected? 
I can create a specific issue if needed but I don't know what's the best option

Mussuranae:
I prefer to do the second comment in an other tickets. I created one for that : see #4311 
It will be a little more complicated because of who ask for invitation or who invite the buyers. We will do it in a second time.

Jackseed:
Alright 👍

PR #4312, opened 2020-11-23
debug screenings banner

SamirTouira:
Number of sessions modified (5 sessions max), select option reviewed and adapted and css a bit modified
Fix: #4183

### DIFF
--- a/libs/movie/src/lib/movie/components/upcoming-screenings/upcoming-screenings.component.html
+++ b/libs/movie/src/lib/movie/components/upcoming-screenings/upcoming-screenings.component.html
@@ -2,22 +2,12 @@
   <ng-container *ngIf="screenings$ | async as screenings">
     <h3>Next Screenings</h3>
     <div fxLayout="row" fxLayoutAlign="center center" fxLayoutGap="32px">
-      <ng-container *ngFor="let screening of screenings; let index = index">
+      <ng-container *ngFor="let screening of screenings; let index = index; let isLast = last">
         <div fxLayout="column" fxLayoutAlign="center center">
-          <h5>{{ index === 0 ? 'FIRST SESSION' : 'SECOND SESSION' }}</h5>
-          <span class="mat-body-1" fxLayout="row" fxLayoutAlign="center start">
-            {{ screening | eventRange }}
-          </span>
-          <span class="mat-body-1" fxLayout="row" fxLayoutAlign="center start">
-            <!-- Start hour -->
-            {{ screening.start | date: 'h a' }}
-            <!-- Splitter -->
-            -
-            <!-- End Hour -->
-            {{ screening.end | date: 'h a'}}
-          </span>
+          <h5>{{ sessions[index] }} Session</h5>
+          <span class="mat-body-1">{{ screening | eventRange }}</span>
         </div>
-        <mat-divider *ngIf="index === 0" vertical></mat-divider>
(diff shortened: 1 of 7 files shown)


libs/movie/src/lib/movie/components/upcoming-screenings/upcoming-screenings.component.ts
```diff
@@ -22,12 +22,19 @@ import { OrganizationService } from '@blockframes/organization/+state';
 })
 export class UpcomingScreeningsComponent {
 
-  public sessionCtrl = new FormControl('first');
+  public sessions = {
+    0: 'first',
```
GrandSchtroumpf, line 26: I think you want to have a titlecase here `First`, 'Second', ...

libs/movie/src/lib/movie/components/upcoming-screenings/upcoming-screenings.component.html
```diff
@@ -2,31 +2,24 @@
...
-          </span>
-          <span class="mat-body-1" fxLayout="row" fxLayoutAlign="center start">
-            <!-- Start hour -->
-            {{ screening.start | date: 'h a' }}
-            <!-- Splitter -->
-            -
-            <!-- End Hour -->
-            {{ screening.end | date: 'h a'}}
-          </span>
+          <h5>{{ sessions[index] }} Session</h5>
```
GrandSchtroumpf, line 7: Either ` | titlecase` or apply the titlecase directly in the property "sessions" (see comment below)

libs/movie/src/lib/movie/components/upcoming-screenings/upcoming-screenings.component.ts
```diff
@@ -22,12 +22,19 @@ import { OrganizationService } from '@blockframes/organization/+state';
 })
 export class UpcomingScreeningsComponent {
 
-  public sessionCtrl = new FormControl('first');
+  public sessions = {
```
GrandSchtroumpf, line 25: Use an array when you work with number index in order: `public sessions = ['First', 'Second', 'Thrid' ,...]`

GrandSchtroumpf (changes requested):

libs/movie/src/lib/movie/components/upcoming-screenings/upcoming-screenings.component.html
SamirTouira: We had apply the propety `text-transform: uppercase` to the <h5> for sessions[index] in the scss file, are you ok with that ?

libs/movie/src/lib/movie/components/upcoming-screenings/upcoming-screenings.component.ts
SamirTouira: The titlecase for here is already used in the upcoming-screenings.component.html : 

`     18   <mat-option
      19   *ngFor="let screening of screenings; let index = index"
      20    [value]="index">
      21   {{ sessions[index] | titlecase }} Session
      22   </mat-option>`

Is this correct ?

libs/movie/src/lib/movie/components/upcoming-screenings/upcoming-screenings.component.html
GrandSchtroumpf: ok

GrandSchtroumpf (approved):

PR #4319, opened 2020-11-24
updated wishlist structure

phpgeekfr:


libs/organization/src/lib/organization/components/wishlist-button/wishlist-button.component.ts
```diff
@@ -46,20 +40,14 @@ export class WishlistButtonComponent implements OnInit {
...
 
-  ngOnInit() {
-    this.toggle$ = this.orgQuery.selectActive().pipe(
-      map(org => {
-        return org.wishlist
-          .filter(({ status }) => status === 'pending')
-          .some(({ movieIds }) => movieIds.includes(this.movieId));
-      })
-    );
+  public toggle$() {
```
GrandSchtroumpf, line 49: make it a property : 
```typescript
ngOnInit() {
  this.toggle$ = this.cartQuery.isAddedToWishlist(this.movieId);
}
```
Only if the movieId is not supposed to change. Else we would need to listen on the change (safer) : 
```typescript
private _movieId = new BehaviorSubject('');
toggle$ = this._movieId.pipe(
  switchMap(id => this.cartQuery.isAddedToWishlist(this.movieId))
);
@Input()
set movieId(id: string) {
  this._movieId.next(id);
}
get movieId() {
  return this._movieId.getValue();
}
```

GrandSchtroumpf (approved):

PR #4320, opened 2020-11-24
WIP Feat 3932

phpgeekfr:

PR #4324, opened 2020-11-24
Crm org delete

### LINKED ISSUE #3933: CRM App - Delete an org

Mussuranae:
See #3933 
Prepare the real deletion of an organization for the CRM app.
On organization deletion, I had also deletion of movies / invitations / events / notifications related to this organization.
Fix #3933


apps/backend-functions/src/orgs.ts
```diff
@@ -200,10 +200,59 @@ export async function onOrganizationDelete(
...
   const org = orgSnapshot.data() as OrganizationDocument;
 
+  // Reset the orgId field on user document
+  for (const userId of org.userIds) {
+    const userSnapshot = await db.doc(`users/${userId}`).get();
+    const user = userSnapshot.data() as PublicUser;
+    await removeMemberPermissionsAndOrgId(user);
+  }
+
+  // Delete movies belonging to organization and its sub-collection if there is
```
phpgeekfr, line 210: you can remove `and its sub-collection if there is` because this is done by onMovieDelete trigger (that will also clean other things related to movie)

apps/backend-functions/src/orgs.ts
```diff
@@ -200,10 +200,59 @@ export async function onOrganizationDelete(
...
+    await invit.ref.delete();
+  }
+
+  const invitationsToOrgCollectionRef = db.collection('invitations').where('toOrg.id', '==', org.id);
+  const invitationsToOrgSnap = await invitationsToOrgCollectionRef.get();
+  for (const invit of invitationsToOrgSnap.docs) {
+    await invit.ref.delete();
+  }
+
+  // TODO CLEAN STORAGE
```
phpgeekfr, line 250: I think you can remove it, all medias seems to be deleted in `cleanOrgMedias`

apps/backend-functions/src/orgs.ts
```diff
@@ -200,10 +200,59 @@ export async function onOrganizationDelete(
 
   const org = orgSnapshot.data() as OrganizationDocument;
 
+  // Reset the orgId field on user document
+  for (const userId of org.userIds) {
+    const userSnapshot = await db.doc(`users/${userId}`).get();
+    const user = userSnapshot.data() as PublicUser;
+    await removeMemberPermissionsAndOrgId(user);
```
phpgeekfr, line 207: you use this `removeMemberPermissionsAndOrgId` but the whole permission document should be removed not updated.
You should: remove the permission doc and set orgId: null on each users

phpgeekfr (changes requested):
For the contracts parts :
```
db.collection('contracts').where('partyIds', 'array-contains', org.id);
```
And for each contract returned, remove the org from 'parties' array (the partyIds will be self updated)

Also, the simulation deletion is missing. 
See `simulateDeletion` in https://github.com/blockframes/blockframes/pull/4320/files

example : 
image

libs/admin/src/lib/admin-panel/pages/organization/organization.component.ts
```diff
@@ -195,7 +195,7 @@ export class OrganizationComponent implements OnInit {
         warning: 'You will also delete everything regarding this organization',
         confirmationWord: 'delete',
         onConfirm: () => {
-          // this.organizationService.remove(this.orgId);
+          this.organizationService.remove(this.orgId);
```
phpgeekfr, line 198: missing await

phpgeekfr (approved):

PR #4352, opened 2020-11-27
Add default queries

GrandSchtroumpf:


apps/cms/src/app/home/org-titles/org-titles.component.ts
```diff
@@ -1,4 +1,4 @@
-import { NgModule, ChangeDetectionStrategy, Component, Input } from '@angular/core';
+import { NgModule, ChangeDetectionStrategy, Component, Input, ChangeDetectorRef } from '@angular/core';
```
fritzschoff, line 1: if imported then use pls

apps/cms/src/app/home/titles/titles.component.ts
```diff
@@ -8,25 +8,25 @@ import { MatRadioChange, MatRadioModule } from '@angular/material/radio';
...
-import { FirestoreFormModule, firestoreQuery, FirestoreQuery } from '../../forms/firestore';
+import { FirestoreFormModule, firestoreQuery, FirestoreQuery, titlesFromApp } from '../../forms/firestore';
 import { getTitlesQueryFn, toMap } from '../pipes';
 import { map, shareReplay, switchMap } from 'rxjs/operators';
 
 interface TitlesSection extends Section {
   _type: 'titles',
   title: string;
-  mode: 'poster' | 'banner' | 'slider';
+  mode: string; // 'poster' | 'banner' | 'slider';
```
fritzschoff, line 22: guess that is WIP

fritzschoff (approved):

apps/cms/src/app/home/titles/titles.component.ts
GrandSchtroumpf: good catch thx

PR #4353, opened 2020-11-27
Session guard

pldespaigne:


libs/event/src/lib/guard/event.guard.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    // if user wasn't invited OR hasn't accepted yet
+    if (!hasUserAccepted) {
+      return this.router.parseUrl(`/c/o/marketplace/event/${event.id}`);
+    }
+
+    return true;
+  }
+
+  canDeactivate(component: any, currentRoute: ActivatedRouteSnapshot, currentState: RouterStateSnapshot, nextState: RouterStateSnapshot): (boolean |Observable<boolean>) {
```
fritzschoff, line 58: break the line. not everybody got a 2389497238947238949 inch screen

libs/event/src/lib/guard/session.guard.ts
```diff
@@ -1,49 +1,39 @@
+
```
fritzschoff, line 1: fairly unusual

fritzschoff (approved):

PR #4354, opened 2020-11-27
provide emulator & bucket helper functions for JSONL migration

george43g:
as per title


apps/backend-ops/src/emulator.ts
```diff
@@ -0,0 +1,11 @@
+import { gracefullyKillEmulator, importPrepareFirestoreEmulatorBackup, startFirestoreEmulatorWithImport, waitForProcOutput } from '@blockframes/firebase-utils';
+import { resolve } from 'path';
+
+
+export async function importEmulatorFromBucket(bucketUrl: string) {
+  const emulatorBackupPath = resolve(process.cwd(), 'tmp', 'test');
```
phpgeekfr, line 6: is `'tmp', 'test'` definitive path ?

phpgeekfr (approved):

apps/backend-ops/src/emulator.ts
george43g: this is temporary, as this will change moving forward. The important thing is that one parameter can change the dir used without depending on multiple vars/hardcoded vars

PR #4355, opened 2020-11-27
debug screener

phpgeekfr:


phpgeekfr:
this fix two issues we have to access the screener room but :
image

pldespaigne (approved):

PR #4371, opened 2020-12-01
Algolia org on user

### LINKED ISSUE #4356: Cannot display org name in invitation search
Following: #4260
Cannot display org name in invitation search because data is coming from Algolia.
If we want to display it, we need to push orgId on Algolia users indexe or fetch orgId from userId coming from Algolia.

Mussuranae:
Update user index on algolia to put the org denomination of the user on it.
Fix #4356

### DIFF
--- a/apps/backend-functions/src/orgs.ts
+++ b/apps/backend-functions/src/orgs.ts
@@ -16,7 +16,7 @@ import { getAdminIds, getAppUrl, getOrgAppKey, getDocument, createPublicOrganiza
 import { ErrorResultResponse } from './utils';
 import { cleanOrgMedias } from './media';
 import { Change, EventContext } from 'firebase-functions';
-import { algolia, deleteObject, storeSearchableOrg, findOrgAppAccess, hasAcceptedMovies } from '@blockframes/firebase-utils';
+import { algolia, deleteObject, storeSearchableOrg, findOrgAppAccess, hasAcceptedMovies, storeSearchableUser } from '@blockframes/firebase-utils';
 
 /** Create a notification with user and org. */
 function notifyUser(toUserId: string, notificationType: NotificationType, org: OrganizationDocument, user: PublicUser) {
@@ -118,8 +118,18 @@ export async function onOrganizationUpdate(change: Change<FirebaseFirestore.Docu
   }
 
   // Update algolia's index
-  if (before.denomination.full !== after.denomination.full) {
-    throw new Error('Organization name cannot be changed !'); // this will require to change the org ENS name, for now we throw to prevent silent bug
+  if (before.denomination.full !== after.denomination.full
+  || before.denomination.public !== after.denomination.public) {
+    after.userIds.forEach(async userId => {
+      const userDocRef = db.doc(`users/${userId}`);
+      const userSnap = await userDocRef.get();
+      const userData = userSnap.data() as PublicUser;
+      await storeSearchableUser(userData);
+    })
+    // Normally, we can't change the name of the organization, but an admin can change it directly on database or on CRM
+    // I let a warning to be aware of this update inside the logs functions
+    //! When users will be able to change their org's name, we have to take care of the org ENS name
+    console.warn('Organization\'s name has been updated, be careful !');
   }
 
   await cleanOrgMedias(before, after);
--- a/apps/backend-functions/src/orgs.ts
+++ b/apps/backend-functions/src/orgs.ts
@@ -126,9 +126,6 @@ export async function onOrganizationUpdate(change: Change<FirebaseFirestore.Docu
       const userData = userSnap.data() as PublicUser;
       await storeSearchableUser(userData);
     }
-    // Normally, we can't change the name of the organization, but an admin can change it directly on database or on CRM
-    // I let a warning to be aware of this update inside the logs functions
-    //! When users will be able to change their org's name, we have to take care of the org ENS name
     console.warn('Organization\'s name has been updated, be careful !');
   }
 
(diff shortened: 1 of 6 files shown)


apps/backend-functions/src/users.ts
```diff
@@ -118,12 +118,13 @@ export async function onUserUpdate(change: functions.Change<FirebaseFirestore.Do
...
 
-  // if name, email or avatar has changed : update algolia record
+  // if name, email, avatar or orgId has changed : update algolia record
   if (
     before.firstName !== after.firstName ||
     before.lastName !== after.lastName ||
     before.email !== after.email ||
-    before.avatar !== after.avatar
+    before.avatar !== after.avatar ||
+    before.orgId !== after.orgId
```
phpgeekfr, line 127: you test the orgId which is fine, but don't we need to test the org name also ? What if it changed ?

libs/firebase-utils/src/lib/algolia/algolia.ts
```diff
@@ -164,19 +165,27 @@ export function storeSearchableMovie(
...
+      orgData = org.data();
+    }
+
     const userRecord: AlgoliaUser = {
       objectID: user.uid,
       email: user.email,
       firstName: user.firstName ?? '',
       lastName: user.lastName ?? '',
       avatar: user.avatar ?? '',
+      orgName: orgData ? orgData.denomination : {}
```
phpgeekfr, line 188: You should use `orgName` here like this we only store in algolia the value we use for the org name.

libs/invitation/src/lib/form/user/user.component.html
```diff
@@ -18,7 +18,8 @@
       <ng-template let-user>
         <div class="profile-item" fxLayout fxLayoutAlign="strt center" fxLayoutGap="8px">
           <img [ref]="user.avatar" asset="profil_user.webp" alt="Profile">
-          <span *ngIf="user.firstName">{{ user | displayUser | async }}</span> <!-- #4356 user doest not have orgId attribute-->
+          <span *ngIf="user.firstName">{{ user.firstName && user.lastName | titlecase }}</span>
```
phpgeekfr, line 21: Why not using `{{ user | displayUser | async }}` here ?

libs/invitation/src/lib/form/user/user.component.html
```diff
@@ -18,7 +18,8 @@
...
+          <span *ngIf="user.orgName.public || user.orgName.full">({{ user.orgName.public ? user.orgName.public : user.orgName.full }})</span>
```
phpgeekfr, line 22: If you use orgName to put data in algolia like I said above, you could just do `<span *ngIf="user.orgName">({{ user.orgName }})</span>`

phpgeekfr (changes requested):

apps/backend-functions/src/users.ts
Mussuranae: So I checked that in the code, and on the `onOrganizationUpdate` funciton, there was something to forbidden the update of the name
``` 
if (!before || !after || !after.denomination.full) {
    console.error('Invalid org data, before:', before, 'after:', after);
    throw new Error('organization update function got invalid org data');
  }
```
in the `orgs.ts` in backend-functions.
So I don't know if it's very useful now :/

libs/invitation/src/lib/form/user/user.component.html
Mussuranae: Because the pipe use the orgId on the user, but as we don't save the orgId but directly the organization's name, I can't use it anymore. I tried to update the pipe but can't succeed between the PublicUser and AlgoliaUser type....

apps/backend-functions/src/users.ts
phpgeekfr: what if an admin change the name directly in db ?

libs/invitation/src/lib/form/user/user.component.html
phpgeekfr: `Because the pipe use the orgId on the user,` I made you review my PR about this

phpgeekfr: I don't see any reason why you can't use `{{ user | displayUser | async }}` since I added a check on orgId to prevent error

phpgeekfr: ` I tried to update the pipe but can't succeed between the PublicUser and AlgoliaUser type` for this pipe to work we only need lastName & firstName. both publicUser and algoliaUsers have this attributes

apps/backend-functions/src/users.ts
phpgeekfr: and it's always better to evaluate the good data. what if in the future we allow orgName changes? Will you remember that this need to be updated?

libs/invitation/src/lib/form/user/user.component.html
Mussuranae: See last commit....

apps/backend-functions/src/orgs.ts
```diff
@@ -118,8 +118,18 @@ export async function onOrganizationUpdate(change: Change<FirebaseFirestore.Docu
...
+  || before.denomination.public !== after.denomination.public) {
+    for(const userId of after.userIds) {
+      const userDocRef = db.doc(`users/${userId}`);
+      const userSnap = await userDocRef.get();
+      const userData = userSnap.data() as PublicUser;
+      await storeSearchableUser(userData);
+    }
+    // Normally, we can't change the name of the organization, but an admin can change it directly on database or on CRM
+    // I let a warning to be aware of this update inside the logs functions
+    //! When users will be able to change their org's name, we have to take care of the org ENS name
```
phpgeekfr, line 131: Outdated comment, org name can be changed on app.

ENS was an ethereum related thing. this can be removed

phpgeekfr (approved):

Issue #4290, opened 2020-11-19
Design review - Meetings


Jackseed:
#### Global
Following this wireframe
![image](https://user-images.githubusercontent.com/14964443/100338057-6a6e7f80-2fd8-11eb-961f-a79a6f99dc89.png)
- [x] Add a subtitle below "My Files" title: "This is your File Storage. Its structure is predetermined. We trust you to upload the right content in the right place!"
- [x] Rename the org section "Company Files" 
- [x] Add an arrow below the breadcrumbs and above the files table that goes back to the previous page
- [x] Sort movies by international title by default
![image](https://user-images.githubusercontent.com/14964443/100338365-c9cc8f80-2fd8-11eb-95c4-1d4516724086.png)
- [x] Rename Still Photo « Images » 
![image](https://user-images.githubusercontent.com/14964443/100339177-d2719580-2fd9-11eb-813b-f38d8c5c72a0.png)
- [x] In title files, change "Main information" for "Poster & Banner" (if possible)
![image](https://user-images.githubusercontent.com/55785823/99797823-3efb1900-2b30-11eb-920c-70300a7991c7.png)

#### Upload box
- [x] Upload box on org docs, change label for « File title » 
![Add File - copie](https://user-images.githubusercontent.com/14964443/100338574-09937700-2fd9-11eb-84c0-81e0ccc8df61.png)
- [x] Upload box on Still Photo is the one from logo and not stills
![The Killing Of Two Lovers  Promotional Elements  Still Photo - copie](https://user-images.githubusercontent.com/14964443/100339002-976f6200-2fd9-11eb-95ba-01db670d0184.png)

- [x] Upload box titles should take the same titles as in the movie form
![elicita  Promotional Elements  Presentation Deck - copie](https://user-images.githubusercontent.com/14964443/100341560-fb475a00-2fdc-11eb-8e03-3f4c08970d6e.png)
![image](https://user-images.githubusercontent.com/14964443/100341581-0306fe80-2fdd-11eb-8e32-a832e0f9bb8d.png)

- [x] Update upload box title for Moodboard / Artistic Deck > "Upload Moodboard / Artistic Deck"

#### File table
- [x] Table for org docs & still photo: remove the sort by on « Type » & « Actions » as all docs have the same types & actions

PL : Idk if this will be easy to since we use a reusable bf-table component under the hood, imo we should try to fix it but no spend to much time on it if it become difficult

[...]

PR #4392, opened 2020-12-03
File preview in file selector of event files

RemcoSimonides:
To do in issue #4290 
- [x] If possible, add a preview button and delete button on the right of the list 
image

### DIFF
--- a/apps/festival/festival/src/app/dashboard/event/edit/edit.component.html
+++ b/apps/festival/festival/src/app/dashboard/event/edit/edit.component.html
@@ -95,9 +95,12 @@
 
           <ng-container *ngIf="!!files.value.length; else empty">
             <mat-list>
-              <mat-list-item *ngFor="let file of files.value">
+              <mat-list-item *ngFor="let file of files.value; let i = index" (click)="previewFile(file)">
                 <mat-icon mat-list-icon [svgIcon]="file | fileTypeImage: 'icon'"></mat-icon>
-                <span>{{file | filePath: ' > ' | async}}</span>
+                <div mat-line>{{file | filePath: ' > ' | async}}</div>
+                <button mat-icon-button (click)="removeSelectedFile(i, $event)">
+                  <mat-icon svgIcon="trash" color="primary"></mat-icon>
+                </button>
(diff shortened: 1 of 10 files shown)


apps/festival/festival/src/app/dashboard/event/edit/edit.component.html
```diff
@@ -95,9 +95,12 @@ <h6>My Files for this Meeting</h6>
 
           <ng-container *ngIf="!!files.value.length; else empty">
             <mat-list>
-              <mat-list-item *ngFor="let file of files.value">
+              <mat-list-item *ngFor="let file of files.value; let i = index" (click)="previewFile(file)">
```
pldespaigne, line 98: Design asked for a preview _button_, here the preview is on the whole item so it force you to use `stopPropagation` & and it's also not a really good ux imo (if I miss-click on delete a dialog will pop in my face)

libs/media/src/lib/components/file-selector/file-selector.component.html
```diff
@@ -14,7 +14,7 @@ <h2 fxFlex="0 1">Choose your files</h2>
         <ng-container *ngIf="orgFiles.length; else noOrgFiles">
           <mat-list-item *ngFor="let file of orgFiles">
             <mat-icon mat-list-icon [svgIcon]="file.path | fileTypeImage : 'icon'"></mat-icon>
-            <div mat-line [matTooltip]="file.path | fileName">{{ file.path | fileFolder }} - {{file.path | fileName}}</div>
+            <div mat-line [matTooltip]="file.path | fileName" (click)="previewFile(file.path)" class="clickable">{{ file.path | fileFolder }} - {{file.path | fileName}}</div>
```
pldespaigne, line 17: This was not requested, but if you do it, I think it's better to do it with a preview button like in the edit page.
Anyway, might be cool to ask UI/X what they think about preview in the select dialog (maybe dialog on top of dialog is not ok, idk)

pldespaigne (changes requested):

pldespaigne (approved):

Issue #4206, opened 2020-11-10
Update emails workflows to integrate financiers emails


phpgeekfr:
Update emails workflows to integrate financiers emails

- [x] update / create templates
- [x] set the correct app name in mails
- [x] set the correct "from" email
 - #2826
- [ ] update images and texts

phpgeekfr:
List of emails along with templates ( second tab but needs to be updated) :
https://docs.google.com/spreadsheets/d/1b1K7YfnquxnfCgGGPxIz5ZNIjr9eK_bUrRJqLvPYGIk/edit?usp=sharing

Jackseed:
@phpgeekfr can we discuss that when you start working on it? That's one the last breaking change to come for MF and I want to be sure we can be ready for the MF release

Mussuranae:
I will work on it this week, we can schedule a call tomorrow if you are available.

phpgeekfr:
Maybe this can be linked to #4263

Mussuranae:
So I created an objet to pass app config in every template of Sendgrid, so it will be easier to get the app name or app logo for example.
But one problem persists, the logo has to be in PNG type, because SVG isn't supported by Gmail. I tried with Outlook and I saw the logo, but I guess a lot of people use Gmail so it will be a problem.

Jackseed:
@Mussuranae there is also the question of the demo request 
- [ ] Add the app name on the demo request emails (sendDemoRequestMail) 
from ticket #4191
Can you have a look and tell me if it can be app specific there plz?

Jackseed:
@Mussuranae I see in the repo that we still have this: 
```
    // Template for sending the verify email
    verifyEmail: 'd-81438bdf511b43cfa866ca63a45a02ae',
    resetPassword: 'd-6a0710945bc841ffb6955e3dc202704c',
```
In fact I think it's not used anymore and we use only this instead:
```
    // Template for welcome message when user created his account by himself
    welcomeMessage: 'd-fc05a8cf5b1548ebae9ca44247a6c256',
```
Could you please confirm and delete that old code so I erase the sendgrid template?

phpgeekfr:
I think this is related to https://github.com/blockframes/blockframes/issues/2875

Jackseed:
Hum not quite sure to see the link between the 2 🤔 
[...]

PR #4394, opened 2020-12-03
App config for emails

Mussuranae:
Create an object app settings parameters for email and pass it to the sendMailFromTemplate function directly.

This will allow admins to use some variables concerning the app (like name or logo), in every Sendgrid template they want. The template should be much more dynamic and flexible.
For developers, we will have these variables only at one place and they will be available for every Sendgrid template. 

Part of issue #4206


apps/backend-functions/src/templates/mail.ts
```diff
@@ -138,15 +138,13 @@ export function userRequestedToJoinYourOrg(request: RequestToJoinOrganization, u
...
   recipient: EmailRecipient,
   orgDenomination: string,
-  appLabel: string,
   eventId: string,
   link: string,
   url: string = appUrl.market
 ): EmailTemplateRequest {
   const data = {
     userFirstName: recipient.name,
     orgName: orgDenomination,
```
phpgeekfr, line 147: are you sure that `templateIds.invitation.attendEvent.created` doest not use appName ?

apps/backend-functions/src/templates/mail.ts
```diff
@@ -167,7 +164,6 @@ export function requestToAttendEventFromUser(
     adminFirstName: recipient.name,
     userFirstName: fromUserFirstname,
     orgName: fromUserOrgName,
-    appName: appLabel,
```
phpgeekfr, line 170: same

apps/backend-functions/src/templates/mail.ts
```diff
@@ -95,10 +95,10 @@ export function userJoinOrgPendingRequest(email: string, orgName: string, userFi
   return { to: email, templateId: templateIds.request.joinOrganization.pending, data };
 }
 
-export function organizationAppAccessChanged(admin: PublicUser, appLabel: string, url: string): EmailTemplateRequest {
+/** Email to let org admin knows that his/her organization has access to a new app */
+export function organizationAppAccessChanged(admin: PublicUser, url: string): EmailTemplateRequest {
   const data = {
     adminFirstName: admin.firstName,
-    appName: appLabel,
```
phpgeekfr, line 101: save as above

libs/utils/src/lib/apps.ts
```diff
@@ -28,6 +35,14 @@ export const sendgridEmailsFrom: Record<App | 'default', EmailJSON> = {
   default: { email: 'team@cascade8.com', name: 'Cascade 8' }
 } as const;
 
+export const appLogo = {
+  catalog: `${appUrl.content}/assets/logo/dark/archipel_content_logo.svg`,
+  festival: `${appUrl.market}/assets/logo/dark/logo_archipel_market_outline.svg`,
+  financiers: `${appUrl.financiers}/assets/logo/dark/logo_media_financiers.svg`,
```
phpgeekfr, line 41: Why SVG if they are not working ? can you create a png from the svg ?
Also you are using the `dark` icons.

apps/backend-functions/src/templates/mail.ts
Mussuranae: So actually, it's used on the template on sendgrid, but on develop, no one uses Sendgrid. We can't change the template now because it will break the email for the production.
But with the new object, Vincent or Marie will just needs to change the `appName` in the template by `app.name`.
Do you think it's too early and we have to do it just before the release ?
Same thing for the 2 others comments.

libs/utils/src/lib/apps.ts
Mussuranae: I need to see with Vincent what logo he wants exactly, and ask Stéphanie to do other logo if needed.

apps/backend-functions/src/templates/mail.ts
phpgeekfr: we should let it as it is for now. Once in production we will be able to change templates to use app.name and then we will be able to remove this.
Add a comment with an issue number to remember that this will be removable once templates are updated

phpgeekfr (approved):

Issue #4393, opened 2020-12-03
Meeting Bug Tracker


pldespaigne:
General Bug Tracking issue for the whole Meeting features

- [x] clicking on the "quit meeting" button, then clicking on "cancel" still kill Twilio (even if the user stays on the session page)
- [x] pdf view for private documents doesn't work for buyers
- [x] when seller quit the meeting, buyers are redirected to lobby instead of the "meeting is over" page
- [ ] After disabling camera and muting microphone in the lobby page, I navigate to the session page and the camera and microphone are active again

_link to master issue #3639_

pldespaigne:
Closing in favor of various per-release Sanity-Team issues

PR #4399, opened 2020-12-04
bug fixes

pldespaigne:
related to Meeting Bug Tracker #4393
- [x] clicking on the "quit meeting" button, then clicking on "cancel" still kill Twilio (even if the user stays on the session page)
- [x] pdf view for private documents doesn't work for buyers

### DIFF
--- a/apps/backend-functions/src/media.ts
+++ b/apps/backend-functions/src/media.ts
@@ -13,6 +13,8 @@ import { MovieDocument } from './data/types';
 import { uploadToJWPlayer } from './player';
 import { HostedVideo } from '@blockframes/movie/+state/movie.firestore';
 import { storage } from 'firebase-functions';
+import { Meeting } from '@blockframes/event/+state/event.firestore';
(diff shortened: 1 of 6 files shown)


apps/backend-functions/src/media.ts
```diff
@@ -143,21 +145,63 @@ async function isAllowedToAccessMedia(ref: string, uid: string): Promise<boolean
...
+
+      // if event is a Meeting and has the file
+      if (eventData.type === 'meeting') {
+
+        // event.meta.files store the raw refs, but the ref in the function params
+        // has been transformed by the front, so we must apply the same transformations
+        // if we want to be able to match ref & files
+        const match = (eventData.meta as Meeting).files.some(file => {
+
+          if (file.startsWith('/')) file = file.substr(1); // remove leading '/' if exists
```
fritzschoff, line 187: seems like this could be a useful helper functions.

fritzschoff (approved):

Issue #4337, opened 2020-11-26
Style issues 


Jackseed:
This ticket will be updated and should be done after the new design system to come (#4176).

### Cropper
- [x] Add 16px margin all around the blue dashed zone (same for all states "uploading" "cropping" "added"...)
Add same margin around tables in the same area. (if space is not enought, have a horizontal scroll inside tables like it's the case in some existing components (see tables inside dashboard)

<img width="783" alt="Capture d’écran 2020-11-23 à 12 14 46" src="https://user-images.githubusercontent.com/66258851/99955967-7f070980-2d85-11eb-8e2b-f893676ec2c2.png">

### Upload statement (movie form)
- [x] Title: put h2 instead of h1. Text centered in the dialog. 
- [x] 24px under the title's divider
- [x] Take off the paper plane in the Update Button.
- [x] In this case, "cancel" button should also be primary since it's already differentiated with "update" by the button shape.
If you can (not mandatory at all) use a lighter shade of primary palette.

<img width="842" alt="Capture d’écran 2020-11-23 à 12 21 11" src="https://user-images.githubusercontent.com/66258851/99956492-65b28d00-2d86-11eb-9f13-6faca4dcf84b.png">

- [x] Action List icons on the right should be stuck on the right and aligned with same icons under and above in table lines. 
- [x] "Document name" and any column's width shouldn't be based on it's content's size and always be the same in every line.

<img width="842" alt="Capture d’écran 2020-11-23 à 12 21 11" src="https://user-images.githubusercontent.com/66258851/99957003-4831f300-2d87-11eb-86aa-b9640816a5ee.png">

### Navbar
- [x] If you can, center the icon inside the light effect. If impossible take it off.

<img width="288" alt="Capture d’écran 2020-11-23 à 12 31 28" src="https://user-images.githubusercontent.com/66258851/99957339-d6a67480-2d87-11eb-9b8a-1f4b78e3f46c.png">

### My Files
- [x] Icons "accept / refuse" have to be the same size (smaller one would be best)
<img width="776" alt="Capture d’écran 2020-11-23 à 12 27 18" src="https://user-images.githubusercontent.com/66258851/99957444-005f9b80-2d88-11eb-8e5e-4915350afc19.png">

### File manager
- [x] Same Padding as other usual buttons for right and left (24px)

[...]

PR #4405, opened 2020-12-07
Polishing file explorer and icon header

RemcoSimonides:
To do's issue #4337 
### Cropper
- [x] Add 16px margin all around the blue dashed zone (same for all states "uploading" "cropping" "added"...)
Add same margin around tables in the same area. (if space is not enought, have a horizontal scroll inside tables like it's the case in some existing components (see tables inside dashboard)

### Upload statement (movie form)
- [x] Title: put h2 instead of h1. Text centered in the dialog. 
- [x] 24px under the title's divider
- [x] Take off the paper plane in the Update Button.
- [x] In this case, "cancel" button should also be primary since it's already differentiated with "update" by the button shape.
If you can (not mandatory at all) use a lighter shade of primary palette.

- [x] Action List icons on the right should be stuck on the right and aligned with same icons under and above in table lines. 
- [x] "Document name" and any column's width shouldn't be based on it's content's size and always be the same in every line.

### Navbar
- [x] If you can, center the icon inside the light effect. If impossible take it off.

### My Files
- [x] Icons "accept / refuse" have to be the same size (smaller one would be best)

### DIFF
--- a/libs/ui/src/lib/layout/marketplace/marketplace.component.scss
+++ b/libs/ui/src/lib/layout/marketplace/marketplace.component.scss
@@ -21,6 +21,10 @@ mat-sidenav-content {
       margin-right: 24px;
     }
 
+    button mat-icon {
+      margin-right: 0px;
+    }
+
     a:first-of-type {
       margin-right: auto;
     }
--- a/apps/festival/festival/src/app/marketplace/marketplace.component.scss
+++ b/apps/festival/festival/src/app/marketplace/marketplace.component.scss
@@ -3,3 +3,7 @@
   display: block;
   height: 100%;
 }
+
+header {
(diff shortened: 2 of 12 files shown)


libs/ui/src/lib/layout/marketplace/marketplace.component.scss
```diff
@@ -21,6 +21,10 @@ mat-sidenav-content {
       margin-right: 24px;
     }
 
+    button mat-icon {
```
RemcoSimonides, line 24: Overwrites the margin-right from the previously added margin
```
* {
  margin-right: 24px;
}
```

pldespaigne (approved):

PR #4408, opened 2020-12-07
bug fix and style improvement

pldespaigne:
related to Meeting Bug Tracker #4393
- [x]  when seller quit the meeting, buyers are redirected to lobby instead of the "meeting is over" page

### DIFF
--- a/apps/festival/festival/src/app/marketplace/event/session/session.component.scss
+++ b/apps/festival/festival/src/app/marketplace/event/session/session.component.scss
@@ -2,7 +2,6 @@
 
 :host {
   height: 100%;
-  display: flex;
   justify-content: center;
   padding: 24px;
 }
@@ -75,6 +74,10 @@ p {
     meeting-video-room {
       width: 60%;
     }
+  }
+
+  img {
+    width: 100%;
   }
 
   media-control {
--- a/apps/festival/festival/src/app/marketplace/event/session/session.component.scss
+++ b/apps/festival/festival/src/app/marketplace/event/session/session.component.scss
@@ -2,7 +2,6 @@
 
 :host {
   height: 100%;
-  justify-content: center;
   padding: 24px;
 }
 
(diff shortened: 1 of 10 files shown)


apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -2,7 +2,6 @@
 
 :host {
   height: 100%;
-  display: flex;
   justify-content: center;
```
fritzschoff, line 5: then remove this as well?

libs/media/src/lib/components/viewers/media-viewer.component.html
```diff
@@ -1,30 +1,28 @@
...
-        </ng-container>
-        <ng-container *ngSwitchDefault>
-          <article class="noFile">
-            <h3>Sorry</h3>
-            <img asset="no_titles.webp" alt="Flying saucer">
-            <p>We don't know how to display this file.</p>
-            <p class="mat-caption">Supported files are: pdfs, images, and videos</p>
-          </article>
-        </ng-container>
+<section fxLayout="column" fxLayoutAlign="space-between" [class]="event.isOwner ? 'owner' : ''">
```
fritzschoff, line 2: you may want here ngClass here. Then you don't need to set it to an empty string?

libs/media/src/lib/components/viewers/pdf-viewer/pdf-viewer.component.html
```diff
@@ -1,18 +1,22 @@
-<div fxLayout="column" #container>
+
+<!-- We need to have div container in order to have the "fullscreen" feature -->
+<div #container>
```
fritzschoff, line 3: why don't you do in the scss file 
``` 
:host {display:block}
```

fritzschoff: and for the container ref, you can't you do a hostbinding in the component

fritzschoff (approved):

PR #4410, opened 2020-12-08
Create cms lib in admin to share types

GrandSchtroumpf:


libs/admin/src/lib/cms/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+  type: 'basic' | 'flat' | 'stroked';
+  color: 'primary' | 'accent' | 'warn' | '';
+}
+
+export type QueryMethods = "where" | "orderBy" | "limit" | "limitToLast";
+export interface CollectionQuery {
+  method: QueryMethods;
+}
+
+export type Conditions = "<" | "<=" | "==" | ">=" | ">" | "array-contains" | "in" | "array-contains-any" | '!=';
```
fritzschoff, line 13: should be singular?

fritzschoff: since it is not an array

fritzschoff (approved):

PR #4452, opened 2020-12-09
Issue mf debug

fritzschoff:
#ref 4406


libs/media/src/lib/directives/image-reference/image-reference.directive.ts
```diff
@@ -98,36 +98,34 @@ export class ImageReferenceDirective implements OnInit, OnDestroy {
       map(([local, global]) => local || global)
     );
 
-    // Adds a 5 (default) sec delay on image loading except for the first one.
-    // This is to prevent imgIx to load the image before it is even uploaded or moved to good directory.
-    // This variable will be updated on the fly inside the `subscribe` to update next observable's behaviour.
+    const obs$: Observable<any>[] = [this.asset$, this.parameters, theme$]
+
     let _delay = 0;
+    if (this.ref$.value) {
```
phpgeekfr, line 104: if this.ref$ start with no value, we never get to enter this: the observable is not updating from "empty" to "image"

phpgeekfr: This is not really a problem but app is less dynamic

phpgeekfr (approved):
please do not push directly to `release/2.0`  use an intermediary branch to push to `release/2.0` only once a day

Issue #4442, opened 2020-12-08
Account settings (org + profile + cookies)


Jackseed:
Tester: @SamirTouira  
Bug fix: @RemcoSimonides

SamirTouira:
Clélia has encountered this issue about cookies with Intercom when she loggin with an account in a clean navigator and seems keep especially the old cookies : 

![issueCookies](https://user-images.githubusercontent.com/52285231/101521155-b4b31180-3985-11eb-9237-92aa8d52721e.png)

I have re-tested on new navigators like microsoft Edge to find this issue, but it seems correct in my side because I haven't Intercom displayed when there are 0 cookies, even after delete or keep cookies in a new navigator.

 I think we just have to keep an eye to see if it will start again

Mussuranae:
The way I encounter this issue : 
- open new chromium window and delete cookie to get clean browser
- login with an buyer account
- accept cookie after login
- close the browser window
- open a new one and go again on the app
At this moment your are already login with the last account and the cookie are still in local storage so intercom appears, but the cookie banner appears too.

SamirTouira:
The button "Update" is pretty stick at the bottom of the page in the organization settings. We just have to add some padding in the footer to resolve it : 

![ButtonUpdate](https://user-images.githubusercontent.com/52285231/101622214-0e175100-3a17-11eb-9df8-3e75950e881f.png)

If you compare the footer in which the "Update" button is located in Profile Settings, the padding is there. In the footer of Organization Settings, it isn't there

Route : /c/o/organization/ID_USER/view/org

RemcoSimonides:
@Mussuranae Does the issue also appear on staging? What browser are you using?

Mussuranae:
Yes on staging and I'm on Linux/Chromium

Jackseed:
- [ ] Upload icon on profile avatar is a bit too large on Safari (cf. screen left Brave, right Safari)
![image](https://user-images.githubusercontent.com/14964443/101778432-23ae7880-3af4-11eb-9237-07fe20886f52.png)

Jackseed:
- [ ] On cookie component, add a title & subtitle to both sections for better understanding:
1. Above Cascade8 (first section):
Title: Essential cookies
Subtitle: These are essential to the operation of the platform's services.
2. Above Third party partners
Title: Optional cookies
Subtitle: These are optional but very useful for us to provide you with optimal services.
[...]

PR #4470, opened 2020-12-10
Persistant cookie

RemcoSimonides:
Fix issue #4442 
Cookies without an expiration date (or max-age) are set to last only for a session. Closing the browser ended the session and thus removing this cookie. By setting the max-age, the cookie is only deleted after one year.

- [x] The button "Update" is pretty stick at the bottom of the page in the organization settings. We just have to add some padding in the footer to resolve it : 
- [x] there are the same issue for the cookies preference page

- [x] issueCookies


SamirTouira (approved):
All seems good

Issue #4417, opened 2020-12-08
Notifications


Jackseed:
Tester: @Jackseed 
Bug fix: @SamirTouira

Mussuranae:
The `more` button redirect to the wrong page when the notification is about a project accepted on the platform.
![image](https://user-images.githubusercontent.com/48033264/101501693-592a5900-3970-11eb-8872-cbd14a38d052.png)
![image](https://user-images.githubusercontent.com/48033264/101501667-4dd72d80-3970-11eb-8e97-dc65745471c2.png)

Jackseed:
> The `more` button redirect to the wrong page when the notification is about a project accepted on the platform.
> ![image](https://user-images.githubusercontent.com/48033264/101501693-592a5900-3970-11eb-8872-cbd14a38d052.png)
> ![image](https://user-images.githubusercontent.com/48033264/101501667-4dd72d80-3970-11eb-8e97-dc65745471c2.png)

Didn't test it yet but correct behaviour would be to redirect to **marketplace** movie view for this film

2amarkina:
- [ ] Need to find other text to explain to a user that the request was sent. I would propose just "Sent" If not - a message is a way too long 
![Uploading Снимок экрана 2020-12-09 в 15.09.51.png…]()

Jackseed:
@Mussuranae for screenings we hide the request access buttons when you are organizer of the event, could we do the same for meetings ?

Mussuranae:
We can but it's on the screening list, do we have somewhere where the meetings appear like the screening and we can ask access to meetings ?

Jackseed:
Right, I tried to reproduce this and the only way you have to request access as the organizer is on the meeting view page on the marketplace
![image](https://user-images.githubusercontent.com/14964443/101659899-d7f1c580-3a46-11eb-88c2-2d39578bba14.png)
![image](https://user-images.githubusercontent.com/14964443/101659919-dde7a680-3a46-11eb-8bb4-7f329c570e00.png)
![image](https://user-images.githubusercontent.com/14964443/101659962-e809a500-3a46-11eb-8ade-f648dfa3c7a7.png)

Jackseed:
- [ ] Existing ticket to take care of #4340

Jackseed:
- [ ] I'll investigate more but it looks like notifications store is not reset after a logout. I just create a new organization and got 15 notif waiting that are not linked to my fresh new account
[...]

PR #4472, opened 2020-12-10
invitation debug when the org is owner

SamirTouira:
Fix : #4417

### DIFF
--- a/libs/event/src/lib/components/screening-item/screening-item.component.ts
+++ b/libs/event/src/lib/components/screening-item/screening-item.component.ts
@@ -1,9 +1,10 @@
 import { Component, ChangeDetectionStrategy, Input } from '@angular/core';
 import { InvitationQuery, Invitation } from '@blockframes/invitation/+state';
 import { ScreeningEvent } from '../../+state';
-import { Observable } from 'rxjs';
+import { Observable, BehaviorSubject } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { Movie } from '@blockframes/movie/+state';
+import { Event } from '../../+state/event.model';
 
 @Component({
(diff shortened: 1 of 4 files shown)


libs/event/src/lib/components/screening-item/screening-item.component.ts
```diff
@@ -16,15 +17,21 @@ export class ScreeningItemComponent {
   public movie: Movie;
   public screening: ScreeningEvent;
   public invitation$: Observable<Invitation>;
+  private _event = new BehaviorSubject<Event>(null);
```
GrandSchtroumpf, line 20: Set the generic type to `ScreeningEvent` directly. It'll avoid to have to cast the type after: 
```typescript
private _event = new BehaviorSubject<ScreeningEvent>(null);
```

libs/event/src/lib/components/screening-item/screening-item.component.ts
```diff
@@ -16,15 +17,21 @@ export class ScreeningItemComponent {
...
+    this._event.next(screening);
     this.screening = screening;
     this.movie = screening.movie;
     this.poster = screening.movie?.poster;
     this.invitation$ = this.invitationQuery.whereCurrentUserIsGuest().pipe(
       map(invits => invits.find(e => e.docId === screening.id))
     );
   }
+  get event() {
+    return this._event.getValue() as ScreeningEvent;
```
GrandSchtroumpf, line 33: With the generic type above you won't need to cast the type : 
```typescript
return this._event.getValue();
```

GrandSchtroumpf (changes requested):

GrandSchtroumpf (approved):

PR #4475, opened 2020-12-10
Pl fixes 1

pldespaigne:


apps/festival/festival/src/app/marketplace/event/session/session.component.ts
```diff
@@ -70,11 +72,13 @@ export class SessionComponent implements OnInit, OnDestroy {
       }
 
       if (event.type === 'screening') {
+        this.dynTitle.setPageTitle(event.title, 'Screening');
         if (!!(event.meta as Screening).titleId) {
           const movie = await this.movieService.getValue(event.meta.titleId as string);
           this.screeningFileRef = movie.promotional.videos?.screener?.ref ?? '';
         }
       } else if (event.type === 'meeting') {
+        this.dynTitle.setPageTitle(event.title, 'Meeting');
```
fritzschoff, line 81: do it after the if else statement if you are using event.title in both if else statements

fritzschoff: so you need to do it only once

libs/ui/src/lib/form/table/form-table.component.scss
```diff
@@ -35,21 +33,6 @@ mat-paginator {
...
-  height: 10px;
-}
-
-::-webkit-scrollbar-track {
-  height: 10px;
-  background-color: var(--active);
-  opacity: 0.5;
-}
-
-::-webkit-scrollbar-thumb {
```
fritzschoff, line 48: so you don't style it anymore?

fritzschoff (approved):

libs/ui/src/lib/form/table/form-table.component.scss
pldespaigne: yup

PR #4477, opened 2020-12-10
Issue 4440

### LINKED ISSUE #4440: MF Dashboard - Movie form + campaign form
Tester: Design + Business teams
Bug fix: @GrandSchtroumpf @fritzschoff

fritzschoff:
fix #4440

### DIFF
--- a/libs/utils/src/lib/form/forms/list.form.ts
+++ b/libs/utils/src/lib/form/forms/list.form.ts
@@ -98,7 +98,7 @@ export class FormList<T, Control extends AbstractControl = any> extends FormArra
   }
 
   /**
-   * Add, remove and update controles to match the value
+   * Add, remove and update controls to match the value
    * @note this method is specific from FormList and is not part and Angular FormArray interface
    */
   patchAllValue(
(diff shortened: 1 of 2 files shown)


libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -109,33 +109,17 @@ export class FormList<T, Control extends AbstractControl = any> extends FormArra
...
     value.forEach((newValue, index) => {
-      // If there is a form already patch it
-      if (this.at(index)) {
-        if (this.at(index)['patchAllValue']) {
-          this.at(index)['patchAllValue'](newValue, {
-            onlySelf: true,
-            emitEvent: options.emitEvent
-          });
-        } else {
-          this.at(index).patchValue(newValue, {
```
fritzschoff, line 120: this else statement was never reached! Tried with every case.

libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -109,33 +109,17 @@ export class FormList<T, Control extends AbstractControl = any> extends FormArra
     } = {}
   ) {
     value.forEach((newValue, index) => {
-      // If there is a form already patch it
-      if (this.at(index)) {
-        if (this.at(index)['patchAllValue']) {
```
fritzschoff, line 114: this statement is always true since patchAllValue because were are in the in a child form that inherirts from the form array.
AbstractControl => Form Array => PerksForm. 
So if you check PerksForm.patchAllValue, it is correct since it inherits it from its parent. Maybe we can check for hasOwnProperty.

libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -109,33 +109,17 @@ export class FormList<T, Control extends AbstractControl = any> extends FormArra
...
-          this.at(index)['patchAllValue'](newValue, {
```
fritzschoff, line 115: If you would call this patchAllValue with an object, this won't do anything since forEach doesn't exist on the object type. See line 111.

libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -109,33 +109,17 @@ export class FormList<T, Control extends AbstractControl = any> extends FormArra
...
-        } else {
-          this.at(index).patchValue(newValue, {
-            onlySelf: true,
-            emitEvent: options.emitEvent
-          });
-        }
-        // else create one
-      } else {
-        this.setControl(index, this.createControl(newValue));
+      this.setControl(index, this.createControl(newValue));
```
GrandSchtroumpf, line 112: I'm ok with this solution **only** if there is no FormList without `.factory`

GrandSchtroumpf (approved):

Issue #3697, opened 2020-09-15
Get rid of JSONL


GrandSchtroumpf:
## JSONL migration list

- [x] rewrite anonymization script to work via emulator (or any DB object passed to it)
- [x] use passthrough command execution for streaming output
- [ ] update backup/restore commands in backend-ops to utilize CLI by running commands locally
- [x] remove backup/restore function from firebase functions
- [ ] write code that automatically stores binary backup in bucket from local emulator
- [x] write code that can pull binary backup from bucket and convert to emulator backup format
- [ ] update CI config to run more prepareForTesting related steps in weekly cron
- [x] update CI to create weekly binary backups instead of jsonl
- [ ] update CI to anonymize in CI: launch emulator, download binary backup, anonymise through emulator, shut down emulator, upload saved state as binary backup

phpgeekfr:
@george43g  does it imply to take care of  "get rid of "admin" backend function that use Express to backup/restore#3973 " before doing this?

PR #4485, opened 2020-12-10
get rid of JSONL - big PR - binary anonymization and full import/export functionality

george43g:
- re Get rid of JSONL #3697
- download and load backups
- run anonymization
- upload backups
- convert between emulator backup and firestore backup (both directions)

- working PFT process
- maintenance mode issue (partially) solved


apps/backend-ops/src/emulator.ts
```diff
@@ -1,11 +1,94 @@
...
+import admin from 'firebase-admin'
+import { backupBucket } from '@env'
 
+/**
+ * This function will download the Firestore backup from specified bucket, import it into
+ * the emulator and run the emulator without shutting it down. This command can be used to run
+ * emulator in background while developing or running other processes.
+ * @param bucketUrl GCS bucket URL for the Firestore backup
+ */
+export async function importEmulatorFromBucket(_bucketUrl: string) {
```
phpgeekfr, line 25: maybe _backup instead of _bucket for `_bucketUrl` ?

apps/backend-ops/src/emulator.ts
```diff
@@ -1,11 +1,94 @@
...
+    proc = await startFirestoreEmulatorWithImport(defaultEmulatorBackupPath);
+    await new Promise(() => {});
+  } catch (e) {
+    await shutdownEmulator(proc);
+    throw e;
+  }
+}
+
+export interface StartEmulatorOptions {
+  importFrom: 'defaultImport' | string
```
phpgeekfr, line 41: is `'defaultImport' | string` usefull ? use only a string or specific values but not both

apps/backend-ops/src/emulator.ts
```diff
@@ -1,11 +1,94 @@
...
+ * This function will download the Firestore backup from specified bucket, import it into
+ * the emulator and run the emulator without shutting it down. This command can be used to run
+ * emulator in background while developing or running other processes.
+ * @param bucketUrl GCS bucket URL for the Firestore backup
+ */
+export async function importEmulatorFromBucket(_bucketUrl: string) {
+  const bucketUrl = _bucketUrl
+    ? _bucketUrl
+    : await getLatestFolderURL(loadAdminServices().storage.bucket(backupBucket));
+  await importPrepareFirestoreEmulatorBackup(bucketUrl, defaultEmulatorBackupPath);
```
phpgeekfr, line 29: is this method `importPrepareFirestoreEmulatorBackup` named well ?

apps/backend-ops/src/users.ts
```diff
@@ -82,7 +82,7 @@ export async function removeUnexpectedUsers(expectedUsers: UserConfig[], auth: A
   return;
 }
 
-function readUsersFromJsonlFixture(db: JsonlDbRecord[]): UserConfig[] {
+function readUsersFromJsonlFixture(db: DbRecord[]): UserConfig[] {
```
phpgeekfr, line 85: is it possible to still be able to use old way (jsonl) until this new one is ready to use and tested?

libs/firebase-utils/src/lib/commands.ts
```diff
@@ -43,7 +43,7 @@ export function runShellCommandUntil(cmd: string, until: string) {
   return { proc, procPromise }
 }
 
-export function waitForProcOutput(proc: ChildProcess, output: string) {
+export function awaitProcOutput(proc: ChildProcess, output: string) {
```
phpgeekfr, line 46: can you add a comment explaining this method ?

apps/backend-ops/src/emulator.ts
```diff
@@ -1,11 +1,94 @@
...
 }
 
+export async function anonymizeLatestProdDb() {
+  if (!('FIREBASE_PRODUCTION_SERVICE_ACCOUNT' in process.env)) {
+    throw new Error('Key "FIREBASE_PRODUCTION_SERVICE_ACCOUNT" does not exist in .env');
+  }
+  const cert = getServiceAccountObj(process.env.FIREBASE_PRODUCTION_SERVICE_ACCOUNT);
+
+  const prodApp = admin.initializeApp(
+    { storageBucket: prodBackupBucket,
```
GrandSchtroumpf, line 68: format this to has the `{` in the previous line.

libs/firebase-utils/src/lib/anonymize/lib.ts
```diff
@@ -0,0 +1,195 @@
...
+  };
+}
+
+function updateUser(user: User | PublicUser) {
+  if (!user) return;
+  if (hasKeys<PublicUser>(user, 'uid') && !hasKeys<User>(user, 'watermark')) { // Is public
+    const newUser =  userCache?.[user.uid] || (userCache[user.uid] = processUser(user))
+    return createPublicUser(newUser);
+    // If not in cache, process the user, write to cache, make it a publicUser and return it
+  } else if (hasKeys<User>(user, 'uid')) {
```
GrandSchtroumpf, line 65: Because you're returning in the previous if statement, you don't need the else, I think it's a little bit easier to read: 
```typescript
if (!user) return
if (!hasKeys<PublicUser>(user, 'uid')) {
  throw Error(`Unable to process user: ${JSON.stringify(user, null, 4)}`);
}
// is public
if (!hasKeys<User>(user, 'watermark')) {
  ...
}
// Not cached yet
if (!userCache?.[user.uid]) {
  userCache[user.uid] = processUser(user)
}
return userCache[user.uid];
```

libs/firebase-utils/src/lib/anonymize/lib.ts
```diff
@@ -0,0 +1,195 @@
...
+
+function getPathOrder(path: string): number {
+  if (path.includes('users/')) return 1;
+  if (path.includes('orgs/')) return 2;
+  if (path.includes('invitations/')) return 3;
+  if (path.includes('notifications/')) return 4;
+  return 5;
+}
+
+export async function runAnonymization(db: FirestoreEmulator) {
```
GrandSchtroumpf, line 142: Could we do something like:
```typescript
const updates = {
  users: processUser,
  orgs: processOrg,
  ...
}

export async function runAnonymization(db: FirestoreEmulator) {
  for (const [path, updateDoc] of Object.entries(updates)) {
    const collection = db.collection(path);
    const docs = await collection.listDocuments();
    // Could be improve with chunks
    for (const doc of docs) {
      const data = doc.data();
      doc.update(updateDoc(data));
    }
  }
}
```

libs/firebase-utils/src/lib/anonymize/lib.ts
```diff
@@ -0,0 +1,195 @@
...
+  throwOnProduction();
+  const dbArray = await loadDb(db);
+  const orderedDbArray = dbArray.sort((a, b) => getPathOrder(a.docPath) - getPathOrder(b.docPath))
+  await db.clearFirestoreData({ projectId: firebase.projectId });
+  const anonDb = orderedDbArray.map(anonymizeDocument)
+  await runChunks(anonDb, async ({content, docPath}) => {await db.doc(docPath).set(content)}, 1000)
+  console.log('Anonymization Done!')
+}
+
+export async function loadDb(db: FirebaseFirestore.Firestore) {
```
GrandSchtroumpf, line 152: I think it's overkill. Check comment above.

apps/backend-ops/src/users.ts
george43g: yes of course - this does not change the underlying structure, just changed the name of the interface not to confuse anyone, since we use this interface in the new anonymization process as well (at least for now)

apps/backend-ops/src/emulator.ts
george43g: this needs to be a full URL to a gcs bucket... that holds a backup... hmm... either could work? It's a URL to a folder in a bucket not just the bucket itself, so you're right. Something like bucketBackupURL, 

But KFH wants to not run via shell commands, so gonna have to rename the functions as well

apps/backend-ops/src/emulator.ts
```diff
@@ -1,11 +1,94 @@
...
+import { backupBucket } from '@env'
 
+/**
+ * This function will download the Firestore backup from specified bucket, import it into
+ * the emulator and run the emulator without shutting it down. This command can be used to run
+ * emulator in background while developing or running other processes.
+ * @param bucketUrl GCS bucket URL for the Firestore backup
+ */
+export async function importEmulatorFromBucket(_bucketUrl: string) {
+  const bucketUrl = _bucketUrl
```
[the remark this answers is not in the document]
george43g: no idea why I didn't see this haha

apps/backend-ops/src/emulator.ts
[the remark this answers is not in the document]
george43g: that's because the function was doing multipe things, but now we're going to change it anyway

apps/backend-ops/src/emulator.ts
```diff
@@ -1,11 +1,94 @@
...
+    ? _bucketUrl
+    : await getLatestFolderURL(loadAdminServices().storage.bucket(backupBucket));
+  await importPrepareFirestoreEmulatorBackup(bucketUrl, defaultEmulatorBackupPath);
+  let proc: ChildProcess;
```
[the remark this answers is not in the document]
george43g: using it in the catch block to enable a graceful shutdown

apps/backend-ops/src/emulator.ts
```diff
@@ -1,11 +1,94 @@
...
+  try {
+    proc = await startFirestoreEmulatorWithImport(defaultEmulatorBackupPath);
+    await new Promise(() => {});
```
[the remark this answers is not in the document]
george43g: I'll clean these up and leave them in just in for you to take another look. I'll still convert the anon method and everything as we discussed

apps/backend-ops/src/emulator.ts
george43g: this implies that it's either that particular string, or any path - this is more for self-documenting code reasons. I'll make it more clear and let me know what you think

apps/backend-ops/src/emulator.ts
```diff
@@ -1,11 +1,94 @@
...
-  await importPrepareFirestoreEmulatorBackup(bucketUrl, emulatorBackupPath)
-  const proc = await startFirestoreEmulatorWithImport(emulatorBackupPath);
-  await gracefullyKillEmulator(proc)
+/**
+ * This function will load the Firestore emulator from a default local path, or use the
+ * path provided as an argument without downloading anything from a GCS bucket.
+ * @param param0 this is a relative path to local Firestore backup to import into emulator
+ */
+export async function loadEmulator({ importFrom = 'defaultImport' }: StartEmulatorOptions) {
+  const emulatorPath = importFrom === 'defaultImport' ? defaultEmulatorBackupPath : resolve(process.cwd(), importFrom);
```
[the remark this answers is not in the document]
george43g: I had no idea it was an issue on windows. in this case, ok, but sometimes we need to `resolve` for an absolute path

apps/backend-ops/src/main.ts
```diff
@@ -13,10 +13,11 @@ import { generateFixtures } from './generate-fixtures';
 import { backup } from './admin';
 import { selectEnvironment } from './select-environment';
 import { healthCheck } from './health-check';
-import { importEmulatorFromBucket } from './emulator';
+import { anonymizeLatestProdDb, importEmulatorFromBucket, loadEmulator, uploadBackup } from './emulator';
 
 const args = process.argv.slice(2);
 const [cmd, ...flags] = args;
+const [arg1, arg2, arg3] = flags;
```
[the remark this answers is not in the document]
george43g: this whole main file is a hack. I made an issue to use inquirer that handles args and commands and --help for us. will look into this. the named variables are going to add complexity unless we use a lib (unless you have a simple way of solving this)

libs/firebase-utils/src/lib/anonymize/lib.ts
```diff
@@ -0,0 +1,195 @@
...
+import { PublicOrganization } from '@blockframes/organization/+state/organization.firestore';
+import { FirestoreEmulator } from '../firestore';
+import { firebase } from '@env'
+import { runChunks } from '../firebase-utils';
+
+const userCache: { [uid: string]: User | PublicUser } = {};
+const orgCache: { [id: string]: Organization | PublicOrganization } = {};
+
+const fakeEmail = (name: string) =>
+  `dev+${name.replace(/\W/g, '').toLowerCase()}-${Math.random()
```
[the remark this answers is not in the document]
george43g: ok mind you this has always been in our anon script, this was copied over from old code

libs/firebase-utils/src/lib/anonymize/lib.ts
george43g: I cannot understand how this would work. In future, once we start to make this script more emulator friendly (it's a port of jsonl script atm) it's more likely going to be possible to write/read in chunks, but at the moment since we must process a sorted db (for caching to work properly) we must load the entire db first, sort it, process it, then load it back in.

libs/firebase-utils/src/lib/anonymize/lib.ts
george43g: I did. Again, this is a port of our JSONL algorithm. It doesn't have to be like this forever but it's a safe way to start and get something working IMO.

apps/backend-ops/src/emulator.ts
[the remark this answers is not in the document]
george43g: okay, so there already exist two separate functions. This one was just logical for me.

apps/backend-ops/src/main.ts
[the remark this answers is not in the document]
george43g: I'm just not sure it's worth writing a flag parser. We use a few arguments now

libs/firebase-utils/src/lib/anonymize/lib.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+function hasKeys<T extends object>(doc: object, ...keys: (keyof T)[]): doc is T {
+  return keys.every((key) => key in doc);
+}
+
+function processUser<T extends User | PublicUser>(u: T): T {
+  const firstName = faker.name.firstName();
+  const lastName = faker.name.lastName();
+  const email = fakeEmail(firstName);
```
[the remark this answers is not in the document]
george43g: once DB is anonymised, syncUsers generates firebase auth from what's in the newly imported db

apps/backend-ops/src/main.ts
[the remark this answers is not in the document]
phpgeekfr: and `arg3` is not used

apps/backend-ops/src/firebaseSetup.ts
```diff
@@ -64,6 +64,50 @@ export async function prepareForTesting() {
 
 }
 
+export async function prepareForTestingBeta() {
+  console.log('Importing latest anonymized db from CI...')
+  const cmd = 'gcloud firestore import gs://ci-backups-blockframes/LATEST-ANON-DB'
```
phpgeekfr, line 69: use import { backupBucket } from 'env/env.blockframes-ci' instead of plain `ci-backups-blockframe`

phpgeekfr (approved):

PR #4543, opened 2020-12-15
More fixes

GrandSchtroumpf:


libs/notification/src/lib/+state/notification.store.ts
```diff
@@ -24,13 +25,15 @@ export class NotificationStore extends EntityStore<NotificationState, Notificati
...
+    private auth: AuthService,
     private movieQuery: MovieQuery,
     private firestore: AngularFirestore,
     private routerQuery: RouterQuery,
     private orgService: OrganizationService
   ) {
     super(initialState);
-    this.appName = appName[getCurrentApp(this.routerQuery)]
+    this.appName = appName[getCurrentApp(this.routerQuery)];
+    this.auth.signedOut.subscribe(() => this.remove());
```
RemcoSimonides, line 36: I think you forgot to unsubscribe from this subscription

RemcoSimonides (approved):

libs/notification/src/lib/+state/notification.store.ts
GrandSchtroumpf: you cannot unsubscribe, the instance live during the app life. We would have to destroy it on signout to fix this. But it's much more difficult to achieve

Issue #4424, opened 2020-12-08
Festival Marketplace - Screening list


Jackseed:
Tester: @GrandSchtroumpf  
Bug fix: @SamirTouira

2amarkina:
I was not sure if it concerns specifically Screening List or Movie Page and I don't know if it is difficult to handle right now, but  leave it here. 

When user gets his "accepted", is it possible to change CTA "Ask for invitation" by text "You already have your invitation" or smth  (not sure for the text)? 

As if on the Screening List the invitation was already accepted, there is no need for a user to ask for it one more time 

![FireShot Capture 601 - Film Page - Screening Schedule - Archipel Market - staging archipelmarket com](https://user-images.githubusercontent.com/55499036/101757788-dae8c680-3ad7-11eb-822f-b21a75bf8bf9.png)

Jackseed:
I agree with you @2amarkina, in fact we have an existing issue for that #4311 but I'm afraid it's gonna be too much for the release. 
@SamirTouira and mb @Mussuranae what's your call on this one?

Mussuranae:
I will look at it right now and tell you here, if we can do it quickly. I completely forget this ticket because we don't put it in a sprint since its creation.

maltmayer:
![Enregistrement de l’écran 2020-12-10 à 16 34 28 (1)](https://user-images.githubusercontent.com/55785823/101793711-6416f200-3b06-11eb-850b-461441e6137e.gif)
Public event approval process still weird. We shouldn't see the sentence about the org approval appear. Should directly go to the event page, without this kind of weird jump

maltmayer:
![image](https://user-images.githubusercontent.com/55785823/101794051-c2dc6b80-3b06-11eb-8233-06a67ef49a60.png)
- [ ] Change text to "Add to My Calendar"

maltmayer:
![image](https://user-images.githubusercontent.com/55785823/101794130-d38ce180-3b06-11eb-8907-e994fc37fa95.png)

- [ ] "This event has already started" should be in another color I think. Blue maybe? Smaller text also I think

- [ ] Change "Request access" CTA text to "Request Access"

- This Request access CTA has a weird alignment I think, but I'll let @stephanie-cascade8 do her review

maltmayer:
![image](https://user-images.githubusercontent.com/55785823/101794585-501fc000-3b07-11eb-9377-bc749b4725e3.png)
[...]

PR #4544, opened 2020-12-15
Typo events and poster redirection to event item

SamirTouira:
related to: #4424


libs/event/src/lib/components/screening-item/screening-item.component.html
```diff
@@ -1,4 +1,4 @@
-<a matRipple [routerLink]="['/c/o/marketplace/title', movie.id]">
+<a matRipple [routerLink]="['/c/o/marketplace/event/', screening.id]">
```
fritzschoff, line 1: you updated as well the router module as well?

fritzschoff (approved):

libs/event/src/lib/components/screening-item/screening-item.component.html
SamirTouira: no need, when I click on the poster, it leads correctly to the event object of the list, was that what you wanted to know ?

fritzschoff: yes,! alright!

PR #4570, opened 2020-12-16
added func

fritzschoff:


phpgeekfr (approved):
LGTM But it would prefere if @GrandSchtroumpf can have a look also

Issue #4514, opened 2020-12-14
Post release bug list


Jackseed:
### Meeting session
- [x] Full screen button doesn't work on Safari
![image](https://user-images.githubusercontent.com/14964443/101799825-e1456580-3b0c-11eb-8039-a02bd72ea75d.png)
- [ ] When clicking on an element that has already been shared, it should reset it. Here I clicked on the video already shared and it was still at then end of the video
PL : I'm not sure it's a good idea to do that, imo that would be bad UX, @2amarkina what is your thoughts on that ?

![image](https://user-images.githubusercontent.com/14964443/101798747-beff1800-3b0b-11eb-815f-6795e8300907.png)

### CRM
- [x] Remco: Make row in table clickable to navigate to edit page (instead of the edit icon as a column)
- [x] Remco: Disable sorting on certain columns (@remco did this in file explorer too)
- [x] Remco: For the upload of organization, if you choose `organizations` in the select after downloading your excel, it doesn't work and the spinner is turning forever
![image](https://user-images.githubusercontent.com/48033264/101754805-6ceed000-3ad4-11eb-8669-a456a29fbf01.png)
The upload works fine if we choose our type of document before uploading the excel.
- [x] There is no simulation for the user when we want to delete it. (Clélia)
![image](https://user-images.githubusercontent.com/48033264/101759087-729ae480-3ad9-11eb-8ebc-00a55daffced.png)
Functionality works fine !

### Notifications (Samir)
- [x] Change notification text triggered when movie status is turned to "accepted". Currently it's "Your project was successfully published on the marketplace."; should be instead: "[ Title name ] was successfully published on the Marketplace."

### Invitations (Samir)
- [x] Clicking on 'event name' or 'more details' leads to event marketplace page, 
but clicking on the event in calendar directly leads to the lobby of the meeting. They should all lead to the lobby page
- [x] Right align invitation status so they are aligned even if invitation text is not same size
![image](https://user-images.githubusercontent.com/14964443/101914929-3c3a9380-3bc5-11eb-8902-d855a5c6810e.png)

### Wishlist
- [ ] wish-list button can be hard to spot (@stephanie-cascade8 we need your advice here)
![image](https://user-images.githubusercontent.com/5858657/101618212-f5586c80-3a11-11eb-91a8-11cea0268026.png)

### MF Homepage (Clélia)
[...]

PR #4580, opened 2020-12-17
Disable sort on columns and make row clickable

RemcoSimonides:
To do in issue #4514 
- [x] Make row in table clickable to navigate to edit page (instead of the edit icon as a column)
- [x] Disable sorting on certain columns

### DIFF
--- a/libs/admin/src/lib/admin-panel/pages/contracts/contracts.component.ts
+++ b/libs/admin/src/lib/admin-panel/pages/contracts/contracts.component.ts
@@ -1,6 +1,6 @@
 import { Component, OnInit, ChangeDetectorRef, ChangeDetectionStrategy } from '@angular/core';
 import { getValue } from '@blockframes/utils/helpers';
-import { ActivatedRoute } from '@angular/router';
+import { ActivatedRoute, Router } from '@angular/router';
 import { ContractService } from '@blockframes/contract/contract/+state/contract.service';
 import { Contract } from '@blockframes/contract/contract/+state/contract.model';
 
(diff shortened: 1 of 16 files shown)


libs/admin/src/lib/admin-panel/pages/contracts/contracts.component.ts
```diff
@@ -30,40 +29,32 @@ export class ContractsComponent implements OnInit {
...
     private cdRef: ChangeDetectorRef,
   ) { }
 
   async ngOnInit() {
 
     this.movieId = this.route.snapshot.paramMap.get('movieId');
-    let contracts: Contract[] = [];
     if (this.movieId) {
-      contracts = await this.contractService.getMovieContracts(this.movieId);
+      this.rows = await this.contractService.getMovieContracts(this.movieId);
```
phpgeekfr, line 46: maybe rename this.rows  to this.contracts

phpgeekfr (approved):
I don't know if this is KFH but this is KBH ;)

I think you missed to rowClick on movie list `/c/o/admin/panel/movies`

libs/admin/src/lib/admin-panel/pages/contracts/contracts.component.ts
phpgeekfr: > I think you missed to rowClick on movie list `/c/o/admin/panel/movies`
If this is because of :
image

You can remove this button since it is also in the `/c/o/admin/panel/movie/movieId` page

Issue #3264, opened 2020-07-08
[CRM APP] Add more info to csv exports


phpgeekfr:
After discussion with @maltmayer and @Jackseed  here is what should be added:

user list

- [ ] add activity description 
- [ ] add number of distinct connections to app (in bigQuery)

movies list

- [ ] add number of screening events for this movie
- [ ] add total  number of invited and attendees

org list

- [ ] add org member count
- [ ] add activity description

event list

- [ ] add privacy setting
- [ ] add host org
- [ ] rename number of attendees => number of invitations
- [ ] add the actual number of attendees. (in bigQuery)

invitations list

- [ ] for past events, add if user attended to event (in bigQuery)

phpgeekfr:
@Jackseed  for this point 
`change app access to be app columns with true or false (e.g. Archipel Content > true | Archipel Market > false)`

One column per app with value  "true" if org has access to dashboard or marketplace (or both) ? No need to distinguish access between dashboard or marketplace ? If so, do I make the same change for the real table in app ? 
![image](https://user-images.githubusercontent.com/1194344/102513472-8c18cf00-408b-11eb-96e2-310795c026f4.png)

Jackseed:
@maltmayer I guess you're the one for this question

maltmayer:
> @Jackseed for this point
> `change app access to be app columns with true or false (e.g. Archipel Content > true | Archipel Market > false)`
> 
> One column per app with value "true" if org has access to dashboard or marketplace (or both) ? No need to distinguish access between dashboard or marketplace ? If so, do I make the same change for the real table in app ?
> ![image](https://user-images.githubusercontent.com/1194344/102513472-8c18cf00-408b-11eb-96e2-310795c026f4.png)

Sorry for the confusion! 
One column per app with value "true" if org has access, "false" if no access. Keep the distinction between marketplace and dashboard :) Same changes for the table yes

phpgeekfr:
@maltmayer What you described is what we currently have in the csv: 

```
catalog | festival
dashboard: no - marketplace: no | dashboard: no - marketplace: yes
```
`One column per app with value "true"  if org has access, "false" if no access`
 => one org can have access to "festival dashboard" but not "festival marketplace"

Do you want one column per app and per module (dashboard marketplace), so 6 columns in total ?

maltmayer:
Per module please.
I need to be able to filter users with this info

maltmayer:
[...]

PR #4587, opened 2020-12-17
  [CRM APP] Add more info to csv exports #3264 part 1

phpgeekfr:


libs/admin/src/lib/admin-panel/pages/organizations/organizations.component.ts
```diff
@@ -68,16 +69,25 @@ export class OrganizationsComponent implements OnInit {
...
+        status: r.status,
+        country: r && r.addresses.main.country ? r.addresses.main.country : '--',
+        email: r.email,
+        memberCount: r.userIds.length,
+        activity: !! r.activity ? r.activity : '--',
+      }
+
+      Object.keys(r.appAccess).forEach(a => {
+        row[`${appName[a]} - dashboard`] = r.appAccess[a].dashboard? 'true' : 'false',
+        row[`${appName[a]} - marketplace`] = r.appAccess[a].dashboard? 'true' : 'false'
```
GrandSchtroumpf, line 86: Don't you mean `r.appAccess[a].marketplace` here ?

GrandSchtroumpf: I would have done : 
```typescript
for (const app in r.appAccess) {
  for (const module of modules) {
    row[`${appName[app]} - ${module}`] = r.appAccess[app][module] ? 'true' : 'false',
  }
}
```

GrandSchtroumpf (approved):

libs/admin/src/lib/admin-panel/pages/organizations/organizations.component.ts
phpgeekfr: good catch

PR #4598, opened 2020-12-18
Currency on dialog form

Mussuranae:
Add a currency icon on the investment dialog form.
image

I need suggestion for the getCurrencySymbol function, I have the feeling that it is not the best way but I prefer to use the function instead of doing if statement in the html.

### DIFF
--- a/apps/financiers/financiers/src/app/marketplace/title/view/view.component.html
+++ b/apps/financiers/financiers/src/app/marketplace/title/view/view.component.html
@@ -58,12 +58,12 @@
           <mat-form-field fxFlex appearance="outline">
             <mat-label>From</mat-label>
             <input type="number" matInput formControlName="from" min="0" />
-            <span matPrefix>{{ getCurrencySymbol((campaign$ | async)?.currency, 'wide') }}</span>
+            <mat-icon matPrefix [svgIcon]="(campaign$ | async)?.currency"></mat-icon>
           </mat-form-field>
           <mat-form-field fxFlex appearance="outline">
             <mat-label>To</mat-label>
             <input type="number" matInput formControlName="to" min="0" />
-            <span matPrefix>{{ getCurrencySymbol((campaign$ | async)?.currency, 'wide') }}</span>
+            <mat-icon matPrefix [svgIcon]="(campaign$ | async)?.currency"></mat-icon>
           </mat-form-field>
         </div>
         <mat-form-field appearance="outline">
--- a/apps/financiers/financiers/src/app/marketplace/title/view/view.component.html
+++ b/apps/financiers/financiers/src/app/marketplace/title/view/view.component.html
@@ -58,10 +58,12 @@
           <mat-form-field fxFlex appearance="outline">
             <mat-label>From</mat-label>
             <input type="number" matInput formControlName="from" min="0" />
+            <span matPrefix>{{ getCurrencySymbol((campaign$ | async)?.currency, 'wide') }}</span>
           </mat-form-field>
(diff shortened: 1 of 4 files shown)


apps/financiers/financiers/src/app/marketplace/title/view/view.component.html
```diff
@@ -58,10 +58,12 @@ <h2>What do you want to know about this project?</h2>
           <mat-form-field fxFlex appearance="outline">
             <mat-label>From</mat-label>
             <input type="number" matInput formControlName="from" min="0" />
+            <span matPrefix>{{ getCurrencySymbol((campaign$ | async)?.currency, 'wide') }}</span>
```
fritzschoff, line 61: There is a currency pipe you can use. 
```
 amount | currency: form.getCurrency() : 'symbol' : '1.0-0' 
 ``` 
 blockframes/libs/organization/campaign/src/lib/form/fundings/fundings.component.html

**Review by fritzschoff — CHANGES_REQUESTED** [review]

File: apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts [hunk 2]
```diff
@@ -118,6 +128,7 @@ export class MarketplaceMovieViewComponent implements OnInit {
       for (const user of users) {
         const data = {
           ...emailData,
+          currency: getCurrencySymbol(this.currency, 'wide'),
```
fritzschoff, line 131: whydo you need that?

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
```diff
@@ -91,6 +94,14 @@ export class MarketplaceMovieViewComponent implements OnInit {
     this.campaign$ = this.movieQuery.selectActiveId().pipe(
       switchMap(id => this.campaignService.valueChanges(id))
     );
+
+    this.sub = this.campaign$.subscribe(data => {
+      this.currency = data.currency;
```
fritzschoff, line 99: do you need that?

fritzschoff (changes requested):

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
Mussuranae: To change the template of the email, the currency should be updated depending on the film/campaign. It's a rough data in the template for now but it should be updated after the next hotfix if we implement this commit. And we don't want a `EUR 123` but `€123` in the email.

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
Mussuranae: It's related to the comment above

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
```diff
@@ -91,6 +94,14 @@ export class MarketplaceMovieViewComponent implements OnInit {
     this.campaign$ = this.movieQuery.selectActiveId().pipe(
       switchMap(id => this.campaignService.valueChanges(id))
```
fritzschoff, line 95: add a tap here so you don't need to subscribe to the campaign$ two lines below.
``` 
   this.campaign$ = this.movieQuery.selectActiveId().pipe(
      switchMap(id => this.campaignService.valueChanges(id)),
      tap(campaign => this.currency= campaing.currency)
```

fritzschoff (approved):

PR #4612, opened 2020-12-22
Issue 3908

### LINKED ISSUE #3908: Factorize the confirmExit in shell
Currently we need to reimplement the `confirmExit` on every tunnel shell
A simple solution would be to create a parent component `TunnelShell` that manage the `confirmExit` methods in the shells.
Or create a reusable function just for this method...
I'm open to ideas

fritzschoff:
close #3908


libs/movie/src/lib/movie/form/movie.shell.config.ts
```diff
@@ -76,7 +76,7 @@ export class MovieShellConfig implements FormShellConfig<MovieControl, Movie> {
     return [onMovieChanges, onStatusChanges];
   }
 
-  async onSave({ publishing }: FormSaveOptions): Promise<any> {
+  async onSave({ publishing }: FormSaveOptions): Promise<void> {
```
pldespaigne, line 79: There is no need to explicitly type here, InetlliSense will do it for you

libs/ui/src/lib/tunnel/exit/exit.component.ts
```diff
@@ -1,27 +1,70 @@
...
+      this.navigate();
+    }
+    const dialogRef = this.dialog.open(TunnelConfirmComponent, {
+      width: '80%',
+      data: this.dialogText
+    });
+    this.sub = dialogRef.afterClosed().subscribe(result => {
+      if (typeof result === 'undefined') {
+        return;
+      }
```
pldespaigne, line 51: Is this the intended behavior ?
"If the user click outside of the dialog, we neither save nor navigate"

Mussuranae (approved):

pldespaigne (approved):

libs/ui/src/lib/tunnel/exit/exit.component.ts
fritzschoff: yes we do nothing, just close modal

libs/ui/src/lib/tunnel/exit/exit.component.ts
```diff
@@ -1,27 +1,70 @@
...
+      result ? this.save() : this.navigate()
+    });
+  }
+
+  private async save() {
+    for (const name in this.configs) {
```
GrandSchtroumpf, line 57: Use Promise.all

libs/ui/src/lib/tunnel/exit/exit.component.ts
```diff
@@ -1,27 +1,70 @@
...
+    @Inject(FORMS_CONFIG) private configs: ShellConfig,
+    private dialog: MatDialog
   ) { }
 
   ngOnInit() {
     this.routeBeforeTunnel = this.exitRedirect || '/c/o/';
   }
 
-  async redirect() {
+  redirect() {
```
GrandSchtroumpf, line 39: This is not the right way to handle exit. What happens if the user click on the browser back button ?

GrandSchtroumpf (changes requested):

Issue #4153, opened 2020-11-04
Algolia autocomplete selects wrong value


RemcoSimonides:
First value in the results is selected instead of the one clicked

.gif is a little fast. I click on "Murphy, Ullrich and Hills" but "Murphy Group" is selected

![wrong organization selected](https://user-images.githubusercontent.com/27687382/98130046-1d582b80-1eba-11eb-9817-fac95e6c5900.gif)

phpgeekfr:
Bug it still there, it takes the first result

PR #4616, opened 2020-12-22
Mano algolia autocomplete  Fixes#4153

xmano:
Fix algolia autocomplete #4153


libs/ui/src/lib/algolia/autocomplete/algolia-autocomplete.component.ts
```diff
@@ -140,15 +138,12 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
...
    */
-  public displayFn() {
-    if (this.resetInput) {
-      return ''
-    }
+  public getSelectedOrg(orgName: string) {
     const value = this.lastValue$.getValue();
     if (value) {
-      return getDeepValue(value[0], this.keyToDisplay);
+      const orgSelected = value.filter(org => org[this.keyToDisplay] === orgName);
```
fritzschoff, line 144: same here for the var name

libs/ui/src/lib/algolia/autocomplete/algolia-autocomplete.component.ts
```diff
@@ -140,15 +138,12 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
    * This function is used internally by the MatAutocomplete to decide
    * what to display in the input field when an option has been selected
    */
-  public displayFn() {
```
fritzschoff, line 143: why do you deleted that function? Also if the developer sets `this.resetInput` to true, you won't reset the input

libs/ui/src/lib/algolia/autocomplete/algolia-autocomplete.component.ts
```diff
@@ -140,15 +138,12 @@ export class AlgoliaAutocompleteComponent implements OnInit, OnDestroy {
...
-    if (this.resetInput) {
-      return ''
-    }
+  public getSelectedOrg(orgName: string) {
```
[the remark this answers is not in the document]
fritzschoff: Same for the parameter

fritzschoff (changes requested):

libs/ui/src/lib/algolia/autocomplete/algolia-autocomplete.component.ts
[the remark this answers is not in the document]
xmano: ok.. will change the names.

libs/ui/src/lib/algolia/autocomplete/algolia-autocomplete.component.ts
xmano: ok.. will change the var name

libs/ui/src/lib/algolia/autocomplete/algolia-autocomplete.component.ts
xmano: That function is not needed. But I need to check where to handle `resetInput`. Will update.

xmano: `
  public selected(result: any) {
    this.selectionChange.emit(result);
    if (this.resetInput) {
      this.control.reset();
    }
  }
`

wont the above code handle it?

xmano: function removed entirely.

fritzschoff:
looks good now

PR #4621, opened 2020-12-23
Prevent user from double invitation

### LINKED ISSUE #4311: Prevent user to ask for several invitation for a screener
From @Jackseed 
```
Feedback from prod: this banner enables user to have multiple invitations on the same event 
(either by previous invitation or by requesting access through the event page/list).
This is a non wanted behaviour as there could be one accepted and one refused 
and it's confusing for the sales agent to receive a request from a user already invited.

Could we limit user invitation for a screening to one only? 
Or disable the "Ask for invitation" button if the user already has an invitation on the event selected?
```

See [#4183](https://github.com/blockframes/blockframes/issues/4183) Debug screening banner on movie view.

Mussuranae:
Users could ask for invitation multiple time on the screening banner. 
Fix #4311


phpgeekfr (approved):
LGTM

PR #4624, opened 2020-12-23
MemberAdded email

### LINKED ISSUE #4529: Update memberAdded email
This email should be triggered only when a user accepts an invitation to join an org. 
Also some variables should be added:
- new user first name & last name
- org name
- admin first name

Mussuranae:
Update memberAdded email with new variables that can be used inside the template of sendgrid.
Fix #4529


apps/backend-functions/src/internals/invitations/organizations.ts
```diff
@@ -15,6 +15,7 @@ import {
 } from '../../templates/mail';
 import { getAdminIds, getDocument, getAppUrl, getOrgAppKey } from '../../data/internals';
```
phpgeekfr, line 16: remove unsed import `getUserMail`

phpgeekfr (approved):

PR #4625, opened 2020-12-23
Email for request to event accepted by org

### LINKED ISSUE #4092: New email - Request to attend a screening accepted
Buyer should receive an email when their request to attend a screening is accepted. 
Mail is ready on sendgrid, template id `d-df7b0a372b994a3090a48dc6cf17ba3e`

Mussuranae:
Add a new function to create email when organization accepted a request from an user to attend one of its event.
Will need several changes on the template on Sendgrid :
- OrgName variable in the subject need to be changed
- I don't put different variable for the time and the date of the event. It's in one variable (instead of 'month/day/year' and 'hours/min/sec', I have 'month/day/year hours:min:sec GMT')

See with Vincent but maybe we need an URL redirecting to the event page ?
Fix #4092

### DIFF
--- a/apps/backend-functions/src/internals/invitations/events.ts
+++ b/apps/backend-functions/src/internals/invitations/events.ts
@@ -4,12 +4,13 @@ import { NotificationDocument, OrganizationDocument, PublicUser } from "../../da
 import { createNotification, triggerNotifications } from "../../notification";
 import { db, getUser } from "../firebase";
 import { getAdminIds, getDocument } from "../../data/internals";
-import { invitationToEventFromOrg, requestToAttendEventFromUser } from '../../templates/mail';
+import { invitationToEventFromOrg, requestToAttendEventFromUser, requestToAttendEventFromUserAccepted } from '../../templates/mail';
 import { sendMailFromTemplate } from '../email';
 import { EventDocument, EventMeta } from "@blockframes/event/+state/event.firestore";
 import { EmailRecipient, getEventEmailData, EventEmailData } from "@blockframes/utils/emails/utils";
 import { App, applicationUrl } from "@blockframes/utils/apps";
 import { orgName, canAccessModule } from "@blockframes/organization/+state/organization.firestore";
+import { format } from "date-fns";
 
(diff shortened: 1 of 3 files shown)


apps/backend-functions/src/internals/invitations/events.ts
```diff
@@ -154,6 +157,21 @@ async function onInvitationToAnEventAccepted({
...
+      const organizerOrgName = orgName(org);
+      const event = await getDocument<EventDocument<EventMeta>>(`events/${eventId}`);
+      const eventStartDate = event.start.toDate();
+      const eventEndDate = event.end.toDate();
+      /**
+       * @dev Format from date-fns lib, here the date will be 'month/day/year, hours:min:sec am/pm GMT'
+       * See the doc here : https://date-fns.org/v2.16.1/docs/format
+       */
+      const eventStart = format(eventStartDate, 'Pppp');
+      const eventEnd = format(eventEndDate, 'Pppp');
```
phpgeekfr, line 170: You could have done this directly in `requestToAttendEventFromUserAccepted` an pass entire `event` object to it

apps/backend-functions/src/internals/invitations/events.ts
```diff
@@ -144,6 +144,9 @@ async function onInvitationToAnEventAccepted({
 
   const notifications: NotificationDocument[] = [];
 
+  // @TODO (#2848) forcing to festival since invitations to events are only on this one
+  const appKey: App = 'festival';
```
phpgeekfr, line 148: declare appKey only if used (inside else if)

phpgeekfr (approved):

apps/backend-functions/src/internals/invitations/events.ts
phpgeekfr: Why don't use `getEventEmailData` that you created in another PR ?

Mussuranae: Because I created it after this one, I will merge the other one before and update this one after :)

Issue #3189, opened 2020-06-25
[CRM APP] - Missing features bis


phpgeekfr:
This is the updated version of #2090.

- [x]  Add a way to delete a movie / contract / user / ... #3933  #3932 
- [ ] change user avatar (lower priority)
- [x] change movie images (high priority) (draft PR Clélia)

- [x] remove `Image sizes` that is now obsolete

phpgeekfr:
@Jackseed is `change user avatar (lower priority)` still a required feature ?

Jackseed:
No we can close this one and open it if the need comes back

PR #4630, opened 2020-12-28
Crm movie picture

Mussuranae:
Add a component on movie page on CRM to be able to change poster, banner and still photos of the movie.

- [x] Banner
- [x] Poster
- [x] Stills

See issue #3189 

EDIT : I change also a bit the form for the productionStatus, we were able to change it one time but not more without refreshing the page. There was also a bug on the appAccess because of a missing data.

### DIFF
--- a/libs/admin/src/lib/admin-panel/pages/movie/movie.component.html
+++ b/libs/admin/src/lib/admin-panel/pages/movie/movie.component.html
@@ -1,108 +1,105 @@
-<main fxLayout="column" fxFlex>
-  <h1><a routerLink="/c/o/admin/panel/">Administration panel</a> > Movie > {{ movieId }} </h1>
+<h1><a routerLink="/c/o/admin/panel/">Administration panel</a> > Movie > {{ movieId }} </h1>
 
-  <mat-card *ngIf="movie">
-    <movie-header [movie]="movie"></movie-header>
-  </mat-card>
+<article *ngIf="movie">
+  <movie-header [movie]="movie"></movie-header>
+</article>
 
-  <article *ngIf="movie" fxLayout="row" fxLayoutAlign="start" fxLayoutGap="24px" fxFlex>
-    <mat-card>
-      <h2>
-        {{ movie.title.international }} ({{ movie.internalRef || 'missing internal ref' }})
-        <admin-go-to type="title" [id]="movieId"></admin-go-to>
-      </h2>
+<section *ngIf="movie" fxLayout="row" fxLayoutAlign="start" fxLayoutGap="24px" fxFlex>
+  <article class="surface">
+    <h2>
+      {{ movie.title.international }} ({{ movie.internalRef || 'missing internal ref' }})
+      <admin-go-to type="title" [id]="movieId"></admin-go-to>
+    </h2>
 
-      <h4>({{ movie.title.original }})</h4>
+    <h4>({{ movie.title.original }})</h4>
 
-      <ng-container *ngIf="movieForm">
-        <form [formGroup]="movieForm" fxLayout="column">
-          <mat-form-field appearance="outline">
-            <mat-label>Internal ref</mat-label>
-            <input matInput type="text" placeholder="Internal ref" formControlName="internalRef" />
-          </mat-form-field>
-          <mat-form-field appearance="outline">
-            <mat-label>Movie Store Status</mat-label>
-            <mat-select formControlName="storeStatus" required>
-              <mat-option *ngFor="let status of storeStatus | keyvalue" [value]="status.key">
-                {{ status.value }}
-              </mat-option>
-            </mat-select>
-          </mat-form-field>
-          <mat-form-field appearance="outline">
-            <mat-label>Movie Store Type</mat-label>
-            <mat-select formControlName="storeType" required>
-              <mat-option *ngFor="let type of storeType | keyvalue" [value]="type.key">
-                {{ type.value }}
-              </mat-option>
-            </mat-select>
-          </mat-form-field>
-          <mat-form-field appearance="outline">
-            <mat-label>Production Status</mat-label>
-            <mat-select formControlName="productionStatus" required>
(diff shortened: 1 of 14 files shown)


libs/admin/src/lib/admin-panel/pages/movie/movie.component.html
```diff
@@ -1,11 +1,11 @@
-<main>
+<main fxLayout="column" fxFlex>
```
GrandSchtroumpf, line 1: looks like the full page is in a main. I would recommend removing it and do : 
```scss
:host {
  display: flex;
  flex-direction: column;
  height: 100%;
}
```

libs/admin/src/lib/admin-panel/pages/movie/movie.component.html
```diff
@@ -68,35 +68,41 @@ <h2>Application Access</h2>
...
-        </ng-template>
-        <ng-template colRef="terms" let-terms>
-          <strong>From : </strong>{{ terms | termDate: 'start' }} <strong>To : </strong>{{ terms | termDate: 'end' }}
-        </ng-template>
-        <ng-template colRef="status" let-status>
-          {{ status | toLabel: 'distributionRightStatus'  }}
-        </ng-template>
-      </bf-table-filter>
-    </mat-card>
+  <mat-card>
```
GrandSchtroumpf, line 72: Don't use mat-card if you only want the background color. 
Use a regular section with the `var(--background-card)` instead

libs/admin/src/lib/admin-panel/pages/movie/movie.component.html
```diff
@@ -68,35 +68,41 @@ <h2>Application Access</h2>
...
+        <strong>From : </strong>{{ terms | termDate: 'start' }} <strong>To : </strong>{{ terms | termDate: 'end' }}
+      </ng-template>
+      <ng-template colRef="status" let-status>
+        {{ status | toLabel: 'distributionRightStatus'  }}
+      </ng-template>
+    </bf-table-filter>
+  </mat-card>
+
+  <section fxFlex fxLayout="row" fxLayoutGap="24px">
+    <mat-card fxFlex *ngIf="movie">
```
GrandSchtroumpf, line 98: same as above, you can do `<article class="surface">` to apply color and padding

libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -179,7 +179,7 @@ export function createMovieOriginalRelease(
   return {
     country: null,
     ...params,
-    date: toDate(params.date),
+    date: toDate(params.date) ?? null,
```
GrandSchtroumpf, line 182: The problem is with undefined not managed by firestore ? Maybe changing the `toDate` function might be more straghtforward.

libs/movie/src/lib/movie/components/picture-upload/picture-upload.component.html
```diff
@@ -0,0 +1,31 @@
+<section fxFlex fxLayout="column">
```
GrandSchtroumpf, line 1: you don't need the section you can remove it and style the `:host`

libs/movie/src/lib/movie/components/picture-upload/picture-upload.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  async uploadPicture() {
+    if (!this.moviePictureForm.valid) {
+      this.snackBar.open('Form invalid, please check error messages', 'close', { duration: 2000 });
+      return;
+    }
+
+    const { documentToUpdate, mediasToUpload } = extractMediaFromDocumentBeforeUpdate(this.moviePictureForm);
+
+    const pictures = {
```
GrandSchtroumpf, line 42: I don't see the benefit of creating a `picture` variable as you're not using it after : 
```typescript
this.movie.poster = documentToUpdate.poster;
```

libs/movie/src/lib/movie/components/video-upload/video-upload.component.html
```diff
@@ -1,64 +1,60 @@
...
-        <file-upload
-          [form]="form.screener.get('ref')"
-          [storagePath]="getPath('screener')"
-          allowedFileType="video"
-          [filePrivacy]="filePrivacy">
-            <h3 title>Upload Screener</h3>
-        </file-upload>
+<form [formGroup]="form" fxLayout="column" fxLayoutAlign="center" fxFlex>
+  <h2>Upload screening</h2>
+  <div [formGroup]="form.screener" fxLayout="column">
```
GrandSchtroumpf, line 3: Prefere `formGroupName="screener"`

GrandSchtroumpf (changes requested):

libs/admin/src/lib/admin-panel/pages/movie/movie.component.html
Mussuranae: I use the surface class you send, that's better than put variable everywhere

libs/movie/src/lib/movie/+state/movie.model.ts
Mussuranae: Yes it is. I put a condition inside the `toDate` function, so that we won't have to put all the time `?? null`

libs/movie/src/lib/movie/components/picture-upload/picture-upload.component.ts
Mussuranae: Yeah, I used it but I can completely change this and use directly `documentToUpdate`. :+1:

libs/utils/src/lib/helpers.ts
```diff
@@ -61,6 +61,9 @@ export function toDate(date: firestore.Timestamp | Date): Date {
   if (typeof date === 'string' || typeof date === 'number') {
     return new Date(date);
   }
+  if (!date) {
+    return null;
```
GrandSchtroumpf, line 65: Add a comment `// Return null to avoid undefined error with firestore`

GrandSchtroumpf (approved):

PR #4635, opened 2020-12-30
Admin invite user to join org

### LINKED ISSUE #739: When a user is added to an org, clear other invitations
users can have only one org, clear other invites and block orgId updates on users

### LINKED ISSUE #2638: Admin invite problem
- [ ] when you're admin on Archipel and you invite someone to join your org, it works the first time. But if you decide to delete the invitation. And for some reason you want to invite the same person again, it doesn't work, they never receive the other email(s).

Mussuranae:
Here is the problem :
when an admin invite user to join org, then remove the invitation, and invite again the same user, there is only one email sent to the user to invite him on the platform, but 2 invitations were sent.
This is because we don't remove the user in the authentication part when we delete the first invitation, so the second email isn't sent correctly.

I create a backend function when an invitation is removed to delete also the user in the authentication part IF the user hasn't orgId already (that would mean he joined an other org or create his own) and if there is no other invitation (if 2 organizations sent him an invitation, we don't want to delete his account if one of the org cancel its invitation).

I'm opened to more improvement and in case I didn't think about an other scenario.

Fix #2638 
Fix #739

### DIFF
--- a/apps/backend-functions/src/invitation.ts
+++ b/apps/backend-functions/src/invitation.ts
@@ -44,6 +44,7 @@ export async function onInvitationWrite(
       const invitationCollectionRef = db.collection('invitations')
         .where('toUser.uid', '==', user.uid)
         .where('mode', '==', 'invitation')
+        .where('status', '==', 'pending')
       const existingInvitation = await invitationCollectionRef.get();
 
       // If there is an other invitation or the user has already an org, we don't want to delete its account
--- a/apps/backend-functions/src/invitation.ts
+++ b/apps/backend-functions/src/invitation.ts
@@ -1,5 +1,5 @@
 import { getDocument, createPublicOrganizationDocument, createPublicUserDocument } from './data/internals';
-import { db, functions, getUser } from './internals/firebase';
+import { db, getUser } from './internals/firebase';
 import { InvitationOrUndefined, OrganizationDocument } from './data/types';
 import { onInvitationToJoinOrgUpdate, onRequestToJoinOrgUpdate } from './internals/invitations/organizations';
 import { onInvitationToAnEventUpdate } from './internals/invitations/events';
@@ -10,8 +10,9 @@ import { ErrorResultResponse } from './utils';
 import { CallableContext } from "firebase-functions/lib/providers/https";
 import { App } from '@blockframes/utils/apps';
 import { EventDocument, EventMeta, MEETING_MAX_INVITATIONS_NUMBER } from '@blockframes/event/+state/event.firestore';
-import { Change } from 'firebase-functions';
 import { EventEmailData, getEventEmailData } from '@blockframes/utils/emails/utils';
+import { Change, EventContext } from 'firebase-functions';
+import * as admin from 'firebase-admin';
 
 /**
  * Handles firestore updates on an invitation object,
@@ -97,6 +98,36 @@ interface UserInvitation {
   invitation: Partial<InvitationBase<any>>;
   app?: App;
 }
+
+export async function onInvitationDelete(
+  invitSnapshot: FirebaseFirestore.DocumentSnapshot<InvitationOrUndefined>,
+  context: EventContext
+): Promise<any>{
+  const invitation = invitSnapshot.data() as InvitationOrUndefined;
+  const toUser = invitation.toUser;
+
+  // Remove user in users collection and in authentication part
+  if(invitation.type === "joinOrganization"
+  && !!toUser
+  && (invitation.status === "pending")) {
+
+    // Fetch potential other invitations
+    const invitationCollectionRef = db.collection('invitations')
+      .where('toUser.uid', '==', toUser.uid)
+      .where('mode', '==', 'invitation')
+      .where('type', '==', 'joinOrganization')
+    const invitationsSnap = await invitationCollectionRef.get();
+
+    // If there is an other invitation or the user has already an org, we don't want to delete its account
+    if(invitationsSnap.docs.length > 1 || !!toUser.orgId) {
+      return;
+    }
+
+    admin.auth().deleteUser(toUser.uid);
+    db.doc(`users/${toUser.uid}`).delete();
+  }
+}
+
 
 /**
  * Invite a list of user by email.
--- a/apps/backend-functions/src/invitation.ts
+++ b/apps/backend-functions/src/invitation.ts
@@ -11,8 +11,8 @@ import { CallableContext } from "firebase-functions/lib/providers/https";
 import { App } from '@blockframes/utils/apps';
 import { EventDocument, EventMeta, MEETING_MAX_INVITATIONS_NUMBER } from '@blockframes/event/+state/event.firestore';
 import { EventEmailData, getEventEmailData } from '@blockframes/utils/emails/utils';
-import { Change, EventContext } from 'firebase-functions';
-import * as admin from 'firebase-admin';
+import { Change } from 'firebase-functions';
+import { invitationStatus } from '@blockframes/invitation/+state/invitation.firestore';
 
 /**
  * Handles firestore updates on an invitation object,
@@ -34,7 +34,7 @@ export async function onInvitationWrite(
   const invitationDoc = after.data() as InvitationOrUndefined;
 
   // Doc was deleted
-  if (!invitationDoc) {
+  if (!invitationDoc && !!invitationDocBefore.toUser) {
     const user = await getUser(invitationDocBefore.toUser.uid)
 
     // Remove user in users collection
@@ -54,6 +54,8 @@ export async function onInvitationWrite(
 
       db.doc(`users/${user.uid}`).delete();
     }
+
+    return;
   }
 
   // Because of rules restrictions, event creator might not have access to other informations than the id.
@@ -186,15 +188,13 @@ That would have exceeded the current limit which is ${MEETING_MAX_INVITATIONS_NU
 }
 
 export async function isInvitationToJoinOrgExist(userEmails: string[]) {
-  for (const email of userEmails) {
-    const invitationRef = db.collection('invitations')
-    .where('type', '==', 'joinOrganization')
-    .where('toUser.email', '==', email);
-    const invitationDocs = await invitationRef.get();
 
-    if(invitationDocs.docs.length) {
-      return true
-    }
-  }
-  return false;
+  const invitationPromises = userEmails.map(email => db.collection('invitations')
+    .where('type', '==', 'joinOrganization')
+    .where('toUser.email', '==', email)
+    .where('status', 'in', invitationStatus.filter(s => s !== 'declined'))
+    .get());
+  const invitationQuery = await Promise.all(invitationPromises);
+  const isInvitationExist = invitationQuery.some(d => d.docs.length > 0)
+  return isInvitationExist;
 }
--- a/apps/backend-functions/src/invitation.ts
+++ b/apps/backend-functions/src/invitation.ts
@@ -33,8 +33,27 @@ export async function onInvitationWrite(
   const invitationDocBefore = before.data() as InvitationOrUndefined;
   const invitationDoc = after.data() as InvitationOrUndefined;
 
-  // Doc was deleted, ignoring...
-  if (!invitationDoc) { return; }
+  // Doc was deleted
+  if (!invitationDoc) {
+    const user = await getUser(invitationDocBefore.toUser.uid)
+
+    // Remove user in users collection
+    if(invitationDocBefore.mode === "invitation" && !!user && invitationDocBefore.status === "pending") {
+
+      // Fetch potential other invitations to join org
+      const invitationCollectionRef = db.collection('invitations')
+        .where('toUser.uid', '==', user.uid)
+        .where('mode', '==', 'invitation')
+      const existingInvitation = await invitationCollectionRef.get();
+
+      // If there is an other invitation or the user has already an org, we don't want to delete its account
+      if(existingInvitation.docs.length > 1 || !!user.orgId || !!user.firstName || !!user.lastName) {
+        return;
+      }
+
+      db.doc(`users/${user.uid}`).delete();
+    }
+  }
 
   // Because of rules restrictions, event creator might not have access to other informations than the id.
   // We consolidate invitation document here.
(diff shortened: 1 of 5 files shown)


apps/backend-functions/src/invitation.ts
```diff
@@ -97,6 +98,36 @@ interface UserInvitation {
...
+  && !!toUser
+  && (invitation.status === "pending")) {
+
+    // Fetch potential other invitations
+    const invitationCollectionRef = db.collection('invitations')
+      .where('toUser.uid', '==', toUser.uid)
+      .where('mode', '==', 'invitation')
+      .where('type', '==', 'joinOrganization')
+    const invitationsSnap = await invitationCollectionRef.get();
+
```
Mussuranae, line 119: Is it better here to check the uid or the email of the user ?

apps/backend-functions/src/main.ts
```diff
@@ -98,6 +98,9 @@ export const onPermissionDeleteEvent = onDocumentDelete('permissions/{orgID}',on
 /** Trigger: when an invitation is updated (e. g. when invitation.status change). */
 export const onInvitationUpdateEvent = onDocumentWrite('invitations/{invitationID}', onInvitationWrite);
 
+/** Trigger: when an invitation is updated (e. g. when invitation.status change). */
```
phpgeekfr, line 101: wrong comment

apps/backend-functions/src/invitation.ts
```diff
@@ -97,6 +98,36 @@ interface UserInvitation {
...
 
+export async function onInvitationDelete(
+  invitSnapshot: FirebaseFirestore.DocumentSnapshot<InvitationOrUndefined>,
+  context: EventContext
+): Promise<any>{
+  const invitation = invitSnapshot.data() as InvitationOrUndefined;
+  const toUser = invitation.toUser;
+
+  // Remove user in users collection and in authentication part
+  if(invitation.type === "joinOrganization"
```
phpgeekfr, line 109: Should `mode === 'invitation'` be also tested ?

apps/backend-functions/src/invitation.ts
```diff
@@ -97,6 +98,36 @@ interface UserInvitation {
...
+
+    // Fetch potential other invitations
+    const invitationCollectionRef = db.collection('invitations')
+      .where('toUser.uid', '==', toUser.uid)
+      .where('mode', '==', 'invitation')
+      .where('type', '==', 'joinOrganization')
+    const invitationsSnap = await invitationCollectionRef.get();
+
+    // If there is an other invitation or the user has already an org, we don't want to delete its account
+    if(invitationsSnap.docs.length > 1 || !!toUser.orgId) {
```
phpgeekfr, line 121: can you rename `invitationsSnap` to existingInvitationToJoinOrg or someting like this ?

apps/backend-functions/src/invitation.ts
phpgeekfr: If you look at `inviteUsers` you should see that every inviation of this kind will have toUser.uid and toUser.email

apps/backend-functions/src/invitation.ts
phpgeekfr: Because if this is a 'request' user don't want to delete his account if we want to cancel his request to join org

phpgeekfr (changes requested):

libs/organization/src/lib/organization/components/member-add/member-add.component.ts
```diff
@@ -56,14 +56,14 @@ export class MemberAddComponent {
       this._isSending.next(true);
       const emails = this.form.value;
       const invitationsExist = await this.invitationService.orgInvitationExists(emails);
-      if (invitationsExist) throw new Error('You already send an invitation to one or more of these users');
+      if (invitationsExist) throw new Error('There is already invitation existing for one or more of these users');
```
phpgeekfr, line 59: there is already *an* invitation no ?

libs/invitation/src/lib/+state/invitation.service.ts
```diff
@@ -56,14 +56,8 @@ export class InvitationService extends CollectionService<InvitationState> {
...
   public async orgInvitationExists(userEmails: string[]): Promise<boolean> {
-    const orgId = this.authQuery.orgId;
-    const orgInvitations = await this.getValue(ref => ref.where('mode', '==', 'invitation')
-      .where('type', '==', 'joinOrganization')
-      .where('fromOrg.id', '==', orgId));
-
-    return orgInvitations.some(
-      invitation => userEmails.includes(invitation.toUser.email) && invitation.status === 'pending'
-    );
+    const f = this.functions.httpsCallable('isInvitationToJoinOrgExisted');
```
phpgeekfr, line 59: You can declare httpCallable like this 
```
private isInvitationToJoinOrgExisted = this.functions.httpsCallable('isInvitationToJoinOrgExisted');
```

Also, the name `isInvitationToJoinOrgExisted` is not correct I think

apps/backend-functions/src/invitation.ts
```diff
@@ -163,3 +199,19 @@ That would have exceeded the current limit which is ${MEETING_MAX_INVITATIONS_NU
     }
   })
 }
+
+export async function isInvitationToJoinOrgExists(userEmails: string[]) {
```
phpgeekfr, line 203: I think that parameters are wrong, an httpcallable always have  `data` and `context`

apps/backend-functions/src/invitation.ts
```diff
@@ -98,6 +99,41 @@ interface UserInvitation {
...
+
+  // Remove user in users collection and in authentication part
+  if(invitation.mode === "invitation" && !!toUser) {
+
+    // Fetch potential other invitations to join org
+    const invitationCollectionRef = db.collection('invitations')
+      .where('toUser.uid', '==', toUser.uid)
+      .where('mode', '==', 'invitation')
+      .where('type', '==', 'joinOrganization');
+    const existingInvitationToJoinOrg = await invitationCollectionRef.get();
```
phpgeekfr, line 117: Is this check useful ? This should not append thank to your new `isInvitationToJoinOrgExists` method

apps/backend-functions/src/invitation.ts
```diff
@@ -163,3 +199,19 @@ That would have exceeded the current limit which is ${MEETING_MAX_INVITATIONS_NU
     }
   })
 }
+
+export async function isInvitationToJoinOrgExists(userEmails: string[]) {
+  let invitationExists = false;
```
phpgeekfr, line 204: this is incorrect. Your method will return directly after the first item of your array

you can do 
```
  for (let email of userEmails) {
    const invitationRef = db.collection('invitations')
    .where('toUser.email', '==', email);
    const invitationDocs = await invitationRef.get();

    if(invitationDocs.docs.length) {
      return true;
    }
  }
 return false;
  ```

Comment (phpgeekfr) [reply apps/backend-functions/src/invitation.ts:204]: you could also have done a promise.all to make the queries in parallel

File: apps/backend-functions/src/invitation.ts [hunk 10]
```diff
@@ -98,6 +99,41 @@ interface UserInvitation {
...
+      .where('mode', '==', 'invitation')
+      .where('type', '==', 'attendEvent');
+    const existingInvitationToAttendEvent = await invitationEventCollectionRef.get();
+
+    // If there is an other invitation or the user has already an org, we don't want to delete its account
+    if(existingInvitationToJoinOrg.docs.length > 1 || existingInvitationToAttendEvent.docs.length > 1 || !!toUser.orgId) {
+      return;
+    }
+
+    admin.auth().deleteUser(toUser.uid);
```
phpgeekfr, line 131: do not delete auth. let `onUserDelete` do the job

apps/backend-functions/src/invitation.ts
```diff
@@ -98,6 +99,41 @@ interface UserInvitation {
...
+
+    // Fetch potential invitations to attend event
+    const invitationEventCollectionRef = db.collection('invitations')
+      .where('toUser.uid', '==', toUser.uid)
+      .where('mode', '==', 'invitation')
+      .where('type', '==', 'attendEvent');
+    const existingInvitationToAttendEvent = await invitationEventCollectionRef.get();
+
+    // If there is an other invitation or the user has already an org, we don't want to delete its account
+    if(existingInvitationToJoinOrg.docs.length > 1 || existingInvitationToAttendEvent.docs.length > 1 || !!toUser.orgId) {
```
phpgeekfr, line 127: I think you should only check if user have an orgId or a lastName/firstName  (by fetching real user from DB, not by using toUser)

apps/backend-functions/src/invitation.ts
```diff
@@ -98,6 +99,41 @@ interface UserInvitation {
   app?: App;
 }
 
+export async function onInvitationDelete(
+  invitSnapshot: FirebaseFirestore.DocumentSnapshot<InvitationOrUndefined>,
+  context: EventContext
```
phpgeekfr, line 104: why this ? you don't use context here

apps/backend-functions/src/invitation.ts
phpgeekfr: So if invitation is deleted, user that received his credentials by emails will not be able to login ?

phpgeekfr (changes requested):

libs/organization/src/lib/organization/components/member-add/member-add.component.ts
Mussuranae: Yeah, good catch ! or maybe plural also. I will change that !

apps/backend-functions/src/invitation.ts
```diff
@@ -98,6 +99,41 @@ interface UserInvitation {
...
+): Promise<any>{
+  const invitation = invitSnapshot.data() as InvitationOrUndefined;
+  const toUser = invitation.toUser;
+
+  // Remove user in users collection and in authentication part
+  if(invitation.mode === "invitation" && !!toUser) {
```
phpgeekfr, line 110: should'nt you check the invitation.status also ?

apps/backend-functions/src/invitation.ts
Mussuranae: Yeah I think you're right, but as it is not implemented on prod, we can maybe have the case of several invitation to join an organization ? Do we plan to clean the db for that ? What do we plan for invitation that are pending for x months etc ?
There is no `onInvitationUpdate` backend function, maybe we can create one and delete others invitations to join org when an user accepted one ?

apps/backend-functions/src/invitation.ts
Mussuranae: I wasn't sure because if user accepted to join an org, and several months later he changed to an other org (by asking us for example), there would be always an invitation accepted but not pending or declined, and we don't want to delete the user if there is other invitation. What do you think about that ?

apps/backend-functions/src/main.ts
```diff
@@ -98,6 +98,12 @@ export const onPermissionDeleteEvent = onDocumentDelete('permissions/{orgID}',on
 /** Trigger: when an invitation is updated (e. g. when invitation.status change). */
 export const onInvitationUpdateEvent = onDocumentWrite('invitations/{invitationID}', onInvitationWrite);
 
+/** Trigger: when an invitation is deleted. */
+export const onInvitationDeleteEvent = onDocumentDelete('invitations/{invitationID}', onInvitationDelete);
```
phpgeekfr, line 102: remove this and make it part of `onInvitationUpdateEvent`

apps/backend-functions/src/invitation.ts
Mussuranae: Yes, but the user is deleted only if there is no other invitation (to join organization or to attend an event). I added the check of the firstname and lastname, so normally it will remove the user only if he/she has no other invitation and never finish his/her register on the app.

apps/backend-functions/src/invitation.ts
```diff
@@ -32,8 +33,28 @@ export async function onInvitationWrite(
   const invitationDocBefore = before.data() as InvitationOrUndefined;
   const invitationDoc = after.data() as InvitationOrUndefined;
 
-  // Doc was deleted, ignoring...
-  if (!invitationDoc) { return; }
+  // Doc was deleted
+  if (!invitationDoc) {
```
phpgeekfr, line 37: this if statement need to return somehting no matter what , rest of code will crash otherwise

apps/backend-functions/src/invitation.ts
```diff
@@ -163,3 +184,17 @@ That would have exceeded the current limit which is ${MEETING_MAX_INVITATIONS_NU
...
+  for (const email of userEmails) {
+    const invitationRef = db.collection('invitations')
+    .where('type', '==', 'joinOrganization')
+    .where('toUser.email', '==', email);
+    const invitationDocs = await invitationRef.get();
+
+    if(invitationDocs.docs.length) {
+      return true
+    }
+  }
```
phpgeekfr, line 198: still possible to use promise.all to speed up process

apps/backend-functions/src/invitation.ts
```diff
@@ -32,8 +33,28 @@ export async function onInvitationWrite(
   const invitationDocBefore = before.data() as InvitationOrUndefined;
   const invitationDoc = after.data() as InvitationOrUndefined;
 
-  // Doc was deleted, ignoring...
-  if (!invitationDoc) { return; }
+  // Doc was deleted
+  if (!invitationDoc) {
+    const user = await getUser(invitationDocBefore.toUser.uid)
```
phpgeekfr, line 38: you also need to check if toUser exists

phpgeekfr (changes requested):

apps/backend-functions/src/invitation.ts
```diff
@@ -163,3 +186,15 @@ That would have exceeded the current limit which is ${MEETING_MAX_INVITATIONS_NU
...
+
+export async function isInvitationToJoinOrgExist(userEmails: string[]) {
+
+  const invitationPromises = userEmails.map(email => db.collection('invitations')
+    .where('type', '==', 'joinOrganization')
+    .where('toUser.email', '==', email)
+    .where('status', 'in', invitationStatus.filter(s => s !== 'declined'))
+    .get());
+  const invitationQuery = await Promise.all(invitationPromises);
+  const isInvitationExist = invitationQuery.some(d => d.docs.length > 0)
```
phpgeekfr, line 198: missing `;` and you can directly do ` return invitationQuery.some(d => d.docs.length > 0);` if you want

apps/backend-functions/src/invitation.ts
```diff
@@ -32,8 +33,30 @@ export async function onInvitationWrite(
   const invitationDocBefore = before.data() as InvitationOrUndefined;
   const invitationDoc = after.data() as InvitationOrUndefined;
 
-  // Doc was deleted, ignoring...
-  if (!invitationDoc) { return; }
+  // Doc was deleted
+  if (!invitationDoc && !!invitationDocBefore.toUser) {
```
phpgeekfr, line 37: you must `return; ` event if toUser is undefined. 
What if invitation is deleted bu with fromUser?

phpgeekfr (approved):

Issue #4200, opened 2020-11-09
Unit Test: Contracts Security Rules


xmano:
Unit Test: Contracts Security Rules

xmano:
Fixed by
https://github.com/blockframes/blockframes/pull/4637

PR #4637, opened 2021-01-04
Update(UT): Add unit tests for security rules - contracts.

xmano:
#4200 

Contracts security rules unit tests


xmano:
Test Passing:

image

apps/rules/src/tests/firestore/fixtures/data.ts
```diff
@@ -32,6 +32,9 @@ export const testFixture = {
     canCreate: true,
     ownerId: 'O001',
   },
+  'permissions/O001/documentPermissions/C001': {
```
phpgeekfr, line 35: mixing `0` and `O` is very dangerous ;)

apps/rules/src/tests/firestore/contracts.spec.ts
```diff
@@ -1,18 +1,105 @@
...
+    describe('Read Contracts', () => {
+      it('should be able to read contract', async () => {
+        const contractRef = db.doc('contracts/C001');
+        await assertSucceeds(contractRef.get());
+      });
+    });
+
+    describe.skip('Create Contracts', () => {
+
+    });
```
phpgeekfr, line 67: can be removed

apps/rules/src/tests/firestore/fixtures/data.ts
xmano: using console font it is clearly visible. I don't see a problem here but in other scenarios like UI it can be problem.

I hope it is fine.

apps/rules/src/tests/firestore/contracts.spec.ts
xmano: Ok will remove this.

apps/rules/src/tests/firestore/contracts.spec.ts
```diff
@@ -1,18 +1,105 @@
...
-//TODO: 4200
-describe.skip('Contracts Rules Tests', () => {
+
+describe('Contracts Rules Tests', () => {
   const projectId = `ctrules-spec-${Date.now()}`;
   let db: Firestore;
 
-  beforeAll(async () => {
-    db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-user2' });
+  describe('With User not in org', () => {
```
phpgeekfr, line 14: can your explain `With User not in org` ?  Because an user without org can not be party of  a contract

phpgeekfr (approved):
Even if not going to be updated soon, can you check if `describe` or `it` sentences are good ? 

For skipped test, can you add a comment explaining why ?

apps/rules/src/tests/firestore/contracts.spec.ts
xmano:
```
      match /versions/{versionId} {
        allow read: if (isContractParty(contractId) || orgCan('canRead', userOrgId(), contractId)) && notInMaintenance();
        allow write: if false;
      }
```

User need not be in same org but can be party..

PR #4640, opened 2021-01-04
Fix 4585: wrong image shown

### LINKED ISSUE #4585: Wrong image shown on first item on movie list
Wrong image shown in first item of movie list after searching **if movie doesn't have a poster**.

In the example below I removed the poster of The Killing Of Two Lovers and then searched for it in the movie list.
bug movie list

RemcoSimonides:
Fix #4585 

Couldn't find any bugs caused by removing the `if (!image) return` line. Also can't find why it was added in the first place. Do you know potential bugs?

### DIFF
--- a/libs/media/src/lib/directives/image-reference/image-reference.directive.ts
+++ b/libs/media/src/lib/directives/image-reference/image-reference.directive.ts
@@ -33,7 +33,6 @@ export class ImageReferenceDirective implements OnInit, OnDestroy {
 
   /** the image to display */
   @Input() set ref(image: string) {
-    if (!image) return;
     this.ref$.next(image);
   }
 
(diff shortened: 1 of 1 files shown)


phpgeekfr:
Did you try to upload a new image like profile picture or org logo ? 

This `if (!image) return` seems to be here since a while..
Can you double check with @pldespaigne  ?

RemcoSimonides:
> Did you try to upload a new image like profile picture or org logo ?
> 
> This `if (!image) return` seems to be here since a while..
> Can you double check with @pldespaigne ?

Yea, I did try that and it's still working. The only thing I noticed is that 5000 delay wasn't enough for larger image uploads and slow internet 🤔

pldespaigne (approved):

phpgeekfr:
> The only thing I noticed is that 5000 delay wasn't enough for larger image uploads and slow internet

yes me too.. But 10 sec will be to much I think. We need to find another solution to prevent this but I have no clue for now

phpgeekfr (approved):

PR #4644, opened 2021-01-04
Event invitation emails

### LINKED ISSUE #4142: Update event invitation emails
Templates have been updated on Sendgrid, there is new data to load and this is the most important email of this app so please test it before enabling it. @Jackseed should confirm the PR before merge. 

The 2 templates are:
- Organization invites user to attend event (`Template ID: d-1a7cc9ca846c4ae1b4e8cf8a76455cc5`)
- Invited user to event with credentials (Archipel Market) (`Template ID: d-ce3e57248a694cefacad49bc4c820078`)

New data to implement:
- sessionURL (event session page)
- eventStartDate
- eventStartTime
- eventEndDate
- eventEndTime

Be careful there are multiple urls in these emails. 
The last version created on Sendgrid should be set to "Active" when this issue is done & validated.

Mussuranae:
Rework 2 email templates when organization invites someone to its event, someone with an account or not.
Fix #4142


apps/backend-functions/src/invitation.ts
```diff
@@ -137,14 +138,14 @@ That would have exceeded the current limit which is ${MEETING_MAX_INVITATIONS_NU
       }
     }
 
-    let eventName = '';
+    let eventData: EventEmailData = getEventEmailData();
     if (invitation.type === 'attendEvent' && !!invitation.eventId) {
```
phpgeekfr, line 142: if invitation.type is joinOrg, `eventData` will not be used in template right ?

apps/backend-functions/src/templates/mail.ts
```diff
@@ -68,14 +68,17 @@ export function userInvite(
...
     userEmail: email,
     userPassword: password,
     orgName,
-    eventName,
-    pageURL: `${pageURL}${USER_CREDENTIAL_INVITATION}`
+    eventName: eventData.title,
+    eventStartDate: eventData.start,
+    eventEndDate: eventData.end,
+    pageURL: `${pageURL}${USER_CREDENTIAL_INVITATION}`,
+    sessionURL: `${pageURL}/c/o/marketplace/event/${eventData.id}`
```
phpgeekfr, line 81: what append if `eventData` is undefined ? add some test maybe

phpgeekfr (approved):

apps/backend-functions/src/invitation.ts
Mussuranae: yes, the function will received an empty object, but depending on the type of the invitation, it's not the same template that will be used and the object isn't used in the template.
`backend-functions/internals/users.ts line 41-43`
image

apps/backend-functions/src/templates/mail.ts
Mussuranae: Isn't it better to check if the event exists ? And if not, return an error ?

phpgeekfr: > And if not, return an error ?

No because this method is also used for joinOrg and there is no event for this.

I was thinking of something like `eventName: eventData.title || '' ,`

Mussuranae: Ok I will do that !

PR #4662, opened 2021-01-06
Issue tunnel clean up

### LINKED ISSUE #3908: Factorize the confirmExit in shell

fritzschoff:
had to move the config interfaces to different file cause of circular dependency
close #3908


apps/festival/festival/src/app/dashboard/tunnel/summary/summary.component.ts
```diff
@@ -32,18 +33,18 @@ export class TunnelSummaryComponent implements OnInit, OnDestroy {
...
 
-  ngOnDestroy(): void {
+  ngOnDestroy() {
     this.subscription.unsubscribe();
   }
 
   public async submit() {
     if (this.form.valid) {
-      await this.shell.update({ publishing: true });
+      await this.shell.layout.update(true);
```
GrandSchtroumpf, line 47: keep the option format. Here it's very difficult what the "true" is doing.

libs/campaign/src/lib/form/campaign.shell.config.ts
```diff
@@ -27,7 +27,7 @@ export class CampaignShellConfig implements FormShellConfig<CampaignControls, Ca
     return [sub];
   }
 
-  async onSave(options: FormSaveOptions): Promise<any> {
+  async onSave(publishing: boolean): Promise<any> {
```
GrandSchtroumpf, line 30: keep option

libs/movie/src/lib/movie/form/shell/shell.component.html
```diff
@@ -2,13 +2,26 @@
...
+      <mat-icon svgIcon="floppy"></mat-icon>
       <span>Save</span>
     </button>
   </header>
 
   <aside>
     <tunnel-nav [steps]="steps$ | async"></tunnel-nav>
   </aside>
+  <ng-template #confirmExit>
+    <div class="confirm-exit" fxLayout="column" fxLayoutAlign="start center">
```
GrandSchtroumpf, line 15: You can use this style : 
https://material.angular.io/components/dialog/overview#dialog-content

GrandSchtroumpf: Also it doesn't feel like we are improving the structure, the goal was get rid of repeating confirm method, now we need to repeat the full component...

libs/ui/src/lib/tunnel/layout/layout.component.ts
```diff
@@ -97,6 +125,54 @@ export class TunnelLayoutComponent implements OnInit, OnDestroy {
     this.previous = getPage(this.steps, url, -1);
   }
 
+  /** Save the form and display feedback to user */
+  async save() {
+    await this.update(false);
+    await this.snackBar.open('Title saved', '', { duration: 1000 }).afterDismissed().toPromise();
```
GrandSchtroumpf, line 131: You cannot do that. Tunnel will b use for other things than title

libs/ui/src/lib/tunnel/layout/layout.component.ts
```diff
@@ -97,6 +125,54 @@ export class TunnelLayoutComponent implements OnInit, OnDestroy {
...
+    if (publishing) {
+      for (const name in this.configs) {
+        const form = this.getForm(name as any);
+        if (form.invalid) {
+          const fields = findInvalidControls(form);
+          throw new Error(`Form "${name}" should be valid before publishing. Invalid fields are: ${fields.join()}`);
+        }
+      }
+    }
+    for (const name in this.configs) {
```
GrandSchtroumpf, line 171: Could you use Promise.all ?

GrandSchtroumpf (changes requested):

libs/ui/src/lib/tunnel/layout/layout.component.ts
```diff
@@ -97,6 +125,54 @@ export class TunnelLayoutComponent implements OnInit, OnDestroy {
     this.previous = getPage(this.steps, url, -1);
   }
 
+  /** Save the form and display feedback to user */
+  async save() {
+    await this.update({ publishing: false });
+    const configNames = Object.keys(this.configs).map(key => key.charAt(0).toUpperCase().concat(key.substring(1))).join(' & ');
```
GrandSchtroumpf, line 131: It's a little bit difficult to understand what it does. Also I fear it won't work with future keys of config. Maybe it would be better to just not have config names... Or having a Record<key, label>

GrandSchtroumpf (approved):

libs/ui/src/lib/tunnel/layout/layout.component.ts
fritzschoff: can add something

Issue #3580, opened 2020-08-28
Support for Safari version 12.1 and older


RemcoSimonides:
A decision needs to be made which versions of Safari we still support and which not. Apple is doing a good job on forcing users to upgrade to newer versions of Safari but there may still be some users left.

Vincent looked into the analytics to see which version of Safari our users are using, but he wasn't able to. We could get this insight if we add code to send browser and browser version to Firebase analytics.

Suggestions to tackle this issue:
- [ ] Edit warning for Safari users if they are on version 12.1 or lower to upgrade their Safari version
- [ ] Deny access for Safari users if they are on version 12.1 or lower

Open bugs related to older versions of Safari
- [ ] Screening List not displaying events eventhough there should be
- [ ] Calendar icon from marketplace navbar doesn't work ([flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) method isn't supported)

RemcoSimonides:
A source for a working method to get the browser including browser version https://stackoverflow.com/questions/5916900/how-can-you-detect-the-version-of-a-browser

fritzschoff:
@RemcoSimonides in you rlast PR you tagged this issue as fixed. Close it?

RemcoSimonides:
No sorry, that wasn't a fix. That's just to get insight in which browser and version our users are using. Dependent on the results from that we decide what we do with this issue ;)

fritzschoff:
I mean this issue is a year old. We sure we want to support 12.1? Since this will become deprecated pretty quickly

RemcoSimonides:
What do you mean deprecated? Safari v12?
The users need to update their Safari version, right?

But this issue is indeed about whether we should support Safari v12, or show a big red warning that they should upgrade their browser before continueing

fritzschoff:
currently we are on safari 14. Until we will support safari 12.1 this will become not relevant anymore, I guess

RemcoSimonides:
Jep, but we never decided on which version we still support and which we don't

fritzschoff:
@Jackseed here

Jackseed:
Thanks as told during the sprint launch @SamirTouira  we can conclude that issue by just removing the footer banner for Safari users. Safari is doing a good job at updating their browser version so let's rely on them and we just won't support Safari 12.xx & lower
[...]

PR #4666, opened 2021-01-06
3580 Save browser name and version in analytics

RemcoSimonides:
Fix issue #3580 

Adding browser name and browser version to Google Analytics data. Based on the gathered data we can determine which versions we should support or not.


apps/catalog/catalog/src/app/app.module.ts
```diff
@@ -115,6 +116,9 @@ export class AppModule {
     if (yandex) yandexService.insertMetrika();
     intercom && intercomId ? intercomService.enable() : intercomService.disable();
 
+    const browser = getBrowserWithVersion();
+    analytics.setUserProperties({ browser_name: browser.name, browser_version: browser.version });
```
GrandSchtroumpf, line 120: If `browser_name` defined by Google Analytics ? If yes, maybe modify the `getBrowerWithVersion` to return the same naming : 
```typescript
analytics.setUserProperties(getBrowserWithVersion());
```

libs/utils/src/lib/analytics/app-analytics.ts
```diff
@@ -29,4 +29,8 @@ export class FireAnalytics {
       this.analytics.logEvent(name, { ...params });
     }
   }
+
+  public setUserProperties(properties: Record<AnalyticsUserProperties, any>) {
```
GrandSchtroumpf, line 33: This is a weird way to type. Better define the type directly in an interface : 
```typescript
interface AnalyticsUserProperties {
  browser_name: string;
  browser_version: string;
}
```

libs/utils/src/lib/browser/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+    tem = /\brv[ :]+(\d+)/g.exec(ua) || [];
+    return { name:'IE',version:(tem[1] || '') };
+  }
+  if (M[1] === 'Chrome') {
+    tem = ua.match(/\b(OPR|Edge)\/(\d+)/);
+    if (tem != null) return { name:tem[1].replace('OPR', 'Opera'), version:tem[2] };
+  }
+  M = M[2] ? [M[1], M[2]] : [navigator.appName, navigator.appVersion, '-?'];
+  if ((tem = ua.match(/version\/(\d+)/i)) != null) M.splice(1, 1, tem[1]);
+  return { name:M[0], version:M[1] };
```
GrandSchtroumpf, line 35: Return `{ browser_name: M[0], browser_version: M[1] }`

GrandSchtroumpf (approved):

PR #4670, opened 2021-01-06
Add app version

### LINKED ISSUE #1219: Add app versionning (in the footer for example)
It would be great to have a version number for the app. Maybe increment automatically at each new push to master ?

RemcoSimonides:
Fix #1219

### DIFF
--- a/libs/ui/src/lib/maintenance/maintenance.service.ts
+++ b/libs/ui/src/lib/maintenance/maintenance.service.ts
@@ -20,4 +20,10 @@ export class MaintenanceService {
       tap(_ => window.location.reload())
     )
   }
+
+  getAppVersion() {
+    return this.db.doc<{ currentVersion: number }>(`${META_COLLECTION_NAME}/_VERSION`).valueChanges().pipe(
+      map(data => data.currentVersion)
+    )
+  }
 }
--- a/libs/ui/src/lib/maintenance/maintenance.service.ts
+++ b/libs/ui/src/lib/maintenance/maintenance.service.ts
@@ -20,10 +20,4 @@ export class MaintenanceService {
       tap(_ => window.location.reload())
     )
   }
-
-  getAppVersion() {
-    return this.db.doc<{ currentVersion: number }>(`${META_COLLECTION_NAME}/_VERSION`).valueChanges().pipe(
-      map(data => data.currentVersion)
-    )
-  }
 }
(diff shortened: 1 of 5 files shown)


libs/ui/src/lib/maintenance/maintenance.service.ts
```diff
@@ -20,4 +20,10 @@ export class MaintenanceService {
       tap(_ => window.location.reload())
     )
   }
+
+  getAppVersion() {
+    return this.db.doc<{ currentVersion: number }>(`${META_COLLECTION_NAME}/_VERSION`).valueChanges().pipe(
```
RemcoSimonides, line 25: interface `IVersionDoc` is in `backend-ops/src/maintance` so I can't access it here. Should I move that interface?

Also, I'm not sure if I should add this function to `maintenance.service.ts`, I could directly add it in `widget.component.ts` instead?

What do you think?

libs/user/src/lib/auth/components/widget/widget.component.html
```diff
@@ -8,7 +8,7 @@
       <button mat-icon-button [widgetTarget]="menu" test-id="profile-avatar" test-id="auth-user-avatar">
         <img asset="profil_user.webp" [ref]="user.avatar" alt="Profile picture">
       </button>
-      <span class="mat-caption bf-admin" *ngIf="isBfAdmin | async">Blockframes Admin</span>
+      <span class="mat-caption bf-admin" *ngIf="isBfAdmin | async">Blockframes Admin - v{{ appVersion$ | async }}</span>
```
phpgeekfr, line 11: actually, it's not app version (v2.0) that you are displaying but db version (47). Can you mention this ?

libs/ui/src/lib/maintenance/maintenance.service.ts
phpgeekfr: Yes, can you move `IVersionDoc` to `libs/utils/src/lib/maintenance/index.ts` ?

phpgeekfr: >Also, I'm not sure if I should add this function to maintenance.service.ts, I could directly add it in widget.component.ts instead?

As you want, if you think that we are not going to use this elsewere you can move it

phpgeekfr: also there is `loadDBVersion` in  `apps/backend-ops/src/migrations.ts` can you check if there is some refactoring to do ?
or at least, update it to use `META_COLLECTION_NAME` instead of plain `_META`

phpgeekfr: and maybe create `export const DB_DOCUMENT_NAME = '_VERSION';` and use it.
or you can even use `export const dbVersionDoc = `${META_COLLECTION_NAME}/${DB_DOCUMENT_NAME}`;`

phpgeekfr (approved):

PR #4674, opened 2021-01-07
Logo "Screen International" leads to the marketplace home

SamirTouira:
Fixed

### DIFF
--- a/apps/festival/festival/src/app/marketplace/marketplace.component.html
+++ b/apps/festival/festival/src/app/marketplace/marketplace.component.html
@@ -38,7 +38,7 @@
   <header fxFlex fxLayout fxLayoutAlign="space-between center" class="marketplace-header">
     <div class="festival-logo" fxLayout>
       <mat-divider vertical></mat-divider>
-      <a routerLink="home" routerLinkActive="active">
+      <a routerLink="home">
         <img src="assets/images/screen-logo.png" />
       </a>
     </div>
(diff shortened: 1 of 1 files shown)


phpgeekfr (approved):

apps/festival/festival/src/app/marketplace/marketplace.component.html
```diff
@@ -38,11 +38,13 @@
   <header fxFlex fxLayout fxLayoutAlign="space-between center" class="marketplace-header">
     <div class="festival-logo" fxLayout>
       <mat-divider vertical></mat-divider>
-      <img src="assets/images/screen-logo.png" />
+      <a routerLink="home" routerLinkActive="active">
```
GrandSchtroumpf, line 41: Remove routerLinkActive, there is no reason for that I think.
Did you check the UI is still working as expected ? height & width ?

GrandSchtroumpf (approved):

apps/festival/festival/src/app/marketplace/marketplace.component.html
SamirTouira: understood, I remove this
sorry what do you mean by height & width ?

SamirTouira: oh, yes, I measured and we've still got the same size if that's what you wanted to know

PR #4675, opened 2021-01-07
Preview and videos

### LINKED ISSUE #2658: Use JW Player for promotional videos
## Description
Now that we have our own in-app video player, we should use it to display promotional videos like trailers, promo reels,... instead of using external urls. 
Still, both systems should co-exist as it shouldn't block our uers if they don't have the file available. 

## Acceptance criteria
- [ ] UX updated to add video player on marketplace and file management on dashboard
- [ ] Update the db accordingly
- [ ] Implement wireframes

pldespaigne:
- [x] preview modal handle any kind of medias (images, pdf, videos)
- [x] fix #2658

### DIFF
--- a/libs/media/src/lib/components/viewers/media-viewer.component.html
+++ b/libs/media/src/lib/components/viewers/media-viewer.component.html
@@ -12,7 +12,7 @@
         <event-video-viewer [eventId]="event.id" [ref]="event.meta.selectedFile" [control]="event.meta.controls[event.meta.selectedFile]" ></event-video-viewer>
       </ng-container>
       <ng-container *ngSwitchDefault>
-        <article class="noFile">
+        <article>
           <h3>Sorry</h3>
           <img asset="no_titles.webp" alt="Flying saucer">
           <p>We don't know how to display this file.</p>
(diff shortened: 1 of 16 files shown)


libs/media/src/lib/components/viewers/media-viewer.component.html
```diff
@@ -12,7 +12,7 @@
         <event-video-viewer [eventId]="event.id" [ref]="event.meta.selectedFile" [control]="event.meta.controls[event.meta.selectedFile]" ></event-video-viewer>
       </ng-container>
       <ng-container *ngSwitchDefault>
-        <article class="noFile">
```
fritzschoff, line 15: if you removed the class, update the css as well pls

libs/movie/src/lib/movie/form/summary/main/main.component.ts
```diff
@@ -15,7 +15,7 @@ export class SummaryMainComponent {
     return this.movie.get('title');
   }
 
-  public get genres() {
```
fritzschoff, line 18: lol nice, good catch

fritzschoff (approved):

libs/media/src/lib/components/viewers/media-viewer.component.html
pldespaigne: this class was not present in the css, that's why I deleted it

PR #4691, opened 2021-01-08
added view count

### LINKED ISSUE #4677: Bug analytics - View counts on dashboard film list not working
Dashboard film list displays every film with #View column to 0 though it should have views. 
This number is supposed to show the total view count of each film view page since it's available on the marketplace.

fritzschoff:
close #4677

### DIFF
--- a/apps/festival/festival/src/app/dashboard/title/list/list.component.ts
+++ b/apps/festival/festival/src/app/dashboard/title/list/list.component.ts
@@ -2,11 +2,12 @@ import { Component, ChangeDetectionStrategy, OnInit, OnDestroy } from '@angular/
 import { Router, ActivatedRoute } from '@angular/router';
 import { FormControl } from '@angular/forms';
 import { startWith, map, switchMap, tap } from 'rxjs/operators';
-import { Observable, Subscription } from 'rxjs';
-import { Movie } from '@blockframes/movie/+state/movie.model';
+import { combineLatest, Observable, Subscription } from 'rxjs';
+import { getMovieTotalViews, Movie, MovieAnalytics } from '@blockframes/movie/+state/movie.model';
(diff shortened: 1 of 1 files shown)


apps/festival/festival/src/app/dashboard/title/list/list.component.ts
```diff
@@ -37,16 +45,17 @@ export class ListComponent implements OnInit, OnDestroy {
...
 
   ngOnInit() {
     this.sub = this.service.valueChanges(fromOrg(this.orgQuery.getActive().id)).pipe(
       switchMap(movies => this.service.syncWithAnalytics(movies.map(m => m.id)))
     ).subscribe();
-
-    this.titles$ = this.service.valueChanges(fromOrg(this.orgQuery.getActive().id)).pipe(
-      map(movies => movies.filter(movie => !!movie)),
-      map(movies => movies.filter(movie => movie.storeConfig.appAccess.festival)),
+    const movieAnalytics$ = this.movieQuery.analytics.selectAll();
```
GrandSchtroumpf, line 55: It's really not optimal. We should get analytics from firestore instead (#4693), but in the meantime you can do that: 
```typescript
const orgId = this.orgQuery.getActive().id;
const titles$ = this.service.valueChange(fromOrg(orgId)).pipe(shareReplay(1));
const analytics$ = titles$.pipe(
  map(movies => movies.map(m => m.id)),
  distuntUntilChanged((a, b) => a.length === b.length), // good approximation
  switchMap(movieIds => getAnalytics(movieIds)),
  startWith([]) // we shouldn't block rendering while waiting for the analytics
);
this.title$ = combineLatest([titles$, analytics$]).pipe(
  map(([movies, analytics]) => createTitleList(movies, analytics))
);
```

GrandSchtroumpf (changes requested):

PR #4692, opened 2021-01-08
Feat 2296

### LINKED ISSUE #2296: De-indexed staging and demo from Google search engin

phpgeekfr:
fix #2296

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -228,6 +228,7 @@ jobs:
       - slack/notify:
           message: '🚚 Deploying to production...'
       - prepare_environment
+      - replace_prod_files
       - build_deploy_upgrade
   deploy-to-staging:
     executor: blockframes-ci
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -81,6 +81,12 @@ commands:
       - checkout_project
       - prepare_gcloud
       - run: cp $ENV_TS ./env/env.ts
+  replace_prod_files:
```
george43g, line 84: hey man, this seems like the perfect thing that angular.json configurations are used for. What if there's a manual deploy without CI? There's a special way you can use fileReplacements to do this I believe. And I think that's definitely the way it should be done if possible.

george43g (changes requested):

.circleci/config.yml
phpgeekfr: > What if there's a manual deploy without CI
Indeed, we will have to remember to replace file manually if we need to deploy without CI

george43g (approved):

Issue #4029, opened 2020-10-22
Clean script improvement - Check wishlists for undefined movie


Jackseed:
Small improvement of the clean script would be to check the wishlist for:
- draft movies (if the movie went from published to draft)
- deleted movies
And remove the movieId if that matches

PR #4715, opened 2021-01-12
Clean script improvement - Check wishlists for undefined movie #4029

phpgeekfr:
This PR adds new features to db-cleaning script to remove deleted or not accepted movies from orgs wishlist.

Updated movies (updated from accepted status to draft for example) will also be removed from orgs wishlist.

It also update unit tests.


apps/backend-ops/src/db-cleaning.ts
```diff
@@ -226,13 +227,22 @@ export function cleanOrganizations(
       await orgDoc.ref.set(org);
     }
 
-    const { userIds } = org as OrganizationDocument;
+    const { userIds, wishlist } = org as OrganizationDocument;
```
GrandSchtroumpf, line 230: Set the default values here, it would be cleaner I think : 
```typescript
const { userIds = [], wishlist = [] } = org as ...:
```

GrandSchtroumpf (approved):

Issue #4704, opened 2021-01-12
Caroussel showing only 4 items total instead of more


RemcoSimonides:
Uploaded 6 still images but only 4 are shown
![caroussel bug](https://user-images.githubusercontent.com/27687382/104293205-0d625680-54be-11eb-82e3-17e6f5c8a502.gif)

PR #4721, opened 2021-01-13
Fix carousel

RemcoSimonides:
Fix issue #4704

### DIFF
--- a/libs/media/src/lib/directives/image-reference/image-reference.directive.ts
+++ b/libs/media/src/lib/directives/image-reference/image-reference.directive.ts
@@ -25,7 +25,7 @@ export class ImageReferenceDirective implements OnInit, OnDestroy {
   @HostBinding('srcset') srcset: string;
   @HostBinding('src') src: string;
   @HostBinding('alt') alt: string;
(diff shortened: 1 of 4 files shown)


libs/media/src/lib/directives/image-reference/image-reference.directive.ts
```diff
@@ -25,7 +25,7 @@ export class ImageReferenceDirective implements OnInit, OnDestroy {
   @HostBinding('srcset') srcset: string;
   @HostBinding('src') src: string;
   @HostBinding('alt') alt: string;
-  @HostBinding('loading') loading = 'lazy';
+  @HostBinding('loading') _loading: 'lazy' | 'eager' = 'lazy';
```
pldespaigne, line 28: You are trying to battle the `image-reference`'s hostBinding here. I'm afraid it could result in inconsistent behavior sometimes.
I think it would be better to add an optional input `loading` to the image reference. It's value should be 'lazy' by default but can be changed to 'eager' when needed.

pldespaigne (changes requested):

pldespaigne (approved):

PR #4730, opened 2021-01-13
Feat 2277 - Clean how models/interfaces are declared in front

### LINKED ISSUE #2277: Clean how models/interfaces are declared in front
We should have something like :

.firestore
```
export interface MovieBase<D extends Timestamp | Date> {...}
export type MovieDocument = MovieBase<Timestamp>
```
.model
```
import { MovieBase } from './***.firestore'
export interface Movie extends MovieBase<Date> {
  specificFrontEndStuffIfNeeded;
}
```

To avoid switching between 3 files to find  "MovieDocumentWithDate"

phpgeekfr:
fix #2277


libs/notification/src/lib/+state/notification.firestore.ts
```diff
@@ -39,11 +40,11 @@ export interface NotificationOptions {
   organization?: PublicOrganization;
   /** @dev Type of the notification */
   type: NotificationType;
-}
-
-/** Generic informations for a Notification. */
-export interface NotificationDocument<D extends Date | firestore.Timestamp = firestore.Timestamp> extends NotificationOptions {
```
phpgeekfr, line 45: @GrandSchtroumpf I don't understand why `NotificationDocument` was previously declared with `<D extends Date | firestore.Timestamp = firestore.Timestamp>` instead using firestore.Timestamp only .  Do you have a hint ?
This is only used in backend so it should always handle only timestamps

phpgeekfr:
I did not touch `distributionRights` since it is going to be reworked from scratch

libs/notification/src/lib/+state/notification.firestore.ts
GrandSchtroumpf: Notifications are only used in the backend ? In the frontend we don't do something like that ?
```typescript
interface Notification extends NotificationDocument<Date> {}
```
If not then you can force it to Timestamp

phpgeekfr: NotificationDocument are only used in backend and  for front we have `export interface Notification extends NotificationBase<Date> {`

fritzschoff (approved):
lgtm

GrandSchtroumpf (approved):

Issue #3208, opened 2020-06-26
[Festival] Sales Agent filter - mat-autocomplete


Jackseed:
## Description
Page `/c/o/marketplace/title`
Sales agent filter should use a mat-autocomplete in order for the user to see the list of all the sales agents when he/she clicks on the input (same as genre autocomplete). 
We need to use `algolia-autocomplete` with the organization index.

![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/a7c446b2-6a80-4121-ab05-aa4c2c015e7c)
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/de77ffa9-bccb-425d-a79f-0d7d2c87ea99)

## Acceptance criteria
- [ ] Add a mat-autocomplete list on the filter Sales Agents
- [ ] List the sales agents by alphabetic order
- [ ] Be sure that Archipel Content (= central org id) is not listed in the sales agents

GrandSchtroumpf:
Investigate why we have "cascade8" in the result.
Exclude organization from unsubmitted movie.
It's not easy to use a select instead of an autocomplete. So let's not do it.

Jackseed:
@GrandSchtroumpf any recommandation so we may achieve this one day?

GrandSchtroumpf:
We have a algolia-autocomplete in the event form, we could take this example. I think the main difficulty is going to filter the organization to only display sellers of the specific app...

Mussuranae:
I look for this issue, and we use the `algolia-chips-autocomplete` component, so it works well. The only 'problem' I see is that we have to write the first letter to see a list of organization appears.
Also, it looks like the filter show only organizations that are accepted and that have movies accepted.

Jackseed:
@Mussuranae a simple fix for this issue would be to display the whole list when the user selects the field and then filter it if s/he types letters. Do you think that's doable?

Mussuranae:
No idae at all. I can look at it tomorrow.

Jackseed:
Yes please have a look it would be a good improvement since it's a filter common to all apps

RemcoSimonides:
@Jackseed Should this 'show autocomplete results on focus' be implemented everywhere where algolia-chips-autocomplete is implemented? Or only for Sales Agent Filter

Everywhere is:
- Sales Agent filter
- Invite users on event

Jackseed:
[...]

PR #4743, opened 2021-01-14
Show autocomplete results on focus

RemcoSimonides:
Awaiting confirmation from Vincent
Fix issue #3208


RemcoSimonides:
Component will get too difficult to maintain if too many features are added. Therefore we close the PR & issue.

PR #4744, opened 2021-01-14
Productionstatus notdisplayed

SamirTouira:
The Steps and currentStep component had a wrong display when you initialize the first time to the movie-tunnel when we want Edit a movie (First step in movie-tunnel, Production Status and buttons Next and Previous weren't displayed)

Fixed with Clélia helps


libs/movie/src/lib/movie/form/shell/shell.component.ts
```diff
@@ -126,9 +126,8 @@ export class MovieFormShellComponent implements TunnelRoot, OnInit, AfterViewIni
     const subs: Observable<any>[] = Object.values(this.configs).map((config: any) => config.onInit()).flat();
     this.sub = combineLatest(subs).subscribe(() => this.cdr.markForCheck());
     this.exitRoute = `/c/o/dashboard/title/${this.query.getActiveId()}`;
-  }
 
-  ngAfterViewInit() {
```
fritzschoff, line 131: that easy? The last time I touched that and why we moved that into ngAfterViewInit it was because of the routes not getting rendered in time. But I checked out your branch and it is working. 
Anyway it is working so good job! 
Maybe keep in mind if you encounter another issue in the movie tunnel that maybe data is undefined but it should defined

fritzschoff (approved):

Issue #4734, opened 2021-01-13
Sort line-up movies by _meta.createdAt (org page)


Jackseed:
Page: `/c/o/marketplace/organization/orgId/title`

Movies should be sorted by their `_meta.createdAt` parameter on that page as most of the time the last movies created are the most important for Sales Agents & Partners (MF).

PR #4746, opened 2021-01-14
4734 sort lineup by createdAt

RemcoSimonides:
Fix issue #4734

### DIFF
--- a/apps/festival/festival/src/app/marketplace/event/list/list.component.ts
+++ b/apps/festival/festival/src/app/marketplace/event/list/list.component.ts
@@ -32,8 +32,7 @@ export class ListComponent implements OnInit {
 
     this.events$ = combineLatest(events$, orgIds$).pipe(
       map(([ events, orgs ]) => this.filterByOrgIds(events, orgs.map(org => org.objectID))),
-     // We can't filter by meta.titleId directly in the query because
-     // it will neglect the function of .startAt(new Date()) to only query events in the future
+     // We can't filter by meta.titleId directly in the query because range and not equals comparisons must all filter on the same field
       map(events => events.filter(event => !!event.meta.titleId)),
       tap(events => this.setTitle(events.length))
     )
--- a/apps/festival/festival/src/app/marketplace/event/list/list.component.ts
+++ b/apps/festival/festival/src/app/marketplace/event/list/list.component.ts
@@ -27,11 +27,14 @@ export class ListComponent implements OnInit {
 
   ngOnInit() {
     const orgIds$ = this.searchForm.valueChanges.pipe(startWith(this.searchForm.value));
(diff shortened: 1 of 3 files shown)


apps/festival/festival/src/app/marketplace/event/list/list.component.ts
```diff
@@ -27,11 +27,14 @@ export class ListComponent implements OnInit {
...
     const orgIds$ = this.searchForm.valueChanges.pipe(startWith(this.searchForm.value));
-    const query = ref => ref.where('meta.titleId', '!=', '').orderBy('meta.titleId').orderBy('end').startAt(new Date());
+    const query = ref => ref.orderBy('end').startAt(new Date());
     const events$ = this.service.queryByType(['screening'], query);
 
     this.events$ = combineLatest(events$, orgIds$).pipe(
       map(([ events, orgs ]) => this.filterByOrgIds(events, orgs.map(org => org.objectID))),
+     // We can't filter by meta.titleId directly in the query because
+     // it will neglect the function of .startAt(new Date()) to only query events in the future
+      map(events => events.filter(event => !!event.meta.titleId)),
```
RemcoSimonides, line 37: I brought this one back because the query with `meta.titleId, '!=', ''` destroys the function of `.startAt(new Date())`
`startAt()` works together with `orderBy()`, but I had to make `orderBy('meta.titleId')` the _first_ orderBy.
Trying to fix this with `startAt(null, new Date())` or `startAt('', new Date())` didn't solve it.

firestore.indexes.json
```diff
@@ -19,11 +19,12 @@
     "collectionGroup": "movies",
     "queryScope": "COLLECTION",
     "fields": [
+      { "fieldPath": "orgIds", "arrayConfig": "CONTAINS" },
       { "fieldPath": "storeConfig.appAccess.festival", "order": "ASCENDING" },
       { "fieldPath": "storeConfig.status", "order": "ASCENDING" },
-      { "fieldPath": "release.year", "order": "DESCENDING" }
```
RemcoSimonides, line 24: I'm quite sure we don't use this index anymore. Looking at git blame you were the one who added it. Can you confirm?
It's probably still in there because we replaced it with `_meta.createdAt` later

GrandSchtroumpf: We are using it in the CMS. Please keep and and don't add the orgIds like that.
We need one index per query. If you add one "where" in the compound query, then we need to create a full new index

GrandSchtroumpf (changes requested):

apps/festival/festival/src/app/marketplace/event/list/list.component.ts
GrandSchtroumpf: Ok, change the comment for the documentation : 
"In a compound query, range (<, <=, >, >=) and not equals (!=) comparisons must all filter on the same field."

GrandSchtroumpf (approved):

PR #4747, opened 2021-01-15
Clean asset

Mussuranae:
Delete images, icons that are not used anymore.
I need to work with Stéphanie to continue the cleaning.


phpgeekfr:
@Mussuranae is this still in progress or can we close ?

Mussuranae:
Still in progress, but I need to wait the trainee to continue the cleaning, maybe we can close and I will reopen one when I work again on that ?

Issue #4733, opened 2021-01-13
Homepage prod status carousels - Add random order


Jackseed:
Movie carousels sort by production status on the homepage haven't change since months because movies are sorted by movieIds.
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/4c1ddd3a-c7ad-4235-90c6-53849e37c8fd)

It doesn't reflect the app activity and that's a pity as we had a lot of new films since. On the other hand we don't want to sort these movies by createdAt parameter as it's also a chance for old movies to be highlighted. 

A good compromise would be to add a random order to these movies. Random should be calculated client-side so each time users reload the page they will have different movies. 
This will give them an extensive look at movies available in the app while give to every movie their chance. 

This random sorting will most likely be used on future carousels so it should be made to be re-usable and as a sort possibility in the CMS.

PR #4749, opened 2021-01-15
4733 Add random sorting on titles

RemcoSimonides:
Fix issue #4733 for movie titles

### DIFF
--- a/apps/festival/festival/src/app/marketplace/home/home.component.html
+++ b/apps/festival/festival/src/app/marketplace/home/home.component.html
@@ -47,7 +47,7 @@
       </a>
     </header>
     <bf-carousel flex>
-      <ng-container *ngFor="let movie of section | queryTitles | async | sortArrayBy:section.sorting">
+      <ng-container *ngFor="let movie of section | queryTitles | async | sortBy:section.sorting">
         <movie-card [movie]="movie" [size]="section.mode" [link]="['../title', movie.id]" [col]="3" carouselItem>
         </movie-card>
       </ng-container>
--- a/apps/festival/festival/src/app/marketplace/home/home.component.html
+++ b/apps/festival/festival/src/app/marketplace/home/home.component.html
@@ -47,7 +47,7 @@
       </a>
     </header>
     <bf-carousel flex>
-      <ng-container *ngFor="let movie of section | queryTitles | async">
+      <ng-container *ngFor="let movie of section | queryTitles | async | sortArrayBy:section.sorting">
         <movie-card [movie]="movie" [size]="section.mode" [link]="['../title', movie.id]" [col]="3" carouselItem>
         </movie-card>
       </ng-container>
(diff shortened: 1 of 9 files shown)


apps/festival/festival/src/app/marketplace/home/home.component.html
```diff
@@ -47,7 +47,7 @@ <h2>{{ section.title }}</h2>
       </a>
     </header>
     <bf-carousel flex>
-      <ng-container *ngFor="let movie of section | queryTitles | async">
+      <ng-container *ngFor="let movie of section | queryTitles | async | sortArrayBy:section.sorting">
```
GrandSchtroumpf, line 50: Rename `sortArrayBy` into `sortBy`. We can only sort list anyway

libs/utils/src/lib/pipes/sort-array.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type SortingOptions = 'default' | 'random';
+export const sortingOptions: SortingOptions[] = ['default', 'random'];
+
+@Pipe({ name: 'sortArrayBy' })
+export class SortArrayByPipe implements PipeTransform {
+  transform(array: any[], method: SortingOptions) {
+    if (Array.isArray(array)) {
+      if (method === 'random') {
+        for (let i = array.length - 1; i > 0; i--) {
```
GrandSchtroumpf, line 11: Use the `array.sort((a, b) => )` method : 
```typescript
const sortFn = {
  random: (a, b) => Math.random() - Math.random() // or  Math.random() - 0.5
}
export type SortingOption = 'default' | keyof typeof sortFn;

@Pipe({ name: 'sortBy' })
export class SortByPipe implements PipeTransform {
  transform(array: any[], method: SortingOptions = 'default') {
    if (method === 'default' || !Array.isArray(array)) return array;
    return array.sort(sortFn[method]);
  }
}
```

GrandSchtroumpf (changes requested):

libs/utils/src/lib/pipes/sort-array.pipe.ts
RemcoSimonides: image

pldespaigne: Hum interesting 🤔 
More details in the "solution" of [this page](https://javascript.info/task/shuffle)

TL;DR :
`.sort(() => .5 Math.random())` is bad if you use it on large inputs or if you want to have an evenly distributed results over many iterations.

IMO our use case is simple enough to use `.sort(() => .5 Math.random())`, moreover because performance doesn't seems to be an issue in this case, I think we should favor simplicity of code.

GrandSchtroumpf (approved):

libs/utils/src/lib/pipes/sort-array.pipe.ts
GrandSchtroumpf: lol. yes we just want an insignificant shuffle.

PR #4750, opened 2021-01-15
OrgId check for invitation

### LINKED ISSUE #739: When a user is added to an org, clear other invitations

Mussuranae:
Check if user invited has already an orgId when we want to invite him/her on our organization
Fix #739

### DIFF
--- a/apps/backend-functions/src/invitation.ts
+++ b/apps/backend-functions/src/invitation.ts
@@ -196,14 +196,9 @@ export async function hasUserAnOrgOrIsAlreadyInvited(userEmails: string[]) {
   .where('email', '==', email)
   .get());
   const userQuery = await Promise.all(userPromises);
-  const orgIdExists = [];
-  userQuery.forEach(d => d.docs.forEach(u => {
-    const data = u.data();
-    orgIdExists.push(!!data.orgId);
-    return orgIdExists;
-  }));
 
-  const isUserOrgIdExist = orgIdExists.some(d => !!d);
+  const hasUserAlreadyAnOrg = userQuery.some(d => d.docs && d.docs.some(u => !!u.data().orgId));
+  if(hasUserAlreadyAnOrg) return true;
 
   const invitationPromises = userEmails.map(email => db.collection('invitations')
     .where('type', '==', 'joinOrganization')
@@ -212,9 +207,5 @@ export async function hasUserAnOrgOrIsAlreadyInvited(userEmails: string[]) {
     .get());
   const invitationQuery = await Promise.all(invitationPromises);
 
-  const isInvitationExist = invitationQuery.some(d => d.docs.length > 0);
-
-  if (isUserOrgIdExist === true && isInvitationExist === true) return true;
-  if (isUserOrgIdExist === true || isInvitationExist === true) return true;
-  else return false;
+  return invitationQuery.some(d => d.docs.length > 0);
 }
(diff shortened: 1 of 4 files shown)


libs/invitation/src/lib/+state/invitation.service.ts
```diff
@@ -58,7 +58,7 @@ export class InvitationService extends CollectionService<InvitationState> {
 
   /** Return true if there is already a pending invitation for a list of users */
   public async orgInvitationExists(userEmails: string[]): Promise<boolean> {
```
phpgeekfr, line 60: maybe also rename `orgInvitationExists`

apps/backend-functions/src/invitation.ts
```diff
@@ -190,13 +190,31 @@ That would have exceeded the current limit which is ${MEETING_MAX_INVITATIONS_NU
...
     .where('status', 'in', invitationStatus.filter(s => s !== 'declined'))
     .get());
   const invitationQuery = await Promise.all(invitationPromises);
-  return invitationQuery.some(d => d.docs.length > 0);
+
+  const isInvitationExist = invitationQuery.some(d => d.docs.length > 0);
+
+  if (isUserOrgIdExist === true && isInvitationExist === true) return true;
+  if (isUserOrgIdExist === true || isInvitationExist === true) return true;
+  else return false;
```
phpgeekfr, line 219: Why so much complexity ?
And the `if` at the end does not make sense.

You can do 
```
export async function hasUserAnOrgOrIsAlreadyInvited(userEmails: string[]) {

  const userPromises = userEmails.map(email => db.collection('users')
  .where('email', '==', email)
  .get());
  const userQuery = await Promise.all(userPromises);
  const orgIdExists = [];

  const someUserAlreadyHaveOrg = userQuery.some(d =>d.docs && d.docs.some(u => !!u.data().orgId));

  if(someUserAlreadyHaveOrg) { return true };

  const invitationPromises = userEmails.map(email => db.collection('invitations')
    .where('toUser.email', '==', email)
    .where('status', 'in', invitationStatus.filter(s => s !== 'declined'))
    .get());
  const invitationQuery = await Promise.all(invitationPromises);

  return invitationQuery.some(d => d.docs.length > 0);
}
```

phpgeekfr (changes requested):

apps/backend-functions/src/invitation.ts
Mussuranae: Thanks, it's modified

apps/backend-functions/src/invitation.ts
```diff
@@ -190,13 +190,22 @@ That would have exceeded the current limit which is ${MEETING_MAX_INVITATIONS_NU
...
-export async function isInvitationToJoinOrgExist(userEmails: string[]) {
+export async function hasUserAnOrgOrIsAlreadyInvited(userEmails: string[]) {
+
+  const userPromises = userEmails.map(email => db.collection('users')
+  .where('email', '==', email)
+  .get());
+  const userQuery = await Promise.all(userPromises);
+
+  const hasUserAlreadyAnOrg = userQuery.some(d => d.docs && d.docs.some(u => !!u.data().orgId));
+  if(hasUserAlreadyAnOrg) return true;
```
phpgeekfr, line 201: bad indent

phpgeekfr (approved):

PR #4752, opened 2021-01-15
Feat 4158 - firebase emulator setup for front (with functions and auth)

phpgeekfr:
fix #4752 
You will need to run `npm i`

How to use : https://www.notion.so/cascade8/Emulator-79492738d2614b35b6435eb80584ff26


libs/firebase-utils/src/lib/firestore/emulator.ts
```diff
@@ -45,8 +45,8 @@ export async function importFirestoreEmulatorBackup(gcsPath: string, emulatorBac
  * This promise resolves when it detects that the emulator has started and is ready.
  * @param emuPath path to the root Firebase emulator export
  */
-export async function startFirestoreEmulatorWithImport(emuPath: string) {
-  const cmd = `firebase emulators:start --only firestore --import ${emuPath} --export-on-exit`;
+export async function startFirestoreEmulatorWithImport(emuPath: string, exportOnExit: boolean = true) {
+  const cmd = `firebase emulators:start --only firestore --import ${emuPath} ${exportOnExit ? '--export-on-exit' : ''}`;
```
GrandSchtroumpf, line 49: Create a const : 
```typescript
const export = exportOnExit ? 'export-on-exit' : '';
const cmd =  `firebase emulators:start --only firestore --import ${emuPath} ${export}`;
```

libs/user/src/lib/auth/components/widget/widget.component.ts
```diff
@@ -37,4 +38,19 @@ export class AuthWidgetComponent {
     const mode = checked ? 'dark' : 'light';
     this.themeService.theme = mode;
   }
+
+  getEmulators(): string {
```
GrandSchtroumpf, line 42: This should be a property not a method.
```typescript
const emulatorList = Object.entries(emulators).filter(([key, value]) => !!value).map(([key, value]) => key);
const emulators = emulatorList.length ? emulatorList.join(' - ') : 'none'
```

GrandSchtroumpf (approved):

PR #4754, opened 2021-01-15
George/refactor

george43g:
- new import export commands with appropriate glue
- automatic backup date-sorting
- function to enable maintenance in firestore backup
- preparations for a backup/restore solution
- migrationBeta - backup with new import/export methods
- upgradeBeta - using migrationBeta


phpgeekfr (commented):
I want to approve your PR because the code looks OK but I'm lost with all methods: old ones, beta, not used yet ..  can you clarify by naming properly, adding comment to say `@deprecated use NewOne instead` ?

Or even put  'old' and 'beta"  in separate files ?

phpgeekfr:
If you can add comment explaining what does a method and when it should be executed ? 

Or if you can update Notion with articles you started a while ago about PFT, backup, restore ..

apps/backend-ops/src/main.ts
```diff
@@ -10,10 +10,10 @@ import { disableMaintenanceMode, displayCredentials, isMigrationRequired, showHe
...
 import { clearUsers, createUsers, printUsers, generateWatermarks, syncUsers } from './users';
 import { generateFixtures } from './generate-fixtures';
-import { backup } from './admin';
+import { backup, exportFirestore, importFirestore } from './admin';
 import { selectEnvironment } from './select-environment';
 import { healthCheck } from './health-check';
-import { anonDbLocal, anonymizeLatestProdDb, downloadProdDbBackup, importEmulatorFromBucket, loadEmulator, uploadBackup } from './emulator';
+import { anonDbLocal, anonymizeLatestProdDb, downloadProdDbBackup, importEmulatorFromBucket, loadEmulator, switchOnMaintenance, uploadBackup } from './emulator';
 
 const args = process.argv.slice(2);
```
phpgeekfr, line 18: as @GrandSchtroumpf suggested, is it possible to enter arguments like this : ` npm run backend-ops someFunc --someArg  /path/to/lol --localPath value` ?

george43g: Yes it's possible, but we're going to have to use a package to parse CLI arguments. I'm not writing that from scratch (unless I have to?)

phpgeekfr: Yes, use an external package. Can you do this in a separate PR ?

phpgeekfr: https://github.com/blockframes/blockframes/issues/4779

phpgeekfr (approved):

Issue #4769, opened 2021-01-18
Screening session - UI bugs


Jackseed:
Spinner & error message + illustration should be centered both horizontally & vertically in the page
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/60f14e1c-dd66-4e27-9fa0-e6a7df77571b)![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/25f9f634-b7b3-4275-867c-ccbed603536d)

RemcoSimonides:
Spinner seems to be horizontally aligned already. Can you check again @Jackseed? Closing for now

It sounds really easy to do but we can't center the loader vertically... Well, at least not with side effects in the footer
![image](https://user-images.githubusercontent.com/27687382/105062031-73c91500-5a7a-11eb-9b6a-72f1bfba1041.png)

PR #4785, opened 2021-01-19
height/width loader spinner centered

SamirTouira:
related to: #4769 

made with @RemcoSimonides

### DIFF
--- a/apps/festival/festival/src/app/marketplace/event/session/session.component.html
+++ b/apps/festival/festival/src/app/marketplace/event/session/session.component.html
@@ -62,7 +62,7 @@
 
 
 <ng-template #loading>
-  <section fxLayout="column" fxLayoutAlign="center center">
+  <section class="loading-spinner" fxLayout="column" fxLayoutAlign="center center">
     <mat-spinner></mat-spinner>
   </section>
 </ng-template>
--- a/apps/festival/festival/src/app/marketplace/event/session/session.component.scss
+++ b/apps/festival/festival/src/app/marketplace/event/session/session.component.scss
@@ -92,6 +92,15 @@ p {
   }
 }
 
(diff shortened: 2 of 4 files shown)


apps/festival/festival/src/app/marketplace/event/session/session.component.html
```diff
@@ -62,7 +62,7 @@ <h2>by {{ event.movie.directors | displayName }}</h2>
 
 
 <ng-template #loading>
-  <section fxLayout="column" fxLayoutAlign="center center">
+  <section class="loading-spinner" fxLayout="column" fxLayoutAlign="center center">
```
fritzschoff, line 65: guess you don't need the flex layout anymore if you are using position absolute?

fritzschoff (approved):

apps/festival/festival/src/app/marketplace/event/session/session.component.scss
```diff
@@ -92,6 +92,15 @@ p {
   }
 }
 
+.loading-spinner {
+  position:absolute;
+	left:0;
+	right:0;
+	top:0;
+	bottom:0;
+	margin:auto;
```
RemcoSimonides, line 101: You can remove `margin: auto`

RemcoSimonides (changes requested):
Wait

RemcoSimonides (approved):

apps/festival/festival/src/app/marketplace/event/session/session.component.html
SamirTouira: I can remove the `fxLayout="column"` indeed, but I have to keep `fxLayoutAlign="center center"` to keep the spinner centered. I just let this if you are okay

PR #4804, opened 2021-01-21
Fix(UT): Add notification model and increase timeout for the test. Fixes #4739

### LINKED ISSUE #4739: Fix Notification unit-test
Notification unit-tests are broken after new config.

image

xmano:
Fixes #4739

### DIFF
--- a/libs/notification/src/lib/+state/notification.service.spec.ts
+++ b/libs/notification/src/lib/+state/notification.service.spec.ts
@@ -2,6 +2,7 @@
 
 import { NotificationService } from './notification.service';
 import { NotificationStore } from './notification.store';
+import { Notification } from './notification.model';
 import { AngularFireModule } from '@angular/fire';
 import { SETTINGS, AngularFirestoreModule, AngularFirestore } from '@angular/fire/firestore';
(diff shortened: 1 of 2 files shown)


libs/notification/src/lib/+state/notification.service.spec.ts
```diff
@@ -53,7 +54,8 @@ describe('Notifications Test Suite', () => {
     await db.doc('notifications/1').set({ isRead: false });
     await service.readNotification({ id: '1' });
     const doc = await db.doc('notifications/1').ref.get();
-    expect(doc.data().isRead).toBeTruthy();
+    const notif = doc.data() as Notification;
```
RemcoSimonides, line 57: I'd prefer the use of full names instead of abbreviation; use full name instead `const notification`

RemcoSimonides (approved):

libs/notification/src/lib/+state/notification.service.spec.ts
xmano: ok will change

xmano:
Made changes to var names

image

phpgeekfr (approved):

Issue #4760, opened 2021-01-18
Typing few characters in country field does not bring up matching countries in drop-down list


xmano:
In movie-form Additional Information (Step 6/8)
![image](https://user-images.githubusercontent.com/8731418/104907200-b58a8a80-59aa-11eb-9ff3-a38b264e7156.png)

This needs to be checked with other country fields also.

SamirTouira:
Strangly I can not trigger this bug, when I type the two firsts letters, that's display correctly :
![image](https://user-images.githubusercontent.com/52285231/104914868-0d67c800-5990-11eb-9c6b-95f2da44ebef.png)

Jackseed:
Tested also myself on staging (develop from last Friday afternoon) and working
![image](https://user-images.githubusercontent.com/14964443/104922513-1f029d00-599b-11eb-839c-b4475325a165.png)
I think the test failing comes from the territory list that has been updated, "United States of America (the)" is now "United States of America"
![image](https://user-images.githubusercontent.com/14964443/104922675-51ac9580-599b-11eb-9f87-efcfd9ca3d75.png)
![image](https://user-images.githubusercontent.com/14964443/104922683-54a78600-599b-11eb-86e0-0783039733a4.png)

xmano:
@Jackseed : it was on latest develop .. i will again confirm it

xmano:
@Jackseed & @SamirTouira I am able to recreate this in a different way. The issue may be with any similar component of UI.

Recreate steps:
1. While editing movie in Addition info give any country in the country release field (say Afghanistan).
2. Save the info and navigate to another page  say Main Information.
3. Return back to Additional Information step. Edit the release info. Now clear out country field and type Fr - list is not updated.

This may be a corner case but bug exists and it may be in all input-drop component.

I also found another issue about date formatting in the same form:

![image](https://user-images.githubusercontent.com/8731418/105028432-07004b80-5a77-11eb-8278-811286d02e80.png)

Jackseed:
Thanks @xmano for the steps to reproduce, it's indeed here and was a hard one to catch!
@SamirTouira mb you can have a look to this one? We won't wait for a fix for the release as it's an edge case but would be definitely good to fix. 

Also regarding the date format it's the placeholder that is wrong:
instead of "YYYY/MM/DD" it should be "MM/DD/YYYY" 
Samir can you change that also on this issue?

@fritzschoff I tag you so you may help Samir since it's movie form related

[...]

PR #4806, opened 2021-01-21
country suggestion list fixed when we type the few first letters

SamirTouira:
Related to: #4760 

Special thanks to @RemcoSimonides and @fritzschoff for the quick help


libs/ui/src/lib/form/list/form-list.component.ts
```diff
@@ -83,7 +84,7 @@ export class FormListComponent<T> implements OnInit, OnDestroy {
 
   /** Add a clean form */
   add() {
-    this.formItem = this.form.createControl();
+   return this.formItem = this.form.createControl();
```
RemcoSimonides, line 87: You can remove this return

libs/ui/src/lib/form/list/form-list.component.ts
```diff
@@ -107,6 +108,7 @@ export class FormListComponent<T> implements OnInit, OnDestroy {
     this.formItem = this.form.at(index);
     this.activeValue = this.formItem.value;
     this.cdr.markForCheck();
+
```
RemcoSimonides, line 111: This line doesn't have to be here ;)

libs/ui/src/lib/form/list/form-list.component.ts
```diff
@@ -124,7 +126,8 @@ export class FormListComponent<T> implements OnInit, OnDestroy {
       this.activeIndex--;
     }
     if (this.isFormEmpty) {
-      this.add();
+     this.add();
+     delete this.formItem;
```
RemcoSimonides, line 130: I think these changes can be reverted too, right? As it isn't needed anymore to solve the bug

RemcoSimonides (changes requested):

libs/ui/src/lib/form/list/form-list.component.ts
SamirTouira: After have tried, we should keep the `delete this.formItem;`, because deleting the formItem object avoid this bug, I have tested to remove this line but the bug comes back since the form isn't reinitialized when we delete the latest country with the delete button

libs/ui/src/lib/form/list/form-list.component.ts
```diff
@@ -69,12 +69,13 @@ export class FormListComponent<T> implements OnInit, OnDestroy {
     ]).pipe(
       map(([list, reverse]) => reverse ? list.reverse() : list)
     );
-
-    this.add();
+    if (!this.form.controls.length) {
```
RemcoSimonides, line 72: Actually it's probably more neat to use `!this.isFormEmpty` here

libs/ui/src/lib/form/list/form-list.component.ts
RemcoSimonides: Issue with this solution is that
this.add() does `this.formItem = this.form.createControl()`
and right after that you want to delete it again with `delete this.formItem`
So I guess the better solution would be to remove both `this.add()` and `delete this.formItem`

SamirTouira: If I delete both `this.add()` and `delete this.formItem` in the `remove()` function, and if inside `ngOnInit()` I  write

`    if (!this.isFormEmpty) {
      this.add();
    }`

the bug is triggered again. But if I add `delete this.formItem` in `ngOnInit()`, then in 

`if (!this.isFormEmpty) {
      this.add();
      delete this.formItem;
    }`
 
it works correctly. It seems clean with that way, what do you think about it?

RemcoSimonides: Sorry, remove `!` before `this.isFormEmpty` in the `ngOnInit()`

RemcoSimonides: And don't do both `this.add()` and `delete this.formItem`. The only thing you're doing then is adding and removing it immediately again

SamirTouira: Indeed, and with `    if (this.isFormEmpty) {
      this.add();
    }` it works perfectly. nice one

libs/ui/src/lib/form/list/form-list.component.ts
```diff
@@ -123,8 +124,5 @@ export class FormListComponent<T> implements OnInit, OnDestroy {
     if (this.activeIndex > index) {
       this.activeIndex--;
     }
-    if (this.isFormEmpty) {
-      this.add();
-    }
```
RemcoSimonides, line 128: Add back
```
if (this.isFormEmpty) {
  this.formItem = this.form.createControl({});
}
```

RemcoSimonides (approved):

PR #4807, opened 2021-01-22
Fix(UT): Need to set env var. Also increase timeout for test. Fixes #4740

### LINKED ISSUE #4740: Fix twilio unit test
Twilio unit tests need to be refactored to work with new nx 11 new configs.

image

xmano:
Fix #4740

### DIFF
--- a/apps/backend-functions/src/test-setup.ts
+++ b/apps/backend-functions/src/test-setup.ts
@@ -1,2 +1,4 @@
-import 'jest-preset-angular';
+﻿import 'jest-preset-angular';
 import 'jest';
+
+jest.setTimeout(10000);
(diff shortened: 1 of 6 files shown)


RemcoSimonides (approved):
Title of PR says 'need to set env var' but I don't see anything about that? Other than that LGTM

xmano:
> Title of PR says 'need to set env var' but I don't see anything about that? Other than that LGTM

It goes into .env file and is gitignored. We could possibly add a comment regarding this.

Also need to fix build issue. Working on it.

RemcoSimonides:
Ah ye, would be useful to write it down in #dev channel for example if everyone needs it ;)

phpgeekfr (approved):

Issue #4789, opened 2021-01-20
Movie form - Festival premiere: uncheck an option


Jackseed:
URL: /c/o/dashboard/tunnel/movie/movieId/reviews
Section: Festival & awards

Currently there is no way to uncheck a Premiere clicked on the toggle button. User should be able to uncheck the option when clicking again on the active element. 
![image](https://user-images.githubusercontent.com/14964443/105166371-aecd5600-5b17-11eb-80cf-4a108eaf843d.png)

Related question: https://stackoverflow.com/questions/51282112/uncheck-active-angular-material-toggle-buttons-on-click

phpgeekfr:
closing this since #4808 indicate that issue is fixed.

PR #4808, opened 2021-01-22
toggles could be uncheck

SamirTouira:
related to: #4789 

The buttons Premiere can be unchecked now


libs/movie/src/lib/movie/form/reviews/reviews.component.html
```diff
@@ -34,7 +34,8 @@ <h3>Festival & Awards</h3>
 
             <h6 class="mat-caption">Premiere</h6>
             <div fxLayout="column" class="toggle-btn">
-              <mat-button-toggle-group [formControl]="prizesForm.get('premiere')" name="Premiere" aria-label="Premiere">
+              <mat-button-toggle-group [formControl]="prizesForm.get('premiere')" name="Premiere" aria-label="Premiere"
+                #group="matButtonToggleGroup" (change)="onChange()" (click)="onClick(group)">
```
fritzschoff, line 38: it is a bit redundant if you setting on every change the variable to true.
can you try something like that?
```typescript
,mat-button-toggle ... (change)="isChecked($event);

and in your ts file

isChecked(change: MatButtonToggleChange) {
if(this.form.get('premiereType').value === change.value) this.form.get('premiereType').reset();
}

Comment (fritzschoff) [reply libs/movie/src/lib/movie/form/reviews/reviews.component.html:38]: if so, then you also can delete the ref variable group and the click event listener.

File: libs/movie/src/lib/movie/form/reviews/reviews.component.ts [hunk 2]
```diff
@@ -11,10 +12,19 @@ import { DynamicTitleService } from '@blockframes/utils/dynamic-title/dynamic-ti
 })
 export class MovieFormReviewscComponent {
   form = this.shell.getForm('movie');
-
+  toggleChanged: boolean;
```
fritzschoff, line 15: you can remove that

fritzschoff (changes requested):

libs/movie/src/lib/movie/form/reviews/reviews.component.html
SamirTouira: It strangly doesn't work. its toLabel('premiereType' isn't recognized... I don't really know how to get the value of these Premiere buttons after have tried many things

fritzschoff (approved):

PR #4809, opened 2021-01-22
Feat 4046 - Notification settings  part 1

phpgeekfr:


apps/backend-functions/src/main.ts
```diff
@@ -195,6 +196,12 @@ export const sendMailAsAdmin = functions.https.onCall(skipInMaintenance(_sendMai
  */
 export const sendMailWithTemplate = functions.https.onCall(skipInMaintenance(_sendMailWithTemplate));
 
+/**
+ * Look at notifications collection to find emails to send
+ */
+export const sendNotificationEmails = functions.pubsub.schedule('*/5 * * * *') // every 5 minutes
```
phpgeekfr, line 202: onNotificationCreate

libs/user/src/lib/auth/forms/notifications-form/notifications-form.component.scss
```diff
@@ -0,0 +1,16 @@
...
+section {
+  margin-bottom: 36px;
+}
+
+mat-checkbox {
+  margin-bottom: 24px;
+}
+
+button.mat-flat-button {
+  margin: 136px 0 36px;
```
GrandSchtroumpf, line 15: Hmmm I think there is an issue here. Always but a symmetric margin (except for gaps). If you're not sure about the css just remove the margin and let @Mussuranae do this part ;)

libs/user/src/lib/auth/forms/notifications-form/notifications-form.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  styleUrls: ['./notifications-form.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class NotificationsFormComponent implements OnInit, OnDestroy {
+
+  @Input() form: NotificationsForm;
+
+  public masterToggleStatus: 'accept' | 'reject' | 'other'
+  private sub: Subscription;
+  public appName;
```
GrandSchtroumpf, line 27: type it to `App`

libs/user/src/lib/auth/forms/notifications-form/notifications-form.component.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  constructor(private routerQuery: RouterQuery) { }
+
+  ngOnInit() {
+    this.sub = this.form.valueChanges.pipe(startWith(this.form.value)).subscribe(data => {
```
GrandSchtroumpf, line 32: It could be an observable directly that you can manage with a | async in the html : 
```typescript
this.form.valueChanges.pipe(
  startWith(this.form.value),
  map(data => {
      const allAccepted = Object.keys(data).every(key => data[key]);
      const allRejected = Object.keys(data).every(key => !data[key]);
      return allAccepted ? 'accept' : allRejected ? 'reject' : 'other';
  })
)
```

libs/user/src/lib/auth/pages/login/login.component.ts
```diff
@@ -70,9 +71,16 @@ export class LoginComponent implements OnInit {
     try {
       const { email, password, firstName, lastName } = signupForm.value;
       const createdFrom = getCurrentApp(this.routerQuery);
-      await this.service.signup(email.trim(), password, { ctx: { firstName, lastName, _meta: { createdFrom } } });
       const privacyPolicy = await this.service.getPrivacyPolicy();
-      await this.service.update({ privacyPolicy: privacyPolicy });
+      await this.service.signup(email.trim(), password, {
```
GrandSchtroumpf, line 75: Create a variable : 
```typescript
const ctx = { ... };
await this.service.signup(email.trim(), password, { ctx });
```

libs/user/src/lib/auth/pages/notifications/notifications.component.html
```diff
@@ -0,0 +1,11 @@
+<mat-card>
+  <mat-card-content>
+    <user-notifications-form [form]="form"></user-notifications-form>
+  </mat-card-content>
+</mat-card>
+<footer fxLayout fxLayoutAlign="center center">
```
GrandSchtroumpf, line 6: There is one update per page ? It's not shared by the whole form of the user ?

libs/user/src/lib/user/+state/user.firestore.ts
```diff
@@ -11,12 +13,43 @@ export interface User extends PublicUser {
...
+    ...notifications
   }
 }
 
+export function createUserSettings(settings: Partial<UserSettings> = {}): UserSettings {
+  return {
+    ...settings,
+    notifications: {
+      ...settings.notifications,
+      ...createNotificationSettings(),
```
GrandSchtroumpf, line 49: I think you want to do : 
```typescript
return {
  ...settings,
  notifications: createNotificationSettings(settings.notifications),
}
```

libs/utils/src/lib/emails/utils.ts
```diff
@@ -37,6 +37,27 @@ export interface EventEmailData {
   end: string
 }
 
+export type EmailErrorCodes = 'E01-unauthorized' | 'E02-general-error' | 'E03-missing-api-key' | 'E04-no-template-available';
+
+export const emailErrorCodes = {
+  E01: {
```
GrandSchtroumpf, line 43: I think the key should be more explicit: 
```typescript
unauthorized: { ... }
general: { ... }
missingKey: { ... }
noTemplate: { ... }
```

apps/backend-functions/src/notification.ts
```diff
@@ -1,28 +1,100 @@
...
 }
 
+// @TODO #4046 should contain all notificationTypes in the end
+const customizableNotificationsTypes: NotificationType[] = ['memberAddedToOrg', 'memberRemovedFromOrg'];
+
+async function appendNotificationSettings(notification: NotificationDocument) {
+  // get user notification settings
+  const user = await getDocument<User>(`users/${notification.toUserId}`);
+
+  if (customizableNotificationsTypes.includes(notification.type)) {
```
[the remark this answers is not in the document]
phpgeekfr: `  if (customizableNotificationsTypes.includes(notification.type)) {` is here to check if we handle this case in the code for this particular `notification.type`, not related to user settings.
Check for overrided config in user.settings will be handled in next part of PR

apps/backend-functions/src/notification.ts
```diff
@@ -1,28 +1,100 @@
...
     ...notification
   };
 }
+
+export async function onNotificationCreate(snap: FirebaseFirestore.DocumentSnapshot): Promise<any> {
+  const notification = snap.data() as NotificationDocument;
+
+  if (notification.email?.isSent === false) {
+    // Update notification state
+    if (customizableNotificationsTypes.includes(notification.type)) {
```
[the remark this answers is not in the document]
phpgeekfr: here, user.settings have already been applied on notification document so no need to check users preferences.
This if is  here to check if we handle this case in the code for this particular notification.type (we need a template id for example)

libs/user/src/lib/auth/forms/notifications-form/notifications-form.component.scss
phpgeekfr: This is actually @Mussuranae work ;) We merged her PR into this one.  Should I remove the margin ?

libs/user/src/lib/auth/forms/notifications-form/notifications-form.component.ts
phpgeekfr: To do this I need to cast `this.appName = appName[getCurrentApp(this.routerQuery)]`  as  `App` , should I do it ?

libs/user/src/lib/auth/pages/notifications/notifications.component.html
phpgeekfr: yes, not shared. This is a different form than profile page for example and it has is own `update()` method

libs/user/src/lib/user/+state/user.firestore.ts
phpgeekfr: I wanted to do this but I can't. Since  `NotificationSettings` is a record, it expect to have all possible values of the record.

Doing 
```
    notifications: {
      ...settings.notifications,
      ...createNotificationSettings(),
```
is to trick the linter letting him believe that each value of record is filled

libs/user/src/lib/auth/forms/notifications-form/notifications-form.component.ts
phpgeekfr: This code was a copy/paste from another component and is not used, I removed it

apps/backend-functions/src/notification.ts
```diff
@@ -1,28 +1,97 @@
...
 }
 
+// @TODO #4046 should contain all notificationTypes in the end
+const types: NotificationType[] = ['memberAddedToOrg', 'memberRemovedFromOrg'];
+
+async function appendNotificationSettings(notification: NotificationDocument) {
+  // get user notification settings
+  const user = await getDocument<User>(`users/${notification.toUserId}`);
+
+  if (types.includes(notification.type)) {
```
GrandSchtroumpf, line 35: For me it should be : 
```typescript
const updateNotif = ({ email, app }: { email: boolean, app: boolean }) => {
  if (email) notification.email = { isSent: false };
  if (app) notification.app = { isRead: false };
}
if (!user.settings?.notifications?.default) {
  updateNotif({ app: true });
} else if (notification.type in user.settings.notifications) {
  updateNotif(user.settings.notifications[notification.type])
} else {
  updateNotif(user.settings.notifications.default);
}
return notification;
```

apps/backend-functions/src/notification.ts
```diff
@@ -1,28 +1,97 @@
...
+    // Update notification state
+    if (types.includes(notification.type)) {
+      const user = await getDocument<User>(`users/${notification.toUserId}`);
+
+      // Send email
+      // const appKey = await getOrgAppKey(user.orgId); //@TODO also use notification.type to guess appKey
+      // await sendMailFromTemplate({ to: user.email, templateId: 'TODO#4046', data: {} }, appKey); // @TODO #4046
+      await sendMail({ to: user.email, subject: notification.type, text: 'test' })
+        .then(_ => notification.email.isSent = true)
+        .catch(e => notification.email.error = e.message);
```
GrandSchtroumpf, line 90: Don't you want to keep the full error ? I think we could store code & message in error, but if you're happy with just the code it's fine.

apps/backend-functions/src/internals/email.ts
```diff
@@ -42,20 +42,18 @@ export function sendMailFromTemplate({ to, templateId, data }: EmailTemplateRequ
...
-    return;
+    throw new Error(emailErrorCodes.missingKey.code);
   }
 
   SendGrid.setApiKey(sendgridAPIKey);
-  return SendGrid.send(msg).catch(e => {
+  return SendGrid.send(msg).then(_ => true).catch(e => {
     if (e.message === 'Unauthorized') {
-      console.log('API key is not authorized to send mails. Please visit: https://www.notion.so/cascade8/Setup-SendGrid-c8c6011ad88447169cebe1f65044abf0 ');
+      throw new Error(emailErrorCodes.unauthorized.code);
```
GrandSchtroumpf, line 53: I think you should emit the code & the message else message is useless.

libs/user/src/lib/auth/forms/notifications-form/notifications-form.component.scss
GrandSchtroumpf: yes please. I'm going to have a conversation with her XD

libs/user/src/lib/user/+state/user.firestore.ts
```diff
@@ -11,12 +13,43 @@ export interface User extends PublicUser {
...
+
+export interface UserSettings {
+  notifications: NotificationSettings,
+}
+
+export interface NotificationSettingsTemplate { email: boolean, app: boolean };
+
+export type NotificationSettings = Record<NotificationType | InvitationType | 'default', NotificationSettingsTemplate>;
+
+export function createNotificationSettings(notifications: Partial<NotificationSettings> = {}): Partial<NotificationSettings> {
```
GrandSchtroumpf, line 33: This should return a `NotificationSettings` and not partial

libs/user/src/lib/user/+state/user.firestore.ts
GrandSchtroumpf: You should be able to do that with comment above

GrandSchtroumpf (approved):

libs/user/src/lib/user/+state/user.firestore.ts
phpgeekfr: I don't know how  I can accomplish this. Linter is not happy because  since input is partial. All required values for Record may not be present.

Issue #4800, opened 2021-01-20
Upload Widget Bug


pldespaigne:
1) After a successful upload, the widget doesn't close itself.
2) Moreover, clicking on the blue check circle button inside the widget will delete the uploaded file from the db & storage.

![image](https://user-images.githubusercontent.com/5858657/105192250-3034e080-5b38-11eb-83d8-dfa301ecf830.png)

PR #4811, opened 2021-01-22
4800 workaround AngularFire bug

RemcoSimonides:
Workaround for #4800

AngularFireUploadTask state is not updated with the latest state of the firebase.UploadTask. Therefore we use a workaround to get the state directly from firebase UploadTask until AngularFire fixes this bug.

This bug exists since update to firebase version 8 and higher.

Also, fixed the bug where it would remove the file when clicking on the button of completed uploads.


pldespaigne (changes requested):
please update your PR with following
- [x] change target branch to be `media-refactoring`
- [ ] the change should be in `file-uploader.service.ts` instead of `media.service.ts`
- [x] update the code to be sure that the fix still works after all the work we did on media-refactoring

RemcoSimonides:
Closing in favor of https://github.com/blockframes/blockframes/pull/5012

PR #4824, opened 2021-01-27
WIP

pldespaigne:


apps/backend-functions/src/media.ts
```diff
@@ -1,92 +1,296 @@
...
+    // (1) Security checks
+
+    // Is the user allowed to upload this file ?
+    switch (metadata.collection) {
+      case 'users': {
+        // only the user is allowed to upload files about himself
+        if (metadata.docId !== metadata.uid) return cleanAndReturn(false);
+        break;
+
+      } case 'movies':
```
GrandSchtroumpf, line 65: Format: 
```typescript
switch (metadata.collection) {
  case: 'users': { }
  case: 'movies': { }
}
```

apps/backend-functions/src/media.ts
```diff
@@ -1,92 +1,158 @@
...
+  // which has been verified by storage rules and can be trusted
+  const metadata = data.metadata as FileMetaData;
+  const isValid = isValidMetadata(metadata, { uidRequired: true });
+
+  const [ tmp ] = data.name.split('/');
+  if ( tmp === tempUploadDir ) {
+
+    // (1) Security checks, (2) copy file to final destination, (3) update db, (4) delete tmp/file
+
+    const cleanAndReturn = async (success: boolean) => {
```
GrandSchtroumpf, line 49: Personally I don't see the point of this function you can directly do everywhere
```typescript
return file.delete();
```
Except if you really need to return a boolean after. But honestly I think you don't need

apps/backend-functions/src/media.ts
```diff
@@ -1,92 +1,158 @@
...
+
+    if (!isValid) return cleanAndReturn(false);
+
+    // (1) Security checks
+
+    // Is the user allowed to upload this file ?
+    switch (metadata.collection) {
+      case 'users': {
+        // only the user is allowed to upload files about himself
+        if (metadata.docId !== metadata.uid) return cleanAndReturn(false);
```
GrandSchtroumpf, line 62: For example here you could do : 
```typescript
if (metadata.docId !== metadata.uid) return file.delete();
```

apps/backend-functions/src/media.ts
```diff
@@ -1,92 +1,158 @@
...
-    if (Array.isArray(savedRef)) {
-      savedRef = savedRef.map(e => e.ref || e).find(ref => ref === filePath) || '';
-    } else if (!!savedRef.ref) {
-      savedRef = savedRef.ref;
+    const segments = data.name.split('/');
+    segments.shift(); // remove tmp/
+    const finalPath = segments.join('/');
+    const to = bucket.file(finalPath);
+    const [exists] = await file.exists();
+    if (!exists) {
```
GrandSchtroumpf, line 96: Why this only happens here ? Shouldn't it be the first check ?

apps/backend-functions/src/media.ts
```diff
@@ -1,92 +1,158 @@
...
+    // (4) delete tmp/file
+    return cleanAndReturn(true);
+
+  } else {
+
+    if (!isValid) throw new Error('Invalid metadata after file copy');
+
+    // Post processing such as: signal end of upload flow, trigger upload to JWPlayer, ...
+
+    if (data.contentType.indexOf('video/') === 0 && metadata.collection === 'movies') {
```
GrandSchtroumpf, line 125: Create a variable with a good name that describe this.

libs/ui/src/lib/form/list/form-list.component.ts
```diff
@@ -84,7 +84,7 @@ export class FormListComponent<T> implements OnInit, OnDestroy {
 
   /** Add a clean form */
   add() {
-    this.formItem = this.form.createControl();
+    this.formItem = this.form.createControl(undefined, this.form.length);
```
GrandSchtroumpf, line 87: hmmmm this is weird. Is it suppose to stay ? You shouldn't need to do that, especially on a reusable class like that.

libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -29,17 +29,17 @@ function hasValue(value: any): boolean {
 /** A list of FormField */
 export class FormList<T, Control extends AbstractControl = any> extends FormArray {
   private _value: T[];
-  createControl: (value?: Partial<T>) => Control = createControlForm;
+  createControl: (value?: Partial<T>, index?: number) => Control = createControlForm;
```
GrandSchtroumpf, line 32: What's the goal of the index ? You shouldn't change this highly reusable file.

libs/ui/src/lib/form/list/form-list.component.ts
pldespaigne: I don't think it will stay, but this is the only way of specifying a second optional param without specifying the first optional param. This how it's done in the TS handbook documentation I think.

libs/utils/src/lib/form/forms/list.form.ts
pldespaigne: This add the possibility for the `createControl` defined in `factory` to know which index it's instantiating.
Its an optional param so I don't think it will cause trouble, and it's also a pretty natural signature to have `element, index` in array method

GrandSchtroumpf (approved):

Issue #4819, opened 2021-01-26
Screening Invitations E2E test fail due to unavailability of title


xmano:
Felicita movie seems to missing from the title drop-down

![image](https://user-images.githubusercontent.com/8731418/105827673-49320b80-5fe8-11eb-9dac-55bc4f136c3b.png)

Not sure if this is a permission issue on CI. Need to investigate and fix the root cause and make the tests green.

xmano:
Note: This issue would be fixed by:
https://github.com/blockframes/blockframes/pull/4825

However we will update the unit-test to check this condition.

PR #4825, opened 2021-01-27
fix rules for reading multiple movies

phpgeekfr:
This fixes rules update #4799 .

The  rule change introduced in #4799 is causing permissions errors when we attempt to fetch movies in `/c/o/dashboard/title` . The query should have worked since every movie document had correct permissions and if we attempt to fetch movies one by one, no issue with current rules. So ... My guess is that ` orgCan('canUpdate', userOrgId(), existingData().id) )` for "n" movies was making to much operation and we hit the limit..
Using `userOrgId() in existingData().orgIds`, no problem (and it fit the actual query in front  `fromOrg = (orgId: string): QueryFn => ref => ref.where('orgIds', 'array-contains', orgId);`).  But only users belonging to org that is in movie.orgIds will be allowed, if user from another org but with "canRead" permissions for this movie document will not be allowed. I don't think this is a problem, we don't give access to draft movies to users that are not belonging to our org.

@RemcoSimonides Can you confirm that this new rule is ok for you ? Do you also think that the problem for reading multiple movies is related to operations limit or another thing that I missed?  

@xmano  Is it possible to create an unit test to read multiples movies (with accepted and draft status, all movies belonging to user's org) and check it we can read all at once ? Like this we could try the previous rules set to check if it falls.  But I don't know if emulator have the same operation limits than real firestore while evaluating rules. Do you have an idea?

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -288,7 +288,7 @@ service cloud.firestore {
       allow read: if userHasValidOrg()
         && notInMaintenance()
         && ( existingData().storeConfig.status == 'accepted'
-         || ( isOrgMember(userOrgId()) && userOrgId() in existingData().orgIds ) 
+         || ( isOrgMember(userOrgId()) && ( userOrgId() in existingData().orgIds )) 
         )
       allow create: if userHasValidOrg()
         && isRequestFieldValueEqualTo(['storeConfig', 'status'], 'draft')
--- a/firestore.rules
+++ b/firestore.rules
@@ -288,7 +288,7 @@ service cloud.firestore {
       allow read: if userHasValidOrg()
         && notInMaintenance()
         && ( existingData().storeConfig.status == 'accepted'
-         || ( isOrgMember(userOrgId()) && orgCan('canUpdate', userOrgId(), existingData().id) )
+         || ( isOrgMember(userOrgId()) && userOrgId() in existingData().orgIds ) 
         )
       allow create: if userHasValidOrg()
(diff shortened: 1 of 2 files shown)


firestore.rules
```diff
@@ -526,7 +526,7 @@ service cloud.firestore {
 		// Parameter "action" can either be "canCreate", "canRead", "canUpdate" or "canDelete"
     function orgCan(action, orgId, docId) {
       return (getDocumentPermissions(orgId, docId).ownerId == orgId)
-        || (getDocumentPermissions(orgId, docId).admin == true)
+        || (getDocumentPermissions(orgId, docId).isAdmin == true)
```
phpgeekfr, line 529: typo here from ages. I think we did not discovered yet because `(getDocumentPermissions(orgId, docId).ownerId == orgId)` always returns true first.

RemcoSimonides (approved):
LGTM

firestore.rules
xmano: I would like to have parenthesis around `userOrgId() in existingData().orgIds` to make the intent clear.

Also orgCan() change impact seems significant. It will be good to run all the unit-test suites and even application to make sure this does not break any other assumptions.

I will write new unit-tests suggested as part of #4819 . You may merge your changes. LGTM

xmano (approved):
Approved with some comments. Otherwise looks Ok.

phpgeekfr:
I added suggested parenthesis.  Change in orgCan must be tested but I think will not make any changes, `(getDocumentPermissions(orgId, docId).ownerId == orgId)` is likely always true

PR #4828, opened 2021-01-28
4693 import analytics in firestore and use

### LINKED ISSUE #4693: Store structured analytics data into Firestore
Create a schedule functions to get the analytics every **day**, process them & upload them to firestore.
On the frontend, call the analytics collection instead of the callable function.

RemcoSimonides:
Fix #4693
- [x] Create rules so movie owners can read analytics

### DIFF
--- a/apps/catalog/catalog/src/app/dashboard/home/home.component.ts
+++ b/apps/catalog/catalog/src/app/dashboard/home/home.component.ts
@@ -1,11 +1,13 @@
 import { DynamicTitleService } from '@blockframes/utils/dynamic-title/dynamic-title.service';
-import { Component, OnInit, ChangeDetectionStrategy, OnDestroy, Optional } from '@angular/core';
-import { MovieQuery } from '@blockframes/movie/+state/movie.query';
+import { Component, OnInit, ChangeDetectionStrategy, Optional, OnDestroy } from '@angular/core';
 import { Observable, Subscription } from 'rxjs';
 import { MovieAnalytics } from '@blockframes/movie/+state/movie.firestore';
-import { MovieService } from '@blockframes/movie/+state/movie.service';
 import { Intercom } from 'ng-intercom';
-import { switchMap } from 'rxjs/operators';
+import { map, switchMap, tap } from 'rxjs/operators';
+import { AnalyticsService } from '@blockframes/utils/analytics/analytics.service';
+import { fromOrg, MovieService } from '@blockframes/movie/+state/movie.service';
+import { OrganizationQuery } from '@blockframes/organization/+state/organization.query';
+import { Movie } from '@blockframes/movie/+state/movie.model';
 
 const isAcceptedInApp = movie => movie.storeConfig.status === 'accepted' && movie.storeConfig.appAccess.catalog === true;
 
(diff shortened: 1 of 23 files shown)


libs/movie/src/lib/movie/components/movie-analytics-chart/movie-analytics-chart.component.ts
```diff
@@ -57,15 +56,16 @@ function toYMD(date: Date) {
...
   public chartData: any[] = [];
 
-  public delayTime$ = new BehaviorSubject(20000);
+  private delayTime$ = new BehaviorSubject(3000);
+  private isLoading = new BehaviorSubject(true);
   public loadingState$ = combineLatest([
-    this.movieQuery.analytics.selectLoading(),
-    this.delayTime$,
+    this.isLoading.asObservable(),
+    this.delayTime$
```
RemcoSimonides, line 68: In `home.component.ts` this component is used. When that page is first loaded, it triggers `orgQuery.selectActive()` 3 times, causing the graphs to rerender multiple times. Therefore I kept this workaround with a delay that waits 3 seconds before showing the graphs. Preferably I'd rather see the `orgQuery.selectActive()` to trigger only once but I didn't manage to do this yet. Is this a known issue? Or do you know a solution? @GrandSchtroumpf

libs/utils/src/lib/analytics/analytics.store.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export interface AnalyticsState extends EntityState<MovieAnalytics, string>, ActiveState<string> {}
+
+@Injectable({ providedIn: 'root' })
+@StoreConfig({ name: 'analytics', idKey: 'id' })
+export class AnalyticsStore extends EntityStore<AnalyticsState> {
+  constructor() {
+    super()
+  }
+}
```
RemcoSimonides, line 13: This `analytics.store.ts` and `analytics.service.ts` are the only added files. Do you think it's better to create a separate lib for analytics?

I could move components and interfaces for analytics from Movie lib to Analytics lib then. Same for org, event or whatever more analytics will be added later.

apps/catalog/catalog/src/app/dashboard/home/home.component.ts
```diff
@@ -17,25 +19,34 @@ const isAcceptedInApp = movie => movie.storeConfig.status === 'accepted' && movi
...
-    this.movieAnalytics$ = this.movieQuery.analytics.selectAll();
-    this.movieQuery.getCount()
-      ? this.dynTitle.setPageTitle('Seller\'s Dashboard')
-      : this.dynTitle.setPageTitle('New Title')
+    this.movies$ = this.orgQuery.selectActive().pipe(
+      switchMap(({ id }) => this.movieService.valueChanges(fromOrg(id)))
+    )
+
+    this.movieAnalytics$ = this.orgQuery.selectActive().pipe(
+      switchMap(({ id }) => this.movieService.valueChanges(fromOrg(id))),
```
GrandSchtroumpf, line 39: You're doing twice the query. This is not a good idea. Better would be to load the analytics directly from the HTML with a pipe : 
```html
<ng-template colRef="anaytics" let-movie="item">{{ movie.id | getTitleAnalytics | async }}</ng-template>
```
And the pipe : 
```typescript
@Pipe({ name: 'getTitleAnalytics' })
export class GetTitleAnalyticsPipe implements PipeTransform {
  constructor(private analyticsService: AnalyticsService) {}
  transform(titleId: string) {
    return this.analyticsService.valueChanges(titleId);
  }
}
```

apps/festival/festival/src/app/dashboard/title/activity/activity.component.ts
```diff
@@ -31,7 +31,7 @@ export class TitleActivityComponent implements OnInit {
   ngOnInit() {
     this.dynTitle.setPageTitle('Title page', 'Marketplace Activity');
     this.movieId = this.movieQuery.getActiveId();
-    this.movieAnalytics$ = this.movieService.getMovieAnalytics([this.movieId]);
+    this.movieAnalytics$ = this.analyticsService.valueChanges([this.movieId]);
```
GrandSchtroumpf, line 34: I think you don't need the array in the params of `valueChanges` here.

libs/movie/src/lib/movie/components/movie-analytics-chart/movie-analytics-chart.component.ts
GrandSchtroumpf: I would have to do it with you directly I think to understand better the whole situation.

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -36,7 +37,13 @@ export class HomeComponent implements OnInit, OnDestroy {
     const _isDataLoaded$ = new BehaviorSubject<boolean>(false);
     this.isDataLoaded$ = _isDataLoaded$.asObservable();
 
-    this.movieAnalytics$ = this.movieQuery.analytics.selectAll();
+    this.movieAnalytics$ = this.orgQuery.selectActive().pipe(
```
GrandSchtroumpf, line 40: We should leverage the pipe system described above I think. It would make the code easier. Let's discuss that during a call.

libs/utils/src/lib/analytics/analytics.store.ts
GrandSchtroumpf: We could create a lib if we've got some analytics related components. But for now we can keep it here.

GrandSchtroumpf (changes requested):

apps/catalog/catalog/src/app/dashboard/home/home.component.ts
RemcoSimonides: On the `home.component` the movieAnalytics is sent as one to movie analytics chart component
`<movie-analytics-chart [analyticsData]="movieAnalytics$ | async"></movie-analytics-chart>`
So it can't be split up here

There is already a AnalyticsPipe which does what you suggest here. It is used on dashboard/titles page.
`<ng-template colRef="view" let-movie="item">{{ movie.id | getViews | async }}</ng-template>`

apps/festival/festival/src/app/dashboard/title/activity/activity.component.ts
RemcoSimonides: If I pass in an array, it returns an array and that's what I need

apps/catalog/catalog/src/app/dashboard/home/home.component.ts
```diff
@@ -1,48 +1,40 @@
...
-      return this.movieService.syncWithAnalytics(ids)
-    })).subscribe();
-    this.movieAnalytics$ = this.movieQuery.analytics.selectAll();
-    this.movieQuery.getCount()
-      ? this.dynTitle.setPageTitle('Seller\'s Dashboard')
-      : this.dynTitle.setPageTitle('New Title')
+    this.titleAnalytics$ = this.orgQuery.selectActive().pipe(
+      switchMap(({ id }) => this.movieService.valueChanges(fromOrg(id))),
+      map(titles => titles.map(t => t.id)),
+      switchMap(ids => this.analyticsService.valueChanges(ids))
```
GrandSchtroumpf, line 32: You forgot to add the title : 
```typescript
tap(ids => ids.length ? this.dynTitle.setPageTitle('Seller\'s Dashboard') : this.dynTitle.setPageTitle('New Title'))
```

GrandSchtroumpf (approved):

PR #4829, opened 2021-01-28
fix movie list bug caused by new rules

phpgeekfr:


apps/financiers/financiers/src/app/marketplace/organization/title/title.component.ts
```diff
@@ -11,22 +13,27 @@ import { DynamicTitleService } from '@blockframes/utils/dynamic-title/dynamic-ti
...
-    ); // TODO query can be improved after issue #3498
+      switchMap(org => {
+        return this.service.valueChanges(ref => ref
+          .where('orgIds', 'array-contains', org.id)
+          .where('storeConfig.status', '==', 'accepted')
+          .where('storeConfig.appAccess.financiers', '==', true)
+          .orderBy('_meta.createdAt', 'desc')
+          )
+      }),
+      map(movies => movies.sort((a, b) => sortMovieBy(a, b, 'Production Year')))
```
GrandSchtroumpf, line 35: What's the point of ordering by createdAt if you order by production year after. OrderBy production year direction in the query instead.

libs/organization/src/lib/organization/components/card/card.component.ts
```diff
@@ -23,7 +23,7 @@ export class OrganizationCardComponent implements OnInit {
   constructor(private movieService: MovieService, private routerQuery: RouterQuery) { }
 
   ngOnInit() {
-    this.orgMovieCount$ = this.movieService.valueChanges(fromOrg(this.org.id)).pipe(
+    this.orgMovieCount$ = this.movieService.valueChanges(fromOrgAndAccepted(this.org.id)).pipe(
       map(movies =>
         movies.filter(movie => movie.storeConfig.status === "accepted" && movie.storeConfig.appAccess[this.app]).length
```
GrandSchtroumpf, line 28: You don't need to filter by accepted anymore then.

firestore.indexes.json
```diff
@@ -3,18 +3,11 @@
     "collectionGroup": "movies",
     "queryScope": "COLLECTION",
     "fields": [
```
GrandSchtroumpf, line 5: You shouldn't remove the existing indexes, they are used for the cms on the homepage.

GrandSchtroumpf (changes requested):

apps/financiers/financiers/src/app/marketplace/organization/title/title.component.ts
GrandSchtroumpf: Can you modify the `sortMovieBy` with : 
```typescript
const sortMovieBy = (a: Movie, b: Movie) => (field: string) => {
  ...
}
```
Like that you can do : 
```typescript
map(movies => movies.sort(sortMovieBy('Production Year')))
```

GrandSchtroumpf (approved):

PR #4834, opened 2021-01-29
fix notification creation (missing email & uid)

phpgeekfr:


phpgeekfr:
@george43g this will prevent next notification of this kind to miss email  or uid but for exinsting documents in this situation you have two solutions:
- dont take thise kind of document into account and skip it from anon db
- anonomize name with random string

GrandSchtroumpf (approved):

PR #4843, opened 2021-02-02
Notifications & Emails

Mussuranae:
Continue to work on notification and emails
- orgAppAccessChanged

### DIFF
--- a/apps/backend-functions/src/orgs.ts
+++ b/apps/backend-functions/src/orgs.ts
@@ -19,6 +19,7 @@ import { ErrorResultResponse } from './utils';
 import { cleanOrgMedias } from './media';
 import { Change, EventContext } from 'firebase-functions';
 import { algolia, deleteObject, storeSearchableOrg, findOrgAppAccess, hasAcceptedMovies, storeSearchableUser } from '@blockframes/firebase-utils';
+import { getDocument } from './data/internals';
 
 /** Create a notification with user and org. */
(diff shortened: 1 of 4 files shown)


apps/backend-functions/src/orgs.ts
```diff
@@ -287,14 +288,20 @@ export const accessToAppChanged = async (
 
   const adminIds = await getAdminIds(orgId);
   const admins = await Promise.all(adminIds.map(id => getUser(id)));
-  const appKey = await getOrgAppKey(orgId);
-  const appUrl = await getAppUrl(orgId);
 
+  const notifications : NotificationDocument[] = [];
   await Promise.all(admins.map(admin => {
```
phpgeekfr, line 293: you don't need await Promise.all anymore

apps/backend-functions/src/orgs.ts
```diff
@@ -287,14 +288,20 @@ export const accessToAppChanged = async (
...
-    const template = organizationAppAccessChanged(admin, appUrl);
-    return sendMailFromTemplate(template, appKey).catch(e => console.warn(e.message));
+    const notification = createNotification({
+      toUserId: admin.uid,
+      docId: admin.orgId,
+      type: 'orgAppAccessChanged'
```
phpgeekfr, line 297: add attribute organization in notification. you will need it to display in front and also use it in `sendOrgAppAccessChangedEmail`

instead of  
```
  const app = await getOrgAppKey(recipient.orgId);
  const url = await getAppUrl(recipient.orgId);
```

phpgeekfr (approved):

PR #4846, opened 2021-02-02
Feat 4046 fix

phpgeekfr:


libs/notification/src/lib/+state/notification.store.ts
```diff
@@ -41,56 +41,76 @@ export class NotificationStore extends EntityStore<NotificationState, Notificati
...
         return {
-          _meta : {...notification._meta, createdAt: toDate(notification._meta.createdAt)},
+          _meta: { ...notification._meta, createdAt: toDate(notification._meta.createdAt) },
           message: `Your organization was accepted by the ${this.appName} team.`,
           imgRef: notification.organization?.logo,
           placeholderUrl: 'empty_organization.webp',
           url: `/c/o/organization/${notification.organization.id}/view/org`,
         };
-      case 'invitationFromUserToJoinOrgDecline':
+      case 'invitationFromUserToJoinOrgDecline': // @TODO #4046 handle or not ?
```
Mussuranae, line 52: I don't know if someone can refuse an invitation. If users haven't an account, the invitation will create one and at their first connexion and they will join automatically. And normally, we can't send invitation to join organization if people have already an org now. Maybe it will change when we will handle the refactor of the connexion flow...

libs/notification/src/lib/+state/notification.store.ts
```diff
@@ -41,56 +41,76 @@ export class NotificationStore extends EntityStore<NotificationState, Notificati
...
         const organizationName = orgName(notification.organization) || 'Organization with no name';
         return {
-          _meta : {...notification._meta, createdAt: toDate(notification._meta.createdAt)},
+          _meta: { ...notification._meta, createdAt: toDate(notification._meta.createdAt) },
           message: `${organizationName} submitted a contract.`,
           placeholderUrl: 'gears.webp',
           url: `/c/o/dashboard/deals/${notification.docId}`, // TODO check url : see  #2716
         };
-      case 'contractInNegotiation':
+      case 'contractInNegotiation': // @TODO #4046 handle or not ? or clean
```
Mussuranae, line 104: I think it would be better to remove this one and the one above, and Max and François will create new one when they will work on it. But for now, it's dead code no ?

libs/notification/src/lib/+state/notification.store.ts
```diff
@@ -107,12 +127,13 @@ export class NotificationStore extends EntityStore<NotificationState, Notificati
...
         })
         return {
-          _meta : {...notification._meta, createdAt: toDate(notification._meta.createdAt)},
+          _meta: { ...notification._meta, createdAt: toDate(notification._meta.createdAt) },
           message: `Your project was successfully published on the marketplace.`,
           imgRef: this.getPoster(notification.docId),
           placeholderUrl: 'empty_poster.webp',
           url: `/c/o/dashboard/title/${notification.docId}/main`,
         };
+      // @TODO case : 'orgAppAccessChanged' ?? sinon a mettre dans appendNotificationSettings
```
Mussuranae, line 136: I think we have to do it as we propose to user to see all notifications in app or by email. Except for the "ghost" notification, all the others has to appear somewhere in the front.

Mussuranae (approved):

PR #4849, opened 2021-02-03
Clean packages

GrandSchtroumpf:
update packages
cleanup after


libs/media/src/lib/+state/media.model.ts
```diff
@@ -24,7 +23,7 @@ export function extractMediaFromDocumentBeforeUpdate(
     MoviePictureAdminForm
 ): { documentToUpdate: any, mediasToUpload: HostedMediaFormValue[] } {
 
-  const cleanedDocument = cloneDeep(form.value);
+  const cleanedDocument = JSON.parse(JSON.stringify(form.value));
```
phpgeekfr, line 26: this is same behavior than `cloneDeep` ?

package.json
```diff
@@ -8,7 +8,6 @@
   "scripts": {
     "ng": "nx",
     "start": "nx serve --hmr --disable-host-check",
```
phpgeekfr, line 10: you should also update `apps/backend-functions/src/package.json` accordingly if some libs are also in it

phpgeekfr (approved):

libs/media/src/lib/+state/media.model.ts
GrandSchtroumpf: yes. Also this function will be removed by @pldespaigne's PR.

PR #4854, opened 2021-02-03
update cloud functions package.json

GrandSchtroumpf:
Move to node 12
Remove unused dependancies


apps/backend-functions/src/package.json
```diff
@@ -42,11 +27,7 @@
     "faker":"5.1.0",
     "generate-password": "^1.4.2",
     "jwplatform": "0.0.5",
-    "line-reader": "^0.4.0",
```
phpgeekfr, line 45: Can also be removed from main package.json i guess

apps/backend-functions/src/package.json
```diff
@@ -42,11 +27,7 @@
...
     "lodash": "^4.17.11",
-    "request": "^2.88.0",
```
phpgeekfr, line 47: same

apps/backend-functions/src/package.json
```diff
@@ -4,24 +4,10 @@
   "license": "MIT",
   "main": "./main.js",
   "engines": {
-    "node": "10"
+    "node": "12"
```
phpgeekfr, line 7: @george43g can this be a problem inside CI ? DockerFile is using node 10

phpgeekfr (approved):

apps/backend-functions/src/package.json
george43g: not really... it's only if we're running jest tests of cloud functions, then it's useful to use the same version of node for the tests (CI container) as is used in the actual cloud functions

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.2",
   "license": "MIT",
   "engines": {
-    "node": "10"
+    "node": "12"
```
george43g, line 6: why is this here at all?

george43g (approved):
nice work!

PR #4856, opened 2021-02-03
Issue excel import

fritzschoff:


libs/contract/src/lib/contract/+state/contract.model.ts
```diff
@@ -252,3 +252,52 @@ export function getValidatedContracts(contracts: Contract[]): Contract[] {
   const validStatus = 'paid' || 'waitingpayment' || 'accepted';
   return contracts.filter(contract => contract.lastVersion.status === validStatus)
 }
+
+///////////////////////////// NEW DATA MODEL OF CONTRACT
+
+export function createMandate(params?) {
```
phpgeekfr, line 258: seems not used and not typed. I guess it is in WIP ?

libs/contract/src/lib/term/+state/term.model.ts
```diff
@@ -0,0 +1,15 @@
...
+  titleId: string;
+  orgId: string;
+  territories: string[];
+  media: string[];
+  exclusive: boolean;
+  duration: { from: Date, to: Date };
+  // End Query //
+  // Vanity information
+  languages: Record<string, { subtitle: boolean }>;
+  criteria: any[];
```
phpgeekfr, line 13: is `any` ok ?

libs/contract/src/lib/term/+state/term.model.ts
```diff
@@ -0,0 +1,15 @@
+export interface Term {
+  id: string;  // Use same id than right & terms
+  // Start Query //
+  titleId: string;
+  orgId: string;
+  territories: string[];
+  media: string[];
+  exclusive: boolean;
+  duration: { from: Date, to: Date };
```
phpgeekfr, line 9: Since it has dates, we would need a TermBase<D> in term.firstore.ts  and then 
```
export interface Term extends TermBase<Date>
```

libs/utils/src/lib/static-model/static-model.ts
```diff
@@ -1931,3 +1931,8 @@ export function getISO3166TerritoryFromSlug(slug: Territory) {
     fr: territoriesFR[territory],
   }
 }
+
+export function getKeyFromValue(scope: Scope, searchValue: string) {
```
phpgeekfr, line 1935: you cannot use `getKeyIfExists` instead ?

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -75,22 +70,23 @@ export class ViewExtractedContractsComponent implements OnInit {
...
-    const titlesFieldsCount = Object.keys(SpreadSheetContractTitle).length / 2; // To get enum length
     const matSnackbarRef = this.snackBar.open('Loading... Please wait', 'close');
     for (const spreadSheetRow of sheetTab.rows) {
-      // CONTRACT ID
-      // Create/retreive the contract
-      let contract = createContract();
+      const trimmedRow = spreadSheetRow.map(row => {
+        if (typeof row === 'string') row.trim()
+        return row
+      })
```
phpgeekfr, line 78: in `spreadSheetRow.map(row => {`  , `row` is confusing . I guess it is a `cell`

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -99,352 +95,228 @@ export class ViewExtractedContractsComponent implements OnInit {
...
-            }
+          /* Create term */
+          const term = createTerm({ orgId: this.orgQuery.getActiveId(), titleId: contract.titleId })
+          if (trimmedRow[SpreadSheetContract.territories].length) {
+            const territoryValues: TerritoryValue[] = (trimmedRow[SpreadSheetContract.territories]).split(this.separator)
+            const territories = territoryValues.map(territory => getKeyFromValue('territories', territory.trim()))
+            term.territories = territories;
           } else {
             importErrors.errors.push({
               type: 'warning',
```
phpgeekfr, line 147: this should not be `error` then ?

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -99,352 +95,228 @@ export class ViewExtractedContractsComponent implements OnInit {
...
+          } else {
             importErrors.errors.push({
-              type: 'error',
-              field: 'titleIds',
-              name: 'Film Code',
-              reason: `Duplicate film code found : ${titleDetails.titleId}`,
+              type: 'warning',
+              field: 'term.language.dubbed',
+              name: 'Language dubbed',
+              reason: `Please provide dubbed version if available.`,
```
phpgeekfr, line 224: unnecessary back tick "`"

phpgeekfr (commented):
So this PR is about cleaning unused import files and prepare the new contract import with new data model. Is this right ?

libs/contract/src/lib/term/+state/term.model.ts
GrandSchtroumpf: We don't actually know what it'll look like. Let's discuss that on the catalog-app channel

libs/contract/src/lib/contract/+state/contract.model.ts
GrandSchtroumpf: Max, please type your function even if it's a WIP ;)

libs/contract/src/lib/term/+state/term.model.ts
fritzschoff: true but we will see where we go with the data model. So I will leave it as it is for now

libs/utils/src/lib/static-model/static-model.ts
fritzschoff: I was looking for that function all over! Thanks!

phpgeekfr (approved):

Issue #4837, opened 2021-02-01
Connexion issue


Jackseed:
We have a good number of users that reported a connexion issue. The workflow is always the same but the bug doesn't appear every time.
Steps to reproduce:
1. Create an org through classic sign up flow
2. Invite a new user to that org through CRM as blockframes admin (`/c/o/admin/panel/organization/orgId`)
3. Try to connect as if you were that new user with the pre-generated password received by email

Bug appears step 3, users receive a snackbar message "The password is invalid or the user does not have a password".

SamirTouira:
Finally, after many tries, impossible to reproduce this. However, when I copy/paste the pregenerated password with the space at the start in the credentials email, I got the same error than these users. I thought about one thing, simply my question is : did you asked them if they have correctly copied the pregenerated password? By removing the space from the beginning of the password?

![image](https://user-images.githubusercontent.com/52285231/106791036-b41cbb80-6654-11eb-859a-818784d5b292.png)

Jackseed:
Ok thanks @SamirTouira, good idea for the space, maybe we can add something to remove spaces on password... @GrandSchtroumpf ?
Cause on the email when you double click on the password it doesn't take the space anyway

GrandSchtroumpf:
This is a little bit dangerous if the user actually has a space at the beginning or end of his password. For my part I've got a bunch of spaces in my pwd, but not at the beginning or end. @phpgeekfr do you know if it safe to trim a password ?

phpgeekfr:
@GrandSchtroumpf Yes it is safe to trim() the input password. same for the login input we can trim

SamirTouira:
Alright, the trim solution works. Thank you for your answers

Jackseed:
I reopen this issue because I'm a bit worried about changing password handling that fast. 
@phpgeekfr @GrandSchtroumpf will this be retro-compatible with old passwords that have white spaces in it?

phpgeekfr:
@Jackseed erf you are right .. It is indeed possible to have a spaces in password. I did'nt knew this was allowed by google

I created an account with password `     1` and it worked .

phpgeekfr:
so maybe we need to revert ? Now that we understand the problem we could tell users to not copy with first space from email ?

phpgeekfr:
But maybe we should prevent new users to create a password containing spaces ? Unfortunately we cannot know if there is some existing users in this situation..

Jackseed:
[...]

PR #4865, opened 2021-02-04
trim on password connexion for credentials

SamirTouira:
fix: #4837 

trim() on the password works correctly to avoid the space from the beginning or end.


libs/user/src/lib/auth/pages/login/login.component.ts
```diff
@@ -46,7 +46,7 @@ export class LoginComponent implements OnInit {
     }
     try {
       const { email, password } = signinForm.value;
-      await this.service.signin(email.trim(), password);
+      await this.service.signin(email.trim(), password.trim());
```
phpgeekfr, line 49: please do the same for `public async signup(signupForm: SignupForm) {`

phpgeekfr (approved):

PR #4871, opened 2021-02-05
list comp clean up

fritzschoff:


apps/catalog/catalog/src/app/dashboard/right/list/list.component.ts
```diff
@@ -84,8 +15,14 @@ function createContractTab(allContracts: Contract[]): ContractTab[] {
...
 export class RightListComponent {
-  public tabs$ = this.query.sales$.pipe(map(createContractTab));
-  public salesLoading$ = this.query.selectLoading();
+  columns = {
+    id: 'Id',
+    status: 'Status'
+  };
+  initialColumns = ['id', 'status'];
+  filter = new FormControl();
+  filter$: Observable<any> = this.filter.valueChanges.pipe(startWith(this.filter.value));
```
RemcoSimonides, line 24: Isn't this of type string?

RemcoSimonides (approved):

apps/catalog/catalog/src/app/dashboard/right/list/list.component.ts
fritzschoff: Well this is going to be a Contract Status when the data model is ready. BUt I can put a string now

Issue #4327, opened 2020-11-25
add a consents collection to save users consents when creating or accessing documents


phpgeekfr:
This issue is part of #4309 .

This functionality is for saving the consent when an user/org create a new document ( a movie for example) and also for saving the consent when an user/org want to access a file ( movie's financiers pdf documents for example).

**Steps ( 1 PR per step)**

- [x] create interface/models in appropriate location in code & update draw.io accordingly
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/80fedee1-3877-4a9c-a6b6-a33cbd379fd6)

- [x] create httpCallable function to save consent with parameters `consentType : 'access' | 'share' , ip: string, docId: string, filePath?: string` 
 This functions will retrieve uid and orgId from authContext (see existing examples) and create / update (push into access or share array )  the consent document with id = orgId.
Check if IP is accessible from httpsCallable call context, this will allow us to remove ip parameter from the callable function

- [x] update db rules to prevent anyone from writing or reading this doc

- [x] test with existing CrmFormDialogComponent 
This dialog will then be used to get user consent on movie creation and at each time users wants to access a pdf file.

- [x] implement it where it should be (maybe see with @Jackseed )

Jackseed:
> implement it where it should be (maybe see with @Jackseed )
We shall save cookie consent here.
It also should be implemented on MF before submitting a movie on a movie form (dashboard) and before downloading financial elements on movie view (marketplace). But I think we don't have the modale existing yet since we wanted to have this consent saved first

phpgeekfr:
@Jackseed 
>We shall save cookie consent here.
This collection is to save consent about sharing or downloading a file/ a db document and is done at the  org level.

Saving cookie consent is individual (on user, not org)  and should be saved on userDocument

Jackseed:
Ok then it's only for MF, I'll create a dedicated issue to have this modale and refer to that issue to save consent

PR #4878, opened 2021-02-09
consents collection assertFails consents.spec.ts

SamirTouira:
This PR is related to: #4327 

The tests goes on the right way for read or right a consent document.


apps/rules/src/tests/firestore/consents.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+import { testFixture } from './fixtures/data';
+import { Firestore, initFirestoreApp } from '@blockframes/testing/firebase/functions';
+import { Consents } from '@blockframes/consents/+state/consents.firestore';
+
+describe('Consents Rules Tests', () => {
+  const projectId = `consentrules-spec-${Date.now()}`;
+  let db: Firestore;
+
+  describe('Any user', () => {
+    const newConsentId = 'CI-007';
```
phpgeekfr, line 11: you can move this line inside ` test('should not be able to write consent collection', async () => `

phpgeekfr (approved):

Issue #4815, opened 2021-01-25
Analytics refactoring


Jackseed:
## Objectives
Analytics need refactoring with 3 objectives in mind:
- debug the existing stats for users
- speed up stats creation and loading for users
- start an admin dashboard with general analytics

## Battle plan
1. [Store on firebase users’ analytics](https://github.com/blockframes/blockframes/issues/4693) & [debug existing queries](https://github.com/blockframes/blockframes/issues/4677)
A list of analytics data displayed to users is [here](https://docs.google.com/spreadsheets/d/1VXZKbd1KPE7GPIjsFA_h6G-_xl166p1IJz58suzhoVQ/edit#gid=1148924705).
2. [Get firebase data on BIgQuery](https://github.com/blockframes/blockframes/issues/3772) to be able to join analytics to firebase 
- Start with users collection as a test
3. Create admin queries on BigQuery
- I'll add an issue with the most interesting stats for which we shall already have the data needed 
4. Create dashboard on Data Studio 
- Following #3 would help business side to look at that data, should be done with the few stats created above

**Nice to have:**

5. Implement Yandex Metrica for hit maps #3317 

6. Add parameters on page_view event
- Need to add `orgId` and `movieId` on some page view events to facilitate queries

7. Create new click events
- Some UX analysis need users' click monitoring, we should start first with a test on landing & sign up page

RemcoSimonides:
To do:
- [ ] Combine intraday events table with the other table (scheduled query)
- [ ] Fetch data from Sendgrid

phpgeekfr:
@RemcoSimonides The 3 issues of this epic are closed. Does it means that we can close this epic?

RemcoSimonides:
@phpgeekfr No, not yet. What we have currently is the bare minimum. When there's some more time I can continue with this one.

PR #4888, opened 2021-02-10
4815 Embed user dashboard

RemcoSimonides:
Issue #4815

image

### DIFF
--- a/libs/admin/src/lib/admin-panel/pages/user/user.component.html
+++ b/libs/admin/src/lib/admin-panel/pages/user/user.component.html
@@ -103,4 +103,14 @@
       </form>
     </article>
   </section>
+  <section fxLayout fxLayoutAlign="center">
+    <iframe 
+      width="780px"
+      height="1100px"
+      [src]="reportURL"
+      frameborder="0"
+      style="border:0"
+      allowfullscreen>
+    </iframe>
+  </section>
 </main>
--- a/libs/admin/src/lib/admin-panel/pages/user/user.component.html
+++ b/libs/admin/src/lib/admin-panel/pages/user/user.component.html
@@ -103,14 +103,13 @@
       </form>
     </article>
   </section>
-  <section fxLayout fxLayoutAlign="center">
-    <iframe 
-      width="780px"
-      height="1100px"
-      [src]="reportURL"
-      frameborder="0"
-      style="border:0"
-      allowfullscreen>
-    </iframe>
-  </section>
+  <ng-container *ngIf="!!dashboardURL">
+    <section fxLayout fxLayoutAlign="center">
+      <iframe
+        [src]="dashboard"
+        frameborder="0"
+        allowfullscreen>
+      </iframe>
+    </section>
+  </ng-container>
 </main>
--- a/libs/admin/src/lib/admin-panel/pages/user/user.component.html
+++ b/libs/admin/src/lib/admin-panel/pages/user/user.component.html
@@ -106,7 +106,7 @@
   <ng-container *ngIf="!!dashboardURL">
     <section fxLayout fxLayoutAlign="center">
       <iframe
-        [src]="dashboard"
+        [src]="dashboardURL"
         frameborder="0"
         allowfullscreen>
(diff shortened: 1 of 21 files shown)


libs/admin/src/lib/admin-panel/pages/user/user.component.ts
```diff
@@ -29,6 +30,8 @@ export class UserComponent implements OnInit {
   public userForm: UserAdminForm;
   private originalOrgValue: string;
 
+  public reportURL: SafeResourceUrl = this.sanitizer.bypassSecurityTrustResourceUrl('');
```
GrandSchtroumpf, line 33: Don't initialize it with no value. `*ngIf` on reportUrl instead in the html

libs/admin/src/lib/admin-panel/pages/user/user.component.html
```diff
@@ -103,4 +103,14 @@ <h3>Other settings</h3>
       </form>
     </article>
   </section>
+  <section fxLayout fxLayoutAlign="center">
+    <iframe 
+      width="780px"
+      height="1100px"
+      [src]="reportURL"
+      frameborder="0"
+      style="border:0"
```
GrandSchtroumpf, line 112: use css

libs/admin/src/lib/admin-panel/pages/user/user.component.html
```diff
@@ -103,4 +103,14 @@ <h3>Other settings</h3>
       </form>
     </article>
   </section>
+  <section fxLayout fxLayoutAlign="center">
+    <iframe 
+      width="780px"
```
GrandSchtroumpf, line 108: Use css

libs/admin/src/lib/admin-panel/pages/user/user.component.scss
```diff
@@ -9,3 +9,7 @@ img {
 main {
   margin: 40px;
 }
+
+iframe {
+  min-height: 105%;
```
GrandSchtroumpf, line 14: Why 105% ? Prefer 
```scss
width: 100%;
min-height: 100%;
```

libs/admin/src/lib/admin-panel/pages/user/user.component.ts
```diff
@@ -54,6 +58,11 @@ export class UserComponent implements OnInit {
 
       this.userForm = new UserAdminForm(this.user);
       this.isUserBlockframesAdmin = await this.userService.isBlockframesAdmin(this.userId);
+
+      const prms = JSON.stringify({ "ds2.user_id": this.userId });
+      const encodedPrms = encodeURIComponent(prms);
+      this.reportURL = this.sanitizer.bypassSecurityTrustResourceUrl(`https://datastudio.google.com/embed/reporting/1564ae35-5e86-4632-bfef-ef7f4db7a865/page/P9czB?params=${encodedPrms}`);
```
GrandSchtroumpf, line 64: Hmmm does it updates over time with the same url ?

libs/admin/src/lib/admin-panel/pages/user/user.component.scss
RemcoSimonides: I need a little more than `100%` in order to don't have a scroll-bar in the iframe (overflow: hidden doesn't work)

`width: 100%` is what I want because then the width doesn't have to change if the dashboard get wider. But the dashboard in the iframe is a fixed width (currently 780px). I could make it 100%, but it will look different on different screen sizes.

So it's either a fixed dashboard width of like 1920px ([something that's often used](https://gs.statcounter.com/screen-resolution-stats/desktop/europe)) and `100% width` in app. Or change the fixed width in the app everytime the dashboard gets wider.

libs/admin/src/lib/admin-panel/pages/user/user.component.ts
RemcoSimonides: Yes, it does

GrandSchtroumpf (approved):

PR #4893, opened 2021-02-10
big cleaning of invitation & events

phpgeekfr:


libs/event/src/lib/+state/event.service.ts
```diff
@@ -97,8 +95,9 @@ export class EventService extends CollectionService<EventState> {
 
     // remove possible undefined values to avoid
     // FirebaseError: [code=invalid-argument]: Unsupported field value: undefined
-    if (!e.movie) delete e.movie;
-    if (!e.organizedBy) delete e.organizedBy;
+    if (!!e.movie) delete e.movie;
+    if (!!e.organizedBy) delete e.organizedBy;
+    if (!!e.org) delete e.org;
```
Mussuranae, line 100: Is it not `if (!e.movie)` with only one `!` ? We want to delete the field is there is no value right ?

libs/firebase-utils/src/lib/firestoreMigrations/0053.ts
```diff
@@ -12,6 +12,21 @@ export async function upgrade(db: Firestore) {
...
+
+    if (!!data.meta?.organizerId) {
+      data.meta.organizerUid = data.meta.organizerId;
+      delete data.meta.organizerId;
+    }
+
+    // Cleaning fields that should not be here CF formatToFirestore in event.service
+    if (!!data.movie) delete data.movie;
+    if (!!data.organizedBy) delete data.organizedBy;
+    if (!!data.org) delete data.org;
```
Mussuranae, line 28: Sooo, if there is a movie on the event, we delete it ?

Mussuranae (approved):

libs/firebase-utils/src/lib/firestoreMigrations/0053.ts
phpgeekfr: yes, this is here to clean the bug introduced (and corrected in this PR ) in `libs/event/src/lib/+state/event.service.ts `

We had objects on invitations that are supposed to be only in fron

libs/event/src/lib/+state/event.service.ts
phpgeekfr: I think this was a typo. invitation.movie or invitation.organizeBy & invitation.org are only for front & should be deteled when we save on db

libs/firebase-utils/src/lib/firestoreMigrations/0053.ts
Mussuranae: Ah ok, thanks !

libs/event/src/lib/+state/event.service.ts
Mussuranae: perfect then !

PR #4895, opened 2021-02-10
proposal for Connexion issue #4837

phpgeekfr:
Quick proposal for Connexion issue #4837 to allow more verbosity in password creation.
This will help us to gather more information when the case described in #4837 occurs

### DIFF
--- a/apps/backend-functions/src/internals/users.ts
+++ b/apps/backend-functions/src/internals/users.ts
@@ -32,21 +32,21 @@ export const getOrInviteUserByMail = async (email: string, fromOrgId: string, in
     const user = await getDocument<PublicUser>(`users/${uid}`);
     return user || { uid, email }
   } catch {
-    const newUser = await createUserFromEmail(email);
-
-    // User does not exists, send him an email.
-    const fromOrg = await getDocument<OrganizationDocument>(`orgs/${fromOrgId}`);
-    const urlToUse = applicationUrl[app];
-
-    const templateId = templateIds.user.credentials[invitationType];
-    const template = userInvite(email, newUser.password, orgName(fromOrg), urlToUse, templateId, eventData);
-
     try {
+      const newUser = await createUserFromEmail(email);
+
+      // User does not exists, send him an email.
+      const fromOrg = await getDocument<OrganizationDocument>(`orgs/${fromOrgId}`);
+      const urlToUse = applicationUrl[app];
+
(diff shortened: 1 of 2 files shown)


SamirTouira:
What it means verbosity? Is this useful if an account with a pre-generated password is resent if an user encounter again the issue #4837 ?

phpgeekfr:
I'm not event sure this will be of any help, since the user is created, it means that we don't enter into the `catch`  I added.
Maybe we should  display generated password in functions log to be able to compare ? Any security issue about this ?

phpgeekfr:
> What it means verbosity ?

Verbosity level is  the amount of data displayed in logs for a particular piece of code. 
Increasing verbosity == adding more console.log or console.error to better see what occured

SamirTouira:
> > What it means verbosity ?
> 
> Verbosity level is the amount of data displayed in logs for a particular piece of code.
> Increasing verbosity == adding more console.log or console.error to better see what is going on

Perfectly understood, thank you.

I'm not sure if I'm able to be a reviewer and approve alone this, but your temporary solution to find the right issue of this is a good idea for me

apps/backend-functions/src/internals/users.ts
```diff
@@ -59,14 +59,18 @@ export const createUserFromEmail = async (email: string): Promise<{ user: Public
...
   const user = await auth.createUser({
     email,
     password,
     emailVerified: true,
     disabled: false
+  }).catch(e => {
+    throw new Error(`There was an error while creating user (email: "${email}" | password: "${password}"): ${e.message}`);
   });
 
+  console.log(`Successfuly created user "${user.uid}" with email : "${email}" and password: "${password}"`);
```
GrandSchtroumpf, line 72: Why not using the firebase native logger ?

GrandSchtroumpf (approved):

apps/backend-functions/src/internals/users.ts
phpgeekfr: What is this ?

SamirTouira: I'm curious but using logger.log instead of console.log is a better refactoring for which reason? Firebase doc evoke both of them

phpgeekfr:
I updated PR with logger.info, first time we use it on project, let's try ;)

SamirTouira (approved):

phpgeekfr:
image

PR #4925, opened 2021-02-11
Unsubscribe group

### LINKED ISSUE #4710: Sendgrid - Include unsubscribe link directly in email core
Currently Unsubscribe link is added on the footer email through Subscription tracking (https://app.sendgrid.com/settings/tracking). This way we can't change text format and adapt it with our template. 

Instead, we should use their replacement tag that should enable use to include an unsubscribe link directly in our template. I tried to do it without success, would be good that a dev tries it. 

/!\ Be careful when touching unsubscribe settings as it's currently used on prod. Best would be to test with another Sendgrid account.

Mussuranae:
Add a new parameter to let us send to sendgrid an group id. It will allow us to split email into several group and users will be able to unsubscribe to some group or every group depending on which emails they want to receive.

Close #4710

### DIFF
--- a/apps/backend-functions/src/internals/email.ts
+++ b/apps/backend-functions/src/internals/email.ts
@@ -14,19 +14,19 @@ import { EmailJSON } from '@sendgrid/helpers/classes/email-address';
  *
  * Handles development mode: logs a warning when no sendgrid API key is provided.
  */
-export async function sendMail({ to, subject, text }: EmailRequest, from: EmailJSON = getSendgridFrom(), unsubId?: number): Promise<any> {
+export async function sendMail({ to, subject, text }: EmailRequest, from: EmailJSON = getSendgridFrom(), groupId?: number): Promise<any> {
   const msg: MailDataRequired = {
     from,
     to,
     subject,
     text,
-    asm: { groupId: unsubId }
+    asm: { groupId: groupId }
   };
 
   return send(msg);
 }
 
-export function sendMailFromTemplate({ to, templateId, data }: EmailTemplateRequest, app: App, unsubId?: number): Promise<any> {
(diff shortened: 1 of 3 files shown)


apps/backend-functions/src/notification.ts
```diff
@@ -23,12 +23,14 @@ import {
...
 } from './templates/mail';
-import { templateIds } from './templates/ids';
+import { templateIds, unsubscribeGroupIds } from './templates/ids';
 import { canAccessModule, orgName } from '@blockframes/organization/+state/organization.firestore';
 import { App, applicationUrl } from '@blockframes/utils/apps';
 
 // @TODO (#2848) forcing to festival since invitations to events are only on this one
 const eventAppKey: App = 'festival';
+// This is for letting user unsubscribe from every email except the critical ones as reset password.
+const unsubscribeId = unsubscribeGroupIds.allExceptCriticals
```
phpgeekfr, line 33: missing `;`

apps/backend-functions/src/internals/email.ts
```diff
@@ -14,18 +14,19 @@ import { EmailJSON } from '@sendgrid/helpers/classes/email-address';
...
     subject,
     text,
+    asm: { groupId: unsubId }
   };
 
   return send(msg);
 }
 
-export function sendMailFromTemplate({ to, templateId, data }: EmailTemplateRequest, app: App): Promise<any> {
+export function sendMailFromTemplate({ to, templateId, data }: EmailTemplateRequest, app: App, unsubId?: number): Promise<any> {
```
phpgeekfr, line 29: maybe rename `unsubId` to `groupId`

phpgeekfr (approved):

Issue #4916, opened 2021-02-11
LinkFile is crashing when upsertWatermark is triggered


phpgeekfr:
When a new user is created, we generate a watermark for him with `upsertWatermark` .
At watermark creation, `linkFile` backend-function is triggered and crash saying that file meta-data are incorrect.

![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/54612bda-50bb-4479-929b-74b1ea99b258)

RemcoSimonides:
Can't reproduce.
Tried regular sign-up flow and via event invite. Both ways I got the watermark without an error in the logs

PR #4926, opened 2021-02-11
4916 Fix invalid metadata for watermark

RemcoSimonides:
Fix issue #4916

### DIFF
--- a/libs/firebase-utils/src/lib/watermark.ts
+++ b/libs/firebase-utils/src/lib/watermark.ts
@@ -32,10 +32,9 @@ export async function upsertWatermark(user: PublicUser, bucketName: string, stor
 
   const ref = `${privacy}/users/${user.uid}/watermark/${user.uid}.svg`;
   const file = storage ? storage.bucket(bucketName).file(ref) : admin.storage().bucket(bucketName).file(ref);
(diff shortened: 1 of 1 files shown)


libs/firebase-utils/src/lib/watermark.ts
```diff
@@ -32,10 +32,9 @@ export async function upsertWatermark(user: PublicUser, bucketName: string, stor
 
   const ref = `${privacy}/users/${user.uid}/watermark/${user.uid}.svg`;
   const file = storage ? storage.bucket(bucketName).file(ref) : admin.storage().bucket(bucketName).file(ref);
+  const metadata = { metadata: { uid: user.uid, privacy: 'public', collection: 'users', field: 'watermark', docId: user.uid } };
 
-  await new Promise(res => {
-    file.createWriteStream({ contentType: 'image/svg+xml' }).end(watermark, () => res());
-  });
+  await file.save(watermark, { metadata });
```
RemcoSimonides, line 37: I could also do it with `file.createWriteStream()` but I like `file.save()` more. contentType is set correctly in Storage.

pldespaigne (approved):

PR #4927, opened 2021-02-12
keep stabilizing

pldespaigne:
more stabilizing


apps/backend-functions/src/media.ts
```diff
@@ -360,12 +360,12 @@ export async function cleanMovieMedias(before: MovieDocument, after?: MovieDocum
 
   const mediaToDelete: string[] = [];
   if (!!after) { // Updating
-    if (needsToBeCleaned(before.banner, after.banner)) {
-      mediaToDelete.push(before.banner);
+    if (needsToBeCleaned(before.banner.storagePath, after.banner.storagePath)) {
```
GrandSchtroumpf, line 363: Can't you update `needsToBeCleaned` to do the `.storagePath` directly so you can do : 
```typescript
if (needsToBeCleaned(before.banner, after.banner))
```

GrandSchtroumpf (approved):

PR #4938, opened 2021-02-16
4002 refactor media

RemcoSimonides:

### DIFF
--- a/libs/movie/src/lib/movie/components/promotional-links/promotional-links.pipe.ts
+++ b/libs/movie/src/lib/movie/components/promotional-links/promotional-links.pipe.ts
@@ -2,6 +2,7 @@ import { Pipe, PipeTransform, NgModule } from '@angular/core';
 import { Movie } from '@blockframes/movie/+state/movie.model';
 import { MediaService } from '@blockframes/media/+state/media.service';
 import { promotionalElementTypes } from '@blockframes/utils/static-model/static-model';
+import { StorageFile } from '@blockframes/media/+state/media.firestore';
 
 @Pipe({
   name: 'promotionalLinks',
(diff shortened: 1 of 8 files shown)


libs/movie/src/lib/movie/components/promotional-links/promotional-links.pipe.ts
```diff
@@ -15,10 +16,15 @@ export class PromotionalLinksPipe implements PipeTransform {
 
   async transform(links: string[], movie: Movie): Promise<any[]> {
     const _links = await Promise.all(links.map(async link => {
-      const ref = (movie.promotional[link] as string);
+      const ref = (movie.promotional[link] as StorageFile | string);
       const useImgIx = ['scenario', 'presentation_deck', 'moodboard'].includes(link);
+
+      if (!useImgIx && typeof ref !== 'string') {
+        throw new Error('Referene should be of type string')
+      }
```
RemcoSimonides, line 24: When useImgIx is used, the type of ref should always be of type `StorageFile`. But just to be sure I added this catch

libs/movie/src/lib/movie/components/promotional-links/promotional-links.pipe.ts
```diff
@@ -15,10 +16,15 @@ export class PromotionalLinksPipe implements PipeTransform {
 
   async transform(links: string[], movie: Movie): Promise<any[]> {
     const _links = await Promise.all(links.map(async link => {
-      const ref = (movie.promotional[link] as string);
+      const ref = (movie.promotional[link] as StorageFile | string);
```
pldespaigne, line 19: Promotional link are a legacy feature, we keep them for now to be backward compatible.
But link are always strings and those strings should always be full URLs, so no need for ImgIx nor StorageFile

pldespaigne (changes requested):

pldespaigne (approved):

PR #4943, opened 2021-02-16
Replace file and image upload component

RemcoSimonides:


libs/ethers/src/lib/wallet/components/wallet-import-key-form/wallet-import-key-form.component.html
```diff
@@ -15,10 +15,10 @@ <h2>Import your Key</h2>
       </mat-slide-toggle>
     </div>
     <ng-container *ngIf="!isMnemonic; else: mnemonicImport">
-      <file-uploader
+      <!-- <file-uploader
         allowedFileType="json"
         (uploaded)="importedJson($event)"
-      ></file-uploader>
+      ></file-uploader> -->
     </ng-container>
```
pldespaigne, line 22: please delete, better to not leave dead code and on top of that dead code that will crash if un-commented ^^

pldespaigne (changes requested):

pldespaigne (approved):

PR #4944, opened 2021-02-16
clean

pldespaigne:
remove deprecated code


libs/media/src/lib/file/explorer/list/list.component.ts
```diff
@@ -26,9 +26,8 @@ export class FileExplorerListComponent {
 
   public openView(item: Partial<StorageFile>, event: Event) {
     event.stopPropagation();
-    const ref = item.storagePath;
-    if (ref) {
-      this.dialog.open(FilePreviewComponent, { data: { ref }, width: '80vw', height: '80vh' });
+    if (item) {
```
RemcoSimonides, line 29: Didn't you teach me to use `!!`?

RemcoSimonides (approved):
LGTM

libs/media/src/lib/file/explorer/list/list.component.ts
pldespaigne: you are right, I just updated the code without using my brain ^^

PR #4945, opened 2021-02-16
CI cleanup & refactor GCP env setup - unify methods

george43g:
- unify env selection method in CI
- simplify project checkout/env setup
- update env selection to set PROJECT_ID so that nx commands work reliably
- in preparation for generic env deployment / PFT being run from CI
- introduce caching for dependencies and repo folder
- cache NX build cache for faster builds
- use PFTBeta in CI PFT tasks
- manually disable maintenance since we no longer introduce 8min delay (rather than waiting 8mins in CI)
- simplify env config and delete unneeded files
- refactor env selection algo
- clean up package.json consistent with ci config


apps/backend-ops/src/select-environment.ts
```diff
@@ -1,69 +1,124 @@
...
+/**
+ * Will return true if SAK project corresponds to given project.
+ */
+function SAKIsCorrect(projectId: string) {
   let key: any;
   try {
-    key = getServiceAccountObj(process.env.GOOGLE_APPLICATION_CREDENTIALS);
+    key = getServiceAccountObj(process.env[SAKKeyName]);
+    console.log('Current SAK projectId:', key.project_id);
   } catch (e) {}
```
phpgeekfr, line 21: no need to display error ?

phpgeekfr (approved):
LGTM

apps/backend-ops/src/select-environment.ts
george43g: nope, if it fails it means its not set, in which case we find it and try to set it

PR #4956, opened 2021-02-17
Org Algolia

### LINKED ISSUE #4946: Add orgActivity into algolia orgs index
To allow us to display org main data to user when he is creating his account and wants to join existing org

Mussuranae:
Add the activity of the org on its algolia index
Fix #4946 
Screen of an org after the `upgradeAlgoliaOrgs` command
image


phpgeekfr (changes requested):
you need to update `onOrganizationUpdate` also if org activity has changed

phpgeekfr (approved):

PR #4963, opened 2021-02-18
Fix event creation issue.

xmano:
Fix event creation issue with Create screening

### DIFF
--- a/apps/catalog/catalog-e2e/src/integration/marketplace/create-wishlist.spec.ts
+++ b/apps/catalog/catalog-e2e/src/integration/marketplace/create-wishlist.spec.ts
@@ -18,7 +18,8 @@ beforeEach(() => {
   cy.wait(1000);
 });
 
-describe('Test wishlist features from library, detail page',  () => {
+//TODO: Rework the tests after App re-write
+describe.skip('Test wishlist features from library, detail page',  () => {
   it('User clears all movies from Wish List', () => {
     const p1 = new HomePage();
     acceptCookie();
--- a/apps/catalog/catalog-e2e/src/integration/marketplace/create-wishlist.spec.ts
+++ b/apps/catalog/catalog-e2e/src/integration/marketplace/create-wishlist.spec.ts
@@ -18,7 +18,7 @@ beforeEach(() => {
   cy.wait(1000);
 });
 
-//TODO: Rework the tests after App re-write
+//TODO: Issue 4965 Rework the tests after App re-write
 describe.skip('Test wishlist features from library, detail page',  () => {
   it('User clears all movies from Wish List', () => {
     const p1 = new HomePage();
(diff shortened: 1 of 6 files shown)


libs/event/src/lib/form/edit-details/edit-details.component.html
```diff
@@ -1,7 +1,7 @@
 <form [formGroup]="form" fxLayout="column">
   <mat-form-field appearance="outline">
     <mat-label>Title</mat-label>
-    <input matInput formControlName="title" test-id="event-title"/>
+    <input test-id="event-title" matInput formControlName="title" />
```
RemcoSimonides, line 4: test-id needs to be the first attribute?

apps/catalog/catalog-e2e/src/integration/marketplace/create-wishlist.spec.ts
```diff
@@ -18,7 +18,8 @@ beforeEach(() => {
   cy.wait(1000);
 });
 
-describe('Test wishlist features from library, detail page',  () => {
+//TODO: Rework the tests after App re-write
```
RemcoSimonides, line 21: Add issue number so we know which issue is blocking

RemcoSimonides (approved):

libs/event/src/lib/form/edit-details/edit-details.component.html
xmano: it can be anywhere.. i generally prefer first for quick ID

apps/catalog/catalog-e2e/src/integration/marketplace/create-wishlist.spec.ts
xmano: https://github.com/blockframes/blockframes/issues/4965

apps/catalog/catalog-e2e/src/integration/marketplace/create-wishlist.spec.ts
```diff
@@ -18,7 +18,8 @@ beforeEach(() => {
   cy.wait(1000);
 });
 
-describe('Test wishlist features from library, detail page',  () => {
+//TODO: Issue 4965 Rework the tests after App re-write
+describe.skip('Test wishlist features from library, detail page',  () => {
```
phpgeekfr, line 22: We also have wishlist feature on festival. So I think this one should me moved to festival tests and fixed

phpgeekfr (commented):
LGTM but have a look at my comment

apps/catalog/catalog-e2e/src/integration/marketplace/create-wishlist.spec.ts
xmano: A separate PR will address that.

PR #4964, opened 2021-02-18
fixing algolia and init avails specs

fritzschoff:


libs/contract/src/lib/avails/avails.spec.ts
```diff
@@ -0,0 +1,204 @@
+import { Term } from '../term/+state/term.model';
+import { isLicensed, filterSales } from './avails';
+import { mandateTerms as acTerms } from './fixtures/mandateTerms'
+import { saleTerms as acSaleTerms } from './fixtures/saleTerms';
+
+describe('Avails', () => {
+    let mandateTerms: Term<Date>[] = (acTerms as unknown[]).map(parseTimestampsOnTerms)
+    let saleTerms: Term<Date>[] = (acSaleTerms as unknown[]).map(parseTimestampsOnTerms)
+
+    it(`Mandate test (territory)
```
GrandSchtroumpf, line 10: Create a function to return the text based on params so you don't have to repeat yourself

libs/contract/src/lib/avails/avails.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+        expect(ACRights).toBe(true);
+        const toDateSales = saleTerms.map(parseTimestampsOnTerms);
+        const avails = filterSales(
+            { duration: { to: new Date('06/30/2021'), from: new Date('01/01/2021') }, exclusive: true, territories: ['canada'], medias: ['freeTv'] },
+            toDateSales);
+        expect(avails).toBe(false);
+    });
+})
+
+function parseTimestampsOnTerms(term: Term<Date>): Term<Date> {
```
GrandSchtroumpf, line 200: can you rename it `toDate` ? It's easier to understand. Also the params should be `term: Term<Timestamp>` no ?

libs/contract/src/lib/avails/avails.ts
```diff
@@ -0,0 +1,39 @@
+import { Media, Territory } from "@blockframes/utils/static-model";
+import { Term } from "../term/+state/term.model";
+
+interface AvailsFilter { medias: Media[], duration: { from: Date, to: Date }, territories: Territory[], exclusive: boolean }
```
GrandSchtroumpf, line 4: format the interface on several lines please.

libs/contract/src/lib/term/+state/term.model.ts
```diff
@@ -1,6 +1,6 @@
 import { MovieLanguageSpecification } from "@blockframes/movie/+state/movie.firestore";
 
-export interface Term {
+export interface Term<T> {
```
GrandSchtroumpf, line 3: If you want you can do : `Terms<T extends Date | Timestamp = Date>`. Like that we only need to specify if it's Timestamp.

libs/firebase-utils/src/lib/algolia/algolia.ts
```diff
@@ -37,13 +37,15 @@ export function clearIndex(indexName: string, adminKey?: string) {
...
 
-export function setIndexConfiguration(indexName: string, config: AlgoliaConfig, adminKey?: string) {
+export function  setIndexConfiguration(indexName: string, config: AlgoliaConfig, adminKey?: string) {
   if (!algolia.adminKey && !adminKey) {
     console.warn('No algolia id set, assuming dev config: skipping');
     return Promise.resolve(true);
   }
 
-  return indexBuilder(indexName, adminKey).setSettings(config);
+  return indexBuilder(indexName, adminKey).setSettings({
```
GrandSchtroumpf, line 46: format on several lines for readability please.

package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "ng": "nx",
     "start": "nx serve --hmr --disable-host-check",
+    "start:catalog": "nx serve catalog --hmr",
```
GrandSchtroumpf, line 8: Do we really need to add that ? I feel like package.json is already too fat ^^

GrandSchtroumpf (approved):

package.json
fritzschoff: no fat shaming please ;)

PR #4969, opened 2021-02-18
Mano e2e nightly issue fix feb

xmano:
Fixes create event in "Organiser creates screening & invites 2 users to the screening"

image


xmano:
Other tests need to be worked on.. will do so subsequently

phpgeekfr (approved):

PR #4970, opened 2021-02-18
Email Input Implementation

Mussuranae:
Implement the new email input


libs/organization/src/lib/organization/+state/organization.model.ts
```diff
@@ -30,8 +30,11 @@ export function createOrganization(
 /** Convert an organization object into a public organization */
 export function createPublicOrganization(org: Partial<Organization>): PublicOrganization {
```
phpgeekfr, line 31: I'm not sure if this is a good idea to do this. Because it means that every time we use `createPublicOrganizationDocument` it will add new data  new unnecessary data (for createNotification()) for example.

Maybe we should make `getInvitationLinkedToEmail` return a Org instead of public org?

phpgeekfr: or make `getInvitationLinkedToEmail` return a custom object with what we need or a `AlgoliaOrganization` ?

Mussuranae: Yeah, you're right, I will change it and return an algoliaOrg

PR #4974, opened 2021-02-19
fix  TypeError: templateRef.createEmbeddedView is not a function 

phpgeekfr:


libs/ui/src/lib/list/page/list-page.component.html
```diff
@@ -2,7 +2,7 @@
   <ng-template pageBar targetId="app-bar">
     <div class="pageBar" fxLayout fxLayoutAlign="start center">
       <app-menu></app-menu>
-      <ng-template [ngTemplateOutlet]="appBar"></ng-template>
+      <ng-content [ngTemplateOutlet]="appBar"></ng-content>
```
phpgeekfr, line 5: Issue is solved but now sure if this is the right thing to do. Can you confirm @RemcoSimonides  ?

RemcoSimonides: `ngTemplateOutlet` is not an attribute of `ng-content`. You can remove the attribute and it will still work (somehow)
The source of the issue is caused by this change: https://github.com/blockframes/blockframes/commit/c495a445e66dae4e0c367985ffe3649e2f59349f
In which 
` @ContentChild(PageAppBarSearchDirective) appBar: PageAppBarSearchDirective;` doesn't have `{ read: TemplateRef }` in there. That's what is causing the error. If you add `{ read: TemplateRef }`, then `*ngIf="!!appBar"` doesn't work anymore.

@GrandSchtroumpf  do you have an idea how to fix this?

RemcoSimonides (changes requested):

phpgeekfr:
So I can just remove it ? Since it still works with only `<ng-content></ng-content>`

libs/ui/src/lib/list/page/list-page.component.html
```diff
@@ -2,7 +2,7 @@
   <ng-template pageBar targetId="app-bar">
     <div class="pageBar" fxLayout fxLayoutAlign="start center">
       <app-menu></app-menu>
-      <ng-template [ngTemplateOutlet]="appBar"></ng-template>
+      <ng-content></ng-content>
```
RemcoSimonides, line 5: Alright, I got it. Add a select
`<ng-content select="list-page-app-bar"></ng-content>`

RemcoSimonides (changes requested):

RemcoSimonides (approved):

PR #4982, opened 2021-02-19
fix twilio unit test

phpgeekfr:
fix test issue discovered in https://github.com/blockframes/blockframes/pull/4981


xmano (approved):
Can be simplified if you want:

testEvents[0].start = testEvents[0].end = new Date();

phpgeekfr:
funny part is that tests were passing depending of the current day time

PR #4984, opened 2021-02-22
many fixes

pldespaigne:
multiple bug fix


libs/ui/src/lib/preview-modal/preview.component.html
```diff
@@ -6,7 +6,7 @@
 
 <ng-template #dialogTemplate let-ref>
   <article fxLayout="column" fxLayoutAlign="center center">
-    <ng-container [ngSwitch]="ref | fileType">
+    <ng-container [ngSwitch]="ref.storagePath | fileType">
```
RemcoSimonides, line 9: It would be more neat if `ref` was renamed to `file`

RemcoSimonides (approved):

PR #4986, opened 2021-02-22
Fix screening invitation issue with notification not getting created

xmano:
Event was not getting created and I found an issue in code debugging with console.log :

https://github.com/blockframes/blockframes/blob/c1cff826f6a98dd4f14d7c03426126218d7718ad/libs/invitation/src/lib/form/user/user.component.ts#L82

fromOrg remains undefined for me.

### DIFF
--- a/apps/festival/festival-e2e/src/integration/dashboard/screening-invitations-2688.spec.ts
+++ b/apps/festival/festival-e2e/src/integration/dashboard/screening-invitations-2688.spec.ts
@@ -39,7 +39,7 @@ describe('Organiser invites other users to private screening', () => {
     p1.clickSignup();
   });
 
-  it('Organiser creates screening & invites 2 users to the screening', () => {
+  it.only('Organiser creates screening & invites 2 users to the screening', () => {
     signIn(users[UserIndex.Organiser]);
     acceptCookie();
 
--- a/libs/invitation/src/lib/form/user/user.component.ts
+++ b/libs/invitation/src/lib/form/user/user.component.ts
@@ -79,7 +79,9 @@ export class UserComponent implements OnInit {
       const emails = this.form.value.map(guest => guest.email);
       this.form.reset([]);
       this.sending.next(true);
-      const fromOrg = this.ownerOrgId ? await this.orgService.getValue(this.ownerOrgId) : undefined;
+      console.log(`E0-> Attend event : ${this.ownerOrgId}`);
+      const fromOrg = this.ownerOrgId ? undefined : await this.orgService.getValue(this.ownerOrgId);
+      console.log(`E-> Attend event : ${fromOrg} : ${JSON.stringify(emails)}`);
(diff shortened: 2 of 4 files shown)


xmano:
With this change : invitations were being received.

image

It might need some more debugging though. Need to analyze and recreate properly.

libs/invitation/src/lib/form/user/user.component.ts
```diff
@@ -79,7 +79,9 @@ export class UserComponent implements OnInit {
       const emails = this.form.value.map(guest => guest.email);
       this.form.reset([]);
       this.sending.next(true);
-      const fromOrg = this.ownerOrgId ? await this.orgService.getValue(this.ownerOrgId) : undefined;
+      console.log(`E0-> Attend event : ${this.ownerOrgId}`);
+      const fromOrg = this.ownerOrgId ? undefined : await this.orgService.getValue(this.ownerOrgId);
```
phpgeekfr, line 83: I don't understand this line `const fromOrg = this.ownerOrgId ? undefined : await this.orgService.getValue(this.ownerOrgId);`

If this.ownerOrgId is undefined, what is the point of `this.orgService.getValue(undefined) ` ? This will retreive all orgs

phpgeekfr: since we are making an invitation, fromOrg is mandatory and should always be set

apps/festival/festival-e2e/src/integration/dashboard/screening-invitations-2688.spec.ts
```diff
@@ -39,7 +39,7 @@ describe('Organiser invites other users to private screening', () => {
     p1.clickSignup();
   });
 
-  it('Organiser creates screening & invites 2 users to the screening', () => {
+  it.only('Organiser creates screening & invites 2 users to the screening', () => {
```
phpgeekfr, line 42: this was for testing purposes?

libs/invitation/src/lib/form/user/user.component.ts
phpgeekfr: if `this.ownerOrgId` is undefined (most of the time except in admin panel)  we use `fromOrg: Organization = this.orgQuery.getActive()`  in invitationService.invite

xmano: I guess there is a better fix available.. i can revert this

phpgeekfr: This is not a fix, this introduces a new bug

phpgeekfr: In this test context, fromOrg should stay undefined, and we use this.orgQuery.getActive() in `invitationService.invite`

xmano:
As I understand the fix provided here will not solve the real purpose of getting fromOrgId.. hence closing it.

PR #4995, opened 2021-02-23
Org request access

Mussuranae:
Add a page to let users ask for new app access when they connect on an app their org can't access.
It sends an email to C8 Admin to let them know an organisation asks for a new app access.
image

image


apps/festival/festival/src/app/festival.module.ts
```diff
@@ -29,6 +29,10 @@ const routes: Routes = createRoutes({
       path: 'dashboard',
       canActivate: [AppGuard],
       loadChildren: () => import('./dashboard/dashboard.module').then(m => m.DashboardModule)
+    },
+    {
+      path: 'request-access',
+      loadChildren: () => import('@blockframes/organization/pages/request-access/request-access.module').then(m => m.OrgRequestAccessModule)
     }
```
Mussuranae, line 36: Do we have to protect this page ?

apps/backend-functions/src/orgs.ts
```diff
@@ -276,3 +276,11 @@ export const accessToAppChanged = async (
     result: 'OK'
   };
 }
+
+/** Send an email to C8 Admins when an organization requests to access to a new platform */
+export const onRequestFromOrgToAccessApp = async (data: { app: App, orgId: string}) => {
```
phpgeekfr, line 281: maybe add a check on orgId and app .
You can also use context to check if user is logged in ( will prevent any user not logged in to call this functions)

apps/festival/festival/src/app/festival.module.ts
phpgeekfr: We could add a guard preventing to access this page if org already have app access to this app  but I don't think this is mandatory

phpgeekfr (approved):

PR #5015, opened 2021-02-25
Feat 4932 ter

phpgeekfr:


libs/user/src/lib/auth/pages/login/login.component.scss
```diff
@@ -51,5 +51,6 @@ mat-sidenav {
   a {
     font-weight: bold;
     cursor: pointer;
+    margin-left: 120px;
```
Mussuranae, line 54: Same, try with the flex layout

Mussuranae (approved):

PR #5016, opened 2021-02-25
Fix extra metadata overwrite

RemcoSimonides:


libs/movie/src/lib/movie/form/sales-pitch/sales-pitch.component.ts
```diff
@@ -25,4 +31,17 @@ export class MovieFormSalesPitchComponent {
     return this.form.promotional.get('salesPitch');
   }
 
+  ngOnInit() {
+    this.sub = this.salesPitch.get('description').valueChanges.subscribe(description => {
```
RemcoSimonides, line 35: Because description is on a different component than the file-uploader, we need to listen to it's changes and update the metadata of the file in the queue.

pldespaigne (approved):

PR #5028, opened 2021-02-26
Support new media structure

RemcoSimonides:
- [x] Migrate media on CMS
- [x] New media form

### DIFF
--- a/apps/cms/src/app/template/template.component.ts
+++ b/apps/cms/src/app/template/template.component.ts
@@ -8,6 +8,7 @@ import { CmsService, CmsParams } from '../cms.service'
 import { Subscription } from 'rxjs';
 import { sections as homeSection } from '../home';
 import { switchMap } from 'rxjs/operators';
+import { createStorageFile } from '@blockframes/media/+state/media.firestore';
 
 const templateSections = {
   home: (params: TemplateParams) => homeSection(params),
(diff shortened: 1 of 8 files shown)


apps/cms/src/app/template/template.component.ts
```diff
@@ -78,6 +79,33 @@ export class TemplateComponent implements OnInit, OnDestroy {
 
   save() {
     const params = this.route.snapshot.params as CmsParams;
-    this.service.save(this.form.value, params);
+    const value = this.updateMediaMetadata(this.form.value, params)
+    this.service.save(value, params);
+  }
+
+  updateMediaMetadata(value: CmsTemplate<Section>, params: CmsParams) {
```
RemcoSimonides, line 86: Only when saving we know which section has media. This is when we insert the correct index in `field`.

pldespaigne (approved):

apps/cms/src/app/template/template.component.ts
```diff
@@ -78,6 +79,33 @@ export class TemplateComponent implements OnInit, OnDestroy {
...
+    const { app, page, template } = params;
+    const getCollection = () => ['cms', app, page].filter(v => !!v).join('/');
```
GrandSchtroumpf, line 88: if you don't put any params, you don't need to create a function for that. The value of the collection won't change depending on the section : 
```typescript
const collection = ['cms', app, page].filter(v => !!v).join('/');
```

apps/cms/src/app/template/template.component.ts
```diff
@@ -78,6 +79,33 @@ export class TemplateComponent implements OnInit, OnDestroy {
...
+
+    value.sections.forEach((section, i) => {
+      let fields = [];
```
GrandSchtroumpf, line 91: Better to have a mapping for that : 
```typescript
// Outside of the component
const mediaFields = {
  banner: ['background', 'image']
  hero: ['background']
}
....
const fields = mediaFields[section._type] ?? [];
```

apps/cms/src/app/template/template.component.ts
```diff
@@ -78,6 +79,33 @@ export class TemplateComponent implements OnInit, OnDestroy {
...
+      if (section._type === 'banner') {
+        fields = ['background', 'image'];
+      } else if (section._type === 'hero') {
+        fields = ['background'];
+      }
+
+      fields.forEach(field => {
```
GrandSchtroumpf, line 98: Prefere `for (const field of fields)` over forEach if you don't need the index.

GrandSchtroumpf (approved):

PR #5029, opened 2021-02-26
George/refactor

george43g:
- ci maintenance delay fix


libs/firebase-utils/src/lib/maintenance.ts
```diff
@@ -28,8 +28,23 @@ export function startMaintenance(db?: FirebaseFirestore.Firestore) {
...
+  if (ago) {
+    const time = new Date(new Date().getTime() - (ago * 1000))
+    return maintenanceRef(db).set(
+      {
+        endedAt: admin.firestore.Timestamp.fromDate(time),
+        startedAt: null,
+      },
+      { merge: false }
+    );
+  }
```
phpgeekfr, line 47: Maybe a more elegant way :
```
export function endMaintenance(db?: FirebaseFirestore.Firestore, ago?: number) {
  if (process.env.BLOCKFRAMES_MAINTENANCE_DISABLED) return;

  let endedAt = admin.firestore.FieldValue.serverTimestamp();
  if (ago) {
    const time = new Date(new Date().getTime() - ago)
    endedAt = admin.firestore.Timestamp.fromDate(time);
  }
  return maintenanceRef(db).set(
    {
      endedAt,
      startedAt: null,
    },
    { merge: true }
  );
}
 ```

File: apps/backend-ops/src/main.ts [hunk 2]
```diff
@@ -26,7 +26,7 @@ async function runCommand() {
     case 'prepareForTestingBeta':
       await startMaintenance(db);
       await prepareForTestingBeta();
-      await endMaintenance(db);
+      await endMaintenance(db, 8 * 60);
```
phpgeekfr, line 29: can you use existing `EIGHT_MINUTES_IN_MS` and remove `(ago * 1000)` ?

phpgeekfr (approved):

PR #5040, opened 2021-03-01
add guard for request access page

Mussuranae:
Add a guard to check when we refresh the page if organization has now access to the platform and redirect the user if yes.

### DIFF
--- a/libs/organization/src/lib/organization/guard/request-access.guard.ts
+++ b/libs/organization/src/lib/organization/guard/request-access.guard.ts
@@ -40,9 +40,9 @@ export class RequestAccessGuard extends CollectionGuard<OrganizationState> {
               if (org.status === 'accepted') {
                 const app = getCurrentApp(this.routerQuery);
                 if (!!org.appAccess[app].marketplace) {
-                  return '/c/o/marketplace/home'
+                  return '/c/o/marketplace/home';
                 }
-                if (!!org.appAccess[app].dashboard) {
+                else if (!!org.appAccess[app].dashboard) {
                   return '/c/o/dashboard/home';
                 }
                 return;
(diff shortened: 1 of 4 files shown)


libs/organization/src/lib/organization/guard/request-access.guard.ts
```diff
@@ -0,0 +1,57 @@
...
+            map(_ => this.query.getActive()),
+            map(org => {
+              if (org.status === 'accepted') {
+                const app = getCurrentApp(this.routerQuery);
+                if (!!org.appAccess[app].marketplace) {
+                  return '/c/o/marketplace/home'
+                }
+                if (!!org.appAccess[app].dashboard) {
+                  return '/c/o/dashboard/home';
+                }
```
phpgeekfr, line 48: maybe less lines with:

return `/c/o/${org.appAccess[app].marketplace? 'marketplace': 'dashboard'}/home`

libs/organization/src/lib/organization/guard/request-access.guard.ts
```diff
@@ -0,0 +1,57 @@
...
+        if (!user.orgId) {
+          return of('/auth/identity');
+        } else {
+          return this.service.syncActive({ id: user.orgId }).pipe(
+            map(_ => this.query.getActive()),
+            map(org => {
+              if (org.status === 'accepted') {
+                const app = getCurrentApp(this.routerQuery);
+                if (!!org.appAccess[app].marketplace) {
+                  return '/c/o/marketplace/home'
```
phpgeekfr, line 44: missing `;` and you can use `else if`

phpgeekfr (approved):

PR #5043, opened 2021-03-01
Clean todo 4932

Mussuranae:
- Allow admin to choose from which platform he wants to create user when creating an org

Next : use UniqueOrgName in indentity form

### DIFF
--- a/libs/admin/src/lib/admin-panel/components/organization/create-organization/create.component.html
+++ b/libs/admin/src/lib/admin-panel/components/organization/create-organization/create.component.html
@@ -1,20 +1,25 @@
-<section fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="50px">
+<section fxLayout="column" fxLayoutAlign="center center">
   <h1>Create Organization </h1>
   <!-- SUPER ADMIN -->
   <form [formGroup]="superAdminForm">
-    <mat-card>
-      <mat-form-field appearance="outline">
-        <mat-label>Super Admin Email</mat-label>
-        <input matInput type="email" placeholder="Email" aria-label="email" formControlName="email" required />
-        <mat-error *ngIf="superAdminForm.get('email').hasError('email') && !superAdminForm.get('email').hasError('required')">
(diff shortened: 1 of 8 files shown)


libs/invitation/src/lib/+state/invitation.service.ts
```diff
@@ -111,6 +111,8 @@ export class InvitationService extends CollectionService<InvitationState> {
 
         const f = this.functions.httpsCallable('inviteUsers');
         const app = getCurrentApp(this.routerQuery); // @TODO #4932 problem here if invite comes from CRM
+        //? Is it really a problem ? When the user will log for the first time, he will update his profile,
+        //? and we will have a new createdFrom or updatedFrom
```
phpgeekfr, line 115: The problem here is that we use `createdFrom`  to send the first mail (links in mail, the sender address etc).
We need to test but  if we invite someone from here, I think that festival will be used as a fallback.

Can you remove comment or add an issue number reminding us that we need to test this case ?

libs/admin/src/lib/admin-panel/components/organization/create-organization/create.component.html
```diff
@@ -1,21 +1,26 @@
...
+      <mat-error *ngIf="superAdminForm.get('email').hasError('taken')">
+        User of this email address already has an organization.
+      </mat-error>
+    </mat-form-field>
   </form>
 
+  <mat-form-field appearance="outline">
+    <mat-label>From App</mat-label>
+    <mat-select [formControl]="fromApp" placeholder="From app" required>
+      <mat-option *ngFor="let app of appList | keyvalue" [value]="app.key">{{ app.value }}</mat-option>
```
phpgeekfr, line 20: can you remove `crm` and `cms` from this select?

phpgeekfr (approved):

libs/invitation/src/lib/+state/invitation.service.ts
Mussuranae: Done with issue #5049

PR #5053, opened 2021-03-02
fix algolia filters on guest list for events

### LINKED ISSUE #4899: Algolia issue when inviting guest to an event that already have guests
If I try to invite a new guest to an event that already have guests I get the following error from Algolia: 
```
message: "Unknown parameter: 1"
```

data sent:
```
data: "{\"0\":\"email:-dev+jessyca-eee@blockframes.io\",\"1\":\"email:-dev+jessyca-eee@blockframes.io\",\"query\":\"e\"}"
```

phpgeekfr:
fix #4899

### DIFF
--- a/libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
+++ b/libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
@@ -103,9 +103,7 @@ export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
       this.displayWithPath = 'value';
     }
 
-    let indexSearch: SearchIndex;
-
-    indexSearch = this.algoliaService.getIndex(this.index)
+    const indexSearch = this.algoliaService.getIndex(this.index);
 
     // create search functions
     const regularSearch = (text: string) => indexSearch.search(text, { facetFilters: this.getFilter() }).then(result => result.hits);
--- a/libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
+++ b/libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
@@ -108,7 +108,7 @@ export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
     indexSearch = this.algoliaService.getIndex(this.index)
 
     // create search functions
-    const regularSearch = (text: string) => indexSearch.search(text, this.getFilter()).then(result => result.hits);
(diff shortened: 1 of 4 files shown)


libs/ui/src/lib/algolia/chips-autocomplete/algolia-chips-autocomplete.component.ts
```diff
@@ -103,12 +103,10 @@ export class AlgoliaChipsAutocompleteComponent implements OnInit, OnDestroy {
...
     }
 
-    let indexSearch: SearchIndex;
-
-    indexSearch = this.algoliaService.getIndex(this.index)
+    const indexSearch = this.algoliaService.getIndex(this.index);
 
     // create search functions
-    const regularSearch = (text: string) => indexSearch.search(text, this.getFilter()).then(result => result.hits);
+    const regularSearch = (text: string) => indexSearch.search(text, { facetFilters: this.getFilter() }).then(result => result.hits);
```
phpgeekfr, line 109: This is what we used to do before migration to algolia v4 . So I guess we need to use facetFilters and not filters (that only accept strings)

libs/utils/src/lib/algolia/algolia.service.ts
```diff
@@ -17,25 +17,24 @@ export class AlgoliaService {
...
         }
-        return this.indices[name];
+        return this.indices[name] as SearchIndex;
 
     }
 
-    query<K extends keyof AlgoliaQueries>(name: K, config: AlgoliaQueries[K]): Promise<AlgoliaRecord<AlgoliaObject[K]>> {
-        return this.getIndex(name).search({
-            query: config.text ?? '',
+    query<K extends keyof AlgoliaQueries>(name: K, config: AlgoliaQueries[K]): Promise<SearchResponse<AlgoliaObject[K]>> {
```
phpgeekfr, line 32: after adding some types to `getIndex(name: 'movie' | 'org' | 'user'): SearchIndex` I realized that this method was  also in error

libs/utils/src/lib/algolia/algolia.service.ts
```diff
@@ -17,25 +17,24 @@ export class AlgoliaService {
...
+        return this.getIndex(name).search<AlgoliaObject[K]>(config.text ?? '', {
             hitsPerPage: config.limitResultsTo,
             page: config.activePage,
             facetFilters: parseFacets(config.facets),
             filters: parseFilters(config.filters)
-        })
+        }).then(e => e);
```
phpgeekfr, line 38: to remove the `readOnly` attribute

fritzschoff (approved):

PR #5068, opened 2021-03-03
Avails Form Component

fritzschoff:


libs/movie/src/lib/movie/form/filters/avails/avails-filter.component.scss
```diff
@@ -0,0 +1,3 @@
+mat-divider {
```
GrandSchtroumpf, line 1: worth case scenario you can use a `hr` to have the same effect

GrandSchtroumpf (approved):

libs/movie/src/lib/movie/form/filters/avails/avails-filter.component.html
```diff
@@ -0,0 +1,30 @@
+<mat-divider></mat-divider>
+<!-- TODO #4997 implement territories search select component -->
+
+<h3>Terms*</h3>
+<div fxLayout fxLayoutAlign="center" fxLayoutGap="24px">
+  <mat-form-field appearance="outline">
+    <mat-label>Licence Start Date</mat-label>
+    <input matInput [matDatepicker]="datepickerFrom" [formControl]="form.get('duration').get('from')">
```
GrandSchtroumpf, line 8: input should be required

PR #5077, opened 2021-03-03
fix movie notes & backend delete file list

pldespaigne:


RemcoSimonides (approved):
LGTM

apps/backend-functions/src/media.ts
```diff
@@ -234,6 +254,14 @@ function needsToBeCleaned(before: StorageFile | undefined, after: StorageFile |
   return !!before.storagePath && before.storagePath !== after.storagePath && !after.storagePath;
 };
 
+function checkFileList(before: StorageFile[] | undefined, after: StorageFile[] | undefined) {
+  const filesToClean: StorageFile[] = [];
+  before?.forEach(beforeFile => {
+    const existsInAfter = after?.some(afterFile => afterFile.storagePath === beforeFile.storagePath);
```
fritzschoff, line 260: not sure but you can use every and avoid the ! in the if statement, just cleaner to read.

fritzschoff (approved):

PR #5088, opened 2021-03-04
Issue contract update

fritzschoff:
got rid of distribution rights, old contract code and contract version code

Sometimes we want to keep the component, cause the structure is correct but the content ain't


libs/admin/src/lib/admin-panel/admin-panel.module.ts
```diff
@@ -22,7 +22,6 @@ import { OrganizationAdminModule } from './pages/organization/organization.modul
 import { AlgoliaAutocompleteModule } from '@blockframes/ui/algolia/autocomplete/algolia-autocomplete.module';
 import { TermDateModule } from '@blockframes/utils/pipes/term-date.pipe';
 import { AppPipeModule, ToLabelModule } from '@blockframes/utils/pipes';
-import { EditTitleModule } from './components/edit-title/edit-title.module';
```
fritzschoff, line 25: @phpgeekfr

phpgeekfr: This refer to a dialog used in libs/admin/src/lib/admin-panel/pages/contract/contract.component.ts that you already removed. So there should be no error

GrandSchtroumpf (approved):

PR #5089, opened 2021-03-04
hide running time

Mussuranae:
When the status of running time is 'to be determined', we hide the running time 'time'
image
OR
image

### DIFF
--- a/libs/movie/src/lib/movie/form/main/main.component.html
+++ b/libs/movie/src/lib/movie/form/main/main.component.html
@@ -62,7 +62,7 @@
       </ng-container>
       <article flxLayout="row" fxLayoutAlign="start start" fxLayoutGap="24px">
 
-        <ng-container *ngIf="form.get('runningTime').get('status').value != 'tobedetermined'">
+        <ng-container *ngIf="form.get('runningTime').get('status').value !== 'tobedetermined'">
           <mat-form-field appearance="outline" fxFlex>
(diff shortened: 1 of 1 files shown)


libs/movie/src/lib/movie/form/main/main.component.html
```diff
@@ -62,13 +62,15 @@ <h3>Reference</h3>
...
       <article flxLayout="row" fxLayoutAlign="start start" fxLayoutGap="24px">
 
-        <mat-form-field appearance="outline" fxFlex>
-          <mat-label>{{ (form.contentType.value | typeDictionary).runningTime }}</mat-label>
-          <input test-id="run-time" matInput type="number" name="runningTime"
-            [formControl]="form.runningTime.get('time')" min="0" required />
-          <mat-hint>In minutes.</mat-hint>
-          <mat-error>{{ (form.contentType.value | typeDictionary).runningTimeError }}</mat-error>
-        </mat-form-field>
+        <ng-container *ngIf="form.get('runningTime').get('status').value != 'tobedetermined'">
```
GrandSchtroumpf, line 65: I think we have a bunch of pipe to do this kind of stuff. Also use "!==" instead of "!=". It' doesn't change anything here, but it's a good practice to do it all the time.

Mussuranae: I don't find any pipe for this one :/

GrandSchtroumpf (approved):

PR #5092, opened 2021-03-04
prevent register festival without name

### LINKED ISSUE #5086: Movie form - Other Festivals - Festival name mandatory when section dirty
user: dev+kenneth-ixf@blockframes.io
url: https://staging.archipelmarket.com/c/o/dashboard/tunnel/movie/02L9gsoY4WTdGbxpOKha/reviews#custom-festival

Other festival section must not be mandatory BUT if a user fills in one of the 3 fields inside that section (name, awards or year), then other festival name needs to become mandatory.

Otherwise we have festivals with no name 😬

Mussuranae:
Add the required state for the name of festival.
Also bold the title in the sidenav for marketplace
Fix #5086


libs/movie/src/lib/movie/form/reviews/reviews.component.html
```diff
@@ -9,7 +9,7 @@ <h3>Festival & Awards</h3>
 
             <div fxLayout="row" fxLayoutGap="24px">
               <static-select test-id="festival-name" [control]="prizesForm.get('name')" scope="festival"
-                [withoutValues]="['']" fxFlex>
+                [withoutValues]="['']" fxFlex required>
```
GrandSchtroumpf, line 12: can you check why there is a "withoutValues" with empty sting ?

GrandSchtroumpf (approved):

PR #5098, opened 2021-03-04
Fix form out of date

RemcoSimonides:


libs/media/src/lib/file/explorer/explorer.component.ts
```diff
@@ -130,6 +132,20 @@ export class FileExplorerComponent implements OnInit, AfterViewInit {
...
+      emptyStorageFile[metadata.field] = createStorageFile({
+        collection: null,
+        docId: null,
+        field: null,
+        privacy: null,
+        storagePath: null
+      })
+      this.db.doc(`${metadata.collection}/${metadata.docId}`).update(emptyStorageFile)
+    }
+  }
```
RemcoSimonides, line 148: Automatically delete file from db when it's removed from the component.

The form in file-explorer does not represent the data of the `file-uploader` component and therefore can't update the db with the "Save" button

libs/media/src/lib/file/file-uploader/file-uploader.component.ts
```diff
@@ -69,7 +73,8 @@ export class FileUploaderComponent implements OnDestroy {
   }
 
 
-  @Input() index: number;
+  @Input() queueIndex: number;
+  @Input() formIndex: number;
```
RemcoSimonides, line 77: @pldespaigne chose to add another Input() instead of making index more complex

libs/movie/src/lib/movie/form/media-notes/notes.component.html
```diff
@@ -23,7 +23,8 @@ <h3>Note / Statement</h3>
               test-id="intent-note"
               [form]="noteForm"
               [meta]="['movies', 'notes', movieId]"
-              [index]="index | fileListIndex: form.promotional.get('notes').value"
+              [queueIndex]="index | fileListIndex: form.promotional.get('notes').value"
```
pldespaigne, line 26: please put the `queueIndex` input before the `form` input since the `form` needs the `queueIndex`

pldespaigne (approved):

PR #5110, opened 2021-03-05
setup static group selection

GrandSchtroumpf:


libs/ui/src/lib/static-autocomplete/group/group.component.html
```diff
@@ -0,0 +1,44 @@
+<ng-container *ngIf="groups$ | async as groups">
+	<ng-container *ngIf="value$ | async as value">
+		<mat-form-field appearance="outline">
+			<mat-label>{{ label }}</mat-label>
```
fritzschoff, line 4: why not content projection? we do it in every other comp

libs/ui/src/lib/static-autocomplete/group/group.component.html
```diff
@@ -0,0 +1,44 @@
...
+							{{ item | toLabel: scope }}
+						</mat-option>
+					</ng-container>
+				</ng-container>
+				<mat-select-trigger>
+					{{ value | triggerDisplay: groups: displayAll }}
+				</mat-select-trigger>
+			</mat-select>
+		</mat-form-field>
+		<mat-hint>{{ hint }}</mat-hint>
```
fritzschoff, line 42: same here

libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -0,0 +1,185 @@
...
+  form = new FormControl([]);
+  value$ = this.form.valueChanges.pipe(
+    startWith(this.form.value),
+    shareReplay(1)
+  );
+  hidden: Record<string, boolean> = {}
+  
+  @Input() displayAll = '';
+  @Input() label = '';
+  @Input() hint = '';
```
fritzschoff, line 83: would go for content projection

libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -0,0 +1,185 @@
...
+  @Input() disabled: boolean = false;
```
fritzschoff, line 84: use the @boolean decorator

fritzschoff (changes requested):

fritzschoff (approved):

Issue #5104, opened 2021-03-05
broken css for import excel


phpgeekfr:
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/e7b1ad02-d92d-4102-84df-1b2b2860e746)

PR #5111, opened 2021-03-05
Fix file import ui

RemcoSimonides:
Fix issue #5104


libs/utils/src/lib/utils.ts
```diff
@@ -35,31 +35,31 @@ export interface FileDefinition { mime: string[], extension: string[] };
...
     mime: ['application/json'],
-    extension: ['json']
+    extension: ['.json']
   },
   csv: {
     mime: ['text/csv'],
-    extension: ['csv']
+    extension: ['.csv']
   }
 }
```
RemcoSimonides, line 64: Fixes the `accept` on `<input [accept]=".csv"`. It wasn't working ^^ probably never had

pldespaigne (approved):

PR #5126, opened 2021-03-09
StaticGroup territories

RemcoSimonides:


libs/utils/src/lib/static-model/static-model.ts
```diff
@@ -784,7 +784,7 @@ export const territoriesISOA2 = {
   aruba: 'AW',
   australia: 'AU',
   austria: 'AT',
-  azerbaidjan: 'AZ',
+  azerbaijan: 'AZ',
```
GrandSchtroumpf, line 787: You need a migration script for that

libs/firebase-utils/src/lib/firestoreMigrations/0055.ts
```diff
@@ -0,0 +1,114 @@
...
+export async function upgrade(db: Firestore) {
+
+  // MOVIE
+  const movies = await db.collection('movies').get()
+  await runChunks(movies.docs, async (movieDoc) => {
+    const data = movieDoc.data() as Movie;
+    let update = false;
+
+    // boxOffice[].territory
+    if (!!data.boxOffice) {
```
GrandSchtroumpf, line 25: Maybe something like that: 
```typescript
// You can use deepkey if it's a deep key
function replace(doc, keys: string[]) {
  let replaced = false;
  for (const key of keys) {
    if (doc[key] && doc[key] === 'christmas- island') {
      doc[key] = 'christmas-island';
      replaced = true;
    }
  }
  return replaced;
}
await runChunks(movies.docs, async (movieDoc) => {
  const data = movieDoc.data() as Movie;
  const isUpdated = replace(data, ['boxOffice', 'rating', ...]);
  if (isUpdated) await movieDoc.ref.set(data)
}
```

GrandSchtroumpf (approved):

PR #5129, opened 2021-03-09
Issue contract import update

fritzschoff:
clean up


libs/utils/src/lib/common-interfaces/timestamp.ts
```diff
@@ -0,0 +1,3 @@
+import type firebase from 'firebase';
+
+export type Timestamp = firebase.firestore.Timestamp;
```
GrandSchtroumpf, line 3: Hmmmm are you using this ? ^^

fritzschoff: yes,

PR #5133, opened 2021-03-09
Add bucket in toolbar

### LINKED ISSUE #5001: Bucket button in the toolbar of the app
- [x] Create a button in the header linked to the bucket. 
- [x] The badge should display the amount of contracts in the selection.
should be on the left side of the mat divider

GrandSchtroumpf:
fix #5001


RemcoSimonides (approved):
LGTM

libs/contract/src/lib/bucket/+state/bucket.model.ts
```diff
@@ -0,0 +1,22 @@
+import { Media, Territory } from '@blockframes/utils/static-model';
+
+export interface Bucket {
+  currency: 'euro' | 'dollar';
```
RemcoSimonides, line 4: missing field `id`

Issue #5138, opened 2021-03-10
Disable Sign up button once clicked


Jackseed:
url: https://staging.archipelmarket.com/auth/connexion#signup

It would prevent users clicking more than once, thus trying to create multiple users while they are just impatient.

Good to be done right now and/or keep it for the new auth flow

PR #5163, opened 2021-03-10
Disable signup button 5138

SamirTouira:
Sign up button disabled once clicked. Thanks to Mussuranae and Jackseed for the little help

fix: #5138


Mussuranae (approved):
Nice job !

PR #5185, opened 2021-03-11
rebase

### LINKED ISSUE #4998: Add to Selection button
- [x] Add button on the movie card to select the title.
- [x] `select(titleId: string)` should get the content of the avail form to create a right. Add it to the bucket (directly on firestore).

Note: Use a transaction to update the bucket to avoid race condition : 
```typescript
add(titleId: string) {
  // Get the parent term
  const parentTermId = this.terms.getValue()[titleId];
  if (!parentTermsId) throw new Error('no available term for this title');
  const term = this.form.value;
  
  this.bucket.update(orgId, (bucket) => {
    const contracts = bucket.contracts || [];
    // Check if there is already a contract that apply on the same parentTermId
    const index = contracts.findIndex(contract => contract.parentTermId === parentTermId);
    
    if (index !== -1) { // If yes, append its's terms with the new one.
      contracts[index].terms.push(term);
      return { contracts };
    } else {  // Else create a new contract
      const contract = { titleId, parentTermsId, terms: [term] };
      return { contracts: [ ...contracts, contract ] };
    }
  });
}
```

fritzschoff:
close #4998

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
+++ b/apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
@@ -135,18 +135,18 @@ export class ListComponent implements OnInit, OnDestroy {
       return;
     }
     // Get the parent term
-    const mandateContract = await this.contractService.getValue(this.mandateMemo[titleId].contractId)
-    if (!mandateContract) throw new Error('no available term for this title');
+    const parentTermId = this.mandateMemo[titleId].id
+    if (!parentTermId) throw new Error('no available term for this title');
     const term = this.availsForm.value;
     this.bucketService.update(this.orgQuery.getActiveId(), (bucket) => {
       const contracts = bucket.contracts || [];
       // Check if there is already a contract that apply on the same parentTermId
-      const index = contracts.findIndex(contract => contract.parentTermId === mandateContract.parentTermId);
+      const index = contracts.findIndex(contract => contract.parentTermId === );
       if (index !== -1) { // If yes, append its's terms with the new one.
         contracts[index].terms.push(term);
         return { ...bucket, contracts };
       } else {  // Else create a new contract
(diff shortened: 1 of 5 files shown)


apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
```diff
@@ -118,6 +129,29 @@ export class ListComponent implements OnInit, OnDestroy {
     return terms.flat() as Term<Date>[];
   }
 
+  async addAvail(titleId: string) {
+    if (this.availsForm.invalid) {
+      this.snackbar.open('Specify the avails before adding to selection', 'close', { duration: 3000 })
+      return;
+    }
+    // Get the parent term
+    const mandate = await this.contractService.getValue(this.mandateMemo[titleId].contractId);
```
GrandSchtroumpf, line 138: I don't think you need the mandate : 
```typescript
const parentTermId = this.mandateMemo[titleId];
if (!parentTermId)  throw
...
const index = contracts.findIndex(contract => contract.parentTermId === parentTermId);
```

apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
```diff
@@ -118,6 +129,29 @@ export class ListComponent implements OnInit, OnDestroy {
     return terms.flat() as Term<Date>[];
   }
 
+  async addAvail(titleId: string) {
+    if (this.availsForm.invalid) {
+      this.snackbar.open('Specify the avails before adding to selection', 'close', { duration: 3000 })
+      return;
+    }
+    // Get the parent term
+    const parentTermId = this.mandateMemo[titleId].id
```
GrandSchtroumpf, line 138: Put a "?" Before .id to avoid "cannot read I'd of undefined"

GrandSchtroumpf (approved):

PR #5208, opened 2021-03-12
migration 56

### LINKED ISSUE #5202: Meeting files migration
Meeting files have been forgotten from the migration 54, we should definitely migrate them!

pldespaigne:
fix #5202

### DIFF
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0056.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0056.ts
@@ -19,14 +19,14 @@ export async function upgrade(db: Firestore) {
 
     if (!!meeting.meta.files) {
 
-      const filesPromises = meeting?.meta?.files?.map(async ref => {
+      const filesPromises = meeting.meta.files.map(async ref => {
 
         const segments = ref.split('/').filter(part => !!part);
         const privacy = segments.shift();
         const collection = segments.shift();
         const docId = segments.shift();
-        const field = segments.shift();
-        const fileName = segments.shift();
+        const fileName = segments.pop();
+        const field = segments.join('.');
 
         if (!privacy || !collection || !docId || !field || !fileName) return;
 
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0056.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0056.ts
@@ -1,0 +1,55 @@
+import { Event, MeetingEvent } from '@blockframes/event/+state/event.model';
+import { Firestore } from '@blockframes/firebase-utils';
+import { StorageFile } from '@blockframes/media/+state/media.firestore';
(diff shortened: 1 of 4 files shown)


libs/firebase-utils/src/lib/firestoreMigrations/old-types.ts
```diff
@@ -177,3 +178,12 @@ export function createOldHostedVideo(params: Partial<OldHostedVideo>): OldHosted
     ...params,
   }
 }
+
+export interface OldMeeting {
```
phpgeekfr, line 182: nice move to create old interfaces

phpgeekfr (approved):
LGTM

libs/firebase-utils/src/lib/firestoreMigrations/0056.ts
```diff
@@ -0,0 +1,55 @@
...
+  /*
+    Media Refactoring for meetings files:
+    `string` -> `StorageFile`
+  */
+  return runChunks(meetings.docs, async (meetingDoc) => {
+    const meeting = meetingDoc.data() as Event<OldMeeting>;
+
+    if (!!meeting.meta.files) {
+
+      const filesPromises = meeting?.meta?.files?.map(async ref => {
```
RemcoSimonides, line 22: Remove question marks since you already check if it exists in the `if` statement above

libs/firebase-utils/src/lib/firestoreMigrations/0056.ts
```diff
@@ -0,0 +1,55 @@
...
+
+        const segments = ref.split('/').filter(part => !!part);
+        const privacy = segments.shift();
+        const collection = segments.shift();
+        const docId = segments.shift();
+        const field = segments.shift();
+        const fileName = segments.shift();
```
RemcoSimonides, line 29: Can you perhaps retrieve the old function to get these values? I remember there may be a case where field could consist of string which includes a `/``. Therefore I think you should first remove the fileName, and then join('/') the leftovers to get the field

libs/firebase-utils/src/lib/firestoreMigrations/0056.ts
```diff
@@ -0,0 +1,55 @@
...
+
+          const storagePath = `${collection}/${docId}/${field}/${fileName}`;
+          return storageFile.find(storage => storage.storagePath === storagePath);
+        } else {
+          return storageFile;
+        }
+      });
+      const files = await Promise.all(filesPromises);
+      // convert old event into new event
+      (meeting as unknown as MeetingEvent).meta.files = files.filter(file => !!file);
```
RemcoSimonides, line 50: I would just have cheated by casting it to `any` but this is a nice trick too

RemcoSimonides (changes requested):

RemcoSimonides (approved):

Issue #5216, opened 2021-03-15
UI fix - Screening list - invitation status spacing


Jackseed:
url https://staging.archipelmarket.com/c/o/marketplace/event
user: dev+carson-yck@blockframes.io

When an invitation is requested to an event, the invitation status is a bit lower than the "Request access" button. Both should be at the same level, good one being the Request access button
![Kapture 2021-03-15 at 11.48.45.gif](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/419ce1a7-300b-4967-a8c5-aa272dbabd7e)
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/79a98641-f24f-4dc4-96e7-012aefbd56f3)

PR #5240, opened 2021-03-16
fix spacing excess between inv button and status

SamirTouira:
related issue: #5216


libs/invitation/src/lib/components/action/action.component.scss
```diff
@@ -8,8 +8,11 @@ section button {
   margin-bottom: 12px;
 }
 
-.status span {
-  padding: 0 40px;
+.status {
+  padding: 14px 32px;
+  margin-bottom: 12px;
+  width: 286.42px;
```
Mussuranae, line 14: No way you put a width of xx.xx pixel, do you want to die ? xD
If the status is for the button, don't put a width. The button will take the place depending on what text it has inside. Instead of a margin-bottom and a padding, can you mix both and just put padding ?

Mussuranae (changes requested):

Mussuranae (approved):
Much better ! Good job !

PR #5252, opened 2021-03-17
fix 5213

### LINKED ISSUE #5213: Movie form - Running time TBD blocks saving
url: https://staging.archipelmarket.com/c/o/dashboard/tunnel/movie/uquO6xBo5Hb0r4oWiHMM/summary
user: dev+brandyn-xio@blockframes.io

Movie form has been changed to hide "Running time field" when status "To be determined" is selected on main page.

The problem is that "Running time" stays mandatory so the form is blocked if no value has been filled in

Best would be that "Running time" is not considered mandatory when TBD is selected. Idk if that's possible.
Also on the summary + dashboard view, "Running time" should be ignored if TBD is selected so in the screenshot above, it should be:
"Running time: To be determined"

Instead of "Running Time Mandatory info missing (To be determined)"

pldespaigne:
fix #5213

### DIFF
--- a/libs/movie/src/lib/movie/form/main/main.component.ts
+++ b/libs/movie/src/lib/movie/form/main/main.component.ts
@@ -47,17 +47,13 @@ export class MovieFormMainComponent implements OnInit, OnDestroy {
     this.dynTitle.setPageTitle('Main Information')
     this.valuesCustomGenres$ = this.form.customGenres.valueChanges.pipe(startWith(this.form.customGenres.value));
 
-    this.sub = this.form.runningTime.valueChanges.pipe(distinctUntilChanged()).subscribe(runningTime => {
-      const status = runningTime.status;
-      const time = runningTime.time;
-      if (status === "confirmed" && !time) {
-        return this.form.runningTime.get('time').setErrors({ required: true });
-      }
-    });
+    this.validateRunningTime(this.form.runningTime.value);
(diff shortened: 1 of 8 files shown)


libs/movie/src/lib/movie/form/main/main.component.ts
```diff
@@ -78,4 +74,13 @@ export class MovieFormMainComponent implements OnInit, OnDestroy {
   get original() {
     return this.form.title.get('original');
   }
+
+  private validateRunningTime({ status, time }: { status: string, time: number }) {
+    if ((status === 'confirmed' || status === 'estimated') && (!time || time <= 0)) {
```
phpgeekfr, line 79: you can also do :
```
[ 'confirmed','estimated'].includes(status)
```

libs/movie/src/lib/movie/form/summary/main/main.component.html
```diff
@@ -100,7 +100,12 @@ <h6>
     </li>
     <li test-id="info-runtime">
       <strong>Running Time </strong>
-      <missing-control [control]="movie.runningTime.get('time')" [link]="link" fragment="general-information">
+      <missing-control
+        *ngIf="movie.runningTime.get('status').value != 'tobedetermined'"
```
phpgeekfr, line 104: use !== instead of !=

phpgeekfr (approved):
LGTM

Issue #5019, opened 2021-02-25
Send mail on offer creation


GrandSchtroumpf:
- [ ] On offer created send a mail to Christophe (email template will be provided by @fritzschoff )
You'll need to query the contracts & the terms based on the offer.

fritzschoff:
make sure to NOT to send an array as root, sendgrid can't handle that. You need an intermediate key. 
```
{
offers:[BUCKERT]
}
```

PR #5259, opened 2021-03-17
5019

RemcoSimonides:
Fix issue #5019

Email result:
image

@GrandSchtroumpf should I change `payTv` to `Pay TV`, and `EUR` to `euro` and such?

### DIFF
--- a/libs/contract/src/lib/bucket/+state/bucket.service.ts
+++ b/libs/contract/src/lib/bucket/+state/bucket.service.ts
@@ -7,8 +7,12 @@ import { Contract, ContractService } from '../../contract/+state';
 import { OfferService } from '../../offer/+state';
 import { IncomeService } from '../../income/+state';
 import { OrganizationQuery } from '@blockframes/organization/+state';
-import { centralOrgID } from '@env';
+import { centralOrgID, supportEmails } from '@env';
 import type firebase from 'firebase';
+import { SendgridService } from '@blockframes/utils/emails/sendgrid.service';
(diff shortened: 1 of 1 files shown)


libs/contract/src/lib/bucket/+state/bucket.service.ts
```diff
@@ -97,5 +106,25 @@ export class BucketService extends CollectionService<BucketState> {
...
+      contract['title'] = title.title.international;
+      contract['orgName'] = org.denomination.full;
+      for (const term of contract.terms) {
+        term.duration.from = toDate(term.duration.from).toDateString() as any
+        term.duration.to = toDate(term.duration.to).toDateString() as any
+      }
+    }
+    const app: App = 'catalog';
+    const request = {
+      to: supportEmails[app],
```
RemcoSimonides, line 124: For production this points to support@archipelcontent.com, which is redirected to business@archipelcontent.com

libs/contract/src/lib/bucket/+state/bucket.service.ts
```diff
@@ -97,5 +106,25 @@ export class BucketService extends CollectionService<BucketState> {
       // We empty the selection but leave the currency
       return { contracts: [] };
     })
+
+    const org = this.orgQuery.getActive()
```
GrandSchtroumpf, line 110: Maybe create a new method for that, just for dividing the logic.

GrandSchtroumpf (approved):

PR #5266, opened 2021-03-18
added bucket

fritzschoff:


apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
```diff
@@ -88,14 +88,15 @@ export class ListComponent implements OnInit, OnDestroy {
       debounceTime(300),
       switchMap(async ([_, availsValue]) => [await this.searchForm.search(), availsValue]),
     ).subscribe(([movies, availsValue]: [SearchResponse<Movie>, AvailsFilter]) => {
+      const bucketQuery = this.bucketQuery.getActive()?.contracts?.map(contract => contract.terms).flat() || [];
```
GrandSchtroumpf, line 91: you'll need to use a stream here instead if the bucket changes but the not the search & avails. Typically when you add a contract to your selection : 
```typescript
function allBucketTerms(bucket?: Bucket) {
  if (!bucket?.contracts) return [];
  return bucket.contracts.map(contract => contract.terms).flat();
}
....
this.sub = combineLatest([
  this.searchForm.valueChanges.pipe(startWith(this.searchForm.value)),
  this.availsForm.valueChanges.pipe(startWith(this.availsForm.value)),
  this.bucket.selectActive().pipe(map(allBucketTerms))
])
```
Normally even if there are no bucket in firestore, it should work.
**If not** you can add : `startWith(undefined)` before the map.

GrandSchtroumpf (approved):

PR #5280, opened 2021-03-18
5019

RemcoSimonides:
- [x] Move send email and empty bucket to onOfferCreate
- [x] Add Org and User data to email

Email result:
image

### DIFF
--- a/apps/backend-functions/src/offer.ts
+++ b/apps/backend-functions/src/offer.ts
@@ -5,9 +5,10 @@ import { App } from '@blockframes/utils/apps';
 import { Offer } from '@blockframes/contract/offer/+state/offer.model';
 import { OrganizationDocument } from '@blockframes/organization/+state';
 import { Movie } from '@blockframes/movie/+state';
-import { supportEmails } from '@env';
+import { appUrl, supportEmails } from '@env';
 import { staticModel } from '@blockframes/utils/static-model';
 import { format } from "date-fns";
+import { User } from '@blockframes/user/types';
 
 
 export async function onOfferCreate(snap: FirebaseFirestore.DocumentSnapshot): Promise<void> {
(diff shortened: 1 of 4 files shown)


apps/backend-functions/src/offer.ts
```diff
@@ -0,0 +1,44 @@
...
+  const app: App = 'catalog';
+  const request: EmailTemplateRequest = {
+    to: supportEmails[app],
+    templateId: 'd-94a20b20085842f68fb2d64fe325638a',
+    data: { org, bucket }
+  }
+  sendMailFromTemplate(request, app);
+
+  // Empty bucket
+  db.doc(`buckets/${orgId}`).update({ contracts: [] });
```
GrandSchtroumpf, line 43: You can clean the bucket before getting the movies, like that the UI in the front would be updated faster

apps/backend-functions/src/offer.ts
```diff
@@ -0,0 +1,44 @@
...
+
+
+export async function onOfferCreate(snap: FirebaseFirestore.DocumentSnapshot): Promise<void> {
+  const offer = snap.data() as Offer;
+  const orgId = offer.buyerId;
+  const [ org, bucket ] = await Promise.all([
+    getDocument<OrganizationDocument>(`orgs/${orgId}`),
+    getDocument<any>(`buckets/${orgId}`)
+  ]);
+
```
GrandSchtroumpf, line 20: At that point you should clean the bucket because you got it in memory

apps/backend-functions/src/offer.ts
```diff
@@ -0,0 +1,44 @@
...
+      term.duration.from = format(term.duration.from.toDate(), 'dd/MM/yyyy');
+      term.duration.to = format(term.duration.to.toDate(), 'dd/MM/yyyy');
+    }    
+  }
+
+  const app: App = 'catalog';
+  const request: EmailTemplateRequest = {
+    to: supportEmails[app],
+    templateId: 'd-94a20b20085842f68fb2d64fe325638a',
+    data: { org, bucket }
```
GrandSchtroumpf, line 38: You might also want to get the connected user with the context.uid data.

libs/contract/src/lib/bucket/+state/bucket.service.ts
```diff
@@ -43,19 +41,19 @@ export class BucketService extends CollectionService<BucketState> {
...
+    const bucket = value.entities[value.active];
 
     const get = <T>(tx: firebase.firestore.Transaction, path: string): Promise<T> => {
       const ref = this.db.doc(path).ref;
       return tx.get(ref).then(snap => snap.data() as T);
     }
      // Run tx
-    await this.update(orgId, async (bucket: Bucket, tx) => {
-       offer = bucket;
+    await this.runTransaction(async (tx) => {
```
GrandSchtroumpf, line 52: You should return now instead of awaiting

GrandSchtroumpf (approved):

GrandSchtroumpf (approved):

PR #5292, opened 2021-03-18
ok signup changes

phpgeekfr:


libs/user/src/lib/auth/components/data-validation/data-validation.component.scss
```diff
@@ -15,8 +15,13 @@ section {
   top: 8px;
   width: 12px;
   height: 12px;
+  margin-left: 3px;
```
Mussuranae, line 18: for all margin/padding, etc, can you put multiple of 8 ? 4px here and 8 or 16px below for the button

Mussuranae (approved):

Issue #5241, opened 2021-03-16
Should only query contracts of Archipel Content


RemcoSimonides:
Currently on catalog movie titles page, we query all the contracts. Instead we should only query the Mandates of Archipel Content and only Sales which match the contract of Archipel Content.

See tagged issue number in code.

This issue has relatively low priority since no other organizations have mandates on Archipel Content yet.

PR #5305, opened 2021-03-19
5241

RemcoSimonides:
Fix issue #5241

### DIFF
--- a/libs/contract/src/lib/contract/+state/contract.service.ts
+++ b/libs/contract/src/lib/contract/+state/contract.service.ts
@@ -12,7 +12,6 @@ export class ContractService extends CollectionService<ContractState> {
   }
 
   getContractsOfOrg(orgId = centralOrgID) {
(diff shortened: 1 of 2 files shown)


libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -9,4 +10,12 @@ export class ContractService extends CollectionService<ContractState> {
   constructor(store: ContractStore) {
     super(store);
   }
+
+  getContractsOfOrg(orgId = centralOrgID) {
+    // Only the mandate of Archipel Content && only sales which match the contract of Archipel Content
+    return Promise.all([
+      this.getValue(ref => ref.where('type', '==', 'mandata').where('buyerId', '==', orgId)),
```
GrandSchtroumpf, line 17: typo: "mandata" -> "mandate"

GrandSchtroumpf (approved):
I find it a little bit stupid to query mandates & sales, merge them, and divide them again after. Maybe we could improve that

Issue #5291, opened 2021-03-18
Movie form - Summary page - Hide Preview button when no image


Jackseed:
On Summary page, when there is no file (image, pdf, video) the "eye" button should be hidden as there is no preview

![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/1002a72d-0a40-4065-a787-6e7e69cfb972)

PR #5308, opened 2021-03-19
fix eye preview displayed when no image

SamirTouira:
fix: #5291


Mussuranae (changes requested):
Wrap the modal and the span in a div to put the if on it. There is no need to call the pipe in the span etc if there is no value.

SamirTouira:
> Wrap the modal and the span in a div to put the if on it. There is no need to call the pipe in the span etc if there is no value.

ok it's done

PR #5309, opened 2021-03-19
update e2e

Mussuranae:
image


libs/e2e/src/lib/pages/organization/OrganizationPendingPage.ts
```diff
@@ -0,0 +1,22 @@
...
+      expect(loc.pathname).to.eq(CREATEPATH);
+    });
+  }
+
+  public assertMoveToOrgJoinPage() {
+    cy.location().should(loc => {
+      expect(loc.pathname).to.eq(JOINPATH);
+    });
+  }
+}
```
xmano, line 22: I think these functions can be refactored to a utility function in e2e utils to be used from any spec file:

```
  public assertMoveTo(pathExpected: string) {
    cy.location().should(loc => {
      expect(loc.pathname).to.eq(pathExpected);
    });
  }
```

It would be helpful

xmano (approved):
Approved with a comment to refactor to utility function.

Thanks

Issue #5289, opened 2021-03-18
Even if stills are empty, displays "empty images" on summary page


Jackseed:
Login : dev+howell-mxf@blockframes.io
url https://staging.archipelmarket.com/c/o/dashboard/title/02L9gsoY4WTdGbxpOKha/main

- [ ] Promotional elements / Images / Stills : Even if stills are empty, displays "empty images" on summary page 
![8](https://user-images.githubusercontent.com/79836137/111616738-8a71bb00-87e2-11eb-9047-9dde37843c45.gif)

PR #5318, opened 2021-03-19
Files explorer

pldespaigne:
lots of small fixes

### DIFF
--- a/libs/media/src/lib/video/viewer/viewer.component.ts
+++ b/libs/media/src/lib/video/viewer/viewer.component.ts
@@ -33,6 +33,11 @@ export class VideoViewerComponent implements AfterViewInit, OnDestroy {
   private signalPlayerReady: () => void;
   private waitForPlayerReady: Promise<void>;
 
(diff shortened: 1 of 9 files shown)


pldespaigne:
will fix #5289

libs/media/src/lib/video/viewer/viewer.component.ts
```diff
@@ -134,15 +153,14 @@ export class VideoViewerComponent implements AfterViewInit, OnDestroy {
       }
     } catch(error) {
       this.loading$.next(false);
+      console.warn(error);
```
phpgeekfr, line 156: is it wanted ?

phpgeekfr (approved):
LGTM

libs/media/src/lib/video/viewer/viewer.component.ts
pldespaigne: yup, it's to help dev in case of errors, otherwise the error message is only displayed in the snackbar which disappear after some seconds.

PR #5323, opened 2021-03-19
add spinner

### LINKED ISSUE #5316: Dashboard homepage - Add a spinner while statistics are loading
### DASHBOARD HOMEPAGE
- [ ] I get an empty page for a few seconds before it loads. There should be a centered spinner instead while stats are loading

image

Mussuranae:
I added a spinner when we wait for analytics card to display. But I think it's weird, I have a feeling it is more long with the spinner than without, I don't know how I can improve this behavior.
Close #5316


RemcoSimonides (approved):
Doesn't feel slower to me ;)

Issue #5311, opened 2021-03-19
[Bonus] add modal selection


GrandSchtroumpf:
[zeplin](https://app.zeplin.io/project/5fecb41a2e806651584e265c/screen/6036abd91025b82b9e42068e)

- [x] Add modal selection page
- [x] Update the bucket datamodel with `specificity: string` & `delivery: string`.
- [x] Update offer datamodel with `delivery: string`
- [x] Update contract datamodel with  `specificity: string`
- [x] Update the `createOffer` to add `delivery: string` on the offer, & `specific: string` on every contract created by the bucket
- [x] Update mail with content

PR #5339, opened 2021-03-22
5311 Add specific terms

RemcoSimonides:
Fix issue #5311

### DIFF
--- a/libs/contract/src/lib/bucket/+state/bucket.service.ts
+++ b/libs/contract/src/lib/bucket/+state/bucket.service.ts
@@ -37,7 +37,7 @@ export class BucketService extends CollectionService<BucketState> {
     return bucket;
   }
 
-  async createOffer(data: { specificity: string, delivery: string }) {
+  async createOffer(specificity: string, delivery: string) {
     const orgId = this.orgQuery.getActiveId();
 
     const get = <T>(tx: firebase.firestore.Transaction, path: string): Promise<T> => {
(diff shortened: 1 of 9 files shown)


libs/contract/src/lib/bucket/+state/bucket.service.ts
```diff
@@ -37,9 +37,8 @@ export class BucketService extends CollectionService<BucketState> {
     return bucket;
   }
 
-  async createOffer() {
+  async createOffer(data: { specificity: string, delivery: string }) {
```
GrandSchtroumpf, line 40: Better to put two params instead of an object here: 
```
async createOffer(specificity: string, delivery: string)
```

libs/contract/src/lib/bucket/+state/bucket.service.ts
```diff
@@ -100,7 +101,7 @@ export class BucketService extends CollectionService<BucketState> {
       }
 
       // Adding uid to add user data to email sent to business team
-      return { uid: this.authQuery.userId }
+      return { ...data, uid: this.authQuery.userId }
```
GrandSchtroumpf, line 104: There is no reason to keep this data around if you don't set it back in the form of SpecificTermsComponent.

GrandSchtroumpf (approved):

libs/contract/src/lib/bucket/+state/bucket.service.ts
RemcoSimonides: This return is in the bucket update callback function and, correct me if I'm wrong, the returned value in this callback function is what the bucket is updated with.

PR #5348, opened 2021-03-23
e2e ci err fix

xmano:
Test results were not generated properly due to ffmpeg video err. The Cypress 6.8.0 rel has fix for this.

image

For festival create screening test fail - added some wait and timeout to make the element available in more stable way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,7 +155,7 @@
     "commander": "^7.0.0",
     "concurrently": "^5.0.0",
     "cross-env": "^7.0.2",
-    "cypress": "^5.2.0",
+    "cypress": "^6.8.0",
     "cypress-wait-until": "^1.7.1",
     "dotenv": "8.2.0",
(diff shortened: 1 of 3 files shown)


phpgeekfr (approved):
lgtm

package.json
```diff
@@ -155,7 +155,7 @@
     "commander": "^7.0.0",
     "concurrently": "^5.0.0",
     "cross-env": "^7.0.2",
-    "cypress": "^5.2.0",
+    "cypress": "^6.8.0",
```
phpgeekfr, line 158: no side effects with this new version ? is electrum working like chrome or still having issues?

xmano: It fixed the video issue for sure.

But we are now having chrome browser as default. So for now Electron is not priority support. Instead several new Cypress features that are now available will be useful for new developments.

Issue #5276, opened 2021-03-18
UI review


Jackseed:
Review from @stephanie-cascade8 
- placeholder to come

- [x] (**Clélia**) Dashboard home - empty stats. Ask for placeholders for empty content because it looks like a bug. 
<img width="1153" alt="Capture d’écran 2021-03-17 à 14 43 39" src="https://user-images.githubusercontent.com/66258851/111477353-51c4d980-872f-11eb-92b4-35cfe8d6ad12.png">

- [x] (**Clélia**) Center loading spinners in the layouts
![Capture d’écran 2021-03-17 à 14 53 44](https://user-images.githubusercontent.com/66258851/111478691-98ff9a00-8730-11eb-8df8-1b150e147c26.png)

- [x] (**Clélia**) Margin-top: 72px before "load more titles"
<img width="1252" alt="Capture d’écran 2021-03-17 à 14 56 28" src="https://user-images.githubusercontent.com/66258851/111479267-188d6900-8731-11eb-933f-e4e54bce4009.png">

- [x] (**Clélia**) 32 px margin before the Sales Agent Card
<img width="495" alt="Capture d’écran 2021-03-17 à 15 01 16" src="https://user-images.githubusercontent.com/66258851/111480057-d6b0f280-8731-11eb-96df-1e9afb270681.png">

- [x] (**Clélia**) No divider after the last element of a list ( if you can target the last with css nth-child for example)
<img width="1261" alt="Capture d’écran 2021-03-17 à 15 03 37" src="https://user-images.githubusercontent.com/66258851/111480410-458e4b80-8732-11eb-9824-c0bb0e1f8d28.png">

- [x] (**Clélia**) basket icon should be aligned at the extreme bottom of the card.
I know it's inside a button so I tested with flexboxes on the button and on the span inside (it works)
<img width="696" alt="Capture d’écran 2021-03-17 à 15 46 48" src="https://user-images.githubusercontent.com/66258851/111486945-37432e00-8738-11eb-83b2-3cc3c3c908af.png">

- [x] Invitations 'accept' refuse' : one of the icons is bigger. Can you choose the smallest one?   (mano)
<img width="317" alt="Capture d’écran 2021-03-17 à 15 48 56" src="https://user-images.githubusercontent.com/66258851/111487207-72ddf800-8738-11eb-96db-d886352397f8.png">

- [x] (**PL**) My Files section - when you go on a nested folder the back arrow is a bit lower than before, it should be at the same level
[...]

PR #5352, opened 2021-03-23
Last UI fixes

Mussuranae:
Last UI fixes from #5276 
Add a placeholder when statistic card are empty
image


libs/movie/src/lib/movie/components/movie-analytics-chart/movie-analytics-chart.component.ts
```diff
@@ -131,4 +131,12 @@ export class MovieAnalyticsChartComponent {
       return 0;
     }
   }
+
+  displayPlaceholder(eventName: MovieAnalyticsEventName) {
+    const totalHits = this.totalHitsOnCurrentMonth(eventName);
+    if (!!totalHits || totalHits === 0) {
```
GrandSchtroumpf, line 137: You can directly return the value. I think you want :
```typescript
return typeof this.totalHitsOnCurrentMonth(eventName) === 'number';
```

GrandSchtroumpf (approved):

PR #5358, opened 2021-03-23
Multiple low prio

pldespaigne:


apps/backend-functions/src/media.ts
```diff
@@ -56,89 +56,97 @@ export async function linkFile(data: storage.ObjectMetadata) {
...
+        // Is the user allowed to upload this file ?
+        switch (metadata.collection) {
+          case 'users': {
+            // only the user is allowed to upload files about himself
+            await assertFile(metadata.docId === metadata.uid, notAllowedError);
+            break;
+
+          } case 'movies': case 'campaigns': { // campaigns have the same ids as movies and business upload rules are the same
+            // only users members of orgs which are part of a movie, are allowed to upload to this movie/campaign
+            const userRef = db.collection('users').doc(metadata.uid);
```
phpgeekfr, line 84: Why is getDocument<Obj> not used anymore?

apps/backend-functions/src/media.ts
```diff
@@ -56,89 +56,97 @@ export async function linkFile(data: storage.ObjectMetadata) {
...
-
-        } default: {
-          await assertFile(false, `UNKNOWN COLLECTION : ${metadata.collection}`);
+    await db.runTransaction(async transaction => {
+      // because of possible nested map and arrays, we need to retrieve the whole document
+      // modify it, then update the whole doc with the new (modified) version
+      const docRef = db.collection(metadata.collection).doc(metadata.docId);
+      const docSnap = await transaction.get(docRef);
+      await assertFile(docSnap.exists, `Document ${metadata.collection}/${metadata.docId} does not exists`);
+      const doc = docSnap.data()!;
```
phpgeekfr, line 65: what is the use of `!` (curiosity) ?

phpgeekfr (approved):
LGTM

apps/backend-functions/src/media.ts
pldespaigne: it tells TypeScript that the result cannot be `undefined` (because I checked `docSnap.exists` the line above)

apps/backend-functions/src/media.ts
pldespaigne: because the code is in a firestore atomic transaction and `getDocument` does not support transaction

phpgeekfr (approved):

PR #5359, opened 2021-03-23
Fix stakeholder array

RemcoSimonides:


libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -138,13 +137,13 @@ export class ViewExtractedContractsComponent implements OnInit {
...
             if (trimmedRow[SpreadSheetContract.stakeholders]) {
-              let orgs: Organization[];
-              /* if (Array.isArray(trimmedRow[SpreadSheetContract.stakeholders])) {
-                 orgs = await Promise.all(trimmedRow[SpreadSheetContract.stakeholders].map(orgName => this.orgService.getValue(ref => ref.where('denomination.public', '==', orgName))));
-               //} else {*/
-              orgs = await this.orgService.getValue(ref => ref.where('denomination.public', '==', trimmedRow[SpreadSheetContract.stakeholders]))
-              //}
+              const orgs: Organization[] = [];
+              const stakeholders = trimmedRow[SpreadSheetContract.stakeholders].split(this.separator);
+              const promises = stakeholders.map(orgName => this.orgService.getValue(ref => ref.where('denomination.public', '==', orgName)));
```
phpgeekfr, line 142: you should  add a trim() to orgName

phpgeekfr: don't we use full instead of public ?

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -138,13 +137,13 @@ export class ViewExtractedContractsComponent implements OnInit {
...
+              const result = await Promise.all(promises);
+              result.forEach(organizations => organizations.forEach(org => orgs.push(org)));
               contract.stakeholders = orgs.filter(org => !!org).map(org => org.id);
+              console.log('stakeholders: ', contract.stakeholders)
```
phpgeekfr: just for dev purposes I guess

phpgeekfr (approved):

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
RemcoSimonides: Probably yea, public is optional so I'll change it to public

fritzschoff (approved):

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -190,7 +186,7 @@ export class ViewExtractedContractsComponent implements OnInit {
             if (trimmedRow[SpreadSheetContract.territories]) {
               const territoryValues = (trimmedRow[SpreadSheetContract.territories]).split(this.separator)
               const territories = territoryValues.map(territory => getKeyIfExists('territories', territory.trim())).filter(territory => !!territory)
-              term.territories = territories;
+              term.territories = (territories.length === 1 && territories[0] === 'world') ? parseToAll('territories', 'world') : territories;
```
GrandSchtroumpf, line 189: Maybe : `territories.every(t => t === 'world') ? parseToAll('territories', 'world') : territories`

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -275,7 +271,8 @@ export class ViewExtractedContractsComponent implements OnInit {
 
             if (trimmedRow[SpreadSheetContract.dubbed]) {
               const languageValues = (trimmedRow[SpreadSheetContract.dubbed]).split(this.separator);
-              const languages: Language[] = languageValues.map(language => getKeyIfExists('languages', language.trim()))
+              let languages: Language[] = languageValues.map(language => getKeyIfExists('languages', language.trim()))
+              if (languages.length === 1 && languages[0] == 'all') languages = parseToAll('languages', 'all');
```
GrandSchtroumpf, line 275: Same as above

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -293,7 +290,8 @@ export class ViewExtractedContractsComponent implements OnInit {
 
             if (trimmedRow[SpreadSheetContract.subtitled]) {
               const languageValues = (trimmedRow[SpreadSheetContract.subtitled]).split(this.separator);
-              const languages: Language[] = languageValues.map(language => getKeyIfExists('languages', language.trim()))
+              let languages: Language[] = languageValues.map(language => getKeyIfExists('languages', language.trim()));
```
GrandSchtroumpf, line 293: Create a reusable function for that :
```typescript
function getLang(rawValue: string) {
  return rawValue === 'world'
  ? parseToAll('languages', 'all')
  : rawValue.split(this.separator);
}
```
Or somehting like that

GrandSchtroumpf (approved):

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -10,10 +10,9 @@ import { getKeyIfExists } from '@blockframes/utils/helpers';
...
 import { ContractsImportState } from '../../../import-utils';
 import { AuthQuery } from '@blockframes/auth/+state';
-import { Organization, OrganizationQuery, OrganizationService } from '@blockframes/organization/+state';
-import { Language, LanguageValue, MediaValue, TerritoryValue } from '@blockframes/utils/static-model';
-import { TermService } from '@blockframes/contract/term/+state/term.service'
-import { Term } from '@blockframes/contract/term/+state/term.model';
+import { OrganizationQuery, OrganizationService } from '@blockframes/organization/+state';
+import { Language, parseToAll } from '@blockframes/utils/static-model';
+import { TermService } from '@blockframes/contract/term/+state/term.service';
 import { AngularFirestore } from '@angular/fire/firestore';
```
phpgeekfr, line 17: concerning 
```
const trimmedRow: string[] = spreadSheetRow.map(cell => {
        if (typeof cell === 'string') return cell.trim()
        return cell.toString()
      })
```
For movie I used something like this (so it mean that xlsx returns number or string only)

```
if (value) {
          return isNaN(value) ? value.trim() : value.toString();
        } else {
          return '';
        }
```

phpgeekfr (approved):

PR #5369, opened 2021-03-24
Fix rules with createOffer

### LINKED ISSUE #5346: Insufficient permission to create offer
Reproduce
- Login with account that is not a blockframes admin and try to create an offer

GrandSchtroumpf:
fix #5346


RemcoSimonides (approved):
Let's see ^^

PR #5377, opened 2021-03-24
Contract import v2

RemcoSimonides:
Contract import refactored.

### DIFF
--- a/libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
+++ b/libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
@@ -18,7 +18,8 @@ import { centralOrgID } from '@env';
 import { Scope } from '@blockframes/utils/static-model/static-model';
 
 const separator = ';'
-const errorsMap: {[key: string]: SpreadsheetImportError} = {
+type errorCodes = 'no-title-id' | 'no-seller-id' | 'no-buyer-id' | 'no-stakeholders' | 'no-territories' | 'no-medias' | 'no-duration-from' | 'no-duration-to';
+const errorsMap: {[key in errorCodes]: SpreadsheetImportError} = {
   'no-title-id': {
     type: 'warning',
     field: 'contract.titleId',
@@ -84,14 +85,15 @@ function split(cell: string) {
 // Time is MM/DD/YYYY
 function getDate(time: string) {
   if (isNaN(+time)) {
-    return new Date(new Date(time).setHours(0));
+    const [month, day, year] = time.split(/[/.]+/).map(t => parseInt(t, 10));
+    return new Date(year, month - 1, day, 0, 0, 0);
   } else {
     return new Date(Math.round(+time - 25569) * 86400 * 1000);
   }
 }
 
 function getStatic(scope: Scope, value: string) {
-  if (value === 'all' || (scope === 'territories' && value === 'world')) return parseToAll(scope, value);
+  if (value.toLowerCase() === 'all') return parseToAll(scope, 'all');
   return split(value).map(v => getKeyIfExists(scope, v)).filter(v => !!v);
 }
 
--- a/libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
+++ b/libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
@@ -8,32 +8,91 @@ import { ContractService } from '@blockframes/contract/contract/+state/contract.
 import { Intercom } from 'ng-intercom';
 import { getKeyIfExists } from '@blockframes/utils/helpers';
 import { DynamicTitleService } from '@blockframes/utils/dynamic-title/dynamic-title.service';
-import { ContractsImportState } from '../../../import-utils';
+import { ContractsImportState, SpreadsheetImportError } from '../../../import-utils';
 import { AuthQuery } from '@blockframes/auth/+state';
-import { OrganizationQuery, OrganizationService } from '@blockframes/organization/+state';
-import { Language, parseToAll } from '@blockframes/utils/static-model';
-import { TermService } from '@blockframes/contract/term/+state/term.service';
+import { OrganizationService } from '@blockframes/organization/+state';
+import { parseToAll } from '@blockframes/utils/static-model';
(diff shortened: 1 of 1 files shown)


libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -8,32 +8,91 @@ import { ContractService } from '@blockframes/contract/contract/+state/contract.
...
+  return cell.split(separator).filter(v => !!v).map(v => v.trim());
+}
+
+// Time is MM/DD/YYYY
+function getDate(time: string) {
+  if (isNaN(+time)) {
+    return new Date(new Date(time).setHours(0));
+  } else {
+    return new Date(Math.round(+time - 25569) * 86400 * 1000);
+  }
```
RemcoSimonides, line 90: `new Date('12.01.2021')` => 1st december 2021
`new Date('12/01/2021')` => 1st december 2021
So no need to split first, then get year, month and days, and then to put it in a date.

Also, sometimes it comes through as a number, then Max's formula applies

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -8,32 +8,91 @@ import { ContractService } from '@blockframes/contract/contract/+state/contract.
...
-  originalLanguageLicensed,
-  dubbed,
-  subtitled,
-  closedCaptioning,
-  contractId,
-  parentTermId,
-  titleId,
-  stakeholders,
+const separator = ';'
+const errorsMap: {[key: string]: SpreadsheetImportError} = {
```
RemcoSimonides, line 21: How do I type the key? So I get intellisense when getting a value from this object.

GrandSchtroumpf: Either: 
```typescript
type ErrorCode = 'no-title-id' | '...';
const errors = { [key in ErrorCode]: SpreadsheetImportError } = { ... };
```
or : 
```typescript
const errors = { .... } as const; // this tells Typescript that it's readonly and will not be mutated
```

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
GrandSchtroumpf: The issue was to set hours & sec to 0. By doing `new Date(year, month - 1, day, 0, 0)` we have a clean way to set the time (instead of a Date inside a Date).
You can use this regex to split either by / or by . : `const  [month, day, year] = time.split(/[/.]+/)`
In which condition is the time is a number ? And is it a time in second or millisecond ?

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -8,32 +8,91 @@ import { ContractService } from '@blockframes/contract/contract/+state/contract.
...
+function getDate(time: string) {
+  if (isNaN(+time)) {
+    return new Date(new Date(time).setHours(0));
+  } else {
+    return new Date(Math.round(+time - 25569) * 86400 * 1000);
+  }
+}
+
+function getStatic(scope: Scope, value: string) {
+  if (value === 'all' || (scope === 'territories' && value === 'world')) return parseToAll(scope, value);
```
GrandSchtroumpf, line 94: Don't try to put everything on one line.
```typescript
if (scope === 'territories' && value === 'world') return parseToAll(scope, value);
if (value === 'all') return parseToAll(scope, value);
return split(value).map(v => getKeyIfExists(scope, v)).filter(v => !!v);
```
But to be honest I would rather not support world as it makes thing more confusing for everybody

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -69,281 +131,134 @@ export class ViewExtractedContractsComponent implements OnInit {
...
-                name: 'Duration from',
-                reason: 'Archipel Content needs to know the starting date of the contract.',
-                hint: 'Edit corresponding sheet field.'
-              })
-            }
+      // CONTRACT
+      let baseContract: Partial<Mandate | Sale> = {};
+      if (_contractId) {
+        baseContract = await this.contractService.getValue(_contractId as string);
+        if (!baseContract) throw new Error('No contract found for id' + _contractId);
```
GrandSchtroumpf, line 211: Should it throw or add a message & fallback to `this.fire.createId()` ?

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -69,281 +131,134 @@ export class ViewExtractedContractsComponent implements OnInit {
...
-        matSnackbarRef.dismissWithAction(); // loading ended */
-      }
+      this.contractsToCreate.data.push({
+        contract,
+        newContract: !_contractId,
+        errors,
+        terms: [term]
+      });
+      // Forcing change detection
+      this.contractsToCreate.data = [...this.contractsToCreate.data];
```
GrandSchtroumpf, line 257: I think it's a hack we could work around, but let's do that later

GrandSchtroumpf (approved):

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
RemcoSimonides: The incoming value is `44197` which should be 1st of January 2021.
I guess it's days since 44197 is too low to be seconds and `24 * 60 * 60 === 86400` which is in the formula

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
RemcoSimonides: I think it should throw since you try to update a contract, not to create a new one.
Honestly, I think it should also throw if there is no titleId, sellerId or buyerId. Currently those are only warnings but the contract is just incorrect if they are not defined.

Issue #5347, opened 2021-03-23
Set appAccess to true when importing contract


RemcoSimonides:
Set `movie.appAccess.catalog` to true when importing a contract for the specified app.

At this line: libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts at line 329
you can get the title id from the row and call the movieservice to update the movie.

RemcoSimonides:
Blocked by broken contract import

PR #5379, opened 2021-03-25
5347 and added back languages to import

RemcoSimonides:
Fix issue #5347

### DIFF
--- a/libs/import/src/lib/components/table-extracted-elements/contracts/contracts.component.ts
+++ b/libs/import/src/lib/components/table-extracted-elements/contracts/contracts.component.ts
@@ -11,6 +11,8 @@ import { sortingDataAccessor } from '@blockframes/utils/table';
 import { ContractsImportState, SpreadsheetImportError } from '../../../import-utils';
 import { TermService } from '@blockframes/contract/term/+state/term.service';
 import { MovieService } from '@blockframes/movie/+state/movie.service';
+import { getCurrentApp } from '@blockframes/utils/apps';
+import { RouterQuery } from '@datorama/akita-ng-router-store';
 
 const hasImportErrors = (importState: ContractsImportState, type: string = 'error'): boolean => {
   return importState.errors.filter((error: SpreadsheetImportError) => error.type === type).length !== 0;
@@ -46,7 +48,8 @@ export class TableExtractedContractsComponent implements OnInit {
     private dialog: MatDialog,
     private contractService: ContractService,
     private termService: TermService,
-    private movieService: MovieService
+    private movieService: MovieService,
+    private routerQuery: RouterQuery
   ) { }
 
   ngOnInit() {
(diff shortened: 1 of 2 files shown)


libs/import/src/lib/components/table-extracted-elements/contracts/contracts.component.ts
```diff
@@ -109,6 +111,13 @@ export class TableExtractedContractsComponent implements OnInit {
     const termIds = await this.termService.add(importState.terms);
     importState.contract.termIds = termIds;
     await this.contractService.add(importState.contract);
+    const titleId = importState.contract.titleId;
+    if (!!titleId) {
+      const title = await this.movieService.getValue(titleId);
+      title.storeConfig.appAccess.catalog = true;
+      this.movieService.update(title);
+    }
```
RemcoSimonides, line 119: There is probably a nicer way to do this.
I considered
`this.movieService.update(id, { 'storeConfig.appAccess.catalog': true });`
but then it doesn't match the movie type

Also, would a transaction be useful here since I first get the movie and then update it?

libs/import/src/lib/components/table-extracted-elements/contracts/contracts.component.ts
```diff
@@ -109,6 +111,13 @@ export class TableExtractedContractsComponent implements OnInit {
     const termIds = await this.termService.add(importState.terms);
     importState.contract.termIds = termIds;
     await this.contractService.add(importState.contract);
+    const titleId = importState.contract.titleId;
+    if (!!titleId && importState.contract.type === 'mandate') {
+      const title = await this.movieService.getValue(titleId);
+      title.storeConfig.appAccess.catalog = true;
```
GrandSchtroumpf, line 117: This should depend in the current app : 
```
title.storeConfig.appAccess[currentApp]
```

libs/import/src/lib/components/view-extracted-elements/contract/view-extracted-contracts.component.ts
```diff
@@ -241,11 +243,17 @@ export class ViewExtractedContractsComponent implements OnInit {
...
       if (!medias.length) errors.push(errorsMap['no-medias']);
 
-      // Languages
-      // TODO add languages
-
       const termId = this.fire.createId();
       const term = createTerm({ id: termId, contractId, orgId, duration, territories, medias, exclusive });
+
+      // Languages
+      Object.entries({ dubbed, subtitle, closedCaptioning }).forEach(([key, value]) => {
```
GrandSchtroumpf, line 250: Transform it into a for loop : 
```typescript
for (const [key, value] of Object.entries({ dubbed, subtitle, closedCaptioning })) {

}
```

GrandSchtroumpf (approved):

PR #5382, opened 2021-03-25
Remove transaction & fix flickering

GrandSchtroumpf:
- Remove transaction as it was blocking the rules, and there is not much risk of failure
- Keep memory of the array of title to avoid flickering


RemcoSimonides (approved):
LGTM

Issue #5383, opened 2021-03-25
End date can be in front of Start date


RemcoSimonides:
It should not be possible to choose an End date before the Start date

RemcoSimonides:
Also, you can select dates that are in the past

PR #5388, opened 2021-03-25
5383 dates should be correct

RemcoSimonides:
Fix issue #5383

### DIFF
--- a/libs/contract/src/lib/avails/form/avails.form.ts
+++ b/libs/contract/src/lib/avails/form/avails.form.ts
@@ -3,14 +3,34 @@ import { AvailsFilter } from '../avails';
 import { FormStaticValueArray } from '@blockframes/utils/form/forms/static-value.form';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
 
+function compareDates(form: FormControl) {
+  const eventForm = form?.parent as FormGroup;
+  if (eventForm) {
+    const { from, to } = eventForm.value;
+    if (from && to && from > to) {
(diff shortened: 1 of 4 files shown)


libs/contract/src/lib/avails/form/avails.form.ts
```diff
@@ -3,14 +3,34 @@ import { AvailsFilter } from '../avails';
 import { FormStaticValueArray } from '@blockframes/utils/form/forms/static-value.form';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
 
+function compareDates(form: FormControl) {
```
GrandSchtroumpf, line 6: We don't have something like that already in the validator file ?

GrandSchtroumpf (approved):

libs/utils/src/lib/form/validators/validators.ts
```diff
@@ -150,3 +150,33 @@ export function isKeyArrayValidator(scope: Scope): ValidatorFn {
...
+      const from = keyOnControl === fromKey ? form.value : eventForm.value[fromKey];
+      const to = keyOnControl === toKey ? form.value : eventForm.value[toKey]
+
+      if (from && to && from > to) {
+        return { startOverEnd: true }
+      }
+    }
+    return null;
+  }
+}
```
RemcoSimonides, line 173: Adding the keyOnControl parameter helps fixing a bug where validation came one valueChange in the form too late.
The value of the changed control was not yet in the parent form

PR #5406, opened 2021-03-30
UT for permission collection

xmano:
Unit tests suite for checking the permission and documentpermission collection.

image

### DIFF
--- a/apps/rules/src/tests/firestore/fixtures/data.ts
+++ b/apps/rules/src/tests/firestore/fixtures/data.ts
@@ -44,6 +44,10 @@ export const testFixture = {
   'orgs/O002': {
     status: 'accepted'
   },
+  'orgs/O003': {
+    userIds: ['uid-user3'],
+    status: 'pending'
+  },
 
   //Users Collection
   'users/uid-bfAdmin': {
--- a/apps/rules/src/tests/firestore/fixtures/data.ts
+++ b/apps/rules/src/tests/firestore/fixtures/data.ts
@@ -45,7 +45,7 @@ export const testFixture = {
     status: 'accepted'
   },
   'orgs/O003': {
-    userIds: ['uid-user3'],
+    userIds: ['uid-user2'],
     status: 'pending'
   },
 
--- a/apps/rules/src/tests/firestore/fixtures/data.ts
+++ b/apps/rules/src/tests/firestore/fixtures/data.ts
@@ -8,6 +8,7 @@ export const testFixture = {
       'uid-user2': 'member',
       'uid-sAdmin': 'superAdmin',
     },
+    id: 'O001',
   },
 
   'permissions/O001/documentPermissions/D001': {
--- a/apps/rules/src/tests/firestore/permission.spec.ts
+++ b/apps/rules/src/tests/firestore/permission.spec.ts
@@ -1,9 +1,8 @@
-﻿import { apps } from '@firebase/rules-unit-testing';
+﻿import { apps, assertSucceeds } from '@firebase/rules-unit-testing';
 import { testFixture } from './fixtures/data';
 import { Firestore, initFirestoreApp } from '@blockframes/testing/firebase/functions';
 
 
-//TODO: 4198
 describe('Permission Rules Tests', () => {
   const projectId = `permrules-spec-${Date.now()}`;
   let db: Firestore;
@@ -15,16 +14,35 @@ describe('Permission Rules Tests', () => {
 
     afterAll(() => Promise.all(apps().map((app) => app.delete())));
 
-    test('test', async () => {});
+    test('user with member role should be able to read document', async () => {
+      const permissionDocRef = db.doc('permissions/O001');
+      await assertSucceeds(permissionDocRef.get());
+    });
   });
 
   describe('With User as Org Admin', () => {
     beforeAll(async () => {
-      db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-user2' });
+      db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-admin' });
     });
 
     afterAll(() => Promise.all(apps().map((app) => app.delete())));
 
-    test('test', async () => {});
+    test('user with admin role should be able to read document', async () => {
+      const permissionDocRef = db.doc('permissions/O001');
+      await assertSucceeds(permissionDocRef.get());
+    });
   });
+
+  describe('With User as Super Admin', () => {
+    beforeAll(async () => {
+      db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-sAdmin' });
+    });
+
+    afterAll(() => Promise.all(apps().map((app) => app.delete())));
+
+    test('user with superadmin role should be able to read document', async () => {
+      const permissionDocRef = db.doc('permissions/O001');
+      await assertSucceeds(permissionDocRef.get());
+    });
+  });  
 });
--- a/apps/rules/src/tests/firestore/permission.spec.ts
+++ b/apps/rules/src/tests/firestore/permission.spec.ts
@@ -44,5 +44,5 @@ describe('Permission Rules Tests', () => {
       const permissionDocRef = db.doc('permissions/O001');
       await assertSucceeds(permissionDocRef.get());
     });
-  });  
+  });
 });
--- a/apps/rules/src/tests/firestore/permission.spec.ts
+++ b/apps/rules/src/tests/firestore/permission.spec.ts
@@ -8,11 +8,23 @@ describe('Permission Rules Tests', () => {
   const projectId = `permrules-spec-${Date.now()}`;
   let db: Firestore;
 
-  beforeAll(async () => {
-    db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-user2' });
+  describe('With User as Org Member', () => {
+    beforeAll(async () => {
+      db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-user2' });
+    });
+
+    afterAll(() => Promise.all(apps().map((app) => app.delete())));
+
+    test('test', async () => {});
   });
 
-  afterAll(() => Promise.all(apps().map((app) => app.delete())));
+  describe('With User as Org Admin', () => {
+    beforeAll(async () => {
+      db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-user2' });
+    });
 
-  test('test', async () => {});
+    afterAll(() => Promise.all(apps().map((app) => app.delete())));
+
+    test('test', async () => {});
+  });
 });
--- a/apps/rules/src/tests/firestore/permission.spec.ts
+++ b/apps/rules/src/tests/firestore/permission.spec.ts
@@ -24,6 +24,19 @@ describe('Permission Rules Tests', () => {
       await assertFails(permissionDocRef.update({note: 'This is a test'}));
     });
 
+    test('id different from orgId, should not be able to create document', async () => {
+      const docName = 'permissions/O123';
+      const docData = {id: 'O111', note: 'id is different from orgId'};
+      const permissionDocRef = db.doc(docName);
+      await assertFails(permissionDocRef.set(docData));
+    });
+
+    test('id not set; org status pending, should be able to create document', async () => {
+      const docName = 'permissions/O003';
+      const docData = {note: 'id is different from orgId'};
+      const permissionDocRef = db.doc(docName);
+      await assertSucceeds(permissionDocRef.set(docData));
+    });
   });
 
   describe('With User as Org Admin', () => {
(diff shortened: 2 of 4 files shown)


xmano:
One observation with respect to firestore.rules, 

https://github.com/blockframes/blockframes/blob/565997f751c288af8981783cea3045e3e34de50c/firestore.rules#L530

The line 
```
          exists(/databases/$(database)/documents/orgs/$(orgId)) == false
```

seems redundant and can be removed. Let me know..

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,155 @@
...
+    test('changing doc Id, should not be able to update document', async () => {
+      const permissionDocRef = db.doc('permissions/O001/documentPermissions/D001');
+      await assertFails(permissionDocRef.update({id: 'P111', note: 'This is a test'}));
+    });
+
+    test('should be able to update document', async () => {
+      const permissionDocRef = db.doc('permissions/O001/documentPermissions/D001');
+      await assertSucceeds(permissionDocRef.update({note: 'This is a test'}));
+    });
+  });
```
phpgeekfr, line 154: can you also add a test to check if `isBrandNewDocument` is working as expected upon documentPermissions creation?

phpgeekfr:
can you change target for fix-for-release-3-0 ?

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,155 @@
...
-  beforeAll(async () => {
-    db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-user2' });
+  describe('With User as Org Member', () => {
+    beforeAll(async () => {
+      db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-user2' });
+    });
+
+    afterAll(() => Promise.all(apps().map((app) => app.delete())));
+
+    test('should be able to read document', async () => {
```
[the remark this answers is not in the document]
xmano: See common comment for test description below.

I have added test for user not belonging to org.

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,155 @@
...
+      const permissionDocRef = db.doc('permissions/O001');
+      await assertSucceeds(permissionDocRef.get());
+    });
+
+    test('should not be able to update document', async () => {
```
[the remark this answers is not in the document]
xmano: see common note below..

xmano:
Common comment about change in test description - it should be already covered in describe("comment")

image

So I think this should be fine.l

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,155 @@
...
+      const permissionDocRef = db.doc('permissions/O001');
+      await assertSucceeds(permissionDocRef.get());
+    });
+
+    test('should be able to update document', async () => {
+      const permissionDocRef = db.doc('permissions/O001');
+      await assertSucceeds(permissionDocRef.update({note: 'This is a test'}));
+    });
+
+    test('when updating id, should not be able to update document', async () => {
```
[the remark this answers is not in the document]
xmano: This test is already done with user role. 

```
      allow create: if isRequestFieldValueEqualToOrNotSetted(['id'], orgId)
        && isCreatingOrganization(orgId)
        && notInMaintenance();
```

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,155 @@
...
+      const permissionDocRef = db.doc('permissions/O001');
+      await assertFails(permissionDocRef.update({id: 'O002', note: 'This is a test'}));
+    });
   });
+});
```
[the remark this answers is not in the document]
xmano: I am looking into this..

apps/rules/src/tests/firestore/permission.spec.ts
xmano: Added :

```
    test('docIndex exists, should not be able to create document', async () => {
      const permissionDocRef = db.doc('permissions/O001/documentPermissions/DP03');
      await assertFails(permissionDocRef.set({note: 'This is a create test for existing docIndex'}));
    });
```

And it passes.

xmano:
Updated and ran tests.. All green :green_heart: :green_heart:

image

apps/rules/src/tests/firestore/permission.spec.ts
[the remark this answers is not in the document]
xmano: added the test:

```
    //test the scenario when org and permission is created at the same time..
    test('org not yet created, should be able to create document', async () => {

```

apps/rules/src/tests/firestore/fixtures/data.ts
```diff
@@ -1,13 +1,19 @@
...
   'permissions/O001': {
     roles: {
       'uid-admin': 'admin',
       'uid-user2': 'member',
       'uid-sAdmin': 'superAdmin',
     },
+    id: 'O001',
   },
 
   'permissions/O001/documentPermissions/D001': {
```
phpgeekfr, line 19: regarding new rules, every `documentPermissions` should have an `id` as attribute

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,180 @@
...
+      const permissionDocRef = db.doc('permissions/O003');
+      await assertFails(permissionDocRef.get());
+    });
+
+    test('should not be able to update document', async () => {
+      const permissionDocRef = db.doc('permissions/O001');
+      await assertFails(permissionDocRef.update({note: 'This is a test'}));
+    });
+
+    test('id different from orgId, should not be able to create document', async () => {
```
phpgeekfr, line 32: this `assertFails` is not only failing because of bad orgId but also because org document already exists.
To test only `id different from orgId` use a transaction were you create the org but use another id for doc permission.

If you make the same test with good orgId instead, it will fail

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,180 @@
...
+      const docName = 'permissions/O004';
+      const docData = {id: 'O004', roles: { 'uid-user2': 'member' }, note: 'Creating org and permission doc'};
+      const permissionDocRef = db.doc(docName);
+      const orgDoc = 'orgs/O004';
+      const orgUserIds = ['uid-user2'];
+
+      await db.runTransaction(async tx => {
+        // create org
+        const orgRef = db.doc(orgDoc);
+        //orgRef.set({status: 'pending', userIds: orgUserIds});
```
phpgeekfr, line 50: unnecessary comment

apps/rules/src/tests/firestore/permission.spec.ts
phpgeekfr: maybe create a test "permission doc must be created along with org in same transaction" ?

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,180 @@
...
+    });
+
+    afterAll(() => Promise.all(apps().map((app) => app.delete())));
+
+    test('should be able to read document', async () => {
+      const permissionDocRef = db.doc('permissions/O001/documentPermissions/D001');
+      await assertSucceeds(permissionDocRef.get());
+    });
+
+    test('id not set, should be able to create document', async () => {
```
phpgeekfr, line 127: not accurate anymore

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,180 @@
...
+
+    afterAll(() => Promise.all(apps().map((app) => app.delete())));
 
-  test('test', async () => {});
+    test('should be able to read document', async () => {
+      const permissionDocRef = db.doc('permissions/O001/documentPermissions/D001');
+      await assertSucceeds(permissionDocRef.get());
+    });
+
+    test('id not set, should be able to create document', async () => {
```
phpgeekfr, line 160: same as above

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,180 @@
...
+  let db: Firestore;
+
+  describe('With User as Org Member', () => {
+    beforeAll(async () => {
+      db = await initFirestoreApp(projectId, 'firestore.rules', testFixture, { uid: 'uid-user2' });
+    });
+
+    afterAll(() => Promise.all(apps().map((app) => app.delete())));
+
+    test('should be able to read document', async () => {
```
phpgeekfr, line 122: add another test with user not in org (same thing that you did for permission)

phpgeekfr (changes requested):

apps/rules/src/tests/firestore/permission.spec.ts
xmano: fixed

apps/rules/src/tests/firestore/permission.spec.ts
xmano: fixed

apps/rules/src/tests/firestore/fixtures/data.ts
xmano: added

apps/rules/src/tests/firestore/permission.spec.ts
xmano: removed

apps/rules/src/tests/firestore/permission.spec.ts
xmano: added
image

apps/rules/src/tests/firestore/permission.spec.ts
xmano: There are 2 tests. 
Test 1 : when id is set and different from Org Id (Here we do not check isCreatingOrganization condition) because the first condition itself fails.

Test 2 : org and permission doc created in the same transaction - it is alread there..

```
    test('id different from orgId, should not be able to create document', async () => {
      const docName = 'permissions/O123';
      const docData = {id: 'O111', note: 'id is different from orgId'};
      const permissionDocRef = db.doc(docName);
      await assertFails(permissionDocRef.set(docData));
    });

    //test the scenario when org and permission is created at the same time..
    test('org not yet created, should be able to create document', async () => {
      const docName = 'permissions/O004';
      const docData = {id: 'O004', roles: { 'uid-user2': 'member' }, note: 'Creating org and permission doc'};
      const permissionDocRef = db.doc(docName);
      const orgDoc = 'orgs/O004';
      const orgUserIds = ['uid-user2'];

      await db.runTransaction(async tx => {
        const orgRef = db.doc(orgDoc);
        tx.set(orgRef, {status: 'pending', userIds: orgUserIds});
        tx.set(permissionDocRef, docData);
      });

      await assertSucceeds(permissionDocRef.get());
    });
```

phpgeekfr: regarding `maybe create a test "permission doc must be created along with org in same transaction" ?` / test2:
I was thinking of a falling test when you  use  same `id` attribute as document Id but without creating the org in same transaction.

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,181 @@
...
+      const permissionDocRef = db.doc('permissions/O003');
+      await assertFails(permissionDocRef.get());
+    });
+
+    test('should not be able to update document', async () => {
+      const permissionDocRef = db.doc('permissions/O001');
+      await assertFails(permissionDocRef.update({note: 'This is a test'}));
+    });
+
+    test('id different from orgId, should not be able to create document', async () => {
```
phpgeekfr, line 32: in fact this is not `id different from orgId` but `id attribute different from documentId`. At this point we don't know if documentId is indeed an orgId

apps/rules/src/tests/firestore/permission.spec.ts
```diff
@@ -1,18 +1,181 @@
...
+      const docName = 'permissions/O123';
+      const docData = {id: 'O111', note: 'id is different from orgId'};
```
phpgeekfr, line 34: `id is different from orgId` => `id is different from documentId` (which should indeed  be an orgId)

apps/rules/src/tests/firestore/permission.spec.ts
xmano: Added the test as discussed.

xmano:
image

And documentation is here:
https://www.notion.so/cascade8/Security-Rules-3f3f1bc4b3a24090b75d2cc8e7e307b9

apps/rules/src/tests/firestore/permission.spec.ts
xmano: updated the description.

phpgeekfr (approved):
;)

PR #5421, opened 2021-03-30
fix Weird behavior when creating an org and permissions

### LINKED ISSUE #2710: Weird behavior when creating an org and permissions
Weird behavior when creating an org and permissions.

```
  sync() {
      // @TODO This async is needed to prevent permission denied
      // in the case where user.orgId exists but not the permissions document.
      // This seems to be caused by the creation/update order when we do
      // orgService.add()
      switchMap(async user => {
        if (!user) {
          return of('/');
        }
        if (!user.orgId) {
          return of('c/organization');
        } else {
          return this.service.syncActive({ id: user.orgId }).pipe(
            map(_ => this.query.getActive()),
            map(permissions => {
              if (!permissions) {
                return 'c/organization';
              }
            })
          );
        }
      })
    );
  }
```

phpgeekfr:
Thanks to @xmano help and new authflow (guard update I think), the weird behavior is gone.

fix #2710

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -526,11 +526,7 @@ service cloud.firestore {
 
     function isCreatingOrganization(orgId) {
 			return userId() in getAfter(/databases/$(database)/documents/orgs/$(orgId)).data.userIds
-      	&& (
-          exists(/databases/$(database)/documents/orgs/$(orgId)) == false
-          // Needed if org creation and permissions doc associated are not made in the same batch write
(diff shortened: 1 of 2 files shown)


xmano (approved):
LGTM.

Just need to check if there is still any case where org status  "Pending" state is involved. Maybe Clelia can review it?

Mussuranae (approved):
Look good to me also !

PR #5422, opened 2021-03-30
fix preview video bug

pldespaigne:
fix the bug that prevent to preview video from the file explorer

### DIFF
--- a/libs/media/src/lib/file/explorer/explorer.model.ts
+++ b/libs/media/src/lib/file/explorer/explorer.model.ts
@@ -1,5 +1,5 @@
 import { MediaRatioType } from '../../image/uploader/uploader.component';
-import { MovieForm } from "@blockframes/movie/form/movie.form";
+import { MovieForm, MovieVideoForm } from "@blockframes/movie/form/movie.form";
 import { OrganizationForm } from "@blockframes/organization/forms/organization.form";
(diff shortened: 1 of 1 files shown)


libs/media/src/lib/file/explorer/explorer.model.ts
```diff
@@ -89,6 +89,16 @@ function getFormListStorage(object: { id: string }, collection: CollectionHoldin
   return FormList.factory<StorageFile>(value, file => new StorageFileForm(file));
 }
 
+function getFormStorageVideo(object: { id: string }, collection: CollectionHoldingFile, label: FileLabel) {
```
RemcoSimonides, line 92: It wasn't possible to give the object a proper type?

RemcoSimonides (approved):
LGTM

libs/media/src/lib/file/explorer/explorer.model.ts
pldespaigne: it's not a type problem, the function `getFormStorage` is calling `new StorageFileForm` which does not create a field for `jwPlayerId` that's why I created a new function that instantiate the good form :)

RemcoSimonides: Ah ye I see, but I was talking about `object: { id: string }`. Isn't one of them a StorageFile and the other one a StorageVideo?

Issue #5411, opened 2021-03-30
RELEASE MVP: Invitations, notifications & emails


fritzschoff:
Trigger Notifications
Invitations (screening: check the status of the invitation)
Emails => Timezones

Jackseed:
user dev+neha-fja@blockframes.io
url https://staging.archipelcontent.com/c/o/marketplace/invitations
- [ ] As discussed, we should add the correct and full url of an app on invitations / notifications so they may worked cross-app. 
Example: Here I click on a screening invitation link from `catalog` which leads to `https://staging.archipelcontent.com/c/o/marketplace/event/Nxq2CNi7DscC0oekm43g`. It's a dead link as screenings only happen on archipelmarket.com.
If I manually change the url for `https://staging.archipelmarket.com/c/o/marketplace/event/Nxq2CNi7DscC0oekm43g` then it's working properly
![Kapture 2021-03-30 at 17 38 51](https://user-images.githubusercontent.com/14964443/113017683-02939580-9180-11eb-88de-e38152bc5a08.gif)
 
![image](https://user-images.githubusercontent.com/14964443/113017931-47b7c780-9180-11eb-883e-388a5bcf2321.png)

Jackseed:
- [ ] Change demo email text, remove "Archipel Content" to avoid misunderstandings from "A user wants to schedule a demo of Archipel Content." --> "A user wants to schedule a demo."

Jackseed:
- [ ] Movie accepted email is always triggering for Archipel Market even if the movie is only on Content (didn't test on MF but I guess it's the same)
template id: `Template ID: d-bfcf2760bcb7484ab55f864a59331d26`

Jackseed:
Invitations have flashes on loading:
https://user-images.githubusercontent.com/14964443/113551442-f29f0a00-95f4-11eb-8453-5f0772270189.mov

We had the same problem on other lists, here is a comment from @RemcoSimonides about it:
> The first time the screening appear on the screening list page, the data probably comes from the Cache. After that the data comes from firestore and thus rerenders the page.
> I see 4 blinks though, so probably this happens twice.
> On homepage I resolved something similar with a beautiful
> distinctUntilChanged((a, b) => JSON.stringify(a) === JSON.stringify(b) or something. That may solve half of the blinks if the data from firestore is the same as in the cache

Jackseed:
[...]

PR #5439, opened 2021-03-31
Full URL notification

Mussuranae:
Change relative url on invitation and notification to full url (cross app problem)
Part of #5411


libs/invitation/src/lib/components/item/item.component.ts
```diff
@@ -52,24 +55,25 @@ export class ItemComponent {
...
         }
         else {
-          return [`/c/o/marketplace/event`, this._invitation.eventId, 'session'];
+          return `${this.applicationUrl.festival}/c/o/marketplace/event/${this._invitation.eventId}/session`;
         }
       }
     } else if (this._invitation.type === 'joinOrganization') {
       const orgId = this._invitation.fromOrg ? this._invitation.fromOrg.id : this._invitation.toOrg.id;
-      return `/c/o/organization/${orgId}/view/members`;
+      return `${this.applicationUrl[this.app]}/c/o/organization/${orgId}/view/members`;
```
phpgeekfr, line 76: so now, even if invitation is not for an event, we have the complete url ? Does it make the entire page reload if we click on an absolute url but we stay on same url  ?

phpgeekfr (approved):
LGTM

libs/invitation/src/lib/components/item/item.component.ts
Mussuranae: Yes it reload the page and redirect you to the new url

libs/utils/src/lib/apps.ts
```diff
@@ -130,6 +132,18 @@ export function getOrgAppAccess(org: OrganizationDocument | Organization, first:
   }
 }
 
+/** Return an array of the app access of the movie */
+export function getMovieAppAccess(movie: MovieDocument | Movie): App[] {
+  const apps: App[] = [];
+  for(const a of app) {
+    const hasAccess = app.some(a => !!movie.storeConfig.appAccess[a]);
```
phpgeekfr: shadowed variable `a`

Issue #5414, opened 2021-03-30
RELEASE MVP: Meetings: File explorers, CRUD, stats, privacy, share documents, videoconference, invitations


fritzschoff:
Screening form (find someone by the orgname)

fritzschoff:
![image](https://user-images.githubusercontent.com/39312833/112992420-71fd8b00-9168-11eb-8be8-e13036065b98.png)
there is a file that is in the org which just got created by prepare for testing?

fritzschoff:
![image](https://user-images.githubusercontent.com/39312833/112992965-fb14c200-9168-11eb-975c-4bba66f45df6.png)
when I left the meeting, I got an error

Jackseed:
- [ ] If that's not too big an issue, add an eye as a preview button on the bottom left side for screeners & sales pitch in My Files section, same as the button for other videos
![image](https://user-images.githubusercontent.com/14964443/113300504-44e2e100-92fe-11eb-8a9d-15e097d49834.png)

Jackseed:
- [ ] My Files: upload "Other video" buggy - mb missing a video type?
user: dev+maryam-och@blockframes.io
url: https://staging.archipelmarket.com/c/o/dashboard/files 
sentry: https://sentry.io/organizations/cascade-8/issues/2318969405/?referrer=slack

![Kapture 2021-04-05 at 15 01 46](https://user-images.githubusercontent.com/14964443/113576486-ec725300-961f-11eb-92aa-92d08035c9ec.gif)

steps to reproduce: 
- go to url above with dashboard user
- choose a title and other video folder
- upload a video
- refresh after upload, Files aren't loading anymore

PR #5441, opened 2021-03-31
Meeting file bugs

pldespaigne:
fix bugs (0), (2) & (5) from issue #5414


libs/event/src/lib/components/meeting/+state/twilio.service.ts
```diff
@@ -31,6 +31,8 @@ export class TwilioService {
   private room: Room;
   private getAccessToken = this.functions.httpsCallable('getAccessToken');
 
+  private preference: {[K in TrackKind]: boolean} = { 'video': true, 'audio': true };
```
phpgeekfr, line 34: you can also do:
```typescript
private preference: Record<TrackKind, boolean> = { 'video': true, 'audio': true };
```

phpgeekfr (approved):
LGTM

Issue #5410, opened 2021-03-30
RELEASE MVP: Avails: create, delete, view


fritzschoff:
Create avail, check the selection page, create offer and validate the email
Cards on selection page

Jackseed:
- [ ] the avails filters were cleared, but the btn remains active. If filters cleared need to change back to initiate state. 
![avails](https://user-images.githubusercontent.com/55499036/113128316-f78d4380-9219-11eb-94aa-ac4662de548d.gif)

Jackseed:
### My Selection page

- [ ] After touching the filters, the column "Action" disappears 
![10](https://user-images.githubusercontent.com/79836137/113127780-7170fd00-9219-11eb-8db7-06603dec4404.gif)

- [ ] Background visible behind rounded corner of drop down
<img width="1440" alt="Screenshot 2021-03-31 at 12 07 48" src="https://user-images.githubusercontent.com/79836137/113128244-e47a7380-9219-11eb-8216-553f7c9c5640.png">

Jackseed:
### Avails filter
- [ ] If possible, could we select by default the typing zone when user clicks on the "Territories" select?
- [ ] When typing a wrong value on Territories, the select doesn't open anymore
https://user-images.githubusercontent.com/14964443/113309166-27664500-9307-11eb-8747-bf293dcbca70.mp4

Jackseed:
- [ ] Change the territory list displayed in the avails filter for this one
```

export const territoriesGroup: StaticGroup<"territories">[] = [
  {
    label: "Europe",
    items: [
      "aland-islands",
      "albania",
      "andorra",
      "austria",
      "belgium",
      "bosnia-and-herzegovina",
      "bulgaria",
      "croatia",
      "czech",
      "denmark",
      "estonia",
      "faroe-islands",
      "finland",
      "france",
      "germany",
      "gibraltar",
      "greece",
      "guernsey",
      "vatican",
      "hungary",
      "iceland",
      "ireland",
      "isle-of-man",
      "italy",
      "jersey",
      "latvia",
      "liechtenstein",
      "lithuania",
      "luxembourg",
      "malta",
      "monaco",
      "montenegro",
      "netherlands",
      "north-macedonia",
      "norway",
      "poland",
      "portugal",
      "romania",
      "san-marino",
      "serbia",
      "slovakia",
      "slovenia",
      "spain",
      "svalbard-and-jan-mayen",
      "sweden",
      "switzerland",
      "ukraine",
[...]

PR #5454, opened 2021-04-01
5410 fix avails filter

RemcoSimonides:
Fix issue #5410 
- [x] the avails filters were cleared, but the btn remains active. If filters cleared need to change back to initiate state.
- [x] Not able to select various territories. When selection a new territory by typing, it deletes the one previously selected
- [x] Add a tooltip "Clear text" on the icon of the round arrow

### DIFF
--- a/libs/ui/src/lib/static-autocomplete/group/group.component.ts
+++ b/libs/ui/src/lib/static-autocomplete/group/group.component.ts
@@ -11,7 +11,7 @@ import {
   NG_VALUE_ACCESSOR
 } from "@angular/forms";
 import { BehaviorSubject, combineLatest, Observable, Subscription, defer } from "rxjs";
-import { map, startWith, shareReplay } from "rxjs/operators";
+import { map, startWith, shareReplay, pairwise } from "rxjs/operators";
 import { Scope, StaticGroup, staticGroups, staticModel } from '@blockframes/utils/static-model';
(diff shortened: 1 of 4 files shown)


libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -66,6 +66,7 @@ function getRootMode(groups: StaticGroup[], value: string[]): GroupMode {
 export class StaticGroupComponent implements ControlValueAccessor {
   private _scope: Scope;
   private sub?: Subscription;
+  private formSub: Subscription;
```
fritzschoff, line 69: you also just have a subs: Subscription[], and you push the subscription into it and in the ngOnDestroy you loop over it an unsubscribe

libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -120,6 +134,7 @@ export class StaticGroupComponent implements ControlValueAccessor {
 
   ngOnDestroy() {
     this.sub?.unsubscribe();
+    this.formSub.unsubscribe();
```
fritzschoff, line 137: maybe add ? if every this sub didn't started

fritzschoff (approved):

fritzschoff (approved):

Issue #5413, opened 2021-03-30
RELEASE MVP: Screenings: CRUD, invitations, privacy, stats, view, session, banners


fritzschoff:
Screening lobby + session (modale when leaving; + check that modale when log out)
Screening form (find someone by the orgname)

RemcoSimonides:
Please check with Vincent if it's even worth fixing this:
- [ ] I created a screening without selecting a movie. When I attend the screening and I want to leave, it shows me this warning but it's not necessary now
![image](https://user-images.githubusercontent.com/27687382/113112708-c9076c80-9209-11eb-95ed-e6ae832b0de8.png)

- [ ] It is possible to create a screening without a title and/or without selecting a movie. The event in the calendar shows up as just an empty block. Should it even be possible to create a screening without these fields filled in?
![image](https://user-images.githubusercontent.com/27687382/113112888-fd7b2880-9209-11eb-9e3c-0c03a3332a59.png)
The email sent to the invited user not contain the title, which makes the sentence incorrect but actually not too bad
![image](https://user-images.githubusercontent.com/27687382/113114187-5eefc700-920b-11eb-90db-11d3ee268b79.png)

RemcoSimonides:
- [ ] Event without a movie (meta.titleId) selected doesn't show up on [screeninglist page](https://staging.archipelmarket.com/c/o/marketplace/event)

Jackseed:
@RemcoSimonides that's indeed an issue but it should be handled on the screening creation more than session & email. That's good that this screening isn't referenced in the screening list as it shouldn't exist in the first place. 

The thing is that since the small event form is shared with meeting we can't ask for a movie in the first place. 

We should take care of that in the screening form but then it becomes an UX issue as some users do create an event before having the movie prepared (screener file mainly). 

So we should just leave for now your 1st and 3rd tick box. Though the 2nd one is a good catch and we should add event title as a mandatory field that can't be empty

RemcoSimonides:
- [ ] Cannot find some users when searching by org name
This is probably caused because of Algolia not having the org name yet, but I think it should have because the user is part of an org.
![image](https://user-images.githubusercontent.com/27687382/113118560-dfb0c200-920f-11eb-8973-51a10548a5cd.png)
In Algolia:
![image](https://user-images.githubusercontent.com/27687382/113118581-e50e0c80-920f-11eb-8018-8fcca242622f.png)

Jackseed:
[...]

PR #5456, opened 2021-04-01
Feat 5413 fix blinking

phpgeekfr:
part of #5413  . I added a `debounceTime` 500ms to prevent blinking.
This seems to do the trick but IMO best solution would be not using observables where we don't need it.


RemcoSimonides (approved):
LGTM

Issue #5407, opened 2021-03-30
RELEASE MVP: Movie CRUD, list, view, wishlist 


fritzschoff:
TEST:
Movie list
Movie view => tabs (all apps)
Movie filters (all apps) 
wishlist (check the CRM delete movie implications)
Delete: CRM

Jackseed:
- [ ] Marketplace title list, update "Format" filter list with the same as movie form ordered alphabetically i.e. 
- Documentary
- Feature Film
- Flow
- Short Film
- TV Film 
- TV Series
marketplace
![image](https://user-images.githubusercontent.com/14964443/113005787-0bcb3500-9175-11eb-959a-30bc2af373f3.png)
vs movie form
![image](https://user-images.githubusercontent.com/14964443/113005801-0ff75280-9175-11eb-91e2-b52da162867d.png)

fritzschoff:
@phpgeekfr I see in the code you did something to the component, is this issue done?

phpgeekfr:
I did the first "checkbox" but I did not closed it so we can ad new bugs if any

fritzschoff:
alright

Jackseed:
user vchoukroun+1@cascade8.com
url https://staging.archipelcontent.com/c/o/dashboard/title
- [ ] Small UI bug on empty catalog dashboard, button text are a bit different on empty "My Titles" than the empty Dashboard page. The space from empty dashboard is the good one

dashboard:
![image](https://user-images.githubusercontent.com/14964443/113126720-391cef00-9218-11eb-9ee1-f0de32fa7f21.png)

my titles
![image](https://user-images.githubusercontent.com/14964443/113126735-3de1a300-9218-11eb-9080-2614725bb3a1.png)

fritzschoff:
done

Jackseed:
### Dashboard 
#### Archipel Market
- [ ] Change text "Import Movies" to "Import Titles"

#### Archipel Content
**My Titles**

- [ ] Add "Import Titles" button and "+" for adding one title as it's in Market 

- [ ] Align the title and the buttons (same as Market)

from ![Снимок экрана 2021-03-31 в 10.34.01.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/db2a50a1-fea4-4959-8beb-86f5dbf2acbd)

to 
[...]

PR #5460, opened 2021-04-01
Issue 5407 movie form

fritzschoff:
ref #5407


libs/utils/src/lib/apps.ts
```diff
@@ -62,7 +63,7 @@ export const applicationUrl: Record<App, string> = {
   financiers: appUrl.financiers
 }
 
-export function getCurrentApp(routerQuery: any): App {
+export function getCurrentApp(routerQuery: RouterQuery): App {
```
phpgeekfr, line 66: I don't know if this is a good idea because some functions (`getAppName` for example) are used in backend. Adding this import will include akita-ng-router into backend

phpgeekfr (approved):

PR #5464, opened 2021-04-01
Add data on notification

Mussuranae:
Add the `createdFrom` on the email to get easily the app from which the email is sent instead of guessing it by the org app access that isn't so relevant for some emails.
Part of #5411


apps/backend-functions/src/orgs.ts
```diff
@@ -11,21 +11,23 @@ import { sendMail } from './internals/email';
...
 import { ErrorResultResponse } from './utils';
 import { cleanOrgMedias } from './media';
 import { Change, EventContext } from 'firebase-functions';
 import { algolia, deleteObject, storeSearchableOrg, findOrgAppAccess, hasAcceptedMovies, storeSearchableUser } from '@blockframes/firebase-utils';
 import { CallableContext } from 'firebase-functions/lib/providers/https';
 
 /** Create a notification with user and org. */
-function notifyUser(toUserId: string, notificationType: NotificationTypes, org: OrganizationDocument, user: PublicUser) {
+async function notifyUser(toUserId: string, notificationType: NotificationTypes, org: OrganizationDocument, user: PublicUser) {
+  const createdFrom = await getOrgAppKey(org);
```
phpgeekfr, line 24: instead of transforming this method into an async one, since you already have org object you should use `getOrgAppAccess(org)[0]` instead of `getOrgAppKey(org)`

apps/backend-functions/src/orgs.ts
```diff
@@ -53,7 +55,8 @@ async function notifyOnOrgMemberChanges(before: OrganizationDocument, after: Org
     const userSnapshot = await db.doc(`users/${userAddedId}`).get();
     const userAdded = userSnapshot.data() as PublicUser;
 
-    const notifications = after.userIds.filter(userId => userId !== userAdded.uid).map(userId => notifyUser(userId, 'orgMemberUpdated', after, userAdded));
+    const promises = after.userIds.filter(userId => userId !== userAdded.uid).map(userId => notifyUser(userId, 'orgMemberUpdated', after, userAdded));
```
phpgeekfr, line 58: you can revert to old code if you update `notifyUser` as suggested above

apps/backend-functions/src/orgs.ts
```diff
@@ -64,7 +67,8 @@ async function notifyOnOrgMemberChanges(before: OrganizationDocument, after: Org
 
     await removeMemberPermissionsAndOrgId(userRemoved);
 
-    const notifications = after.userIds.map(userId => notifyUser(userId, 'orgMemberUpdated', after, userRemoved));
+    const promises = after.userIds.map(userId => notifyUser(userId, 'orgMemberUpdated', after, userRemoved));
```
phpgeekfr, line 70: you can revert to old code if you update `notifyUser` as suggested above

apps/backend-functions/src/orgs.ts
```diff
@@ -121,12 +125,14 @@ export async function onOrganizationUpdate(change: Change<FirebaseFirestore.Docu
   const becomeAccepted = before.status === 'pending' && after.status === 'accepted';
 
   if (becomeAccepted) {
+    const appAccess = await getOrgAppKey(after)
```
phpgeekfr, line 128: same, I thing you can use directly `getOrgAppAccess(org)[0]`

phpgeekfr (approved):

PR #5470, opened 2021-04-02
5410 fix avails filter

RemcoSimonides:
Todo's in issue #5410
- [x] After touching the filters, the column "Action" disappears
- [x] Background visible behind rounded corner of drop down
- [x] If possible, could we select by default the typing zone when user clicks on the "Territories" select?
- [x] When typing a wrong value on Territories, the select doesn't open anymore

### DIFF
--- a/libs/ui/src/lib/overlay-widget/overlay-widget.component.ts
+++ b/libs/ui/src/lib/overlay-widget/overlay-widget.component.ts
@@ -1,4 +1,4 @@
-import { Component, TemplateRef, ViewChild, Directive, ViewEncapsulation, ViewContainerRef, OnDestroy, ElementRef, Output, EventEmitter } from '@angular/core';
+import { Component, TemplateRef, ViewChild, Directive, ViewEncapsulation, ViewContainerRef, OnDestroy, ElementRef, Output, EventEmitter, Input } from '@angular/core';
 import { trigger, state, style, transition, animate } from '@angular/animations';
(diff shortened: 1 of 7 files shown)


libs/ui/src/lib/overlay-widget/overlay-widget.component.ts
```diff
@@ -18,16 +18,12 @@ const fade = trigger('fade', [
...
   template: `
   <ng-template #ref>
-    <div tabindex="-1" role="menu" @fade class="bf-widget">
+    <div tabindex="-1" role="menu" @fade class="bf-widget" [ngStyle]="{ 'overflow': overflow }">
       <ng-content></ng-content>
     </div>
   </ng-template>
   `,
   styles: [`
-    .bf-widget {
```
fritzschoff, line 27: you removed the class but referencing it in line 21

fritzschoff: also I dont see the class overflow?

fritzschoff (changes requested):

libs/ui/src/lib/overlay-widget/overlay-widget.component.ts
RemcoSimonides: class is still used for `.bf-widget:focus`

fritzschoff (approved):

Issue #5415, opened 2021-03-30
RELEASE MVP: CRM: stats


fritzschoff:

Jackseed:
- [ ] On movie view links to marketplace pages are dead (currently `c/o/marketplace/movieId/view/main` while it should be `/c/o/marketplace/title/movieId/main`
![image](https://user-images.githubusercontent.com/14964443/113272288-c07f6680-92db-11eb-9f7f-667fd5f1e2a0.png)

Jackseed:
- [ ] User list loading is still very long, don't know if that's something that can be fixed for the release

Jackseed:
url https://staging.crm.blockframes.io/c/o/admin/panel/movie/ZG4aXHI76Wvxb6CbUBk9
user dev+hettie-fvh@blockframes.io
- [ ] Same as above it's more of a new feature than bug fix so it can be skipped for the release if too big but would be great to add the sales pitch video upload box & the stills images upload box on the CRM movie page for the ressource section
![image](https://user-images.githubusercontent.com/14964443/113281887-76e84900-92e6-11eb-9345-0e8a6dfa0dee.png)

RemcoSimonides:
> url https://staging.crm.blockframes.io/c/o/admin/panel/movie/ZG4aXHI76Wvxb6CbUBk9
> user [dev+hettie-fvh@blockframes.io](mailto:dev+hettie-fvh@blockframes.io)
> 
> * [ ]  Same as above it's more of a new feature than bug fix so it can be skipped for the release if too big but would be great to add the sales pitch video upload box & the stills images upload box on the CRM movie page for the ressource section

Promotional Images are still photos, right? They are in there already. sales pitch is added :)

Jackseed:
Cool thanks @RemcoSimonides! and good catch for the promo images I missed it ><

PR #5480, opened 2021-04-02
5415 Improve users loading, add sales pitch, fix link

RemcoSimonides:
Fix issue #5415 
- [x] On movie view links to marketplace pages are dead (currently c/o/marketplace/movieId/view/main while it should be /c/o/marketplace/title/movieId/main
- [x] User list loading is still very long, don't know if that's something that can be fixed for the release
- [x] Same as above it's more of a new feature than bug fix so it can be skipped for the release if too big but would be great to add the sales pitch video upload box & the stills images upload box on the CRM movie page for the ressource section

### DIFF
--- a/libs/movie/src/lib/movie/components/video-upload/video-upload.component.ts
+++ b/libs/movie/src/lib/movie/components/video-upload/video-upload.component.ts
@@ -35,7 +35,7 @@ export class MovieVideoUploadComponent implements OnInit {
 
   async ngOnInit() {
     this.form = new MovieVideosForm(this.movie.promotional.videos);
-    this.salesPitchForm = new MovieSalesPitchForm(this.movie.promotional.salesPitch)
+    this.salesPitchForm = new MovieSalesPitchForm(this.movie.promotional.salesPitch);
     // Add empty upload zone
     this.form.otherVideos.add({ ref: '' });
     this.cdr.markForCheck();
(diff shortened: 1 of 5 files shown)


libs/movie/src/lib/movie/components/video-upload/video-upload.component.ts
```diff
@@ -27,11 +30,12 @@ export class MovieVideoUploadComponent implements OnInit {
     private cdr: ChangeDetectorRef
   ) {
     this.form = new MovieVideosForm();
+    this.salesPitchForm = new MovieSalesPitchForm();
   }
 
   async ngOnInit() {
     this.form = new MovieVideosForm(this.movie.promotional.videos);
-
+    this.salesPitchForm = new MovieSalesPitchForm(this.movie.promotional.salesPitch)
```
phpgeekfr, line 38: missing `;`

phpgeekfr (approved):

libs/movie/src/lib/movie/components/video-upload/video-upload.component.ts
RemcoSimonides: Ohnooooooo! how could I?!

phpgeekfr: You should get fired for this !

PR #5487, opened 2021-04-06
Issue 5407 wishlist

fritzschoff:


libs/ui/src/lib/layout/marketplace/marketplace.component.ts
```diff
@@ -38,11 +41,21 @@ export class MarketplaceComponent implements OnInit {
...
+    private movieService: MovieService,
+    private routerQuery: RouterQuery
   ) { }
 
   ngOnInit() {
     this.wishlistCount$ = this.orgQuery.selectActive().pipe(
-      map(org => org.wishlist?.length || 0)
+      map(org => org.wishlist),
+      switchMap(movieIds => this.movieService.getValue(movieIds)),
+      map(movies => movies.filter(movie => {
```
GrandSchtroumpf, line 52: I think you need to return after the for loop.
Also, create an external function for that :
```typescript
export const filterMovieByAppAccess = (app: App) => (movie: Movie) => {
  for (const app in movie.storeConfig.appAccess) {
    if (movie.storeConfig.appAccess[app] && currentApp === app) return true;
  }
  return false;
}
....
map(movies => movies.filter(filterMovieByAppAccess(getCurrentApp(this.routerQuery)))
```

GrandSchtroumpf (approved):

PR #5488, opened 2021-04-06
Fix for release 3 0

phpgeekfr:


libs/movie/src/lib/movie/form/movie.shell.config.ts
```diff
@@ -77,6 +77,15 @@ export class MovieShellConfig implements FormShellConfig<MovieControl, Movie> {
         if (this.form.productionStatus.value) {
           movie.productionStatus = this.form.productionStatus.value
         }
+
+        if (this.form.release.get('status').value) {
+          movie.release.status = this.form.release.get('status').value
+        }
+
+        if (this.form.runningTime.get('status').value) {
+          movie.runningTime.status = this.form.runningTime.get('status').value
```
SamirTouira, line 86: is it ok if there aren't semicolon here? if no need you can skip this comment

SamirTouira (approved):

libs/movie/src/lib/movie/form/movie.shell.config.ts
phpgeekfr: Better with `;` but we can skip

PR #5489, opened 2021-04-06
quick fix

pldespaigne:
fix (6) & (7) from #5414

### DIFF
--- a/libs/media/src/lib/file/explorer/explorer.component.html
+++ b/libs/media/src/lib/file/explorer/explorer.component.html
@@ -43,7 +43,10 @@
   <file-uploader [form]="dir.form" [meta]="dir.meta" [accept]="dir.accept" (selectionChange)="change($event, dir.meta)">
     <h3 title>Upload {{ dir.name }}</h3>
   </file-uploader>
-  <section fxLayout="row" fxLayoutAlign="end center" class="save">
+  <section fxLayout="row" fxLayoutAlign="space-between center" class="save">
+    <div>
+      <bf-preview-modal *ngIf="!!dir.form.value.storagePath" [ref]="dir.form.value"></bf-preview-modal>
+    </div>
     <button mat-stroked-button (click)="update()">
       <mat-icon svgIcon="save"></mat-icon>
       <span>Save</span>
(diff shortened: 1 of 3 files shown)


libs/movie/src/lib/movie/+state/movie.model.ts
```diff
@@ -336,7 +336,7 @@ export function createMovieVideos(params: Partial<MovieVideos>): MovieVideos {
   return {
     ...params,
     screener: createMovieVideo(params?.screener),
-    otherVideos: params?.otherVideos?.map(video => createMovieVideo(video)) || [],
+    otherVideos: params?.otherVideos?.map?.(video => createMovieVideo(video)) || [],
```
fritzschoff, line 339: you check if this an array?

libs/media/src/lib/file/explorer/explorer.component.html
```diff
@@ -43,7 +43,10 @@
   <file-uploader [form]="dir.form" [meta]="dir.meta" [accept]="dir.accept" (selectionChange)="change($event, dir.meta)">
     <h3 title>Upload {{ dir.name }}</h3>
   </file-uploader>
-  <section fxLayout="row" fxLayoutAlign="end center" class="save">
+  <section fxLayout="row" fxLayoutAlign="space-between center" class="save">
+    <div>
```
fritzschoff, line 47: div because you need to put the modal as block element?

fritzschoff (approved):

libs/media/src/lib/file/explorer/explorer.component.html
pldespaigne: div because the parent has a flex `space-between` and the bf-modal an `ngIf` so the div is used a wrapper so that if the bf-modal does not appear the save button don't jump to the left

libs/movie/src/lib/movie/+state/movie.model.ts
pldespaigne: somehow the data in db has been corrupted for some movie, this is just a quick-fix/safe guard to avoid breaking the app for the users if they got one of the corrupted movies, I open this issue to find & patch the root cause #5494

Issue #5049, opened 2021-03-02
CRM invitation


Mussuranae:
We need to test and maybe think about a new way for inviting people to event and org from CRM.
Presently, we use the same component that the apps, but it's maybe not the best way as it takes the current app to register some data. Or on CRM we don't want to  register those datas with CRM.
Check Todo in the code

Issue #5505, opened 2021-04-07
3.0 Remaining bug list


Jackseed:
- [x] **Remco:** Org member invitations from CRM use "Blockframes CRM" as app.name variable in the invitation email. Instead it should be an app that the org has access to, if many take the first one.
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/50c0e766-2d5d-47d3-9840-59f9ec7b6320)
url: https://staging.crm.blockframes.io/c/o/admin/panel/organization/035zZAYcEXjSkcf8BmSi
user: dev+hettie-fvh@blockframes.io

- [x] **Remco:** Dashboard title list: sorting doesn't work on the "# Views" column

- [x] **Remco:** Still an image issue on catalog landing page for Safari
Brave
![image](https://user-images.githubusercontent.com/14964443/113825785-b1883080-9781-11eb-8e28-ed01c53e9087.png)
vs Safari
![image](https://user-images.githubusercontent.com/14964443/113825834-bd73f280-9781-11eb-9b4a-c1db8507a4a1.png)

- [x] Remove "Offers & Deals" from marketplace catalog sidenav 
![image](https://user-images.githubusercontent.com/14964443/113826014-f57b3580-9781-11eb-80a7-15a70a3c7665.png)

- [x] catalog - marketplace title list - Clear all filters button doesn't clear Avails filter
![image](https://user-images.githubusercontent.com/14964443/113827322-7555cf80-9783-11eb-8b4b-a59b5bb1e1ae.png)

PR #5506, opened 2021-04-07
5505 5049 replace crm by first accessible app

RemcoSimonides:
Fix issue #5049 && to do in #5505
- [x] Org member invitations from CRM use "Blockframes CRM" as app.name variable in the invitation email. Instead it should be an app that the org has access to, if many take the first one.
- [x] Dashboard title list: sorting doesn't work on the "# Views" colum

### DIFF
--- a/libs/invitation/src/lib/+state/invitation.service.ts
+++ b/libs/invitation/src/lib/+state/invitation.service.ts
@@ -10,7 +10,7 @@ import { Invitation, createInvitation } from './invitation.model';
 import { InvitationDocument } from './invitation.firestore';
 import { cleanInvitation } from '../invitation-utils';
 import { RouterQuery } from '@datorama/akita-ng-router-store';
-import { getCurrentApp } from '@blockframes/utils/apps';
+import { App, getCurrentApp } from '@blockframes/utils/apps';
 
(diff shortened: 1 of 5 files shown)


fritzschoff (approved):

libs/invitation/src/lib/+state/invitation.service.ts
```diff
@@ -104,9 +104,12 @@ export class InvitationService extends CollectionService<InvitationState> {
         const recipients = Array.isArray(idOrEmails) ? idOrEmails : [idOrEmails];
 
         const f = this.functions.httpsCallable('inviteUsers');
-        const app = getCurrentApp(this.routerQuery); // @TODO #5049 problem here if invite comes from CRM
+        let app = getCurrentApp(this.routerQuery);
+        if (app === 'crm') {
+          // Instead use first found app where org has access to
+          app = Object.keys(fromOrg.appAccess).find((app: App) => Object.keys(fromOrg.appAccess[app]).some(module => fromOrg.appAccess[app][module])) as App
+        }
```
Mussuranae, line 111: You have the `getOrgAppAccess` function for that in the lib/utils/apps.ts

PR #5509, opened 2021-04-07
5486 Improve performance and avoid flashes

RemcoSimonides:


libs/event/src/lib/components/screening-item/screening-item.component.html
```diff
@@ -1,12 +1,16 @@
-<a matRipple [routerLink]="['/c/o/marketplace/event/', screening.id]">
-  <img fxHide.lt-sm [ref]="poster" asset="empty_poster.webp" alt="Movie poster of the screening">
-</a>
+<ng-container *ngIf="screening.meta.titleId | getTitle | async as title">
+  <a matRipple [routerLink]="['/c/o/marketplace/event/', screening.id]">
+    <img fxHide.lt-sm [ref]="title.poster" asset="empty_poster.webp" alt="Movie poster of the screening">
+  </a>
+</ng-container>
```
RemcoSimonides, line 5: Wrapping the image with an `ngIf` results in image being rendered later. We're talking about milliseconds here, but it's not the prettiest.

GrandSchtroumpf: You only need the title on the img. So you can do something like that : 
```html
<a matRipple [routerLink]="['/c/o/marketplace/event/', screening.id]">
  <img fxHide.lt-sm
    [ref]="(screening.meta.titleId | getTitle | async).poster"
    asset="empty_poster.webp"
    width="WIDTH TO ADD" height="HEIGHT TO ADD"
    alt="Movie poster of the screening">
</a>
```
By specifying the width we expect, you let the browser keep this space until img is loaded.

libs/event/src/lib/components/screening-item/screening-item.component.html
```diff
@@ -1,12 +1,16 @@
...
 <article fxLayout="column" fxLayoutAlign="space-between start">
   <div fxLayout="column" fxLayoutAlign="start start">
     <h3><a [routerLink]="['/c/o/marketplace/event/', screening.id]">{{ screening.title }}</a></h3>
     <h5>{{ screening.isPrivate ? 'Private' : 'Public' }} Screening </h5>
     <pre class="mat-body-1">{{ screening | eventRange }}</pre>
-    <org-chip [org]="screening.org"></org-chip>
+    <ng-container *ngIf="screening.ownerOrgId | getOrg | async as org">
```
GrandSchtroumpf, line 11: You can do : 
```html
<org-chip [org]="screening.ownerOrgId | getOrg | async"></org-chip>
```

RemcoSimonides: Making it an observable will first emit `undefined` and org-chip doesn't like that. org-chip can be wrapped with the ng-container in my opinion

libs/media/src/lib/image/directives/image.directive.ts
```diff
@@ -102,27 +102,16 @@ export class ImageDirective implements OnInit, OnDestroy {
       map(([local, global]) => local || global)
     );
```
phpgeekfr, line 104: `@Input() delay = 5000;` Is now useless I guess

libs/organization/src/lib/organization/components/chip/chip.component.ts
```diff
@@ -1,13 +1,16 @@
-import { Component, OnInit, ChangeDetectionStrategy, Input } from '@angular/core';
+import { Component, ChangeDetectionStrategy, Input, HostBinding } from '@angular/core';
 import { Organization } from '@blockframes/organization/+state';
+import { fade } from '@blockframes/utils/animations/fade';
 
 @Component({
   selector: 'org-chip',
   templateUrl: './chip.component.html',
   styleUrls: ['./chip.component.scss'],
+  animations: [fade],
```
GrandSchtroumpf, line 9: It'll fade everywhere it's used. I think it's better to apply it only on the event list page to avoid side effect.

GrandSchtroumpf (approved):

Issue #5342, opened 2021-03-22
CRM - Movie list table - change original title with international title


Jackseed:
Users tend to fill in international title so that should be the one used for sorting

PR #5514, opened 2021-04-07
CRM movie-user

### LINKED ISSUE #4877: CRM - User page - Add invitation list(s)
For admin purposes it would help to have a list of all the pending invitations of a user (org invitations + event invitations). It should be on the user page. 

If that's easier to separate org invitations from event invitations and have 2 different tables/lists that's more than ok.

Mussuranae:
Change the original to international title on movie list and add invitation table on user profile
Fix #4877 & #5342


libs/admin/src/lib/admin-panel/pages/user/user.component.html
```diff
@@ -1,115 +1,130 @@
...
+      </mat-form-field>
+      <mat-card-actions align="end" fxLayoutGap="24px">
+        <button type="submit" mat-flat-button color="primary" (click)="update()">Update</button>
+        <button type="submit" mat-flat-button color="warn" (click)="deleteUser()">Delete</button>
+      </mat-card-actions>
+    </form>
+  </mat-card>
+</section>
+
+<section *ngIf="user">
```
phpgeekfr, line 107: if is about `user` but you don't use it after

phpgeekfr (approved):

libs/admin/src/lib/admin-panel/pages/user/user.component.html
Mussuranae: Yeah, if there is no user, there will be no invitation. So it was to let time to the component to load data. I can change it for if invitation and not user

PR #5523, opened 2021-04-07
Invitation email to event

### LINKED ISSUE #5519: Event invitation is not sent if an org invitation is pending
Classic flow for concierges is:
- inviting a new user to an org
- then inviting that user to the event
This way when the user activates her/his account, the event invitation is here. 

Now this flow doesn't work, the event invitation isn't sent anymore.

Mussuranae:
Fix the invitation email not sent when an user is invited to join org first.
Fix #5519

### DIFF
--- a/apps/backend-functions/src/internals/invitations/events.ts
+++ b/apps/backend-functions/src/internals/invitations/events.ts
@@ -6,6 +6,7 @@ import { getUser } from "../utils";
 import { createDocumentMeta, createPublicInvitationDocument, getAdminIds, getDocument } from "../../data/internals";
 import { EventDocument, EventMeta } from "@blockframes/event/+state/event.firestore";
 import * as admin from 'firebase-admin';
+import { hasUserAnOrgOrIsAlreadyInvited } from '../../invitation';
 
 /**
  * Handles notifications and emails when an invitation to an event is created.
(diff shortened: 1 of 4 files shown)


apps/backend-functions/src/internals/invitations/events.ts
```diff
@@ -32,14 +32,16 @@ async function onInvitationToAnEventCreate(invitation: InvitationDocument) {
...
   if (!!invitation.toUser) {
     /**
-     * @dev We wants to send this email only if user have an orgId. If not, this means that he already received an
+     * @dev We wants to send this email only if user have an orgId and a validated account. If not, this means that he already received an
      * email inviting him along with his credentials.
     */
     const user = await getDocument<PublicUser>(`users/${invitation.toUser.uid}`);
-    if (!user.orgId) {
+    const hasOrgOrOrgInvitation = await hasUserAnOrgOrIsAlreadyInvited([invitation.toUser.email]);
+    if (!user.orgId && !hasOrgOrOrgInvitation) {
```
phpgeekfr, line 40: I think you can use only `hasOrgOrOrgInvitation` because we already check in this method if user have an orgId

phpgeekfr (approved):

apps/backend-functions/src/internals/invitations/events.ts
Mussuranae: Yeah true, I change it

PR #5533, opened 2021-04-08
removed form country component and implement static select component

### LINKED ISSUE #5239: remove `form-country` component and use `static-select` instead
We should get rid of `FormCountryComponent` component and use the `static-select` instead. This will reduce the amount of code to maintain.

see https://github.com/blockframes/blockframes/pull/4958#discussion_r595064065

fritzschoff:
close #5239


GrandSchtroumpf (approved):
❤️

PR #5537, opened 2021-04-08
order

### LINKED ISSUE #5357: File explorer - sorting
Comments from @maltmayer - 2.3 Release reviews (#5261)
- [ ] the files in each movie are ordered alphabetically. they should not, it does not make sense

I think it should be in more logical order:
- Poster
- Banner
- Images
- Screener
- Sales Pitch
- Other Videos
- Scenario
- Presentation Deck
- Moodboard / Artistic Deck
- Notes & Statements

Comment from @2amarkina 
>  i would propose to put the folders first and then put the single documents in alphabetical order OR put everything as they appear on the movie tunnel

Comment from @pldespaigne 
> having a predefined/hardcoded order will be really hard or ugly

 But on the contrary, the movies should be ordered alphabetically in the File storage. We can't guess the logic here: 

PL:
> PL: It's currently ordered by ID, probably possible with a sort function or an order clause in the query, this might have side effects though

pldespaigne:
fix #5357

### DIFF
--- a/libs/media/src/lib/file/explorer/explorer.component.ts
+++ b/libs/media/src/lib/file/explorer/explorer.component.ts
@@ -1,7 +1,7 @@
 import { ChangeDetectionStrategy, Component, ViewChild, TemplateRef, Pipe, PipeTransform, Input, AfterViewInit, OnInit } from '@angular/core';
 
 // Blockframes
-import { MovieService } from '@blockframes/movie/+state';
+import { Movie, MovieService } from '@blockframes/movie/+state';
 import { RouterQuery } from '@datorama/akita-ng-router-store';
 import { App } from '@blockframes/utils/apps';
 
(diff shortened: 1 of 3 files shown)


libs/media/src/lib/file/explorer/explorer.component.ts
```diff
@@ -85,6 +87,10 @@ export class FileExplorerComponent implements OnInit, AfterViewInit {
       this.org$.asObservable(),
       this.movieService.valueChanges(query)
     ]).pipe(
+      map(([org, titles]) => {
+        const sortedTitles = titles.sort((movieA, movieB) => movieA.title.international < movieB.title.international ? -1 : 1);
```
GrandSchtroumpf, line 91: Move the sort on `this.movieService.valueChanges(query)`
```typescript
const titles$ = this.movieService.valueChanges(query).pipe(
  map(titles => titles.sort((a, a) => a.title.international < b.title.international ? -1 : 1)
);
this.root$ = combineLatest([this.org$, titles$]).pipe(
  map(([org, titles]) => getDirectories(org, titles)),
)
```

GrandSchtroumpf (approved):

PR #5544, opened 2021-04-09
cart cleaning

phpgeekfr:
remove unused code


libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -147,4 +149,32 @@ export class OrganizationService extends CollectionService<OrganizationState> {
...
+      });
+    } else {
+      wishlist.push(movie.id);
+      this.analytics.event('addedToWishlist', {
+        movieId: movie.id,
+        movieTitle: movie.title.original
+      });
+    }
+
+    this.update({ ...orgState, wishlist });
```
GrandSchtroumpf, line 178: You should be able to update only the wishlist. remove ...orgState.

GrandSchtroumpf (approved):

libs/organization/src/lib/organization/+state/organization.service.ts
phpgeekfr: updated it to `this.update(orgState.id, { wishlist });`

PR #5545, opened 2021-04-09
Fix compile error in tests

xmano:
Hot fix for resolving test compile errors.

Need to add proper fix subsequently if it was meant to make new test pass @fritzschoff 

image


fritzschoff (approved):
Lgtm

PR #5565, opened 2021-04-16
add exclude regex to gsutil transfer function helper

george43g:
as per title
we add a string regex option (will always BE python regex)
we exclude mov and mp4 files under protected dir when generating golden data

voila

### DIFF
--- a/libs/firebase-utils/src/lib/restore-storage-from-ci.ts
+++ b/libs/firebase-utils/src/lib/restore-storage-from-ci.ts
@@ -21,11 +21,12 @@ export async function restoreStorageFromCi(ciApp: admin.app.App) {
     console.log('Latest backup:', folderName);
 
     console.log("Mirroring storage bucket from blockframe-ci to your local project's storage bucket...");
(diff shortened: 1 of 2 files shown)


libs/firebase-utils/src/lib/restore-storage-from-ci.ts
```diff
@@ -25,6 +25,7 @@ export async function restoreStorageFromCi(ciApp: admin.app.App) {
       quiet: true,
       mirror: true,
       from: `gs://${CI_STORAGE_BACKUP}/${folderName}`,
+      exclude: 'protected.*\.mp4$|protected.*\.mov$',
```
phpgeekfr, line 28: can you make the regex support the file types listed in storage.rules ?
```
       request.resource.contentType.matches('video/x-msvideo') ||
          request.resource.contentType.matches('video/x-matroska') ||
          request.resource.contentType.matches('video/mp4') ||
          request.resource.contentType.matches('video/3gpp') ||
          request.resource.contentType.matches('video/quicktime') ||
          request.resource.contentType.matches('video/x-ms-wmv')
```

phpgeekfr (approved):

Issue #5184, opened 2021-03-11
Change release media


Jackseed:
url: /c/o/dashboard/tunnel/movie/movieId/additional-information (production status to "released")

Currently we display 3 media for release:
- Theatrical
- Video
- Hotels
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/ef12c789-1726-4054-a3d3-dbee95bf1d1b)

This list should be changed for:
- Theatrical
- Video
- TV

Since we remove "Hotels", db should be checked before changing that to look for a movie that has hotel as release media. I don't think there is one so it should be fine, but better safe than sorry

RemcoSimonides:
Hee @Jackseed ,
No matter whether there is 'Hotels' in the database, I think it's still best to create a migration for it. Just to be save ;) But what value would you change it to? video?

Also, this is the list we have to choose from
```
  payTv: 'Pay TV', // pay-tv
  freeTv: 'Free TV', // free-tv
  payPerView: 'Pay Per View', // pay-per-view
  est: 'EST',
  nVod: 'N-VOD', // n-vod
  aVod: 'A-VOD', // a-vod
  fVod: 'F-VOD', // f-vod
  sVod: 'S-VOD', // s-vod
  theatrical: 'Theatrical',
  video: 'Video',
  planes: 'Planes',
  boats: 'Boats',
  hotels: 'Hotels',
  educational: 'Educational',
  rental: 'Rental',
  through: 'Sell Through'
  ```
  which doesn't contain a 'TV'; only Pay TV and Free TV. What should I do?

Jackseed:
@RemcoSimonides ok then you can use theatrical, video, payTv and freeTv

PR #5577, opened 2021-04-20
5184 change originalRelease hotels for video

RemcoSimonides:
Fix issue #5184

### DIFF
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0059.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0059.ts
@@ -10,12 +10,16 @@ export async function upgrade(db: Firestore) {
   */
   return runChunks(movies.docs, async (movieDoc) => {
     const movie = movieDoc.data();
+    let update = false;
     for (const release of movie.originalRelease) {
       if (release.media === 'hotels') {
         release.media = 'video';
+        update = true;
       }
     }
   
-    await movieDoc.ref.set(movie);
+    if (update) {
+      await movieDoc.ref.set(movie);
+    }
   });
 }
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0059.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0059.ts
@@ -1,0 +1,21 @@
+import { Firestore } from '@blockframes/firebase-utils';
+import { runChunks } from '../firebase-utils';
+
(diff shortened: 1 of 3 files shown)


libs/firebase-utils/src/lib/firestoreMigrations/0059.ts
```diff
@@ -0,0 +1,25 @@
...
+  const movies = await db.collection('movies').get();
+
+  /*
+    Replace originalRelease media 'hotels' with 'video'
+  */
+  return runChunks(movies.docs, async (movieDoc) => {
+    const movie = movieDoc.data();
+    let update = false;
+    for (const release of movie.originalRelease) {
+      if (release.media === 'hotels') {
```
RemcoSimonides, line 15: Chance exists that no movie uses this value so I only update the movie if it's needed

libs/firebase-utils/src/lib/firestoreMigrations/0059.ts
```diff
@@ -0,0 +1,25 @@
...
+        release.media = 'video';
+        update = true;
+      }
+    }
+  
+    if (update) {
```
phpgeekfr, line 21: (since we are in a migration, we can afford to update every docs, even not touched)

libs/movie/src/lib/movie/form/additional-information/additional-information.component.html
```diff
@@ -16,7 +16,7 @@ <h3>Release</h3>
                 </static-select>
 
                 <static-select test-id="release-media" [control]="releaseForm.get('media')" scope="medias" fxFlex
-                  [withoutValues]="['payTv', 'freeTv', 'payPerView', 'est', 'nVod','aVod', 'fVod', 'sVod', 'planes', 'boats', 'educational', 'rental','through']">
+                  [withoutValues]="['payPerView', 'est', 'nVod','aVod', 'fVod', 'sVod', 'planes', 'boats', 'educational', 'rental','through', 'hotels']">
```
phpgeekfr, line 19: is `hotels` still used elsewere ? or can we remove it ?

phpgeekfr (approved):

libs/movie/src/lib/movie/form/additional-information/additional-information.component.html
RemcoSimonides: It is used ;)

PR #5583, opened 2021-04-20
mirrored local video

### LINKED ISSUE #5503: Vertical symmetry for video conference on meetings
Standards like Zoom, Gmeet,.. mirror the video with a vertical symmetry. We should do the same to keep up with good practice if that's not too big an issue.

pldespaigne:
fix #5503

### DIFF
--- a/libs/event/src/lib/components/meeting/video/video.component.html
+++ b/libs/event/src/lib/components/meeting/video/video.component.html
@@ -1,5 +1,5 @@
 
-<video #video [ngClass]="{ hidden: !(remoteTrackState$.video | async) || !(localTrackState$.video | async) }"></video>
+<video #video [ngClass]="{ hidden: !(remoteTrackState$.video | async) || !(localTrackState$.video | async), mirrored: local }"></video>
 <audio #audio></audio>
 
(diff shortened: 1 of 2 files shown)


libs/event/src/lib/components/meeting/video/video.component.html
```diff
@@ -1,5 +1,5 @@
 
-<video #video [ngClass]="{ hidden: !(remoteTrackState$.video | async) || !(localTrackState$.video | async) }"></video>
+<video #video [ngClass]="{ hidden: !(remoteTrackState$.video | async) || !(localTrackState$.video | async), mirrored: local }"></video>
```
phpgeekfr, line 2: what is this about ?

phpgeekfr (approved):

libs/event/src/lib/components/meeting/video/video.component.html
pldespaigne: It's about #5503 : we mirror the video of the local user.
It's a UX good practice _(used by Google & Zoom)_ so that if you raise your right hand, the raised hand on the video will also be on the right side of the screen.

So this line of code ☝️  just apply the class `mirrored` if `this.local = true`

phpgeekfr: ok simple ^^ I was first thinking of a mirrored server to download videos

Issue #5467, opened 2021-04-02
Email validation after creating an account should be mandatory to access app(s)


phpgeekfr:
Email validation after creating an account should be mandatory to access app(s).

Suggestions:
- [x] create a guard to redirect users without validated email to the "send validation email" page
- [x] display this guard only to new created users, not existing one
- [ ] add a information popin in users settings page inviting existing users to validate their email

Jackseed:
@phpgeekfr 
> - create a guard to redirect users without validated email to the "send validation email" page

Is it that page you're talking about? 
![image](https://user-images.githubusercontent.com/14964443/113549522-e2d1f680-95f1-11eb-9130-c0b895f07a52.png)

> - display this guard only to new created users, not existing one
👍 

> - add a information popin in users settings page inviting existing users to validate their email
do we have a way to know how many existing users didn't validate their email address?

RemcoSimonides:
When user wants to validate their email again, they should have a button to re-send a verification email.
For this; search for `@TODO (#2821)` for this

RemcoSimonides:
> do we have a way to know how many existing users didn't validate their email address?
I don't think we have an easy way, but maybe you know @phpgeekfr ?

phpgeekfr:
we could add this information in CRM. Or if is only for a one shot: `firebase auth:export /home/bruce/Bureau/cascade8/ninja/auth/auth-prod-$(date +%d-%m-%Y).json --format=json`
Info will be in json

phpgeekfr:
for auth-prod-07-04-2021:

"emailVerified": false => 1398
 "emailVerified": true, => 2270

Jackseed:
Thanks @phpgeekfr for the figures, we can't block that many users with a guard - especially if they have a wrong email address in the first place. 
For now I would go with your recommendation to propose to users to validate it. 

Could we add a CTA button "Verify Email Address" on the right of the email address input that is visible only for users that didn't verify their email?
Also a snackbar should be added when button clicked with "Validation email sent." for 4s and no action.

![image](https://user-images.githubusercontent.com/14964443/116096186-5cb54700-a6a9-11eb-8775-d63b23c22321.png)

RemcoSimonides:
Do we have a different email template? Because this one contains 'Thanks for signing up to Archipel! '. Or is it ok?
![image](https://user-images.githubusercontent.com/27687382/116434011-53171500-a84a-11eb-8ca5-d7d74f407352.png)

[...]

PR #5596, opened 2021-04-21
5467 enforce email verification

RemcoSimonides:
Fix issue #5467


libs/user/src/lib/auth/guard/email-verified.guard.ts
```diff
@@ -0,0 +1,26 @@
...
+})
+export class EmailVerifiedGuard implements CanActivate  {
+  constructor(private authQuery: AuthQuery, private router: Router) {}
+
+  canActivate() {
+    return this.authQuery.select().pipe(
+      map(user => {
+        // Before we forced users to verify email, the user didnt have a createdAt value yet.
+        // Therefore we still give access to these old users by checking if they have this value.
+        if (!!user.profile._meta?.createdAt && !user.emailVerified) {
```
phpgeekfr, line 17: What I had in mind was to check that `createdAt` date was after date that **we introduced this guard** to affect only newly created users. Like this only newly created users will be forced to verify email.

For older customers, I would have added a message in `c/o/account/profile/view/settings` to inform them that they can improve account security by verifying there email address.

This is maybe a question to ask to business team.

phpgeekfr: So, if we only force new users to verify email and if we don't let them access the app until they do (because of IdentityGuard) do we need this one  ?

RemcoSimonides: yea, without this guard you can navigate to any page beyond `/c/o`...

phpgeekfr (approved):

libs/user/src/lib/auth/guard/identity.guard.ts
```diff
@@ -35,6 +35,9 @@ export class IdentityGuard extends CollectionGuard<AuthState> {
           map(async user => {
             if (!hasDisplayName(user)) { return true; }
             if (!!user.orgId) {
+              const { emailVerified } = await this.service.user;
+              if (!emailVerified) return this.router.navigate(['c/organization/join-congratulations']);
```
phpgeekfr, line 39: this needs to be inside `if (!!user.orgId) {` ?  No above ?

Mussuranae (approved):

Issue #5340, opened 2021-03-22
Change Org invitation email template-ids


Jackseed:
We currently use 3 different emails when an org invite a member, one per app:
- AC template id: `d-a34ce9ea59c5477f9feae8f556157b6b`
- AM template id: `d-f0c4f1b2582a4fc6ab12fcd2d7c02f5c` 
- MF template id: `d-a34ce9ea59c5477f9feae8f556157b6b`

Now everything related to apps have been changed to variables so we can switch to only one template used for every app. 

The correct id is the one currently used for AM:  `d-f0c4f1b2582a4fc6ab12fcd2d7c02f5c`

RemcoSimonides:
@Jackseed do you want me to delete or deactive the templates in SendGrid too?

PR #5597, opened 2021-04-21
5340 remove app specific template ids

RemcoSimonides:
Fix issue #5340 
catalog and financiers weren't even used

### DIFF
--- a/apps/backend-functions/src/templates/mail.ts
+++ b/apps/backend-functions/src/templates/mail.ts
@@ -69,7 +69,7 @@ export function userInvite(
   password: string,
   orgName: string,
   pageURL: string = appUrl.market,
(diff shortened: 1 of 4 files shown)


apps/backend-functions/src/templates/mail.ts
```diff
@@ -69,7 +69,7 @@ export function userInvite(
   password: string,
   orgName: string,
   pageURL: string = appUrl.market,
-  templateId: string = templateIds.user.credentials.joinOrganization.festival,
+  templateId: string = templateIds.user.credentials.joinOrganization,
```
Mussuranae, line 72: Did you check for this function if we pass a template sometimes ?

Mussuranae (approved):

PR #5600, opened 2021-04-22
implementing new link animation

fritzschoff:


libs/user/src/lib/auth/pages/identity/identity.component.html
```diff
@@ -88,15 +88,15 @@ <h1>Create your account</h1>
 
         <div class="privacy" fxLayout="column" fxLayoutAlign="start">
           <auth-accept-conditions [control]="form.get('termsOfUse')">
-            <span conditionText>I hereby accept {{ appName }}</span>
-            <span conditionLink>Terms and Conditions</span>
+            <span conditionText>I hereby accept {{ appName }}&nbsp;</span>
+            <a conditionLink>Terms and Conditions</a>
```
GrandSchtroumpf, line 92: You shouldn't have a link without href/routerLink. If you want to apply the style only, create a class "link" that have the same style as the links.

GrandSchtroumpf (approved):

Issue #5384, opened 2021-03-25
Improve staging & prod logs


Jackseed:
Currently we use Sentry for staging & prod logs but we should at least keep error logs in the console as well to facilitate debugs. 

Re backend there is another existing issue #5011 

@pldespaigne @phpgeekfr I let you give more details there as you experimented a lot this issue during the release process

phpgeekfr:
@Jackseed Is this issue about front end ? You want to display *front* errors into users console ?   Or are you talking about backend only ? In that case, this is a duplicate of #5011

phpgeekfr:
This make sense for staging or other testing env where sentry is enabled but not for production I guess

PR #5601, opened 2021-04-22
fix #5011 and #5384

### LINKED ISSUE #5011: Remove sentry from backend functions
Sentry is used in the backend but no data are actually sent to sentry dashboard.
That means that error log are lost.

- [ ] remove sentry from `backend-functions`
- [ ] check that functions exceptions are logged correctly in firebase loggs

phpgeekfr:
- Errors in backend will now be catched by sentry but still logged in firebase functions log.
- Errors in frontend are still pushed to sentry if `sentryDsn` is enabled but if environment is not production, errors will still be logged into browser console for debugging


pldespaigne (approved):
Awesome 🙌

Issue #5356, opened 2021-03-23
Notifications - improvements


Jackseed:
- [x] New texts for notifications have been updated here https://docs.google.com/spreadsheets/d/1b1K7YfnquxnfCgGGPxIz5ZNIjr9eK_bUrRJqLvPYGIk/edit#gid=1590038159
It concerns only lines in orange

- [x] `d-f84d8c5a70884316870ca4ef657e368f` is spawned for every new member in the org but in fact it would be better if it was used only when a user accepts an invitation to join an org.
Cause here it's sent also for requests being accepted but we already have 3 mails related to requests, it's a bit too much

- [x] Small improvement for Event request refused `d-4e36dd72f50d4e7e9ec738476efa84a9`, would be good to add a `pageUrl` that leads to the marketplace org view (/c/o/marketplace/organization/orgId/title) of the organizer org

- [x] We need a more accurate variable for app access email, it shall give us the new app name that has been given during the update. Send email for every app the org gained access to. Also, update notification text. (issue #5471)

- [x] Don't send email if the app access are removed and not added

- [x] Add app name and organization name in email the to cascade8 admin when a user/org requests access to an app

RemcoSimonides:
> - [ ] d-f84d8c5a70884316870ca4ef657e368f is spawned for every new member in the org but in fact it would be better if it was used only when a user accepts an invitation to join an org.
Cause here it's sent also for requests being accepted but we already have 3 mails related to requests, it's a bit too much

This one is actually quite difficult because the notification is triggered when length of array of `userIds` is changed on org update. How important is it to remove this notification? 😅

PR #5604, opened 2021-04-22
5356 notifications improvements

RemcoSimonides:
Some to do's on issue #5356
- [x] d-f84d8c5a70884316870ca4ef657e368f is spawned for every new member in the org but in fact it would be better if it was used only when a user accepts an invitation to join an org.
Cause here it's sent also for requests being accepted but we already have 3 mails related to requests, it's a bit too much
- [x] Small improvement for Event request refused d-4e36dd72f50d4e7e9ec738476efa84a9, would be good to add a pageUrl that leads to the marketplace org view (/c/o/marketplace/organization/orgId/title) of the organizer org
- [x] New texts for notifications (2/3)

### DIFF
--- a/apps/backend-functions/src/orgs.ts
+++ b/apps/backend-functions/src/orgs.ts
@@ -55,8 +55,16 @@ async function notifyOnOrgMemberChanges(before: OrganizationDocument, after: Org
     const userSnapshot = await db.doc(`users/${userAddedId}`).get();
     const userAdded = userSnapshot.data() as PublicUser;
 
-    const notifications = after.userIds.filter(userId => userId !== userAdded.uid).map(userId => notifyUser(userId, 'orgMemberUpdated', after, userAdded));
-    return triggerNotifications(notifications);
+    // Only send notification if invitations is invitation and not a request
+    const invitationsSnapshot = await db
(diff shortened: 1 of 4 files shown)


apps/backend-functions/src/orgs.ts
```diff
@@ -55,8 +55,16 @@ async function notifyOnOrgMemberChanges(before: OrganizationDocument, after: Org
     const userSnapshot = await db.doc(`users/${userAddedId}`).get();
     const userAdded = userSnapshot.data() as PublicUser;
 
-    const notifications = after.userIds.filter(userId => userId !== userAdded.uid).map(userId => notifyUser(userId, 'orgMemberUpdated', after, userAdded));
-    return triggerNotifications(notifications);
+    // Only send notification if invitations is invitation and not a request
```
Mussuranae, line 58: for this to work, don't you have to check for the mode `invitation` and not `request` ?

Mussuranae: Nevermind you can do the trick with `toUser` and `fromOrg`.

Mussuranae (approved):
LGTM

PR #5605, opened 2021-04-22
Feat 5156

### LINKED ISSUE #5156: Create org ids for apps
Currently we've got a "centralOrgID" in env.
We should be able to create orgId per app: 
```typescript
export const centralOrgId = {
  cascade8: 'jnbHKBP5YLvRQGcyQ8In',
  catalog: 'nBM8sy5EEAOtVBCX4nTJ', // This has been created in prod
};
```
The id of catalog should be available through PFT starting on the 15th of March

- [x] Update algolia search and analytics to iterate on every org.
- [x] Update excel import & createOffer to use catalog id instead

phpgeekfr:
fix #5156

### DIFF
--- a/apps/backend-functions/src/movie.ts
+++ b/apps/backend-functions/src/movie.ts
@@ -4,7 +4,7 @@ import { triggerNotifications, createNotification } from './notification';
 import { createDocumentMeta, getDocument, getOrganizationsOfMovie } from './data/internals';
 import { removeAllSubcollections } from './utils';
 
-import { centralOrgID } from './environments/environment';
+import { centralOrgId } from './environments/environment';
 import { orgName } from '@blockframes/organization/+state/organization.firestore';
 import { cleanMovieMedias } from './media';
 import { Change, EventContext } from 'firebase-functions';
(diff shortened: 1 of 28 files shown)


apps/backend-functions/src/movie.ts
```diff
@@ -104,7 +104,7 @@ export async function onMovieUpdate(
   const appAccess = apps.filter(a => !!after.storeConfig.appAccess[a]);
 
   if (isMovieSubmitted) { // When movie is submitted to Archipel Content
-    const archipelContent = await getDocument<OrganizationDocument>(`orgs/${centralOrgID}`);
+    const archipelContent = await getDocument<OrganizationDocument>(`orgs/${centralOrgId.catalog}`);
```
phpgeekfr, line 107: I think this should now be `catalog` but I would like a confirmation for this please

apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
```diff
@@ -122,7 +122,7 @@ export class ListComponent implements OnInit, OnDestroy {
 
   private async getContract(type: Contract['type']) {
     const contracts = type === 'mandate'
-      ? await this.contractService.getValue(ref => ref.where('type', '==', 'mandate').where('buyerId', '==', centralOrgID).where('status', '==', 'accepted'))
+      ? await this.contractService.getValue(ref => ref.where('type', '==', 'mandate').where('buyerId', '==', centralOrgId.catalog).where('status', '==', 'accepted'))
```
phpgeekfr, line 125: I think this should now be `catalog` but I would like a confirmation for this please

apps/backend-functions/src/movie.ts
GrandSchtroumpf: I suppose yes

GrandSchtroumpf (approved):

PR #5606, opened 2021-04-22
update selection page

### LINKED ISSUE #5526: remove use of debounceFactory in MarketplaceSelectionComponent
remove use of debounceFactory in MarketplaceSelectionComponent

`debouncedUpdatePriceControl = debounceFactory((index, price) => this.updatePrice(index, price), 1000);`

fritzschoff:
close #5526


libs/utils/src/lib/pipes/reduce.pipe.ts
```diff
@@ -0,0 +1,18 @@
+import { NgModule } from '@angular/core';
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'reduce',
+  pure: false
```
GrandSchtroumpf, line 6: Why isn't it pure ?

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.ts
```diff
@@ -56,9 +58,13 @@ export class MarketplaceSelectionComponent {
     this.bucketService.update(id, { currency });
   }
 
-  updatePrice(index: number, price: string) {
+  cachePrice(index: number, price: string) {
+    this.cachedPrice[index] = +price
```
GrandSchtroumpf, line 62: Prefere `parseFloat(price)`

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.html
```diff
@@ -44,8 +45,8 @@ <h3>{{ title.directors | displayName }}</h3>
             </section>
             <mat-form-field appearance="outline">
               <mat-label>Price</mat-label>
-              <input matInput type="number" min="0" [value]="contract.price"
-                (input)="debouncedUpdatePriceControl(i, $event.target.value)" [attr.test-id]="'price-' + i"/>
+              <input matInput type="number" min="0" [value]="contract.price" [attr.test-id]="'price-' + i"
+              (input)="cachePrice(i, $event.target.value)" (blur)="updatePrice(i, $event.target.value)" />
```
GrandSchtroumpf, line 49: tbh, I would just get rid of the (input) event altogether.

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.html
```diff
@@ -5,7 +5,7 @@
       <div class="main-title" fxLayout="column" fxLayoutAlign="center center">
         <h1>My Selection</h1>
         <p>There are {{ bucket.contracts.length }} titles in my selection</p>
-        <p>Total price: {{ bucket | bucketTotalPrice | currency:bucket.currency }}</p>
+        <p>Total price: {{ cachedPrice | reduce | currency:bucket.currency }}</p>
```
GrandSchtroumpf, line 8: What's wrong wiht bucketTotalPrice ? You removed it from here but you didn't remove the pipe class, is it because it's used somewhere else ?

libs/utils/src/lib/pipes/reduce.pipe.ts
```diff
@@ -0,0 +1,18 @@
+import { NgModule } from '@angular/core';
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'reduce',
```
GrandSchtroumpf, line 5: I think it should be call `total` or `sum`

libs/utils/src/lib/pipes/reduce.pipe.ts
fritzschoff: cause the data can change and it needs to be reflected in the ui and need to trigger a rerender

GrandSchtroumpf: pure means that the output won't change if the input doesn't change. For example : 
```typescript
@Pipe({ name: 'interval', pure: false })
class Interval {
  transform(time: number) {
    const cb = () => Math.random();
    setInterval(cb, time);
    return cb;
  }
}
```

GrandSchtroumpf: In your case if you're not mutating the array input (which you shouldn't do) the pipe should be pure.

GrandSchtroumpf (approved):

PR #5607, opened 2021-04-22
5356 notification improvements

RemcoSimonides:
To do in issue #5356
- [x] Don't send email if the app access are removed and not added
- [x] We need a more accurate variable for app access email, it shall give us the new app name that has been given during the update. Send email for every app the org gained access to. Also, update notification text.

### DIFF
--- a/libs/notification/src/lib/+state/notification.firestore.ts
+++ b/libs/notification/src/lib/+state/notification.firestore.ts
@@ -41,6 +41,7 @@ export const notificationTypesPlus = [
   'movieSubmitted', // (catalog only)
   'organizationAcceptedByArchipelContent',
   'orgAppAccessChanged',
+  'orgAppAccessRemoved'
 ] as const;
 
 export type NotificationTypesBase = typeof notificationTypesBase[number];
--- a/libs/notification/src/lib/+state/notification.firestore.ts
+++ b/libs/notification/src/lib/+state/notification.firestore.ts
@@ -6,6 +6,7 @@ import { firestore } from 'firebase-admin';
 import { DocumentMeta } from '@blockframes/utils/models-meta';
 import { EmailErrorCodes } from '@blockframes/utils/emails/utils';
 import { Bucket } from '@blockframes/contract/bucket/+state/bucket.model';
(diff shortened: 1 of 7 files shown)


libs/notification/src/lib/+state/notification.store.ts
```diff
@@ -126,6 +126,7 @@ export class NotificationStore extends EntityStore<NotificationState, Notificati
           url: `/c/o/dashboard/title/${notification.docId}/main`,
         };
       case 'orgAppAccessChanged':
+      case 'orgAppAccessRemoved':
         // @TODO #4046 Update text if needed
```
Mussuranae, line 130: Can you delete this todo please ?

Mussuranae (approved):

libs/notification/src/lib/+state/notification.store.ts
```diff
@@ -127,14 +127,16 @@ export class NotificationStore extends EntityStore<NotificationState, Notificati
           url: `/c/o/dashboard/title/${notification.docId}/main`,
         };
       case 'orgAppAccessChanged':
-        // @TODO #4046 Update text if needed
+        const message = !!notification.appAccess
+          ? `Your organization has now access to ${notification.appAccess}`
```
Mussuranae, line 131: Use the `getAppName` function or directly the const like `appName[app]` (check `appName` in the code)

libs/notification/src/lib/+state/notification.firestore.ts
```diff
@@ -60,6 +61,7 @@ export interface NotificationBase<D> {
   organization?: PublicOrganization;
   invitation?: PublicInvitation;
   bucket?: Bucket;
+  appAccess?: App;
```
Mussuranae, line 64: Just call it `app`, it can help for some other notification in the future and it wont be access automatically..

Mussuranae (approved):

libs/notification/src/lib/+state/notification.firestore.ts
RemcoSimonides: App is already taken 😥

PR #5639, opened 2021-04-27
clean maintenance mode delay

phpgeekfr:
clean maintenance mode delay


apps/backend-ops/src/main.ts
```diff
@@ -27,7 +26,7 @@ async function runCommand() {
     case 'prepareForTestingBeta':
       await startMaintenance(db);
       await prepareForTestingBeta();
-      await endMaintenance(db, EIGHT_MINUTES_IN_MS);
+      await endMaintenance(db);
```
phpgeekfr, line 29: @george43g Do you remember why we added `EIGHT_MINUTES_IN_MS` here instead of removing this delay from front ? 

I thing it very confusing to have an ended date set in the past

george43g: because there is a delay, we need it to ensure it ends before we start e2e tests otherwise the tests will. see maintenance page on front endRather than the site. this is because frontendStill checks 8Min past endMaintenance

george43g: butMaybe this has changed

Issue #5169, opened 2021-03-11
Marketplace Screening view - Change "Access Screening Room" button visibility


Jackseed:
### Context
url: c/o/marketplace/event/eventId (screening)

Currently "Access Screening Room" CTA appears only when the user is accepted to the event and the event has started. Thus users don't know where to go when the event start since they don't know that button. 

invited & started event:
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/1301bd51-bcb1-431c-8c36-627be42a1ba7)
not invited: 
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/9ca3a6c7-959b-4fdc-9191-8148eae53ab8)

Instead we'd like to make that button appears when users are invited even when the event hasn't started yet. Though, the button should be disabled as long as event hasn't started yet with a tooltip saying "You'll be able to join screening when it starts". 

Also, one more use case is public events: it should then be displayed all the time but disable with the tooltip as long as the event hasn't started. 

### Acceptance criteria
- [ ] "Access Screening Room" CTA is visible when:
- event (public & private) is not finished
**&&**
- event is public
**or**
- event is private & user is invited

- [ ] "Access Screening Room" CTA is disabled + tooltip when:
- event hasn't started

RemcoSimonides:
In my opinion the disabled button looks like a normal button in this context.
![image](https://user-images.githubusercontent.com/27687382/116193360-82852f00-a72f-11eb-9c3e-85246abead60.png)

Therefore I propose this text instead:
![image](https://user-images.githubusercontent.com/27687382/116193299-6da89b80-a72f-11eb-98ba-b2115fa4a1e9.png)

Also, what text do you want to have on the tooltip? Something like
"You can access once the event has started."?

Jackseed:
Looks ok to me. Tooltip text is: "You'll be able to join screening when it starts"

@maltmayer @2amarkina could you confirm it's what you wanted?

maltmayer:
> Looks ok to me. Tooltip text is: "You'll be able to join screening when it starts"
> 
> @maltmayer @2amarkina could you confirm it's what you wanted?

"You'll be able to join the screening when it starts." (with a the)

Jackseed:
![image](https://user-images.githubusercontent.com/14964443/120656061-908f4380-c483-11eb-9d9c-abaa953300e8.png)
lgtm 👍  > closed

PR #5663, opened 2021-04-29
5169 Access screening room button before start

RemcoSimonides:
Fix issue #5169

### DIFF
--- a/libs/event/src/lib/layout/view/view.component.ts
+++ b/libs/event/src/lib/layout/view/view.component.ts
@@ -44,7 +44,7 @@ export class EventViewComponent implements OnInit, OnDestroy {
       this.event$.pipe(filter(event => !!event)),
       this.invitationService.guestInvitations$
     ]).pipe(
-      map(([event, invitations]) => invitations.find(invitation => invitation.eventId === event.id))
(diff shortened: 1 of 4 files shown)


libs/event/src/lib/layout/view/view.component.ts
```diff
@@ -44,7 +44,7 @@ export class EventViewComponent implements OnInit, OnDestroy {
       this.event$.pipe(filter(event => !!event)),
       this.invitationService.guestInvitations$
     ]).pipe(
-      map(([event, invitations]) => invitations.find(invitation => invitation.eventId === event.id))
+      map(([event, invitations]) => invitations.find(invitation => invitation.eventId === event.id) ?? null)
```
pldespaigne, line 47: `find()` already return `undefined` if nothing is found, did you really need to return `null` instead ?

pldespaigne (approved):

libs/event/src/lib/layout/view/view.component.ts
RemcoSimonides: ye, we have another state here. undefined, null and defined. This fix should have been in an earlier PR already.

Issue #5626, opened 2021-04-27
User: CRUD


Jackseed:
Tester: Samir
Debugger: Max

Jackseed:
- [ ] There shouldn't be a space before the question mark 
![Screenshot 2021-04-28 at 10.08.21.png](https://images.zenhubusercontent.com/603cae1b5c058cfd0a584f5f/a0fd67bb-0ffe-42fa-97cf-8f1a10fec259)

- [ ] There shouldn't be a space before the question mark 
![Screenshot 2021-04-28 at 10.10.55.png](https://images.zenhubusercontent.com/603cae1b5c058cfd0a584f5f/a20b644d-439e-4edf-b0e6-40200c3c0cec)

- [ ] Text of selected should be in contrast colour (white for light theme)
![Screenshot 2021-04-28 at 10.14.58.png](https://images.zenhubusercontent.com/603cae1b5c058cfd0a584f5f/e20417fe-4589-40c0-8f61-14f639f7dcc8)

### Sign up confirmation page ###
- [ ] Underline should be the same colour as text
![Screenshot 2021-04-28 at 10.21.46.png](https://images.zenhubusercontent.com/603cae1b5c058cfd0a584f5f/37278667-594f-4dd9-86cf-640614570bb7)

- [ ] There shouldn't be a white background on this icon
![Screenshot 2021-04-28 at 10.22.40.png](https://images.zenhubusercontent.com/603cae1b5c058cfd0a584f5f/9317887c-0487-4fc9-87ef-f7cbbbf9a6d1)

- [ ] Add a tooltip on the refresh button: "Refresh page."
![A.gif](https://images.zenhubusercontent.com/603cae1b5c058cfd0a584f5f/b162325d-46d9-49f5-9f55-07a739cb4d6b)

Jackseed:
- [ ] The email validation guard is blocking old users!! It shouldn't as we have no way to help them for now. This guard should only block **new** user while they haven't verify their email address
linked to #5647

Jackseed:
**Auth Login**

- [ ] Put back the background-800 on cards. (same as above)

- [ ] Signing-up : Line height problem here (didn't know how to sole it with inspector css)
![Capture d’écran 2021-04-28 à 16 25 34](https://user-images.githubusercontent.com/66258851/116420941-b69b4580-a83e-11eb-9eba-d7fd2942877d.png)

- [ ] Color white on the text (contrast color for primary_500 backgrounds in the builder)
[...]

PR #5669, opened 2021-04-29
Issue 5626

fritzschoff:
ref #5626

### DIFF
--- a/styles/components/mat-toggle-button.scss
+++ b/styles/components/mat-toggle-button.scss
@@ -11,7 +11,7 @@
     border-left: solid 1px mat-color($primary, 500);
   }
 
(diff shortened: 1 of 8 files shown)


GrandSchtroumpf (approved):

styles/components/mat-toggle-button.scss
```diff
@@ -23,5 +23,6 @@
 
   .mat-button-toggle-checked.mat-button-toggle-appearance-standard {
     background-color: mat-color($primary, 800);
+    color: white
```
GrandSchtroumpf, line 26: It should be `mat-color($primary, contrast-800)` or something like that instead.

PR #5681, opened 2021-04-30
Issue 5626

fritzschoff:
ref #5626


libs/user/src/lib/auth/components/accept-conditions/accept-conditions.component.html
```diff
@@ -1,8 +1,8 @@
 <mat-checkbox test-id="checkbox" (change)="toggleCondition()" color="primary">
   <span class="mat-caption"><ng-content select="[conditionText]"></ng-content>
-    <p (click)="openTermsOfUse(displayConditions)" class="terms-link">
+    <a (click)="openTermsOfUse(displayConditions)" class="terms-link">
```
GrandSchtroumpf, line 3: if you use a you need to have routerLink or a href. You can transform the `opentermsofUse` into an observable of a link maybe ?

libs/user/src/lib/auth/components/widget/widget.component.scss
```diff
@@ -28,3 +28,7 @@ widget-card {
   transform: translate(-50%);
   color: var(--warn);
 }
+// .mat-button class has a default margin-left which doesnt make sense here
+.logout-btn {
```
GrandSchtroumpf, line 32: Where does this margin come from ? It's not by default in angular material. Overriding here is not a good idea, it looks like a deprecated rule that we override locally instead of fixing it

styles/components/anchor.scss
```diff
@@ -0,0 +1,19 @@
+@mixin anchor($theme) {
+  $primary: map-get($theme, primary);
+
+  a:not(.mat-button) {
```
GrandSchtroumpf, line 4: use a:not(.mat-button-base) instead, like that we cover all material button cases

GrandSchtroumpf (approved):

libs/user/src/lib/auth/components/widget/widget.component.scss
fritzschoff: as stated in the comment, it comes from .mat-button class

PR #5683, opened 2021-04-30
Create BucketForm

GrandSchtroumpf:
Create a bucket form
Add method to toggle a territory on the form

### DIFF
--- a/libs/contract/tsconfig.lib.json
+++ b/libs/contract/tsconfig.lib.json
@@ -12,7 +12,7 @@
     "experimentalDecorators": true,
     "importHelpers": true,
     "types": [],
-    "lib": ["dom", "es2018"]
+    "lib": ["dom", "es2018", "ES2019.Array"]
   },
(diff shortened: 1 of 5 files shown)


libs/contract/tsconfig.lib.json
```diff
@@ -12,7 +12,7 @@
     "experimentalDecorators": true,
     "importHelpers": true,
     "types": [],
-    "lib": ["dom", "es2018"]
+    "lib": ["dom", "es2018", "ES2019.Array"]
```
phpgeekfr, line 15: what is this doing ?

phpgeekfr (approved):
LGTM

libs/contract/tsconfig.lib.json
GrandSchtroumpf: It allows to use `[["a"], ["b"]].flat() // ["a", "b"]`

PR #5691, opened 2021-04-30
WIP avails form map

phpgeekfr:


apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
```diff
@@ -38,18 +44,97 @@ export class MarketplaceMovieAvailsComponent implements OnInit {
...
+        })
+
+
+        return availableTerritories;
+      })
+    );
+
+    // Territories that are not under a mandate
+    // @TODO #5573 usefull ? &  use form values 
+    this.notLicensedTerritories$ = this.termService.valueChanges(this.mandates.map(m => m.termIds).flat()).pipe(
```
GrandSchtroumpf, line 112: From what I understand it's all territories by default, so you don't need to do a query on it. In the constructor you can do 
```typescript
this.allTerritories = Object.keys(territories).filter(t => !!territoriesISOA3[t]).map(t => ({
  isoA3: territoriesISOA3[t],
  label: territories[t],
}));
```

libs/contract/src/lib/avails/form/avails.form.ts
```diff
@@ -5,22 +5,27 @@ import { FormControl, FormGroup, Validators } from '@angular/forms';
...
+
+  return controls;
 }
 
 type AvailControl = ReturnType<typeof createAvailControl>
 
 export class AvailsForm extends FormEntity<AvailControl, AvailsFilter> {
-  constructor(avail: Partial<AvailsFilter> = {}) {
-    super(createAvailControl(avail))
+  constructor(mode: 'single' | 'multi' = 'multi', avail: Partial<AvailsFilter> = {}) {
```
GrandSchtroumpf, line 28: I would rather do something like that: 
```typescript
constructor(avail: Partial<AvailsFilter> = {}, required:  (keyof BucketTerms)[]))
```

RemcoSimonides (approved):

Issue #5625, opened 2021-04-27
Movie: CRUD, list, view, wishlist


Jackseed:
Tester: Max;
Debugger: Samir

Jackseed:
- [ ] Wrong color foreground-base in light theme and dark theme for the gradient here.

first color of gradient should be #8DA4FA in light
![Capture d’écran 2021-04-28 à 16 19 34](https://user-images.githubusercontent.com/66258851/116419634-8acb9000-a83d-11eb-956f-432b7252ddb5.png)

corrected
![Capture d’écran 2021-04-28 à 16 14 20](https://user-images.githubusercontent.com/66258851/116418932-dcbfe600-a83c-11eb-89fe-b60aa8b6e56f.png)

first color of gradient should be #30344A in dark
![Capture d’écran 2021-04-28 à 16 16 44](https://user-images.githubusercontent.com/66258851/116419278-30323400-a83d-11eb-814d-b1c5568ebedf.png)

corrected 
![Capture d’écran 2021-04-28 à 16 16 33](https://user-images.githubusercontent.com/66258851/116419247-28728f80-a83d-11eb-934c-344936d34458.png)



**Movie: CRUD, list, view, wishlist**
- [ ] Builder/theme colors error on the titles and paragraph.
all h1, h2, h3... should be associated with the var foreground-800
all p, body-1, body-2 foreground-700
all captions foreground-600
This should improve the contrast btw different text types.
<img width="676" alt="Capture d’écran 2021-04-28 à 10 32 45" src="https://user-images.githubusercontent.com/66258851/116375692-cf403700-a80f-11eb-9f12-c3aa3c20964f.png">

- [ ] Chips color background should be associated to background-600
which should correspond to #E7E7E7 in light and #16192E in dark.

On the screenshot, the corrected version (what it should look like)
<img width="389" alt="Capture d’écran 2021-04-28 à 10 56 09" src="https://user-images.githubusercontent.com/66258851/116376590-b5532400-a810-11eb-910d-c77dcf2e9b0b.png">
<img width="412" alt="Capture d’écran 2021-04-28 à 10 57 11" src="https://user-images.githubusercontent.com/66258851/116376598-b6845100-a810-11eb-819a-1c315993f589.png">

- [ ] h5 font-weight from 700 to 600
(I know we're supposed to take off all h5 but we won't be able to work on that soon)
See difference expected here:
[...]

PR #5692, opened 2021-04-30
UI card

pldespaigne:
multiple ui fixes related to #5625

### DIFF
--- a/styles/components/mat-tabs.scss
+++ b/styles/components/mat-tabs.scss
@@ -1,0 +1,15 @@
+@mixin mat-bf-tabs-theme($theme) {
+  $primary: map-get($theme, primary);
+  $accent: map-get($theme, accent);
+  $warn: map-get($theme, warn);
+  $background: map-get($theme, background);
+  $foreground: map-get($theme, foreground);
+
+  .mat-tab-label:hover {
+    .mat-icon {
+      path, g, polygon {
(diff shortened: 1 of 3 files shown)


styles/components/mat-tabs.scss
```diff
@@ -0,0 +1,15 @@
...
+  $warn: map-get($theme, warn);
+  $background: map-get($theme, background);
+  $foreground: map-get($theme, foreground);
+
+  .mat-tab-label:hover {
+    .mat-icon {
+      path, g, polygon {
+        fill: mat-color($foreground, base);
+      }
+    }
```
Mussuranae, line 13: You could put it directly on the mat-icon file no ?

pldespaigne: I could but that would put `.mat-tab-label` in the mat-icon file.
Imo both solution are ok and none of them are perfect ^^

Mussuranae: And we want this behavior only when the icon is inside a tab ?

Mussuranae (approved):

styles/components/mat-tabs.scss
Mussuranae: If yes, so that's okay, if no and we want it for all icons, you can move it to the icon file :)

Issue #5655, opened 2021-04-28
Avails Search Criteria


RemcoSimonides:
- [x] Create Avail Search Criteria UI
![image](https://user-images.githubusercontent.com/27687382/116384890-93f63600-a818-11eb-9e73-d6099c44a592.png)
- [x] Clear button resets the form
- [x] Apply filters gets the terms and updates the map
- [x] Rework AvailsFilterForm to have dynamic validation because the AvailsFilterForm doesn't need territories on this page. Quick sketch by Francois as inspiration:
![image](https://user-images.githubusercontent.com/27687382/116386526-2945fa00-a81a-11eb-9709-ef1313842e01.png)

(there is a [separate issue](https://github.com/blockframes/blockframes/issues/5646) for modal behind "learn how it works" link)

PR #5719, opened 2021-05-05
5655 isSold without territories

RemcoSimonides:
Issue #5655

### DIFF
--- a/libs/contract/src/lib/avails/avails.ts
+++ b/libs/contract/src/lib/avails/avails.ts
@@ -53,6 +53,13 @@ export function isSold(
   { medias, duration, territories, exclusive }: AvailsFilter,
   terms: Term<Date>[], // Terms of all sales of the title
 ) {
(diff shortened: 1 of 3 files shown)


libs/contract/src/lib/avails/avails.ts
```diff
@@ -83,7 +83,7 @@ export function isSold(
       continue;
     }
   }
-  return false;
+  return result;
```
phpgeekfr, line 86: good idea, does the method need a renaming ?

like `getSoldTerms` ?

And another one :
```typescript
export function isSold(...) {
  return this.getSoldTerms(...).length;
}
```

phpgeekfr (approved):

libs/contract/src/lib/avails/avails.ts
phpgeekfr: doing this should correct the broken unit tests

PR #5720, opened 2021-05-05
More css

pldespaigne:
last fix about link & button


styles/components/mat-button.scss
```diff
@@ -14,9 +14,9 @@
   .mat-stroked-button,
   .mat-button,
   .mat-icon-button {
-    .mat-button-wrapper {
+    // .mat-button-wrapper {
```
phpgeekfr, line 17: temporary ? If not, can you remove ?

phpgeekfr (approved):

PR #5721, opened 2021-05-05
Merge back r 3 1 develop

phpgeekfr:


libs/event/src/lib/layout/view/view.component.scss
```diff
@@ -30,6 +30,12 @@ section {
   .link {
     text-decoration: underline;
   }
+
+  div {
```
GrandSchtroumpf, line 34: I doesn't look like the right way to handle that. Might create headache for future devs

libs/media/src/lib/file/explorer/explorer.component.scss
```diff
@@ -1,63 +1,43 @@
...
-    .not-selected {
-      margin-right: 5px;
-    }
   }
 
   section {
-    width: 100%;
+    flex: 1;
     padding: 0;
+    height: fit-content;
```
GrandSchtroumpf, line 18: it doesn't make sense to have fit-content & flex: 1. It feels like something is not well defined here.

libs/ui/src/lib/list/filter/list-filter.component.scss
```diff
@@ -25,3 +25,8 @@ mat-card {
   margin-top: 32px;
   min-width: 400px;
 }
+
+ /* To remove the vertical scrollbar on Safari */
+body {
```
GrandSchtroumpf, line 30: NO ! I can understand that Safari is not the best browser but : 
1. It can handle scrollbar correctly
2. You don't override the body in a component

This need to be clean in high priority.

libs/user/src/lib/auth/components/accept-conditions/accept-conditions.component.html
```diff
@@ -1,8 +1,8 @@
 <mat-checkbox test-id="checkbox" (change)="toggleCondition()" color="primary">
   <span class="mat-caption"><ng-content select="[conditionText]"></ng-content>
-    <p (click)="openTermsOfUse(displayConditions)" class="terms-link">
+    <a (click)="openTermsOfUse(displayConditions)" class="terms-link">
```
GrandSchtroumpf, line 3: `<a>` must have `href` or `routerLink`. Use `<button>` instead.

libs/user/src/lib/auth/pages/identity/identity.component.scss
```diff
@@ -66,6 +66,25 @@ section {
   }
 }
 
+.pass-info {
+  align-self: flex-start;
+  display: flex;
+  align-items: center;
+  margin-bottom: 8px;
+  p {
+    margin: 0 6px;
```
GrandSchtroumpf, line 75: spacing should always be multiple of 4px

libs/user/src/lib/auth/pages/identity/identity.component.scss
```diff
@@ -66,6 +66,25 @@ section {
...
+    button {
```
GrandSchtroumpf, line 76: add a comment with issue id. This should be removed after updating the button design in `<p>`

libs/utils/src/lib/spreadsheet/format.ts
```diff
@@ -376,6 +376,8 @@ export function formatSingleValue(value: string, scope: Scope, path: string, mov
         movie[pathParts[0]] = key;
       } else if (pathParts.length === 2) {
         movie[pathParts[0]][pathParts[1]] = key;
+      } else if (pathParts.length === 3) {
```
GrandSchtroumpf, line 379: This should be done recursively.

styles/components/anchor.scss
```diff
@@ -0,0 +1,6 @@
+@mixin anchor($theme) {
+  a:not(.mat-button):not(.mat-flat-button):not(.mat-stroked-button),
```
GrandSchtroumpf, line 2: Change to : 
```scss
a:not(.mat-button-base),
a:not(.mat-button-base):visited 
```

GrandSchtroumpf (approved):

apps/catalog/catalog/src/app/landing/landing.component.scss
```diff
@@ -22,6 +22,7 @@ landing-content {
     margin-top: 24px;
     img {
       max-width: 600px;
+      max-height: 418px;
```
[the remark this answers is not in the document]
SamirTouira: I specified 418px because on Google Chrome for example, when I have inspected the images about this issue with devtools, I saw '418px' in height. So I wanted to put the same value to get the same size on Safari to keep the same image size for every navigators. What is the best solution for this please? What is the expected value in vh to avoid this issue?

Issue #5530, opened 2021-04-08
Angular pipe updates


phpgeekfr:
in  `ToLabelPipe` and `DisplayNamePipe` remove the use of `limit` or `max`  and use  `css text-overflow: elipsis` instead.

Created from https://github.com/blockframes/blockframes/pull/5513

![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/c54dc8cd-eb2d-480a-b535-f6c9ecbc2500)

fritzschoff:
I dont this will work. SInce this css property needs absolute units from the parent container, if not provided, then there won't be any `...`

SamirTouira:
I don't know if this is the right way, but I have added a  `class="elipsis"` in the `<section>` for the territories in the selection component (catalog) just for the test. Then I've wrote this `text-overflow: elipsis` in the css, it not seems to work so I think the parameter limit?: sounds good, correct me if i'm wrong but what about replacing only `...` instead of 'limit.lenght more' inside the pipe if the text-overflow property gives like a `...` render ?
![image](https://user-images.githubusercontent.com/52285231/116102629-2a0e4d00-a6af-11eb-9277-52860d13ea45.png)

PR #5755, opened 2021-05-07
limit? toLabel and max? displayName pipes removed, css instead

SamirTouira:
issue: #5530


libs/ui/src/lib/marketplace/pages/wishlist/wishlist.component.scss
```diff
@@ -19,3 +19,7 @@ img {
   width: 380px;
   margin: 24px 0;
 }
+
+.ellipsis {
+  width: 300px;
```
Mussuranae, line 24: It would be better to put `max-width` I guess

Mussuranae (approved):

Issue #5677, opened 2021-04-29
OTHER - Business Review 


maltmayer:
## DASHBOARD MOVIE FORM

**wait for confirmation**
- [x] Shooting Information should go after Production Information 
![image](https://user-images.githubusercontent.com/55785823/116568267-a18ae900-a908-11eb-9fae-9bd2ee760612.png)
Actually I'm wondering if "Shooting information" shouldn't go into "Production information"

**wait for confirmation**
- [x] Box office fields should be in this order : 1. Figures 2. Metrics 3. Country (all three together on one line if possible) Otherwise 1. Figures 2. Metrics (together on one line) 3. Country
![image](https://user-images.githubusercontent.com/55785823/116578654-eb2c0180-a911-11eb-8599-f292ebdf6f0f.png)

## DASHBOARD EVENT FORM
**Ok for fix (Samir)**
- [x] For usability, the first displayed hour in the list here should be the one already selected
![image](https://user-images.githubusercontent.com/55785823/116576123-91c2d300-a90f-11eb-8414-e560ff681715.png)

### MARKETPLACE MOVIE VIEW

**Ok for fix (Clélia)**
- [x] Subtitles behave like they're clickable but they're not > should be changed for all the mat-chips in the app **that doesn't have a clickable action (like cancel button)**
![image](https://user-images.githubusercontent.com/55785823/117002315-b211d980-ace3-11eb-935f-232e53f4bad3.png)

## NOTIFICATIONS
- [x] Mark all as read icon should be consistent with the Mark as read icon I think (with the little checks)
![image](https://user-images.githubusercontent.com/55785823/116577639-f29edb00-a910-11eb-906e-34171dd3c17b.png)
![image](https://user-images.githubusercontent.com/55785823/116577831-211cb600-a911-11eb-88c0-acc6922b2a59.png)

maltmayer:
## TOOLBAR ICONS

- [ ] I feel like we should work on creating icons that are all the same size, as these have random size and therefore give an impression of weirdly aligned icons
![image](https://user-images.githubusercontent.com/55785823/116581406-963dba80-a914-11eb-9c04-f547cf2ddda1.png)

maltmayer:
## MARKETPLACE MOVIE VIEW 
- [ ] Maybe we should find a way to limit a movie title to 2 lines... otherwise the banner is not pretty
[...]

PR #5778, opened 2021-05-11
Business fixes

Mussuranae:
3 fixes from #5677

image
image
image


apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.ts
```diff
@@ -10,15 +10,17 @@ import { MatDialog } from '@angular/material/dialog';
...
 @Component({
   selector: 'catalog-selection',
   templateUrl: './selection.component.html',
   styleUrls: ['./selection.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class MarketplaceSelectionComponent {
-  currencies = movieCurrencies;
+export class MarketplaceSelectionComponent implements OnInit {
+  withoutCurrencies = Object.keys(movieCurrencies).filter(currency => !['EUR', 'USD'].includes(currency));
```
GrandSchtroumpf, line 22: This is weird:
- keys should not be in capital letters
- Use includes when more than 3 items
```typescript
withoutCurrencies = Object.keys(movieCurrencies).filter(currency => currency !== 'eur' || currency !== 'usd');
```

GrandSchtroumpf (approved):

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.ts
Mussuranae: Sorry for the capital letters, but actually, I can do nothing they are all like that in the static model. It would need a migration to change it. I won't do that in my PR.
image

Issue #5760, opened 2021-05-10
Review Avails WIP 


2amarkina:
**Texts** 

- [x] Fill in the following fields to know where, when and in which conditions this title is available. (Dot is missing missing)
- [x]   This phrase should start from newline: «Search results use an "and" logic for multiple conditions, not an "or" logic.»
- [x] «Available Territories»  with capital letter

**Main**
- [x] we need to change a switcher (on/off ; activate/desactivate) to toggle buttons choosing between same-level elements but related to different user tasks. 
![Снимок экрана 2021-05-10 в 13.08.02.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/bad95912-4b05-4001-a785-3f7f33a52e3f)

- [x] when switch to «single territory view» the «avails search criteria» should also change 
- [x] need different colors for "not licensed" and "available" 
- [x] when moving the map strange bug happens: map is getting cut
![moving.gif](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/a3b8e696-7764-4b8e-a99a-fa3f441f5da7)

- [x] when switch to other page (i.g. Contact Company) warning modal should also appear (screenshot) (as it does when clicking on Selection icon for example)
if click here ![Снимок экрана 2021-05-10 в 13.15.49.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/e21b9725-9fd5-46ed-bc1c-a36bcba329f9) get also this ![Снимок экрана 2021-05-10 в 12.24.45.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/597a0ad6-1ca2-4c7c-9eca-eea052a86df2) 

- [x] Can't use "select all available countries" 

**Rights Card**
- [x] on the Rights card "edit" and "delete" are missing 
- [x] when adding 1 country, three Rights instead of 1 are created: 
- one card OK (the last one)
- one card added with all non-exclusive countries (initial search was for exclusive) (a long country list)
- one card with two countries (Hungary, Australia in example below). 
![add countries.gif](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/63bcfa2f-e3be-4770-b03e-9de79938e79e)

- [x] when deleting a country it works only for 1 rights card, (which is a good one in previous paragraph), other cards with many countries do not move. 

**Adding to Selection/Selection**
- [x] preselected rights and CTA appear, but no rights added
[...]

PR #5785, opened 2021-05-12
5760 Review Avails

RemcoSimonides:
Fix issue #5760 

@pldespaigne, I did some text changes in the calendar component too. I guess it doesn't collide with your code too much.

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
+++ b/apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
@@ -15,6 +15,8 @@ import { ConfirmComponent } from '@blockframes/ui/confirm/confirm.component';
 
 import { ExplanationComponent } from './explanation/explanation.component';
 import { switchMap } from 'rxjs/operators';
+import { MatSnackBar } from '@angular/material/snack-bar';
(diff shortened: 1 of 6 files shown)


apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
```diff
@@ -39,30 +41,38 @@ export class MarketplaceMovieAvailsComponent implements OnInit, OnDestroy {
 
   public terms$ = this.bucketForm.selectTerms(this.movie.id);
 
-  private sub: Subscription;
+  private subs: Subscription[] = [];
```
pldespaigne, line 44: I'm not sure to understand why do you need an Array of subs here 🤔

pldespaigne (approved):

apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
RemcoSimonides: Could also add another property in this class for the second subscription and check if it exists on destroy. With the array I can just loop over the subscriptions. Doesn't really matter which solution to use I guess

PR #5786, opened 2021-05-12
height logo for whole toolbar

SamirTouira:
issue: #5677 
comment:
image


libs/ui/src/lib/layout/app-logo/app-logo.component.scss
```diff
@@ -1,6 +1,6 @@
 :host {
-  height: 48px;
   display: block;
+  margin-top: 8px;
```
GrandSchtroumpf, line 3: You should never put a margin style in a host element. Margin is to organize element between each other. A component style is supposed to be self contained.
Either we defined a specific size for the app-logo, and it's the same everywhere in the app (as we had before), or we apply the height on every case.
I would recommend to define the height of the app logo so it's the same everywhere, else we'll end up having uncoherent design across apps.
Note: The height can change depending on the size of the screen if need be.

Ask @stephanie-cascade8 what should be the right height for the app-logo.

GrandSchtroumpf (changes requested):

GrandSchtroumpf (approved):

Issue #5653, opened 2021-04-28
Map component


RemcoSimonides:
https://app.zeplin.io/project/5fecb41a2e806651584e265c/screen/6086ad1bcd8a35129d08f3e3
( no need to support "Select all available territories" button for now )
- [x] Link to org contact page (text above map)
- [x] For now we only need to be able to select 1 country at a time on the map
- [x] Selected territories are added to preselected rights

PR #5787, opened 2021-05-13
5653 Restyle map

RemcoSimonides:
tag issues #5760 #5653
Improve colors map


apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
```diff
@@ -67,22 +67,22 @@
 
       <!-- MAP LAYERS -->
       <ng-container *ngFor="let tag of (available$ | async); trackBy: trackByTag">
-        <map-feature [tag]="tag.isoA3" color="color-1" (click)="addTerritory(tag)"
+        <map-feature [tag]="tag.isoA3" color="primary" [weight]="1" (click)="addTerritory(tag)"
```
GrandSchtroumpf, line 70: weight=1 should be default, you shouldn't need to specify it

libs/ui/src/lib/map/map.component.ts
```diff
@@ -22,15 +22,24 @@ import { startWith, switchMap, map } from 'rxjs/operators';
...
 
   @Input()
   set color(color: string) {
-    this.color$.next(color);
+    this.state = color ? { fillColor: `var(--${color})`, fillOpacity: 1 } : { fillOpacity: 0 };
   }
-  get color(): string {
-    return this.color$.getValue();
+  @Input()
+  set weight(weight: number) {
```
GrandSchtroumpf, line 40: You could also support string value to do `weight="2"` without the [] :
```typescript
set weight(value: number | string) {
  const weight = typeof value === "string" ? parseInt(value) : (value ?? 1);
  this.state = { weight };
}
```
Using `??` will keep 0 if input is 0

GrandSchtroumpf (approved):

PR #5795, opened 2021-05-14
5760 Update avails tab UI

RemcoSimonides:
issue #5760

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
+++ b/apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
@@ -57,17 +57,33 @@
 <mat-card class="avails-card">
   <mat-card-header>
     <mat-card-title>Available Territories</mat-card-title>
+  </mat-card-header>
+  <mat-divider></mat-divider>
+  <mat-card-header>
     <mat-card-subtitle *ngIf="org$ | async as org">
-      Click on the territories you're interested in to add them to your Selection. For territories that are not licensed, feel free to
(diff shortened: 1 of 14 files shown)


apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
```diff
@@ -57,17 +57,33 @@
...
+        <ng-container *ngIf="status === 'VALID' else invalid">
+        </ng-container>
+      </ng-container>
+      <ng-template #invalid>
+        <section class="overlay" fxLayoutAlign="center center">
+          <article class="mat-body-2">
+            Please fill in Avail Search Criteria section to show results on the Map.
+          </article>
+        </section>
+      </ng-template>
```
RemcoSimonides, line 82: availsForm.statusChanges doesn't fire at the start. The availsForm has to change first.

As `*ngIf` doesn't work for undefined, I decided to use the else here.

styles/components/mat-form-field.scss
```diff
@@ -23,6 +23,11 @@
         display: none;
       }
     }
+
+    mat-icon[matPrefix] {
+      margin-left: 10px;
+      margin-right: 10px;
```
RemcoSimonides, line 29: This is not a multiple of 4 but this is what comes closest to the date fields
image

apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
GrandSchtroumpf: You can do in the component: 
```typescript
status$ = this.availsForm.statusChanges.pipe(startWith('INVALID'));
```

GrandSchtroumpf (approved):

PR #5802, opened 2021-05-17
5760 Bugs and added days to runs duration

RemcoSimonides:
- [x] Fix observables not triggering on map component
- [x] Fix no starting value on currency field
- [x] Added days to runs duration

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.ts
+++ b/apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.ts
@@ -38,9 +38,9 @@ export class MarketplaceSelectionComponent implements OnDestroy {
     distinctUntilChanged()
   );
 
-  trackById = (i: number, doc: { id: string }) => doc.id;
-
   private sub = this.currencyForm.valueChanges.pipe(distinctUntilChanged()).subscribe(value => this.updateCurrency(value));
+
+  trackById = (i: number, doc: { id: string }) => doc.id;
 
   constructor(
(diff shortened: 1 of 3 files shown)


apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.ts
```diff
@@ -50,14 +52,15 @@ export class MarketplaceSelectionComponent implements OnInit {
   ) {
     this.bucket$ = this.bucketQuery.selectActive().pipe(
       tap(bucket => {
-        this.setTitle(bucket?.contracts.length)
-        bucket?.contracts.forEach((contract, i) => this.setPrice(i, contract.price))
-      }),
+        this.setTitle(bucket?.contracts.length);
+        if (!!bucket?.currency) this.currencyForm.setValue(bucket.currency);
```
RemcoSimonides, line 56: This does unnecessarily update the bucket once with the starting value of currency when initializing this page. But I couldn't find a way to avoid this :/ do you know?

Mussuranae: I guess you have to subscribe to your `currencyForm` (line 41) after initialized it ? Your subscription leads to the `updateCurrency` function that update the bucket, and you set a value here, so it triggers the `updateCurrency` function.
I don't know if the constructor is set before the `OnInit` of component, but maybe if you let your subscription in a `ngOnInit`, it won't happen ?

Mussuranae (approved):

PR #5803, opened 2021-05-17
Business fixes

Mussuranae:
Fixed some plural, order title in event form, and color of the missing control component


styles/components/bf-missing-control.scss
```diff
@@ -2,11 +2,11 @@
...
   $warn: map-get($theme, warn);
 
-  .missing-control {
+  a.missing-control span {
     color: #F6A92C;
   }
 
-  .mandatory-control {
+  a.mandatory-control {
     color: mat-color($warn, 500);
```
Mussuranae, line 10: If I can suggest to remove the `bf-missing-control.scss` because it was here to set the color of the missing component. But for the design rework of the anchor tag, we create the `anchor` file, and those files have conflict between them. We could merge them completely. The `anchor` file can also be used for the link in blue (exemple with the link to dowload file/images) and all other links we want to custom. I'm afraid that we will currently put a lot of `color: var(--primary);` everywhere instead of trying to stay consistent...
image

apps/festival/festival/src/app/marketplace/event/ended/meeting-ended.component.html
```diff
@@ -12,7 +12,7 @@ <h2>You just left the Meeting.</h2>
     <h4 *ngIf="duration | duration: '' as pipedDuration">It lasted {{ pipedDuration }}.</h4>
 
     <ng-container *ngIf="attendees$ | async as attendees">
-      <h3>There were {{attendees.length}} attendee{attendees.length, plural, =1 {} other {s}}</h3>
+      <h3>There were {{attendees.length}} attendee{attendees.length, plural, =0 {} =1 {} other {s}}</h3>
```
RemcoSimonides, line 15: It's supposed to be "There were 0 attendees", right?
With an 's' when it's 0

libs/media/src/lib/file/upload-widget/upload-widget.component.html
```diff
@@ -1,7 +1,7 @@
 <mat-accordion *ngIf="(tasks.$ | async)?.length as itemCount">
   <mat-expansion-panel expanded>
     <mat-expansion-panel-header i18n>Uploading {{ itemCount }} item{ itemCount,
-      plural, =1 {} other {s} }
+      plural, =0 {} =1 {} other {s} }
```
RemcoSimonides, line 4: Same here. I think with 0 it's correct to include the 's'. "0 items", not "0 item".

libs/ui/src/lib/static-autocomplete/check-boxes/check-boxes.component.ts
```diff
@@ -48,10 +48,14 @@ export class StaticCheckBoxesComponent implements OnInit, OnDestroy {
     ).subscribe(_ => this.checkboxes.forEach(box => box.checked = false))
   }
 
+  compare() {
+    return (a, b) => 1;
+  }
```
RemcoSimonides, line 53: What's this magic doing? 😅

styles/components/bf-missing-control.scss
RemcoSimonides: This one is for Francois to determine

apps/festival/festival/src/app/marketplace/event/ended/meeting-ended.component.html
Mussuranae: Before the changes yes. With the changes it will be `there were 0 attendee`, that's not better ? In fact if we want to do it very well that would be `there was none attendee` no ?

libs/ui/src/lib/static-autocomplete/check-boxes/check-boxes.component.ts
Mussuranae: It's François' power xD
Order the select with the same order in the static model x) The keyvalue pipe order the list by alphabetic order, but we want the original order (the one of the static model). That's the magic here ^^'

GrandSchtroumpf (approved):

PR #5814, opened 2021-05-19
tests for avail-calendar

phpgeekfr:
Still some cases to cover but this is a start


libs/contract/src/lib/avails/tests/calendar.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    const mandateA = createMandate({ id: 'MandateA', termIds: ['termA'] });
+    const mandateB = createMandate({ id: 'MandateB', termIds: ['termB'] });
+
+    const termA = createTerm({ id: 'termA', contractId: mandateA.id });
+
+    const selectedA = toDurationMarker([mandateA, mandateB], termA);
+    expect(selectedA.contract.id).toEqual(mandateA.id);
+
+    const selectedB = toDurationMarker([mandateA, mandateB], termA);
+    expect(selectedB.contract.id).toEqual(mandateA.id);
```
RemcoSimonides, line 31: You're testing twice the same thing here

RemcoSimonides: termA should be termB

RemcoSimonides (approved):

libs/contract/src/lib/avails/tests/calendar.spec.ts
phpgeekfr: good catch thx

PR #5817, opened 2021-05-20
lint user lib

GrandSchtroumpf:


libs/user/src/lib/auth/forms/profile/profile.component.html
```diff
@@ -26,7 +26,7 @@
         <mat-label>Email Address</mat-label>
         <input matInput test-id="phone-form" placeholder="Email Address" type="tel" [formControl]="form.get('email')" />
       </mat-form-field>
-      <div *ngIf="!(this.authQuery.hasVerifiedEmail$ | async)" fxLayoutAlign="start center">
+      <div *ngIf="(authQuery.hasVerifiedEmail$ | async) === (false || null)" fxLayoutAlign="start center">
```
RemcoSimonides, line 29: How is hasVerifiedEmail ever null?
I didn't know about `x === (1 || 2)` yet so thanks

libs/user/src/lib/user/+state/user.store.ts
```diff
@@ -2,8 +2,7 @@ import { Injectable } from "@angular/core";
 import { StoreConfig, EntityStore, EntityState } from "@datorama/akita";
 import { User } from "@blockframes/auth/+state/auth.store";
 
-export interface UserState extends EntityState<User> {
-}
+export type UserState = EntityState<User>;
```
RemcoSimonides, line 5: So this is how to solve this one :D I was removing the empty interfaces

RemcoSimonides (approved):
LGTM

PR #5820, opened 2021-05-20
Notification Lint

Mussuranae:
Fix Notification lib ES Lint errors and warnings.


libs/notification/src/lib/+state/notification.service.spec.ts
```diff
@@ -65,7 +65,7 @@ describe('Notifications Test Suite', () => {
   it('Formats notification', () => {
     const notificationStore = TestBed.inject(NotificationStore)
     notificationStore.formatNotification = jest.fn();
-    service.formatFromFirestore({} as any);
+    service.formatFromFirestore({} as Notification);
```
GrandSchtroumpf, line 68: I think we should update the `formatFromFirestore` function to support a `Partial<Notification>` instead.

libs/notification/src/lib/components/item/item.component.html
```diff
@@ -1,7 +1,7 @@
 
 <mat-list-item>
   <img matListAvatar [asset]="notification.placeholderUrl" [ref]="notification.imgRef" alt="Notification picture">
-  <p *ngIf="!(xs$ | async)" matLine class="mat-body-1" test-id="notification-message">{{ notification.message }}</p>
+  <p *ngIf="(xs$ | async) === (false || null || undefined)" matLine class="mat-body-1" test-id="notification-message">{{ notification.message }}</p>
```
GrandSchtroumpf, line 4: remove the false. xs$ cannot be false.

GrandSchtroumpf (approved):

libs/notification/src/lib/+state/notification.service.spec.ts
Mussuranae: I will do it in a next PR, I already worked a lot on the utils lib fixes, and this function is a little more complex to fix than "just" change `Notification` by `Partial<Notification>`.

PR #5821, opened 2021-05-20
Fix linter errors. Few warnings still need to be sorted out.

xmano:
I need help with few more lines of code..

libs\media\src\lib\file\viewers\utils.ts

```
(container.nativeElement as any).webkitRequestFullscreen();

(document as any).webkitExitFullscreen();

```

libs\media\src\lib\image\uploader\reference-path.pipe.ts
```
    const getRef = referencesPaths[type][key] as any;
    return getRef(docId)
```

libs\media\src\lib\file\explorer\explorer.component.ts  - can I remove ...args here?
```
  keepOrder(...args) { return 0; }

```

libs\media\src\lib\+state\file-uploader.service.ts
Promise.allSettled() .. configs to be changed by you.


libs/media/.eslintrc.json
```diff
@@ -11,11 +11,11 @@
       "parserOptions": { "project": ["libs/media/tsconfig.*?.json"] },
       "rules": {
         "@angular-eslint/directive-selector": [
-          "warning",
+          1,
```
GrandSchtroumpf, line 14: It should be "warn" instead of 1

libs/media/src/lib/file/carousel/carousel.component.ts
```diff
@@ -18,7 +18,7 @@ export class FileCarouselComponent {
   @Input() files: StorageFile[];
 
   /** Emit the file path that has been selected by the user */
-  @Output() select = new EventEmitter<string>();
+  @Output() selectedPath = new EventEmitter<string>();
```
GrandSchtroumpf, line 21: Change to `selectPath`, make sure it's updated in every html file that use it (if any)

libs/media/src/lib/image/directives/imgix-helpers.ts
```diff
@@ -53,7 +52,7 @@ export function getImgSize(ref?: string) {
 export function formatParameters(parameters: ImageParameters): string {
 
   const query = Object.entries(parameters)
-    .filter(([key, value]) => !!value)
+    .filter(([value]) => !!value)
```
GrandSchtroumpf, line 55: No, this is not at all the same thing. Change for that: 
```typescript
const query = Object.keys(parameters)
  .filter(key => parameters[key])
  .map(key => `${key}=${parameters[key]}`)
  .join('&');
```

libs/media/.eslintrc.json
xmano: fixed

libs/media/src/lib/file/carousel/carousel.component.ts
xmano: Changed and verifying the clients of this component.

libs/media/.eslintrc.json
xmano: done

libs/media/src/lib/image/directives/imgix-helpers.ts
xmano: Fixed.

xmano:
Seeing some build errors with the changes.. Have to fix those.

PR #5822, opened 2021-05-20
linter cleaning bruce part 2

phpgeekfr:

### DIFF
--- a/libs/e2e/src/lib/utils/functions.ts
+++ b/libs/e2e/src/lib/utils/functions.ts
@@ -97,13 +97,8 @@ let currentID = 0;
 export const randomID = (): string => (`${new Date().getTime()}-${currentID++}`);
 export const createFakeScript = (title: string) => cy.task('random:pdf', title);
 
-export function getTomorrowDay(date: Date) {
-  const tomorrowDay = date.getDay() === 6 ? 0 : date.getDay() + 1;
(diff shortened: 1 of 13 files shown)


libs/e2e/src/lib/utils/functions.ts
```diff
@@ -80,32 +80,32 @@ export function selectAction(element: string,
...
+export const createFakeScript = (title: string) => cy.task('random:pdf', title);
 
 export function getTomorrowDay(date: Date) {
   const tomorrowDay = date.getDay() === 6 ? 0 : date.getDay() + 1;
   return tomorrowDay;
 }
 
 export interface FormOptions {
-  inputValue: any;
+  inputValue;
```
GrandSchtroumpf, line 106: Is there no way to specify the type ? Like `string | number` for example ?

libs/e2e/src/lib/utils/queryinterface.ts
```diff
@@ -1,8 +1,7 @@
-
 export interface QueryInferface {
   exist?: boolean;
   index?: number;
   key?: string;
-  value?: any;
+  value?;
```
GrandSchtroumpf, line 5: Same here.

GrandSchtroumpf (approved):

Issue #5706, opened 2021-05-04
Avails Calendar


pldespaigne:
<details>
<summary>Implement the Avails Calendar</summary>

![prototype](https://user-images.githubusercontent.com/79836137/115029561-64653680-9ec6-11eb-94aa-d0fae0381db6.gif)
![image](https://user-images.githubusercontent.com/5858657/116987286-28f0a780-accf-11eb-9ff7-20b0c653e8eb.png)
https://zpl.io/bl3rPDz

From a dev POV, it can look something like that:

```html
<calendar>
  <ng-container *ngFor="let term of availables">
    <calendar-selection (click)="" color=""></calendar-selection>
  </ng-container>
  <ng-container *ngFor="let term of sold">
    <calendar-selection disabled color=""></calendar-selection>
  </ng-container>
</calendar>
```
</details>

---

- [x] Ask design for more info about behavior & edge cases
- [x] Setup Calendar component & integrate it to the title avails page
- [x] Implement simple design
- [x] Implement calendar behavior
- [x] implement Markers
- [x] split calendar & map in separate pages & factorize some code to parent
- [x] Show Modal when Avails Criteria is not filled in yet
- [x] When delete Preselect Right, also remove the selection in the calendar
- [x] Clear button should also clear calendar
- [x] Edit button on Preselected Rights should fill in the Avails Criteria form + Calendar
- [ ] if possible, add a tooltip when hovering a month with the availability status (eg "Available", "Not Licensed" or "Rights Sold")
- [x] Add same "Rights added." snackbar from the world map when a user selects an end date on the calendar
- [ ] UX review
- [x] UI review
- [x] update legend to use the same color as the calendar
- [x] distinguish sold from in bucket with a third color (don't forget to update the legend accordingly)
- [x] fix small bug when editing in bucket terms
- [x] remove calendar card margin & align card to the left

PR #5826, opened 2021-05-20
5706 Add to preselected rights

RemcoSimonides:
tag issue #5706


libs/contract/src/lib/avails/calendar/calendar.component.ts
```diff
@@ -37,23 +36,36 @@ export class AvailsCalendarComponent implements OnInit {
...
-    this.stateMatrix = markersToMatrix(markers, this.stateMatrix, 'avail');
+    this._licensedMarkers = markers;
+    this.updateMatrix();
   }
 
   @Input() set soldMarkers(markers: DurationMarker[] | undefined) {
     if (!markers) return;
-    this.stateMatrix = markersToMatrix(markers, this.stateMatrix, 'sold');
+    this._soldMarkers = markers;
+    this.updateMatrix();
```
RemcoSimonides, line 51: I put this in a separate function because I wasn't sure whether licensedMarkers could overwrite soldMarkers. this.stateMatrix is set multiple times now though. But I don't see any side-effect in the front-end

libs/contract/src/lib/avails/calendar/calendar.component.ts
```diff
@@ -29,23 +32,36 @@ export class AvailsCalendarComponent implements OnInit {
...
   }
 
+  @Output() selected = new EventEmitter<Duration<Date>>();
```
GrandSchtroumpf, line 50: Prefer `select` (like `click` instead of `clicked`)

libs/contract/src/lib/avails/calendar/calendar.component.ts
```diff
@@ -29,23 +32,36 @@ export class AvailsCalendarComponent implements OnInit {
...
+
   ngOnInit() {
     const state = this.state$.getValue();
     state.highlightedRange = this.rows.map(_ => this.columns.map(__ => false));
 
     this.state$.next(state);
   }
 
+  updateMatrix() {
+    this.stateMatrix = this.rows.map(_ => this.columns.map(__ => 'empty'));
```
GrandSchtroumpf, line 60: For linter
```typescript
this.stateMatrix = this.rows.map(() => this.columns.map(() => 'empty'));
```
Also, it's not a good think to update several time the same property, better: 
```typescript
const matrix = this.rows.map(_ => this.columns.map(__ => 'empty'));
if (this._licensedMarkers.length) matrix = markersToMatrix(this._licensedMarkers, matrix , 'available');
if (this._soldMarkers.length) matrix  = markersToMatrix(this._soldMarkers, matrix , 'sold');
this.stateMatrix = matrix;
```

libs/contract/src/lib/avails/avails.ts
```diff
@@ -154,10 +154,18 @@ export function findSameTermIndex(terms: BucketTerm[], avail: AvailsFilter) {
  * @param avail
  * @returns
  */
-export function isSameTerm(term: BucketTerm, avail: AvailsFilter) {
+export function isSameTerm(term: BucketTerm, avail: AvailsFilter, unrequired?: 'territories' | 'duration') {
   if (term.exclusive !== avail.exclusive) return false;
-  if (!avail.duration?.from || term.duration.from.getTime() !== avail.duration.from.getTime()) return false;
-  if (!avail.duration?.to || term.duration.to.getTime() !== avail.duration.to.getTime()) return false;
+
+  if (unrequired !== 'duration') {
```
pldespaigne, line 160: this `if` will be executed if `unrequired` is `undefined`, is this the wanted behavior ?

libs/contract/src/lib/avails/calendar/calendar.component.ts
pldespaigne: It will trigger a component render every time you call `updateMatrix`, but It's not a big deal imo

pldespaigne (approved):

apps/catalog/catalog/src/app/marketplace/title/avails/calendar/avails-calendar.component.ts
```diff
@@ -68,16 +70,14 @@ export class MarketplaceMovieAvailsCalendarComponent {
...
-    this.mandateTerms$
+    this.mandateTerms$,
+    this.availsForm.value$
   ]).pipe(
-    map(([mandates, selected, sold, inSelection, mandateTerms]) => {
+    map(([mandates, mandateTerms]) => {
       if (this.availsForm.invalid) return [];
-      return availableDurations(selected, sold, inSelection, mandates, mandateTerms);
+      return getDurationMarkers(mandates, mandateTerms);
     })
```
GrandSchtroumpf, line 81: You can `shareReplay(1)` here to avoid combining all over again each time you run `selected`

apps/catalog/catalog/src/app/marketplace/title/avails/calendar/avails-calendar.component.ts
```diff
@@ -90,4 +90,20 @@ export class MarketplaceMovieAvailsCalendarComponent {
   clear() {
     this.availsForm.reset();
   }
+
+  async selected(duration: Duration<Date>) {
+    const licensed = await this.licensed$.pipe(take(1)).toPromise();
```
GrandSchtroumpf, line 95: Here you're creating a new observer for the whole observable. Using `shareReplay` as described above might help.

libs/contract/src/lib/avails/avails.ts
```diff
@@ -154,10 +154,18 @@ export function findSameTermIndex(terms: BucketTerm[], avail: AvailsFilter) {
  * @param avail
  * @returns
  */
-export function isSameTerm(term: BucketTerm, avail: AvailsFilter) {
+export function isSameTerm(term: BucketTerm, avail: AvailsFilter, unrequired?: 'territories' | 'duration') {
```
GrandSchtroumpf, line 157: I don't like this kind of third argument flag that will introduce more if statements in the function. I would rather split the function into small function and combine them for territory or duration: 
```typescript
function hasSameDuration(avail, term) {
  if (!avail.duration?.from || term.duration.from.getTime() !== avail.duration.from.getTime()) return false;
  if (!avail.duration?.to || term.duration.to.getTime() !== avail.duration.to.getTime()) return false;
  return true;
}
function hasSameTerritories() {...}
function hasSameMedia() {...}
```
Then you can use the function combinaison you want for map & calendar

libs/contract/src/lib/bucket/form.ts
```diff
@@ -189,9 +189,51 @@ export class BucketForm extends FormEntity<BucketControls, Bucket> {
...
+
+  /**
+   * This function is based on Duration Marker
+   */
+  getFromBucket(avails: AvailsFilter, marker: DurationMarker): { contractIndex: number, termIndex: number } | undefined {
+    const { term } = marker;
+    const bucket = this.value;
+
+    const contractIndex = bucket.contracts.findIndex(c => c.parentTermId === term.id);
+    if (contractIndex === -1) return undefined;
```
GrandSchtroumpf, line 231: You don't need to specify undefined. Just `return;`

PR #5829, opened 2021-05-20
KLH

phpgeekfr:

### DIFF
--- a/apps/catalog/catalog/src/app/dashboard/right/list/list.component.html
+++ b/apps/catalog/catalog/src/app/dashboard/right/list/list.component.html
@@ -1,6 +1,6 @@
 <h1>Offers & Deals</h1>
 <header mat-tab-nav-bar fxLayout="row" fxLayoutAlign="space-between end">
-  <button mat-tab-link (click)="resetFilter()" [active]="!(filter$ | async)">All Offers</button>
+  <button mat-tab-link (click)="resetFilter()" [active]="(filter$ | async) === (false || null || undefined)">All Offers</button>
   <button mat-tab-link (click)="applyFilter('draft')" [active]="(filter$ | async) === 'draft'">Offers</button>
   <button mat-tab-link (click)="applyFilter('submitted')" [active]="(filter$ | async) === 'submitted'">Ongoing
     Deals</button>
--- a/apps/catalog/catalog/src/app/dashboard/right/list/list.component.html
+++ b/apps/catalog/catalog/src/app/dashboard/right/list/list.component.html
@@ -1,6 +1,6 @@
 <h1>Offers & Deals</h1>
 <header mat-tab-nav-bar fxLayout="row" fxLayoutAlign="space-between end">
-  <button mat-tab-link (click)="resetFilter()" [active]="(filter$ | async) === (null || undefined || '')">All Offers</button>
+  <button mat-tab-link (click)="resetFilter()" [active]="(filter$ | async) === '' ">All Offers</button>
   <button mat-tab-link (click)="applyFilter('draft')" [active]="(filter$ | async) === 'draft'">Offers</button>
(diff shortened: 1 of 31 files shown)


apps/catalog/catalog/src/app/dashboard/right/list/list.component.html
```diff
@@ -1,6 +1,6 @@
 <h1>Offers & Deals</h1>
 <header mat-tab-nav-bar fxLayout="row" fxLayoutAlign="space-between end">
-  <button mat-tab-link (click)="resetFilter()" [active]="!(filter$ | async)">All Offers</button>
+  <button mat-tab-link (click)="resetFilter()" [active]="(filter$ | async) === (false || null || undefined)">All Offers</button>
```
GrandSchtroumpf, line 3: hmmm what it the exact value ? null or undefined or '' ? Be careful here.

apps/catalog/catalog/src/app/dashboard/title/list/list.component.html
```diff
@@ -15,7 +15,7 @@ <h1>My Titles</h1>
     </header>
 
     <div mat-tab-nav-bar fxLayout="row" fxLayoutAlign="space-between end">
-      <button mat-tab-link (click)="resetFilter()" [active]="!(filter$ | async)">All</button>
+      <button mat-tab-link (click)="resetFilter()" [active]="(filter$ | async) === (false || null || undefined)">All</button>
```
GrandSchtroumpf, line 18: Same as above

apps/catalog/catalog/src/app/marketplace/marketplace.component.html
```diff
@@ -39,7 +39,7 @@
     <a mat-icon-button routerLink="selection">
       <mat-icon svgIcon="shopping_basket"
         [matBadge]="contractCount$ | async"
-        [matBadgeHidden]="!(contractCount$ | async)"
+        [matBadgeHidden]="(contractCount$ | async) === (false || null || undefined)"
```
GrandSchtroumpf, line 42: Hmm I think this is 0 here.

libs/invitation/src/lib/+state/invitation.service.spec.ts
```diff
@@ -44,15 +46,15 @@ describe('Invitations Test Suite', () => {
   it('Formats invitation from firestore', () => {
     const is = TestBed.inject(InvitationService);
     is.formatFromFirestore = jest.fn();
-    is.formatFromFirestore({} as any);
+    is.formatFromFirestore({} as InvitationDocument);
```
GrandSchtroumpf, line 49: formatFromFirestore should accept a Partial<InvitationDocument> instead I think.

phpgeekfr: Since formatFromFirestore is supposed to handle data from DB, my guess is that is should accept only full document.

If I update it to accept partial I will have to pass it into CreateInvitationDocument (that does not exists yet)

libs/invitation/src/lib/+state/invitation.service.spec.ts
```diff
@@ -44,15 +45,15 @@ describe('Invitations Test Suite', () => {
   it('Formats invitation from firestore', () => {
     const is = TestBed.inject(InvitationService);
     is.formatFromFirestore = jest.fn();
-    is.formatFromFirestore({} as any);
+    is.formatFromFirestore({ ...createInvitation(), date: new firebase.firestore.Timestamp(new Date().getTime(), 1) });
```
GrandSchtroumpf, line 48: ah no. You are adding firebase in the build here. Better leave it with the casting than doing that.

GrandSchtroumpf (approved):

PR #5842, opened 2021-05-21
Fix Linter error for testing lib

xmano:
Fix  es-lint errors 

Remove ts-lint rules


libs/testing/src/lib/firebase/functions.ts
```diff
@@ -29,7 +29,9 @@ export function initFunctionsTestMock(emulator = true, overrideConfig?: AppOptio
   try {
     // tslint:disable-next-line: no-eval
     runtimeConfig = eval('require')(join(process.cwd(), './.runtimeconfig.json'));
-  } catch (e) { }
+  } catch (e) { 
+    console.log(e);
```
GrandSchtroumpf, line 33: Better use the logger of firebase. But leave it like that for now.

GrandSchtroumpf (approved):

PR #5844, opened 2021-05-21
financiers linter

RemcoSimonides:
2 any types are still used which I couldn't remove.

### DIFF
--- a/apps/financiers/financiers/src/polyfills.ts
+++ b/apps/financiers/financiers/src/polyfills.ts
@@ -69,21 +69,13 @@ import '@angular/localize/init';
  /***************************************************************************************************
  * ALGOLIA CONFIG
  */
-(window as any).process = {
-  env: { DEBUG: undefined },
-};
-
-/***************************************************************************************************
-* ETHERS
-*/
-(window as any).global = window;
-
+window.process.env = { DEBUG: undefined };
 
 /***************************************************************************************************
 * allSettled
 */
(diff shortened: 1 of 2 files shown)


apps/financiers/financiers/src/polyfills.ts
```diff
@@ -69,21 +69,13 @@ import '@angular/localize/init';
...
  * ALGOLIA CONFIG
  */
-(window as any).process = {
-  env: { DEBUG: undefined },
-};
-
-/***************************************************************************************************
-* ETHERS
-*/
-(window as any).global = window;
```
RemcoSimonides, line 79: As Ethers lib is removed I figured I can delete this now

apps/financiers/financiers/src/polyfills.ts
```diff
@@ -69,21 +69,13 @@ import '@angular/localize/init';
...
-
+window.process.env = { DEBUG: undefined };
 
 /***************************************************************************************************
 * allSettled
 */
-if (!(Promise as any).allSettled) {
-  (Promise as any).allSettled = (promises: Promise<any>[]) =>
+if (!Promise.allSettled) {
+  (Promise as any).allSettled = (promises: Promise<unknown>[]) =>
```
GrandSchtroumpf, line 78: You can remove the as any here too.

GrandSchtroumpf (approved):

PR #5874, opened 2021-05-26
added new calendar tests

phpgeekfr:


libs/contract/src/lib/avails/tests/bucketForm.spec.ts
```diff
@@ -288,3 +288,247 @@ describe('Test BucketForm behaviors', () => {
...
+      },
+      // Set from form
+      exclusive: true,
+      territories: ['france', 'germany'],
+      medias: ['theatrical']
+    };
+
+    bucketForm.addDuration(availDetails, markerA);
+    const index = bucketForm.getTermIndexForCalendar(availDetails, markerB);
+    expect(index).toBeUndefined;
```
GrandSchtroumpf, line 531: `toBeUndefined` shouldn't be a method ?

GrandSchtroumpf (approved):

libs/contract/src/lib/avails/tests/bucketForm.spec.ts
phpgeekfr: It is a method. I added `()` but this also work without

Issue #5186, opened 2021-03-11
Sendgrid email updates


Jackseed:
A lot has been done within Epic #4046  but there is still some we can do to improve our systems. 

### Standardised objects
One good practice that has been implemented with `events` and `app` objects is to send the whole object standardised so we may change texts without sending new variables and avoid wrong variable name in the emails. 
As an example, `event` is sent with parameters `type`, `start`, `end`, `viewUrl`, `sessionUrl`, `title`. 

We could do the with orgs and users. So far parameters used are:
Org
- name
- contact email
- orgId

User
- first name
- last name
- email
- password

### Mailto
For emails where a second user is mentioned (eg. "Hello XX, **YYY** would like to join your org") we'd like to add his/her email address as a mailTo since we're a marketplace that connects the business industry it looks right to give users a way for direct contact

Jackseed:
@Mussuranae I updated the mailTo needs ;)

Mussuranae:
Can you list all email templates that need the org object, those that need the user object (all ?) and those that need the Mailto object ? I will begin to change some data in the code, but it would be good to have a double check to not forget some template. Thanks.

PR #5876, opened 2021-05-26
Org object email

Mussuranae:
Add a new org object for email templates.
Part of #5186

### DIFF
--- a/apps/backend-functions/src/internals/invitations/organizations.ts
+++ b/apps/backend-functions/src/internals/invitations/organizations.ts
@@ -12,6 +12,7 @@ import { getAdminIds, getDocument, getOrgAppKey, createPublicOrganizationDocumen
 import { wasAccepted, wasDeclined, wasCreated } from './utils';
 import { unsubscribeGroupIds } from '../../templates/ids';
 import { applicationUrl } from '@blockframes/utils/apps';
+import { getOrgEmailData } from '@blockframes/utils/emails/utils';
 
 async function addUserToOrg(userId: string, organizationId: string) {
   const db = admin.firestore();
@@ -153,7 +154,8 @@ async function onRequestFromUserToJoinOrgAccept({
   await addUserToOrg(fromUser.uid, toOrg.id);
   const app = await getOrgAppKey(toOrg.id);
   const urlToUse = applicationUrl[app];
-  const template = userJoinedAnOrganization(fromUser.email, urlToUse, toOrg.denomination.full, fromUser.firstName);
+  const org = getOrgEmailData(toOrg)
+  const template = userJoinedAnOrganization(fromUser.email, urlToUse, org, fromUser.firstName);
   return sendMailFromTemplate(template, app, unsubscribeGroupIds.allExceptCriticals);
 }
 
--- a/apps/backend-functions/src/internals/invitations/organizations.ts
+++ b/apps/backend-functions/src/internals/invitations/organizations.ts
@@ -154,7 +154,7 @@ async function onRequestFromUserToJoinOrgAccept({
   await addUserToOrg(fromUser.uid, toOrg.id);
   const app = await getOrgAppKey(toOrg.id);
   const urlToUse = applicationUrl[app];
-  const org = getOrgEmailData(toOrg)
+  const org = getOrgEmailData(toOrg);
   const template = userJoinedAnOrganization(fromUser.email, urlToUse, org, fromUser.firstName);
   return sendMailFromTemplate(template, app, unsubscribeGroupIds.allExceptCriticals);
 }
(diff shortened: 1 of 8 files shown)


libs/utils/src/lib/emails/utils.ts
```diff
@@ -90,3 +97,11 @@ export function getEventEmailData(event?: Partial<EventDocument<EventMeta>>): Ev
     sessionUrl: event?.id ? `/c/o/marketplace/event/${event.id}/session` : ''
   }
 }
+
+export function getOrgEmailData(org: Partial<OrganizationDocument>) {
```
phpgeekfr, line 101: missing return `OrgEmailData`

apps/backend-functions/src/users.ts
```diff
@@ -221,6 +222,11 @@ export const sendUserMail = async (data: { subject: string, message: string, app
 export const createUser = async (data: { email: string, orgName: string, app: App }, context: CallableContext): Promise<PublicUser> => {
   const { email, orgName, app } = data;
 
+  const orgRef = await db.collection('orgs').where('denomination.full', '==', orgName).get();
+  const orgDoc = orgRef.docs;
+  const org = orgDoc[0].data()
+  const orgEmailData = getOrgEmailData(org);
+
```
phpgeekfr, line 229: Sometimes, in import for example, the user is created before the organization CF `libs/import/src/lib/components/table-extracted-elements/organizations/organizations.component.ts` => `addOrganization`.

So this will fail because the org does not exists yet.

Could be cool to be able to pass orgId directly to createUser when available and to have a fallback when we only have the orgName (because it does not exists yet)

apps/backend-functions/src/templates/mail.ts
```diff
@@ -146,34 +146,35 @@ export function invitationToJoinOrgDeclined(admin: PublicUser, user: PublicUser)
...
     userLastName: userRemoved.lastName,
     userEmail: userRemoved.email,
-    orgName: org.denomination.full,
+    org: org,
     pageURL: `${ORG_HOME}${org.id}/view/members`
   };
   return { to: admin.email, templateId: templateIds.org.memberRemoved, data };
 }
 
+// TODO this one needs a small refactoring. It will be done mainly with the user email object creation.
```
phpgeekfr, line 170: can add an issue comment to this todo ?

apps/backend-functions/src/templates/mail.ts
```diff
@@ -146,34 +146,35 @@ export function invitationToJoinOrgDeclined(admin: PublicUser, user: PublicUser)
...
 /** Send email to org admin to inform him that an user has left his org */
-export function userLeftYourOrganization (admin: PublicUser, userRemoved: PublicUser, org: PublicOrganization): EmailTemplateRequest {
+export function userLeftYourOrganization (admin: PublicUser, userRemoved: PublicUser, org: OrgEmailData): EmailTemplateRequest {
   const data = {
     adminFirstName: admin.firstName,
     userFirstName: userRemoved.firstName,
     userLastName: userRemoved.lastName,
     userEmail: userRemoved.email,
-    orgName: org.denomination.full,
+    org: org,
```
phpgeekfr, line 164: `org,` directly

apps/backend-functions/src/internals/invitations/organizations.ts
```diff
@@ -12,6 +12,7 @@ import { getAdminIds, getDocument, getOrgAppKey, createPublicOrganizationDocumen
 import { wasAccepted, wasDeclined, wasCreated } from './utils';
 import { unsubscribeGroupIds } from '../../templates/ids';
 import { applicationUrl } from '@blockframes/utils/apps';
+import { getOrgEmailData } from '@blockframes/utils/emails/utils';
```
phpgeekfr, line 15: is it possible to use `publicOrg` instead of  `OrgEmailData` ? Like this we don't have another model that could be mistaken with orgDocument. You did it like this because you did not wanted to have `org.denomination.public` for example ?

apps/backend-functions/src/internals/invitations/organizations.ts
```diff
@@ -153,7 +154,8 @@ async function onRequestFromUserToJoinOrgAccept({
   await addUserToOrg(fromUser.uid, toOrg.id);
   const app = await getOrgAppKey(toOrg.id);
   const urlToUse = applicationUrl[app];
-  const template = userJoinedAnOrganization(fromUser.email, urlToUse, toOrg.denomination.full, fromUser.firstName);
+  const org = getOrgEmailData(toOrg)
```
phpgeekfr, line 157: missing `;`

phpgeekfr (changes requested):

apps/backend-functions/src/internals/invitations/organizations.ts
Mussuranae: Yes, there is some different data between those two interface. We haven't the logo and the activity on the `OrgEmailData,` but there are on `PublicOrganization`, etc... And yes, it is easier for the denomination to just put `org.denomination`.

PR #5877, opened 2021-05-26
Reset on clear

pldespaigne:
related to #5706

### DIFF
--- a/libs/contract/src/lib/avails/calendar/calendar.component.ts
+++ b/libs/contract/src/lib/avails/calendar/calendar.component.ts
@@ -34,7 +34,7 @@ export class AvailsCalendarComponent implements OnInit {
 
   private _licensedMarkers: DurationMarker[] = [];
   private _soldMarkers: DurationMarker[] = [];
-  /** Includes available, sold, selected, and inselection markers */
+  /** Includes available, sold, selected, and in selection markers */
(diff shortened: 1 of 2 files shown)


libs/contract/src/lib/avails/calendar/calendar.component.ts
```diff
@@ -57,6 +57,9 @@ export class AvailsCalendarComponent implements OnInit {
   }
 
   updateMatrix() {
+    const highlightedRange = this.rows.map(() => this.columns.map(() => false));
```
GrandSchtroumpf, line 60: Would be a good idea to create a method to get a Matrix filled with a default value. To avoid repeating this line several time.

libs/contract/src/lib/avails/calendar/calendar.component.ts
```diff
@@ -57,6 +57,9 @@ export class AvailsCalendarComponent implements OnInit {
...
+    this.state$.next(createAvailCalendarState({ highlightedRange }));
```
GrandSchtroumpf, line 61: Can you create a intermediary variable ?
```typescript
const goodName = createAvailCalendarState({ highlightedRange });
this.sate$.next(goodName);
```

GrandSchtroumpf (approved):

PR #5882, opened 2021-05-27
fix edit avail btn

phpgeekfr:
image


apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
```diff
@@ -134,11 +128,12 @@ export class MarketplaceMovieAvailsComponent implements OnDestroy {
     const mode = this.router.url.split('/').pop();
 
     if (mode === 'map') {
+      this.bucketForm.patchValue({}); // Force observable to reload
       this.avails.mapForm.setValue({ exclusive, duration, medias, territories: [] });
     }
 
     if (mode === 'calendar') {
-      this.avails.calendarForm.setValue({ exclusive, medias, territories });
+      this.avails.calendarForm.setValue({ exclusive, medias, territories, duration });
```
RemcoSimonides, line 136: Are you sure this isn't causing side-effects? 😅

RemcoSimonides: In some functions we check if duration has value in availsForm. If not, then we don't have to validate these values. If yes, then we do validate them. For calendar we don't use them in availsForm I think.

phpgeekfr: It fixes an error displayed in snackbar, for sure ;) because the form is expecting some value for duration
I did not saw any side-effect.

apps/catalog/catalog/src/app/marketplace/title/avails/calendar/avails-calendar.component.html
```diff
@@ -1,5 +1,5 @@
 
-<aside>
+<aside id="avails">
```
RemcoSimonides, line 2: thanks for this one :)

apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
phpgeekfr: I can remove it if you want to let more informed dev to tackle this ;)

phpgeekfr: or put :

```
duration: { from: '', to :'' } 
```
?

RemcoSimonides: I would go for the latter yes. Just to be sure

RemcoSimonides (approved):

PR #5896, opened 2021-05-31
User email object

Mussuranae:
Created a new user email object to get constant data about the user to who we send an email.
About the notion page as well
https://www.notion.so/cascade8/Email-Data-Object-8ed9d64e8cd4490ea7bc0e469c04043e

Second part of #5186

### DIFF
--- a/apps/backend-functions/src/templates/mail.ts
+++ b/apps/backend-functions/src/templates/mail.ts
@@ -1,12 +1,14 @@
 /**
  * Templates for transactional emails we send to user and to cascade8 admins.
+ * Look at the notion page to know objects relative to emails
+ * https://www.notion.so/cascade8/Email-Data-Object-8ed9d64e8cd4490ea7bc0e469c04043e
  */
 import { supportEmails, appUrl } from '../environments/environment';
 import { EmailRequest, EmailTemplateRequest } from '../internals/email';
 import { templateIds } from './ids';
 import { RequestToJoinOrganization, RequestDemoInformations, OrganizationDocument, PublicOrganization } from '../data/types';
 import { PublicUser, User } from '@blockframes/user/+state/user.firestore';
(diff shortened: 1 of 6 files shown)


libs/utils/src/lib/emails/utils.ts
```diff
@@ -42,6 +43,20 @@ export interface OrgEmailData {
   id: string
 }
 
+/**
+ * This interface is used mainly for the user that will received its email.
```
phpgeekfr, line 47: that will receive the mail ?

phpgeekfr (approved):

apps/backend-functions/src/templates/mail.ts
```diff
@@ -290,30 +289,30 @@ export function requestToAttendEventFromUserRefused(
...
-export function movieAcceptedEmail(toUser: PublicUser, movieTitle: string, movieUrl: string): EmailTemplateRequest {
-  const data = { userFirstName: toUser.firstName, movieTitle, movieUrl };
+export function movieAcceptedEmail(toUser: UserEmailData, movieTitle: string, movieUrl: string): EmailTemplateRequest {
+  const data = { user: toUser, movieTitle, movieUrl };
   return { to: toUser.email, templateId: templateIds.movie.accepted, data };
 }
 
-/** */
-export function offerCreatedConfirmationEmail(org: OrganizationDocument, bucket: Bucket, user: User): EmailTemplateRequest {
+/** ???? */
```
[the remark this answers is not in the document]
Mussuranae: I love it too :D

libs/utils/src/lib/emails/utils.ts
Mussuranae: The user to who we sent the email. I have to find an other way to describe it... :/

PR #5918, opened 2021-05-31
UserSubject email object

Mussuranae:
Last step of #5186 
Introduce the `userSubject` object in some templates. It is the same object as the `userEmailData` used for a second user that is sometimes the subject of the email.
I'm not sure about the variable name, maybe it's better `subUser` or something like that ? I'm open to better idea :D
Example :
```
Hi user.firstName,
userSubject.firstName would like to join your org.
```

Need #5896 to be merged before and do some test.


phpgeekfr (approved):
LGTM

PR #5920, opened 2021-06-01
Reset password page

### LINKED ISSUE #5916: Change password reset url
Currently password reset page has `auth/email-verification` url. 
Instead it should be `auth/reset-password`

Mussuranae:
Renamed reset password component and change password component.
Changed route for those component.
Fix #5916


libs/user/src/lib/auth/auth.module.ts
```diff
@@ -49,9 +49,9 @@ export const AuthRoutes: Routes = [
     loadChildren: () => import('./pages/identity/identity.module').then(m => m.IdentityModule)
   },
   // @TODO (#2875) rename to password-reset
```
Mussuranae, line 51: I can't find the issue :/

GrandSchtroumpf: You can remove the comment then :)

libs/user/src/lib/auth/pages/change-password/change-password.component.scss
```diff
@@ -1,8 +1,10 @@
 @import '~@angular/flex-layout/mq';
 
 section {
+  background-color: var(--background-card);
   padding-top: 50px;
   padding-bottom: 50px;
+  height: 100vh;
```
GrandSchtroumpf, line 7: It's usually not a good idea to use vh except for overlay (dialog, ...). If it's a page, it's better to use % of the parent. Also considering smaller screen, what should happen when content overflow ? If you already checked, ignore this comment.

libs/user/src/lib/auth/pages/reset-password/reset-password.component.scss
```diff
@@ -1,8 +1,10 @@
 @import '~@angular/flex-layout/mq';
 
 section {
+  background-color: var(--background-card);
   padding-top: 50px;
   padding-bottom: 50px;
+  height: 100vh;
```
GrandSchtroumpf, line 7: Same as above.

GrandSchtroumpf (approved):

PR #5928, opened 2021-06-02
Fix for release 3 2

phpgeekfr:


apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.scss
```diff
@@ -4,6 +4,10 @@ p {
   line-height: 24px;
 }
 
+a {
```
GrandSchtroumpf, line 7: remove this to avoid override

apps/catalog/catalog/src/app/marketplace/title/view/view.component.ts
```diff
@@ -49,9 +51,10 @@ export class MarketplaceMovieViewComponent implements OnInit {
   }
 
   navigateToAvails(titleId: string) {
```
GrandSchtroumpf, line 53: Where is this method used ? If it's in the HTML, it should be a routerLink

apps/festival/festival/src/app/marketplace/event/list/list.component.scss
```diff
@@ -14,4 +14,10 @@ mat-divider {
   width: 100%;
   display: flex;
   justify-content: center;
-}
\ No newline at end of file
+}
+
+div {
```
GrandSchtroumpf, line 19: it's not a good practice to target a div directly as it might create css issues in the future. Prefer setting a class to the img instead.

GrandSchtroumpf (approved):

apps/catalog/catalog/src/app/marketplace/title/view/view.component.ts
phpgeekfr: Looks like that is is used nowhere

Issue #5906, opened 2021-05-31
Excel import: orgs, movies


RemcoSimonides:

xmano:
Found issue with import movie by admin it is not reflected in Catalog App. Use Movie.xlsx in fixtures.

Movies imported successfully.
![image](https://user-images.githubusercontent.com/8731418/120602542-c115b400-c468-11eb-9e9e-801f9bffb348.png)

But can't search in library  and there create-bucket will fail (user: Camilla)

![image](https://user-images.githubusercontent.com/8731418/120602738-eefaf880-c468-11eb-9661-bafe0f22ee61.png)


Exists in release 3.2 branch and also in develop

RemcoSimonides:
Tested on Safari 14.1
- [ ] Replace with the good illustration in svg (probably a problem of webp not working on safari here but the current illustration is obsolete and needs to be updated with a new one anyway. Ask it to the UI team)

<img width="589" alt="Capture d’écran 2021-06-02 à 16 53 10" src="https://user-images.githubusercontent.com/66258851/120503053-4050ac80-c3c3-11eb-980f-f06cc85e91ce.png">

phpgeekfr:
> Found issue with import movie by admin it is not reflected in Catalog App. Use Movie.xlsx in fixtures.
> 
> Movies imported successfully.
> ![image](https://user-images.githubusercontent.com/8731418/120602542-c115b400-c468-11eb-9e9e-801f9bffb348.png)
> 
> But can't search in library and there create-bucket will fail (user: Camilla)
> 
> ![image](https://user-images.githubusercontent.com/8731418/120602738-eefaf880-c468-11eb-9661-bafe0f22ee61.png)
> 
> Exists in release 3.2 branch and also in develop

This is due to the fact that you did not used latest template.  The latest also had a lot of errors that will be corrected in next PR

phpgeekfr:
>  'Need help' button should be a mat button (neither filled nor stroke), in foreground-800 and on click should open the intercom

For the `and on click should open the intercom` part, this is working bug user need to accept cookies before

Jackseed:
Movie import works fine, 
- [ ] I modified a bit the column order to be same as movie form:
    - Logline from BI to AF
    - Social responsibility goals BJ to AX
    - Target Audience BK to AW
Unfortunately it moves all the other columns after BJ as well...

These templates also have an updated "Content type" list, without Documentary, to go along with #5887 
[...]

PR #5946, opened 2021-06-03
Feat 5906 - Excel import: orgs, movies #5906

phpgeekfr:


libs/import/src/lib/components/import-spreadsheet/import-spreadsheet.component.html
```diff
@@ -30,7 +30,7 @@ <h1>{{ pageTitle }} </h1>
 
   <section class="help-section" fxLayout="column" fxLayoutAlign="center center">
     <p>Once you have filled it out, please upload it below.</p>
-    <a [routerLink]="['./']" (click)="openIntercom()"><strong>Need help ?</strong></a>
+    <button mat-button [routerLink]="['./']" (click)="openIntercom()"><strong>Need help ?</strong></button>
```
GrandSchtroumpf, line 33: change [routerLink]="['./']" to `routerLink="."`. It should work the same.

GrandSchtroumpf (approved):

libs/import/src/lib/components/import-spreadsheet/import-spreadsheet.component.html
phpgeekfr: In fact, since it is now  a `button` and not a `a` I can remove routerLink

Issue #5908, opened 2021-05-31
Landings


RemcoSimonides:

RemcoSimonides:
Now we've got this screen after sending a request. 
![Снимок экрана 2021-06-02 в 14.03.57.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/4ae91b51-026f-4794-bcc5-8f2f112aca63)

- [ ] We can go for using different states for this button: 

1. Send Request ![Снимок экрана 2021-06-02 в 14.06.07.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/8e9ddd41-2276-49dc-a9c5-08c9d6f5cf34)
2. Sending Request... ![Снимок экрана 2021-06-02 в 14.06.07.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/676cd1bb-1796-45bc-8fd6-112ac35ec0fd)
3. Request Sent ![Снимок экрана 2021-06-02 в 14.11.03.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/102fdb74-c50b-4044-acf4-77954ee411ef)

PR #5947, opened 2021-06-03
5908 fixes for release 3.2 (Send Request button)

SamirTouira:
issue: https://github.com/blockframes/blockframes/issues/5908#issuecomment-853749072


libs/landing/src/lib/shell/shell.component.ts
```diff
@@ -41,6 +41,7 @@ export class LandingFooterComponent { }
 export class LandingShellComponent {
   public submitted = false;
   public appName = getAppName(getCurrentApp(this.routerQuery));
+  public buttonText = 'Send Request'
```
Mussuranae, line 44: miss a `;`

libs/landing/src/lib/shell/shell.component.ts
```diff
@@ -70,15 +71,22 @@ export class LandingShellComponent {
 
   /** Triggers when a user click on the button from LearnMoreComponent.  */
   public sendRequest(form: FormGroup) {
+
     if (form.invalid) {
       this.snackBar.open('Please fill the required informations.', 'close', { duration: 2000 });
+      this.buttonText = 'Send Request';
```
Mussuranae, line 77: Do you really need this one as you put a default text for the button ?

libs/landing/src/lib/shell/shell.component.html
```diff
@@ -86,7 +86,7 @@ <h3>Schedule a demo with our team!</h3>
             <tel-input [form]="form.get('phoneNumber')" placeholder="Phone number"></tel-input>
           </mat-form-field>
         </article>
-        <button [disabled]="submitted" fxFlexAlign="center" mat-flat-button color="accent" type="submit">Send Request</button>
+        <button [disabled]="submitted" fxFlexAlign="center" mat-flat-button color="accent" type="submit">{{buttonText}}</button>
```
RemcoSimonides, line 89: Add space `{{ buttonText }}`

libs/landing/src/lib/shell/shell.component.ts
RemcoSimonides, line 44: `;` at the end of the line

libs/landing/src/lib/shell/shell.component.ts
```diff
@@ -70,15 +71,22 @@ export class LandingShellComponent {
...
       return;
     }
+
     try {
+      if (form.status === "VALID"){
+        this.buttonText = 'Sending Request...';
+      }
```
RemcoSimonides, line 82: It happens that status is not valid here?
Line 75 you check for `form.invalid` already

Mussuranae: Agree with Remco, I'm not sure you need this check. And I think you can also write `if (form.valid)` (for a next time ;) )

libs/landing/src/lib/shell/shell.component.ts
SamirTouira: Indeed, no need finally, thanks 👍

libs/landing/src/lib/shell/shell.component.ts
SamirTouira: Alright, thank you!

SamirTouira:
I changed that 👍

Mussuranae (approved):
good job !

Issue #5910, opened 2021-05-31
General layout: header, footer, navbar,...


RemcoSimonides:

phpgeekfr:
There is tiny shift on the "Schedule a demo" button depending on the app we are on:

![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/2d38f702-6dee-4463-a857-52e2c6c16dbd)
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/40b44afa-5c29-4c0b-bfcc-ca8811f08978)
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/534bb7a3-c223-4eca-9f2f-8f5c6a2fb68e)

RemcoSimonides:
You have a good eye for details 🤣 I don't think this needs 'fixing' since users to have all three apps open at the same time and then care about this tiny UI difference

SamirTouira:
Haha if an user find this tiny difference I let him my place and the beer is on me

RemcoSimonides:
App logo, Footer & filled or stroke buttons
### Localization
- App name: Market
- Page: https://staging.archipelmarket.com/c/o/marketplace/home
- Url: https://staging.archipelmarket.com/c/o/dashboard/files

### Fix
- [ ] Logo (tag element app logo) should be 48px height instead of 60px.

<img width="502" alt="Capture d’écran 2021-06-02 à 11 05 05" src="https://user-images.githubusercontent.com/66258851/120452875-d3252300-c392-11eb-9c25-35a844467c82.png">
<img width="395" alt="Capture d’écran 2021-06-02 à 11 06 49" src="https://user-images.githubusercontent.com/66258851/120452889-d7514080-c392-11eb-995f-f5af7054fc05.png">

- [ ] Buttons need to be fixed sizes not fluid (just like before the material updates)
related doc for integration here

<img width="490" alt="Capture d’écran 2021-06-02 à 11 09 00" src="https://user-images.githubusercontent.com/66258851/120453257-2b5c2500-c393-11eb-9698-4f1b9d3efed0.png">
<img width="789" alt="Capture d’écran 2021-06-02 à 11 09 10" src="https://user-images.githubusercontent.com/66258851/120453276-30b96f80-c393-11eb-9597-5886149e2742.png">

- [ ] Footer. Reduce padding from 40px (top and bottom) to 24px.
[...]

Issue #5666, opened 2021-04-29
404 page


Jackseed:
Let's get crazy =)

RemcoSimonides:
Hee @jdrns , illustrations looked good and is approved by Vincent. Can you add the link and move this issue to 'Done'? :)

jdrns:
@RemcoSimonides, the illustration is available [here](https://drive.google.com/drive/folders/1vSEd5bHX7GpbHFi-GJ8xo3AwKhKDSdEH?usp=sharing)
Preview of what it should look like with text (not sure if the background will be white or grey though but both are fine) 
<img width="629" alt="Screenshot 2021-05-25 at 10 10 38" src="https://user-images.githubusercontent.com/79836137/119462845-75189000-bd41-11eb-977f-0f79a37a892d.png">


Do we also need a dark one?

RemcoSimonides:
Thanks! Yea, we will also need one for dark mode ;) The background is transparent, right? So it will change dependent of the theme used.

jdrns:
Backgrounds are transparent yes! 

Both illustrations (light and dark) are available [here](https://zpl.io/bz3GOXX)

SamirTouira:
Should I convert these .png images to .webp or let .png for the moment ?

RemcoSimonides:
Convert them or ask Stephanie to do it ;)

PR #5956, opened 2021-06-03
404 not found reworked + time hours list event rolled back

SamirTouira:
comment: https://github.com/blockframes/blockframes/issues/5910#issuecomment-853833806
original ticket: #5666

### DIFF
--- a/libs/ui/src/lib/error/not-found/not-found.component.html
+++ b/libs/ui/src/lib/error/not-found/not-found.component.html
@@ -1,7 +1,6 @@
 <section bgAsset="background_404.webp">
   <article>
-      <span>Page not found</span>
-      <p>The page you're looking for cannot be</p>
-      <p> found.</p>
+    <h3 class="mat-caption">Page not found</h3>
+    <p>The page you're looking for cannot be found.</p>
   </article>
 </section>
--- a/libs/ui/src/lib/error/not-found/not-found.component.html
+++ b/libs/ui/src/lib/error/not-found/not-found.component.html
@@ -1,6 +1,6 @@
-<section bgAsset="background_404.webp">
+<section fxLayoutAlign="center" fxFlex bgAsset="background_404.webp">
   <article>
-    <h3 class="mat-caption">Page not found</h3>
+    <h3>Page not found</h3>
     <p>The page you're looking for cannot be found.</p>
   </article>
 </section>
--- a/libs/ui/src/lib/error/not-found/not-found.component.html
+++ b/libs/ui/src/lib/error/not-found/not-found.component.html
@@ -1,10 +1,7 @@
-<section class="background" bgAsset="background_404.webp">
-  <mat-card>
-    <mat-card-header>
-      <mat-card-title>Page not found</mat-card-title>
-    </mat-card-header>
-    <mat-card-content>
-      <p>The page you're looking for cannot be found.</p>
-    </mat-card-content>
-  </mat-card>
+<section bgAsset="background_404.webp">
+  <article>
+      <span>Page not found</span>
+      <p>The page you're looking for cannot be</p>
+      <p> found.</p>
+  </article>
 </section>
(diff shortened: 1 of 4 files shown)


libs/ui/src/lib/error/not-found/not-found.component.html
```diff
@@ -1,10 +1,7 @@
...
-    </mat-card-header>
-    <mat-card-content>
-      <p>The page you're looking for cannot be found.</p>
-    </mat-card-content>
-  </mat-card>
+<section bgAsset="background_404.webp">
+  <article>
+      <span>Page not found</span>
+      <p>The page you're looking for cannot be</p>
+      <p> found.</p>
```
Mussuranae, line 5: indentation

libs/ui/src/lib/error/not-found/not-found.component.scss
```diff
@@ -1,13 +1,25 @@
...
   background-repeat: no-repeat;
+  background-attachment: fixed;
 }
 
-mat-card {
-  max-width: 400px;
+article {
+  max-width: 600px;
   margin-left:auto;
   margin-right:auto;
```
Mussuranae, line 13: `margin: auto` should be enough

libs/ui/src/lib/error/not-found/not-found.component.scss
```diff
@@ -1,13 +1,25 @@
...
+  position: relative;
+  top: 120px;
+  span {
+      margin-bottom: 0;
+      font-size: 20px;
+      color: #666b81;
+      font-weight: bold;
```
Mussuranae, line 20: NEEEVVVEEERRRRRRR !!!! :O
You can use some class for typography, I guess the one you want here is ` mat-h3`the typo corresponding to H3, so try to put h3 instead of span or use the class `mat-subheading-2`

libs/ui/src/lib/error/not-found/not-found.component.html
RemcoSimonides, line 5: Alignment of code too far indented.
You can keep the sentence in one paragraph ;)

Mussuranae (changes requested):

libs/ui/src/lib/error/not-found/not-found.component.scss
```diff
@@ -1,13 +1,25 @@
...
+article {
+  max-width: 600px;
   margin-left:auto;
   margin-right:auto;
+  position: relative;
+  top: 120px;
+  span {
+      margin-bottom: 0;
+      font-size: 20px;
+      color: #666b81;
```
RemcoSimonides, line 19: We have to use a color variable

libs/ui/src/lib/error/not-found/not-found.component.scss
```diff
@@ -1,13 +1,25 @@
...
+      font-weight: bold;
+    }
+  p {
+    margin: 0;
+  }
```
RemcoSimonides, line 24: Remove this one if you change the line back to one paragraph

libs/ui/src/lib/error/not-found/not-found.component.scss
```diff
@@ -1,13 +1,25 @@
...
 }
 
-mat-card {
-  max-width: 400px;
+article {
+  max-width: 600px;
   margin-left:auto;
   margin-right:auto;
+  position: relative;
+  top: 120px;
```
RemcoSimonides, line 15: I think a % would be more smart than pixels here. On small screens this isn't going to look good

GrandSchtroumpf: Normally "top: 120px" should have no effect here. It only applies for element with position absolute, fixed or sticky. Position static & relative shouldn't be impacted by the top property.
You set "position: relative" when you want an element inside to have an `absolute` position `relative` to this parent.

libs/ui/src/lib/error/not-found/not-found.component.scss
```diff
@@ -1,13 +1,31 @@
...
 
-mat-card {
+article {
   max-width: 400px;
-  margin-left:auto;
-  margin-right:auto;
+  margin: auto;
+  position: sticky;
+  top: 5%;
+  right: 50%;
```
Mussuranae, line 17: You need it with the `margin: auto` ?

Mussuranae (approved):

Issue #5900, opened 2021-05-31
Avails: Create, Delete, View


RemcoSimonides:
* Avails filter
* Add to selection button
* Cards on selection page
* World map view
* Calendar view

RemcoSimonides:
I selected this movie : "CONTACT WITH LEGENDARY FISHES: Nile Perch, Egypt" by Charles Vallee
Then I clicked inside form fields without filling them, which put the fields in their error state.

- [ ] Alignement is destroyed because there is no invisible hint in the neighbor fields (hints should always be here even if not visible for horizontal alignements)
- [ ] Error color should also be on the icons of the fields.
- [ ] the error text should be "mandatory" instead of date "should be in the future". check the right text with Marie.
- [ ] "Exclusif" -> the "c" is missing.
- [ ] "Media" is also mandatory and should also take the error color state.
<img width="1405" alt="Capture d’écran 2021-06-02 à 15 17 16" src="https://user-images.githubusercontent.com/66258851/120491287-6a04d600-c3b9-11eb-9866-60b9c372aaf7.png">

RemcoSimonides:
- [ ] Weird transition behaviour when clicking on Search and Buy Rights
![g](https://user-images.githubusercontent.com/79836137/120481891-e3e49180-c3b0-11eb-8548-1a13dd57c3c6.gif)

RemcoSimonides:
### Calendar
Calendar cells are not all the same width, the width is depending on the content. Can we add a min-width?
![h](https://user-images.githubusercontent.com/79836137/120491863-e0093d00-c3b9-11eb-96ce-0e442c26b8f7.gif)

RemcoSimonides:
### Selection page
- [ ] In the field under please select a currency, there is no placeholder (when not having a bucket doc yet). Either add a placeholder or select a default currency

RemcoSimonides:
### Preselected Rights
- [ ] As Terms and Rights are the part of "Vocabulary" in Capital Letter Documentation, change the text "Search for new Terms and Rights" 
![Снимок экрана 2021-06-02 в 17.15.17.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/12fbcafe-1538-4097-af5f-3df9cffd57f9)

pldespaigne:
> - [ ] Error color should also be on the icons of the fields.
> - [ ] "Media" is also mandatory and should also take the error color state.
> - [ ] Hint says 'Date must be in the future' even though no date is input
[...]

PR #5957, opened 2021-06-04
5900 edit btn on selection page and currency label

RemcoSimonides:
To do's in issue #5900 
- [x] In the field under please select a currency, there is no placeholder (when not having a bucket doc yet). Either add a placeholder or select a default currency
- [x] Edit button should not only lead to Film Page, but should have the same behavior as edit btn in Preselection Rights Card (fill in avails criteria, show selected countries).

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
+++ b/apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
@@ -4,7 +4,7 @@ import { ActivatedRoute, Router } from '@angular/router';
 import { Component, ChangeDetectionStrategy, OnDestroy, AfterViewInit } from '@angular/core';
 
 import { filter, switchMap } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.ts
```diff
@@ -29,6 +29,7 @@ import { ExplanationComponent } from './explanation/explanation.component';
 export class MarketplaceMovieAvailsComponent implements AfterViewInit, OnDestroy {
   private sub: Subscription;
   private fragSub: Subscription;
+  private paramsSub: Subscription;
```
GrandSchtroumpf, line 32: Better to have a sub list and unsubscribe for each : 
```typescript
subs: Subscription[] = [];
....
const paramsSub = ...
this.subs.push(paramsSub);
...
this.subs.forEach(sub => sub.unsubscribe())
```

GrandSchtroumpf (approved):

PR #5960, opened 2021-06-04
Feat 5898 c

phpgeekfr:
fix import/update issue  for #5906

### DIFF
--- a/libs/utils/src/lib/spreadsheet/format.ts
+++ b/libs/utils/src/lib/spreadsheet/format.ts
@@ -359,18 +359,8 @@ export function formatRatings(ratings: { country: string, value: string }[], sta
 }
 
 export function formatAudienceGoals(audience: { goals: string, targets: string }[]) {
-  const goals = [];
-  const targets = [];
-  audience.forEach(a => {
-    if (a.goals) {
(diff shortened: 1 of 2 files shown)


libs/utils/src/lib/spreadsheet/format.ts
```diff
@@ -358,22 +358,20 @@ export function formatRatings(ratings: { country: string, value: string }[], sta
...
 
-export function formatAudienceGoals(audience: {goals: string, targets: string[]}[], state: MovieImportState) {
-  const movieGoals = [];
-  const movieTargets = [];
-  audience.map(a => {
-    let goalKey;
+export function formatAudienceGoals(audience: { goals: string, targets: string }[]) {
+  const goals = [];
+  const targets = [];
+  audience.forEach(a => {
```
GrandSchtroumpf, line 364: This would be easier to read. Except if audience is a huge array it shouldn't change much:
```typescript
const goals = audience.filter(a => a.goals).map(a => getKeyIfExists('socialGoals', a.goals));
const targets = audience.filter(a => a.targets);
return createAudienceGoals({ targets, goals });
```

GrandSchtroumpf (approved):

SamirTouira (approved):

Issue #5902, opened 2021-05-31
Screenings: CRUD, Invitations, Privacy, Stats, View, Session, Banners


RemcoSimonides:
* Screening form (find someone by the orgname)
* Screening lobby + session (modale when leaving; + check that modale when log out)

SamirTouira:
- [ ] I saw that the filter of organizations in the selection list is working well, but when initializing the component and we click on an organization to filter the first time, there is this error 'id':

![image](https://user-images.githubusercontent.com/52285231/120490861-0d092000-c3b9-11eb-9d89-af46277e02b1.png)

Same thing here but with an other text error:

![image](https://user-images.githubusercontent.com/52285231/120490936-1db99600-c3b9-11eb-9d80-b275ae089be9.png)

if this one is untriggerable, ignore this for the moment it is maybe from my environment

SamirTouira:
- [ ] I discovered a strange issue when you are the owner of an event. Steps to reproduce this:

- Create a screening event
- Go to this screening session
- Click on your profile on top right and click on Log out

That will return to the home page, with the profile logo changed. If you click on this, strangely it will redirect you to the home page:

![image](https://user-images.githubusercontent.com/52285231/120506495-43996780-c3c6-11eb-8717-7f54bd100c44.png)

RemcoSimonides:
## Your environment
OS: MacOS Big Sur
Browser : Safari 14.0.3 (16610.4.3.1.7)

## Review item
Screening Page leaving Modal

### Localization
- App name: Archipel Market
- Page: Screening Page
- Url: https://staging.archipelmarket.com/c/o/marketplace/event/zkjQ1kCJ09QNO0KW88Sn/session
- User used (if anonymized): 
dev+trevor-kcu@blockframes.io 

### Context

- Need to homogenize the modal windows everywhere following [Modals documentation.](https://www.notion.so/cascade8/Modals-warnings-0c8d7ccd3bab4c1c904da5c4b8aa48e4)
- Following the [Modals Text Inventory ](https://docs.google.com/spreadsheets/d/1N06uryQf5J7g56kyenJgpZ_HtiXmYb_Evf798E4fwsY/edit#gid=678850512) change the text

### Fix
- [ ] add close button 
- [ ] title "You are about to leave this screening."
- [ ] main text "You might not be able to come back as its access is time-limited."
- [ ] main CTA "Leave anyway" should be above secondary CTA "Stay"

### Images
This modal 
[...]

PR #5968, opened 2021-06-04
fixed background issue in dark theme + svg replacement (empty illustration)

SamirTouira:
comment: https://github.com/blockframes/blockframes/issues/5902#issuecomment-853946403

I just have to review some details with Clélia about .svg files before merging this


Mussuranae (commented):
Best is to get only SVG images, check pages where they are used to see if the size of the image is correct

Mussuranae (approved):
Nice !

PR #5986, opened 2021-06-08
#6005 org, movie & contract delete : update bucket

### LINKED ISSUE #6005: Deleting a title from CRM app does not remove a bucket list entry if created for that movie.
Deleting a title from CRM app does not remove a bucket list entry if created for that movie.

image

To recreate:
1. Create a new title from AM & check it is listed fine.
2. Import contract for this
3. Create a bucket for this movie. Optional add the movie to wishlist
4. Delete the title in CRM App
5. Observe that wishlist is removed but bucket still shows notification and entry for the movie.

(Same issue for organization removal)

_Originally posted by @xmano in https://github.com/blockframes/blockframes/issues/5897#issuecomment-856713724_

phpgeekfr:
fix #6005


phpgeekfr:
Not mandatory for the release and to much test to perform. This will be pushed to develop once the release is done

phpgeekfr:
Also need to look at new collections `offers` and `incomes`

apps/backend-functions/src/contracts.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  // Delete incomes documents, if any
+  const incomesCollectionRef = db.collection('incomes').where('contractId', '==', contract.id);
+  const incomesSnap = await incomesCollectionRef.get();
+  for (const income of incomesSnap.docs) {
+    await income.ref.delete();
+  }
+
+  console.log(`Contract ${contract.id} removed`);
```
GrandSchtroumpf, line 27: Do you want this to stay ?

apps/backend-functions/src/movie.ts
```diff
@@ -73,15 +74,23 @@ export async function onMovieDelete(
...
   }
 
-  // Update contracts
-  const contracts = await db.collection('contracts').where('titleIds', 'array-contains', movie.id).get();
-  contracts.docs.forEach(c => {
-    const contract = c.data();
-    if (contract.lastVersion?.titles[movie.id]) {
-      delete contract.lastVersion.titles[movie.id];
+  // Delete contracts
+  const contractsCollectionRef = await db.collection('contracts').where('titleId', '==', movie.id).get();
```
GrandSchtroumpf, line 78: Looks like you should do a transaction if you read data... if it's not critic you can leave it as batch.

apps/backend-functions/src/terms.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export async function onTermDelete(termSnapshot: FirebaseFirestore.DocumentSnapshot<Term>) {
+
+  const term = termSnapshot.data() as Term;
+
+  // Delete contracts that have this term as parent
+  const contractsCollectionRef = db.collection('contracts').where('parentTermId', '==', term.id);
+  const contractsSnap = await contractsCollectionRef.get();
+  for (const contract of contractsSnap.docs) {
+    await contract.ref.delete();
```
GrandSchtroumpf, line 13: you might want to do that in a transaction also no ?

GrandSchtroumpf (approved):

apps/backend-functions/src/contracts.ts
phpgeekfr: Yes, we should even add this to all onDocumentWrite to know which ID has been updated/deleted ... We don't have enough logs for digging when we encounter a weird issue

Issue #5897, opened 2021-05-31
Movie: CRUD, List View, Wishlist


RemcoSimonides:
* Movie list
* Movie view + Financial + Campaign view (movie view tabs)
* Movie filters > test on MF for specific filter
* Delete: CRM

RemcoSimonides:
## Movie View
- [ ] Ask a new illustration with the right colors to the design team (those are old wrong colors)
<img width="315" alt="Capture d’écran 2021-06-02 à 11 25 13" src="https://user-images.githubusercontent.com/66258851/120456283-bdfdc380-c395-11eb-881b-3be9427f37ce.png">

## Movie list
When filtering by Languages & Versions
 - [ ] Add more margin-right between inline checkboxes
12px
 - [ ] Add more margin-bottom between inline checkboxes
12px
_(the effect is more obvious when clicking inside the checkbox, the ripple effect crosses on close elements)_

<img width="604" alt="Capture d’écran 2021-06-02 à 11 56 06" src="https://user-images.githubusercontent.com/66258851/120461164-0fa84d00-c39a-11eb-971e-b317e01f6f5e.png">
<img width="488" alt="Capture d’écran 2021-06-02 à 12 05 06" src="https://user-images.githubusercontent.com/66258851/120463013-c22cdf80-c39b-11eb-8584-5e1574f40976.png">

RemcoSimonides:
Bug on Safari:
- [ ] Like on Chrome or mozilla, scroll should be disabled when a filter is open
![f](https://user-images.githubusercontent.com/79836137/120482072-0ecee580-c3b1-11eb-8fc3-137405e2e837.gif)

Jackseed:
Bug on Chrome
https://staging.archipelmarket.com/c/o/dashboard/home
user dev+armando-nwl@blockframes.io

- [ ] Not sure it's linked to movie but it looks like there is an issue with movie stats loading on dashboard homepage, reloading all the time:

![Kapture 2021-06-03 at 11 22 11](https://user-images.githubusercontent.com/14964443/120622028-bce49900-c45e-11eb-8140-379f93b50c4e.gif)

RemcoSimonides:
### Summary Page
- [ ] Add info about Production Status in Reference 
- [ ] Make Production Status mandatory 
[...]

PR #5988, opened 2021-06-08
input autocomplete for territories

Mussuranae:
Related to this [comment](https://github.com/blockframes/blockframes/issues/5897#issuecomment-855789315)


libs/utils/src/lib/static-model/static-model.ts
```diff
@@ -471,6 +471,7 @@ export const festival = {
   venice: 'Venice International Film Festival',
 } as const
 
+// Please keep it in an alphabetic order !
```
xmano, line 474: Alphabetic order of country values?

Also would it be good to group Cypress & Cypress, Northern together? Easier to access

xmano (approved):
Approved with comments.

Check about autocomplete with others.. for me it works good.

libs/utils/src/lib/static-model/static-model.ts
Mussuranae: Yes, Anastasiia agrees with you, so I change it again. I will ask Vincent and Marie for the alphabetic order. But I guess we change this list only when they ask us to do so. It should not often be a problem.

PR #5989, opened 2021-06-08
Remove offers & deals

SamirTouira:
comment: https://github.com/blockframes/blockframes/issues/5910#issuecomment-853982296

I need a confirmation if all things related to Offers & Deals should be removed or only the tab in the sidenav.


Mussuranae (changes requested):
You did too much ^^ Remove only the tab in the sidenav of dashboard and marketplace on catalog ;)

Mussuranae (approved):

PR #5994, opened 2021-06-09
Movie release fixes

Mussuranae:
Status at the end of the table on the Dashboard title list  page
Comment [here](https://github.com/blockframes/blockframes/issues/5897#issuecomment-853643385)

### DIFF
--- a/libs/ui/src/lib/list/table-filter/table-filter.component.ts
+++ b/libs/ui/src/lib/list/table-filter/table-filter.component.ts
@@ -19,7 +19,7 @@ import { MatTableDataSource } from '@angular/material/table';
 import { MatPaginator } from '@angular/material/paginator';
 import { MatSort } from '@angular/material/sort';
 import { Observable } from 'rxjs';
-import { map, startWith, tap } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


libs/ui/src/lib/list/table-filter/table-filter.component.ts
```diff
@@ -19,7 +19,7 @@ import { MatTableDataSource } from '@angular/material/table';
 import { MatPaginator } from '@angular/material/paginator';
 import { MatSort } from '@angular/material/sort';
 import { Observable } from 'rxjs';
-import { map, startWith } from 'rxjs/operators';
+import { map, startWith, tap } from 'rxjs/operators';
```
phpgeekfr, line 22: Looks like that `tap` is not used here

phpgeekfr (approved):
LGTM

libs/ui/src/lib/list/table-filter/table-filter.component.ts
Mussuranae: Yep, you're right, I used it to console log things ^^

PR #5999, opened 2021-06-09
fixed(image directive):corrected image height flicker

iPR0D1G1:
Correct bug of image size flicker on load.


libs/media/src/lib/image/directives/image.directive.ts
```diff
@@ -37,6 +38,22 @@ export class ImageDirective implements OnInit, OnDestroy {
     this.ref$.next(file);
   }
 
+  /** Set the aspect ratio of the image */
+  @Input() set size(_size: 'poster' | 'banner' | 'avatar') {
```
GrandSchtroumpf, line 42: Prefer `ratio` to avoid typo with `sizes` which is a native attribute of the `` tag.
Also the "_" before "size" is not required in the params as it will not be a shadow name.

libs/media/src/lib/image/directives/image.directive.ts
```diff
@@ -24,6 +24,7 @@ export class ImageDirective implements OnInit, OnDestroy {
   private ref$ = new BehaviorSubject<StorageFile>(undefined);
 
   @HostBinding('srcset') srcset: string;
+  @HostBinding('style') style: string;
```
GrandSchtroumpf, line 27: Better to set the `@HostBinding('width')` and `@HostBinding('height')` because Chromium and Firefox wil automaticall apply the `aspect-ratio: auto (width) / (height)` based on theses attributes:
```typescript
@HostBinding('width') width: number;
@HostBinding('height') height: number;
```

libs/media/src/lib/image/directives/image.directive.ts
```diff
@@ -37,6 +38,22 @@ export class ImageDirective implements OnInit, OnDestroy {
     this.ref$.next(file);
   }
 
+  /** Set the aspect ratio of the image */
+  @Input() set size(_size: 'poster' | 'banner' | 'avatar') {
+    let value = '3/4';
```
GrandSchtroumpf, line 43: Better to create a mapping outside of the directive: 
```typescript
const ratios = {
  poster: { w: 3, h: 4 },
  banner: { w: 16, h: 9 },
  avatar: { w: 1, h: 1 },
};
....
@Input() set ratio(ratio: 'poster' | 'banner' | 'avatar') {
  const { w, h } = ratios[ratio];
  this.width = w;
  this.height = h;
}
```

libs/movie/src/lib/movie/components/card/card.component.html
```diff
@@ -1,6 +1,6 @@
 <figure (mouseenter)="hasOverlay = true" (mouseleave)="hasOverlay = false" matRipple [matRippleRadius]="50">
   <ng-container *ngIf="(ltLg$ | async) === false  else forTouchScreen">
-    <img [ref]="movie | movieImage: size | storageFile: 'movies':size:movie.id" [asset]="placeholderAsset" alt="Movie Illustration">
+    <img [ref]="movie | movieImage: size | storageFile: 'movies':size:movie.id" [size]="size" [asset]="placeholderAsset" alt="Movie Illustration" >
```
GrandSchtroumpf, line 3: I think you'll need to set `height: auto` on the img css when applying with width & height attribute.

GrandSchtroumpf (changes requested):

libs/media/src/lib/image/directives/image.directive.ts
GrandSchtroumpf: Quick correction, the `aspect-ratio` is generated only on Chromium with the arguments. Firefox doesn't apply it automatically: 
image

libs/media/src/lib/image/directives/image.directive.ts
```diff
@@ -38,19 +40,20 @@ export class ImageDirective implements OnInit, OnDestroy {
...
-  /** Set the aspect ratio of the image */
-  @Input() set size(_size: 'poster' | 'banner' | 'avatar') {
-    let value = '3/4';
-    if (_size === 'banner') {
-      value = '16/9';
-    } else if (_size === 'avatar') {
-      value = '1/1';
-    }
-    // If this.style is undefined, the appended text will be `undefined aspect-ratio...`
+  aspectRatios = {
```
GrandSchtroumpf, line 43: Put that outside of the class, there is no reason to keep it inside the class as it's the same for every img.

libs/media/src/lib/image/directives/image.directive.ts
GrandSchtroumpf: following the comment just above we need to set the aspect ratio. Prefer creating a specific hostbinding, instead of updating the whole style. It would avoid duplication of string in the style
```typescript
@HostBinding('style.aspectRatio') aspectRatio?: string;
```
Using string concatenation is not a good practice because each time there is a change in the input you'll push the new aspect-ratio instead of overwriting it.

libs/media/src/lib/image/directives/image.directive.ts
```diff
@@ -38,19 +40,20 @@ export class ImageDirective implements OnInit, OnDestroy {
...
+    avatar: { w: 1, h: 1 },
+  };
+  @Input() set aspectRatio(ratio: 'poster' | 'banner' | 'avatar') {
+    const { w, h } = this.aspectRatios[ratio];
+    this.width = w;
+    this.height = h;
+    /** Set the aspect ratio of the image */
     if (this.style) {
-      this.style += ` aspect-ratio: ${value};`;
+      this.style += ` aspect-ration: ${w}/${h}; height: auto;`
```
GrandSchtroumpf, line 54: This is dangerous to put the height: auto here as we might need a specific height in px sometime, and local style will always be stronger that local css. Following my comment above: 
```typescript
@Input() set ratio(ratio: 'poster' | 'banner' | 'avatar') {
  const { w, h } = ratios[ratio];
  this.width = w;
  this.height = h;
  this.aspectRatio = `auto ${w} / ${h}`;
}
```

GrandSchtroumpf (changes requested):

phpgeekfr:
@iPR0D1G1  please don't merge into this branch. 
- If it is supposed to be a hotfix, create a hotfix branch from production.
- If not, please wait until  https://github.com/blockframes/blockframes/pull/6004 is merged so you can change target to develop

PR #6003, opened 2021-06-09
General layout #5910

SamirTouira:
issue: #5910
linked to: https://github.com/blockframes/blockframes/issues/5910#issuecomment-856757745 & https://github.com/blockframes/blockframes/issues/5910#issuecomment-856823719

Everything is working well except one thing about the "escape" key. When we close the component in fullscreen with the escape key, the fullScreen variable doesn't change to false, then the button for leave the fullscreen is bugged because it doesn't recognize that `this.document.exitFullscreen`. The button design stays like 'exit fullscreen' even with the small size
I didn't succeded to manage this

I have also a remark about the fullscreen files in safari (or I hope it happens only in BrowserStack), once we clicked on the fullscreen mode, the button to revert the fullscreen is "hidden" due to the screen too big, so what should we do if it is the case?


phpgeekfr:
@SamirTouira please don't merge into this branch. 
- If it is supposed to be a hotfix, create a hotfix branch from production.
- If not, please wait until  https://github.com/blockframes/blockframes/pull/6004 is merged so you can change target to develop

SamirTouira:
> @SamirTouira please don't merge into this branch.
> 
> * If it is supposed to be a hotfix, create a hotfix branch from production.
> * If not, please wait until  #6004 is merged so you can change target to develop

Sure, no worries, thanks for your explanation. I prepared this in advance and I have to finish some stuff before merging this

PR #6015, opened 2021-06-09
Fix error in bucket test due to change in movie import template.

xmano:
Modified templates according to `libs\import\src\lib\components\view-extracted-elements\movies\view-extracted-movies.component.ts`

image

!! Bucket test passing
image

Fixed one unrelated issue due to pre-filled date in avails form:
image


xmano:
I will increase some timeouts to improve the reliability of the tests

phpgeekfr (approved):
LGTM

PR #6016, opened 2021-06-09
avail bug fix for prod

phpgeekfr:


GrandSchtroumpf (approved):
I would rather use the ActivatedRoute params instead of the Movie query to limite the dependency to Akita

Issue #5985, opened 2021-06-08
Improve Avail Search Criteria errors


RemcoSimonides:
- [x] Error color should also be on the icons of the fields.
- [x] "Media" is also mandatory and should also take the error color state.
- [x] Hint says 'Date must be in the future' even though no date is input
<img width="1405" alt="Capture d’écran 2021-06-02 à 15 17 16" src="https://user-images.githubusercontent.com/66258851/120491287-6a04d600-c3b9-11eb-9866-60b9c372aaf7.png">

pldespaigne:
This is not really related to the avails, it's more related to the re-usable input components we designed.
We should update them to solve the listed issues, but we should keep in mind that those component are probably used elswhere in the app and we must ensure that the style stays consistent

RemcoSimonides:
I think "Error color should also be on the icons of the fields." is the only one that applies to all input fields with an icon, but the other ones are specific for avails criteria search, right?

pldespaigne:
"Media" is a `static-group` component and doesn't handle really well validation of the parent form, so it needs refactoring, which might impact everywhere we use static-groups

GrandSchtroumpf:
The `static-group` will require a huge rework to support the status update. We need to move forward with Archipel Content CRM, therefore I ask PL to postpone the work on this issue for now.

yspringfield:
As it stands, it shows all the various error states. Upon testing, I noticed *all* the fields no longer show an error on reset clicked. Don't know if that's intentional. Just mentioning for what it's worth.

PR #6026, opened 2021-06-10
improve style

pldespaigne:
solve some items of #5985

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
+++ b/apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
@@ -2,6 +2,8 @@
   <section fxLayoutGap="12px" fxFlex>
     <static-group fxFlex formControlName="medias" scope="medias" displayAll="All Rights" required placeholder="Specify right(s) to filter" [withoutValues]="['theatrical']" test-id="medias" icon="play_arrow">
       <mat-label>Media</mat-label>
+
+      <span groupError *ngIf="availsForm.hasError('required', 'medias')">This field is mandatory</span>
     </static-group>
 
     <mat-form-field fxFlex appearance="outline">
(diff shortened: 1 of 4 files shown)


libs/ui/src/lib/static-autocomplete/group/group.component.html
```diff
@@ -3,9 +3,9 @@
 		<mat-form-field appearance="outline">
       <mat-icon *ngIf="!!icon" matPrefix [svgIcon]="icon"></mat-icon>
 			<mat-label>
-				<ng-content select="mat-label"></ng-content>{{ required ? ' *' : '' }}
```
RemcoSimonides, line 6: If you remove `{{ required ? ' *' : '' }}` here then you should add the * in the `<mat-label>Media *</mat-label>` and all other static-group places where the field is mandatory

apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
```diff
@@ -2,6 +2,8 @@
   <section fxLayoutGap="12px" fxFlex>
     <static-group fxFlex formControlName="medias" scope="medias" displayAll="All Rights" required placeholder="Specify right(s) to filter" [withoutValues]="['theatrical']" test-id="medias" icon="play_arrow">
       <mat-label>Media</mat-label>
+
```
RemcoSimonides, line 5: Remove empty line :)

RemcoSimonides (approved):
Nice!

libs/ui/src/lib/static-autocomplete/group/group.component.html
pldespaigne: This is not needed because I added `[required]="required"` on the mat-form-field which automatically add a `*` next to the label if needed.

PR #6027, opened 2021-06-10
fix production status

### LINKED ISSUE #5529: observable issue in MovieShellConfig
observable issue in MovieShellConfig.

Check for `#5529` in code.

Created from https://github.com/blockframes/blockframes/pull/5513#discussion_r608702898

Mussuranae:
Fix prod status bug
Close #5529


phpgeekfr (approved):
A self solved issue ... ?
@Jackseed we will need to test this to be sure we are is not putting any old bug back

@Mussuranae can you also close #5529 ?

Mussuranae:
> A self solved issue ... ?
> @Jackseed we will need to test this to be sure we are is not putting any old bug back

Looks like yes :/
I agree for testing. I tested for the production status on catalog and festival, and it was all working but I didn't test much more except coming back to the movie view and navigate  on different movie view on dashboard.

Issue #5984, opened 2021-06-08
[CRM] Archipel Content


GrandSchtroumpf:
[Zeplin](https://app.zeplin.io/project/6075614f83ab0350cbea888a)

## Pages

---
## IceBox (Do Not implement)
Contract List
- Query the list of contracts of an offer: `contractService.valueChanges(ref => ref.where('offerId', '==', offerId))`.
- Group everything related to the title (name, id, poster) in one column and use a pipe to query it in the template (@2amarkina approved).
- Use `<bf-table-filter>` (`libs\ui\src\lib\list\table-filter\table-filter.component.ts`) for the table.

---

# Already Done
## Pages
- [x] Contract Form (#6082) (@jeremie74)
[zeplin](https://app.zeplin.io/project/6075614f83ab0350cbea888a/screen/6081beb10b9148145de61fd6)

- [x] Offer View Page (#6072) **(@iPR0D1G1 )**
[zeplin](https://app.zeplin.io/project/6075614f83ab0350cbea888a/screen/6081be7f042d9a1c0d4856f3)

- [x] Shell Component (#6080)
- [x] Offer List page (#6070) **(@iPR0D1G1 )**
[zeplin](https://app.zeplin.io/project/6075614f83ab0350cbea888a/screen/6081be794300b91e99bf6ad4)
Use `<bf-table-filter>` (`libs\ui\src\lib\list\table-filter\table-filter.component.ts`) for the table.
- [x] Contract/Title View (#6081)
[zeplin](https://app.zeplin.io/project/6075614f83ab0350cbea888a/screen/6081be816b999f14b7d03718)

[Dialog Validation on delete](https://app.zeplin.io/project/6075614f83ab0350cbea888a/screen/6081be848d30082175f65245): 
- open dialog when clicking on delete action button
- [x] Offer View
  - use a `<ng-template>` in the template of the offer view instead of creating another component
- [x] Contract View
### Offer

- [x] Update the Offer interface in `libs\contract\src\lib\offer\+state\offer.model.ts` to have `buyerUserId` which should be userId that triggered the contract.
- [x] Update `libs\contract\src\lib\bucket\+state\bucket.service.ts` to add `specificity`, `buyerId` & `buyerUserId` also in the offer when we create it.

### Contract
- [x] Update the Offer interface in `libs\contract\src\lib\bucket\+state\bucket.model.ts` to have `buyerUserId` which should be userId that triggered the contract.
[...]

PR #6035, opened 2021-06-11
CRM Prework

pldespaigne:
resolve some items of the CRM pre-work defined in #5984

### DIFF
--- a/libs/contract/src/lib/contract/+state/contract.model.ts
+++ b/libs/contract/src/lib/contract/+state/contract.model.ts
@@ -11,10 +11,6 @@ export interface Contract {
   offerId?: string;
   /** The id of the buyer's org, can be undefined if external sale */
   buyerId?: string;
-  /** The user id of the buyer, can be undefined if external sale */
-  buyerUserId?: string;
-  /** User generated text description */
-  specificity?: string;
   /** Id of the direct seller. AC in the Archipel Content app */
   sellerId: string;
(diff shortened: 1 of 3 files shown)


libs/contract/src/lib/contract/+state/contract.model.ts
```diff
@@ -9,8 +9,12 @@ export interface Contract {
   termIds: string[];
   /** Offer in which the contract is included is any */
   offerId?: string;
-  /** The id of the buyer, can be undefined if external sale */
+  /** The id of the buyer's org, can be undefined if external sale */
   buyerId?: string;
+  /** The user id of the buyer, can be undefined if external sale */
+  buyerUserId?: string;
+  /** User generated text description */
+  specificity?: string;
```
GrandSchtroumpf, line 17: Should be only on the sale

libs/contract/src/lib/contract/+state/contract.model.ts
```diff
@@ -36,6 +40,8 @@ export function createMandate(params: Partial<Mandate> = {}): Mandate {
     termIds: [],
     parentTermId: '',
     buyerId: '', // For external sales this is undefined
+    buyerUserId: '', // For external sales this is undefined
```
GrandSchtroumpf, line 43: also update the on the createSale.

GrandSchtroumpf (approved):

PR #6052, opened 2021-06-16
Mano e2e with emulator demo

xmano:
Cleaned up configs for e2e with emulator.

Note this is tested with manual setup of emulator and serve running in different terminal tabs


libs/e2e/src/lib/utils/screenings.ts
```diff
@@ -4,7 +4,7 @@ export const EVENTS: Partial<Screening>[] = [
   {
     event: 'TKOTL Private Screening',
     by : {
-      uid: '2OJUZoWtTVcew27YDZa8FQQdg5q2'
+      uid: '7i57eOi2t1Zlik9HUoub7KvJqfg3'
```
phpgeekfr, line 7: this is some uid you created manually ?

libs/e2e/src/lib/fixtures/users.ts
```diff
@@ -4,7 +4,8 @@ import userFixture from 'tools/fixtures/users.json';
 
 export enum USER {
   Vincent = 'MDnN2GlVUeadIVJbzTToQQNAMWZ2',
-  Jean = '2OJUZoWtTVcew27YDZa8FQQdg5q2',
+  Jean1 = '2OJUZoWtTVcew27YDZa8FQQdg5q2',
```
phpgeekfr, line 7: this is some uid you created manually ? 
Can you add a comment explaining the situation?

libs/e2e/src/lib/utils/screenings.ts
xmano: Since auth is not working, we are unable to login with same email id associated with the fixture uid.

So I had to create an auth user manually and use that UID.

libs/e2e/src/lib/fixtures/users.ts
xmano: Sure will add.

However our strategy might change depending on use of auth emulator or we might choose app commands (Cypress plugins), Dynamic fixtures, etc.. 

So this code is going to change.

george43g:
I think we should close this as I implemented most of these changes a while ago

Issue #6041, opened 2021-06-14
Extract contract-item


pldespaigne:
Extract the contract table on the terms from the
selection page (`apps\catalog\catalog\src\app\marketplace\title\selection\selection.component.html`)
into a `<contract-item>` component inside the `libs/contracts/lib/contracts/item` folder.

It should look like that:

```html
<contract-item [contract]="contract">
  <contract-price>
    <!-- Can be a form or display -->
  </contract-price>
  <contract-actions>
    <!-- Action buttons for the contract -->
  </contact-actions>
  <ng-template termAction let-term let-index="index">
    <!-- Action buttons to display in the action cell per term -->
  </ng-template>
</contract-item>
```

iPR0D1G1:
Please, can you give me the url to a page wherein this selection component is used?

PR #6054, opened 2021-06-17
Contract item

iPR0D1G1:
This PR extract the contract-item component, according to issue #6041


pldespaigne (changes requested):
Please also rename the `contract-item` folder simply into `item`. since it's already in a `contract folder`.

The path should now be like that `libs/contract/src/lib/contract/item/contrac-item.component.html`

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.html
```diff
@@ -17,71 +17,51 @@ <h1>My Selection</h1>
...
-            <ng-template colRef="exclusive" let-exclusive>
-              {{ exclusive ? 'Yes' : 'No' }}
-            </ng-template>
-            <ng-template colAction="action" label="Actions" let-termIndex="index">
-              <a mat-icon-button [routerLink]="['/c/o/marketplace/title', contract.titleId, 'avails']" [queryParams]="{ contract: i, term: termIndex }" matTooltip="Edit">
+    <ng-container *ngIf="(bucket$ |async) as bucket">
+      <ng-container *ngFor="let contract of bucket.contracts; trackBy: trackById; let i = index">
+        <contract-item
+          [contract]="contract"
+          [columns]="columns"
```
GrandSchtroumpf, line 30: Columns are the same in every contract-item modulo the action column. Better would be to add the action column only if there is a `colActionTemplate` inside the contract-item.
I think you can do something like that : 
```typescript
actionTemplate?: TemplateRef<unknown>;
@ContentChild('colActionTemplate') set colActionsTemplate(template: TemplateRef<unknown>) {
  if (template) {
    this.column.push('action');
    this.actionTemplate = template;
    this.cdr.marckForCheck();
  }
}
```

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.html
```diff
@@ -17,71 +17,51 @@ <h1>My Selection</h1>
...
-            <button mat-button (click)="removeContract(i, title)" test-id="delete-contract">
-              Delete all rights for this title
-            </button>
-            <a mat-flat-button [routerLink]="['../title', title.id, 'avails']" color="primary">
-              Add Rights
-            </a>
-          </footer>
-        </section>
+            <contract-footer>
+              <footer fxLayout fxLayoutAlign="space-between" class="footer">
```
GrandSchtroumpf, line 51: As PL pointed out, I think it's better to do `<footer contractFooter>` to avoid adding unecessary nodes.

libs/contract/src/lib/contract/contract-item/contrac-item.component.html
```diff
@@ -0,0 +1,42 @@
+
```
GrandSchtroumpf, line 1: Please format

libs/contract/src/lib/contract/contract-item/contrac-item.component.html
```diff
@@ -0,0 +1,42 @@
...
+                </p>
+              </section>
+            </ng-template>
+            <ng-template colRef="medias" let-medias>
+              {{ medias | toLabel:'medias' }}
+            </ng-template>
+            <ng-template colRef="exclusive" let-exclusive>
+              {{ exclusive ? 'Yes' : 'No' }}
+            </ng-template>
+            <ng-template  *ngIf="showColActionsTemplate" colAction="action" label="Actions" let-termIndex="index">
```
GrandSchtroumpf, line 35: Use `*ngIf="colActionsTemplate"` instead of creating another `showColActionsTemplate` variable.

libs/contract/src/lib/contract/contract-item/contrac-item.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  @ContentChild('priceTemplate') priceTemplate: TemplateRef<unknown>;
+  @ContentChild('colActionTemplate') colActionsTemplate: TemplateRef<unknown>;
+  @Input() columns: Record<string, string>;
+  @Input() initialColumns: string[];
+  @Output() openDetails = new EventEmitter<{ terms: string, scope: Scope }>();
+
+  trackById = (i: number, doc: { id: string }) => doc.id;
+
+  get showColActionsTemplate() {
+    return Boolean(this.colActionsTemplate);
```
GrandSchtroumpf, line 40: In the repo we usually cast into a boolean with "!!" instead of "Boolean()"

libs/contract/src/lib/contract/contract-item/contrac-item.component.ts
```diff
@@ -0,0 +1,47 @@
+import { Component, Input, Output, EventEmitter, TemplateRef, ContentChild } from '@angular/core';
+import { BucketContract } from '@blockframes/contract/bucket/+state/bucket.model';
+import { Scope } from '@blockframes/utils/static-model';
+
+@Component({
```
GrandSchtroumpf, line 5: For ng-content it's better to create a Directive instead of a full component, it will prevent the compiler to run the template & style AST parsers for this.

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.html
GrandSchtroumpf: As described in the comment above, you shouldn't need to specify the columns at all in the input. It should be a property of the component and should me updated only if there is a colActionTemplate

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.html
```diff
@@ -17,71 +17,45 @@ <h1>My Selection</h1>
...
-          <footer fxLayout fxLayoutAlign="space-between">
-            <button mat-button (click)="removeContract(i, title)" test-id="delete-contract">
-              Delete all rights for this title
-            </button>
-            <a mat-flat-button [routerLink]="['../title', title.id, 'avails']" color="primary">
-              Add Rights
-            </a>
-          </footer>
-        </section>
+            <footer fxLayout fxLayoutAlign="space-between" class="footer" footer>
```
GrandSchtroumpf, line 45: you don't need a directive "footer" here. The tag of the component should be enough. Also no need to provide a class to select it in the css. Better to access it with hierarchy selector: 
```css
contract-item footer { ... }
```

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.scss
```diff
@@ -8,27 +8,9 @@
...
-.contract {
-  padding: 24px;
-  margin-bottom: 72px;
-
-  img {
-    max-width: 150px;
-    max-height: 200px;
-    cursor: pointer;
-  }
+footer.footer {
```
GrandSchtroumpf, line 12: See comment above

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.html
```diff
@@ -17,71 +17,44 @@ <h1>My Selection</h1>
...
-            </ng-template>
-            <ng-template colRef="medias" let-medias>
-              {{ medias | toLabel:'medias' }}
-            </ng-template>
-            <ng-template colRef="exclusive" let-exclusive>
-              {{ exclusive ? 'Yes' : 'No' }}
-            </ng-template>
-            <ng-template colAction="action" label="Actions" let-termIndex="index">
-              <a mat-icon-button [routerLink]="['/c/o/marketplace/title', contract.titleId, 'avails']" [queryParams]="{ contract: i, term: termIndex }" matTooltip="Edit">
+            <ng-template #colActionTemplate label="Actions" let-termIndex="index" let-titleId="titleId">
```
GrandSchtroumpf, line 37: As described in [this issue](https://github.com/blockframes/blockframes/issues/5984) the `colActionTemplate` should have this API : 
```typescript
<ng-template termAction let-term let-index="index">
```
The `let-term` should be $implicit & and the `let-index` should be names "index" and not `termIndex`.
The label here is unused. You should remove it.
You can replace the directive `termAction` with a pseudo id `#termAction` if you prefer as it doesn't require any Input.

You don't need to provide the titleId as you can get it from the contract which is in your scope.

libs/contract/src/lib/contract/item/contract-item.component.html
```diff
@@ -0,0 +1,42 @@
...
+        </section>
+      </ng-template>
+      <ng-template colRef="medias" let-medias>
+        {{ medias | toLabel:'medias' }}
+      </ng-template>
+      <ng-template colRef="exclusive" let-exclusive>
+        {{ exclusive ? 'Yes' : 'No' }}
+      </ng-template>
+      <ng-template *ngIf="actionTemplate" colAction="action" label="Actions" let-termIndex="index">
+        <ng-template [ngTemplateOutlet]="actionTemplate"
```
GrandSchtroumpf, line 35: See comment above above template outlet context

libs/contract/src/lib/contract/item/contract-item.component.ts
```diff
@@ -0,0 +1,47 @@
...
+    duration: 'Terms',
+    territories: 'Territories',
+    medias: 'Rights',
+    exclusive: 'Exclusivity'
+  };
+  actionTemplate?: TemplateRef<unknown>;
+
+  @ContentChild('colActionTemplate') set colActionsTemplate(template: TemplateRef<unknown>) {
+    if (template) {
+      this.initialColumns.push('action');
```
GrandSchtroumpf, line 34: you'll also need to update the columns here.

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.html
```diff
@@ -17,71 +17,44 @@ <h1>My Selection</h1>
...
-                <div fxLayout="column">
-                  <h3>{{ title.directors | displayName }}</h3>
-                  <p>{{ title | movieFeature }}</p>
-                </div>
-              </article>
-            </section>
-            <mat-form-field appearance="outline">
+    <ng-container *ngIf="(bucket$ |async) as bucket">
+      <ng-container *ngFor="let contract of bucket.contracts; trackBy: trackById; let i = index">
+        <contract-item
```
GrandSchtroumpf, line 28: You can format it on one line now :)

libs/contract/src/lib/contract/item/contract-item.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  styleUrls: ['./contract-item.component.scss']
+})
+export class ContractItemComponent {
+
+  constructor(
+    private cdr: ChangeDetectorRef,
+    private dialog: MatDialog,
+  ) { }
+
+  @Input() contract: BucketContract;
```
GrandSchtroumpf, line 21: Put all the properties (contract, priceTemplate, actionTemplate, colActionTemplate, trackId) above the constructor. In most programming language, a class is usually built like this  :
```typescript
class MyClass {
  property1;
  property2;
  constructor() {}
  method1() {}
  method2() {}
}
```

libs/contract/src/lib/contract/item/contract-item.component.ts
```diff
@@ -0,0 +1,47 @@
...
+} from '@angular/core';
+import { BucketContract } from '@blockframes/contract/bucket/+state/bucket.model';
+import { Scope } from '@blockframes/utils/static-model';
+import { MatDialog } from '@angular/material/dialog';
+import { DetailedTermsComponent } from '@blockframes/contract/term/components/detailed/detailed.component';
+
+@Component({
+  selector: 'contract-item',
+  templateUrl: './contract-item.component.html',
+  styleUrls: ['./contract-item.component.scss']
```
GrandSchtroumpf, line 12: always use `ChangeDetectionStrategy.OnPush` for the component

GrandSchtroumpf (changes requested):

apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.html
iPR0D1G1: I just verified and there's nowhere implicit variable term from let-term is used within the ng-template.
can we simply not put it...

apps/catalog/catalog/src/app/marketplace/title/selection/selection.module.ts
```diff
@@ -33,23 +25,15 @@ import { MatTooltipModule } from '@angular/material/tooltip';
     GetTitlePipeModule,
     BucketPipesModule,
     DeepKeyPipeModule,
-    ImageModule,
```
pldespaigne, line 36: This module is probably needed for the images to be shown

pldespaigne (changes requested):

pldespaigne (approved):

Issue #5975, opened 2021-06-07
Confirmation modale - Disable CTA button when form is invalid


Jackseed:
Modale visible on different places:
- MF: movie form summary page when submitting a new movie
- CRM: org & user view, delete button

CTA button should be disabled as long as the text typed is different from the expected text.
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/ea957d58-5ae4-4683-95cb-89c844e431b4)

PR #6074, opened 2021-06-21
Disable CTA button when form is invalid fixed

SamirTouira:
issue: #5975

### DIFF
--- a/libs/admin/src/lib/admin-panel/components/crm-form-dialog/crm-form-dialog.component.html
+++ b/libs/admin/src/lib/admin-panel/components/crm-form-dialog/crm-form-dialog.component.html
@@ -18,6 +18,6 @@
 
   <article fxLayout="row" fxLayoutAlign="space-evenly">
     <button mat-stroked-button (click)="close()">Cancel</button>
-    <button mat-flat-button color="warn" (click)="confirm()">{{ data.confirmButtonText | firstWordTitlecase }}</button>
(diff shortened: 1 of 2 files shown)


libs/admin/src/lib/admin-panel/components/crm-form-dialog/crm-form-dialog.component.html
```diff
@@ -18,6 +18,6 @@ <h3 *ngIf="data.title"> {{ data.title }} </h3>
 
   <article fxLayout="row" fxLayoutAlign="space-evenly">
     <button mat-stroked-button (click)="close()">Cancel</button>
-    <button mat-flat-button color="warn" (click)="confirm()">{{ data.confirmButtonText | firstWordTitlecase }}</button>
+    <button type="submit" [disabled]="data.confirmationWord.toUpperCase() !== actionConfirm.value" mat-flat-button color="warn" (click)="confirm()">{{ data.confirmButtonText | firstWordTitlecase }}</button>
```
Mussuranae, line 21: Maybe you can put it in the TS file and add the `form.invalid` check. It's not super lisible here.
```
const isValid = this.actionConfirm.value === this.data.confirmationWord.toUpperCase() && this.actionConfirm.valid;
```

and in the html :
```
<button [disabled]="!isValid">
```

You will be able to reuse the `isValid` const in the `confirm` function as well ;)

Mussuranae (changes requested):

Mussuranae (approved):
Perfect !

Issue #6070, opened 2021-06-21
[CRM] Offer List


GrandSchtroumpf:
![image](https://user-images.githubusercontent.com/5858657/122736104-f18a8a00-d27f-11eb-90e9-9af0c60b621d.png)
[Zeplin](https://app.zeplin.io/project/6075614f83ab0350cbea888a/screen/6081be794300b91e99bf6ad4)

---

## Pre-work

- [x] In the CRM app, remove the drop-down menu for "Contracts" as it will be used only for one single page, then rename "Contract" into "Offers".

![image](https://user-images.githubusercontent.com/5858657/122734149-209ffc00-d27e-11eb-86dc-2c47a9511172.png)

## (A) Page
- [x] create a new"Offers" page component and its route in the crm app, when you click on the "Offers" button it shoudl redirect you to this new empty page, this is the (A) page/component int the above screenshot
- the page should be `libs\admin\src\lib\admin-panel\pages\offers`
- route are defined in the `admin-panel.module.ts`
---

## (B) List component

Generate component here : `libs\contract\src\lib\offer\list`

- [x] Update the model of Offer (`libs\contract\src\lib\offer\+state\offer.model.ts`) with these new status options: 
```typescript
status: 'pending' | 'negociating' | 'accepted' | 'signing' | 'signed' | 'declined';
```
- [ ] Create a filter like in the title list example (`apps\catalog\catalog\src\app\dashboard\title\list\list.component.html`).
- [ ] Query all offers with their contracts. You can use the `queryChanges` property from "akita-ng-fire": 
```typescript
import { queryChanges } from "akita-ng-fire";
...
queryChanges.call(this.offerService, {
  path: 'offers',
  contracts: (offer: Offer) => this.contractService.valueChanges(ref => ref.where('offerId', '==', offer.id))
})
```
- [ ] Display them in a `<bf-table-filter>` (`libs\ui\src\lib\list\table-filter\table-filter.component.ts`)

_main issue #5984_

PR #6075, opened 2021-06-21
Offer List #6070

iPR0D1G1:
Offer List #6070


libs/admin/src/lib/admin-panel/admin-panel.module.ts
```diff
@@ -57,6 +59,7 @@ export const panelRoutes: Routes = [
   { path: 'organization/:orgId', component: OrganizationComponent },
   { path: 'movie/:movieId', component: MovieComponent },
   { path: 'users', component: UsersComponent },
+  { path: 'offers', component: OffersComponent },
```
GrandSchtroumpf, line 62: Best would be to create a module for this component and lazyload it. I know we don't do that for the other component of the module, but we plan to improve that progressively : 
```typescript
{ path: 'offers', loadChildren: () => import('./pages/offer/list/list.module').then(m => OfferListModule) }
```

libs/admin/src/lib/admin-panel/pages/offers/offers.component.ts
```diff
@@ -0,0 +1,37 @@
...
+@Component({
+  selector: 'offers',
+  templateUrl: './offers.component.html',
+  styleUrls: ['./offers.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class OffersComponent implements OnDestroy {
+
+  subs: Subscription[] = [];
+  offers = queryChanges.call(this.offerService, {
```
GrandSchtroumpf, line 19: We usually put a $ at the end of the properties when it's an observable: `offers$`

libs/admin/src/lib/admin-panel/pages/offers/offers.component.ts
```diff
@@ -0,0 +1,37 @@
...
+    path: 'offers',
+    //@ts-ignore
```
GrandSchtroumpf, line 21: Create a type for that in the same file and cast the query : 
```typescript
type OfferWithContracts = Offer & { contracts: Contract[] };
const queryOffer: Query<OfferWithContracts> = {
  path: 'offers',
  contracts: offer => ({
    path: 'contracts',
    queryFn: ref => ref.where('offerId', '==', offer.id)
  })
}
@Component({ .... })
export class OfferListComponent {
  offers$ = queryChanges.call(this.offerService, queryOffer);
}
```

libs/admin/src/lib/admin-panel/pages/offers/offers.component.ts
```diff
@@ -0,0 +1,37 @@
...
+
+@Component({
+  selector: 'offers',
+  templateUrl: './offers.component.html',
+  styleUrls: ['./offers.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class OffersComponent implements OnDestroy {
+
+  subs: Subscription[] = [];
```
GrandSchtroumpf, line 18: You don't have any subscriptions. Remove it, and the ngOnDestroy

libs/admin/src/lib/admin-panel/pages/offers/offers.module.ts
```diff
@@ -0,0 +1,14 @@
...
+
+@NgModule({
+  declarations: [
+    OffersComponent,
+  ],
+  imports: [
+    CommonModule,
+  ]
+})
+export class OffersModule { }
```
GrandSchtroumpf, line 14: If you import the component directly in the AdminPanelModule you don't need a module for the component.
But following the comment above I would like you to lazy load this component, so you need a module, but you'll need to import every dependancy required in the component (`MatTabsModule`, ...) and a `RouterModule.forChild()`

libs/admin/src/lib/admin/toolbar-top/toolbar-top.component.html
```diff
@@ -21,8 +21,8 @@
   </button>
 
 
-  <button mat-button [matMenuTriggerFor]="contractsMenu">
-    <mat-icon svgIcon="document"></mat-icon><span matLine>Contracts</span>
+  <button mat-button routerLink="/c/o/admin/panel/offers">
```
GrandSchtroumpf, line 24: always use a `a` tag when using `routerLink`. Also you might want to add a routerLinkActive class.

iPR0D1G1: Just verified the design and there's no need for a routerLinkActive class. the design doesn't specifies which link is currently active.

libs/utils/src/lib/pipes/total.pipe.ts
```diff
@@ -0,0 +1,21 @@
...
+    data.forEach(s => something += +getDeepValue(s, path))
+    return something;
+  }
+}
+
+@NgModule({
+  declarations: [TotalPipe],
+  exports: [TotalPipe]
+})
+export class TotalPipeModule { }
```
pldespaigne, line 21: I also created a similar pipe in `/lib/admin-panel/pages/offers/shell/shell.component.ts` (at the bottom of the file)
Could please delete my pipe, and use yours in the `shell.component.html`

pldespaigne (changes requested):

libs/admin/src/lib/admin-panel/pages/offers/list/offer-list.component.scss
```diff
@@ -0,0 +1,27 @@
+nav[mat-tab-nav-bar] button {
+  border: none;
+  background-color: transparent;
+}
```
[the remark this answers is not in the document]
iPR0D1G1: @GrandSchtroumpf  Told me to do this.

libs/utils/src/lib/pipes/join.pipe.ts
```diff
@@ -0,0 +1,15 @@
...
+  transform(data: unknown[], path: string = '', seperator: string = ', ') {
+    return data.map(s => typeof s === 'string' ? s : getDeepValue(s, path)).join(seperator)
+  }
+}
+
+@NgModule({
+  declarations: [JoinPipe],
+  exports: [JoinPipe]
+})
+export class JoinPipeModule { }
```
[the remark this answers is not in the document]
iPR0D1G1: this is intended to be used by @GrandSchtroumpf in the future.

libs/admin/src/lib/admin-panel/pages/offers/list/offer-list.component.ts
```diff
@@ -0,0 +1,83 @@
...
+  ) { }
+
+  /** Dynamic filter of offers for each tab. */
+  applyFilter(filter?: AllOfferStatus) {
+    this.filter.setValue(filter);
+  }
+
+  /* index paramter is unused because it is a default paramter from the filter javascript function */
+  filterByStatus(offer: Offer, index: number, value: AllOfferStatus): boolean {
+    if (value === 'pending') {
```
GrandSchtroumpf, line 71: Prefer switch case here : 
```typescript
switch (value) {
  case 'pending': return offer.status === value;
  case ...
  default: true;
}
```

libs/admin/src/lib/admin/toolbar-top/toolbar-top.component.html
```diff
@@ -21,9 +21,9 @@
   </button>
 
 
-  <button mat-button [matMenuTriggerFor]="contractsMenu">
-    <mat-icon svgIcon="document"></mat-icon><span matLine>Contracts</span>
-  </button>
+  <a mat-button routerLink="/c/o/admin/panel/offers">
```
GrandSchtroumpf, line 24: I think you only need `routerLink="offers"`

libs/contract/src/lib/offer/list/list.component.ts
```diff
@@ -0,0 +1,14 @@
...
+  Component, ChangeDetectionStrategy,
+} from '@angular/core';
+
+@Component({
+  selector: 'offers-list',
+  templateUrl: './list.component.html',
+  styleUrls: ['./list.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ListComponent {
```
GrandSchtroumpf, line 11: Looks like you're not using this component. Can we remove it?

libs/utils/src/lib/pipes/total.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { NgModule, Pipe, PipeTransform } from '@angular/core';
+import { getDeepValue } from './deep-key.pipe';
+
+@Pipe({ name: 'total' })
+export class TotalPipe implements PipeTransform {
+  transform(data: unknown[], path: string,): number {
```
GrandSchtroumpf, line 6: remove trailing "," in the params

GrandSchtroumpf (approved):

libs/admin/src/lib/admin/toolbar-top/toolbar-top.component.html
iPR0D1G1: Bildschirmfoto 2021-06-25 um 11 12 29 AM
We actually need the entire correct url for proper routing.

pldespaigne (approved):

PR #6077, opened 2021-06-21
4815 Users dashboard

RemcoSimonides:
Added dashboard for Users


env/env.blockframes-remco.ts
```diff
@@ -148,7 +148,8 @@ export const bigQueryAnalyticsTable = 'blockframes-remco.analytics_234801912.eve
 // ========
 
 export const datastudio = {
```
phpgeekfr, line 150: you are not updating every env files ?

libs/admin/src/lib/admin-panel/components/dashboard/dashboard.component.html
```diff
@@ -0,0 +1,9 @@
+<ng-container *ngIf="!!dashboardUrl">
+  <section fxLayout fxLayoutAlign="center">
+    <iframe
```
phpgeekfr, line 3: they don't provide anything else than iframe ?

phpgeekfr (approved):
LGTM

env/env.blockframes-remco.ts
RemcoSimonides: No, I only make a dashboard for Staging and Production. I added the staging dashboard to my env.ts so I can see it.

libs/admin/src/lib/admin-panel/components/dashboard/dashboard.component.html
RemcoSimonides: No :/

Issue #6069, opened 2021-06-21
Invitation Code Contact us UI Change


RemcoSimonides:
Small UI update on the sign-up page for the Invitation Code input field. Instead of "Can't find it? Contact Us" we now only have an icon next to the input field.

<img width="690" alt="Screenshot 2021-06-09 at 16 34 35" src="https://user-images.githubusercontent.com/79836137/121374917-959b3980-c940-11eb-882e-b4f3d6f777a5.png">

Mini fab with icon inside. There is no zeplin link yet as the mock-ups are not up to date but let me know if there are any questions. You can find [the icon](https://zpl.io/bLqW01Y) on zeplin though, size is 24px

- [x] On hover, we have a tooltip that says "Contact us to get help"
- [x] On click, it should open the intercom

PR #6079, opened 2021-06-21
Code contact UI change

SamirTouira:
fix: #6069


libs/user/src/lib/auth/pages/identity/identity.component.html
```diff
@@ -83,18 +83,14 @@ <h1>Create your account</h1>
 
         <!-- PASSWORD FIELDS -->
         <ng-container *ngIf="form.get('generatedPassword').enabled">
-          <mat-form-field appearance="outline">
-            <mat-label>Invitation Pass</mat-label>
-            <input matInput type="password" placeholder="Your Invitation Pass" formControlName="generatedPassword"
+          <div fxLayoutAlign="end">
```
GrandSchtroumpf, line 86: I think this div is doing nothing. Try removing this div.

GrandSchtroumpf (approved):

libs/user/src/lib/auth/pages/identity/identity.component.scss
```diff
@@ -47,6 +47,11 @@ section {
       width: 100%;
     }
 
+    .intercom {
+      margin-top: 20px;
+      margin-left: 12px;
```
Mussuranae, line 52: Put 24 and 16px (multiple of 8 always ;) )

libs/user/src/lib/auth/pages/identity/identity.component.html
Mussuranae, line 86: I wonder why do you need the `end` align ? If you want to get only a row, you can put `fxLayout` alone, it is the same as `fxLayout="row"`, also I think you can replace your CSS (margin) by a `fxLayoutGap="16px"`, it will put a space of 16px between elements, and I guess you can play with the fxLayoutAlign a little more. I don't remember we already use the `end` choice ;)

Mussuranae (changes requested):

Mussuranae (approved):

PR #6093, opened 2021-06-22
Currency on bucket test

Mussuranae:


apps/catalog/catalog-e2e/src/integration/marketplace/create-bucket.spec.ts
```diff
@@ -57,6 +57,7 @@ describe('Create a new bucket and finalize a new offer', () => {
     }
     p1.openSidenavMenuAndNavigate('selection');
     assertMoveTo(SELECTION_PATH);
+    cy.wait(3 * SEC);
```
xmano, line 60: Instead of waiting here, the better way is to check in Selection Page if the element is present with a timeout. This will eliminate unwanted wait here.

xmano (changes requested):
Were you able to fix the currency changing issue with these changes?

xmano (approved):
LGTM

PR #6096, opened 2021-06-22
CRM adding: select list of movies

jeremie74:


libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.scss
```diff
@@ -0,0 +1,10 @@
+:host {
+    display: flex;
+    flex-direction: column;
+    max-width: 1135px;
```
pldespaigne, line 4: please use a multiple of 4 for the px

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
```diff
@@ -7,6 +11,12 @@ import { Component, ChangeDetectionStrategy } from '@angular/core';
...
 })
 export class ContractFormComponent {
+  form = new FormGroup({
+    titleId: new FormControl(),
+  })
+  titles$ = this.service.valueChanges(ref => ref.where('app.catalog.status', '==', 'approved'));
 
+  constructor(private service: MovieService){}
 
+  save() {}
```
pldespaigne, line 21: It's weird to have an empty method like that.
If I wasn't aware that you are still working on this page, I would ask you to delete it

pldespaigne (approved):

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
pldespaigne: I just look at the CI logs and this empty method is caught as an error, so will need to remove it to be able to merge

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.scss
jeremie74: Done

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
jeremie74: Done. It was a request from François, i think for the future to be ready.

Issue #6084, opened 2021-06-22
Movie view dashboard: change status, archive, "delete"


Jackseed:
## Description
Dashboard users will now be able to:
- Switch a movie back to draft
- Archive a movie
- "Delete" a movie (= switch active app movie app access to `false`, **not deleting**)

All these actions will be available through a new action list in the movie banner (dashboard side only). 
Tabs will also be added in the dashboard movie list to filter through movie app status. (Already implemented in `catalog` and `financiers` but not `festival`). 

Also, the movie app status will now be available in the banner as a tag, instead of the original title.

## UI 
[New banner](https://zpl.io/2GPrXqr)
[Confirmation modale](https://zpl.io/bPBZwKD)
[Title list tabs](https://zpl.io/2yDNAvw)

### New action list 
This list is linked with the main CTA button on the top right of the banner. The list and main CTA will change depending on:
- movie app status
- current application

[Here](https://docs.google.com/spreadsheets/d/1e4qahYcBxsYJ4Dpuu_RyL_onlgkMRNts4k5kai-Zu20/edit#gid=0) is a recap of different action list status depending on app & movie app status **please respect it (and ask me if not clear)**. 

If only one action is available, the list is replaced with the action button only
![image](https://user-images.githubusercontent.com/14964443/122907453-72ff1c80-d353-11eb-840d-ab509d3a56fd.png)

The possible actions are:
- [x] Submit Title: redirects to the movie form, summary page, anchored to the bottom (CTA Submit)
- [x] Publish on Marketplace: redirects to the movie form, summary page, anchored to the bottom (CTA Publish)
- [x]  Switch back to Draft: switch movie app status to `draft` (opens a snackbar "Title back to Draft." for 4sec)
- [x] Remove from Marketplace: switch movie app status to `draft` (opens a snackbar "Title removed from Marketplace." for 4sec)
- [x] Marketplace Preview: redirects to the marketplace movie view
- [x] Edit Title: redirects to movie form lobby
- [x] Archive Title: switch movie app status to `archived` (opens a snackbar "Title archived." for 4sec)
- [x]  Restore Title: switch movie app status to `draft` (opens a snackbar "Title restored." for 4sec)
- [x] Delete Title: switch movie app access to `false` (opens a modale to validate first, then a snackbar "Title deleted." for 4sec)

### Title list
- [x] Add the new "Archived" tab that show all movies with status `archived`
[...]

PR #6097, opened 2021-06-22
status tabs on festival app + added archived status

phpgeekfr:
part of https://github.com/blockframes/blockframes/issues/6084


apps/festival/festival/src/app/dashboard/title/list/list.component.html
```diff
@@ -13,8 +13,15 @@ <h1>My Titles</h1>
       </div>
     </header>
 
+    <nav mat-tab-nav-bar fxLayout="row" fxLayoutAlign="space-between end">
```
GrandSchtroumpf, line 16: Use header instead of nav. It's not a navigation bar.

apps/financiers/financiers/src/app/dashboard/title/list/list.component.ts
```diff
@@ -11,7 +11,7 @@ import { Intercom } from 'ng-intercom';
 import { appName, getCurrentApp } from '@blockframes/utils/apps';
 import { RouterQuery } from '@datorama/akita-ng-router-store';
 
-type Filters = 'all' | 'draft' | 'ongoing' | 'achieved';
+type Filters = 'all' | 'draft' | 'ongoing' | 'achieved' | 'archived';
```
GrandSchtroumpf, line 14: don't you need to update the Movie interface also ?

GrandSchtroumpf (approved):

apps/financiers/financiers/src/app/dashboard/title/list/list.component.ts
phpgeekfr: New status was added in `libs/utils/src/lib/static-model/static-model.ts`

SamirTouira (approved):

PR #6099, opened 2021-06-22
tag movie app access status added (UI WIP)

SamirTouira:
related to: #6084
Design WIP

### DIFF
--- a/libs/movie/src/lib/movie/dashboard/shell/shell.component.html
+++ b/libs/movie/src/lib/movie/dashboard/shell/shell.component.html
@@ -7,6 +7,9 @@
         <article fxLayout="column" fxFlex>
           <h4>{{ movie | movieFeature }}</h4>
           <h1>{{ movie.title.international }}</h1>
+          <span class="tag">
+            <p>{{ movie.app[appName].status }}</p>
+          </span>
           <h6>{{ movie.directors | displayName }}</h6>
           <p>
             <span>{{ movie.originCountries | toLabel: 'territories'}}</span>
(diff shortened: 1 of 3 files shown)


libs/movie/src/lib/movie/dashboard/shell/shell.component.scss
```diff
@@ -27,6 +27,20 @@ article * {
   min-width: 0;
 }
 
+span.tag {
```
phpgeekfr, line 30: Maybe you can take CSS from zeplin as an example:
image

image

libs/movie/src/lib/movie/dashboard/shell/shell.component.html
```diff
@@ -7,6 +7,9 @@
         <article fxLayout="column" fxFlex>
           <h4>{{ movie | movieFeature }}</h4>
           <h1>{{ movie.title.international }}</h1>
+          <mat-chip class="tag">
+            <p>{{ movie.app[appName].status }}</p>
```
phpgeekfr, line 11: why do you need a `p` tag inside mat-chip ?

phpgeekfr (approved):
LGTM

libs/movie/src/lib/movie/dashboard/shell/shell.component.html
SamirTouira: because it is for configure the text like the color (white) and a margin auto, if I put these 2 properties only in `mat-chip `it won't work even if I remove the margin auto. so if I put the color: white in the mat-chip it won't change the text

PR #6100, opened 2021-06-23
5746 Stop sharing unselects file

RemcoSimonides:
image

### DIFF
--- a/libs/media/src/lib/file/carousel/carousel.component.html
+++ b/libs/media/src/lib/file/carousel/carousel.component.html
@@ -8,7 +8,7 @@
 <ng-container *ngIf="files.length; else noFiles">
   <bf-carousel flex>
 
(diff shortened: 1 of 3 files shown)


libs/media/src/lib/file/carousel/carousel.component.html
```diff
@@ -8,7 +8,7 @@ <h3>Your Files</h3>
 <ng-container *ngIf="files.length; else noFiles">
   <bf-carousel flex>
 
-    <button carouselItem mat-button (click)="selectFile('')" [disabled]="isSelected('')" matTooltip="Stop presenting file">
```
RemcoSimonides, line 11: I didn't have to remove this but I like that the stop sharing doesn't keep it's disabled color but the carousel goes to 'clean' state again

pldespaigne (approved):

PR #6109, opened 2021-06-24
6089 Add email verified to user meta data

RemcoSimonides:
Work in progress!

### DIFF
--- a/libs/user/src/lib/user/+state/user.service.ts
+++ b/libs/user/src/lib/user/+state/user.service.ts
@@ -4,6 +4,7 @@ import { UserState, UserStore } from './user.store';
 import { OrganizationQuery } from '@blockframes/organization/+state/organization.query';
 import { map } from 'rxjs/operators';
 import { User, AuthQuery, AuthStore } from '@blockframes/auth/+state';
+import { DocumentMeta } from '@blockframes/utils/models-meta';
 
 
 @Injectable({ providedIn: 'root' })
@@ -21,6 +22,7 @@ export class UserService extends CollectionService<UserState> {
     private organizationQuery: OrganizationQuery,
   ) {
     super(store);
+    this.updateEmailVerified();
   }
 
   //////////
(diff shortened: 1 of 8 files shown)


libs/user/src/lib/user/+state/user.service.ts
```diff
@@ -106,4 +108,20 @@ export class UserService extends CollectionService<UserState> {
   public changeRank(rank: string) {
     this.authStore.updateProfile({ financing: { rank } });
   }
+
+  // TODO Once we have a custom email verified page, we can update the users' meta there
+  private async updateEmailVerified() {
```
RemcoSimonides, line 113: So this is executed every time the page loads. On the 'Thank you for joining us' page, you'll need to refresh the page in order to check if you've got access to the page. So once you've verified your email, you are obligatory to refresh the page anyway so then this user._meta.emailVerified is updated anyway.

Would be better to have a dedicated page for 'Thank you for verifying your email' instead of the default one from Firebase. Then we can execute this functionality there.

libs/user/src/lib/user/+state/user.service.ts
```diff
@@ -106,4 +108,20 @@ export class UserService extends CollectionService<UserState> {
   public changeRank(rank: string) {
     this.authStore.updateProfile({ financing: { rank } });
   }
+
+  // TODO #6113 nce we have a custom email verified page, we can update the users' meta there
```
phpgeekfr, line 112: `nce` => `once` ?

libs/user/src/lib/user/+state/user.service.ts
phpgeekfr: so once we have this custom email verified page, we will not need to reload page?

phpgeekfr (approved):

libs/user/src/lib/user/+state/user.service.ts
RemcoSimonides: This story about reloading the page just makes sure this function is always called. If the user doesn't have to reload the 'Thank you for joining us page' this code wouldn't trigger because it's only triggered when UserService is constructed.

That's why this custom email verified page would be a better spot to have this code. The user will always go there when he clicks the 'Verify email' button on the email.

Issue #5766, opened 2021-05-11
E2E Test failure on CI


xmano:
Catalog App E2E tests failed to run

> nx run catalog-e2e:e2e:ci 
- Generating browser application bundles...
✔ Browser application bundle generation complete.

Initial Chunk Files                                                                                              | Names                                                                                                         |      Size
vendor.js                                                                                                        | vendor                                                                                                        |   8.70 MB
main.js                                                                                                          | main                                                                                                          |   1.03 MB
polyfills-es5.js                                                                                                 | polyfills-es5                                                                                                 | 570.42 kB
styles.css                                                                                                       | styles                                                                                                        | 466.43 kB
scripts.js                                                                                                       | scripts                                                                                                       | 460.89 kB

[snip]

Build at: 2021-05-11T00:39:39.565Z - Hash: 8ae0b224c5366731aa3c - Time: 78138ms

Warning: /home/circleci/repo/apps/catalog/catalog/src/app/dashboard/right/list/list.component.ts is part of the TypeScript compilation but it's unused.
Add only entry points to the 'files' or 'include' properties in your tsconfig.

** Angular Live Development Server is listening on localhost:4200, open your browser on http://localhost:4200/ **

✔ Compiled successfully.

Here E2E should have started but next app E2E tests are launched.

> nx run festival-e2e:e2e:ci   <============== Festival App E2E is run
- Generating browser application bundles...
✔ Browser application bundle generation complete.

Catalog App compiles fine but E2E tests don't run..

xmano:
Festival E2E did not run fully:

![image](https://user-images.githubusercontent.com/8731418/117934512-60d79a80-b320-11eb-8e58-6b638d84ce62.png)

The test run get interrupted by another app E2E Test.

xmano:
@phpgeekfr I tried with wait workaround in CRM & Financiers E2E test but it is not working all the time. I think we need to investigate at CI job level.

xmano:
Update : 11th June 2021. Issue still persists. Festival E2E did not run.

```
> nx run festival-e2e:e2e:ci 
- Generating browser application bundles...

Initial Chunk Files                                                                                     | Names         |      Size
vendor.js                                                                                               | vendor        |   8.60 MB

:
:

** Angular Live Development Server is listening on localhost:4200, open your browser on http://localhost:4200/ **


✔ Compiled successfully.
✔ Browser application bundle generation complete.

>>>> Here the tests don't run.. CI starts with next App <<<<

[...]

PR #6115, opened 2021-06-24
tests for E2E Test failure on CI #5766

phpgeekfr:


george43g (approved):
I'll refactor/clean this up at some point but suitable for now.

xmano (approved):
LGTM

If there is a way to loop through all the apps by way of some scripting code it will be good. Else we have to do this for every additional app.

PR #6116, opened 2021-06-24
Css movie view dashboard shell improvement - feature 6084 - WIP

SamirTouira:
issue: #6084 

Let me know what do you think about this CSS flow. This PR fix the CSS bug button "Delete Title" exceeded, some typos,  I moved the origin countries and the release year to bottom with a `margin-top: auto;`  to correspond with the zeplin design, I can also do a` margin-top: auto;` to the Actions button instead of the origin countries and the release year but I don't know if that's a good practice for Blockframes


libs/movie/src/lib/movie/dashboard/shell/shell.component.html
```diff
@@ -37,19 +37,19 @@ <h6>{{ movie.directors | displayName }}</h6>
         </ng-template>
 
         <ng-container *ngIf="movie.app[appName].status === 'archived' else actionsMenu">
-          <button mat-icon-button (click)="removeAppAccess()" matTooltip="Delete title">
+          <button class="fit-content" mat-icon-button (click)="removeAppAccess()" matTooltip="Delete Title">
```
GrandSchtroumpf, line 40: I never use fit-content in the app, so I'm surprise you need it here.

libs/movie/src/lib/movie/dashboard/shell/shell.component.scss
```diff
@@ -38,3 +41,14 @@ h1 {
 h6 {
   margin-top: 8px;
 }
+
+aside {
+  a.outlines {
```
GrandSchtroumpf, line 46: You should never override padding of a mat-button. This makes the UI very hard to maintain.

libs/movie/src/lib/movie/dashboard/shell/shell.component.scss
```diff
@@ -38,3 +41,14 @@ h1 {
...
+    margin: 0;
+    padding: 0;
+  }
+  button.fit-content {
+    width:fit-content;
```
GrandSchtroumpf, line 51: Same here, do not override the width of a mat-icon-button, it's fixed to 40px and should remain like that. If we need to reduce the size we should think of a way to do it globally.

GrandSchtroumpf (changes requested):

libs/movie/src/lib/movie/dashboard/shell/shell.component.scss
```diff
@@ -38,3 +41,30 @@ h1 {
...
 }
+
+aside {
+  button.delete-title-button {
+    width:100%;
+    margin-bottom: 8px;
+  }
+}
+
+mat-chip.capitalize-first-letter {
```
GrandSchtroumpf, line 52: Avoid using class when you don't need. Here you can just target the mat-chip without specifying the class.

SamirTouira: Alright, my bad

libs/movie/src/lib/movie/dashboard/shell/shell.component.html
```diff
@@ -37,19 +40,23 @@ <h6>{{ movie.directors | displayName }}</h6>
         </ng-template>
 
         <ng-container *ngIf="movie.app[appName].status === 'archived' else actionsMenu">
-          <button mat-icon-button (click)="removeAppAccess()" matTooltip="Delete title">
+          <button class="delete-title-button" mat-button (click)="removeAppAccess()" matTooltip="Delete Title">
```
GrandSchtroumpf: remove unused class

libs/movie/src/lib/movie/dashboard/shell/shell.component.html
```diff
@@ -37,19 +40,23 @@ <h6>{{ movie.directors | displayName }}</h6>
...
+            <span>Delete Title</span>
           </button>
         </ng-container>
         <ng-template #actionsMenu>
-          <!-- TODO #6084 finnish action menu -->
-          <button mat-button [matMenuTriggerFor]="actionMenuMenu">Actions</button>
+          <!-- TODO #6084 finish action menu -->
+          <button mat-button [matMenuTriggerFor]="actionMenuMenu">
+            <mat-icon class="more-horiz" svgIcon="more_horiz"></mat-icon>
+            Actions
```
GrandSchtroumpf, line 52: If you've got a mat-icon in a mat-button, then the text should always be in a span, because there is a global style that apply 8px margin to span that are before or after mat-icon in a button

libs/movie/src/lib/movie/dashboard/shell/shell.component.scss
```diff
@@ -38,3 +40,14 @@ h1 {
...
   margin-top: 8px;
 }
+
+mat-chip {
+  text-transform: capitalize;
+}
+
+mat-icon{
+  &.flip-icon {
+    -webkit-transform: scaleX(-1);
```
GrandSchtroumpf, line 50: You don't need to specify webkit, sass is going to add it post compilation.

GrandSchtroumpf (approved):

SamirTouira:
I can confirm that these 2 checkboxes do the same thing. When the title is Accepted, if the user click on "Switch back to Draft", the movie is removed from the Marketplace because the Accepted title is switched to Draft. If we add a "Remove from Marketplace" button when the title is Accepted, the title will back to Draft aswell, so it will be removed from the Marketplace. It does exactly the same thing

PR #6128, opened 2021-06-29
fix rule datamodel

pldespaigne:
Incomes data model was updated to remove `contractId` and instead use the contract id as the income id.
However the firestore rules have been updated to reflect this change.

This PR fix that.


phpgeekfr (approved):
lgtm

GrandSchtroumpf (approved):

PR #6130, opened 2021-06-29
Feat 6127 - fiix anon db verified email & disable maintenance screen for next release

phpgeekfr:
Feat 6127 - fiix anon db verified email & disable maintenance screen for next release


george43g (approved):
Seems fine to me, if your goal is to disable maintenance mode. Please note that functions will still be skipped in maintenance mode with these changes. Just the frontend and rules have been disabled but not function skips

phpgeekfr:
> Please note that functions will still be skipped in maintenance mode with these changes. Just the frontend and rules have been disabled but not function skips

@george43g Yes this is intentional, this is to not disturb users watching a screening while we update the app. So from the backend side, no changes at all.

PR #6132, opened 2021-06-29
ContractView confirm on delete

pldespaigne:
Add an input-confirm dialog when deleting a term in the CRM's ContractView.

### DIFF
--- a/libs/admin/src/lib/admin-panel/pages/offers/list/offer-list.component.html
+++ b/libs/admin/src/lib/admin-panel/pages/offers/list/offer-list.component.html
@@ -38,7 +38,8 @@
           {{ specificity ? 'YES' : '-' }}
         </ng-template>
         <ng-template colRef="incomes" let-incomes>
-          {{incomes | total:'price' | currency:incomes[0].currency:'symbol'}}
+          <span *ngIf="incomes.length !== 0">{{ incomes | total: 'price' | currency:incomes[0].currency }}</span>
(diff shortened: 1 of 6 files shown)


libs/admin/src/lib/admin-panel/pages/offers/list/offer-list.component.html
```diff
@@ -38,7 +38,8 @@ <h1>Offers</h1>
           {{ specificity ? 'YES' : '-' }}
         </ng-template>
         <ng-template colRef="incomes" let-incomes>
-          {{incomes | total:'price' | currency:incomes[0].currency:'symbol'}}
+          <span *ngIf="incomes.length !== 0">{{ incomes | total: 'price' | currency:incomes[0].currency }}</span>
```
GrandSchtroumpf, line 41: Best would be that total pipe return 0 if the array is empty. Like that you don't have to change that.

GrandSchtroumpf (approved):

libs/admin/src/lib/admin-panel/pages/offers/list/offer-list.component.html
pldespaigne: it's not a problem about the total pipe but about accessing `incomes[0]` in the currency pipe, normally incomes is never empty but some data is wrong in the pft so that just an extra layer of safety to avoid breaking the page

Issue #6018, opened 2021-06-10
Change Terms of Service and Privacy Policy modal to page


RemcoSimonides:
- [x] Convert privacy policy and terms of service to a page.
- [x] On the sign-up page, when you open the Privacy Policy or Terms of Service, it should open in a new tab instead of modal.
- [x] Check if everything is still working when you navigate to the Privacy Policy or Terms of Service if you click on it from the footer (no need to open the page in a new tab then, just navigate to it).

This should also fix the back navigation on the page since there is no need for these buttons anymore. See issue #5743

Jackseed:
Could we remove the back button on the new pages? Since we open a new tab they are useless

GrandSchtroumpf:
@Jackseed we want to remove the back button all the time or only when it's opened in a new tab ?

Jackseed:
@GrandSchtroumpf only when it opens a new tab.. is it possible?

GrandSchtroumpf:
Normally yes, but it was supposed to work like that with the current state of the application. We are investigating why it's not working yet...

Jackseed:
Ok thanks, best is to do it on hotfix branch if you find a fix before the release

GrandSchtroumpf:
It's actually working as expected right now.
If you click on a link to a new tab you shouldn't see the back button. But if you open a empty tab then paste the url then you see the back button (as there is an history in the browser).
Handling the use case describe above (empty tab + link) would add a lot of overhead, and seems outside of the scope of this issue to me.

Jackseed:
Right my bad I didn't think it distinguished where you came from but was different on urls. Ok to close 👍

PR #6157, opened 2021-07-01
fixed(privacy and terms):remove back button

iPR0D1G1:
Fixed:  #6018


RemcoSimonides:
In what case do we still need to see this button? I think we don't need this button anymore

pldespaigne:
We need it in case it's open in the app (not as a new tab)
https://github.com/blockframes/blockframes/issues/6018

pldespaigne (changes requested):
Perfect ! but prefer `camelCase` over `snake_case` for the `no_back` variable.
Also this need to be in the release, so you should change your target branch from `develop` to `fix-for-release-3-2-1`

pldespaigne:
Closed, because #6018 was mistakenly reopen but there was no real issue

Issue #6111, opened 2021-06-24
Verify email of 'old' users


RemcoSimonides:
Since 2021-05-31 we force users to verify their email address. Older users are however still able to enter our platform without verifying their email.
When running prepareForTesting, new anonymized users are created. Their creation date is the date you run the prepareForTesting and thus are considered new users. However, they don't have their email verified yet because the real user behind this account never did so. Therefore we can't access the app with these 'old but new' users.

My suggestion is to run a migration to verify the emails of these old users. Then we can use these accounts again for development, and I can remove the 'verify email' button from the profile page.

For dev:
- [ ] Remove check for old users in EmailVerifiedGuard
- [ ] Remove 'verify email' from Profile page

@Jackseed ok for you?

Jackseed:
Lgtm 

Though just to be sure: we should keep the verify email button on the CRM in case we have a user blocked that doesn't receive our emails

RemcoSimonides:
👍 yes definitely we keep that button.

pldespaigne:
Just for you to know, verify email through CRM works fine, but the CRM ui is not updated so its always saying "verified email: false" even if the email is indeed verified.

Jackseed:
that definitely looks like an issue 😅

RemcoSimonides:
Ah yea, I forgot about that one. I'll fix it. Thanks 👍

phpgeekfr:
anon-db fix to have all users in PFT with emailVerified  can be found here : https://github.com/blockframes/blockframes/pull/6130

RemcoSimonides:
Export of current status of email verified of users can be found here
https://docs.google.com/spreadsheets/d/1R94eUgTFUS0A78bOMfIypS3ihkTIwJ5u2KInksiwhZw/edit#gid=1285095703

Jackseed:
UI improvements mentioned in #6089 
except from that everything looks good 👍

PR #6158, opened 2021-07-01
6111 Update email verified meta of user

RemcoSimonides:
Task on #6111

### DIFF
--- a/libs/admin/src/lib/admin-panel/pages/user/user.component.ts
+++ b/libs/admin/src/lib/admin-panel/pages/user/user.component.ts
@@ -190,9 +190,9 @@ export class UserComponent implements OnInit {
     });
   }
 
-  async verifyEmail() {
+  verifyEmail() {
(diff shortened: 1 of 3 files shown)


libs/admin/src/lib/admin-panel/pages/user/user.component.ts
```diff
@@ -190,9 +190,15 @@ export class UserComponent implements OnInit {
...
-  verifyEmail() {
+  async verifyEmail() {
     const f = this.functions.httpsCallable('verifyEmail');
-    return f({ uid: this.userId }).toPromise();
+    const success = await f({ uid: this.userId }).toPromise();
+    if (success) {
+      this.userService.update(this.userId, (user) => {
+        user._meta.emailVerified = true;
+        return user;
+      });
```
pldespaigne, line 200: I think it's better to do this in the backend, just before returning `true`.

pldespaigne (approved):

pldespaigne (approved):

PR #6162, opened 2021-07-01
Contract form avails

jeremie74:


libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
```diff
@@ -14,4 +14,15 @@
       <mat-icon matPrefix *ngIf="currency" [svgIcon]="currency"></mat-icon>
     </mat-form-field>
   </header>
+  <bf-form-table [form]="form.get('terms')" [columns]="termColumns" tablePosition="bottom">
+    <!-- Columns -->
+    <ng-template class="duration" colRef="avails.duration" let-duration>{{ duration.from | date : 'M/d/yyyy' }} - {{ duration.to | date : 'M/d/yyyy' }}</ng-template>
+    <ng-template colRef="avails.territories" let-territories><b>{{ territories | toLabel: 'territories' }}</b></ng-template>
```
GrandSchtroumpf, line 20: remove the <b> tag if it's not used in the other columns.

GrandSchtroumpf (approved):

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
jeremie74: Witch tag ?

Issue #6160, opened 2021-07-01
[Static Group] display name of a group of static


GrandSchtroumpf:
## Use case
When we want to display a list of item organized by group we would like to display only the name of the group if all item of a group is selected. 
For example with territories: 
- display "Europe" if all the countries in Europe are selected
- display "World" if all the countries are selected
- display each country if all Europe minus one country is selected

## Pipe
There is an existing pipe here called `triggerDisplay` : `libs\ui\src\lib\static-autocomplete\group\group.component.ts`
Extract this pipe in a `libs\utils\src\lib\pipe\group-label` folder

Rename the pipe `toGroupLabel`

## Implementation
Implement it in the contract-item for territories & media

PR #6163, opened 2021-07-01
refactor getMode pipe

iPR0D1G1:
Fixed: #6160
export refactor pipe to a pipe folder


GrandSchtroumpf:
I think you didn't understood the issue.
> When we want to display a list of item organized by group we would like to display only the name of the group
Your PR if not solving this use case

> Implement it in the contract-item for territories & media
Your PR is not implementing the pipe in the contract-item

Please take more time to : 
- read the issue
- ask if this is not clear
- test in the app if your PR is actually solving the issue

The pipe that needs to be extract is `triggerDisplay` not `getMode`. Currently this pipe is used only in the `static-group-autocomplete` to display the list of static model by group is applicable. We want to use this pipe to display `territories` and `media` in the table of the `contract-item` component. It should look like that : 
```html
<ng-template colRef="territories" let-territories>
  {{ territories | toGroupLabel:territoryGroup:'territories':'World' }}
</ng-template>
```
With territoryGroup beeing the `staticGroup.territories` from this file : `libs\utils\src\lib\static-model\groups.ts#l321`.

PR #6164, opened 2021-07-01
User object email

### LINKED ISSUE #6149: Update user variable on 3 last emails
Email to change from `userFirstName` to `user.firstName`:
- [x] verifyEmail: 'd-81438bdf511b43cfa866ca63a45a02ae'
- [x] welcomeMessage: 'd-fc05a8cf5b1548ebae9ca44247a6c256'
- [x] appAccessRequest: 'd-a740796b97bf403292b3940804977bbe'

Mussuranae:
Change the last email with the user object
Fix #6149


apps/backend-functions/src/templates/mail.ts
```diff
@@ -57,8 +57,8 @@ export function userResetPassword(email: string, link: string): EmailTemplateReq
 }
 
 /** Send email to an user who is requesting new app access */
-export function appAccessEmail(email: string, user: User): EmailTemplateRequest {
-  const data = { userFirstName: user.firstName };
+export function appAccessEmail(email: string, user: UserEmailData): EmailTemplateRequest {
+  const data = { user };
   return { to: email, templateId: templateIds.user.appAccessRequest, data };
```
phpgeekfr, line 62: you can directly: `return { to: email, templateId: templateIds.user.appAccessRequest, { user } };`

libs/utils/src/lib/apps.ts
```diff
@@ -52,8 +52,8 @@ export const sendgridEmailsFrom: Record<App | 'default', EmailJSON> = {
 
 // Those logos have to be in PNG because Gmail doesn't support SVG images
 export const appLogo = {
-  catalog: `${appUrl.content}/assets/email/archipel_content.png`,
-  festival: `${appUrl.market}/assets/email/archipel_market.png`,
+  catalog: `${appUrl.content}/assets/email/archipel-content.png`,
+  festival: `${appUrl.market}/assets/email/archipel-market.png`,
```
phpgeekfr, line 56: images where already renamed ?

phpgeekfr (approved):

libs/utils/src/lib/apps.ts
Mussuranae: Yep yep, that was a part of the cleaning of the asset, a little changes I merged before the big PR to "split" it a little. The logo for the emails are now in their own folder in the assets. That's less confusing and Stéphanie knows also what we use in emails ;)

Issue #6188, opened 2021-07-02
Change format of offers ids


pldespaigne:
Currently offers use the default firestore ids, but want it formatted like that: `<2 FIRST LETTER OF ORG NAME>-<6 RANDOM DIGITS>`

- [x] write a small function to generate those ids
- [x] use this function inside `createOffer()`

PR #6190, opened 2021-07-02
feat(utils):implemented createOfferId method

iPR0D1G1:
Fixed: #6188

### DIFF
--- a/libs/contract/src/lib/bucket/+state/bucket.service.ts
+++ b/libs/contract/src/lib/bucket/+state/bucket.service.ts
@@ -59,7 +59,8 @@ export class BucketService extends CollectionService<BucketState> {
     });
 
     // Create offer
-    const offerId = await this.offerService.add({
+    const offerId = createOfferId(orgName);
+    await this.offerService.add({
       buyerId: orgId,
       buyerUserId: this.authQuery.userId,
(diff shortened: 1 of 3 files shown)


iPR0D1G1 (commented):
reviewed

libs/contract/src/lib/bucket/+state/bucket.service.ts
```diff
@@ -64,7 +66,8 @@ export class BucketService extends CollectionService<BucketState> {
       status: 'pending',
       currency: bucket.currency,
       date: new Date(),
-      delivery
+      delivery,
+      id: createOfferId(orgName.substr(0, 2))
```
GrandSchtroumpf, line 70: Better to do : 
```typescript
const offerId = createOfferId(orgName);
await this.offerService.add({
  id: offerId,
  ...
})
```
And let the `createOfferId` take care of the logic of taking the first two letters

GrandSchtroumpf (approved):

libs/contract/src/lib/bucket/+state/bucket.service.ts
pldespaigne: I was going to say that ^^

PR #6192, opened 2021-07-02
refactor toGroupLabel to return array

iPR0D1G1:
Refactored to return arrays to permit using with mat-chips.

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.html
+++ b/apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.html
@@ -40,14 +40,14 @@
             <article class="expansion-panel-content">
               <ng-container *ngIf="term.territories.length <= maxTerritories; else moreTerritories">
                 <mat-chip-list aria-label="Territories">
-                  <mat-chip *ngFor="let territory of term.territories">{{ territory | toLabel: 'territories' }}</mat-chip>
+                  <mat-chip *ngFor="let territory of term.territories  | toGroupLabel:'territories':'World'">{{ territory }}</mat-chip>
                 </mat-chip-list>
               </ng-container>
               <ng-template #moreTerritories>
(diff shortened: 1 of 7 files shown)


iPR0D1G1 (commented):
Reviewed

apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.html
```diff
@@ -40,14 +40,14 @@ <h2 id="rights">Preselected Rights</h2>
             <article class="expansion-panel-content">
               <ng-container *ngIf="term.territories.length <= maxTerritories; else moreTerritories">
```
GrandSchtroumpf, line 41: You should check the length of the result of the pipe here, not the length of all territories.  Put the pipe at this level. Like that you can access the result in the ng-template to.

GrandSchtroumpf (changes requested):

apps/catalog/catalog/src/app/marketplace/title/avails/avails.component.html
```diff
@@ -38,21 +38,25 @@ <h2 id="rights">Preselected Rights</h2>
...
                   </mat-chip-list>
-                  <button mat-button class="spacing" (click)="openTerritoryModal(term.territories, 'territories')">... See {{ term.territories.length - 30 }} More</button>
-                </div>
-              </ng-template>
+                </ng-container>
+                <ng-template #moreTerritories>
+                  <div fxFlex fxLayout="column" fxLayoutAlign="start end">
+                    <mat-chip-list aria-label="Territories">
+                      <ng-container
+                        *ngFor="let territory of territories, index as i">
```
GrandSchtroumpf, line 51: You can use the splice pipe here instead of the *ngIf on the index.

GrandSchtroumpf (approved):
Nice job

Issue #6066, opened 2021-06-21
Filter by Release Year on Catalog


RemcoSimonides:
The tests showed that buyers were interested in having a possibility to filter the Titles with Production Year. This filter only needs to be implemented in the Catalog app

![Снимок экрана 2021-06-11 в 17.02.41.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/9195627e-e5fa-4539-8fc2-407530fc2273)

SamirTouira:
What do you think about this:


https://user-images.githubusercontent.com/52285231/124799746-2ae81880-df55-11eb-98a0-43995ad38374.mp4



I used the release year from Algolia because we only have this one to be able to filter but nothing about the Production Year (which is different of release year), so should the wording be different than "Production Year"?

Also in the slider, I sliced the years in "10" steps  (5 slices in the slider view). How many steps should we really have?

And the minimal year that I writed in the code is 1970. 2030 for the max year. What is the real values you want?

Thanks by advance for the answer :)

RemcoSimonides:
- [ ] Remove 2030 label

PR #6198, opened 2021-07-02
Filter by Production Year - Catalog

SamirTouira:
issue: #6066 

https://user-images.githubusercontent.com/52285231/124799746-2ae81880-df55-11eb-98a0-43995ad38374.mp4

### DIFF
--- a/libs/movie/src/lib/movie/form/filters/production-year/production-year.component.ts
+++ b/libs/movie/src/lib/movie/form/filters/production-year/production-year.component.ts
@@ -2,6 +2,7 @@ import { Component, ChangeDetectionStrategy, Input, OnInit, OnDestroy } from '@a
 import { FormControl } from '@angular/forms';
 import { Subscription } from 'rxjs';
 import { map, distinctUntilChanged } from 'rxjs/operators';
+import { MovieSearch } from '../../search.form';
 
 export const maxReleaseYear = 2030;
 
@@ -15,10 +16,13 @@ export class ProductionYearFilterComponent implements OnInit, OnDestroy {
 
   @Input() form: FormControl;
 
+  labels: FormControl;
   private sub: Subscription;
   maxReleaseYear = maxReleaseYear;
+  search: number;
 
   ngOnInit() {
+    // this.form.setValue(maxReleaseYear)
     this.sub = this.form.valueChanges.pipe(
       map(res => !!res),
       distinctUntilChanged()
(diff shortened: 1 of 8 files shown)


SamirTouira:
@RemcoSimonides correct me if I'm wrong but I still continue to think that it is not possible technically to fill the colour bar of the slider only to the right if the min value isn't 0 or if the max value is set before the min value, because even if the label is formatted, Algolia find nothing instead of finding all movies under or equal to 2030. Otherwise if it's doable, I am not able / not enough skilled to do it, sorry. I can only do right to left 2030 to 1970 with the colour bar only to the left

image

libs/movie/src/lib/movie/form/filters/production-year/production-year.component.ts
```diff
@@ -14,20 +15,37 @@ export const maxReleaseYear = 2030;
...
+  minYear = minYear;
 
   ngOnInit() {
     this.sub = this.form.valueChanges.pipe(
       map(res => !!res),
       distinctUntilChanged()
     ).subscribe(isDirty => isDirty ? this.form.markAsDirty() : this.form.markAsPristine());
+
+    this.componentForm.setValue(minYear);
+    if (this.componentForm.value >= minYear && this.componentForm.value <= maxYear) {
```
RemcoSimonides, line 32: You just set componentForm to `minYear` and then you check it's value if it's `>= minYear` and `<= maxYear`? It will always be equal to minYear in this case 😅

libs/movie/src/lib/movie/form/filters/production-year/production-year.component.ts
```diff
@@ -14,20 +15,37 @@ export const maxReleaseYear = 2030;
...
+      this.componentForm.setValue(maxYear - this.componentForm.value + minYear);
+      this.componentForm = this.form;
```
RemcoSimonides, line 34: Because of this the componentForm is equal to form... No need to create a separate formControl if you do this.

libs/movie/src/lib/movie/form/filters/production-year/production-year.component.ts
```diff
@@ -14,20 +15,37 @@ export const maxReleaseYear = 2030;
...
+      return this.componentForm.value;
```
RemcoSimonides, line 35: returning a value in `ngOnInit()` function? That value isn't going anywhere

libs/movie/src/lib/movie/form/filters/production-year/production-year.component.ts
```diff
@@ -14,20 +15,37 @@ export const maxReleaseYear = 2030;
...
+    }
   }
 
   ngOnDestroy() {
     this.sub.unsubscribe();
   }
+
+  displayLabel(value: number) {
+    const invertedValue =  maxYear - value + minYear;
+    if (value >= minYear && value <= maxYear) {
```
RemcoSimonides, line 45: value is always equal or bigger than minYear and smaller or equal to maxYear. You can't slide the slider further than these values, can you?

libs/movie/src/lib/movie/form/filters/production-year/production-year.component.ts
```diff
@@ -14,20 +15,37 @@ export const maxReleaseYear = 2030;
...
+      return this.componentForm.value;
+    }
   }
 
   ngOnDestroy() {
     this.sub.unsubscribe();
   }
+
+  displayLabel(value: number) {
+    const invertedValue =  maxYear - value + minYear;
```
RemcoSimonides, line 44: You've repeated this formula twice now ;) that means you can make it reusable

libs/movie/src/lib/movie/form/search.form.ts
```diff
@@ -155,7 +156,7 @@ export class MovieSearchForm extends FormEntity<MovieSearchControl> {
       search['filters'] = `budget >= ${max - this.minBudget.value ?? 0}`;
     }
     if (this.release.value) {
-      search['filters'] = `release.year <= ${this.release.value}`;
+      search['filters'] = `release.year >= ${maxYear - this.release.value + minYear}`;
```
RemcoSimonides, line 159: searchForm shouldn't have inverted value here anymore (1970 instead of expected 2030 for example). It should already be inverted by now. So when there is a bug within the searchForm and you try to debug it, you don't need to follow all the code to figure out why there is 1970 instead of 2030. Keep the logic and complexity within the component for this 'inverted' value.

RemcoSimonides (changes requested):
I'll create my PR and show you the solution. You got close! Nice try 👍 :)
Solution: #6290

RemcoSimonides:
Closing in favor of #6290

PR #6248, opened 2021-07-07
Fix 6084 b

phpgeekfr:
in the title list, it was missing in the mockups but we should have the amount of films in parenthesis aside the tab title, something like this (not the good tab names):


apps/festival/festival/src/app/dashboard/title/list/list.component.scss
```diff
@@ -11,11 +11,9 @@ header.import-header {
 header[mat-tab-nav-bar] button {
   border: none;
   background-color: transparent;
-}
-
-header[mat-tab-nav-bar] button {
-  border: none;
-  background-color: transparent;
+  span {
+    margin-left: 3px;
```
SamirTouira, line 15: Correct me if I'm wrong but we should have only a multiple of 4 or 8 pixel value

apps/financiers/financiers/src/app/dashboard/title/list/list.component.scss
```diff
@@ -11,6 +11,9 @@ header.import-header {
 header[mat-tab-nav-bar] button {
   border: none;
   background-color: transparent;
+  span {
+    margin-left: 3px;
```
SamirTouira, line 15: Same here ^^

SamirTouira (approved):
LGTM

apps/catalog/catalog/src/app/dashboard/title/list/list.component.scss
```diff
@@ -11,6 +11,9 @@ header.import-header {
 header[mat-tab-nav-bar] button {
   border: none;
   background-color: transparent;
+  span {
+    margin-left: 3px;
```
SamirTouira, line 15: :)

PR #6249, opened 2021-07-07
saving terms incomes in contract

jeremie74:


libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
```diff
@@ -59,8 +65,25 @@ export class ContractFormComponent implements OnInit {
...
+    async save() {
+      if (this.form.valid) {
+        const { terms, titleId, price } = this.form.value;
+        const contractId = this.route.snapshot.params.contractId;
+        const write = this.contractService.batch(); // create a batch
+        const termList = (terms as any[]).map(term => toTerm(term, contractId));
+        const termIds = await this.termService.upsert(termList, { write });
+        const existingTermIds = this.contract?.termIds || [];
+        const termIdsToDelete = existingTermIds.filter(id => !termIds.includes(id));
+        await this.termService.remove(termIdsToDelete);
```
GrandSchtroumpf, line 83: Please add the `{ write }` batch as a second argument of your method.

libs/contract/src/lib/term/+state/term.model.ts
```diff
@@ -38,7 +37,6 @@ export interface Term<T extends Date | firebase.firestore.Timestamp = Date> {
 export function createTerm(params: Partial<Term<Date>> = {}): Term<Date> {
   return {
     id: '',
-    orgId: '',
```
GrandSchtroumpf, line 41: Looks like removing this orgId is creating a build issue in this file : `libs\import\src\lib\components\view-extracted-elements\contract\view-extracted-contracts.component.ts`.
It seems like it's the only place where we use the orgId. You can remove it from this file also.

GrandSchtroumpf (approved):

PR #6264, opened 2021-07-09
overload toLabel

GrandSchtroumpf:
Use case: 
```
{{ text | toLabel:'offerStatus' | titlecase }}
```
this would fails, because the output can be an array. by overloading you say that if the first param is a string, then the return value is a string.


iPR0D1G1 (commented):
reviewed, I understand better now.

iPR0D1G1 (approved):

Issue #6210, opened 2021-07-05
MF - Open discussion email is broken


Jackseed:
Template ID: `d-e902521de8684c57bbfa633bad88567a`
This template has been broken with our 2 last changes:
- [x] `userSubject` is not working in this email (it should replace the investor variable)
- [x] `org.denomination` neither
- [x] `from` used to be a number but now it's our email address so we should change the name of that variable in the template (I would say `min` `max` instead of `from` `to`)

Current result of my test:
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/6c204b3a-e852-4ccc-abc9-21c28d956957)

Mussuranae:
This one isn't handle by our backend functions.... Maybe we have to change that ?

Jackseed:
Oh ok got it... how complex would it be to handle it with our backend functions? 
Cause I may still rollback the email template to previous version that doesn't use org and user objects. The only remaining issue would be the `from` that is automatically send with our emails so we would need to change it anyway.

phpgeekfr:
- [ ] add 2E2 test on this

Mussuranae:
@Jackseed this email is now updated. I also update the template for the `from` and `to` to a little object `scope: { from, to }` so that we have no problem to display the value of the form.
It would be good to have a discussion with @phpgeekfr about this email, it looks like it is a little different from the others emails we have and its purpose was a little different. So when you came back from Cannes, maybe we can organize a call together.
I will now work on an e2e test for this.

Jackseed:
Ok good! what's different from the others? with pleasure to discuss that next week. 

Also what do you think of integrating other emails to backend functions like the new offer email (`Template ID: d-94a20b20085842f68fb2d64fe325638a`) and mb other I miss?

Mussuranae:
It can be nice, let's see next week ;)

PR #6268, opened 2021-07-09
E2e financiers

Mussuranae:
E2E test for investment email

Related to #6210 
image


apps/financiers/financiers-e2e/src/integration/app.spec.ts
```diff
@@ -9,7 +9,7 @@ describe('financiers', () => {
   });
 
   it('should display welcome message', () => {
-    cy.wait(300 * SEC);
+    cy.wait(30 * SEC);
     cy.get('financiers-landing h4', {timeout: 30 * SEC})
```
xmano, line 13: `cy.get('financiers-landing h4', {timeout: 30 * SEC})`  you could increate timeout here and eliminate the previous wait. This is equivalent and will be faster.

apps/financiers/financiers/src/app/marketplace/home/home.component.html
```diff
@@ -8,8 +8,8 @@ <h1 class="mat-display-2">The professional financing hub <br />for films and TV
     professional content financiers.
   </p>
   <div class="actions" fxLayout="row" fxLayoutGap="24px">
-    <a mat-flat-button color="accent" test-id="library" routerLink="../title">Projects</a>
-    <a mat-stroked-button test-id="sales agent" routerLink="../organization">Financiers</a>
+    <a mat-flat-button color="accent" routerLink="../title">Projects</a>
+    <a mat-stroked-button routerLink="../organization">Financiers</a>
```
xmano, line 12: Why did you remove test-id here? Is it not needed?

libs/e2e/src/lib/utils/functions.ts
```diff
@@ -9,7 +9,7 @@ export function clearDataAndPrepareTest(path: string = '/auth') {
   cy.clearLocalStorage();
   indexedDB.deleteDatabase('firebaseLocalStorageDb');
   if (path !== '') {
-    cy.visit(path);
+    cy.visit(path, {timeout: 30 * SEC});
```
xmano, line 12: This may be counterproductive.

https://docs.cypress.io/guides/references/configuration#Timeouts

pageLoadTimeout is already set at 60s. I suggest you remove this since our servers are slower and we need 60s timeout.

apps/financiers/financiers-e2e/src/support/pages/marketplace/SearchPage.ts
```diff
@@ -0,0 +1,56 @@
...
+
+
+  ///////////////////////////
+  // ACTIONS ON MOVIE CARD //
+  ///////////////////////////
+  public clickWishlistButton(movieName: string) {
+    cy.get('movie-card', {timeout: 30 * SEC})
+      .contains(movieName).parent().parent()
+      .find('button[test-id=heart-button]').click('bottom', {force: true} );
+    cy.wait(2000);
```
xmano, line 44: 2 * SEC will improve readability of wait time

apps/financiers/financiers-e2e/src/support/pages/marketplace/SearchPage.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  /** Click on a movie-card and navigate to the movie-view */
+  public selectMovie(movieName: string) {
+    cy.log(`=>selectMovie : {${movieName}}`);
+    cy.get('movie-card', {timeout: 30 * SEC})
+      .contains(movieName).parent().parent()
+      .find('a').click();
```
xmano, line 52: I suggest a small wait here of 0.5 SEC after click

libs/e2e/src/lib/utils/functions.ts
```diff
@@ -229,3 +229,11 @@ export function assertMoveTo(path: string) {
     expect(loc.pathname).to.eq(path);
   });
 }
+
+/** Used to navigate from the menu in the navbar */
+export function openSidenavMenuAndNavigate(button: string = '') {
```
xmano, line 234: Can this be achieved by existing function `clickOnMenu()` ?

apps/financiers/financiers-e2e/src/integration/investment-email.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    p4.fillDiscussionForm(discussionData);
+    p4.sendDiscussionEmail();
+
+    // Waiting this snackbar to appear, because it appears after the emails have been sent.
+    cy.get('snack-bar-container', {timeout: 30 * SEC}).should('contain', 'Your email has been sent.');
+    cy.log('Email sent');
+
+    // Check if emails are well sent.
+    cy.log('Checking emails...');
+    cy.mailosaurSearchMessages(serverId, {
```
xmano, line 60: Would be good to check the subject line of the email and content if it is unique.

xmano (changes requested):
In some place we have timeout: 20000 etc - please use 20 * SEC to represent the values as it is easier to read.

apps/financiers/financiers/src/app/marketplace/home/home.component.html
Mussuranae: No it's not used, and that's created some duplicate with the sidenav.

libs/e2e/src/lib/utils/functions.ts
Mussuranae: Yes it can, so we have a duplicate here I guess, because this function is also on some catalog e2e I think.

apps/financiers/financiers-e2e/src/integration/investment-email.spec.ts
Mussuranae: There is the name of the user who send the email, I don't know if I can get it in e2e :/ 
I can do a partial check because the rest of the subject is more or less static.

apps/financiers/financiers-e2e/src/integration/app.spec.ts
Mussuranae: I wonder if we want this test as it only goes on the landing page and look if there is the title....

xmano: For now let the test remain there. I remember putting the delay as workaround for CI Issue. Now that is fixed and we can remove line 12 entirely and add 60s timeout for get. 

We will need this test later for implementing sanity checks for deploy. So let it be.

xmano (approved):
LGTM

Issue #6064, opened 2021-06-18
Test enforce light theme on landing page


RemcoSimonides:
For a new design we need to check if it's possible to enforce light theme on this page only. Ask Stephanie to review it.

Jackseed:
@Mussuranae I can't see it on staging though it looks like it has been merged before we create the release branch so it should be here.
Maybe I'm missing something?

Jackseed:
I can't say where the issue is coming from but it's not working on Brave, Chrome and Safari but it does on Firefox.

On Brave I had even quite the opposite of what we want:
- I logged in and select light mode
- I log out
- landing goes dark.. 
Tested currently on staging

Jackseed:
Update: it looks like it's a Mac issue, MacOs enforce the OS theme to the landing; as I'm in dark mode on Mac all the landing stay dark

Mussuranae:
I tried n my side with Firefox and Chromium (Linux) and it was ok for me. I guess a big part of our users doesn't use their computer the same way as you, and they maybe never look at their preferences or don't look also the preference of the browser.

Anyway, at a moment we won't be able to force the light mode if the user really want to see its browser in dark mode.

Jackseed:
My issue here is not at the browser level but the OS. And all Mac users have to set it as it's one of the onboarding steps you have to pass through when buying a new Mac. 
Since 50% of our users have MacOS I think we should spend a little more time trying to understand if and how we could overcome OS preferences

RemcoSimonides:
@Jackseed test it again when we release to staging this Wednesday. I think I have fixed it

PR #6272, opened 2021-07-12
6064 Force light theme on landing page

RemcoSimonides:
Fix issue #6064


libs/landing/src/lib/shell/shell.component.ts
```diff
@@ -67,9 +67,14 @@ export class LandingShellComponent {
     private snackBar: MatSnackBar,
     private routerQuery: RouterQuery,
     private functions: AngularFireFunctions,
-    theme: ThemeService)
-  {
-    theme.initTheme('light');
+    private theme: ThemeService
+  ) {
+    theme.setTheme('light')
```
RemcoSimonides, line 72: Theme defined as parameter for `initTheme()` is overwritten by browser/system preference and by setting in localStorage. Therefore I set the theme directly for this page.
The theme 'reset' when leaving the landing page. Then the theme is set back to either browser/system preference, localStorage setting or 'light' as defined in the parameter.

Mussuranae (approved):

PR #6273, opened 2021-07-12
Polishing contract form

jeremie74:

### DIFF
--- a/libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
+++ b/libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
@@ -1,24 +1,37 @@
-<ng-container>
-  <form [formGroup]="form" (ngSubmit)="save()">
+<form [formGroup]="form" (ngSubmit)="save()">
     <header>
       <mat-form-field appearance="outline">
-        <mat-label>Select a Title</mat-label>
-        <mat-select formControlName="titleId">
+        <mat-select formControlName="titleId"  placeholder="Select a title">
           <mat-option *ngFor="let title of (titles$ | async)" [value]="title.id">
             {{ title.title.international }}
           </mat-option>
         </mat-select>
+        <mat-error *ngIf="form.controls['titleId'].hasError('required')">Title is required!</mat-error>
       </mat-form-field>
       <mat-form-field appearance="outline">
         <mat-label>Title price</mat-label>
         <input type="number" matInput formControlName="price" required min="0"/>
         <mat-icon matPrefix *ngIf="currency" [svgIcon]="currency"></mat-icon>
+        <mat-error *ngIf="form.hasError('required', 'price')">Price field is required!</mat-error>
       </mat-form-field>
     </header>
+
+    <ng-container *ngIf="showBfFilterTable">
+      <ng-container *ngIf="contracts$ | async as contracts; else loading">
+        <bf-table-filter
+        [source]="contracts" 
+        [columns]="termColumns" 
+        clickable 
+        showFilter 
+        showPaginator link="/c/o/dashboard/contract/">
+      </bf-table-filter>
+      </ng-container>
+    </ng-container>
+
     <bf-form-table [form]="form.get('terms')" [columns]="termColumns" tablePosition="bottom">
       <!-- Columns -->
       <ng-template class="duration" colRef="avails.duration" let-duration>{{ duration.from | date : 'M/d/yyyy' }} - {{ duration.to | date : 'M/d/yyyy' }}</ng-template>
-      <ng-template colRef="avails.territories" let-territories>{{ territories | toLabel: 'territories' }}</ng-template>
+      <ng-template colRef="avails.territories" let-territories>{{ territories | toGroupLabel:'territories':'World' | join }}</ng-template>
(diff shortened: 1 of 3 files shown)


libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
```diff
@@ -1,24 +1,37 @@
...
     <header>
       <mat-form-field appearance="outline">
-        <mat-label>Select a Title</mat-label>
-        <mat-select formControlName="titleId">
+        <mat-select formControlName="titleId"  placeholder="Select a title">
           <mat-option *ngFor="let title of (titles$ | async)" [value]="title.id">
             {{ title.title.international }}
           </mat-option>
         </mat-select>
+        <mat-error *ngIf="form.controls['titleId'].hasError('required')">Title is required!</mat-error>
```
RemcoSimonides, line 9: Use .get() instead: `form.get('titleId').hasError('required')`

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
```diff
@@ -38,6 +40,7 @@ export class ContractFormComponent implements OnInit {
   })
   titles$ = this.service.valueChanges(ref => ref.where('app.catalog.status', '==', 'approved'));
   currency?: string;
+  showBfFilterTable: boolean =false;
```
RemcoSimonides, line 43: No need to type it if you set a default value to it. Typescript already knows it's a boolean then. Also, add a spacebar after the `=` 🙂

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
```diff
@@ -46,27 +49,39 @@ export class ContractFormComponent implements OnInit {
...
+    private orgQuery: OrganizationQuery,
     ){}
 
     async ngOnInit() {
+      if (this.route.snapshot.queryParams.termId) {
+        this.showBfFilterTable = true;
+      }
       const contractId: string = this.route.snapshot.params.contractId;
-      const [ contract, income ] = await Promise.all([
+      const offerId: string = this.route.snapshot.params.offerId;
```
RemcoSimonides, line 73: I don't know what the difference is between queryParams and params but it would be nice if we could do it like this
`const { termId, contractId, offerId } = this.route.snapshot.queryParams`

RemcoSimonides (approved):

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
jeremie74: Query params is a query (an object of key value pair that go in the url), params is the :id that is sending in the URL

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
```diff
@@ -1,24 +1,25 @@
...
+        <mat-error *ngIf="form.hasError('required', 'price')">Price field is required!</mat-error>
       </mat-form-field>
     </header>
+    
     <bf-form-table [form]="form.get('terms')" [columns]="termColumns" tablePosition="bottom">
       <!-- Columns -->
       <ng-template class="duration" colRef="avails.duration" let-duration>{{ duration.from | date : 'M/d/yyyy' }} - {{ duration.to | date : 'M/d/yyyy' }}</ng-template>
-      <ng-template colRef="avails.territories" let-territories>{{ territories | toLabel: 'territories' }}</ng-template>
+      <ng-template colRef="avails.territories" let-territories>{{ territories | toGroupLabel:'territories':'World' | join }}</ng-template>
       <ng-template colRef="avails.medias" let-media>{{ media | toLabel: 'medias' }}</ng-template>
```
GrandSchtroumpf, line 23: You need to put the `toGroupLabel` also on the media

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
```diff
@@ -46,27 +49,39 @@ export class ContractFormComponent implements OnInit {
     'versions': 'Versions',
     'runs': '# of broadcasts'
   }
+  public orgId = this.orgQuery.getActiveId();
+  public contracts$ = this.contractService.valueChanges(
```
GrandSchtroumpf, line 53: you're not using it, so you can remove it. Plus as we are in the CRM we don't want to get the contract of the current org as the current org is always someone from Cascade8.
You can also remove the orgId then.

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
```diff
@@ -46,27 +49,39 @@ export class ContractFormComponent implements OnInit {
...
     private contractService: ContractService,
-    private termService: TermService
+    private termService: TermService,
+    private offerService: OfferService,
+    private orgQuery: OrganizationQuery,
     ){}
 
     async ngOnInit() {
+      if (this.route.snapshot.queryParams.termId) {
+        this.showBfFilterTable = true;
```
GrandSchtroumpf, line 70: you're not using it. You can remove it.

PR #6298, opened 2021-07-15
Crm updating contract income

jeremie74:


libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
```diff
@@ -95,8 +96,12 @@ export class ContractFormComponent implements OnInit {
         const existingTermIds = this.contract?.termIds || [];
         const termIdsToDelete = existingTermIds.filter(id => !termIds.includes(id));
         await this.termService.remove(termIdsToDelete, { write });
-        await this.contractService.update(contractId, { titleId, termIds  }, { write });
-        await this.incomeService.update(contractId, { price }, { write }); // Update the price in the batch
+        if(this.form.value.titleId !== this.contract.titleId) {
```
GrandSchtroumpf, line 99: You already have the variables in memory from `const { terms, titleId, price } = this.form.value;`
```typescript
if (titleId !== this.contract.titleId) {
  await this.contractService.update(contractId, { titleId, termIds }, { write });
}
if (price !== this.income.price) {
  await this.incomeService.update(contractId, { price }, { write });
}
```

GrandSchtroumpf (approved):
Can you change "approved" for "accepted" in the title query ? It was actually a mistake from the beginning : 
```typescript
this.service.valueChanges(ref => ref.where('app.catalog.status', '==', 'accepted'));
```

PR #6306, opened 2021-07-15
6088 webhook listener sendgrid to bigquery

RemcoSimonides:
Listens to SendGrid Events webhook and inserts the data into bigquery
Would be nice to have this in the release 3.3 so we can start gathering data about email traffic

### DIFF
--- a/apps/backend-functions/src/sendgrid.ts
+++ b/apps/backend-functions/src/sendgrid.ts
@@ -1,0 +1,55 @@
+import { BigQuery } from '@google-cloud/bigquery';
+import { Request, Response } from "firebase-functions";
+
+/**
+ * Listens to Events Webhook of Sengrid
+ * 
+ * Event Webhook Schema
+ * email:string, timestamp:timestamp, event:string, asm_group_id:numeric, sg_event_id:string, sg_message_id:string, sg_template_id:string, sg_template_name:string, reason:string, status:string 
+ */
+export const eventWebhook = async (req: Request, res: Response) => {
+
+  const datasetId = 'sendgrid';
+  const tableId = 'events';
+
+  try {
+    const bigquery = new BigQuery();
+    const dataset = bigquery.dataset(datasetId);
+    const table = dataset.table(tableId);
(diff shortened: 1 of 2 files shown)


apps/backend-functions/src/sendgrid.ts
```diff
@@ -0,0 +1,55 @@
...
+      timestamp,
+      event,
+      asm_group_id,
+      sg_event_id,
+      sg_message_id,
+      sg_template_id,
+      sg_template_name,
+      reason,
+      status
+    }));
```
RemcoSimonides, line 44: Perhaps there is a smarter way to get rid of unknown object properties. I can't insert the item into bigquery if the property isn't already in the schema so this way I want to make sure it doesn't have any unknown properties.

apps/backend-functions/src/sendgrid.ts
```diff
@@ -0,0 +1,55 @@
...
+  const datasetId = 'sendgrid';
+  const tableId = 'events';
+
+  try {
+    const bigquery = new BigQuery();
+    const dataset = bigquery.dataset(datasetId);
+    const table = dataset.table(tableId);
+    const body = req.body as Array<any>;
+
+    const cleaned = body.map(({
```
GrandSchtroumpf, line 21: Better to do : 
```typescript
function fromSendgrid(data: SendgridType): BlockframeType {
  return {
    email: data.email,
    timestamp: data.timestamp,
    ...
  }
}
....
const cleaned = body.map(fromSendgrid);
```

apps/backend-functions/src/sendgrid.ts
GrandSchtroumpf: you can do : 
```typescript
function copyFields<T>(from: T, fields: string[]): T {
  const result: Partial<T> = {};
  for (const key in from) {
    if (fields.includes(key)) result[key] = from[key];
  }
  return result as T;
}
```

GrandSchtroumpf (approved):

PR #6310, opened 2021-07-16
Reactive hour list

### LINKED ISSUE #5805: Event form - Reactive hour list
App festival
Event form (short on calendar page + long)
For usability, the first displayed hour in the list here should be the one already selected

Mussuranae:
Fix #5805


libs/ui/src/lib/form/time-picker/time-picker.component.html
```diff
@@ -15,7 +15,8 @@
     formControlName="time"
     (input)="setDate()"
   />
+
 </div>
-<mat-autocomplete #timing="matAutocomplete" (optionSelected)="setDate()">
-  <mat-option *ngFor="let hour of hours" [value]="hour">{{ hour }}</mat-option>
+<mat-autocomplete #timing="matAutocomplete" (optionSelected)="setDate()" (opened)="scrollToOption()">
+  <mat-option [class]="tagID + '-' + i" *ngFor="let hour of hours; let i = index" [value]="hour">{{ hour }}</mat-option>
```
GrandSchtroumpf, line 21: For better readabillty can you do  :
```html
<ng-container *ngFor="let hour of hours; let i = index">
  <mat-option [id]="tagID + '-' + i" [value]="hour"></mat-option>
</ng-container>
```
I use id instead of class, because you shouldn't have the same mat-option with the same id.

libs/ui/src/lib/form/time-picker/time-picker.component.ts
```diff
@@ -203,4 +212,12 @@ export class TimePickerComponent implements ControlValueAccessor, MatFormFieldCo
     this.onChange(value)
     this.timeChange.emit(value);
   }
+
+  /* Allow us to display the "pre-selected" mat-option */
+  scrollToOption() {
+    const time =this.form.get('time').value
+    const index = hours.indexOf(time);
+    const option = this.document.getElementsByClassName(`${this.tagID}-${index}`)[0];
```
GrandSchtroumpf, line 220: Better: 
```typescript
// Get the reference of the host
constructor(privat el: ElementRef) {}

scrollToOption() {
    const time =this.form.get('time').value
    const index = hours.indexOf(time);
    // query in the element and not on the global document
    const option = this.el.nativeElement.getElementsByClassName(`${this.tagID}-${index}`)[0];
}
```

GrandSchtroumpf (approved):

libs/ui/src/lib/form/time-picker/time-picker.component.ts
Mussuranae: If I use the ElementRef instead of the Document, it finds nothing :/ 
Also, I don't know if it will find the `id` of the mat-option if we check for `className`. I tried several things with the `nativeElement` but get always `undefined` as a result for option :|

PR #6313, opened 2021-07-16
Implement Holdbacks

GrandSchtroumpf:
- Creates holdbacks model
- Creates holdbacks form
- Add holdbacks on selection
- Add holdbacks on crm
- Add holdbacks on dashboard
- Add holdbacks on marketplace

### DIFF
--- a/libs/contract/src/lib/contract/+state/contract.model.ts
+++ b/libs/contract/src/lib/contract/+state/contract.model.ts
@@ -2,9 +2,8 @@ import { MovieLanguageSpecification } from "@blockframes/movie/+state/movie.fire
 import { createDocumentMeta, DocumentMeta } from "@blockframes/utils/models-meta";
 import { Media, Territory } from "@blockframes/utils/static-model";
 import { Duration } from '../../term/+state/term.model';
-import type firebase from 'firebase';
 import { createLanguageKey } from "@blockframes/movie/+state";
-type Timestamp = firebase.firestore.Timestamp;
+import { Timestamp } from "@blockframes/utils/common-interfaces/timestamp";
 
 export const contractStatus = ['pending', 'accepted', 'declined', 'archived'] as const;
 
@@ -17,7 +16,7 @@ export interface Holdback<D extends Timestamp | Date = Date> {
   languages: Record<string, MovieLanguageSpecification>;
 }
 
-export interface Contract<D = Date> {
+export interface Contract<D extends Timestamp | Date = Date> {
   _meta: DocumentMeta<D>;
   id: string;
   type: 'mandate' | 'sale';
@@ -37,6 +36,7 @@ export interface Contract<D = Date> {
   sellerId: string;
   /** Org ids that have contract parent of this contract */
   stakeholders: string[];
+  holdbacks: Holdback<D>[];
 }
 
 export type ContractDocument = Contract<Timestamp>;
@@ -76,6 +76,7 @@ export function createMandate(params: Partial<Mandate> = {}): Mandate {
     type: 'mandate',
     status: 'pending',
     stakeholders: [],
+    holdbacks: [],
     ...params
   }
 }
@@ -95,6 +96,7 @@ export function createSale(params: Partial<Sale> = {}): Sale {
     type: 'sale',
     status: 'pending',
     stakeholders: [],
+    holdbacks: [],
     ...params
   }
 }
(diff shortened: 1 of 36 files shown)


libs/contract/src/lib/contract/+state/contract.model.ts
```diff
@@ -85,3 +108,11 @@ export function isMandate(contract: Contract): contract is Mandate {
 export function isSale(contract: Contract): contract is Sale {
   return contract.type === 'sale';
 }
+
+
+export function convertDuration(duration: Duration<Timestamp>): Duration<Date> {
+  return {
+    from: duration.from.toDate(),
+    to: duration.to.toDate(),
+  }
```
RemcoSimonides, line 117: We have a `toDate()` function in utils for this. That one checks if it's really a Timestamp and not a Date to prevent errors

libs/contract/src/lib/contract/holdback/form/form.component.html
```diff
@@ -0,0 +1,93 @@
...
+  <header>
+    <h1>Holdback for "{{ title.title.international }}"</h1>
+    <button mat-icon-button (click)="cancel()">
+      <mat-icon svgIcon="close"></mat-icon>
+    </button>
+  </header>
+  <bf-form-table [columns]="columns" [form]="form">
+    <!-- Table -->
+    <ng-template colRef="duration" let-duration>
+      {{ duration.from | date }} - {{ duration.to | date:'M/d/y' }}
```
RemcoSimonides, line 22: define date format for `duration.from | date`

libs/contract/src/lib/contract/holdback/form/form.component.scss
```diff
@@ -0,0 +1,53 @@
...
+
+.group-label {
+  cursor: pointer;
+}
+
+.actions {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
```
RemcoSimonides, line 49: Why not put this in the html with fxLayout so it's clear from the html what it should look like?

libs/contract/src/lib/contract/holdback/form/form.component.scss
```diff
@@ -0,0 +1,53 @@
+:host {
+  display: block;
+}
+
+header {
+  display: flex;
+  justify-content: space-between;
```
RemcoSimonides, line 7: Why not put this in the html with fxLayout so it's clear from the html what it should look like?

libs/contract/src/lib/contract/holdback/form/form.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  @Input() holdbacks: Holdback[] = [];
+  @Output() change = new EventEmitter<Holdback[]>();
+
+  ref: MatDialogRef<void, void>;
+  form: FormList<Holdback, HoldbackForm>;
+  columns = {
+    duration: 'Duration',
+    territories: 'Territories',
+    medias: 'Media',
+    languages: 'Versions',
```
RemcoSimonides, line 27: Test whether you can sort correctly by these columns, otherwise disable sorting like in the other PR
`duration: { value: 'Duration', disableSorting: true };`

libs/contract/src/lib/contract/holdback/list/list.component.html
```diff
@@ -0,0 +1,28 @@
+<ng-container *ngIf="holdbacks.length">
+  <h3>Holdbacks</h3>
+  <bf-table-filter [source]="holdbacks" [columns]="columns">
+    <ng-template colRef="duration" let-duration>
+      {{ duration.from | date }} - {{ duration.to | date:'M/d/y' }}
```
RemcoSimonides, line 5: Add format for date here too

libs/contract/src/lib/term/components/detailed/detailed.component.scss
```diff
@@ -2,6 +2,18 @@
   display: block;
 }
 
-section {
-  padding: 24px;
+header {
+  display: flex;
+  align-items: flex-start;
+  justify-content: space-between;
```
RemcoSimonides, line 8: define in html

RemcoSimonides (approved):

libs/contract/src/lib/contract/+state/contract.model.ts
GrandSchtroumpf: This function should be used in the formatFormFirestore hooks which are always Timestamps. But ok.

PR #6318, opened 2021-07-19
Price in CRM should be 0 by default if no price entered - hint message

jeremie74:


libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
```diff
@@ -10,9 +10,11 @@
     </mat-form-field>
     <mat-form-field appearance="outline">
       <mat-label>Title price</mat-label>
-      <input type="number" matInput formControlName="price" required min="0"/>
+      <input type="number" matInput formControlName="price" />
```
RemcoSimonides, line 13: Shouldn't the `min="0"` stay? Without it you can enter sub-zero numbers I think

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
```diff
@@ -10,9 +10,11 @@
...
       <mat-icon matPrefix *ngIf="currency" [svgIcon]="currency"></mat-icon>
-      <mat-error *ngIf="form.hasError('required', 'price')">Price field is required!</mat-error>
+      <!-- Please live this comment if PM change idea 
+      <mat-error *ngIf="form.hasError('required', 'price')">Price field is required!</mat-error> -->
+      <mat-hint>If no price entered. Seller will recieve active price.</mat-hint>
```
RemcoSimonides, line 17: comma, receive* and Asking Price*  (receive is incorrectly written in the design - correct way is receive*)
"If no price entered, Seller will receive Asking Price"

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
```diff
@@ -10,9 +10,11 @@
     </mat-form-field>
     <mat-form-field appearance="outline">
       <mat-label>Title price</mat-label>
-      <input type="number" matInput formControlName="price" required min="0"/>
+      <input type="number" matInput formControlName="price" />
       <mat-icon matPrefix *ngIf="currency" [svgIcon]="currency"></mat-icon>
-      <mat-error *ngIf="form.hasError('required', 'price')">Price field is required!</mat-error>
+      <!-- Please live this comment if PM change idea 
+      <mat-error *ngIf="form.hasError('required', 'price')">Price field is required!</mat-error> -->
```
RemcoSimonides, line 16: You can remove it. We aim to keep the repo clean from unused code ;)

RemcoSimonides (changes requested):

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
jeremie74: Ok i copied it from an other page, i had a doubt and yes its receive

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
jeremie74: Yes but is someone change advice, we will have to put it, but ok ill do it

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.html
jeremie74: yes it's true

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
```diff
@@ -108,8 +108,13 @@ export class ContractFormComponent implements OnInit {
           await this.contractService.update(contractId, { titleId, termIds  }, { write });
         }
         if(price !== this.income.price) {
-          await this.incomeService.update(contractId, { price }, { write }); // Update the price in the batch
+          if(price === null) {
+            await this.incomeService.update(contractId, { price: 0 }, { write });
+          } else {
+            await this.incomeService.update(contractId, { price }, { write });
+          }
```
RemcoSimonides, line 115: After small interaction with Francois we decided to not change the price to `0` and keep `null` instead. Can you undo this again? ^^

RemcoSimonides (approved):

libs/admin/src/lib/admin-panel/pages/offers/shell/contract-form/contract-form.component.ts
jeremie74: I understand it's a strange request

PR #6338, opened 2021-07-21
Merge back release 3 3 develop

phpgeekfr:


libs/invitation/src/lib/components/action/action.component.ts
```diff
@@ -24,6 +24,8 @@ export class ActionComponent {
     }
   }
 
+  @Input() display: 'full' | 'icon-only' = 'full';
```
GrandSchtroumpf, line 27: When you've got only two options it's better to use a boolean with the `@boolean` decorator :
```typescript
@Input() @boolean small = true; // use the name you want
```
Like that in the html you can do : 
```html
<invitation-action small></invitation-action> <!-- Icon only -->
<invitation-action></invitation-action> <!-- Icon & text-->
```

GrandSchtroumpf (approved):

libs/invitation/src/lib/components/action/action.component.ts
phpgeekfr: created #6339

PR #6343, opened 2021-07-21
More searchable attribute

### LINKED ISSUE #6138: Algolia - Add searchable data to the movie search bar
When typing a search query in the search bar, it should not only search the text in the 'international title' and 'original title' but also in:

- [x] International Title;
- [x] Original Title;
- [x] Country of Origin (should be ok as 1 field);
- [x] Director's First Name, Director's Last Name; (better focus only on Last name? + can be ambiguous as a table)
- [x] Original Language (should be ok);
- [x] Genre (should be ok);
- [x] Keywords;
- [x] Production Company Name (can be ambiguous as a table);
- [x] International Sales Company Name (can be ambiguous as a table);
- [x] Festival Name (can be ambiguous as a table);
- [x] Seller / Sales Agent.

It might not be easy to add some of these attributes to the searchable data because of the data structure. Discuss with Anastasiia if it's too much effort.

Mussuranae:
Add some attribute to be searchable by user on the movie on algolia.
Allow also user to do several queries, I mean that they will be able to look for different characteristics.
Ex : user types `France, action, tribeca` in the search bar, it will look for every movie that has at least one of this characteristics.

Fix #6138

### DIFF
--- a/libs/firebase-utils/src/lib/algolia/algolia.ts
+++ b/libs/firebase-utils/src/lib/algolia/algolia.ts
@@ -121,8 +121,8 @@ export function storeSearchableMovie(
       keywords: movie.keywords ? movie.keywords : [],
       // Register the entire festival name because it will be used for research by users
       festivals: movie.prizes.map(prize => festival[prize.name]) || [],
-      productionCompany: movie.stakeholders.productionCompany.map(company => company.displayName),
-      salesAgent: movie.stakeholders.salesAgent.map(agent => agent.displayName),
+      productionCompany: movie.stakeholders.productionCompany.map(company => company.displayName) || [],
+      salesAgent: movie.stakeholders.salesAgent.map(agent => agent.displayName) || [],
 
       // facets
       genres: movie.genres ? movie.genres : [],
--- a/libs/firebase-utils/src/lib/algolia/algolia.ts
+++ b/libs/firebase-utils/src/lib/algolia/algolia.ts
@@ -1,7 +1,7 @@
 import algoliasearch from 'algoliasearch';
 import { algolia as algoliaClient, centralOrgId } from '@env';
 import * as functions from 'firebase-functions';
-import { Language } from '@blockframes/utils/static-model';
+import { festival, Language } from '@blockframes/utils/static-model';
 import { App, app, getOrgModuleAccess, modules } from "@blockframes/utils/apps";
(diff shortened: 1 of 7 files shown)


libs/firebase-utils/src/lib/algolia/algolia.ts
```diff
@@ -119,6 +119,10 @@ export function storeSearchableMovie(
         movie.directors.map((director) => `${director.firstName} ${director.lastName}`) :
         [],
       keywords: movie.keywords ? movie.keywords : [],
+      // Register the entire festival name because it will be used for research by users
+      festivals: movie.prizes.map(prize => festival[prize.name]) || [],
+      productionCompany: movie.stakeholders.productionCompany.map(company => company.displayName),
+      salesAgent: movie.stakeholders.salesAgent.map(agent => agent.displayName),
```
RemcoSimonides, line 125: Don't you need fallback (` || [] `) for these too in case property stakeholders or productionCompany doesn't exist on the object?

libs/movie/src/lib/movie/form/search.form.ts
```diff
@@ -161,6 +160,16 @@ export class MovieSearchForm extends FormEntity<MovieSearchControl> {
...
     }
 
+    /*
+    Allow the user to use comma or space to separate their research.
+    ex : `France, Berlinale, Action` can be a research but without the `optionalWords`
+    it will be considered as one string/research and not 3 differents.
+    */
+    if (needMultipleQueries) {
+      const multipleQueries: string[] = this.query.value.split(',' || ' ');
+      search['optionalWords'] = multipleQueries;
```
RemcoSimonides, line 170: This can be one line imo

Also, did you perhaps ask Ana whether we want this feature in more search bars? Looks quite useful 😁

RemcoSimonides (approved):

libs/firebase-utils/src/lib/algolia/algolia.ts
Mussuranae: Oh yeah, of course, thanks !

libs/movie/src/lib/movie/form/search.form.ts
Mussuranae: Yes, I ask her if she wanted this for the org list also, and she said for the moment only on the movie list ;) but it won"t take too much time to add it for the other list we can create/have ^^

PR #6346, opened 2021-07-21
Catalog set price to 0 add hint

jeremie74:


jeremie74:
Hi Remco,

I think that yes it was solved. But maybee better check it with François.

PR #6347, opened 2021-07-21
Cypress Custom command for Login

### LINKED ISSUE #4225: Feat (E2E): Use Login App commands instead of UI Operation
Use API for login instead of UI login to speed up E2E test scenarios

xmano:
Closes #4225


xmano:
Tests passing locally!

image

apps/catalog/catalog-e2e/src/index.d.ts
```diff
@@ -0,0 +1,6 @@
+declare namespace Cypress {
+  interface Chainable<Subject> {
+    login(email: string, password: string): Chainable<any>;
```
GrandSchtroumpf, line 3: Maybe specify what the "any". I guess it's always a user or something like that no ?

apps/catalog/catalog-e2e/src/integration/dashboard/import.spec.ts
```diff
@@ -21,18 +19,18 @@ const MOVIE_IMPORT_TIMEOUT = 120 * SEC;
...
+  cy.login(users[0].email, users[0].password);
+  cy.visit('/c/o/marketplace/home');
 
   cy.location('pathname', {timeout: 120 * SEC})
     .should('include', '/marketplace/home');
 
   cy.log("Switch to dashboard home");
   //Reach dashboard home
-  cy.get('catalog-marketplace button[test-id=menu]', {timeout: 3 * SEC})
+  cy.get('catalog-marketplace button[test-id=menu]', {timeout: 60 * SEC})
```
GrandSchtroumpf, line 31: 60sec is a long time no ? Why do we need to wait that long ?

apps/catalog/catalog-e2e/src/support/commands.ts
```diff
@@ -25,3 +28,22 @@ import 'cypress-mailosaur';
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+//Init the app
+const app = firebase.initializeApp(firebaseConfig('catalog'));
+
```
GrandSchtroumpf, line 34: you'll need to connect to the emulator here I think if you want to login with the emulator : 
```typescript
const auth = app.auh();
auth.useEmulator("http://localhost:9099");
```

GrandSchtroumpf: maybe you'll need to have a check on whether or not to use the emulator. Maybe @george43g can answer that.

xmano: I will do a more thorough E2E test with the Emulator later. The scope of this PR is to get the custom command syntax, code structure correct.

We know that the tests are working with the login service command. Next what remains is to make sure that it works with Emulator as well and add any additional code required to handle that.

GrandSchtroumpf (approved):

apps/catalog/catalog-e2e/src/index.d.ts
xmano: import firebase from 'firebase/app';

declare namespace Cypress {
  interface Chainable<Subject> {
    login(email: string, password: string): Chainable<firebase.auth.UserCredential>;
    logout(): Chainable<void>;
  }
}

Issue #6305, opened 2021-07-15
[Backend-Functions] Update offer on contract status change


GrandSchtroumpf:
We want to update the offer status based on the status of all the contracts inside: 
- `pending`: none of the contract has been accepted
- `negotiating`: at least one of the contract is not pending.
- `accepted`: all the contract are accepted
- `signing`: updated on CRM
- `signed`: updated on CRM
- `declined`: all contracts are declined

In `apps/backend-functions/src/contracts.ts` create a function `onContractUpdated` that :
- [x] checks if the status has changed
- [x] if yes, checks the status of the other contract of the offer & the current status of the offer
- [x] if needed updates the offer status to one of the above

In `apps/backend-functions/src/main.ts`
- [x] create a `onContractUpdatedEvent` cloud function trigger below the `onContractDeleteEvent` (you should use the `onDocumentWrite` helper)

PR #6356, opened 2021-07-22
Offer status

pldespaigne:
solve #6305

### DIFF
--- a/apps/backend-functions/src/contracts.ts
+++ b/apps/backend-functions/src/contracts.ts
@@ -1,5 +1,7 @@
 import { db } from './internals/firebase';
-import { Contract } from '@blockframes/contract/contract/+state/contract.model';
+import { Contract, ContractStatus, Mandate, Sale } from '@blockframes/contract/contract/+state/contract.model';
+import { Change } from 'firebase-functions';
+import { Offer } from '@blockframes/contract/offer/+state';
 
 export async function onContractDelete(contractSnapshot: FirebaseFirestore.DocumentSnapshot<Contract>) {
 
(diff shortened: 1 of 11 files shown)


libs/utils/src/lib/pipes/total.pipe.ts
```diff
@@ -4,6 +4,7 @@ import { getDeepValue } from './deep-key.pipe';
 @Pipe({ name: 'total' })
 export class TotalPipe implements PipeTransform {
   transform(data: unknown[], path: string): number {
```
GrandSchtroumpf, line 6: update type `transform(data?: unknown[] | null, path: string)` to reflect your code

apps/backend-functions/src/contracts.ts
```diff
@@ -32,3 +34,48 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
 
   console.log(`Contract ${contract.id} removed`);
 }
+
+
+export async function onContractUpdate(
+  change: Change<FirebaseFirestore.DocumentSnapshot>
+) {
+
+  const before = change.before;
```
GrandSchtroumpf, line 43: you can do : 
```typescript
const before = change.before?.data();
const after = change.after?.data();
if (!before || !after) {
  // Use HttpsError provides a better login system if I recall correctly
  throw new https.HttpsError('failed-precondition', 'Parameter "change" not found');
}
const isSale = before.type === after.type && before.type === 'sale';
const statusChanged = before.status !== after.status;
if (!isSale || !statusChanged) return;
...
```

apps/backend-functions/src/contracts.ts
```diff
@@ -32,3 +34,48 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+  const isSale = contractBefore.type === contractAfter.type && contractBefore.type === 'sale' // contract is of type 'sale'
+  const statusHasChanged = contractBefore.status !== contractAfter.status // contract status has changed
+
+  if (isSale && statusHasChanged) {
+
+    const offerRef = db.doc(`offers/${contractAfter.offerId}`);
+    const offerSnap = await offerRef.get();
+    const offer = offerSnap.data() as Offer;
+
+    if (offer.status !== 'signed' && offer.status !== 'signing') {
```
GrandSchtroumpf, line 63: I like it better when you return after if : 
```typescript
if (offer.status === 'signed' || offer.status === 'signing') return;
```

apps/backend-functions/src/contracts.ts
```diff
@@ -32,3 +34,48 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+
+      const offerContractsQuery = db.collection('contracts')
+        .where('offerId', '==', contractAfter.offerId)
+        .where('type', '==', 'sale');
+      const offerContractsSnap = await offerContractsQuery.get();
+
+      const contractsStatus: ContractStatus[] = offerContractsSnap.docs.map(doc => doc.data().status);
+
+      let newOfferStatus = offer.status;
+      newOfferStatus = contractsStatus.every(status => status !== 'accepted') ? 'pending' : newOfferStatus;
```
GrandSchtroumpf, line 73: You should remove this line. It does nothing. It's make the code too confusing.

apps/backend-functions/src/contracts.ts
```diff
@@ -32,3 +34,48 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+      newOfferStatus = contractsStatus.some(status => status !== 'pending') ? 'negotiating' : newOfferStatus;
+      newOfferStatus = contractsStatus.every(status => status === 'accepted') ? 'accepted' : newOfferStatus;
+      newOfferStatus = contractsStatus.every(status => status === 'declined') ? 'declined' : newOfferStatus;
+
+      offerRef.update({ status: newOfferStatus });
```
GrandSchtroumpf, line 78: no need to update if status hasn't changed : 
```typescript
if (newOfferStatus === offer.status) return;
offerRef.update({ status: newOfferStatus });
```

GrandSchtroumpf (approved):

apps/backend-functions/src/contracts.ts
pldespaigne: We cannot return something to the client as this function is invoked by an event in the db

PR #6362, opened 2021-07-22
Clean Data Model #4853 & #5350 

### LINKED ISSUE #4853: Clean Data Model
After 2.0 release, we spent some times cleaning the code, but we didn't clean the data model yet.
Indeed there is still some legacy field here and there along with fields that can be renamed and/or moved for more clarity.

For example we have a `movie.promotional.financialDetails` that is not used and seems to be duplicated with `campaign.files.financingPlan` (which is used in Mediafinanciers)

### LINKED ISSUE #5350: Remove old link on movie
There is still some old properties on the movie regarding links for video. Now we can upload video, it should be good to clean them everywhere in the project.

phpgeekfr:
fix #4853 
fix #5350


GrandSchtroumpf (approved):
❤️

PR #6378, opened 2021-07-23
Improve performance

GrandSchtroumpf:
Use new `joinWith` custom operator to query subcollections for tables


RemcoSimonides (approved):
Nice - That will make querying multiple collections easier :)

PR #6384, opened 2021-07-26
Add filters to invitation page

jeremie74:


libs/invitation/src/lib/invitation.component.scss
```diff
@@ -15,3 +15,8 @@ p {
 img {
   width: 380px;
 }
+
+form static-select:first-child {
+  padding-right: 24px;
+}
```
pldespaigne, line 21: this should be a margin instead of a padding,
moreover you can simply use `static-select {` instead of `form static-select:first-child`

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -1,34 +1,43 @@
...
+  form = new FormGroup({
+    type: new FormControl(),
+    status: new FormControl(),
+  });
   // Invitation that require an action
-  invitations$ = this.service.myInvitations$.pipe(
+  invitations$ = this.service.myInvitations$.pipe( 
+    switchMap(invitations => this.form.valueChanges.pipe(
+      startWith({ type: [''], status: [''] }),
+      map((filters) => applyFilters(invitations, { type: filters.type, status: filters.status })))),
```
pldespaigne, line 26: use `combineLatest` instead, double `pipe` is really not clean

```TypeScript
newObs$ = combineLatest([
  obsA$,
  obsB$,
]).pipe(
  tap(([valueA, valueB]) => /* ... */),
);
```

libs/utils/src/lib/static-model/types.ts
```diff
@@ -1,34 +1,35 @@
...
+rating,
+scoring,
+shootingPeriod,
+socialGoals,
+soundFormat,
+stakeholderRoles,
+storeStatus,
+territories,
+territoriesISOA3,
+unitBox
```
pldespaigne, line 32: remove these changes, indentation was good

pldespaigne (changes requested):

libs/invitation/src/lib/components/list/list.component.ts
```diff
@@ -10,7 +10,7 @@ import { Invitation } from '../../+state';
   selector: 'invitation-list',
   templateUrl: './list.component.html',
   styleUrls: ['./list.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush
+  //changeDetection: ChangeDetectionStrategy.OnPush
```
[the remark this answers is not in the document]
jeremie74: Hi PL, i don't think you are on the right commit, because i already change and uncomment.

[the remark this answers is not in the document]
jeremie74: Are u on the right commit, because i don't see it commented in my code, and i am updated.

libs/invitation/src/lib/invitation.component.ts
jeremie74: We need to have a switchmap to detect the state changement of the select form. If we combine the two observables to wait for the last emitted values, we can't use switchmap

PR #6385, opened 2021-07-27
fixed horizontal scrollbar bug

SamirTouira:
Result: 

image

Edit: PR up to date. Found a better solution than position relative and absolute.


libs/event/src/lib/components/screening-item/screening-item.component.html
```diff
@@ -1,30 +1,32 @@
 <ng-container *ngIf="event$ | async as event">
-  <a matRipple [routerLink]="['/c/o/marketplace/event/', event.id]">
-    <img fxHide.lt-sm
+  <section fxLayoutAlign="space-between start">
```
GrandSchtroumpf, line 2: remove this section you don't need it. If you want to apply `space-between` on the row, do it on the `:host` selector in the SCSS

GrandSchtroumpf (changes requested):

GrandSchtroumpf (approved):

Issue #6178, opened 2021-07-01
Add filters to invitations page


RemcoSimonides:
- [ ] Add filter by type and status on invitations page
[Zeplin](https://app.zeplin.io/project/60755fb03239554f39def530/screen/60c718df1ee0fa10be95a53d)
Link to discussion in design issue: #4179

phpgeekfr:
@SamirTouira check with @jeremie74 if he can work on this or not

PR #6392, opened 2021-07-27
Add filter invitation 2

jeremie74:


libs/utils/src/lib/static-model/static-model.ts
```diff
@@ -425,8 +425,20 @@ export const orgActivity = {
 
 /** Status of an Organization, set to pending by default when an Organization is created. */
 export const organizationStatus = {
-  pending: 'Pending',
-  accepted: 'Accepted'
+  accepted: 'Accepted',
+  declined: 'Declined',
```
phpgeekfr, line 429: why do you update this const ?

jeremie74: To display on the alphabetic order of the select, without adding a function

phpgeekfr: but you don't use it in your PR

phpgeekfr (approved):
LGTM ;)

RemcoSimonides:
@jeremie74 do you mind referring the issue number like this #6178 in the description of your PRs? That way I can easily find the PR from the issue, and the issue from the PR.

Issue #6370, opened 2021-07-23
Update admin emails


Jackseed:
- [x] A new organization has been created:
- change the title for "[App Name] - [Org name] was created and needs a review"
- change the text for "You can review it here(link to CRM on here)."
- remove completely the unsubscribe and manage email preference links. If not possible, put it as a link on the words "Unsubscribe" & "Manage email preferences"; add 4 to 5 lines between main text and these links.

- [x] New user connexion:
- add app name to the text: 
 " User [User first name] [User last name] connected for the first time to the [App Name]".
- remove completely the unsubscribe and manage email preference links. If not possible, put it as a link on the words "Unsubscribe" & "Manage email preferences"; add 4 to 5 lines between main text and these links.

PR #6396, opened 2021-07-27
6370 update admin emails

RemcoSimonides:
Fix issue #6370


apps/backend-functions/src/internals/users.ts
```diff
@@ -89,6 +89,6 @@ export const createUserFromEmail = async (email: string, createdFrom: App = 'fes
  */
 export const sendFirstConnexionEmail = async (user: PublicUser) => {
   const mailRequest = await userFirstConnexion(user);
-  const from = await getSendgridFrom(user._meta.createdFrom);
-  return sendMail(mailRequest, from).catch(e => console.warn(e.message));
+  const from = getSendgridFrom(user._meta.createdFrom);
+  return sendMail(mailRequest, from, 0).catch(e => console.warn(e.message));
```
Mussuranae, line 93: Maybe put an explanation on the `sendMail` function to explain why you need `0` and in which case we need to put it. For future coworkers to understand it easily.

Mussuranae (approved):

PR #6399, opened 2021-07-27
Mandate Migration

pldespaigne:
Some mandates have missing `sellerId` & `stakeholders`, this migration will fill those properties with the required values.


libs/firebase-utils/src/lib/firestoreMigrations/0066.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (isRogueMandate) return;
+
+    let update = false;
+
+    if (!mandate.sellerId) {
+      update = true;
+      mandate.sellerId = centralOrgId.catalog;
+    }
+
+    if (!mandate.stakeholders || !mandate.stakeholders.length) {
```
GrandSchtroumpf, line 35: you can do `if (!mandate.stakeholders?.length)`

GrandSchtroumpf (approved):

Issue #6407, opened 2021-07-28
Status "in negotiation" appears to be a way too long. See this with UI team


iPR0D1G1:

iPR0D1G1:
<img width="856" alt="Bildschirmfoto 2021-07-28 um 10 23 25 AM" src="https://user-images.githubusercontent.com/22674720/127298256-e0b506ce-5caa-4526-a136-3de3551470e3.png">

GrandSchtroumpf:
The real solution is to use css propery `white-space: nowrap`. See my PR: https://github.com/blockframes/blockframes/pull/6412

PR #6408, opened 2021-07-28
refactor offer status

iPR0D1G1:
Fixed: #6407


RemcoSimonides (approved):

GrandSchtroumpf:
This is not the right way to fix the issue. See my comment on the issue: https://github.com/blockframes/blockframes/issues/6407

PR #6416, opened 2021-07-28
Invitations filters issue

jeremie74:
Adding message when invitations empty on load or on filters


libs/invitation/src/lib/invitation.component.html
```diff
@@ -10,23 +9,27 @@ <h1>Invitations</h1>
...
+      </button>
       <button mat-button test-id="invitation-accept-all" (click)="acceptAll(invitations)">
         <mat-icon svgIcon="check_circle"></mat-icon>
         <span>Accept All</span>
       </button>
     </header>
-    <invitation-list [invitations]="invitations" class="surface"></invitation-list>
-  </ng-container>
+    <invitation-list *ngIf="!emptyMessage" [invitations]="invitations" class="surface"></invitation-list>
+    <ng-container *ngIf="emptyMessage">
```
phpgeekfr, line 22: use ng-template instead.
```
<invitation-list *ngIf="!emptyMessage else empty" [invitations]="invitations" class="surface">
<ng-template #empty>
 <section fxLayout="column" fxLayoutAlign="center center">
        <h1>You have no invitation yet</h1>
        <p>Reload the page or click on the clear filter button above</p>
        <img asset="empty_invitation.svg" alt="No invitations">
      </section>
</ng-template>
```

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -20,16 +20,19 @@ const applyFilters = (invitations: Invitation[], filters: { type: string[], stat
   styleUrls: ['./invitation.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class InvitationComponent {
+
```
phpgeekfr, line 23: don't add new lines here

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -46,6 +53,14 @@ export class InvitationComponent {
     private orgQuery: OrganizationQuery
   ) { }
 
+  ngOnInit() {
+    this.subscribe = this.invitations$.subscribe(val => val.length ? this.emptyMessage = false : this.emptyMessage = true)
```
phpgeekfr, line 57: use pipe instead, this will prevent you to unsubscribe in onDestroy

phpgeekfr: something like
```
this.emptyMessage$ = this.invitations$.pipe(map(inv => !inv.length))
```

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -59,4 +74,12 @@ export class InvitationComponent {
     const [moduleAccess = 'dashboard'] = getOrgModuleAccess(org, app);
     return this.router.navigate([`/c/o/${moduleAccess}/home`]);
   }
+
+  clearSelect (formName: string[]) {
+    formName.map(val => this.form.get(val).reset());
```
phpgeekfr, line 79: don't use a map here.. you don't return anyting

phpgeekfr: why can't you reset the whole form instead of each control individually ?

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -59,4 +74,12 @@ export class InvitationComponent {
...
+  }
+
+  clearFilters() {
+    this.clearSelect (this.formName);
```
phpgeekfr, line 83: why create a new function `clearSelect` that is only called here ?

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -20,16 +20,19 @@ const applyFilters = (invitations: Invitation[], filters: { type: string[], stat
...
     status: new FormControl([]),
   });
+
   // Invitation that require an action
   invitations$ = combineLatest([
     this.service.myInvitations$,
     this.form.valueChanges.pipe(startWith({ type: [], status: []  }))
-  ]).pipe(
+  ])
+  .pipe(
```
phpgeekfr, line 35: previous formatting was the correct one

libs/invitation/src/lib/invitation.component.html
```diff
@@ -10,23 +9,27 @@ <h1>Invitations</h1>
           <mat-label>Filter by status</mat-label>
         </static-select>
       </form>
+      <button mat-button (click)="clearFilters()" matTooltip="Clear fields">
+        <mat-icon svgIcon="refresh_filters"></mat-icon>
+        Clear filters
```
GrandSchtroumpf, line 14: In the application we've got a global style that applies 8px margin between mat-icon and span. So whenever you use a `mat-icon` next to a text, always put this text into a `span`.

libs/invitation/src/lib/invitation.component.ts
GrandSchtroumpf: The goal is to levarage the `async` pipe of angular. The pipe will automatically manage subscription & unsubscription when it needs it. Also it'll run a `markForCheck` whenever the value is change so it works perfectly even with an OnPush changeDetection strategy.
Only use `subscribe` when you want to update something that is *not* in the template (as you cannot use the `async` pipe).
So it should look like that : 
```typescript
formName = ['type', 'status'];
// Here I'm putting the code directly in the property definition, I don't need to do it in the ngOnInit
emptyMessage = this.invitations$.pipe(map(map(inv => !inv.length));
```
And in the template: 
```html
<ng-container *ngIf="emptyMessage$ | async"
```
---
Now in your case a better solution would be to do the logic directly in the template, without having to create a new observable : 
```html
<ng-container *ngIf="invitations$ | async as invitations; else loading">
  ...
  <ng-container *ngIf="invitations.length else empty">
  
  </ng-container>
</ng-container>
<ng-template #empty>
  
</ng-template>
<ng-template #loading>
  
</ng-template>
```

This is a very common pattern when working with list : 
- show #loading if it's loading
- show #empty if the array is empty
- else show the content

libs/invitation/src/lib/invitation.component.ts
jeremie74: i use map to browse the table and reset all the forms. But i can do an iteration

jeremie74: why can't you reset the whole form instead of each control individually ?
Yes i can do that

jeremie74: form.reset();

libs/invitation/src/lib/invitation.component.html
```diff
@@ -10,14 +10,31 @@ <h1>Invitations</h1>
           <mat-label>Filter by status</mat-label>
         </static-select>
       </form>
+      <button mat-button (click)="clearFilters()" matTooltip="Clear fields">
+        <mat-icon svgIcon="refresh_filters"></mat-icon>
+        <span>Clear filters</span>
+      </button>
```
phpgeekfr, line 16: Clear button should be with other inputs, not in middle of page
image

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -21,23 +21,27 @@ const applyFilters = (invitations: Invitation[], filters: { type: string[], stat
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class InvitationComponent {
+  formName = ['type', 'status'];
```
phpgeekfr, line 24: used ?

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -21,23 +21,27 @@ const applyFilters = (invitations: Invitation[], filters: { type: string[], stat
...
   form = new FormGroup({
     type: new FormControl([]),
     status: new FormControl([]),
-  });
+  })
+  // Invitation count for conditions
+  invitationCount$ = this.service.myInvitations$.pipe(map(inv => inv.length))
```
phpgeekfr, line 30: missing `;` and maybe add a new line between  variable declarations

libs/invitation/src/lib/invitation.component.ts
jeremie74: No it's true, i delete it

libs/invitation/src/lib/invitation.component.ts
jeremie74: Ok

libs/invitation/src/lib/invitation.component.html
```diff
@@ -10,14 +10,31 @@ <h1>Invitations</h1>
...
       <button mat-button test-id="invitation-accept-all" (click)="acceptAll(invitations)">
         <mat-icon svgIcon="check_circle"></mat-icon>
         <span>Accept All</span>
       </button>
     </header>
-    <invitation-list [invitations]="invitations" class="surface"></invitation-list>
+    <ng-container *ngIf="invitations.length">
+      <invitation-list [invitations]="invitations" class="surface"></invitation-list>
```
phpgeekfr, line 23: I'm not sure that class="surface" is used

libs/invitation/src/lib/invitation.component.html
jeremie74: Ok it's done

libs/invitation/src/lib/invitation.component.html
```diff
@@ -10,14 +10,31 @@ <h1>Invitations</h1>
           <mat-label>Filter by status</mat-label>
         </static-select>
       </form>
-      <button mat-button test-id="invitation-accept-all" (click)="acceptAll(invitations)">
+      <button mat-button (click)="clearFilters()" matTooltip="Clear fields">
+        <mat-icon svgIcon="refresh_filters"></mat-icon>
+        <span>Clear filters</span>
+      </button>
+      <button mat-button test-id="invitation-accept-all" class="left" (click)="acceptAll(invitations)">
```
phpgeekfr, line 17: you can do this without creating a class, with fxLayout

libs/invitation/src/lib/invitation.component.scss
```diff
@@ -19,3 +19,7 @@ img {
 static-select {
   margin-right: 24px;
 }
+
+.left {
+  margin-left: auto;
+}
```
phpgeekfr, line 25: use fxLayout as the other 2 existing inputs

libs/invitation/src/lib/invitation.component.html
```diff
@@ -10,14 +10,31 @@ <h1>Invitations</h1>
...
     </header>
-    <invitation-list [invitations]="invitations" class="surface"></invitation-list>
+    <ng-container *ngIf="invitations.length">
+      <invitation-list [invitations]="invitations" class="surface"></invitation-list>
+    </ng-container>
+    <ng-container *ngIf="!invitations.length">
+      <section fxLayout="column" fxLayoutAlign="center center">
+        <p>You have no invitation with this filter.<br>Click on <b>clear filters</b> button to reset.</p>
+        <img asset="empty_invitation.svg" alt="No invitations">
+        <button mat-button (click)="clearFilters()" matTooltip="Clear fields">
```
phpgeekfr, line 29: maybe you can directly do:  <button mat-button (click)="form.reset()" matTooltip="Clear fields">

libs/invitation/src/lib/invitation.component.html
GrandSchtroumpf: surface is a global class that looks like mat-card, without all the burden of mat-card. It's a legit class ;)

phpgeekfr: ok thanks, I did not saw it in component's scss

libs/invitation/src/lib/invitation.component.html
jeremie74: I can't isolate an element with flex to be align in a different position of others. That's why i use css. I need them to be on the same row.

jeremie74: Surface is used for the white background

jeremie74: I found that fxFlexOffset="" i can try it

libs/invitation/src/lib/invitation.component.scss
```diff
@@ -19,3 +19,7 @@ img {
 static-select {
   margin-right: 24px;
 }
+
+.left {
```
phpgeekfr, line 23: still needed ?

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, ChangeDetectionStrategy } from '@angular/core';
+import { Component, ChangeDetectionStrategy} from '@angular/core';
```
phpgeekfr, line 1: why ? use good formatting

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -24,20 +24,25 @@ export class InvitationComponent {
   form = new FormGroup({
     type: new FormControl([]),
     status: new FormControl([]),
-  });
+  })
```
phpgeekfr, line 27: missing ;

libs/invitation/src/lib/invitation.component.scss
jeremie74: no, deleted..

libs/invitation/src/lib/invitation.component.html
phpgeekfr: so you don't do it ?

phpgeekfr (approved):
LGTM

PR #6421, opened 2021-07-29
Download JWP video into storage

pldespaigne:
Create a backend functions that is able to download a video from jwp into bucket storage


apps/backend-functions/src/rescue.ts
```diff
@@ -0,0 +1,104 @@
...
+      res.on('data', chunk => body += chunk);
+      res.on('end', () => {
+        const response = JSON.parse(body);
+        resolve(response);
+      });
+		}).on('error', e => {
+		    reject(e);
+		});
+	});
+}
```
GrandSchtroumpf, line 23: just for formatting : 
```typescript
function fetchJSON(url: string): Promise<any> {
	return new Promise((resolve, reject) => {
		get(url, res => {
      let body = '';
      res.on('data', chunk => body += chunk);
      res.on('end', () => resolve(JSON.parse(body)));
		})
		.on('error', e => reject(e));
	});
}
```

apps/backend-functions/src/rescue.ts
```diff
@@ -0,0 +1,104 @@
...
+
+function fetchRedirect(url: string): Promise<string> {
```
GrandSchtroumpf, line 25: can you add a comment on that ? I'm not sure to understand

apps/backend-functions/src/rescue.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  // GET THE LIST OF ENCODED FILE FOR A GIVEN VIDEO
+
+  const apiResult = await fetchJSON(url);
+
+  // FILTER THE RESPONSE TO GET THE URL OF THE HIGHEST DEFINITION VIDEO FILE
+
+  const fileName = apiResult.title;
+  const videos = apiResult.playlist[0].sources.filter(source => source.type === 'video/mp4').sort();
+  const hdVideo = videos.sort((a, b) => b.filesize - a.filesize)[0];
```
GrandSchtroumpf, line 89: What about : 
```typescript
const [video] = apiResult.playlist[0].sources
  .filter(source => source.type === 'video/mp4')
  .sort((a, b) => b.filesize - a.filesize);
if (!video) ....
const fileUrl = video.file;
```

GrandSchtroumpf (approved):

Issue #5799, opened 2021-05-17
Image modale - add buttons to navigate


Jackseed:
Page: Marketplace movie view
Section Images (bottom)
Url: `/c/o/marketplace/title/g8xnPDdQMFXsSkr1WyZj/main`

### Description
We added a modale to zoom in an image when clicked. 
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/0ebef7cf-e6c2-4294-aba9-aa8af17aab6d)

An update of this component is to add enough margins on the sides of the modale to add previous & next arrows that enables users to navigate through images. We also should have a cross icon to close the modale
Arrows should be the same as the one used in the image carrousel
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/e552fe7b-cdfc-43bf-9216-23218c2fc193)

### Acceptance criteria
- [ ] Left & right arrows in the modale margins
- [ ] Left go to previous image, right to next
- [ ] If first or last image is displayed hide left or right arrow respectively 
- [ ] Cross icon that closes the modale

### Wireframe
[figma](https://www.figma.com/file/vBJGpcE68Fyc8IPNEi5slY/%5B-Media-Ressources-%5D?node-id=1177%3A8481)
![image](https://user-images.githubusercontent.com/14964443/118469627-29913100-b706-11eb-8a58-51ed9152f9f1.png)

PR #6424, opened 2021-07-29
5799 navigate next and previous in file preview modal

RemcoSimonides:
fix issue #5799

### DIFF
--- a/libs/media/src/lib/file/explorer/explorer.component.ts
+++ b/libs/media/src/lib/file/explorer/explorer.component.ts
@@ -16,8 +16,6 @@ import { AngularFirestore, QueryFn } from '@angular/fire/firestore';
 import { Organization } from '@blockframes/organization/+state';
 import { FileUploaderService, MediaService } from '@blockframes/media/+state';
 import { createStorageFile, StorageFile } from '@blockframes/media/+state/media.firestore';
-import { FilePreviewComponent } from '../preview/preview.component';
-import { MatDialog } from '@angular/material/dialog';
 import { getFileMetadata } from '@blockframes/media/+state/static-files';
 
 function getDir(root: Directory, path: string) {
@@ -72,8 +70,7 @@ export class FileExplorerComponent implements OnInit, AfterViewInit {
     private movieService: MovieService,
     private mediaService: MediaService,
     private service: FileUploaderService,
-    private routerQuery: RouterQuery,
-    private dialog: MatDialog,
+    private routerQuery: RouterQuery
   ) {}
 
   ngOnInit() {
@@ -121,13 +118,6 @@ export class FileExplorerComponent implements OnInit, AfterViewInit {
 
   getMeta(dir: FileDirectoryBase, index: number) {
     return [ ...dir.meta, index ];
-  }
-
-  openView(item: Partial<StorageFile>, event: Event) {
-    event.stopPropagation();
-    if (item) {
-      this.dialog.open(FilePreviewComponent, { data: { ref: item }, width: '80vw', height: '80vh', autoFocus: false });
-    }
   }
 
   async downloadFile(item: StorageFile, event: Event) {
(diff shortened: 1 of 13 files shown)


libs/media/src/lib/file/explorer/explorer.component.ts
```diff
@@ -123,13 +120,6 @@ export class FileExplorerComponent implements OnInit, AfterViewInit {
     return [ ...dir.meta, index ];
   }
 
-  openView(item: Partial<StorageFile>, event: Event) {
-    event.stopPropagation();
-    if (item) {
-      this.dialog.open(FilePreviewComponent, { data: { ref: item }, width: '80vw', height: '80vh', autoFocus: false });
-    }
-  }
-
```
RemcoSimonides, line 132: code wasn't used

libs/media/src/lib/file/preview/preview.component.html
```diff
@@ -4,18 +4,28 @@
...
-    <p>Currently supported file are: pdf, images, and videos</p>
+<article fxLayout fxLayoutAlign="start center">
+  <button mat-icon-button (click)="previous()" matTooltip="Previous" *ngIf="index !== 0">
+    <mat-icon svgIcon="chevron_left"></mat-icon>
+  </button>
+  <ng-container [ngSwitch]="ref.storagePath | fileType">
+    <ng-container *ngSwitchCase="'image'">
+      <div>
+        <img [ref]="ref" asset="no_titles.svg" />
+      </div>
```
RemcoSimonides, line 15: This div helps the img to not go too wide

GrandSchtroumpf: When you arrive at this point, it's because there is something weird with how you do things. Please reconsider the problem. Can you set the width and height in px bases on the type of image for example ?

libs/media/src/lib/file/preview/preview.component.scss
```diff
@@ -3,8 +3,17 @@
...
 
-img {
-  object-fit: contain;
-  width: 100%;
+article {
   height: 100%;
+
+  img {
+    object-fit: contain;
+    width: 100%;
```
GrandSchtroumpf, line 11: you should never set the width & height of the img at the same time, it case break the ratio.

libs/media/src/lib/file/preview/preview.component.ts
```diff
@@ -10,10 +10,38 @@ import { StorageFile } from '@blockframes/media/+state/media.firestore';
...
+  ref: StorageFile;
+
+  refs: StorageFile[];
+  index: number = 0;
+
   constructor(
     private dialogRef: MatDialogRef<FilePreviewComponent>,
-    @Inject(MAT_DIALOG_DATA) public data: { ref: StorageFile }
-  ) { }
+    @Inject(MAT_DIALOG_DATA) public data: { ref: StorageFile } | { refs: StorageFile[], index: number }
```
GrandSchtroumpf, line 21: For list we should have another component.

GrandSchtroumpf (changes requested):

libs/media/src/lib/file/preview-list/preview-list.component.scss
```diff
@@ -0,0 +1,19 @@
...
+article {
+  height: 100%;
+
+  div {
+    min-width: $button-width;
+  }
+
+  img {
+    object-fit: contain;
+    width: calc(100% - #{$button-width * 2});
```
RemcoSimonides, line 17: too advanced? 😅
img is still taking full width of dialog instead of taking the buttons next to it in account. So I decrease the full width with 2x button width.

libs/media/src/lib/file/preview/preview.component.scss
```diff
@@ -6,5 +6,4 @@
 img {
```
GrandSchtroumpf, line 6: could you update the code for the preview component too ?

GrandSchtroumpf (approved):

Issue #6433, opened 2021-07-30
Catalog bugs.


iPR0D1G1:
Main issue: #6383
- [x]  Background color is not the good one (Safari Specific). Should be the same as in Chrome version.
from
<img src="https://camo.githubusercontent.com/0d473e55a315e9b2e443d5b70b1aeb04e0b465d88dd732d1c97d66de7d83c8be/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3564643534353238633063636433303030313336373263332f31656439313931652d633830632d343430332d383038612d393964613737663733313834" />

to
<img src="https://camo.githubusercontent.com/5c1f97533ac1f01fdeee3fe923eee435c35d567872ec8a3f3b63beebc33cbe93/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3564643534353238633063636433303030313336373263332f36376238393965372d386331342d346331342d393433332d656266343962646365306266" />

## Contract List

- [x]  If possible, by default order the contracts by the recent date. (The recent on the top)
<img src="https://camo.githubusercontent.com/55a7c20de252de7e246191d131c525d51be34cbf6e992ab91895596becf7d432/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3564643534353238633063636433303030313336373263332f39343339313532662d663333662d346463342d383166302d353034306433616331333035"/>

## Selection Page

- [x] If no versions added, put a "-"
<img src="https://camo.githubusercontent.com/fd49c7e910e443a8229bf42dbd1c3d4c459357c8d3e40ccd0138cecf80611417/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3564643534353238633063636433303030313336373263332f30653735623732632d633562652d346236662d393134652d643835353662656564366364"/>

- [x] If possible, put the recent offers on the first place. Exemple: 07/27/2021 should be on the first line, then 07/26/2021 and 07/23/2021 on the last place.
<img src="https://camo.githubusercontent.com/14b27224bf11e99c96871b82b7070f95e85d400a61144f6f760696cfb33d3f83/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3564643534353238633063636433303030313336373263332f38373933646539362d626661302d343361382d383933382d313236623735653436383236"/>
- [x] If possible, need to align the infos in table
[...]

PR #6437, opened 2021-07-30
feat:sort contract and offer and deals per date created in descending…

iPR0D1G1:
Fixed: #6433


apps/catalog/catalog/src/app/marketplace/offer/list/list.component.ts
```diff
@@ -15,7 +15,9 @@ import { switchMap } from 'rxjs/operators';
 })
 export class ListComponent {
   offers$ = this.orgQuery.selectActiveId().pipe(
-    switchMap(orgId => this.service.valueChanges(ref => ref.where('buyerId', '==', orgId))),
+    switchMap(orgId => this.service.valueChanges(
```
GrandSchtroumpf, line 18: To make it cleaner you can do : 
```typescript
const query: Queryfn = ref => ref.where('buyerId', '==', orgId).orderBy('_meta.createdAt', 'desc');
offers$ = this.orgQuery.selectActiveId().pipe(
  switchMap(orgId => this.service.valueChanges(query)),
  joinWith({
    contracts: offer => this.getContracts(offer.id)
  })
)
```

apps/catalog/catalog/src/app/marketplace/offer/shell/shell.component.scss
```diff
@@ -7,7 +7,8 @@ header {
   display: flex;
   flex-direction: column;
   align-items: start;
-  background-image: linear-gradient(to bottom, var(--primary-lighter), transparent);
+  // https://stackoverflow.com/questions/46309641/what-happens-in-safari-with-the-transparent-color
+  background-image: linear-gradient(to bottom, var(--primary-lighter), #fff0);
```
GrandSchtroumpf, line 11: You should *never* hardcode a color, else it'll break on the other theme. Use the right css variable, for **example** : 
```scss
background-image: linear-gradient(to bottom, var(--primary-lighter), var(--background));
```

libs/contract/src/lib/contract/item/contract-item.component.html
```diff
@@ -39,11 +39,12 @@ <h3>{{ title.directors | displayName }}</h3>
         </ng-template>
       </ng-template>
       <ng-template colRef="languages" let-languages>
-        <ul *ngIf="languages">
+        <ul *ngIf="languages && languages.length > 0; else noLanguage">
```
GrandSchtroumpf, line 42: You can directly do : 
```html
<ul *ngIf="languages?.length > 0; else noLanguage">
```

GrandSchtroumpf (changes requested):

apps/catalog/catalog/src/app/marketplace/offer/list/list.component.ts
```diff
@@ -15,12 +16,16 @@ import { switchMap } from 'rxjs/operators';
 })
 export class ListComponent {
   offers$ = this.orgQuery.selectActiveId().pipe(
-    switchMap(orgId => this.service.valueChanges(ref => ref.where('buyerId', '==', orgId))),
+    switchMap(orgId => this.service.valueChanges(this.query(orgId))),
```
GrandSchtroumpf, line 19: do not create a method if you don't depend on the state (`this`), you can put the function outside of the component: 
```typescript
const query = (orgId: string): QueryFn  => ref => ref.where('buyerId', '==', orgId).orderBy('_meta.createdAt', 'desc');
```

GrandSchtroumpf (approved):

PR #6446, opened 2021-08-16
Fix #6386

### LINKED ISSUE #6386: window is undefined error in backend functions
We can't check `'Cypress' in window` in back-end.
Let's save whether Cypress is used in _META collection and toggle the setting when starting a e2e command in cli

xmano:
Fixes #6386


xmano:
Will correct the branch issue and re-submit

apps/backend-functions/src/templates/mail.ts
```diff
@@ -363,7 +363,7 @@ export async function organizationRequestedAccessToApp(org: OrganizationDocument
 }
 
 export async function userFirstConnexion(user: PublicUser): Promise<EmailRequest> {
-  const supportEmail = ('Cypress' in window) ? testEmail : getSupportEmail(user._meta.createdFrom);
+  const supportEmail = (e2e_mode) ? testEmail : getSupportEmail(user._meta.createdFrom);
```
RemcoSimonides, line 366: No need for `()` around e2e_mode

RemcoSimonides (commented):
What happened with the idea of setting a variable when starting an e2e process instead of defining it statically?

apps/backend-functions/src/templates/mail.ts
xmano: Ok. will take care of this along with other review comments if any.

phpgeekfr:
@xmano I cleaned branch history so this should be clean. But ask me again before merging please

apps/backend-functions/src/templates/mail.ts
xmano: @george43g : plesae let me know if using env config is ok or we want more changes like setting a parameter during e2e config.

xmano:
> @xmano I cleaned branch history so this should be clean. But ask me again before merging please

Sure!

george43g:
hey mate,

apps/backend-functions/src/environments/environment.e2e.ts

create the above file and have it as a build config. 
so

nx build backend-functions --configuration=e2e ...(or =emulator, depending on the real purpose of these changes)

should select that build config (using that file with a fileReplacement) against any projectId

xmano:
I made changes to backend-functions environment. However due to some pre-commit issue I cannot commit.

Here are the main diffs:

```
diff --git a/angular.json b/angular.json
index e57baed21..0e88ec694 100644
--- a/angular.json
+++ b/angular.json
@@ -208,7 +208,7 @@
                 },
                 {
                   "replace": "apps/backend-functions/src/environments/environment.ts",
-                  "with": "apps/backend-functions/src/environments/environment.prod.ts"
+                  "with": "apps/backend-functions/src/environments/environment.ci.ts"
                 }
               ]
             }
```

backend-function environment

```
diff --git a/apps/backend-functions/src/environments/environment.prod.ts b/apps/backend-functions/src/environments/environment.prod.ts
index 20ab18f51..ede1224c6 100644
--- a/apps/backend-functions/src/environments/environment.prod.ts
+++ b/apps/backend-functions/src/environments/environment.prod.ts
@@ -6,6 +6,8 @@
  */

+export const e2e_mode = false;
+
 export {
   factoryContract,
   backupBucket,
```

xmano:
@george43g : kindly review the diffs posted as comment as above. I would like to work on this and close it.

george43g:
you have made the default mode e2e mode. We still havent fully transitioned to this... 
also, you have not referenced your new .ci file in angular.json. This mode is to be used during e2e testing correct?

Issue #6240, opened 2021-07-07
Update Archipel Market Landing page


RemcoSimonides:
[Zeplin](https://app.zeplin.io/project/5de7944cf85691577aacfbb7/screen/60c9c2ef572ec78d4d00e600)

![image](https://user-images.githubusercontent.com/27687382/129704914-d54ecdbf-3c5f-4bf2-b297-0119888e211d.png)

Link to original design issue: #4127

Jackseed:
@maltmayer should we have Security as a benefit?

RemcoSimonides:
1:
![image](https://user-images.githubusercontent.com/27687382/129745090-c2c44d54-92e1-4178-b01d-d9216b67a84e.png)
- [ ] Update image
- [ ] Add "One tool for your year-round 
promotion & acquisitions" text

2;
![image](https://user-images.githubusercontent.com/27687382/129745479-fc2aea43-38f2-4cad-9dfc-c40cfce14122.png)
- [ ] Add this new banner

3:
![image](https://user-images.githubusercontent.com/27687382/129745620-6416d00b-43c4-431f-91fe-5a2eaf4760ae.png)
- [ ] Add description underneath title
- [ ] Sales Agent tab
- [ ] Buyer tab (I'll ask Stephanie to update this)

4:
![image](https://user-images.githubusercontent.com/27687382/129746964-9de854c9-251e-439e-b5ca-0a8797784bcf.png)

5:
![image](https://user-images.githubusercontent.com/27687382/129747062-4b9e2897-fcc3-4c05-b019-dc5fd5456212.png)
Remco: Ask Stephanie for updated text for reviews

6:
![image](https://user-images.githubusercontent.com/27687382/129747225-2add309b-0072-4805-a923-20b23542562d.png)
- [ ]

RemcoSimonides:
Created new issue for subscribe to newsletter functionality #6640

PR #6476, opened 2021-08-18
#6240 Update Archipel market landig page - 1

Poppe-A:


apps/festival/festival/src/app/landing/landing.component.html
```diff
@@ -51,14 +52,15 @@ <h3>{{ feature.title }}</h3>
         </article>
       </mat-tab>
     </mat-tab-group>
-    <section fxLayout="column" fxLayoutAlign="start center" class="join" fxLayoutGap="24px">
-      <p class="mat-display-2">Interested? Join now!</p>
-      <a mat-flat-button color="accent" routerLink="/auth/identity">
-        <span>Sign up</span>
+  </landing-content>
+
+  <landing-join fxLayout="column" fxLayoutAlign="start center" class="join" fxLayoutGap="32px">
```
RemcoSimonides, line 57: `class="join"` isn't used anymore so it can be removed

apps/festival/festival/src/app/landing/landing.component.scss
```diff
@@ -39,10 +37,11 @@ landing-content {
       }
     }
   }
+}
 
-  .join {
-    padding: 48px 0;
-  }
+landing-join {
+  padding: 100px 0;
```
RemcoSimonides, line 43: padding can be defined in the parent component.
If you check `shell.component.scss` you see the same is done for `landing-content` and `landing-header`

libs/landing/src/lib/shell/shell.component.html
```diff
@@ -24,6 +24,10 @@
 
   <ng-content select="landing-content"></ng-content>
 
+  <!-- Join -->
+
+  <ng-content select="landing-join"></ng-content>
```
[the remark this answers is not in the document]
Poppe-A: I've added a content-projection because the "join" part now needs to be full width (and the landig content is not). 
For me the other possibility would be to keep it in the implementation, and to modify the style to:
position: absolute;
left: 0;
width: 100vw

I don't know which solution is the cleanest

RemcoSimonides (approved):
LGTM
I assume we fix the width later?
image

Issue #6472, opened 2021-08-18
[CATALOG] Contract-Item display other user's holdbacks


pldespaigne:
Add a button at the bottom of the `contract-item` component. When the button is clicked it should open a modal that display the list of holdbacks in the contract.
- Hide the button if there is no holdbacks.
- The modal should dispaly all the holdbacks of the contract, we don't care if they are colliding or not.

![image](https://user-images.githubusercontent.com/5858657/129868939-19cda558-f465-41ca-9f0f-473ee25eece7.png)

PR #6477, opened 2021-08-18
feat(crm,catalog):content projected list of existing holdbacks to contract-item

iPR0D1G1:
fixed:  #6472


apps/catalog/catalog/src/app/marketplace/title/selection/selection.component.ts
```diff
@@ -27,6 +27,7 @@ export class MarketplaceSelectionComponent {
   private prices: number[] = [];
   priceChanges = new Subject<number>();
   total$: Observable<number>;
+  orgId:string;
```
GrandSchtroumpf, line 30: you can initialize it directly here : 
```typescript
orgId = this.orgQuery.getActiveId();
```

libs/contract/src/lib/contract/holdback/form/form.component.ts
```diff
@@ -15,6 +16,7 @@ import { HoldbackForm } from '../form';
 })
 export class HolbackFormComponent {
   @Input() title: Movie;
+  @Input() otherOrgHoldBacks:Holdback[]=[];
```
GrandSchtroumpf, line 19: Rename to `@Input() existingHoldbacks?: Holdback[]`.
We use the "?" here because we don't want to display anything if there is no existingHoldbacks

libs/contract/src/lib/contract/holdback/form/form.component.html
```diff
@@ -1,6 +1,9 @@
 <holdback-list [holdbacks]="holdbacks"></holdback-list>
 
-<div class="actions">
+<div class="actions" fxLayout [fxLayoutAlign]="otherOrgHoldBacks?.length ? 'space-between center': 'center center'">
+  <button mat-flat-button *ngIf="otherOrgHoldBacks?.length" (click)="openHoldbackModal()">
```
GrandSchtroumpf, line 4: If you don't initialize the input you can do : `*ngIf="existingHoldbacks"`

libs/contract/src/lib/contract/holdback/modal/holdback-modal.component.html
```diff
@@ -7,7 +7,7 @@ <h1>Holdbacks exist for this right</h1>
 <section>
   <p>
     There are some holdbacks that exists for this right.<br/>
-    You can still add this right to your selection.
+   {{data.addingToSelection?" You can still add this right to your selection.":''}}
```
GrandSchtroumpf, line 10: Use ngIf instead : 
```typescript
<span>There are some holdbacks that exists for this right.</span>
<ng-container *ngIf="data.withWarning">
  <br/>
  <span>You can still add this right to your selection.</span>
</ng-container>
```

libs/contract/src/lib/contract/holdback/modal/holdback-modal.component.ts
```diff
@@ -12,7 +12,7 @@ import { Holdback } from '../../+state';
 export class HoldbackModalComponent {
 
   constructor(
-    @Inject(MAT_DIALOG_DATA) public data: { holdbacks: Holdback[] },
+    @Inject(MAT_DIALOG_DATA) public data: { holdbacks: Holdback[], addingToSelection:boolean },
```
GrandSchtroumpf, line 15: change `addingToSelection` to `withWarning`

libs/movie/src/lib/movie/pipes/get-title-holdbacks.ts
```diff
@@ -0,0 +1,30 @@
...
+    private service: ContractService
+  ) { }
+
+  async transform(titleId: string, excludedOrgs: string[] = [''], ): Promise<Holdback[]> {
+    return this.service.getValue(
+      ref => ref.where('titleId', '==', titleId)
+        .where('status', '==', 'accepted')
+        .where('type', '==', 'sale')
+        .where('buyerId', 'not-in', excludedOrgs)
+    ).then(
```
pldespaigne, line 18: prefer `async`/`await` over `then`

pldespaigne (changes requested):

libs/contract/src/lib/contract/holdback/modal/holdback-modal.component.ts
```diff
@@ -12,7 +12,7 @@ import { Holdback } from '../../+state';
 export class HoldbackModalComponent {
 
   constructor(
-    @Inject(MAT_DIALOG_DATA) public data: { holdbacks: Holdback[] },
+    @Inject(MAT_DIALOG_DATA) public data: { holdbacks: Holdback[], withWarning:boolean },
```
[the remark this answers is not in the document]
GrandSchtroumpf: Hmmm I would recommend to make the `withWarning` as optional instead. If it's undefined, then it'll work as if it were `false`

libs/movie/src/lib/movie/pipes/get-title-holdbacks.ts
```diff
@@ -0,0 +1,26 @@
...
+import { ContractService, Holdback, Sale } from '@blockframes/contract/contract/+state';
+
+@Pipe({ name: 'getTitleHoldbacks' })
+export class GetTitleHoldbacksPipe implements PipeTransform {
+
+  constructor(
+    private service: ContractService
+  ) { }
+
+  async transform(titleId: string, { excludedOrgs = [] }: { excludedOrgs: string[] },): Promise<Holdback[]> {
```
GrandSchtroumpf, line 11: titleId can be null in your case above. If it's null, return `of(undefined)`;

libs/contract/src/lib/contract/item/contract-item.component.html
```diff
@@ -52,4 +52,9 @@ <h3>{{ title.directors | displayName }}</h3>
     </bf-table-filter>
   </ng-container>
   <ng-content select="footer"></ng-content>
+  <ng-container *ngIf="contract?.titleId| getTitleHoldbacks:{excludedOrgs:[orgId]} | async as existingHoldbacks">
```
GrandSchtroumpf, line 55: This pipe is getting a little bit out of control ^^. Can you rename : 
```html
<ng-container *ngIf="contract?.titleId | getExistingHoldbacks:orgId | async as holdbacks">
```
Don't try to see future uses. Keep it simple and stupid

libs/movie/src/lib/movie/pipes/get-title-holdbacks.ts
```diff
@@ -0,0 +1,26 @@
...
+    const contracts = await this.service.getValue(
+      ref => ref.where('titleId', '==', titleId)
+        .where('status', '==', 'accepted')
+        .where('type', '==', 'sale')
+        .where('buyerId', 'not-in', excludedOrgs)
```
GrandSchtroumpf, line 16: keep it simple. Use one `excludedOrg` and use `!=` operator

libs/contract/src/lib/contract/item/contract-item.component.html
pldespaigne: I told him to do that, sorry @iPR0D1G1 it's my bad

libs/contract/src/lib/bucket/+state/bucket.model.ts
```diff
@@ -30,20 +30,7 @@ export interface BucketContract {
...
-  languages: Record<string, MovieLanguageSpecification>;
-  runs?: {
-    broadcasts: number;
-    catchup: {
-      from: Date,
-      duration: number,
-      period: 'day' | 'week' | 'month'
-    }
-  }
+export interface BucketTerm extends TermBase<Date> {
```
pldespaigne, line 33: You should not have empty types, if `BucketTerm` = `TermBase`, then it should be `Term extends BucketTerm`.
There is no use to create a third base type here, I though `BucketTerm` had some other properties not included in the `Term`

pldespaigne (changes requested):

libs/contract/src/lib/bucket/+state/bucket.model.ts
pldespaigne: It's even that which is breaking the CI ^^
image

pldespaigne (approved):

Issue #6382, opened 2021-07-26
Improve the request access


RemcoSimonides:
On all pages:
- [x] Clicking on app logo should navigate the user to landing page (a.k.a. log the user out)

Step 1:
![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/198b6ecf-581c-4545-8c9b-8933a2b08e0d)
[Zeplin](https://zpl.io/VkM4q0g)

Step 2:
- [x] Text for ( i ) icon after 'Choose your role' should contain the text "_Depending on your role, you might have access to different features of the platform_".
- [x] Bonus: Also change the text on Create account page (currently it says '... access 'to different _things_ ...'
- [x] Cancel button navigates the user back to landing page (a.k.a. log the user out)
- [x] The text inside the square brackers is changing depending on the app (Buyer and Sales Agent for Market and Content and Investor and Partner for Financiers)
Note: Check [Zeplin](https://zpl.io/b6Anz7q). Text might have changed.
![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/3564af7e-c3f1-4c75-b1ae-b7dbaed2e6a1)

Step 3:
Note: Check [Zeplin](https://zpl.io/aRRn5dn). Text might have changed.
![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/0217f0eb-353d-43ac-8fb3-1ebc3b15d23c)

Link to original design issue #5783

RemcoSimonides:
@phpgeekfr Hey team! Can you please add your [planning poker estimates with ZenHub](https://app.zenhub.com/workspaces/Dev---Sanity-Team-5f6b0067599064680cd2c617/issues/blockframes/blockframes/6382?planning-poker)

PR #6480, opened 2021-08-18
Feat #6382 part 2

phpgeekfr:
Feat #6382 part 2


libs/organization/src/lib/organization/pages/request-access-pending/request-access-pending.component.html
```diff
@@ -0,0 +1,9 @@
+<app-logo (click)="logout()"></app-logo>
+
+<mat-card fxLayout="column" fxLayoutAlign="center center">
```
GrandSchtroumpf, line 3: If you're not using mat-card directive (mat-card-image, mat-card-title, ...) then do not use mat-card. Use the class="surface" instead.

libs/organization/src/lib/organization/pages/request-access-pending/request-access-pending.component.scss
```diff
@@ -0,0 +1,35 @@
...
+}
+
+mat-card {
+  text-align: center;
+  padding-top: 48px;
+  margin: auto;
+  width: 50%;
+  
+  img {
+    height: auto;
```
GrandSchtroumpf, line 26: remove this height, except if you know exactly the ratio of the image, then set the exact value for height & width. That will avoid potential ump as th browser will reserve this space for the img before fetching it.

libs/organization/src/lib/organization/pages/request-access/request-access.component.html
```diff
@@ -1,15 +1,39 @@
...
-  <p class="mat-body-1">
-    We noticed that your email address is associated with an existing user account of one of our applications.<br/>
-    To access Archipel Content, please request access and wait for our team's approval.
-  </p>
-  <p class="mat-body-2">N.B.: All members from your company will also get granted access to {{ appName }}.</p>
-  <button mat-flat-button color="primary" (click)="requestAccess()" [disabled]="disabledRequest">Request Access</button>
-  <button mat-button color="primary" (click)="refresh()">Refresh Page</button>
+  <ng-container *ngIf="step$ | async as step">
+
+    <ng-container *ngIf="step === 'initial'">
```
GrandSchtroumpf, line 6: Maybe using `ngSwitchCase` make more sense here. But it's not super important

libs/organization/src/lib/organization/pages/request-access/request-access.component.html
```diff
@@ -1,15 +1,39 @@
...
+    </ng-container>
+
+    <ng-container *ngIf="step === 'request'">
+      <h1>What access are your looking for ?</h1>
+      <p class="mat-body-1">
+        To access {{ appName[currentApp] }}, please select a "role". It will determine the features you can acccess once you've joined the platform.
+      </p>
+      <p class="mat-body-2">N.B.: All members from your company will be granted the same access.</p>
+
+      <article fxLayout="row" fxLayoutAlign="space-between center" fxLayout.lt-md="column">
```
GrandSchtroumpf, line 23: you don't nee to specify `fxLayout="row"` as `row` is the default. You can do `fxLayout`

libs/organization/src/lib/organization/pages/request-access/request-access.component.html
```diff
@@ -1,15 +1,39 @@
...
+        <div fxLayout="row">
+          <p>Choose your role * :</p>
+          <mat-icon svgIcon="info"
+            matTooltip="Depending on your role, you might have access to different features of the platform."></mat-icon>
+        </div>
+        <mat-button-toggle-group [ngClass]="{ disabled: false }" name="appAccess" [formControl]="formControl">
```
GrandSchtroumpf, line 29: Hmmm I don't understand why you force the element to NOT have the class `disabled`?

libs/organization/src/lib/organization/pages/request-access/request-access.component.html
```diff
@@ -1,15 +1,39 @@
...
+
+    <ng-container *ngIf="step === 'request'">
+      <h1>What access are your looking for ?</h1>
+      <p class="mat-body-1">
+        To access {{ appName[currentApp] }}, please select a "role". It will determine the features you can acccess once you've joined the platform.
+      </p>
+      <p class="mat-body-2">N.B.: All members from your company will be granted the same access.</p>
+
+      <article fxLayout="row" fxLayoutAlign="space-between center" fxLayout.lt-md="column">
+        <div fxLayout="row">
```
GrandSchtroumpf, line 24: I think a `label` would make more sense here : 
```html
<label>
  <span>Choose your role * :</span>
  <mat-icon ...></mat-icon>
</label>
```
You would also benefit from the global spacing between span & mat-icon like that.

libs/organization/src/lib/organization/pages/request-access/request-access.component.scss
```diff
@@ -36,4 +37,19 @@ mat-card {
...
+  margin-bottom: 8px;
+}
+
+mat-button-toggle {
+  min-width: 125px;
+}
+
+mat-icon {
+  margin-left: 4px;
+  margin-right: 12px;
```
GrandSchtroumpf, line 50: I think this part of the style should be on the label (see my comment above) : 
```css
label {
  padding: 0 12px;
}
```

GrandSchtroumpf (approved):

libs/organization/src/lib/organization/pages/request-access/request-access.component.html
phpgeekfr: bad copy/paste ..

PR #6492, opened 2021-08-20
add partner logos on landing page

Poppe-A:


apps/festival/festival/src/app/landing/landing.component.html
```diff
@@ -11,6 +11,16 @@ <h3>One tool for your year-round <br/>promotion & acquisitions</h3>
   </landing-header>
 
   <landing-content>
+
+    <section class="landing-partners" fxLayout="column" fxLayoutAlign="start" fxLayoutGap="16px">
+      <h1 class="mat-display-1">They're using it</h1>
+      <p class="mat-body-1">Over 1300 buyers are already using Archipel, from companies like:</p>
+      <div fxLayout="row" fxLayoutAlign="space-between">
```
RemcoSimonides, line 18: fxLayout="row" is the default of fxLayout so you can remove `="row"` and keep only `fxLayout`

RemcoSimonides (changes requested):

RemcoSimonides:
I think something like this should be ok
image

```
      <div fxLayout="row wrap" fxLayoutGap="24px">
        <img *ngFor="let partnerLogo of partnerLogoList" [asset]='partnerLogo' [alt]='partnerLogo'>
      </div>
```

RemcoSimonides (approved):

PR #6502, opened 2021-08-24
switched to node for pre-commit script

phpgeekfr:


tools/scripts/pre-commit.ts
```diff
@@ -0,0 +1,23 @@
...
+const dotenv = require('dotenv');
+const forbiddenEntries = dotenv.parse(fs.readFileSync('.env'));
+const childProcess = require('child_process');
+const forbiddenKeys = Object.keys(forbiddenEntries);
+const exec = require('child_process').exec;
+const cmd = 'git commit -v --dry-run';
+const output = childProcess.execSync(cmd).toString();
+
+let errors = false;
+for (let key of forbiddenKeys) {
```
GrandSchtroumpf, line 11: you can use `const key of forbiddenKeys` as you're not updating the same variable as in a regular `for (let i = 0; ...)`

tools/scripts/pre-commit.ts
```diff
@@ -0,0 +1,23 @@
+const fs = require('fs');
```
GrandSchtroumpf, line 1: Please use `import` instead of `require` :
```typescript
import fs from 'fs';
import { exec } from 'child_process';
import dotenv from 'dotenv';
```
or something like that.

tools/scripts/pre-commit.ts
```diff
@@ -0,0 +1,23 @@
...
+const dotenv = require('dotenv');
+const forbiddenEntries = dotenv.parse(fs.readFileSync('.env'));
+const childProcess = require('child_process');
+const forbiddenKeys = Object.keys(forbiddenEntries);
+const exec = require('child_process').exec;
```
GrandSchtroumpf, line 6: looks like you're not using it.

tools/scripts/pre-commit.ts
```diff
@@ -0,0 +1,23 @@
...
+const cmd = 'git commit -v --dry-run';
+const output = childProcess.execSync(cmd).toString();
+
+let errors = false;
+for (let key of forbiddenKeys) {
+  if (!!forbiddenEntries[key] && output.includes(forbiddenEntries[key])) {
```
GrandSchtroumpf, line 12: just for readability : 
```typescript
const value = forbiddenEntries[key];
if (!!value && output.includes(value))
```

GrandSchtroumpf (approved):

tools/scripts/pre-commit.ts
phpgeekfr: not possible as we are in `js`

Issue #6460, opened 2021-08-17
db-cleaning script issues


phpgeekfr:
Currently, the db-cleaning script  is executed during upgrade process with maintenanceMode set to `true`.
In this context, document deletion is not triggering any 'onDocumentDelete' methods in backend-functions.

This made me realize that some deletions performed  during execution of this script are not deleting related document in cascade (IE: if I remove an user, I also want to remove logo on storage, delete notifications linked to this users etc..).

So to have a good cleaning-script, related document should be deleted also but this require a huge refactoring to allow interaction with bucket, algolia, jwplayer etc.. outside of backend-functions.

We could also call this db-cleaning script when app is not in maintenanceMode but this can be very risky as we don't know  all the sides effects and this cannot be fully tested with unit-testing.

My suggestion is to let the script as it is for now (what we does not handle right now  is media deletion when we delete users but there is no other side effect).

Regarding orgs without users, there is only 4 in production, we could delete them manually.

Regarding other documents that may be deleted during db-cleaning script, we need to check how we can manage this cleanly.

_Originally posted by @phpgeekfr in https://github.com/blockframes/blockframes/issues/5371#issuecomment-823973325_

- [x] look for `@TODO #6420` in code

Jackseed:
It would be useful to have a list of objects possibly deleted by the cleaning script, plus the related ones that should be deleted in consequence but won't be as you mentioned. 
Then we may see how critical it is not to delete it and/or find solution object per object.

phpgeekfr:
Yes good idea, will do that

phpgeekfr:
Along with this issue, I will also check how to create a shrinked db for e2e tests on every PR

phpgeekfr:
since anonymous login was activaded `removeUnexpectedUsers` will remove them since they don't have corresponding uid on db. Pay attention to this

Jackseed:
shouldn't we update this script then?

phpgeekfr:
> shouldn't we update this script then?

Not for this release but for the next one yes

phpgeekfr:
- [ ] remove remaining watermarks fields on invitations (fromUser, toUser)  CF #6905

phpgeekfr:
Minimal update have been done in https://github.com/blockframes/blockframes/pull/7222

More work is in https://github.com/blockframes/blockframes/pull/7170 but this is not mandatory anymore

PR #6513, opened 2021-08-25
Feat 6460 - create a shrinked DB for e2e in emulator

phpgeekfr:
Feat 6460 - create a shrinked DB for e2e in emulator

First part of #6460


.circleci/config.yml
```diff
@@ -316,6 +316,11 @@ jobs:
     steps:
       - prepare_environment
       - run: npm run firestore:run-db-anon-process
+  push-shrinked-db-to-ci:
```
george43g, line 319: can we not make any changes to the CI config? Once this is merged (I will either rebase into my branch or merge into develop and then merge in develop) I will integrate the shrinked db. It should just become part of PFT  and anon-db that is run rather than a separate process IMHO. 

The tests should check for the existence of required users and provide and easy error message if they are missing. We are ignoring missing users in fixtures at the moment in PFT and I'm not sure why.

apps/catalog/catalog/src/environment/environment.ts
```diff
@@ -1,3 +1,4 @@
 import { getEmulatorsConfig } from "@blockframes/utils/emulator-front-setup";
+import { emulators } from "@env";
```
george43g, line 2: This defeats the purpose of the separate emulator environment file we have here. We are not relying in values in @env - each env may or may not need to use an emulator in different situations.

apps/cms/src/environment/environment.ts
```diff
@@ -1,3 +1,4 @@
 import { getEmulatorsConfig } from "@blockframes/utils/emulator-front-setup";
+import { emulators } from "@env";
 
-export const emulatorConfig = getEmulatorsConfig({ auth: false, firestore: false, functions: false });
+export const emulatorConfig = getEmulatorsConfig(emulators);
```
george43g, line 4: Same for all of these

apps/backend-ops/src/main.ts
```diff
@@ -49,6 +50,9 @@ async function runCommand() {
     case 'anonProdDb':
       await anonymizeLatestProdDb();
       break;
+    case 'shrinkDb':
+      await shrinkDb();
```
george43g, line 54: This is cool. Just give me the command and I can take over from there :)

apps/backend-ops/src/main.ts
```diff
@@ -37,7 +38,7 @@ async function runCommand() {
       await loadEmulator({ importFrom: arg1 });
       break;
     case 'importEmulator':
-      await importEmulatorFromBucket(arg1);
+      await importEmulatorFromBucket({ importFrom: arg1 });
```
george43g, line 41: I don't like to put boolean arguments in these strings - because of the way strings are passed, there are certain flags that are captured by npm (`npm start`) before they reach the TS code. some flags are also captured by `ts-node` as well, and finally we get access to arguments. So I want to minimise what we include in command lines - this is why that `inquirer` PR got stuck in my stashes. Unless code is compiled and linked on a system (such as tools like firebase and npm itself) we cannot reliably get command line options - so far strings of text without dashes separated by spaces has worked, but this could be a source of hard to catch bugs in future.

I will transition to a compiled version and tool eventually, but for now I would prefer separate commands for
"import data into emulator backup but don't start it"
"import data into emulator and start it, and start the following emulators"

So maybe let me handle this if you explain your use case

sample.env
```diff
@@ -9,6 +9,7 @@ ALGOLIA_API_KEY=""
 ALGOLIA_SEARCH_KEY=""
 JWPLAYER_KEY=""
 JWPLAYER_SECRET=""
+JWPLAYER_APIV2SECRET=""
```
george43g, line 12: good catch!

package.json
```diff
@@ -53,6 +53,7 @@
     "postinstall": "ngcc --properties es2015 browser module main && node ./decorate-angular-cli.js",
     "prerender:festival": "nx build festival --prod && nx run festival:server --configuration production && angular-prerender --browser-target festival:build --server-target festival:server",
     "firestore:run-db-anon-process": "npm run backend-ops anonProdDb",
+    "firestore:run-db-shrink-process": "npm run backend-ops shrinkDb",
```
george43g, line 56: again, not sure we need this - I'm thinking of moving everything under `npm start` since we basically have our own tool to launch apps, emulators and more. This will keep the number of npm commands smaller.

Although we've used this method previously (npm run etc...), I think we're moving away from it now except where needed.

libs/firebase-utils/src/lib/firestore/emulator.ts
```diff
@@ -1,7 +1,6 @@
 import { firebase } from '@env'
 import { initializeApp } from 'firebase-admin'
-import { clearFirestoreData, withFunctionTriggersDisabled } from '@firebase/rules-unit-testing';
```
george43g, line 3: I left this here as a reminder that we can build backend functions in such a way that all triggers are disabled without relying on maintenance. Or the other way is to simply not enable functions emulator at all. Not sure if we'll ever need this.

libs/firebase-utils/src/lib/copy-db-from-ci.ts
```diff
@@ -10,6 +10,8 @@ export const latestAnonDbFilename = 'LATEST-ANONYMIZED.jsonl';
 
 export const latestAnonDbDir = 'LATEST-ANON-DB';
 
+export const latestAnonShrinkedDbDir = 'LATEST-ANON-SHRINKED-DB';
```
george43g, line 13: Shouldn't it be the same DB? we can experiment with the shrinked version In my PR

apps/backend-ops/src/db-shrink.ts
```diff
@@ -0,0 +1,466 @@
...
+  campaigns: [
+    'orgId'
+  ]
+}
+
+export async function shrinkDb() {
+  const { db } = loadAdminServices({ emulator: true });
+
+  // STEP 1 load-latest anon-db into emulator and keep it running with auth & firestore
+  const proc = await importEmulatorFromBucket({ importFrom: `gs://${ciBucketName}/${latestAnonDbDir}`, keepEmulatorsAlive: true });
```
george43g, line 117: I see why you did this now. Partially my fault because I didn't create a full suite of options for devs to use.

george43g (changes requested):
Do you want me to make a PR with my suggested changes?

george43g (approved):

PR #6514, opened 2021-08-26
landing clients opinion + form

Poppe-A:


apps/festival/festival/src/app/landing/landing.component.scss
```diff
@@ -72,12 +86,15 @@ landing-content {
...
 
-  .landing-join {
-    background-image: url("/assets/images/background_hero.png");
-    background-size: cover;
-    background-repeat: no-repeat;
-    background-position: center center;  
-    padding: 96px 0;
+  .opinions > div {
+    padding: 24px 56px;
+    background-color: var(--background-card);
```
RemcoSimonides, line 91: Can you check if we should use "surface" class instead?
I don't remember exactly what surface class does, but I feel like it will make it look like a card somewhat

libs/landing/src/lib/shell/shell.component.html
```diff
@@ -72,18 +72,18 @@ <h3>Schedule a demo with our team!</h3>
             </mat-error>
           </mat-form-field>
         </article>
-        <article fxLayout="row" fxLayoutAlign="space-between start" fxLayout.lt-md="column" fxLayoutGap="24px">
-          <mat-form-field appearance="outline">
+        <article fxLayout="row" fxLayoutAlign="space-between start" fxLayout.lt-md="column" fxLayoutAlign.lt-md=""erf fxLayoutGap="24px">
```
RemcoSimonides, line 75: What's this? `fxLayoutAlign.lt-md=""erf` :p

Poppe-A: wooow sorry ... I must have look more closely

apps/festival/festival/src/app/landing/landing.component.scss
```diff
@@ -25,8 +29,18 @@ landing-content {
     padding: 80px
   }
 
-  h1 {
-    margin-bottom: 12px;
+  .landing-join {
+    background-size: cover;
+    background-repeat: no-repeat;
+    background-position: center center;
```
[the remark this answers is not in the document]
Poppe-A: Yes. Without these lines, the bg won't be centered (i've tested it)

apps/festival/festival/src/app/landing/landing.component.scss
Poppe-A: Surface is kind of the same that mat-card. So it's not really what I want here, because I don't need the borders, the layout and the padding are not the same

RemcoSimonides (approved):

PR #6545, opened 2021-08-30
last festival LP part

Poppe-A:


apps/festival/festival/src/app/landing/landing.component.scss
```diff
@@ -81,6 +80,25 @@ landing-content {
...
+    img {
+      width: 100%;
+      object-fit: contain;
+    }
+    > div { 
+      padding: 24px;
+      > p {
+          margin-top: 24px;
+        }
+    }
```
RemcoSimonides, line 98: Is the `>` needed here?

apps/festival/festival/src/app/landing/landing.component.html
```diff
@@ -100,4 +100,24 @@ <h3>{{ benefit.title }}</h3>
...
   </landing-content>
 
+  <landing-detail>
+      <h1 class="mat-display-1">They're talking about us</h1>
+      <section fxLayout fxLayoutAlign="space-between center" fxLayout.lt-md="column" fxLayoutGap="24px">
+        <ng-container *ngFor="let review of reviews">
+          <div class="surface" fxLayout="column" fxFlexAlign="stretch"> 
+            <img [asset]="review.img" alt="review image">
+            <div fxLayout="column" fxLayoutAlign="space-between" fxLayoutGap="24px" fxFlex="1 0">
+              <div>
```
RemcoSimonides, line 110: This div is not needed

apps/festival/festival/src/app/landing/landing.component.html
```diff
@@ -100,4 +100,24 @@ <h3>{{ benefit.title }}</h3>
...
+                <h3>{{review.title}}</h3>
+                <p>{{review.text}}</p>
+              </div>
+              <div fxLayout fxLayoutAlign="center center" fxFlexAlign="end" fxLayoutGap="24px">
```
RemcoSimonides: I think `fxLayoutAlign="end center"` would do. No need for fxFlexAlign then

apps/festival/festival/src/app/landing/landing.component.scss
```diff
@@ -81,6 +80,25 @@ landing-content {
   }
 }
 
+landing-detail {
+  padding: 80px;
+  background-color: var(--background-card);
+  .surface {
+    padding: 0;
```
RemcoSimonides, line 87: padding is defined globally on the surface style so better not overwrite it.
I see you want the image to hit the sides. I was thinking `mat-card` would actually be a valid option here but that one also has padding

@GrandSchtroumpf can he overwrite the padding here to make the card hit the sides or do you have another suggestion?

libs/landing/src/lib/shell/shell.component.html
```diff
@@ -88,22 +88,55 @@ <h3>Schedule a demo with our team!</h3>
...
-      <h3>Proudly associated with the following partners</h3>
-      <span fxLayout="row wrap" fxLayoutAlign="space-evenly start" fxLayoutGap="80px">
-        <img asset="CNC_logo.png" type="logo" alt="cnc">
-        <img src="/assets/logo/BPI_france.png" alt="BPI France logo" />
-        <img src="/assets/logo/creative_media_europe.png" alt="creatiev media europe logo" />
-        <img asset="IFCIC_logo.png" type="logo" alt="ifcic">
-      </span>
+    <footer fxLayout="column" fxLayoutAlign="start center" fxLayoutAlign.lt-md="center center"
+      fxLayoutGap="32px">
+      <div fxLayout fxLayoutAlign="start">
```
RemcoSimonides, line 104: Don't need this `div` I think

libs/landing/src/lib/shell/shell.component.html
```diff
@@ -88,22 +88,55 @@ <h3>Schedule a demo with our team!</h3>
...
+        <app-logo></app-logo>
+      </div>
+      <mat-divider></mat-divider>
+      <div fxLayout fxLayoutAlign="space-between" fxLayout.lt-md="column" fxLayoutGap.lt-md="24px">
```
RemcoSimonides, line 108: Change to `section`

Do you know how grid works? I think it can be applied well here
image
I see 3 colums and 2 rows here

With this change you can remove quite some divs and flex boxes I think

RemcoSimonides (changes requested):

apps/festival/festival/src/app/landing/landing.component.html
Poppe-A: Without this div, the space between the title and the text are not the same in all the "surfaces"

apps/festival/festival/src/app/landing/landing.component.html
Poppe-A: fxFlexAlign = align-self. It will select an element and align it to the end of its parent.
fxLayoutAlign="end center" will affect the childrens of the component

libs/landing/src/lib/shell/shell.component.html
Poppe-A: By default (without this div) app-logo take the full width and is centered. I've added this div to align the logo to the left

apps/festival/festival/src/app/landing/landing.component.scss
Poppe-A: 2 solutions :

libs/landing/src/lib/shell/shell.component.html
```diff
@@ -88,22 +88,55 @@ <h3>Schedule a demo with our team!</h3>
...
+        <div fxLayout="column" fxLayoutAlign="space-between start">
```
RemcoSimonides, line 109: Can you change the 1x div and 2x span to `article`? Also fix indentation of code a bit :)

apps/festival/festival/src/app/landing/landing.component.html
```diff
@@ -116,4 +116,23 @@ <h3>{{opinion.name}}</h3>
       </article>
     </section>
   </landing-content>
+  
+  <landing-detail>
+      <h1 class="mat-display-1">They're talking about us</h1>
+      <section fxLayout fxLayoutAlign="space-between center" fxLayout.lt-md="column wrap" fxLayoutGap="24px">
```
RemcoSimonides, line 122: Can you add `fxLayoutAlign.lt-md="start center"`? Looks better

apps/festival/festival/src/app/landing/landing.component.scss
GrandSchtroumpf: Please never override any padding from a global component / class.
In this case the UI looks very close to a mat-card indeed (mat-card-image, mat-card-title, mat-card-content-mat-card-footer). So I would use mat-card for that. But don't override the default style.

apps/festival/festival/src/app/landing/landing.component.html
```diff
@@ -116,4 +116,23 @@ <h3>{{opinion.name}}</h3>
...
+              <mat-card-title>{{review.title}}</mat-card-title>
+              <mat-card-content>{{review.text}}</mat-card-content>
+              <div fxLayout fxLayoutAlign="center center" fxFlexAlign="end" fxLayoutGap="16px">
+                <span class="mat-caption">From</span>
+                <img [asset]="review.logo">
+              </div>
+            </div>
+          </mat-card>
+        </ng-container>
+      </section>
```
RemcoSimonides, line 136: Fix indentation

RemcoSimonides (approved):

PR #6555, opened 2021-08-31
refactor static-group component.

iPR0D1G1:
An attempt to fix this :  #5985


libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -60,32 +58,50 @@ function getItems(groups: StaticGroup[]): string[] {
...
+
   @ViewChild('inputEl') input: ElementRef<HTMLInputElement>;
   @Input() displayAll = '';
-  @Input() @boolean required = false;
-  @Input() @boolean disabled = false;
-  @Input() placeholder = 'Tap to filter'
   @Input() withoutValues: string[] = [];
   @Input() scope: GroupScope;
-  @Input() icon: string;
+  @Input() set placeholder(plh) {
```
GrandSchtroumpf, line 90: type your params and use the real name : `placeholder: string`

libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -98,10 +114,27 @@ export class StaticGroupComponent implements ControlValueAccessor, OnInit, OnDes
...
   }
 
+  set value(values: string[] | null) {
+    this.form.setValue(values)
+    this.touched = true;
+    this.stateChanges.next();
+  }
+
+  get empty() {
+    return this.form.value?.length < 1;
```
GrandSchtroumpf, line 124: Better : `return !this.form.value?.length`

libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -138,6 +171,10 @@ export class StaticGroupComponent implements ControlValueAccessor, OnInit, OnDes
   onOpen(opened: boolean) {
     if (opened) {
       this.input.nativeElement.focus();
+      /**@Todo: wip removing the below line causes the determination of error state to not work.
+       * Checkout why, you can ask for help. You've already explored all options known to you.
+       */
+      this.touched = true;
```
GrandSchtroumpf, line 177: looks good, there is no reason to remove this line.

libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -149,19 +186,69 @@ export class StaticGroupComponent implements ControlValueAccessor, OnInit, OnDes
...
   }
+
   registerOnTouched(fn: () => void) {
     this.onTouch = fn;
   }
   setDisabledState(isDisabled: boolean): void {
     this.disabled = isDisabled;
   }
 
+  checkIsControlValid(values: string[]): ValidationErrors | null {
```
GrandSchtroumpf, line 202: This method doesn't use `this`. It could be outside of the component. Also why not using `Validators.minlength(1)` directly ?

libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -149,19 +186,69 @@ export class StaticGroupComponent implements ControlValueAccessor, OnInit, OnDes
...
+      return {
+        mustBePositive: {
+          length: values.length
+        }
+      };
+    }
+    return null;
+  }
+
+  validate(control: AbstractControl): ValidationErrors | null {
```
GrandSchtroumpf, line 213: What the goal of this function ?

libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -149,19 +186,69 @@ export class StaticGroupComponent implements ControlValueAccessor, OnInit, OnDes
...
+    const values = control.value;
+    return this.checkIsControlValid(values)
+  }
+
+  get errorState(): boolean {
+    const value = Boolean(this.checkIsControlValid(this.form.value)) && this.touched;
```
GrandSchtroumpf, line 219: just return it `return Boolean(Validators.minlength(0)(this.form.value)) && this.touched`

libs/ui/src/lib/static-autocomplete/group/group.component.ts
iPR0D1G1: It is an abstract method from the `Validator` interface.

libs/ui/src/lib/static-autocomplete/group/group.component.ts
iPR0D1G1: The component itselContrlValueAccessor, I don't see how we can apply a validator from within the definition of itself.
Could you give me some indications to do that?

libs/contract/src/lib/bucket/form.ts
```diff
@@ -23,7 +23,7 @@ function createBucketTermControl(params: Partial<BucketTerm> = {}) {
   const term = createBucketTerm(params);
   return {
     territories: new FormStaticValueArray<'territories'>(term.territories, 'territories'),
-    medias: new FormStaticValueArray<'medias'>(term.medias, 'medias'),
+    medias: new FormStaticValueArray<'medias'>(term.medias, 'medias',),
```
GrandSchtroumpf, line 26: remove the ","

libs/ui/src/lib/static-autocomplete/group/group.component.ts
```diff
@@ -54,38 +52,68 @@ function getItems(groups: StaticGroup[]): string[] {
   return groups.reduce((items, group) => items.concat(group.items), []);
 }
 
+
+function checkIsControlValid(values: string[]): ValidationErrors | null {
```
GrandSchtroumpf, line 56: I think you don't need that anymore.

GrandSchtroumpf (approved):

PR #6565, opened 2021-08-31
Fix cypress not found in backend issue

### LINKED ISSUE #6561: cannot use "('Cypress' in window)" in backend functions
In 
```
export async function userFirstConnexion(user: PublicUser): Promise<EmailRequest> {
  const supportEmail = ('Cypress' in window) ? testEmail : getSupportEmail(user._meta.createdFrom);
  return {
    to: supportEmail,
    subject: 'New user connexion',
    text: userFirstConnexionTemplate(user)
  };
}
```

Related sentry error : https://sentry.io/organizations/cascade-8/issues/2615348959/?environment=staging&project=1540126&query=is%3Aunresolved

Looks like it is preventing email to be sent.

@xmano already fixed this but on a branch that is not yet on release/3.4

xmano:
Fixes #6561

### DIFF
--- a/apps/backend-functions/src/environments/environment.ts
+++ b/apps/backend-functions/src/environments/environment.ts
@@ -5,6 +5,7 @@
  * Use this setup (non-production) when the execution context is outside firebase
  * functions.
  */
+export const e2e_mode = true;
 
 export {
   factoryContract,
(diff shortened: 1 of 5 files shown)


apps/backend-functions/src/environments/environment.ts
```diff
@@ -5,6 +5,7 @@
  * Use this setup (non-production) when the execution context is outside firebase
  * functions.
  */
+export const e2e_mode = true;
```
phpgeekfr, line 8: is it supposed to be true here ?

xmano: Ok making default `false` and added a comment to clarify.

phpgeekfr (approved):

PR #6580, opened 2021-09-02
fixed(validators.ts):proper cross validation of date fields on update

iPR0D1G1:
related: 6553

### DIFF
--- a/libs/utils/src/lib/form/validators/validators.ts
+++ b/libs/utils/src/lib/form/validators/validators.ts
@@ -177,7 +177,7 @@ export function compareDates(fromKey: string, toKey: string, keyOnControl?: stri
       if (from && to) {
         if (to < from) {
           control.setErrors({ ...errors, startOverEnd: true })
-        } else if (to > from) {
+        } else {
           if ('startOverEnd' in errors) {
             delete errors.startOverEnd;
             if (Object.keys(errors).length < 1) { errors = null }
--- a/libs/utils/src/lib/form/validators/validators.ts
+++ b/libs/utils/src/lib/form/validators/validators.ts
@@ -158,21 +158,24 @@ export function isKeyArrayValidator(scope: Scope): ValidatorFn {
  * @param toKey name of key of to date
  * @returns
  */
-export function compareDates(fromKey: string, toKey: string, keyOnControl?: string): ValidatorFn {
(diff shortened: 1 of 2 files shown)


libs/utils/src/lib/form/validators/validators.ts
```diff
@@ -164,9 +164,37 @@ export function compareDates(fromKey: string, toKey: string, keyOnControl?: stri
       const from = keyOnControl === fromKey ? form.value : eventForm.value[fromKey];
       const to = keyOnControl === toKey ? form.value : eventForm.value[toKey]
 
+      let fromErrors = eventForm.controls['from'].errors;
+      let toErrors = eventForm.controls['to'].errors;
```
pldespaigne, line 168:
```TypeScript
let fromErrors = eventForm.controls['from'].errors ?? {};
let toErrors = eventForm.controls['to'].errors ?? {};
```
and remove the others `?? {}` from everywhere else

libs/utils/src/lib/form/validators/validators.ts
```diff
@@ -164,9 +164,37 @@ export function compareDates(fromKey: string, toKey: string, keyOnControl?: stri
...
+        if (!('startOverEnd' in (fromErrors ?? { }))) {
+          eventForm.controls['from'].setErrors({...(fromErrors??{}), startOverEnd: true })
+        }
+      } else if (keyOnControl === toKey && from && to && from < to) {
+        if (fromErrors && 'startOverEnd' in fromErrors) {
+          delete fromErrors.startOverEnd;
+          if (Object.keys(fromErrors).length < 1) { fromErrors = null }
+          eventForm.controls['from'].setErrors(fromErrors)
+        }
+      }
```
pldespaigne, line 192: The logic is same for both case, could you try to factorize that in a helpers ? The idea is to avoid having the same code written twice.

Also please try to avoid checking several time the same conditions in the same if/else (e.g. `keyOnControl === toKey && from && to`)

pldespaigne (changes requested):

pldespaigne (approved):
you forgot some `;` but otherwise SGTM

GrandSchtroumpf (approved):

GrandSchtroumpf (approved):

PR #6588, opened 2021-09-02
Fix #6527 part 1

### LINKED ISSUE #6527: Movie: CRUD, List, View, Wishlist
- [ ] Movie list
- [ ] Movie view + Financial + Campaign view (movie view tabs)
- [ ] Movie filters > test on MF for specific filter
- [ ] Delete: CRM

Changes that should be tested
#4246
#5799
#6121

phpgeekfr:
#6527 part 1

### DIFF
--- a/apps/backend-functions/src/internals/email.ts
+++ b/apps/backend-functions/src/internals/email.ts
@@ -11,7 +11,7 @@ import { App, getSendgridFrom, AppMailSetting, getAppName, appLogo, applicationU
 import { EmailJSON } from '@sendgrid/helpers/classes/email-address';
 import { logger } from 'firebase-functions';
 
-export const substitutions = {
+const substitutions = { // @TODO #6586 looks like unused
   groupUnsubscribe: "<%asm_group_unsubscribe_raw_url%>",
   preferenceUnsubscribe: "<%asm_preferences_raw_url%>",
   notificationPage: "/c/o/account/profile/view/notifications"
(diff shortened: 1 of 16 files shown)


apps/backend-functions/src/internals/email.ts
```diff
@@ -23,7 +23,7 @@ const criticalsEmailsGroupId = unsubscribeGroupIds.criticalsEmails;
  * Like this, we can avoid showing the criticalEmails group, which is linked for example to the reset password email.
  * Users won't be able to unsubscribe from this group and will always received email from the criticalsEmails group.
 */
-const groupsToDisplay = [unsubscribeGroupIds.allExceptCriticals];
+const groupsToDisplay = [unsubscribeFromAllExceptCriticals];
```
GrandSchtroumpf, line 26: Personally I think that the naming is difficult to maintain. We lose the information that it's an Id. To be honest, I thought it was a function. I would do something like `groupId.unsubcribeAll`. If ever we can force unsbscribe we can do `groupId.forceUnsubcribeAll`.

libs/utils/src/lib/emails/ids.ts
```diff
@@ -77,3 +77,10 @@ export const unsubscribeGroupIds = {
   allExceptCriticals: 15120,
   criticalsEmails: 15136
 }
+
+// This is for letting user unsubscribe from every email except the critical ones as reset password.
+export const unsubscribeFromAllExceptCriticals = unsubscribeGroupIds.allExceptCriticals;
+
+// Use this groupId to remove unsubscribe link at mail bottom. Typically for support emails
+// Note, only "text" emails (as opposition to "html") will remove the unsubscribe link
+export const noUnsubscribeLink = 0;
```
GrandSchtroumpf, line 86: this should be part of a `groupId` object: 
```typescript
const groupId = {
  unsubscribeAll = 15120,
  unsubscribeEmails = 15136,
  noUnsubscribeLink = 0
}
```

GrandSchtroumpf (changes requested):

GrandSchtroumpf (approved):

PR #6589, opened 2021-09-02
Better jwp script

pldespaigne:
- [x] check if jw id exists in jwplayer cdn
- [x] check if storage path exist in firebase storage

### DIFF
--- a/apps/backend-ops/src/rescueJWP.ts
+++ b/apps/backend-ops/src/rescueJWP.ts
@@ -7,15 +7,19 @@ import { delay } from '@blockframes/utils/helpers';
 import { storageFileExist } from 'libs/firebase-utils/src/lib/firebase-utils';
 
 
+type StorageVideoKeys = keyof StorageVideo;
+
+const keysToCheck: StorageVideoKeys[] = [
+  'jwPlayerId',
+  'privacy',
+  'collection',
+  'docId',
+  'field',
+  'storagePath'
+];
+
 function isVideoOK(video: StorageVideo) {
-  return [
-    'jwPlayerId',
-    'privacy',
-    'collection',
-    'docId',
-    'field',
-    'storagePath',
-  ].every(key => !!video[key]);
+  return keysToCheck.every(key => !!video[key]);
 }
 
 function checkStorageFile(video: StorageVideo) {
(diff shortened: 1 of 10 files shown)


apps/backend-ops/src/rescueJWP.ts
```diff
@@ -16,6 +18,20 @@ function isVideoOK(video: StorageVideo) {
   ].every(key => !!video[key]);
```
phpgeekfr, line 18: To avoid some mistakes, maybe you can : 
```

  type StorageVideoKeys = keyof StorageVideo;

  const keysToCheck: StorageVideoKeys[] = [
    'jwPlayerId',
    'privacy',
    'collection',
    'docId',
    'field',
    'storagePath'
  ];

  return keysToCheck.every(key => !!video[key]);
```

phpgeekfr: In fact this will not work because of ` [K: string]: string; // extra-data`

apps/backend-ops/src/rescueJWP.ts
```diff
@@ -39,14 +63,17 @@ export async function rescueJWP() {
...
 
   const okVideos: MovieVideo[] = [];
+  const toCheck: MovieVideo[] = [];
   const emptyVideos: MovieVideo[] = [];
   const notInJWP: MovieVideo[] = [];
   const notInStorage: MovieVideo[] = [];
+  const wrongJWPId: MovieVideo[] = [];
+  const wrongStoragePath: MovieVideo[] = [];
 
   const sortVideo = (video: StorageVideo, movieId: string) => {
```
phpgeekfr, line 73: is sortVideo an appropriate naming ?

phpgeekfr (approved):
LGTM

libs/firebase-utils/src/lib/node-request.ts
```diff
@@ -0,0 +1,26 @@
...
+      ...options.headers,
+      'Content-Type': 'application/json',
+      'Content-Length': Buffer.byteLength(postData),
+    }
+  };
+	return new Promise((resolve, reject) => {
+		const req = request(options.method === 'POST' ? postOptions : options, res => {
+      let body = '';
+      res.on('data', chunk => body += chunk);
+      res.on('end', () => resolve(JSON.parse(body) as T));
```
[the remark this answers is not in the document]
pldespaigne: It's to type the response if you now it. e.g. `const myMovie = sendRequest<Movie>('api/movie/123')`

apps/backend-ops/src/rescueJWP.ts
pldespaigne: I think so ^^
Because the functions takes every videos and sort them into different "bucket" depending on their properties.
A bit like sorting marbles by colors.

Issue #6554, opened 2021-08-31
[Release Test] Contract: creation, email, excel import


RemcoSimonides:
Tester: @iPR0D1G1
Debugger: @pldespaigne 

To test
Avails filter
Add to selection button
Cards on selection page
World map view
Calendar view
Connected as an admin, import of contracts.

Let me know what you've tested for this so I can write those test scenarios in this issue next release

iPR0D1G1:
To test
Avails filter
Add to selection button
Cards on selection page
World map view
Calendar view
Related to: #6553
To avoid merge conflicts, below are only those specific to this tests.

# [Page Avails -> Preselected Rights](https://staging.archipelcontent.com/c/o/marketplace/title/NnhsksCPbqMRzux0hPiT/avails/map)

[] Territories dialog
Doesn't show the list of additional territories.
<img width="1440" alt="Bildschirmfoto 2021-09-01 um 10 25 30 AM" src="https://user-images.githubusercontent.com/22674720/131658257-0de3ba01-d066-4308-bcba-646ac24ab990.png">

iPR0D1G1:
# [Imports page -> Contracts]
(https://staging.archipelcontent.com/c/o/dashboard/import)

- [ ] Importing contracts as admin provides wrong information.
Using the attached template.
[e2e contracts Kopie.xlsx](https://github.com/blockframes/blockframes/files/7090761/e2e.contracts.Kopie.xlsx)

As indicated in the image, it says there was an error.
<img width="1438" alt="Bildschirmfoto 2021-09-01 um 12 31 01 PM" src="https://user-images.githubusercontent.com/22674720/131666295-9b5a33ac-a4c9-45ca-a8fc-0b63e1f30a08.png">

Listing contracts actually indicates there are no contracts.
<img width="1438" alt="Bildschirmfoto 2021-09-01 um 12 50 12 PM" src="https://user-images.githubusercontent.com/22674720/131666426-8072fb60-9bb2-427a-b0ab-db36b2a6bfd2.png">

Jackseed:
url https://staging.archipelcontent.com/c/o/dashboard/contracts/co-517861/efuTTLhCrf7GlQwYqIgZ
user: dev+caesar-val@blockframes.io
- [ ] Contract header loading indefinitely due to permission issue
![image](https://user-images.githubusercontent.com/14964443/131814039-dac1f936-87e7-4941-9e11-64bf02a33b5e.png)

Jackseed:
[...]

PR #6599, opened 2021-09-03
Territories dialog & some other fixes

pldespaigne:
some items from #6554

### DIFF
--- a/libs/utils/src/lib/pipes/group-label.pipe.ts
+++ b/libs/utils/src/lib/pipes/group-label.pipe.ts
@@ -1,5 +1,5 @@
 import { Pipe, PipeTransform, NgModule } from '@angular/core';
-import { StaticGroup, GroupScope, staticGroups } from '@blockframes/utils/static-model';
+import { StaticGroup, GroupScope, staticGroups, staticModel } from '@blockframes/utils/static-model';
 
 @Pipe({ name: 'toGroupLabel' })
 export class GroupLabel implements PipeTransform {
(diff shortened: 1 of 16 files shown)


libs/utils/src/lib/pipes/group-label.pipe.ts
```diff
@@ -1,25 +1,28 @@
...
 
     const allItems = groups.reduce((items, group) => items.concat(group.items), []);
+
     if (allItems.length === value.length) return [all];
+
     return groups.map(group => {
       const items = [];
       for (const item of group.items) {
-        if (value.includes(item)) items.push(staticModel[scope][item]);
+        if (value.includes(item)) items.push(item);
```
GrandSchtroumpf, line 17: Hmmm I don't understand, here you're displaying the key, not the label aren't you ?

libs/utils/src/lib/pipes/to-label.pipe.ts
```diff
@@ -8,7 +8,8 @@ export function toLabel(value: string | string[], scope: Scope): string | string
     if (Array.isArray(value)) {
       return value.map(val => staticModel[scope][val]).join(', ')
     } else {
-      return staticModel[scope][value];
+      // if no value is found return the input value to avoid unexpected empty values
+      return staticModel[scope][value] ?? value;
```
GrandSchtroumpf, line 12: You're hiding the dust under the carpet. `staticModel[scope][value]` shouldn't be undefined. You should fix the root of the problem instead of putting a fallback.

RemcoSimonides (approved):

PR #6605, opened 2021-09-07
6554 last fixes

pldespaigne:
fix last items from #6554

### DIFF
--- a/apps/catalog/catalog/src/app/dashboard/contracts/view/view.guard.ts
+++ b/apps/catalog/catalog/src/app/dashboard/contracts/view/view.guard.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { CanActivate, ActivatedRouteSnapshot, Router } from '@angular/router';
 import { AuthQuery } from '@blockframes/auth/+state';
-import { ContractService, Mandate, Sale } from '@blockframes/contract/contract/+state';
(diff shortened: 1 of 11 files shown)


apps/catalog/catalog/src/app/dashboard/contracts/view/view.guard.ts
```diff
@@ -0,0 +1,24 @@
...
+  constructor(
+    private router: Router,
+    private authQuery: AuthQuery,
+    private contractService: ContractService,
+  ) { }
+
+  async canActivate(route: ActivatedRouteSnapshot) {
+    const { contractId } = route.params;
+
+    const contract = await this.contractService.getValue(contractId) as unknown as Sale | Mandate;
```
GrandSchtroumpf, line 17: The service is not able to know that it's a Contract because `contractId` has type `any`. The real solution would be to use `paramMap` instead of `route.params` because the paramMap knows that the value is a string : 
```typescript
const contractId = route.paramMap.get('contractId');
if (!contractId) return this.router.parseUrl('c/o/dashboard/contracts');
const contract = await this.contractService.getValue(contractId);
const isStakeholder = contract.stakeholders.includes(this.authQuery.orgId);
if (!isStakeholder) return this.router.parseUrl('c/o/dashboard/contracts');
return true;
```

GrandSchtroumpf (approved):

Issue #6577, opened 2021-09-02
Sellers' notification & email on offer creation


Jackseed:
When an offer is created by a buyer it should create a notification and an email for each seller concerned by this offer. 

Notification text: 
```
[ Organization Name] made an offer on [ Title name ], [ Title name ],  [ Title name ] (only but all titles owned by this org within the offer)
```

Sendgrid template id: `d-f7c2e13b7d6b427daf08bdfdf904d69`

Notification settings: within Offer Management section, title: "You receive an offer."

PR #6608, opened 2021-09-07
feat(contract:creation):send email to all stake holder users less the…

iPR0D1G1:
Fixed: #6577


libs/utils/src/lib/emails/ids.ts
```diff
@@ -65,6 +65,9 @@ export const templateIds = {
   },
   offer: {
     created: 'd-94a20b20085842f68fb2d64fe325638a'
+  },
+  contract: {
+    created: 'd-f7c2e13b7d6b427daf08bdfdf904d697'
```
phpgeekfr, line 70: On the issue it is `offer creation` . But actually it is a contract that is created ?

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +44,32 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+        to: user.email,
+        templateId: templateIds.contract.created,
+        data: { user, app: { name: app }, title: { names: title.title.international } }
+      }
+      return sendMailFromTemplate(request, app);
+    })
+  })
+
+  return Promise.all(promises.flat(1))
+
```
phpgeekfr, line 70: If the goal is to create a notification for all stakeholders, you should use `triggerNotifications` That will send the notification. The `triggerNotifications` will need to be updated to trigger the mail also.

In order to let the users decide if they want to receive a mail or not for this kind of notification, you must also update `libs/user/src/lib/auth/forms/notifications-form/notifications-form.component.ts`

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +40,35 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+  const contract = contractSnapshot.data() as Contract;
+  if (contract.type !== 'sale') return;
+  const stakeholders = contract.stakeholders.filter(
+    stakeholder => (stakeholder !== contract.buyerId) && stakeholder !== centralOrgId.catalog
+  );
+
+  if (!stakeholders.length) { return }
+
+  const promises = stakeholders.map(async stakeholder => {
```
GrandSchtroumpf, line 53: There is really no big reason to use Promise.all as we don't care to much about awaiting for all of them to be sent. Actually it's better not to use Promise.all as if one fails, they all fails.
It should look like that : 
```typescript
const title = await getDocument<Movie>(`movies/${contract.titleId}`);
const getNotifications = org => org.userIds.map(userId => createNotification({
  toUserId: userId,
  type: 'contractCreatedConfirmation',
  docId: contract.offerId,
  title,
}));

for (const stakeholder of stakeholders) {
  getDocument<Organization>(`orgs/${stakeholder}`)
    .then(getNotifications)
    .then(triggerNotifications);
}
```

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +40,35 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+    const org = await getDocument<Organization>(`orgs/${stakeholder}`)
+    const title = await getDocument<Movie>(`movies/${contract.titleId}`)
```
GrandSchtroumpf, line 55: this is independent of the stakeholder, you should get it before looping.

libs/notification/src/lib/+state/notification.firestore.ts
```diff
@@ -28,6 +29,7 @@ export const notificationTypesBase = [
   'invitationToAttendScreeningCreated',
 
   // Notifications related to offers
+  'contractCreatedConfirmation',
```
GrandSchtroumpf, line 32: rename `contractCreated`. The `Confirmation` doesn't bring more information.

apps/backend-functions/src/notification.ts
```diff
@@ -445,6 +450,19 @@ async function sendRequestToJoinOrgDeclined(recipient: User, notification: Notif
...
+async function sendContractCreatedConfirmation(recipient: User, notification: NotificationDocument) {
+  const app: App = 'catalog';
+  const toUser = getUserEmailData(recipient);
+  const template: EmailTemplateRequest = {
+    to: toUser.email,
+    templateId: templateIds.contract.created,
+    data: { user: toUser, app: { name: app }, title: { names: notification.title?.title?.international } }
+  }
+
+  await sendMailFromTemplate(template, app, unsubscribeId);
```
GrandSchtroumpf, line 463: You don't need `await`. Use `return` instead.

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +40,35 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
 
+export async function onContractCreate(contractSnapshot: FirebaseFirestore.DocumentSnapshot<Contract>) {
+
+  const contract = contractSnapshot.data() as Contract;
+  if (contract.type !== 'sale') return;
+  const stakeholders = contract.stakeholders.filter(
+    stakeholder => (stakeholder !== contract.buyerId) && stakeholder !== centralOrgId.catalog
+  );
+
+  if (!stakeholders.length) { return }
```
GrandSchtroumpf, line 51: Please don't put `{}` if you inline a `if` statement

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +40,35 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
   console.log(`Contract ${contract.id} removed`);
 }
 
+export async function onContractCreate(contractSnapshot: FirebaseFirestore.DocumentSnapshot<Contract>) {
+
+  const contract = contractSnapshot.data() as Contract;
+  if (contract.type !== 'sale') return;
+  const stakeholders = contract.stakeholders.filter(
```
GrandSchtroumpf, line 47: In the app we usually do like that when you cannot inline a callback on one line:
```typescript
const stakeholders = contract.stakeholders.filter(stakeholder => {
  return stakeholder !== contract.buyerId && stakeholder !== centralOrgId.catalog;
});

File: apps/backend-functions/src/notification.ts [hunk 9]
```diff
@@ -426,6 +431,19 @@ async function sendRequestToJoinOrgDeclined(recipient: User, notification: Notif
...
+function sendContractCreatedConfirmation(recipient: User, notification: NotificationDocument) {
+  const app: App = 'catalog';
+  const toUser = getUserEmailData(recipient);
+  const template: EmailTemplateRequest = {
+    to: toUser.email,
+    templateId: templateIds.contract.created,
+    data: { user: toUser, app: { name: app }, title: { names: notification.title?.title?.international } }
+  }
+
+  return sendMailFromTemplate(template, app, unsubscribeId);
```
phpgeekfr, line 444: update your branch with latest develop and change
`unsubscribeId` to `groupIds.unsubscribeAll`

apps/backend-functions/src/notification.ts
```diff
@@ -1,8 +1,8 @@
 import { InvitationDocument, MovieDocument, NotificationDocument, OrganizationDocument, NotificationTypes } from './data/types';
 import { getDocument, getOrgAppKey, createDocumentMeta } from './data/internals';
 import { NotificationSettingsTemplate, User } from '@blockframes/user/types';
-import { sendMailFromTemplate } from './internals/email';
-import { emailErrorCodes, EventEmailData, getEventEmailData, getOrgEmailData, getUserEmailData } from '@blockframes/utils/emails/utils';
+import { sendMailFromTemplate, sendMail, substitutions } from './internals/email';
```
phpgeekfr, line 4: you can remove `sendMail, substitutions`

apps/backend-ops/src/db-cleaning.ts
```diff
@@ -308,6 +308,7 @@ export function cleanDocsIndex(
  * Check each type of notification and return false if a referenced document doesn't exist
  * @param notification the notification to check
  * @param existingIds the ids to compare with notification fields
+ * @TODO: #6460 & #6608: new notification type `contractCreated` created. Remember to take this into account.
```
phpgeekfr, line 311: For this comment I was more thinking about the `title?:Movie;` attribute you added. But specifying the new notificationType `contractCreated` is also useful. Can you mention both ?

libs/notification/src/lib/+state/notification.firestore.ts
```diff
@@ -60,6 +62,7 @@ export interface NotificationBase<D> {
   organization?: PublicOrganization;
   invitation?: PublicInvitation;
   bucket?: Bucket;
+  title?:Movie;
```
phpgeekfr, line 65: Can you lint the file ? missing a space between `title?:` and `Movie;` ;)

libs/notification/src/lib/notification.component.html
```diff
@@ -7,7 +7,8 @@ <h1>Activity Feed</h1>
         <span>Mark all as read</span>
       </button>
     </header>
-    <notification-list [notifications]="notifications" class="surface"></notification-list>
+    <notification-list [notifications]="notifications" class="surface">
+    </notification-list>
```
phpgeekfr, line 11: No need but ok ;)

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +40,34 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+export async function onContractCreate(contractSnapshot: FirebaseFirestore.DocumentSnapshot<Contract>) {
+  const contract = contractSnapshot.data() as Contract;
+
+  if (contract.type !== 'sale') return;
+
+  const stakeholders = contract.stakeholders.filter(stakeholder => {
+    return (stakeholder !== contract.buyerId) && stakeholder !== centralOrgId.catalog;
+  });
+
+  if (!stakeholders.length) return
```
phpgeekfr, line 52: missing `;` 
If there is no situations where `stakeholders.length` should be equal to zero. Maybe an exception should be thrown ?

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +40,34 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+
+  const title = await getDocument<Movie>(`movies/${contract.titleId}`)
```
phpgeekfr, line 54: missing `;`

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +40,34 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+    toUserId: userId,
+    type: 'contractCreated',
+    docId: contract.offerId,
+    title,
+  }));
+
+  for (const stakeholder of stakeholders) {
+    getDocument<Organization>(`orgs/${stakeholder}`)
+      .then(getNotifications)
+      .then(triggerNotifications);
```
phpgeekfr, line 66: try to avoid the use of `then` can you turn in into `await` ?

libs/notification/src/lib/+state/notification.firestore.ts
phpgeekfr: In fact you seems to be using the new title attribute mainly to get `notification.title.id` in front and `notification.title?.title?.international`. So let's remove this attribute from `NotificationBase<D>` and use docId instead.

so in front: 
`notification.docId` instead of `notification.title.id`

in backend:
```
const movie = await getDocument<MovieDocument>(`movies/${notification.docId}`);
...
movie.title?.international
```

And when creating the notification:
```
createNotification({
    toUserId: userId,
    type: 'contractCreated',
    docId: title.id
  }));
```

Unless you **really** need to keep the offerId. But it looks like not right now.

apps/backend-ops/src/db-cleaning.ts
phpgeekfr: Don't take this comment into account if you remove title attribute from `notificationBase` as suggested in comments of `libs/notification/src/lib/+state/notification.firestore.ts`

libs/user/src/lib/auth/forms/notifications-form/notifications.form.ts
```diff
@@ -31,7 +31,8 @@ function createNotificationsControls(settings: Partial<NotificationSettings> = {
     requestToAttendEventCreated: new  NotificationSettingsForm(settings.requestToAttendEventCreated, true),
     invitationToAttendMeetingCreated: new  NotificationSettingsForm(settings.invitationToAttendMeetingCreated, true),
     invitationToAttendScreeningCreated: new  NotificationSettingsForm(settings.invitationToAttendScreeningCreated, true),
-    offerCreatedConfirmation: new NotificationSettingsForm(settings.offerCreatedConfirmation)
+    offerCreatedConfirmation: new NotificationSettingsForm(settings.offerCreatedConfirmation),
```
phpgeekfr, line 34: not your code but if you could lint the file. It would be great

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +40,34 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+
+  if (!stakeholders.length) return
+
+  const title = await getDocument<Movie>(`movies/${contract.titleId}`)
+
+  const getNotifications = (org: Organization) => org.userIds.map(userId => createNotification({
+    toUserId: userId,
+    type: 'contractCreated',
+    docId: contract.offerId,
+    title,
```
phpgeekfr, line 60: If you can add  `_meta: createDocumentMeta({ createdFrom: appAccess })` it would be great. If no, not an issue.

phpgeekfr (commented):
Good job !

apps/backend-functions/src/contracts.ts
phpgeekfr: maybe you can force to `catalog` as you did in `sendContractCreatedConfirmation`

apps/backend-functions/src/notification.ts
```diff
@@ -426,6 +431,19 @@ async function sendRequestToJoinOrgDeclined(recipient: User, notification: Notif
...
 }
 
+/** Send copy of offer that buyer has created to all non-buyer stakeholders */
+function sendContractCreatedConfirmation(recipient: User, notification: NotificationDocument) {
+  const app: App = 'catalog';
+  const toUser = getUserEmailData(recipient);
+  const template: EmailTemplateRequest = {
+    to: toUser.email,
+    templateId: templateIds.contract.created,
+    data: { user: toUser, app: { name: app }, title: { names: notification.title?.title?.international } }
```
phpgeekfr, line 441: do you need the `app` in the mail template ?

apps/backend-functions/src/contracts.ts
GrandSchtroumpf: I asked him to do that because we don't want to wait between each notification. And we don't want to use `Promise.all` because if one crash, all of them crash. In this case `.then` is the best option in my opinion

phpgeekfr: roger that

apps/backend-functions/src/notification.ts
iPR0D1G1: yes, it's used within the sengrid template.

apps/backend-functions/src/contracts.ts
```diff
@@ -35,6 +39,31 @@ export async function onContractDelete(contractSnapshot: FirebaseFirestore.Docum
...
+    toUserId: userId,
+    type: 'contractCreated',
+    docId: contract.titleId,
+    _meta: createDocumentMeta({ createdFrom: 'catalog' })
+  }));
+
+  for (const stakeholder of stakeholders) {
+    const org = await getDocument<Organization>(`orgs/${stakeholder}`);
+    const notification = await getNotifications(org);
+    triggerNotifications(notification);
```
phpgeekfr, line 63: Sorry I gave you bad info on this. Francois wants to keep the .then for this case

apps/backend-functions/src/notification.ts
```diff
@@ -426,6 +433,15 @@ async function sendRequestToJoinOrgDeclined(recipient: User, notification: Notif
   await sendMailFromTemplate(template, app, groupIds.unsubscribeAll);
 }
 
+/** Send copy of offer that buyer has created to all non-buyer stakeholders */
+async function sendContractCreatedConfirmation(recipient: User, notification: NotificationDocument) {
+  const app: App = 'catalog';
+  const toUser = getUserEmailData(recipient);
+  const movie = await getDocument<Movie>(`movies/${notification.docId}`)
+  const template = contractCreatedEmail(toUser, movie.title?.international, 'catalog')
```
phpgeekfr, line 441: missing `;`

apps/backend-functions/src/templates/mail.ts
```diff
@@ -300,6 +300,12 @@ export function movieAcceptedEmail(toUser: UserEmailData, movieTitle: string, mo
   return { to: toUser.email, templateId: templateIds.movie.accepted, data };
 }
 
+/** Inform user of org whose movie is being bought */
+export function contractCreatedEmail(toUser: UserEmailData, movieTitle: string, app: App): EmailTemplateRequest {
+  const data = { user: toUser, app: { name: app }, title: { names: movieTitle } };
+  return { to: toUser.email, templateId: templateIds.contract.created, data }
```
phpgeekfr, line 306: missing `;`

apps/backend-ops/src/db-cleaning.ts
```diff
@@ -308,6 +308,8 @@ export function cleanDocsIndex(
...
+ * and title?:Movie added in NotiticationBase
```
phpgeekfr, line 312: you can remove `title?:Movie` now that you removed it from notification Interface

phpgeekfr (approved):
Approved with few last comments

apps/backend-functions/src/notification.ts
```diff
@@ -191,6 +193,11 @@ export async function onNotificationCreate(snap: FirebaseFirestore.DocumentSnaps
           .then(() => notification.email.isSent = true)
           .catch(e => notification.email.error = e.message);
         break;
+      case 'contractCreated':
+        await sendContractCreatedConfirmation(recipient, notification)
```
GrandSchtroumpf, line 197: Can you rename `sendContractCreated`

GrandSchtroumpf (approved):

PR #6610, opened 2021-09-07
org in header

pldespaigne:
@Jackseed asked me to replace existing header of catalog dashboard view by this

image

### DIFF
--- a/apps/catalog/catalog/src/app/dashboard/contracts/view/view.component.html
+++ b/apps/catalog/catalog/src/app/dashboard/contracts/view/view.component.html
@@ -4,10 +4,12 @@
       <mat-icon svgIcon="arrow_back"></mat-icon>
     </a>
     <bf-tag>{{contract.status | toLabel: 'contractStatus'}}</bf-tag>
-    <span *ngIf="contract.buyerId | getOrg | async as org" fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="16px">
-      <img height="100" width="100" [ref]="org?.logo" asset="empty_organization.svg" type="logo" alt="Organization Logo">
-      <h1>{{ org | orgName }}</h1>
-    </span>
+    <ng-container *ngIf="contract.buyerId | getOrg | async as org">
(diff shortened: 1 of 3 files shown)


apps/catalog/catalog/src/app/dashboard/contracts/view/view.component.html
```diff
@@ -4,8 +4,11 @@
       <mat-icon svgIcon="arrow_back"></mat-icon>
     </a>
     <bf-tag>{{contract.status | toLabel: 'contractStatus'}}</bf-tag>
-    <h1>Offer Reference: {{contract.offerId}}</h1>
-    <span>Offer created: {{contract._meta.createdAt | date: 'dd/MM/yyyy'}}</span>
+    <span *ngIf="contract.buyerId | getOrg | async as org" fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="16px">
```
RemcoSimonides, line 7: Line is easier to read if this `*ngIf` is in an `ng-container`
`row` is default of fxLayout so no need to specify it

apps/catalog/catalog/src/app/dashboard/contracts/view/view.component.html
```diff
@@ -4,8 +4,11 @@
...
+      <img height="100" width="100" [ref]="org?.logo" asset="empty_organization.svg" type="logo" alt="Organization Logo">
+      <h1>{{ org | orgName }}</h1>
```
RemcoSimonides, line 9: A page should only contain 1x `h1`. Is this the only h1?
If not, change to something like `mat-display-2`. Maybe you don't need to overwrite the margin then either.

RemcoSimonides (approved):

apps/catalog/catalog/src/app/dashboard/contracts/view/view.component.html
pldespaigne: It's indeed the only `h1`, it was already there before, I just moved it and changed its content

Issue #6567, opened 2021-09-01
[Archipel Content] Cleanup


GrandSchtroumpf:
# Title Page
file: `apps\catalog\catalog\src\app\marketplace\title\list\list.component.ts`

- [x] Add cache on title search for contract 
Method `getContract` load the contracts each time we arrive on the page. Using `valueChanges` instead of `getValue` in the services would keep the result in a global cache. Like that we won't have to wait each time we arrive on the page.

- [x] Don't wait for contract before loading
The search is initialized line 124 after waiting for all the contracts. We should not wait for it and start loading all movies as soon as possible.

- [x] Factorize `addAvail` method
This method was created before the avails search mechanism. I think there is room to improve the code and factorize it with existing piece of code.

# Excel Import
file: `libs\import\src\lib\components\view-extracted-elements\contract\view-extracted-contracts.component.ts`

- [x] The main issue is that we've got dependencies between each line, which makes the process more difficult to maintain & avoid error. Discuss with @Jackseed what we can change in the excel design to make it easier to parse without decreasing UX.

related issue (might be close after the cleanup) : 
https://github.com/blockframes/blockframes/issues/5267

- [x] Improve the overall excel import devX. We should look at the existing code of all the import component and check if we can find a pattern there. If we can, create the utils functions to improve the code.

Jackseed:
It’s hard to me to say what you could change in order to clean code but I can give you an overview of what is or could be useful regarding users’ perspective:
- Not completely sure but it feels like we have to fill cells with lower case words (eg “yes”, “mandate”); would be easier to have capital letters (“Yes”, “Mandate”) and in fact even better if you lower case it code-side so users may fill it with whatever caps they want
- International Title, Licensor & Licensee name currently accept names, best would be to have possibility to fill it with ids as well (for admin purposes)
- Territories: it’s really hard to fill territories like “World excluding France”, currently we have to copy paste a list of all the countries and remove France from it. It’s extremely hard to read afterwards and prompt to errors. A while ago we had 2 columns: “Territories included” (eg “World”) and “Territories excluded” (eg “France”); if we could get back to something similar it would be really precious.
It would be also really useful to be able to fill in region names like “Africa” or “Europe” and that every territories from that region are included.
- the “All” keyword is really useful
- I’m not completely sure of the use of the admin columns (purple ones N-P). I guess it’s to attach multiple terms or movies to one contract and that’s useful but I didn’t try it so far and I’m not sure it’s working. Column P should be merged with column A

PR #6663, opened 2021-09-13
[CATALOG] Refactoring Marketplace Title List `addAvail()`

pldespaigne:
solve item of #6567

### DIFF
--- a/libs/contract/src/lib/avails/avails.ts
+++ b/libs/contract/src/lib/avails/avails.ts
@@ -212,22 +212,24 @@ export function getCollidingHoldbacks(holdbacks: Holdback[], terms: BucketTerm[]
 // ----------------------------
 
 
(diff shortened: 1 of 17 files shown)


libs/contract/src/lib/avails/avails.ts
```diff
@@ -212,23 +212,25 @@ export function getCollidingHoldbacks(holdbacks: Holdback[], terms: BucketTerm[]
 // ----------------------------
 
 
+export function getParentTerms(titleId: string, avails: AvailsFilter, mandates: Mandate[], terms: Term[]) {
```
GrandSchtroumpf, line 215: For me you should get the right parentTerm here as you've got the avails.

GrandSchtroumpf (approved):

Issue #6239, opened 2021-07-07
Responsive Large Screens - Landing page


RemcoSimonides:
We start with implementing this responsive capability in the landing page. More pages will follow later.

[Zeplin](https://app.zeplin.io/project/5f3100d317a0041be2ad99c4/screen/60dc4213015aef929f65789f)

Link to discussion in design issue: #5560

GrandSchtroumpf:
The best way to achieve that is to have a global `--max-width` css variable that we can call from anywhere, and create a global class `max-width` (or some better name).
There are two options: 
- use padding
- use margin

# Padding
```scss
.max-width {
  padding-left: minmax(24px, calc((100vh - var(--max-width)) / 2));
  padding-right: minmax(24px, calc((100vh - var(--max-width)) / 2));
  box-sizing: border-box;
}
```
_I put 24px a minimum padding as a reference. Please check with @stephanie-cascade8 for the real value._

This has the advantage of not needing much update on the html : 
```html
<header class="max-width">

</header>
<section class="max-width">

</section>
```

# Margin
```scss
.max-width {
  max-width: var(--max-width);
  margin: auto;
}
```
The css is easier to understand but it means that now we need to wrap the content with a section and apply the background on the section: 
```html
<header>
  <article class="max-width">
  ...
  </article>
</header>
<section>
  <article class="max-width">
  ...
  </article>
</section>
```

RemcoSimonides:
@phpgeekfr Hey team! Can you please add your [planning poker estimates with ZenHub](https://app.zenhub.com/workspaces/Dev---Sanity-Team-5f6b0067599064680cd2c617/issues/blockframes/blockframes/6239?planning-poker)

phpgeekfr:
On large screens (4k) cookie banner does not take the full page:
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/2a1040a4-f2fa-43da-a5df-b393a7ca4038)

RemcoSimonides:
![image](https://user-images.githubusercontent.com/27687382/132825826-55903c5e-f82e-4fe1-a80f-441c1fa5744d.png)
![image](https://user-images.githubusercontent.com/27687382/132825915-86d23e57-f54f-4178-91c6-7a9d558738a2.png)
![image](https://user-images.githubusercontent.com/27687382/132825970-0eadfd9d-3253-4c40-a9d2-c61b007f5bf6.png)

[...]

PR #6668, opened 2021-09-14
6239 responsive landing pages for large screens

RemcoSimonides:
Fix issue #6239

Screenshot 2021-09-15 at 14 13 50

Screenshot 2021-09-15 at 14 14 13


apps/festival/festival/src/app/landing/landing.component.scss
```diff
@@ -18,22 +23,26 @@ landing-content {
...
-  .landing-join {
-    background-size: cover;
-    background-repeat: no-repeat;
-    background-position: center center;  
-    padding: 96px 0;
+  .landing-join-wrapper {
+    background-color: $banner-background;
+
+    .landing-join {
+      background: linear-gradient(91deg, $banner-background 0%, #2a1d34 50%, $banner-background 100%);
```
GrandSchtroumpf, line 43: please put 90deg, or better `to right / to bottom / to left / to top / ...`

apps/catalog/catalog/src/app/landing/landing.component.scss
```diff
@@ -1,16 +1,23 @@
 @import "~@angular/flex-layout/mq";
+$banner-background: rgba(5, 7, 36);
```
GrandSchtroumpf, line 2: couldn't you use an existing variable ? I know that this page is only in light mode, but if we decide to change the variables of the app, it's better to react to them.

RemcoSimonides: Stephanie: we don't have an existing variable. We'd have to add it later
Remco: I noticed that for some sections the theme is overwritten so it will need extra refactoring once we have a dark/light mode dependant variable (see screenshot where it goes wrong)
image

GrandSchtroumpf (approved):

PR #6670, opened 2021-09-14
ok #6635

### LINKED ISSUE #6635: Homepage film slider - display logline instead of synopsis
We changed synopsis to logline on this component a bit ago but it was a mistake since logline isn't mandatory most of the movies are empty. 

Instead we shall use synopsis again but limit the characters display to 350 and then display "..." at the end if synopsis text has more characters than that.

phpgeekfr:
fix #6635


libs/utils/src/lib/pipes/max-length.pipe.ts
```diff
@@ -2,23 +2,39 @@ import { CommonModule } from '@angular/common';
...
  */
 @Pipe({ name: 'maxLength' })
-export class MaxLength implements PipeTransform {
-  transform(text: string, length: number) {
-    if(text?.length && text.length > length) {
-      return text.substr(0, length).concat('...');
-    }
-    return text;
+export class MaxLengthPipe implements PipeTransform {
+  transform(text: string, length: number, keepLastWordComplete = false, addEllipsis = true) {
```
RemcoSimonides, line 30: addEllipsis is never false so far but you just wanted to have that in case it's needed? Not sure if it's ever needed... Also, the comment above the pipe says "and replace it with 3 dots". Remove that if you keep addEllipsis parameter.

GrandSchtroumpf (approved):

Issue #5895, opened 2021-05-31
Rework event form structure


Jackseed:
This form is getting more and more event types and before adding new ones or changing the UI (tabs), we need to work on the structure. 
Main objectives are:
- split common sections from specific ones (like we did with the movie form)
- rework on the association short form (calendar page) + long form because it's blocking the mandatory fields

2amarkina:
- [ ] introduce the "mandatory" flow 
- [ ] mix the image upload solutions (empty + with elements)

2amarkina:
[Access wireframes ](https://www.figma.com/file/HQ2FBvm2u0vyWO0QpIKeTs/%5B-Event-Form-Structure-%5D?node-id=0%3A1)

2amarkina:
- [ ] no mandatory 

- [ ] same mini-form 

- [ ] saving process

- [ ] tabs

2amarkina:
- [ ] Take into account the use case when the buyers try to click on title names mentioned in the description

RemcoSimonides:
@RemcoSimonides Create dev issue

2amarkina:
The new structure is ready for Screenings and Meetings. 

### The Screening Wireframes: 
1. ![Снимок экрана 2021-07-06 в 16.40.33.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/cb9e806e-0e21-4fa7-828f-15168da301dc)

2. ![Снимок экрана 2021-07-06 в 16.40.54.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/b79effa1-1bb1-48eb-9c07-b57befe82068)
![Снимок экрана 2021-07-06 в 16.41.09.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/e0f827c0-95ce-4820-9dcd-707888542490)

3. If possible, keep Attendance tab opened by default, when screening if finished. 
![Снимок экрана 2021-07-06 в 16.41.56.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/e21ef464-e5d2-4a98-8fb8-663f813889a4)


### The Meetings Wireframes
1. ![Снимок экрана 2021-07-06 в 16.43.50.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/4f73d0d8-71fc-4d56-abf3-74d00ea77cb8)

[...]

PR #6688, opened 2021-09-21
Event form - underline tabs + leaving popup

Poppe-A:


phpgeekfr (approved):
Good job ! Can you remove the comment `#5895` that you already taken care of ?

PR #6689, opened 2021-09-21
do not use 4Go/cloud function by default #6622

### LINKED ISSUE #6622: [Cleanup] do not use 4Go/cloud function by default
Currently the cloud functions use 4Go of memory by default. Most of them don't require this amount of memory.
- Update the default value
- Redeploy everything
- Test that it's still working

image

phpgeekfr:
fix #6622


phpgeekfr:
image

GrandSchtroumpf (approved):

PR #6691, opened 2021-09-21
Mano e2e financiers investment email feat

xmano:
Implement investor email  E2E for financiers


apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
```diff
@@ -106,8 +106,7 @@ export class MarketplaceMovieViewComponent implements OnInit {
   }
 
   async sendEmail(emailData: EmailData, title: string, orgs: Organization[]) {
-
-    this.dialogRef.close();
+    let edata = null;
```
phpgeekfr, line 109: looks like `edeta` content is not used. Can you turn it into a boolean if this is realy needed ? 
also, rename it to something more understandable please.

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
```diff
@@ -133,6 +132,12 @@ export class MarketplaceMovieViewComponent implements OnInit {
           title,
         };
 
+        if (('Cypress' in window) && (edata === null) && (toUser.firstName !== userSubject.firstName)) {
```
phpgeekfr, line 135: you are using `('Cypress' in window)` twice. Create a const for this instead.

I don't understand the use of `(toUser.firstName !== userSubject.firstName)` here

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
```diff
@@ -133,6 +132,12 @@ export class MarketplaceMovieViewComponent implements OnInit {
...
+          edata = JSON.stringify(data);
+          window['Cypress_e2e'] = data;
+          console.log(data);
```
phpgeekfr, line 138: can you remove console.log  ?

phpgeekfr (approved):
approved with comments

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
xmano: SendGrid does not seem to send mail when from and to are same.. this breaks my test. To avoid that I am doing this.

I will take of const var.

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
xmano: Renamed to email_ready & converted to boolean. It is needed for correct operation.

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
xmano: `const cyCheck = 'Cypress' in window;`

added

apps/financiers/financiers/src/app/marketplace/title/view/view.component.ts
xmano: removed

xmano:
The test for touser.email is not good since it is overwritten by the test and does not remain unique.
We will need additional conditions and logic to go with test for email id. For now using recipient name and test work nice.

image

PR #6701, opened 2021-09-22
New Table

GrandSchtroumpf:
Rework the bf-table
Implement it in crm organizations


apps/crm/crm/src/app/crm/pages/organizations/organizations.component.ts
```diff
@@ -15,30 +15,8 @@ import { take } from 'rxjs/operators';
...
-    'denomination.public',
-    'addresses.main.country',
-    'status',
-    'activity',
-    'email',
-    'appAccess',
-  ];
-  public orgs$ = this.service.valueChanges().pipe(take(1)).toPromise();
+  // We take two to make sure we don't only get indexedDB
+  public orgs$ = this.service.valueChanges().pipe(take(2));
```
phpgeekfr, line 19: is it possible to create an alias for this ? So we don't have to remember to use take(2) ?

libs/ui/src/lib/list/table/table.component.ts
```diff
@@ -0,0 +1,136 @@
...
+        if (!pageSize) return data;
+        return data.slice(pageIndex * pageSize, (pageIndex + 1) * pageSize);
+      })
+    );
+  }
+
+  private $sort(data: T[]) {
+    const sorting = this.columns.filter(c => c.isSortable).map(column => column.$sort(data));
+    if (!sorting.length) return of(data);
+    return merge(...sorting).pipe(startWith(data)); // Prevent first merging to overfire
```
phpgeekfr, line 125: I don't get this comment `// Prevent first merging to overfire`

apps/crm/crm/src/app/crm/pages/organizations/organizations.component.html
```diff
@@ -1,56 +1,58 @@
...
-      <ng-template colRef="denomination.full" let-denominationFull>
-        {{ denominationFull }}
-      </ng-template>
-      <ng-template colRef="denomination.public" let-denominationPublic>
-        {{ denominationPublic }}
-      </ng-template>
-      <ng-template colRef="addresses.main.country" let-country>
-        {{ country ? country : '--' }}
+  <bf-table clickable [source]="orgs" pagination="5" useFilter (rowClick)="goToEdit($event)">
+    <ng-template colRef="id" let-id sort>{{ id }}</ng-template>
```
phpgeekfr, line 23: so with this new component, no more `columnsToFilter` or `initialColumns` ? We need to specify inside the component all the columns we want to use ?

libs/ui/src/lib/list/table/table.component.scss
```diff
@@ -0,0 +1,168 @@
...
+tr {
+  background: inherit;
+}
+
+th, td {
+  border-bottom-width: 1px;
+  border-bottom-style: solid;
+  max-width: 800px;
+  
+  line-height: 18px;  // Custom Font (validated with Stephanie)
```
phpgeekfr, line 44: do you want to keep this `validated with Stephanie` ?

phpgeekfr (approved):
LGTM 
Depending of the rows displayed , the last column does not have same height at each page: 
image
VS
image


So the paginator is not at the same location

apps/crm/crm/src/app/crm/pages/organizations/organizations.component.ts
GrandSchtroumpf: For now I've got no idea on how to make it better. Ideally it should be on akita-ng-fire...
If we do an alias, we need to do it on every service...

libs/ui/src/lib/list/table/table.component.ts
GrandSchtroumpf: ah yes, good point. I removed the code associated with this comment

libs/ui/src/lib/list/table/table.component.ts
```diff
@@ -0,0 +1,136 @@
...
+@Directive({ selector: '[colRef]' })
+export class ColumnDirective<T> {
+  control = new FormControl(true);
+
+  @Input('colRef') name!: string;
+  @Input() label?: string;
+  @Input() @boolean sticky = false;
+  @Input() @boolean defaultSort: boolean;
+  @Input() sort?: '' | ((a: any, b: any) => number);
+  @Input() filter = (input: string, value: any, row: T) => {
```
pldespaigne, line 39: Could be nice to have some ts-doc comments so that the devs will know exactly what are those inputs and they affect the behavior of the component.

libs/ui/src/lib/list/table/table.component.scss
GrandSchtroumpf: yes, in case someone see this override, so they know it's validated by Stephanie.

libs/ui/src/lib/list/table/table.component.ts
```diff
@@ -0,0 +1,136 @@
...
+  @Input() @boolean useFilter: boolean;
+  @Input() set source(source: T[]) {
+    this.dataSource.next(source);
+  }
+  @Input() set pagination(amount: number | string) {
+    const pageSize = coerceNumberProperty(amount);
+    if (typeof pageSize === 'number') this.paginator.pageSize = pageSize;
+  }
+
+  @Output() rowClick = new EventEmitter<T>();
```
pldespaigne, line 101: Could be nice to have some ts-doc comments so that the devs will know exactly what are those inputs and they affect the behavior of the component.

pldespaigne (approved):

apps/crm/crm/src/app/crm/pages/organizations/organizations.component.html
GrandSchtroumpf: yes

PR #6717, opened 2021-09-23
fixed e2e test in emulator. And maybe some cases in production

phpgeekfr:


libs/user/src/lib/auth/pages/identity/identity.component.ts
```diff
@@ -197,6 +180,8 @@ export class IdentityComponent implements OnInit, OnDestroy {
         return;
       }
 
+      const user = await this.createUser(this.form.value);
```
phpgeekfr, line 183: I had to move the createUser after the `const unique = await this.orgService.uniqueOrgName(denomination.full);`  because the guard in `libs/user/src/lib/auth/guard/identity.guard.ts` was update too fast (because of auth emulator).

Because of 
```
if (!userAuth.emailVerified) {
          return of(this.router.navigate(['c/organization/join-congratulations']));
        }
```
use was redirected before response of `const unique = await this.orgService.uniqueOrgName(denomination.full);`

firestore.rules
```diff
@@ -216,8 +216,7 @@ service cloud.firestore {
     /// RULES ///
 
     match /orgs/{orgId} {
-      allow read: if isSignedIn()
-        && notInMaintenance();
+      allow read: if notInMaintenance();
```
phpgeekfr, line 219: I had to remove `isSignedIn()` because `this.orgService.uniqueOrgName(denomination.full);` needs to be able to read org list. 
This is not a problem since everyone can create an account right now and this will also be needed for the new event type without login-in

apps/catalog/catalog-e2e/src/integration/landing/account-creation.spec.ts
```diff
@@ -75,6 +75,19 @@ describe('User can create new account and create a new organization', () => {
...
+    p1.fillUserInformations(USER);
+
+    const p2 = new OrganizationLiteFormPage();
+    p2.createNewDashboardOrg();
+
+    p1.clickTermsAndCondition();
+    p1.clickPrivacyPolicy();
+    p1.submitForm();
+    assertMoveTo(IDENTITYPATH);
+  })
```
phpgeekfr, line 90: this test `it('use already existing org name', () => {` which was not well named, need the `it('Fill all the fields', () => {` one to be executed . So I moved it to the same `describe`

libs/user/src/lib/auth/pages/identity/identity.component.ts
xmano: Would be good idea to comment the sequence of operation.

Any other place this sequence may be used and need correction?

xmano (approved):
Approved with comments. Please verify.

libs/user/src/lib/auth/pages/identity/identity.component.ts
phpgeekfr: I think this is the only place where we can have this case where `uniqueOrgName` is called without a logged in user

firestore.rules
phpgeekfr: I added new tests with not logged in user

Issue #6643, opened 2021-09-09
Screening - Counter Component


RemcoSimonides:
Create animating counter component for Screening. Discuss the details with Stephanie :)

[Zeplin](https://app.zeplin.io/project/61129744b469ce8a0c470c3e/screen/61156022e602f59b78fe192d)

![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/06da3461-3608-44df-884e-4d4d4d2a75c9)

Original design issue: #6225

PR #6769, opened 2021-09-28
Countdown working

Poppe-A:
#6643


RemcoSimonides (changes requested):
Can you make it so you don't need `event-countdown` component but only the `counter` component?

I think `counter` component only needs start date as input and you can move all the `calcDateDiff()` code into this `counter` component so it's reusable

libs/event/src/lib/components/countdown/counter-display/counter-display.component.html
```diff
@@ -0,0 +1,58 @@
...
+
+  <!-- Minutes -->
+  <span fxLayout="column" fxLayoutAlign="center center"> 
+    <div class="numberContainer" fxLayout fxLayoutAlign="center center">
+      <p class="mat-title" [@testAnim2]="newMinutes">
+        {{ minutes }}
+      </p> 
+      <p 
+      class="mat-title"
+      *ngIf="newMinutes !== false"
```
RemcoSimonides, line 50: No need for `!== false`, `*ngIf="newMinutes` is enough probably (if this component shouldnt show up when newMinutes is undefined, 0, '', or null)

libs/event/src/lib/components/countdown/counter-display/counter-display.component.html
```diff
@@ -0,0 +1,58 @@
...
+      <p
+      class="mat-title"
+      *ngIf="newDays !== false"
+      (@testAnim.done)="onAnimationDone()"
+      [@testAnim]>
+        {{ newDays }}
+      </p>
+    </div>
+    <p class="mat-caption">days</p>
+  </span>
```
RemcoSimonides, line 19: We have 3x the same component but with different values, right? There should be a way to refactor this to 1 ^^ not sure how yet though... let me think

RemcoSimonides:
Can you move the `counter` component to the ui lib btw? as it will not only be used for events

libs/ui/src/lib/countdown/countdown.component.html
```diff
@@ -0,0 +1,27 @@
+<section fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="16px">
+  <span class="mat-body-1">Event will start in :</span>
+  <div class="counter-display" fxLayout="row" fxLayoutAlign="space-between center" fxLayoutGap="24px">
```
RemcoSimonides, line 3: You could rename this element to `article`. This way you only have 1 article and 1 div in this element and you don't need the class names to access the element in the css. `fxLayout="row"` can be just `fxLayout`

libs/ui/src/lib/countdown/countdown.component.html
```diff
@@ -0,0 +1,27 @@
...
+    <ng-container *ngFor="let unit of timeUnits; let last = last">
+      <span fxLayout="column" fxLayoutAlign="center center"> 
+        <div class="numberContainer" fxLayout fxLayoutAlign="center center">
```
RemcoSimonides, line 6: remove `class=numberContainer` if you did previous mentioned change ☝️

libs/event/src/lib/layout/view/view.module.ts
```diff
@@ -11,7 +11,7 @@ import { MatButtonModule } from '@angular/material/button';
 import { MatIconModule } from '@angular/material/icon';
 import { MatTooltipModule } from '@angular/material/tooltip';
 import { ImageModule } from '@blockframes/media/image/directives/image.module';
-
+import { CountdownModule } from '@blockframes/ui/countdown/countdown.module'
```
RemcoSimonides, line 14: `;` and add new line between imports and @NgModule

libs/ui/src/lib/countdown/countdown.component.html
```diff
@@ -0,0 +1,27 @@
...
+            *ngIf="time[unit].current > -1"
+            class="mat-title"  
+            [@numberAnim]="time[unit].new">
+              {{ time[unit].current | number: "2.0"}}
+          </p> 
+          <p
+          class="mat-title"
+          *ngIf="time[unit].new > -1"
+          (@tempNumberAnim.done)="onAnimationDone()"
+          [@tempNumberAnim]>
```
RemcoSimonides, line 17: Make indentation same as previous `<p>`

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,114 @@
...
+        })
+      ),
+    ]),
+  ]);
+}
+
+/* 
+  This component is an adaptable countdown : 
+  It will only display the time units that it will receive as a property array :
+  example : ['hours', 'minutes']
```
RemcoSimonides, line 43: Move this comment above the `@Input()` variable in the class

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  It also needs a date property to calculate the end of the countdown
```
RemcoSimonides, line 44: Above `@Input() date`

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  Important : As we sometime need to display a "0" value, we cannot use 0 as "false" value.
+  Here we'll use -1 to express false or undefined.
```
RemcoSimonides, line 46: Move above `time` variable

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  selector: 'countdown-timer',
+  templateUrl: './countdown.component.html',
+  styleUrls: ['./countdown.component.scss'],
+  animations: [tempAnim(), anim()],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class CountdownComponent implements OnInit, OnDestroy {
+  private subscription: Subscription;
+  @Input() date;
+  @Input() timeUnits;
```
RemcoSimonides, line 58: Add types to these variables

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  
+  public time = {
+    days: { current: -1, new: -1},
+    hours: { current: -1, new: -1},
+    minutes: { current: -1, new: -1},
+    seconds: { current: -1, new: -1}
+  }
+
+  constructor(private ref: ChangeDetectorRef) {
+  }
```
RemcoSimonides, line 68: Move `}` to previous line since there is nothing between the brackets

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  ngOnInit() {
+    this.subscription = interval(1000).pipe(
+      startWith(0),
+      tap(_ => this.calcDateDiff()),
+    ).subscribe()
```
RemcoSimonides, line 74: `;`

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.ref.markForCheck();
+  }
+
+  onAnimationDone() {
+    this.timeUnits.forEach(elm => {
+      if(this.time[elm].new > -1) {
+        this.time[elm].current = this.time[elm].new as number;
+        this.time[elm].new = -1
+      }
+    });
```
RemcoSimonides, line 112: For naming consistency we should name `elm` to `unit`

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,114 @@
...
+      }
+    }
+
+    this.ref.markForCheck();
+  }
+
+  onAnimationDone() {
+    this.timeUnits.forEach(elm => {
+      if(this.time[elm].new > -1) {
+        this.time[elm].current = this.time[elm].new as number;
```
RemcoSimonides, line 109: Do you need `as number`?

RemcoSimonides (changes requested):
Cleaning time

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,115 @@
...
+        style({
+          bottom: '100%',
+        })
+      ),
+    ]),
+  ]);
+}
+
+@Component({
+  selector: 'countdown-timer',
```
RemcoSimonides, line 41: As the `date` and `timeUnits` inputs are mandatory, we should say in this selector they are mandatory by 
`selector: '[date][timeUnits] countdown-timer'`

libs/ui/src/lib/countdown/countdown.component.ts
```diff
@@ -0,0 +1,115 @@
...
+    this.ref.markForCheck();
+  }
+
+  onAnimationDone() {
+    this.timeUnits.forEach(unit => {
+      if(this.time[unit].new > -1) {
+        this.time[unit].current = this.time[unit].new;
+        this.time[unit].new = -1
+      }
+    });
```
RemcoSimonides, line 113: For code consistency you should use either both a `for const ... of ...` or both `forEach`.

RemcoSimonides (approved):

libs/ui/src/lib/countdown/countdown.component.html
```diff
@@ -0,0 +1,27 @@
+<section fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="16px">
+  <span class="mat-body-1">Event will start in :</span>
```
GrandSchtroumpf, line 2: Use `p` or a heading instead of span. `p` & headings are `display: block` so you don't need a column around.

libs/ui/src/lib/countdown/countdown.component.html
```diff
@@ -0,0 +1,27 @@
+<section fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="16px">
```
GrandSchtroumpf, line 1: You don't need this section.

libs/ui/src/lib/countdown/countdown.component.html
```diff
@@ -0,0 +1,12 @@
+  <p class="mat-body-1">Event will start in :</p>
```
RemcoSimonides, line 1: This shouldn't be in this component as it is specific for event. If you want to reuse the component on the maintenace page for example, then this needs to be moved.

PR #6770, opened 2021-09-28
cleaning imports

phpgeekfr:
movie some models to `****.firestore.ts` and updated `apps.ts`

As @GrandSchtroumpf suggested, I will create an issue to move all firestore.ts files into a new `models` lib

@RemcoSimonides can you try to deploy again with these changes ? It worked on my side  but just to be sure ..

### DIFF
--- a/libs/contract/src/lib/bucket/+state/bucket.firestore.ts
+++ b/libs/contract/src/lib/bucket/+state/bucket.firestore.ts
@@ -1,6 +1,6 @@
 import { Holdback } from '@blockframes/contract/contract/+state/contract.firestore';
 import { BucketTerm } from '@blockframes/contract/term/+state/term.firestore';
-import { MovieCurrency } from '@blockframes/utils/static-model';
+import type { MovieCurrency } from '@blockframes/utils/static-model';
 
 export interface Bucket {
   id: string;
--- a/libs/contract/src/lib/bucket/+state/bucket.firestore.ts
+++ b/libs/contract/src/lib/bucket/+state/bucket.firestore.ts
@@ -1,0 +1,28 @@
+import { Holdback } from '@blockframes/contract/contract/+state/contract.firestore';
+import { BucketTerm } from '@blockframes/contract/term/+state/term.firestore';
+import { MovieCurrency } from '@blockframes/utils/static-model';
+
+export interface Bucket {
+  id: string;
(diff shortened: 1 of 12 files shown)


phpgeekfr:
issue https://github.com/blockframes/blockframes/issues/6771

RemcoSimonides:
Deploy succesful :)

libs/utils/src/lib/apps.ts
```diff
@@ -124,7 +122,7 @@ export function getAppName(slug: App, short = false) {
  * getOrgAppAccess(orgA); // ['catalog', 'festival']
  * getOrgAppAccess(orgB); // ['festival']
  */
-export function getOrgAppAccess(org: OrganizationDocument | Organization, first: App = 'festival'): App[] {
+export function getOrgAppAccess(org: OrganizationDocument | OrganizationBase<Date>, first: App = 'festival'): App[] {
```
RemcoSimonides, line 125: It wasn't possible to move Organization to `.firestore.ts`? or would it break too much imports throughout the app?

RemcoSimonides (approved):

phpgeekfr:
> Deploy succesful :)

Ho yeah ! I really don't get why you are the first with this issue. Problem was here for a long time

libs/contract/src/lib/bucket/+state/bucket.firestore.ts
```diff
@@ -0,0 +1,28 @@
+import { Holdback } from '@blockframes/contract/contract/+state/contract.firestore';
+import { BucketTerm } from '@blockframes/contract/term/+state/term.firestore';
+import { MovieCurrency } from '@blockframes/utils/static-model';
```
GrandSchtroumpf, line 3: Maybe be more precise here to not import all the static-model. Another option is to put `import type {...}` like that typescript knows if should remove this line at compile time.

libs/contract/src/lib/contract/+state/contract.firestore.ts
```diff
@@ -0,0 +1,54 @@
+
+import { DocumentMeta } from "@blockframes/utils/models-meta";
+import { Media, Territory } from "@blockframes/utils/static-model";
```
GrandSchtroumpf, line 3: same as above

GrandSchtroumpf (approved):

PR #6774, opened 2021-09-29
6686 account creation bug

### LINKED ISSUE #6686: Archipel Market - Account Creation bug (Cannot read property)
Apparently user has received an invitation as all the infos are filled in. 
Nevertheless, the field "invitation pass" doesn't appear. 

User receives an error snackbar "Cannot read property 'uid' of null" bug. 

Then user makes another attempt, here the invitation pass appears. 

The example can be found [**here**](https://drive.google.com/drive/folders/1oFkA2slQHqmAR00QcRoLmSAoZBX-w_dY?ths=true)
_timecode: 0:25_

UPDT. : user connected on the 6th of September

RemcoSimonides:
Hopefully fix #6686 and otherwise improve performance on signup

### DIFF
--- a/apps/backend-functions/src/main.ts
+++ b/apps/backend-functions/src/main.ts
@@ -22,14 +22,13 @@ import { linkFile, getMediaToken as _getMediaToken } from './media';
 import { onEventDelete } from './event';
 import { getTwilioAccessToken, twilioWebhook as _twilioWebhook } from './twilio';
 import { eventWebhook as sendgridEventWebhook } from './sendgrid';
-import { heavyConfig, superHeavyConfig } from '@blockframes/firebase-utils';
+import { hotConfig, heavyConfig, superHeavyConfig } from '@blockframes/firebase-utils';
 import { onNotificationCreate } from './notification';
 import { importAnalytics } from './pubsub/daily-analytics-import';
 import { onOfferCreate } from './offer';
 import { onContractCreate, onContractDelete, onContractUpdate } from './contracts';
 import { onTermDelete } from './terms';
 import { downloadVideo } from './rescue';
-import { production } from '@env';
 
 console.log('Function instance loaded');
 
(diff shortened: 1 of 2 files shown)


apps/backend-functions/src/main.ts
```diff
@@ -113,7 +114,8 @@ export const onInvitationWriteEvent = onDocumentWrite('invitations/{invitationID
 export const hasUserAnOrgOrIsAlreadyInvited = functions().https.onCall(invitations.hasUserAnOrgOrIsAlreadyInvited);
 
 /** Used to get invitation linked to an email when users signup for the first time */
-export const getInvitationLinkedToEmail = functions().https.onCall(invitations.getInvitationLinkedToEmail);
+/** Keep one instance warm on production in order to show up invitation pass field faster when Signing up */
+export const getInvitationLinkedToEmail = functions({ maxInstances: production ? 1 : 0}).https.onCall(invitations.getInvitationLinkedToEmail);
```
phpgeekfr, line 118: bad indent + with this `{ maxInstances: production ? 1 : 0}` you are going to erase defaultConfig ```
{
  timeoutSeconds: 60,
  memory: '256MB',
};
```

So maybe `{ ...defaultConfig, maxInstances: production ? 1 : 0 } ` or create a specific config in `libs/firebase-utils/src/lib/firebase-utils.ts`

phpgeekfr: I will update https://www.notion.so/cascade8/Release-troubleshooting-8085a54303de4019b0a129116d60fa13 or do it if you want to remind us to check the price of this feature for one or two months

phpgeekfr: can we see on cloud.google or firebase interface that the function is kept as warm ?

phpgeekfr: > I will update https://www.notion.so/cascade8/Release-troubleshooting-8085a54303de4019b0a129116d60fa13 or do it if you want to remind us to check the price of this feature for one or two months

I updated the doc

phpgeekfr (approved):

PR #6790, opened 2021-09-30
Update watermark fix #6754

### LINKED ISSUE #6754: Update watermark when user name is changed
I changed my name within app but watermark didn't. It's important for users with generic email address that regularly pass from one person to another.

phpgeekfr:
fix #6754 

To be able to refresh watermark on imgIx side, we needed to make it have an unique filename so it does not stay the same on imgIx cache.

During this, I noticed that storagePath of watermarks were including `public/`. But this information is on the `privacy` field of the `storageFile` object. This was not noticed before because filename was always the same.


phpgeekfr:
Not ready yet, previous watermarks are not deleted and db is updated with good url but then re-updated with previous one.

pldespaigne (approved):

PR #6792, opened 2021-10-01
Correct storage cleaning

Poppe-A:


apps/backend-ops/src/storage-cleaning.ts
```diff
@@ -132,7 +132,7 @@ export async function cleanWatermarkDir(bucket: Bucket) {
       // Cleaning files that have a too long name
       if (await f.delete()) { deleted++; }
     } else {
-      const userId = f.name.split('/')[2].replace('.svg', ''); // @TODO #6754
+      const userId = f.name.split('/')[2].replace('.svg', '').split('-')[0];
```
phpgeekfr, line 135: In fact this is not useful.
Example: 
```
console.log("public/users/LIXEkDP2jTXPxHcfZ9h05NVdu3p2/watermark/LIXEkDP2jTXPxHcfZ9h05NVdu3p2-aaaaaaaaaaaaaaaaaaaa.svg".split('/')[2].replace('.svg', '').split('-')[0])
```

Or 
```
console.log("public/users/LIXEkDP2jTXPxHcfZ9h05NVdu3p2/watermark/LIXEkDP2jTXPxHcfZ9h05NVdu3p2-aaaaaaaaaaaaaaaaaaaa.svg".split('/')[2].replace('.svg', ''))
```

Will output same result because we look at what is after the second `/`.


this method `cleanWatermarkDir` is really old and not useful anymore. I will remove it

PR #6795, opened 2021-10-04
Excel Import : more refcatoring

pldespaigne:
more refactoring


libs/import/src/lib/utils.ts
```diff
@@ -6,9 +6,13 @@ import { Term } from "@blockframes/contract/term/+state/term.model";
 import { SheetTab } from "@blockframes/utils/spreadsheet";
 
 
+export const spreadsheetImportTypes = [ 'titles', 'organizations', 'contracts' ] as const;
```
RemcoSimonides, line 9: Are you planning on using this array?
Otherwise only `export type SpreadsheetImportType = 'titles' | 'organizations' | 'contracts';` would suffice.

RemcoSimonides (approved):

Issue #6988, opened 2021-10-28
Camille's review - Dashboard Avails & Sales


cams-goussale:
Browser: Chrome

### Page: My sales

- [x] We get an error when trying to import Sales; see Excel attached
![image](https://user-images.githubusercontent.com/93206363/139284819-c3d97cb3-4ca5-495a-8b96-9eef9f9a920f.png)

[TEST_281021_import contract test sept 21.xlsx](https://github.com/blockframes/blockframes/files/7435373/TEST_281021_import.contract.test.sept.21.xlsx)

- [ ] When a sale is accepted, the title disappears from My Avails list. See below example with title CRYSTAL SWAN:

![image](https://user-images.githubusercontent.com/93206363/139281325-e02d0642-f584-4cce-a658-679835b10436.png)

![image](https://user-images.githubusercontent.com/93206363/139281378-0ee770b8-c89f-4073-b5bc-c5437a7fd5e2.png)

- [x] Should be the International Title instead of Original Title

![image.png](https://images.zenhubusercontent.com/617811d92b89e7733866d8a0/0822952f-48f5-4cb3-944e-42087943f539)

- [x] Should be "STATUS" instead of "APPROVAL"

![image.png](https://images.zenhubusercontent.com/617811d92b89e7733866d8a0/d2a02d0d-ba79-4d98-9fb0-5fcac98b61cb)

iPR0D1G1:
> > * [ ]  When a sale is accepted, the title disappears from My Avails list. See below example with title CRYSTAL SWAN:
> 
> @iPR0D1G1 did you forgot to fix this item or simply to mark it as done ?

Not fix nor mark as done. Tried reproducing and once the sale was accepted, the title didn't vanish so I just took it for a false positive. But my environment shouldn't be taken as a litmus test as it's riddled with stale data.

pldespaigne:
Ok it's good to comment something like this, so we are aware of your work:
"Can't reproduce locally" 

Closing for now, feel free to re-open if needed

PR #7017, opened 2021-11-02
Solve some excel import bugs

pldespaigne:
- lot's of bugs
- and also first item of #6988

### DIFF
--- a/libs/import/src/lib/view-extracted-elements/titles/utils.ts
+++ b/libs/import/src/lib/view-extracted-elements/titles/utils.ts
@@ -493,28 +493,22 @@ export async function formatTitle(
   for (const result of results) {
     const { data, errors, warnings } = result;
 
-    const stakeholders: MovieStakeholders = {
-      productionCompany: [],
-      coProductionCompany:  [],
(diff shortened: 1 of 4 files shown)


libs/import/src/lib/view-extracted-elements/titles/utils.ts
```diff
@@ -495,19 +494,35 @@ export async function formatTitle(
     const { data, errors, warnings } = result;
 
     const stakeholders: MovieStakeholders = {
```
GrandSchtroumpf, line 496: You can do : 
```typescript
if (!data.stakeholders) warnings.push(optionalWarning({ field: 'stakeholders', name: 'Stakeholders' }));
if (!data.languages) warnings.push(optionalWarning({ field: 'languages', name: 'Languages' }));
const getStakeholders = (role: string) => data.stakeholders?.filter(s => s.role === role) ?? [];
const stakeholders = {
  productionCompany: getStakeholders('executiveProducer')
  ...
}
```

GrandSchtroumpf (approved):

Issue #6917, opened 2021-10-20
Sending event invitation to user without organisation


Poppe-A:
To send an invitation email depending on the user status (registered or not), we only check if the user doesn't exist in DB, or if it has an organisation.
After sending an invitation to an non registered email, the email adress is added in DB, but the user is not linked to an org.
So this case is not covered by the code (email register but no org) and we don't send anything.

PR #7024, opened 2021-11-03
send invit email to user with no orgId

Poppe-A:
#6917

### DIFF
--- a/apps/backend-functions/src/internals/users.ts
+++ b/apps/backend-functions/src/internals/users.ts
@@ -40,12 +40,12 @@ export const getOrInviteUserByMail = async (
     //if user exists but has no orgId, we still want to send him an invitation email
     if (invitation.type === "attendEvent" && !user.orgId) {
       const invitationTemplateId = templateIds.user.credentials.attendEventRemindInvitationPass;
-      sendMailFromTemplate({
+      await sendMailFromTemplate({
         to: email,
         templateId: invitationTemplateId,
         data: { 
           event: eventData,
-          orgName: invitation.fromOrg.denomination.full
+          org: getOrgEmailData(invitation.fromOrg)
         }
       }, app)
     }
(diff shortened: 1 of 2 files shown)


apps/backend-functions/src/internals/users.ts
```diff
@@ -36,6 +36,20 @@ export const getOrInviteUserByMail = async (
...
+    
+    //if user exists but has no orgId, we still want to send him an invitation email
+    if (invitation.type === "attendEvent" && !user.orgId) {
+      const invitationTemplateId = templateIds.user.credentials.attendEventRemindInvitationPass;
+      sendMailFromTemplate({
+        to: email,
+        templateId: invitationTemplateId,
+        data: { 
+          event: eventData,
+          orgName: invitation.fromOrg.denomination.full
```
phpgeekfr, line 48: can you use `getOrgEmailData(fromOrg);` as we do in `userInvite()` ?

and `getUserEmailData` for user if useful ?

apps/backend-functions/src/internals/users.ts
```diff
@@ -36,6 +36,20 @@ export const getOrInviteUserByMail = async (
   try {
     const { uid } = await auth.getUserByEmail(email);
     const user = await getDocument<PublicUser>(`users/${uid}`);
+    
+    //if user exists but has no orgId, we still want to send him an invitation email
+    if (invitation.type === "attendEvent" && !user.orgId) {
+      const invitationTemplateId = templateIds.user.credentials.attendEventRemindInvitationPass;
+      sendMailFromTemplate({
```
phpgeekfr, line 43: no need to await since the rest of the process is not depending of this email ?

phpgeekfr (approved):

apps/backend-functions/src/internals/users.ts
Poppe-A: Ok for getOrgEmailData,
But I don't think I need getUserEmailData, since I only user user email

PR #7037, opened 2021-11-04
Event reminder

Poppe-A:
#7030


libs/utils/src/lib/emails/utils.ts
```diff
@@ -57,7 +57,8 @@ export interface UserEmailData {
   firstName?: string,
   lastName?: string,
   email: string,
-  password?: string
+  password?: string,
+  isRegistered? : boolean
```
phpgeekfr, line 61: extra space between `isRegistered?` and `:` ;)

phpgeekfr (approved):

PR #7048, opened 2021-11-05
use sessionsStorage instead of akita store

phpgeekfr:


libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -170,13 +172,19 @@ export class AuthService extends FireAuthService<AuthState> {
...
-   * Sign in an user anonymously, he will get an uid and will be stored in firebase auth as anonymous
+   * Sign in an user anonymously if not already connected,
+   * he will get an uid and will be stored in firebase auth as anonymous
    * @returns Promise<firebase.auth.UserCredential>
    */
   async signInAnonymously() {
-    const creds = await firebase.auth().signInAnonymously();
-    this.store.updateAnonymousCredentials({ uid: creds.user.uid }, { reset: true });
-    return creds;
+    const currentUser = firebase.auth().currentUser;
```
GrandSchtroumpf, line 180: you shouldn't use `firebase` directly. The AuthService extends the FireAuthService, which provide the current `auth` : 
```typescript
if (this.user) {
  return this.updateAnonymousCredentials({ uid: this.user.uid });
}
const creds = await this.auth.signInAnonymously();
return this.updateAnonymousCredentials({ uid: creds.user.uid }, { reset: true });
```
Return the creds in the `updateAnonymousCredentials` method

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -241,8 +249,48 @@ export class AuthService extends FireAuthService<AuthState> {
...
     // Clean anonymousCredentials
     await this.onSignout();
-    this.store.reset();
     // Delete (and logout) user
     return firebase.auth().currentUser.delete();
   }
+
+  public resetAnonymousCredentials() {
+    const keys = ['uid', 'role', 'email', 'invitationId', 'lastName', 'firstName'];  // keys of AnonymousCredentials
+    Object.keys(keys).forEach(k => {
```
GrandSchtroumpf, line 258: keys is an array, so `Object.keys(keys)` here is going to be a list of indexes : 
```typescript
const keys = ['uid', 'role', 'email', 'invitationId', 'lastName', 'firstName'];
keys.forEach(key => sessionStorage.removeItem(`anonymousCredentials.${k}`));
```

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -241,8 +249,48 @@ export class AuthService extends FireAuthService<AuthState> {
...
+      sessionStorage.removeItem(`anonymousCredentials.${k}`);
+    });
+  }
+
+  public updateAnonymousCredentials(anonymousCredentials: Partial<AnonymousCredentials>, options?: { reset: boolean }) {
```
GrandSchtroumpf, line 263: rename `anonymousCredentials` to `creds`. We already know it's user credentials (because we are in the `AuthService`) and we know it's anonymous because of the name of the method.

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -241,8 +249,48 @@ export class AuthService extends FireAuthService<AuthState> {
...
+
+    Object.keys(anonymousCredentials).forEach(k => {
+      if (!anonymousCredentials[k]) {
+        sessionStorage.removeItem(`anonymousCredentials.${k}`);
+      } else {
+        sessionStorage.setItem(`anonymousCredentials.${k}`, anonymousCredentials[k]);
+      }
+    })
+
+    this.anonymousCredentials$.next(this.anonymousCredentials);
```
GrandSchtroumpf, line 277: return the credential at the end.

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -241,8 +249,48 @@ export class AuthService extends FireAuthService<AuthState> {
...
+    });
+  }
+
+  public updateAnonymousCredentials(anonymousCredentials: Partial<AnonymousCredentials>, options?: { reset: boolean }) {
+
+    if (options?.reset) { // @TODO #6756 useful ?
+      this.resetAnonymousCredentials();
+    }
+
+    Object.keys(anonymousCredentials).forEach(k => {
```
GrandSchtroumpf, line 269: Prefere : 
```typescript
for (const [key, value] of Object.entries(creds)) {
  !!value
    ? sessionStorage.removeItem(`anonymousCredentials.${key}`)
    : sessionStorage.setItem(`anonymousCredentials.${k}`, value);
}
```

libs/user/src/lib/auth/guard/anonymous-auth-guard.ts
```diff
@@ -20,10 +20,7 @@ export class AnonymousAuthGuard implements CanActivate, CanDeactivate<unknown> {
    * @returns Promise<boolean>
   */
   async canActivate() {
-    const currentUser = await this.afAuth.currentUser;
-    if (!currentUser) {
-      await this.authService.signInAnonymously();
-    }
+    await this.authService.signInAnonymously();
```
GrandSchtroumpf, line 23: this is really not a good idea to automatically signin in a guard I think. But we can discuss that later.

phpgeekfr: OK to discuss it later. Is signs up user only if he is not already. Every pages under /event/eventId required at least anonymous identity

PR #7070, opened 2021-11-09
review - first step

Poppe-A:
Review correction for feat-6756 branch


libs/ui/src/lib/layout/event/event.module.ts
```diff
@@ -10,7 +10,7 @@ import { AuthWidgetModule } from '@blockframes/auth/components/widget/widget.mod
 import { AppBarModule } from '@blockframes/ui/app-bar';
 import { AppLogoModule } from '../app-logo/app-logo.module';
 import { FooterModule } from '../footer/footer.module';
-import { AsideModule } from '../marketplace/aside/aside.module';
+import { AsideModule } from '../../../../../../apps/festival/festival/src/app/marketplace/aside/aside.module';
```
phpgeekfr, line 13: This is not OK. I checked with @GrandSchtroumpf and this means that we have a structure problem. Can update this PR without this change for now please ?

PR #7079, opened 2021-11-09
Excel import: creation of objects in db

pldespaigne:
fixing bugs related to the creation of org/movie/contract in db just after the ecel import

### DIFF
--- a/libs/import/src/lib/table-extracted-elements/contracts/contracts.component.ts
+++ b/libs/import/src/lib/table-extracted-elements/contracts/contracts.component.ts
@@ -112,6 +112,7 @@ export class TableExtractedContractsComponent implements OnInit {
    */
   private async addContract(importState: ContractsImportState): Promise<boolean> {
     const data = this.rows.data;
+    // TODO INVESTIGATE ERROR WHEN CREATING CONTRACTS : FirebaseError: [code=invalid-argument]: Timestamp seconds out of range: 1332713170800
     const termIds = await this.termService.add(importState.terms);
     importState.contract.termIds = termIds;
     await this.contractService.add({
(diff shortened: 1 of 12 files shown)


libs/import/src/lib/table-extracted-elements/contracts/contracts.component.ts
```diff
@@ -112,6 +112,7 @@ export class TableExtractedContractsComponent implements OnInit {
    */
   private async addContract(importState: ContractsImportState): Promise<boolean> {
     const data = this.rows.data;
+    // TODO INVESTIGATE ERROR WHEN CREATING CONTRACTS : FirebaseError: [code=invalid-argument]: Timestamp seconds out of range: 1332713170800
```
GrandSchtroumpf, line 115: This timestamps looks like millisecond and not second.

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -262,7 +266,8 @@ export async function formatContract(
 
     const term = toTerm(data.term, contract.id, firestore);
 
-    if (contract.type === 'sale') {
+    // for **internal** sales we should check the parentTerm
```
GrandSchtroumpf, line 269: Prefer variables with good naming than comment : 
```typescript
const isInternalSale = contract.type === 'sale' && contract.sellerId === centralOrgId.catalog
```

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -94,7 +94,15 @@ export async function parse(
       const value = Array.isArray(values) ? values : [ values ];
       if (last) {
         const promises = value.map(v => transform(v, entity, state, rowIndex));
-        item[field] = await Promise.all(promises);
+        const results = await Promise.all(promises);
+
+        const validResults = results.filter(r => !(r instanceof ValueWithWarning));
```
GrandSchtroumpf, line 99: what about : 
```typescript
// This make sure filter cast the type
const isWarning = (r: any): r is ValueWithWarning  => r instanceof ValueWithWarning;
item[field] = result.map(r => isWarning(r) ? r.value : r);
const warningResults = result.filter(isWarning).map(r => r.warning);
warnings.concat(warningResults);
```

GrandSchtroumpf (approved):

PR #7082, opened 2021-11-09
Finish review correction

Poppe-A:
Event without login review correction


apps/festival/festival/src/app/event/role/role.module.ts
```diff
@@ -19,6 +20,7 @@ import { AppLogoModule } from '@blockframes/ui/layout/app-logo/app-logo.module';
     FlexLayoutModule,
     MatButtonModule,
     MatCardModule,
+    MatIconModule,
```
phpgeekfr, line 23: Looks like you don"t need it

Issue #7067, opened 2021-11-08
No connexion events - review


Jackseed:
OS: Mac
Browser: Safari (& Brave for meetings)
Env: https://festival-blockframes-bruce.web.app 11/08
Users: dev+gerda-isl@blockframes.io (organizer)

### Event form
- [x] When entering the form, I'm redirected to the statistics tab, it should be a redirection to Main tab (Screening / Meeting) when event is present or future and Statistics when it's closed.
- [x] Wording to change for Privacy settings:
PUBLIC - Open to anyone with a link + tooltip with text "Best for Meetings"
PROTECTED - Open to anyone invited to the event (Archipel Market user or not)
PRIVATE - Open to invited, identified Archipel Market users only + tooltip with text "Best for Screenings"
 ![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/3f9a50ad-110f-4aec-8cb9-190cd363a42a)
- [x] Add a snackbar when clicking on "Save" (and saving): "Event saved." for 4s
- [x] Add a snackbar when clicking on Copy the link for screening button: "Link copied." for 4s
- [x] Add a snackbar when clicking on Copy guest emails button: "Guests' emails copied." for 4s
- [x] Statistics didn't show up for this private event https://festival-blockframes-bruce.web.app/c/o/dashboard/event/20R763fRQUQ8BDzPWn2D/edit/statistics (user above)

### Public / Invitation only connexion page
- [x] Split first name & last name inputs in 2 on the same line instead of one above the other
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/a30b7e9b-8c31-44d2-9ec9-0fb242ddd549)

### Anonymous pages 
- [x] Remove the top right settings, name & profile picture to keep only logo on the toolbar
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/ed5ec72e-998c-4e2a-9cb1-62c886b8b0ab)
- [x] Log out automatically the anonymous account if the user navigates to a page different than the event page
- [x] Event view & session pages:  logo doesn't navigate to landing
- [x] I got this error on Brave on the Organizer / guest page while there is no Gmap 
```
Google Maps JavaScript API error: RefererNotAllowedMapError
https://developers.google.com/maps/documentation/javascript/error-messages#referer-not-allowed-map-error
[...]

PR #7085, opened 2021-11-10
Correct wording and split identity fields

Poppe-A:
#7067


libs/event/src/lib/form/edit-details/edit-details.component.html
```diff
@@ -24,14 +24,14 @@
...
+        PUBLIC - Open to anyone with a link
       </mat-radio-button>
       <mat-radio-button  color="primary" value="invitation-only">
-        INVITATION ONLY - upon invited guests can access your event
+        PRIVATE - Open to anyone invited to the event (Archipel Market user or not)
       </mat-radio-button>
-      <mat-radio-button  color="primary" value="private">
-        PRIVATE - only {{ appName }} users can access your event --- best for Screenings
+      <mat-radio-button  color="primary" value="private" matTooltip="Best for screenings">
+        PROTECTED - Open to invited, identified Archipel Market users only
```
phpgeekfr, line 34: @Jackseed updated the issue, this is now `public` 'protected` `private` 

https://github.com/blockframes/blockframes/issues/7067

you also need to update 
```
export const accessibility = {
  public: 'Public',
  'invitation-only': 'Invitation only',
  private: 'Private',
} as const;
```

in libs/utils/src/lib/static-model/static-model.ts

To update the backend-code with this new changes, we will only need to replace 'invitation-only' to protected everywhere in code but lets do this once it is merged into release/3.5

phpgeekfr (approved):
LGTM

PR #7095, opened 2021-11-11
7057

### LINKED ISSUE #7057: Release 3.5 - File explorer
[Test scenarios and Test cases](https://www.notion.so/cascade8/File-Explorer-08ea83d7e82f437c9b96e22e87d8df7c)

Changes in this release so please double check:
- #6824

RemcoSimonides:
Fix #7057

### DIFF
--- a/libs/media/src/lib/file/upload-widget/upload-widget.component.ts
+++ b/libs/media/src/lib/file/upload-widget/upload-widget.component.ts
@@ -6,9 +6,6 @@ import { AngularFirestore } from '@angular/fire/firestore';
 // Blockframes
 import { BehaviorStore } from '@blockframes/utils/observable-helpers';
 import { slideUp } from '@blockframes/utils/animations/fade';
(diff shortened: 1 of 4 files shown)


libs/media/src/lib/file/upload-widget/upload-widget.component.ts
```diff
@@ -50,35 +39,10 @@ export class UploadWidgetComponent {
...
     const tasks = this.tasks.value;
-    if (removeReference) this.removeReference(tasks[index]);
     tasks.splice(index, 1);
     this.tasks.value = tasks;
   }
 
-  /**
-   * Removes the already-set References in the db
-   */
-  async removeReference(task: AngularFireUploadTask) {
```
RemcoSimonides, line 63: I don't think we need this anymore because the reference is only written to firestore when the upload is completed.
These changes were 1+ year old already so I guess it's still from the old way we handled media.

phpgeekfr (approved):

PR #7107, opened 2021-11-12
clean title list dashboard

GrandSchtroumpf:
Cleanup title list in dashboard & use joinWith instead of queryChange


apps/festival/festival/src/app/dashboard/title/list/list.component.ts
```diff
@@ -61,16 +66,4 @@ export class ListComponent {
   resetFilter() {
     this.filter.reset();
```
yspringfield, line 67: I believe we should have something similar to this.
`this.dynTitle.useDefault();` 
in `resetFilter` as used in 
apps/catalog/catalog/src/app/dashboard/title/list/list.component.ts

yspringfield (approved):
SGTM

PR #7110, opened 2021-11-12
fixed

### LINKED ISSUE #7062: Release 3.5 - Avails
[Test scenarios and Test cases](https://www.notion.so/cascade8/Avails-3e975555a25545fb8d66ae498725680d)

Changes in this release so please double check:
- [x] #6120
- [x] #6619
- [ ] #6620
- [x] #6693
- [x] #6695
- [x] #6702
- [x] #6705
- [ ] #6733
- [x] #6751
- [x] #6990
- [x] #6767
- [x] #6766
- [ ] #6765
- [x] #6936

yspringfield:
#7062


libs/contract/src/lib/avails/avails.ts
```diff
@@ -58,9 +58,9 @@ export function termsCollision(avails: AvailsFilter, terms: BucketTerm<Date>[])
 /** Get all the terms that overlap the avails filter */
 export function collidingTerms<T extends BucketTerm>(avails: AvailsFilter, terms: T[]) {
   return terms.filter(term => (avails.exclusive || term.exclusive)
-    && someOf(avails.territories, 'optional').in(term.territories)
-    && someOf(avails.medias, 'optional').in(term.medias)
-    && someOf(avails.duration, 'optional').in(term.duration)
+    && allOf(avails.territories, 'optional').in(term.territories)
```
yspringfield, line 61: I'll like to know if this is okay here?

libs/utils/src/lib/operators.ts
```diff
@@ -73,13 +75,22 @@ export function joinWith<T, Query extends QueryMap<T>>(queries: Query, options:
...
+        /**
+         * https://stackoverflow.com/a/122704/6441976
+         * We need a more robust deep clone method other than
+         * @code JSON.parse(JSON.stringify(my_var))
+         * Dates gets toIsoString-ified and this causes
+         * avails durationCheck in `allOfAvailInTerms` method
+         * to fail as it's verifying greater than on a date and a
+         * stringified date object ( which is at this point a string )
+         */
+        return cloneDeep(entity) as any
```
yspringfield, line 88: I'll like to know if this is okay here?

apps/catalog/catalog/src/app/dashboard/avails/calendar/calendar.component.html
```diff
@@ -1,4 +1,4 @@
-<form [availsFilter]="availsForm" required hide='duration'></form>
+<form [availsFilter]="availsForm" required hide='duration' fxLayout="row wrap"></form>
```
GrandSchtroumpf, line 1: Can you put it in the css instead ? I tend to prefer in the CSS because fxLayout is using Javascript and it set the value in the `style` attribute, which makes it all mighty over any other css properties.

apps/catalog/catalog/src/app/dashboard/avails/map/map.component.html
```diff
@@ -1,4 +1,4 @@
-<form [availsFilter]="availsForm" required hide='territories'></form>
+<form [availsFilter]="availsForm" required hide='territories' fxLayout="row wrap"></form>
```
GrandSchtroumpf, line 1: same as above, and below.

libs/contract/src/lib/avails/avails.ts
GrandSchtroumpf: @pldespaigne can you review this ? To me it looks good.

libs/utils/src/lib/operators.ts
GrandSchtroumpf: We try not to use lodash in the code as it has (had ?) a bad reputation for treeshaking.
If the issue if only for Date and maybe Timestamp (which should be the case because we don't use any other classes in the data), then you can use the reviver parameter of `JSON.parse`: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#using_the_reviver_parameter

There are many example of code doing that (maybe there is already a piece of code in the code doing it actually). Here is one example: https://gist.github.com/davehax/2f32e7b09c3da3531601e6543fcff82e

libs/contract/src/lib/avails/avails.ts
pldespaigne: I find it strange because `allOf in` is more restrictive than `someOf in`
But if unit tests are green, I would say it's ok

libs/ui/src/lib/map/map.component.ts
```diff
@@ -127,7 +130,8 @@ export class MapComponent implements AfterViewInit, OnDestroy {
       fillOpacity: 0,
       stroke: true,
       weight: 1,
-      color: '#06081c'
+      color: '#06081c',
+      className:this.disableSelect? 'no-cursor': '',
```
GrandSchtroumpf, line 134: If you cannot apply the style directly here. Then change the viewEncapsulation of the component to None and set the class `no-cursor` at the level of the component so the children can take advantage of it.
If you do that remember to wrap all the style in the css with `:host {}` to avoir bleeding style

apps/catalog/catalog/src/styles.scss
```diff
@@ -6,6 +6,10 @@
 // @todo(#1617) Lazy Load this
 @import '~leaflet/dist/leaflet.css';
 
+.no-cursor{
```
GrandSchtroumpf, line 9: There is no reason to create a global class for that. Please remove it and apply it at the right place. See my comment on viewEncapsulation

apps/catalog/catalog/src/app/marketplace/offer/shell/shell.component.html
```diff
@@ -11,7 +11,7 @@ <h1>{{ offer.id }}</h1>
       </div>
       <div *ngIf="offer.contracts as contracts">
         <h2>
-          {{ contracts.length }} Titles
+          {{ contracts.length + (contracts.length>1 ? ' Titles':' Title')}}
```
GrandSchtroumpf, line 14: prefer `{{ contracts.length }} {{ contracts.length ? ' Titles':' Title' }}`

libs/contract/src/lib/avails/calendar/calendar.component.html
```diff
@@ -9,7 +9,7 @@
     <th scope="row"  [ngClass]="{ hover: r === state.hoverRow }">{{row}}</th>
     <td
       *ngFor="let col of columns; index as c"
-      [ngClass]="stateMatrix[r][c]"
+      [ngClass]="stateMatrix[r][c] + (disableSelect ? ' select-disabled' : '')"
```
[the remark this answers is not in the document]
yspringfield: Opted for this `[class.select-disabled]="disableSelect"` as the cursor isn't forcibly pointer should disableSelect be false.

apps/catalog/catalog/src/app/marketplace/offer/shell/shell.component.html
yspringfield: We rather want to show the plural form if and only if there're 2 or more contracts

GrandSchtroumpf (approved):

PR #7120, opened 2021-11-15
Import contract stakeholders

pldespaigne:
copy the parent mandate stakeholders into the sale stakeholders

### DIFF
--- a/libs/import/src/lib/view-extracted-elements/contract/utils.ts
+++ b/libs/import/src/lib/view-extracted-elements/contract/utils.ts
@@ -250,7 +250,17 @@ export async function formatContract(
       const exists = await Promise.all(stakeholders.map(id => getUser({ id }, userService, userCache)));
       const unknownStakeholder = exists.some(e => !e);
       if (unknownStakeholder) throw new UnknownEntityError({ field: 'contract.stakeholders', name: 'Stakeholders' });
-      return [ data.contract.buyerId, data.contract.sellerId, ...stakeholders ];
+      if (data.contract.type === 'mandate') {
+        return [ data.contract.buyerId, data.contract.sellerId, ...stakeholders ];
+      } else {
+        if (data.contract.sellerId === centralOrgId.catalog) { // internal sale
+          // seller ID is archipel, we don't need to add it, as mandate stakeholders will be copied here (copy is done bellow ~line 290)
+          return [ data.contract.buyerId, ...stakeholders ];
+        } else { // external sale
+          // if the sale is external the seller is not archipel (it's the owner org), and the buyer is unknown by definition
+          return [ data.contract.sellerId, ...stakeholders ]
+        }
+      }
     },
   };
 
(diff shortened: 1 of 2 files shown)


libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -250,7 +250,17 @@ export async function formatContract(
       const exists = await Promise.all(stakeholders.map(id => getUser({ id }, userService, userCache)));
       const unknownStakeholder = exists.some(e => !e);
       if (unknownStakeholder) throw new UnknownEntityError({ field: 'contract.stakeholders', name: 'Stakeholders' });
-      return [ data.contract.buyerId, data.contract.sellerId, ...stakeholders ];
+      if (data.contract.type === 'mandate') {
+        return [ data.contract.buyerId, data.contract.sellerId, ...stakeholders ];
+      } else {
+        if (data.contract.sellerId === centralOrgId.catalog) { // internal sale
+          // seller ID is archipel, we don't need to add it, as mandate stakeholders will be copied here (copy is done bellow ~line 290)
```
RemcoSimonides, line 257: referring to line number isn't maintainable. I like it more if you then also mention what you mean like you did at line 294. Like `±line 290 - contract.stakeholders.concat(mandate stakeholders)` or something. Or extract it to a seperate function and refer to that function in the comment.

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -270,20 +280,27 @@ export async function formatContract(
     const isInternalSale = contract.type === 'sale' && contract.sellerId === centralOrgId.catalog;
     if (isInternalSale) {
       if (typeof data.parentTerm === 'number') {
-        contract.parentTermId = contracts[data.parentTerm - 2]?.terms[0]?.id; // excel lines start at 1 and first line is the column names
-        if (!contract.parentTermId) errors.push({
+        const mandate = contracts[data.parentTerm - 2];
+        contract.parentTermId = mandate?.terms[0]?.id; // excel lines start at 1 and first line is the column names
```
RemcoSimonides, line 284: What if the excel file changes. Should this then also be changed? Should be defined in 1 spot or calculated so it's easier to maintain imo

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -270,20 +280,27 @@ export async function formatContract(
...
         contract.parentTermId = data.parentTerm;
+        // moreover the corresponding mandate is already in the contractCache so the look up should be efficient
+        const mandate = await checkParentTerm(contract.parentTermId, contractService, contractCache);
+        contract.stakeholders.concat(mandate.stakeholders);
       }
     }
 
-    // TODO FOR EXTERNAL SALE WE SHOULD COPY THE MANDATE STAKEHOLDERS INTO THE SALE STAKEHOLDERS
+    // remove duplicate from stakeholders
+    contract.stakeholders = contract.stakeholders.filter((s, i) => contract.stakeholders.indexOf(s) == i);
```
RemcoSimonides, line 303: `===` instead of `==`

RemcoSimonides: Also, I don't really see how this filters the duplicates? isn't it removing everything because `indexOf(s)` is always equal to `i`?

pldespaigne: - `filter` remove if condition is false
- `indexOf` return the first index of an element 😉

RemcoSimonides (approved):

PR #7128, opened 2021-11-16
George/refactor - minor hotfix

george43g:
env.ts can get outdated and does now allow devs to make repo wide environment changes without everyone running the `use` command again. When this is merged, every dev should select their environment once more using the backend-ops tool to ensure their config stays up to date


george43g:
there is an issue with this branch having a bad commit history

PR #7129, opened 2021-11-16
hotfix to prevent env.ts from being out of sync

george43g:

### DIFF
--- a/apps/backend-ops/src/select-environment.ts
+++ b/apps/backend-ops/src/select-environment.ts
@@ -1,5 +1,4 @@
 import { join, resolve } from 'path';
-import { copyFileSync } from 'fs';
 import { runShellCommand, getServiceAccountObj } from '@blockframes/firebase-utils';
 import { promises as fsPromises } from 'fs';
 import { execSync } from 'child_process';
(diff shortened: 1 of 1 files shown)


apps/backend-ops/src/select-environment.ts
```diff
@@ -109,10 +109,12 @@ export async function selectEnvironment(projectId: string) {
   output = execSync(cmd).toString();
   console.log(output);
 
+  const fileName = `env.${projectId}.ts`;
+  const envLine = `export * from './${fileName}'`;
   const localEnvFile = join(process.cwd(), 'env', 'env.ts');
-  const targetEnvFile = join(process.cwd(), 'env', `env.${projectId}.ts`);
-  copyFileSync(targetEnvFile, localEnvFile);
```
phpgeekfr, line 114: if `copyFileSync`  not used, remove from import

apps/backend-ops/src/select-environment.ts
```diff
@@ -109,10 +109,12 @@ export async function selectEnvironment(projectId: string) {
...
-  console.log(`env.ts file overwritten with env.${projectId}.ts`);
+  // const targetEnvFile = join(process.cwd(), 'env', `env.${projectId}.ts`); //? Do we check if exists?
```
phpgeekfr, line 115: remove if not useful or add an issue number to the comment

apps/backend-ops/src/select-environment.ts
```diff
@@ -109,10 +109,12 @@ export async function selectEnvironment(projectId: string) {
...
+  await writeFile(localEnvFile, envLine);
+  console.log(`env.ts file points to env.${projectId}.ts`);
```
phpgeekfr, line 117: `points to` makes me think that it is a symbolic link. maybe `now using`  or  `now importing all from`

RemcoSimonides (approved):

Issue #7059, opened 2021-11-08
Release 3.5 - General UI


RemcoSimonides:
[Test scenarios and Test cases](https://www.notion.so/cascade8/General-UI-db403143e36c480d9c6b562cc001119f)

Changed in this release so please double check:
- #4731
- #6175
- #6239
- #6372
- #6483
- #6634
- #6640
- #6832

xmano:
Market home page has some images missing, getting traceback:
![image](https://user-images.githubusercontent.com/8731418/141056937-2d928fca-753e-4afe-a86e-2b7eea6c68fe.png)

And some images for In Post-Production section are also missing:
![image](https://user-images.githubusercontent.com/8731418/141056985-8b2ce234-754b-44dc-a60b-9aaf5154c1c6.png)

RemcoSimonides:
The background image of 'Don't miss anything' is working for me. Which browser are you using?

The movies in 'Post-Production' don't have a poster image yet

RemcoSimonides:
## LANDING PAGE
### BANNER

- [ ] Change text to: "One tool for your year-round B2B promotions & acquisitions"


### THEY'RE USING IT

- [ ] Change text to: Over 1400 buyers are already using Archipel, from companies like:
- [ ] Remove 1 or 2 logos so it fits on 1 line instead of 2 (on a standard computer screen like mike - Macbook 13")


### HOW IT WORKS / Sales agents / User Benefits

- [ ] Correct text to: "Simplify your buyers' lives"


### WHAT OUR CLIENTS SAY

- [ ] Correct name with capital letter: "Nate Bolotin"


### WANT TO LEARN MORE? Hero

- [ ] Correct Email Label and Placeholder to "Email Address" (not Email Adress or E-mail address)
- [ ] Correct Phone Placeholder to: "Phone Number" (with capital letter)
- [ ] Change checkbox text to: "Subscribe to our newsletter"


### FOOTER

- [ ] Add hyperlinks to the partners' logos:
- BPI : https://www.bpifrance.fr/
- Creative Europe MEDIA: https://ec.europa.eu/culture/funding-creative-europe/creative-europe-media-strand
- IFCIC: http://www.ifcic.fr/
- CNC: https://www.cnc.fr/

Jackseed:
### Catalog landing banner
- [ ] Change the title for: "The only marketplace offering multi-seller deals"

Jackseed:
### Screening carousel
[...]

PR #7130, opened 2021-11-16
7059 event banner improvements

RemcoSimonides:
to do in #7059

### DIFF
--- a/apps/festival/festival/src/app/marketplace/home/home.component.html
+++ b/apps/festival/festival/src/app/marketplace/home/home.component.html
@@ -102,7 +102,7 @@
   <section class="events-slider" bgAsset="events_slider_background.png">
     <header fxLayout fxLayoutAlign="space-between center" class="dark-contrast-theme">
       <h2>{{ section.title }}</h2>
-      <a mat-button [routerLink]="section.link || '../organization'">
+      <a mat-button [routerLink]="section.link || '../event'">
         <span>See all</span>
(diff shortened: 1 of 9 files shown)


apps/festival/festival/src/app/marketplace/home/home.component.html
```diff
@@ -102,7 +102,7 @@ <h2>{{ section.title }}</h2>
   <section class="events-slider" bgAsset="events_slider_background.png">
     <header fxLayout fxLayoutAlign="space-between center" class="dark-contrast-theme">
       <h2>{{ section.title }}</h2>
-      <a mat-button [routerLink]="section.link || '../organization'">
+      <a mat-button [routerLink]="section.link || '../event'">
```
phpgeekfr, line 105: does this lead to event list ?

libs/admin/src/lib/cms/home.ts
```diff
@@ -47,6 +47,7 @@ export interface SliderSection extends Section {
 export interface EventsSliderSection extends Section {
   _type: 'eventsSlider';
   title: string;
+  link: string;
```
phpgeekfr, line 50: I don't see where in this PR you use this `link`

phpgeekfr (approved):

apps/festival/festival/src/app/marketplace/home/home.component.html
RemcoSimonides: ye

PR #7136, opened 2021-11-16
small fixes on events

phpgeekfr:


libs/event/src/lib/form/edit-details/edit-details.component.scss
```diff
@@ -4,4 +4,8 @@ mat-icon{
 
 mat-radio-group {
   margin-bottom: 16px;
+}
+
+.noAccessibilityChange p {
```
RemcoSimonides, line 9: Makes sense if the h1 is also centered if you center the paragraph

RemcoSimonides (approved):

PR #7146, opened 2021-11-16
safari banner & watermark

phpgeekfr:


apps/festival/festival/src/app/event/event.component.ts
```diff
@@ -1,9 +1,29 @@
...
+export class EventComponent implements OnInit {
+
+  isMeeting$: Observable<boolean>;
+
+  constructor(
+    private route: ActivatedRoute,
+    private service: EventService,
+  ) { }
+
+  async ngOnInit() {
```
RemcoSimonides, line 22: No need for async

RemcoSimonides: We didn't have the event data in the EventComponent already? ^^' Also no way to just pass it through @Input()?

RemcoSimonides (approved):

apps/festival/festival/src/app/event/event.component.ts
phpgeekfr: > We didn't have the event data in the EventComponent already
No this one only took care of layout

> Also no way to just pass it through @input()?

no, this is the very first component of the event pages

Issue #6830, opened 2021-10-08
get rid of https://e2e.blockframes.io/


phpgeekfr:
get rid of https://e2e.blockframes.io/

phpgeekfr:
I stopped docker container on `Hetzner` server (144.76.172.41) 

I'm also taking some time to stop what is not used on this server so we can stop using it.

Once `check_app_size` is removed from `config.yml`

delete the webhook 
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/7e9877a4-cc17-4181-ac01-a1b73d6ad0e7)

phpgeekfr:
stopped crontab that was saving bdd backups on `Hetzner` server

```
0 2 * * * PATH=/home/blockframes/google-cloud-sdk/bin:$PATH gsutil -m rsync -r gs://blockframes-backups /home/blockframes/firestore-backup/data
```

phpgeekfr:
to remove jenkins docker container, `check_app_size` step will have to be removed on develop, lets do this after the release

phpgeekfr:
We will also be able to remove Zapier task : 
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/56a4ac3e-0383-4e93-aac7-5d80e2d6b5a5)

phpgeekfr:
Zapier task is pushing data to a spreadsheet file ...
@Jackseed do you use this ? 
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/aba3a1cd-d43d-4b12-a45e-bd0a5d3721ee)

phpgeekfr:
crontab activated again. We should check server price and talk about how we store data outside google infrastructure @Jackseed

phpgeekfr:
server price: 39.99€

phpgeekfr:
web hook turned off.
zapier task tuned off
jenkins docker container stoped on Hetzner server

PR #7157, opened 2021-11-17
cleaning check_app_size & e2e.blockframes.io #6830

phpgeekfr:


apps/crm/crm/src/app/crm/pages/dev-area/dev-area.component.html
```diff
@@ -9,11 +9,6 @@ <h4>Don't feel alone anymore...</h4>
         </mat-card-header>
         <mat-card-content>
           <ul>
-            <li>
```
george43g, line 12: sure we want to do this already?

george43g (approved):
would rather comment this out for now but fine

apps/crm/crm/src/app/crm/pages/dev-area/dev-area.component.html
phpgeekfr: yeah, no one is using this. And docker ` e2e.blockframes.io` is already down

Issue #6465, opened 2021-08-17
Landing - add an anchor to top of the page on the logo link


Jackseed:
Currently if you're on the landing, scroll down, then click on the logo nothing happens. Instead it should scroll up to the top of the page.

RemcoSimonides:
@Poppe-A if you type "Fix #1234" in the PR, this issue will close automatically when you merge the PR ;)

PR #7185, opened 2021-11-18
scrollToTop landing page

Poppe-A:
#6465


libs/landing/src/lib/shell/shell.component.scss
```diff
@@ -23,8 +23,11 @@ landing-shell-page {
         margin-right: auto;
       }
 
-      a[app-logo] {
+      button[app-logo] {
         margin-right: 96px;
+        border: none;
+        background-color: transparent;
```
phpgeekfr, line 29: on other CSS sheets `button[app-logo]` use `background: none;` .  if  `background-color: transparent;`  does similar, use `background: none;` instead

phpgeekfr (approved):

Issue #7061, opened 2021-11-08
Release 3.5 - Meeting


RemcoSimonides:
[Test scenarios and Test cases](https://www.notion.so/cascade8/Meeting-7d16f2abef1a4b5b85b5a1889dfa1eab)

Changes in this release so please double check:
- #5219
- #6639

cams-goussale:
hello @phpgeekfr and @antoine :)
This is the review from last week's meeting test with 9 persons. Can some of these points be fixed for the release next week?

## Your environement
_Please describe your environment_ 
- OS: Windows and Mac depending on users
- Browser (& version if possible): Chrome, Safari

## Review item
_Copy & fill this section for each item._
### Localization
- App name: Market
- Page: Meeting page
- Url: https://staging.archipelmarket.com/c/o/marketplace/event/OpchYPkh0m4PK1CkUrte/lobby
- User used (if anonymized): 9 users with 2 sellers, 7 buyers. See [details here](https://docs.google.com/spreadsheets/d/1CenZNjfrP8rnHTQyRwrohYYJnYOnD_XeutntDV95Gmc/edit#gid=0)

**1. Color background**
Context: When there are some users who haven't join the meeting yet, there is a white/transparent rectangle for the missing persons.
- [ ] Fix: Change the color to the background layout which is called "--background" in our builder - to test with Stéphanie

![image](https://user-images.githubusercontent.com/93206363/140716179-7f0a9114-3006-4479-b7c7-28e5809d9075.png)
![image](https://user-images.githubusercontent.com/93206363/140719286-06e248fe-ec1e-4621-ab86-2505e02e74de.png)


**2. Too much black for the large view.**
- [ ] Fix: If possible, have the cameras in format 16:9 to avoid all the black spaces.

![image](https://user-images.githubusercontent.com/93206363/140782796-931110d1-4c24-46bf-ac01-5c66c1c15e41.png)


**3. Watermark to be higher**
- [ ] Fix: Place the watermark higher so it doesn't overstep too much on the content displayed.

![image](https://user-images.githubusercontent.com/93206363/140718771-893ebcfa-625c-4632-97ce-715e6a646dc4.png)


**4. Maximize button on the video conference room**
Context: A user can maximise / expand in full screen the content shared, but they can't with the room's cameras.
- [ ] Fix: If possible, add a maximize button for the video conference section.

[...]

PR #7187, opened 2021-11-18
Clean tooltip on event creation

Poppe-A:
#7061

### DIFF
--- a/libs/event/src/lib/form/edit-details/edit-details.component.html
+++ b/libs/event/src/lib/form/edit-details/edit-details.component.html
@@ -24,15 +24,21 @@
   <div *ngIf="showPrivacy" fxLayout="column" fxLayoutGap="12px">
     <h3>Privacy Settings</h3>
     <mat-radio-group fxLayout="column" formControlName="accessibility" (change)="accessibilityChanged($event)">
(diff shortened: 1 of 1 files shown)


libs/event/src/lib/form/edit-details/edit-details.component.html
```diff
@@ -24,14 +24,16 @@
   <div *ngIf="showPrivacy" fxLayout="column" fxLayoutGap="12px">
     <h3>Privacy Settings</h3>
     <mat-radio-group fxLayout="column" formControlName="accessibility" (change)="accessibilityChanged($event)">
-      <mat-radio-button  color="primary" value="public" matTooltip="Best for meetings">
+      <mat-radio-button  color="primary" value="public">
         PUBLIC - Open to anyone with a link
+        <mat-icon svgIcon="info" matTooltip="Best for meetings"></mat-icon>
```
phpgeekfr, line 29: I think I tried this during release fixes and icons where not well aligned with text, did you check that ?

Poppe-A: Yes, it's a bit too high, but it looks like an indication, it didn't seemed weird to me. I'll try to fix it

phpgeekfr (approved):

PR #7191, opened 2021-11-19
Mano import titles e2e

xmano:
Update the test and make import title to pass


xmano:
Import title passing locally and on emulator

image

phpgeekfr (approved):
so movie import is working but not contract yet ? 
What is the next step to make contract import work ?

xmano:
> so movie import is working but not contract yet ? What is the next step to make contract import work ?

Contract is next on the pipeline.
- Review the template and correct it
- But we need to have a Cypress Command to get the movies accepted - some development needed

Issue #7209, opened 2021-11-22
Simplify mailchimp tags for newsletter subscriber


Poppe-A:
remove useles tags, the last one must looks like this: 
landing - appName (festival/catalog/financiers)

we also need to change the user satus (unsubscribed to subscribed)

PR #7212, opened 2021-11-22
correct mailchimp tags

Poppe-A:
#7209

### DIFF
--- a/apps/backend-functions/src/mailchimp.ts
+++ b/apps/backend-functions/src/mailchimp.ts
@@ -22,6 +22,6 @@ export const registerToNewsletters = async (data: {email: string, tags: Mailchim
   return await mailchimp.lists.addListMember(mailchimpListId, 
     {email_address, 
     tags,
-    status: "unsubscribed"}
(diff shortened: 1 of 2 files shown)


apps/backend-functions/src/main.ts
```diff
@@ -80,7 +80,7 @@ export const getEventAnalytics = functions().https.onCall(skipInMaintenance(logE
 /** Trigger: REST call bigQuery to fetch analytics active users */
 export const getAnalyticsActiveUsers = functions().https.onCall(skipInMaintenance(logErrors(bigQuery.getAnalyticsActiveUsers)));
 
-export const registerToNewsletter = functions().https.onCall(skipInMaintenance(logErrors(mailchimp.registerToNewsletters)));
+export const registerToNewsletters = functions().https.onCall(skipInMaintenance(logErrors(mailchimp.registerToNewsletters)));
```
RemcoSimonides, line 83: No need to change the name. This CORS error is only on your env.

RemcoSimonides (approved):

PR #7222, opened 2021-11-23
db-cleaning script - minimal update #6460

phpgeekfr:


apps/backend-ops/src/db-cleaning.spec.ts
```diff
@@ -89,26 +93,29 @@ describe('DB cleaning script', () => {
...
     ];
 
     // Add a new auth user that is not on db
-    const fakeAuthUser = { uid: 'D', email: 'D@fake.com' };
-
+    const fakeAuthUser = { uid: 'D', email: 'D@fake.com', providerData: [{ uid: 'D', email: 'D@fake.com', providerId: 'password' }] };
     adminAuth.users.push(fakeAuthUser);
 
+    // Add an anonymous user that should not be deleted even if not on DB
+    const anonymousUser = { uid: 'anonymous', providerData: [] };
```
RemcoSimonides, line 106: No need to add providerData? `{ uid: 'anonymous', providerId: 'anonymous' }`

apps/backend-ops/src/users.ts
```diff
@@ -62,7 +62,8 @@ export async function removeUnexpectedUsers(expectedUsers: UserConfig[], auth: A
   do {
     const result = await auth.listUsers(1000, pageToken);
 
-    const users = result.users;
+    // Anonymous users filtering
+    const users = result.users.filter(u => u.providerData.length !== 0);
```
RemcoSimonides, line 66: and if you provide providerId in earlier comment, then filter on that instead of array length

RemcoSimonides (approved):

apps/backend-ops/src/db-cleaning.spec.ts
phpgeekfr: no, when user is logged in as anonymous, `providerData` is an empty array

apps/backend-ops/src/users.ts
phpgeekfr: No, we need to test what a real anonymous user will have in `providerData` array. If anonymous , this array is empty 
see https://stackoverflow.com/questions/39640574/how-to-bulk-delete-firebase-anonymous-users

Issue #6985, opened 2021-10-28
Negociation data model


GrandSchtroumpf:
`negociations` is a **subcollection of the contract**. It's used to keep track of the last state of the negotiation between the stakeholders.

We want to keep only the field that can be changed and the field required for the query & rules

## Model
`libs\contract\src\lib\negociation\+state\negociation.firestore.ts`
- [x] Create the interface
```typescript
// We extends the BucketContract with some information for rules
interface Negociation extends BucketContract {
  _meta: Meta;
  id: string;
  /** Id of the org that created the negociation */
  createdByOrg: string;
  /** list of companies involved by this contract */
  stakeholders: string[];
  /** Always Archipel Content in our case */
  sellerId: string;
  /** The org that want to buy the title */
  buyerId: string;
}
```

- [x] Also we need to add the status `negociating` in the `ContractStatus` list : 
`libs\contract\src\lib\contract\+state\contract.firestore.ts`

## Service
We use `akita-ng-fire` for the service which still implies some boilerplate.
- [x] Create the service
`libs\contract\src\lib\negociation\+state\negociation.service.ts`
```typescript
export interface NegociationStateextends EntityState<Negociation, string>, ActiveState<string> {}

@Injectable({ providedIn: 'root' })
@CollectionConfig({ path: 'contracts/:contractId/negociations' })
export class NegociationService extends CollectionService<NegociationState> {
  useMemorization=true;
  constructor(store: NegociationStore ) {
    super(store)
  }
  
  formatFromFirestore(negociation: Negociation<Timestamp>) {
    // Transform Timestamp into Date here
  }
}

// BOILETPLATE
@Injectable({ providedIn: 'root' })
@StoreConfig({ name: 'negociation' })
export class NegociationStore extends EntityStore<NegociationState> {
  constructor() {
    super();
  }
}
```

## Indexes
We want to order negociations by `_meta.createdAt` in `desc` mode, so we need an index for it in `firestore.indexes.json`.
- [x] Create the indexe

## Rules : 
- [x] `create`: if user's orgId is same as `createdByOrg`
- [x] `update`: if field changed is **only** `status` && user's orgId is **not** the `createByOrg` && orgId is in `stakeholders` list
- [x] `read`: if user's orgId is in `stakeholders` list

## Contract Service
`libs\contract\src\lib\contract\+state\contract.service.ts`
- [x] Update the ContractService to get the last negociation of a specific contract :
```typescript
[...]

PR #7234, opened 2021-11-24
feat:added negotiation data model

yspringfield:
Fixed: #6985


firestore.rules
```diff
@@ -371,6 +371,25 @@ service cloud.firestore {
                     && isOnlyChangingContractStatus()
                     && notInMaintenance();
       allow delete: if false;
+
+      ////////////////////////////////
+      // NEGOTIATION RULES            //
+      ////////////////////////////////
+
+      function isUpdatingOnlyStatus(){
```
GrandSchtroumpf, line 379: would be nice to have a global function 
```
function onlyChange(fields) {
  return incomingData().diff(existingData()).affectedKeys().hasOnly(fields);
}
```

PR #7237, opened 2021-11-24
fix onMovieUpdateEvent error

RemcoSimonides:
all video deleted resulted in 'crash' in movieUpdateEvent error
Also, it was trying to move files unnecessarily


libs/firebase-utils/src/lib/node-request.ts
```diff
@@ -15,11 +15,12 @@ export function sendRequest<T = unknown>(options: RequestOptions, data?: unknown
 		const req = request(options.method === 'POST' ? postOptions : options, res => {
       let body = '';
       res.on('data', chunk => body += chunk);
-      res.on('end', () => resolve(JSON.parse(body) as T));
+      res.on('end', () => resolve(body ? JSON.parse(body) : undefined as T));
```
RemcoSimonides, line 18: `as T` isn't needed I just noticed so I'm removing that one

pldespaigne (approved):

Issue #7004, opened 2021-10-29
Mail when an user requests to attend event is sent to org admins but not event creator


phpgeekfr:
If event creator is not org admin and another user makes a request to attend this event, event creator will get a notification but no email.

phpgeekfr:
looks like there is the same issue for `onInvitationToAnEventAcceptedOrRejected`

phpgeekfr:
Cleaning:
In `apps/backend-functions/src/internals/invitations/events.ts`
```
   const adminIds = await **getAdminIds(invitation.toOrg.id);**
    const admins = await Promise.all(adminIds.map(i => getUser(i)));
    admins.forEach(a => recipients.push(a.uid));
```

Looks like we can do : 
```
    const adminIds = await getAdminIds(invitation.toOrg.id);
    adminIds.forEach(uid => recipients.push(uid));
```

PR #7238, opened 2021-11-24
Send mail to user organizer on request to attend

Poppe-A:
#7004

### DIFF
--- a/libs/event/src/lib/form/create/create.component.ts
+++ b/libs/event/src/lib/form/create/create.component.ts
@@ -3,8 +3,9 @@ import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
 import { AuthService } from '@blockframes/auth/+state';
 import { AccessibilityTypes } from '@blockframes/utils/static-model/types';
 import { OrganizationQuery } from '@blockframes/organization/+state';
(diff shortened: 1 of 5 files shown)


libs/event/src/lib/form/create/create.component.ts
```diff
@@ -28,11 +29,9 @@ export class EventCreateComponent {
   }
 
   async createAndRedirect() {
-    const event = this.form.value;
+    const event = this.form.value as Event<Meeting | Screening>;
     event.ownerOrgId = this.orgQuery.getActiveId();
-    if (isMeeting(event)) {
-      event.meta.organizerUid = (await this.authService.user).uid;
-    }
+    event.meta.organizerUid = (await this.authService.user).uid;
```
phpgeekfr, line 34: I don't know which one is better: 
```
const event = this.form.value as Event<Meeting | Screening>;
```

or 

```
if (isMeeting(event) || isScrening(event)) {
```

phpgeekfr (approved):

libs/event/src/lib/form/create/create.component.ts
Poppe-A: I think the first one is makes more sens. It's just about formating the event, while the second one is a test between two things which will fullfil the condition

PR #7245, opened 2021-11-25
6872 added on hold org status

### LINKED ISSUE #6872: Add a new org status "On hold"
We should be able to differentiate current `pending` orgs between the ones that hasn't been reviewed and the others that were reviewed but not accepted.
We won't refuse them for business reasons but instead we should be able to put it "On hold". This new status is exactly the same as `pending` it's just a way to split this category in 2.

RemcoSimonides:
Fix #6872


phpgeekfr (approved):
LGTM

Issue #7160, opened 2021-11-17
Roles and Identity pages for events as anonymous


phpgeekfr:
when joining an event without log in or sign up (from an invitation by email) :

- [x] This part is way too big (the surface needs to be 80% of the container size , which is here the button)
- [x] We need the logo on this page to be in the layout (not on position fixed) so that elements can align properly.
- [x] Buttons should be 215px width and height and border-radius: 100%;
- [x] Stephanie should provide images without the circle, so we can have to rounded button behind react on hover.
<img width="900" alt="Capture d’écran 2021-11-15 à 16 58 09" src="https://user-images.githubusercontent.com/66258851/141813978-e3e65592-0ccb-4ad5-870d-8fd8067889d7.png">

<img width="663" alt="Capture d’écran 2021-11-15 à 17 03 42" src="https://user-images.githubusercontent.com/66258851/141814319-17576348-6706-4202-9c99-fa5c7f636d85.png">

- [x] "organizer" and "guest" should be outside of the button circle

_Originally posted by @RemcoSimonides in https://github.com/blockframes/blockframes/issues/7060#issuecomment-970053950_

- [x]   Events pages are accessible for connected users that does not have access to marketplace : maybe add the "hasAccess" guards that redirect to "Request access" page. moved to => #7266
- [x]  Prevent two user (or more) to connect with the same email to a "invitation-only" event ? moved to => #7254

_Originally posted by @phpgeekfr in https://github.com/blockframes/blockframes/issues/7061#issuecomment-965184688_

### LOBBY PAGE
![image](https://user-images.githubusercontent.com/55785823/142042895-2ea17de0-060d-40e3-837c-cd57803ad773.png)
- [x] Change text to "Please fill in the following information:"
- [x] Correct with capitals: "First Name" / "Last Name" / "Email Address" / "Join Event"
- [x] Correct text to "Log in with existing account"

https://user-images.githubusercontent.com/55785823/142189511-994d95a6-2b80-4a6f-8177-eaca6e81ceb5.mov
- I have a problem logging in with the error message "Missing invitation parameter" but I double checked, I was invited
- Also the behavior is not okay here, it should not take me back to the landing page, it should keep me on the "lobby page"

- [ ] This is minor but I can't go back to the previous page with my browser's "back" button, I have to click on the one in the page. I think we should be able to do both
https://user-images.githubusercontent.com/55785823/142209720-5d0ee005-c404-4ebc-a014-f640fe5f4827.mov

[...]

PR #7260, opened 2021-11-26
redirect to request-access if user has no marketplace access

Poppe-A:
#7160


phpgeekfr:
given this new issue https://github.com/blockframes/blockframes/issues/7266
we can close this PR

PR #7299, opened 2021-11-30
Feat 7045 pdf

phpgeekfr:


libs/media/src/lib/pipes/storageFile.pipe.ts
```diff
@@ -5,19 +5,23 @@ import { CollectionHoldingFile, FileLabel, getFileMetadata } from '../+state/sta
...
 import { Observable } from 'rxjs';
 
+export function toStorageFile(value: StorageFile | string, collection: CollectionHoldingFile, label: FileLabel, docId: string) {
+  if (typeof value === "string") {
+    const metadata = getFileMetadata(collection, label, docId)
+    return createStorageFile({
+      storagePath: value,
+      ...metadata
+    })
+  } else return value;
```
GrandSchtroumpf, line 15:
```suggestion
  if (typeof value !== "string") return value;
  const metadata = getFileMetadata(collection, label, docId)
  return createStorageFile({
    storagePath: value,
    ...metadata
  })
```

libs/utils/src/lib/pdf.service.ts
```diff
@@ -0,0 +1,53 @@
...
+        return {
+          id: m.objectID,
+          posterUrl
+        };
+      }),
+      app
+    };
+
+    // @dev if using emulator, use `http://localhost:5001/${projectId}/${firebaseRegion}/createPdf`
+    const url = `https://${firebaseRegion}-${projectId}.cloudfunctions.net/createPdf`;
```
GrandSchtroumpf, line 34: The code should work the same in both cases : 
```typescript
const url = useEmulators
  ? `http://localhost:5001/${projectId}/${firebaseRegion}/createPdf`
  : `https://${firebaseRegion}-${projectId}.cloudfunctions.net/createPdf`;
```

libs/utils/src/lib/pdf.service.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    await new Promise(resolve => {
+      this.http.post(url, params, { responseType: 'arraybuffer' })
```
GrandSchtroumpf, line 37: use `fetch` if you want to use a promise; 
```typescript
fetch(`${url}?${params}`).then(res => res.blob()).then(...)
```

libs/utils/src/lib/static-model/static-model.ts
```diff
@@ -288,9 +288,9 @@ export const movieFormatQuality = {
 
 export const movieLanguageTypes = {
   original: 'Original',
-  dubbed: 'Dubbings',
-  subtitle: 'Subtitles',
-  caption: 'Closed Captions',
+  dubbed: 'subs',
```
GrandSchtroumpf, line 291: This change will be on every form. Is that ok ?

libs/utils/src/lib/utils.ts
```diff
@@ -214,3 +215,31 @@ export function getWatermark(email: string = '', firstName: string = '', lastNam
 
   return `data:image/svg+xml;utf8,${encodeURIComponent(svg)}`;
 }
+
+export function toLabel(value: string | string[], scope: Scope, joinWith? : string, endWith?: string): string {
```
GrandSchtroumpf, line 219: I thought we already had a function `toLabel`...

apps/backend-functions/src/createPdf.ts
```diff
@@ -0,0 +1,189 @@
...
+  const args = ['--no-sandbox', '--disable-setuid-sandbox'];
+  const browser = await puppeteer.launch({ args: args });
+  const page = await browser.newPage();
+
+  // Wait for the page to load completely
+  await page.setContent(html, { waitUntil: 'networkidle0' });
+  await page.evaluateHandle('document.fonts.ready');
+
+
+  const cssHeader = [];
```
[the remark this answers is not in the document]
phpgeekfr: Headers and footers are added by `puppeteer` as separate html files. So the css in the html cannot be used

libs/utils/src/lib/static-model/static-model.ts
phpgeekfr: I need to confirm with @2amarkina but yes, this is the wording that professionals uses and understands

2amarkina: It's ok for business team to use the following short versions everywhere:

libs/utils/src/lib/utils.ts
phpgeekfr: This is the one and only. I moved it so I can import it in backend without importing angular related stuff

GrandSchtroumpf (approved):

PR #7300, opened 2021-12-01
Negotiation form refactor

yspringfield:
Fixed: negotiation form now respects bucket contract structure


GrandSchtroumpf (commented):
You need to update the crm contract form also to match the new structure of the NegotiationForm

PR #7302, opened 2021-12-02
7242 search with a space

### LINKED ISSUE #7242: Searching for movie with a space in title causes traceback and query seems to hang
Steps to recreate:

1. Go to Library
2. Search for a Movie title with a space  (I used Movie 1)
3. At this time, if you try reload the page, the search box gets filled with Date string : "Mon Jan 01 2001 00:00:00 GMT+0530 (India Standard Time)" and query seems to hang.

image

The traceback is here:

this.query.value.split is not a function
    at MovieSearchForm.push.537.MovieSearchForm.search (search.form.ts:214)
    at ListComponent.<anonymous> (list.component.ts:131)
    at step (apps_catalog_catalog_src_app_marketplace_title_list_list_module_ts.js:83)
    at Object.next (apps_catalog_catalog_src_app_marketplace_title_list_list_module_ts.js:64)
    at apps_catalog_catalog_src_app_marketplace_title_list_list_module_ts.js:58
    at new ZoneAwarePromise (zone.js:1387)
    at push.2704.__awaiter (apps_catalog_catalog_src_app_marketplace_title_list_list_module_ts.js:54)
    at SwitchMapSubscriber.project (list.component.ts:131)
    at SwitchMapSubscriber._next (switchMap.js:30)
    at SwitchMapSubscriber.next (Subscriber.js:49)

However the bigger issue would be slow refresh of the titles - this might cause E2E to timeout.

RemcoSimonides:
Fix #7242

### DIFF
--- a/libs/utils/src/lib/form/form-state-url-encoder.ts
+++ b/libs/utils/src/lib/form/form-state-url-encoder.ts
@@ -1,7 +1,16 @@
 import { ActivatedRoute, Router } from "@angular/router"
 
 function dateReviver(key: string, value: string | number | boolean) {
-  if (typeof value === 'string') {
+
+  /**
+   * The internet doesn't have a waterproof way to check if a string is a date
+   * This method doesn't work for string + number like 'Movie 1'
+   * https://stackoverflow.com/questions/7445328/check-if-a-string-is-a-date-value/30870755#30870755
+   * 
+   * To minimize side effects; only revive date for fields which support date value
+   */
+  const dateKeys = ['from', 'to'];
+  if (dateKeys.includes(key) && typeof value === 'string') {
     // This could be an Invalid date
     const date = new Date(value);
     // check if the getTime is NaN. If not, it's a valid date.
--- a/libs/utils/src/lib/form/form-state-url-encoder.ts
+++ b/libs/utils/src/lib/form/form-state-url-encoder.ts
@@ -1,23 +1,5 @@
 import { ActivatedRoute, Router } from "@angular/router"
 
-function dateReviver(key: string, value: string | number | boolean) {
-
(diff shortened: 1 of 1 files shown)


libs/utils/src/lib/form/form-state-url-encoder.ts
```diff
@@ -1,7 +1,16 @@
...
+
+  /**
+   * The internet doesn't have a waterproof way to check if a string is a date
+   * This method doesn't work for string + number like 'Movie 1'
+   * https://stackoverflow.com/questions/7445328/check-if-a-string-is-a-date-value/30870755#30870755
+   * 
+   * To minimize side effects; only revive date for fields which support date value
+   */
+  const dateKeys = ['from', 'to'];
+  if (dateKeys.includes(key) && typeof value === 'string') {
```
GrandSchtroumpf, line 13: A much cleaner way would be to use the same format for replacing Date in stringify & parse : 
- Stringify: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter
- parse: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#using_the_reviver_parameter

GrandSchtroumpf (changes requested):

RemcoSimonides:
After doing some more tests, I found out the `dateReviver` function still returned the date as a string and all the filters do get their pre-filled data and everything is working.

`dateReviver` function caused some bugs if it recognised a string as a date even if it wasn't a date (e.g. 'movie 1')

GrandSchtroumpf (approved):

PR #7309, opened 2021-12-02
fix contract-form in crm

GrandSchtroumpf:


apps/crm/crm/src/app/crm/pages/offers/shell/contract-form/contract-form.component.ts
```diff
@@ -1,24 +1,19 @@
-import { Component, ChangeDetectionStrategy, OnInit } from '@angular/core';
+import { Component, ChangeDetectionStrategy, OnInit, ChangeDetectorRef } from '@angular/core';
```
yspringfield, line 1: ChangeDetectionRef seems not to be used.

yspringfield (approved):

PR #7316, opened 2021-12-03
Implement unit-tests described in issue 4833

### LINKED ISSUE #4833: Unit test for query movies list
Create a unit test to make this query :
```
return this.movieService.valueChanges(ref => ref
          .where('orgIds', 'array-contains', org.id)
          .where('storeConfig.status', '==', 'accepted')
          .where('storeConfig.appAccess.financiers', '==', true)
          .orderBy('_meta.createdAt', 'desc')
          )
````

Where org.id is not user's org. (CF `apps/financiers/financiers/src/app/marketplace/organization/title/title.component.ts`)

And another with (CF `fromOrg(`)
```
return this.movieService.valueChanges(ref => ref
          .where('orgIds', 'array-contains', org.id)
          )

````
Where org.id is the one user is connect with and with movies that are `accepted` or `draft`

### PR DESCRIPTION
**Pull request by xmano**
Closes #4833

### CODE REVIEW

File: apps/rules/src/tests/firestore/fixtures/data.ts [hunk 1]
```diff
@@ -208,7 +208,7 @@ export const testFixture = {
         acceptedAt: null
       },
       financiers: {
-        status: 'draft',
+        status: 'accepted',
```
phpgeekfr, line 211: is some tests :
- User with admin role
- With User not belonging to any org

movies MI-0d7 is  described as a draftMovie (`draftMovieId`) please rename this var that can be confusing

apps/rules/src/tests/firestore/movies.spec.ts
```diff
@@ -53,6 +53,41 @@ describe('Movies Rules Tests', () => {
...
       })
+
+      test('should be able to fetch movie collection for current org', async () => {
+        const orgId = 'O001';
+        const testMovieTitle = 'movies/MI-078';
+
+        const movieRef = db.collection('movies').where('orgIds', 'array-contains', orgId);
+        const moviesSnap  = await movieRef.get();
+        const movieDocs = moviesSnap.docs;
+        let movies = [];
```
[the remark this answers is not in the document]
xmano: done

apps/rules/src/tests/firestore/movies.spec.ts
```diff
@@ -53,6 +53,41 @@ describe('Movies Rules Tests', () => {
...
+        const orgId = 'O001';
+        const testMovieTitle = 'movies/MI-0d7';
+
+        const movieRef = db.collection('movies')
+                           .where('orgIds', 'array-contains', orgId)
+                           .where('app.financiers.status', '==', 'accepted')
+                           .where('app.financiers.access', '==', true);
+        const moviesSnap  = await movieRef.get();
+        const movieDocs = moviesSnap.docs;
+        let movies = [];
```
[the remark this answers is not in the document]
xmano: done

apps/rules/src/tests/firestore/fixtures/data.ts
xmano: Will rethink this and see if it conflicts the other tests. Get back to you on this.

phpgeekfr: From what I saw, it only needs to be renamed in 

-    User with admin role
-   With User not belonging to any org

apps/rules/src/tests/firestore/movies.spec.ts
```diff
@@ -53,6 +53,41 @@ describe('Movies Rules Tests', () => {
         const movieRef = db.doc(`movies/${existMovieAccepted}`);
         await assertSucceeds(movieRef.get());
       })
+
+      test('should be able to fetch movie collection for current org', async () => {
```
phpgeekfr, line 57: add a new test that should fail if we are asking for movies of a given org if current user is not orgMember

apps/rules/src/tests/firestore/movies.spec.ts
```diff
@@ -53,6 +53,41 @@ describe('Movies Rules Tests', () => {
...
+        let movies = [];
+        movieDocs.forEach(doc => movies.push(doc.data()));
+        expect(movies.length).toEqual(3);
+
+        // testMovieTitle doesn't belong to orgId, shouldn't be fetched
+        const unexpectedMovie = [testFixture[testMovieTitle]];
+        expect(movies).not.toEqual(expect.arrayContaining(unexpectedMovie));
+      })
+
+      test('should be able to fetch movie collection for query condition', async () => {
```
phpgeekfr, line 73: add a new test without specifying `.where('app.financiers.status', '==', 'accepted')` that should fail

phpgeekfr: you are testing a query where connected user org is  'O001 .. so any query should work .. change the user you are connected with for this test for an user that does not belong to   'O001 ..

phpgeekfr (changes requested):

apps/rules/src/tests/firestore/fixtures/data.ts
xmano: ok will rename

apps/rules/src/tests/firestore/movies.spec.ts
xmano: This is covered in the scenario:
`test('should be able to fetch movie collection for current org)'

Here the movie from org 'O003' is not present in the list. So I think it covers the case you mentioned.

Also we have positive confirmation that movies only belonging to the org will be listed here.

phpgeekfr: This does not cover it at all.
The query  `const movieRef = db.collection('movies').where('orgIds', 'array-contains', orgId);`  is asking form movies of orgId O001 how can movies from org 'O003' can be present here ? 

The goal of this test is to check if orgMembers of a given org can fetch movies they own ( despite of the movie status) and that members of others orgs can not

phpgeekfr: so add a new test that should fail if we are asking for movies of a given org if current user is not orgMember

apps/rules/src/tests/firestore/movies.spec.ts
phpgeekfr: I just tried to remove `.where('app.financiers.status', '==', 'accepted')` from query and this still work. So this test is not testing what we want: we want users to be able to fetch only accepted movies of an org they are not member of

apps/rules/src/tests/firestore/movies.spec.ts
```diff
@@ -53,6 +53,41 @@ describe('Movies Rules Tests', () => {
...
+        const movieRef = db.collection('movies').where('orgIds', 'array-contains', orgId);
+        const moviesSnap  = await movieRef.get();
+        const movieDocs = moviesSnap.docs;
+        let movies = [];
+        movieDocs.forEach(doc => movies.push(doc.data()));
+        expect(movies.length).toEqual(3);
+
+        // testMovieTitle doesn't belong to orgId, shouldn't be fetched
+        const unexpectedMovie = [testFixture[testMovieTitle]];
+        expect(movies).not.toEqual(expect.arrayContaining(unexpectedMovie));
```
phpgeekfr, line 70: remove this. as this is not testing what we want and may be confusing

apps/rules/src/tests/firestore/movies.spec.ts
```diff
@@ -53,6 +53,41 @@ describe('Movies Rules Tests', () => {
...
+                           .where('app.financiers.access', '==', true);
+        const moviesSnap  = await movieRef.get();
+        const movieDocs = moviesSnap.docs;
+        let movies = [];
+        movieDocs.forEach(doc => movies.push(doc.data()));
+        expect(movies.length).toEqual(1);
+
+        // testMovieTitle matches query criteria, should be returned
+        const expectedMovie = [testFixture[testMovieTitle]];
+        expect(movies).toEqual(expect.arrayContaining(expectedMovie));
```
phpgeekfr, line 89: remove this. as this is not testing what we want and may be confusing

apps/rules/src/tests/firestore/movies.spec.ts
xmano: Created new test for addressing this and merged your code in right place.

apps/rules/src/tests/firestore/fixtures/data.ts
```diff
@@ -244,7 +244,34 @@ export const testFixture = {
...
+      financiers: {
+        status: 'accepted',
+        access: false,
+        refusedAt: null,
+        acceptedAt: null
+      }
+    },
+    note: '',
+    orgIds: ['O003']
+  },
```
phpgeekfr, line 274: this is not used

xmano: removed

PR #7318, opened 2021-12-03
6235 links in notification

### LINKED ISSUE #6235: Correlation between notifications and Event Page
**_Priority: medium / high_**

Users were confused seeing “screening is about to start” and have a CTA disabled. 

Possible solution: Need to precise exact time when a screening starts or say screening starts in XX minutes. 
Think how exactly it can be shown, if user is able to see the exact time when a message was received.

### LINKED ISSUE #7424: Intercom/cookies issue on Sign Up
When receiving an Invitation to Sign up from an Org, open the Sign up form, the Intercom disappears. Also the "?" next to Invitation Pass leads to nothing.

_Originally posted by @cams-goussale in https://github.com/blockframes/blockframes/issues/7345#issuecomment-989758988_

Remco:
For this to happen I had to disable the cookies (or only the cookie for Intercom). Then none of the '?' buttons in the app will work because Intercom is blocked by user settings.

However, I noticed that Intercom is always launched (despite disabled cookies) on the landing page. I couldn't find anything in our code that can trigger this. @Jackseed do you know if there is a setting on Intercom website that cause this? I'm thinking about a welcome message. If yes, then probably we should trigger this welcome message in our code so we can check if the user accepted the cookies first.

RemcoSimonides:
Fix #6235
Fix #7424

### DIFF
--- a/libs/notification/src/lib/+state/notification.model.ts
+++ b/libs/notification/src/lib/+state/notification.model.ts
@@ -2,7 +2,7 @@ import { StorageFile } from "@blockframes/media/+state/media.firestore";
 import { NotificationBase } from "./notification.firestore";
 
 export interface Notification extends NotificationBase<Date> {
-  message: { text: string, link?: string }[];
+  message: string;
   imgRef?: StorageFile;
   placeholderUrl?: string;
   url?: string;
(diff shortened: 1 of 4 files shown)


GrandSchtroumpf (changes requested):
First we need to understand the problem, then we can think on how we can improve the notification system with a long term solution.
In the meantime, a much better solution would be to put an HTML string in the message and parse it in the frontend with `innerHTML`

RemcoSimonides:
routerLink can't be used in innerHTML and href links doesn't navigate within the app but within the browser instead. Any suggestion on how to solve this?

PR #7353, opened 2021-12-07
update

phpgeekfr:


apps/rules/src/tests/firestore/movies.spec.ts
```diff
@@ -161,6 +137,45 @@ describe('Movies Rules Tests', () => {
     });
   });
 
+  describe('With User not in org', () => {
```
xmano, line 140: we already have a describe block for these tests : User not belonging to organization in movies.orgIds
around L211 - better to have the tests there

xmano (approved):
You could commit these changes into branch and I can take care of organisation if you want.

PR #7363, opened 2021-12-08
feat(cloud functions):function to sync negotiation and sales status

yspringfield:
added cloud function to sync [negotiation status]/ declineReason with that of the concerned sale.

### DIFF
--- a/apps/backend-functions/src/negotiation.ts
+++ b/apps/backend-functions/src/negotiation.ts
@@ -1,0 +1,38 @@
+import { db } from './internals/firebase';
+import { Sale } from '@blockframes/contract/contract/+state/contract.model';
+import { Change, EventContext } from 'firebase-functions';
+import { Negotiation } from '@blockframes/contract/negotiation/+state/negotiation.firestore';
+
+
+export async function onNegotiationUpdate(
+  change: Change<FirebaseFirestore.DocumentSnapshot>, context: EventContext
+) {
+
+  const before = change.before;
+  const after = change.after;
+  const { contractId } = context.params
+
+  if (!before || !after) {
+    throw new Error('Parameter "change" not found');
+  }
+
+  const negotiationBefore = before.data() as Negotiation;
+  const negotiationAfter = after.data() as Negotiation;
+
+  const isStatusUnchanged = negotiationBefore.status === negotiationAfter.status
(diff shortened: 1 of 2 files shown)


apps/backend-functions/src/negotiation.ts
```diff
@@ -0,0 +1,38 @@
...
+  const before = change.before;
+  const after = change.after;
+  const { contractId } = context.params
+
+  if (!before || !after) {
+    throw new Error('Parameter "change" not found');
+  }
+
+  const negotiationBefore = before.data() as Negotiation;
+  const negotiationAfter = after.data() as Negotiation;
```
GrandSchtroumpf, line 20:
```suggestion
  const { contractId } = context.params;
  const before = change.before?.data()  as Negotiation;
  const after = change.after?.data() as Negotiation;
  if (!before || !after) {
    throw new Error('Parameter "change" not found');
  }
```

apps/backend-functions/src/negotiation.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const isStatusUnchanged = negotiationBefore.status === negotiationAfter.status
+  const isDeclined = negotiationAfter.status === 'declined'
+
+  if (isStatusUnchanged) return;
```
GrandSchtroumpf, line 25:
```suggestion
  const hasStatusChanged = before.status !== after.status;
  if (!hasStatusChanged) return;
```

apps/backend-functions/src/negotiation.ts
```diff
@@ -0,0 +1,38 @@
...
+  const saleSnap = await saleRef.get();
+  const sale = saleSnap.data() as Sale<Date>;
+
+  if (sale.status !== 'pending' && sale.status !== 'negotiating') return
+
+  const updates: Partial<Negotiation> = { status: negotiationAfter.status };
+  if (isDeclined)
+    updates.declineReason = negotiationAfter.declineReason;
+
+  saleRef.update(updates);
```
GrandSchtroumpf, line 37: For me you don't need to check the status of the contract: 
```suggestion
  const status = after.status;
  const update = status === 'declined' ? { status, declineReason: after.declineReason } : { status };
  return db.doc(`contracts/${contractId}`).update(update);
```

GrandSchtroumpf (approved):

PR #7365, opened 2021-12-08
Fix for release 3 6

phpgeekfr:


apps/crm/crm/src/app/crm/pages/event/event.component.ts
```diff
@@ -46,6 +50,25 @@ export class EventComponent implements OnInit {
         this.snackBar.open('No title id defined for this screening event..', 'close', { duration: 5000 });
       }
     }
+
+    this.invitations$ = this.invitationService.valueChanges(ref => ref.where('type', '==', 'attendEvent').where('eventId', '==', this.eventId))
+      .pipe(switchMap(async (invitations: InvitationDetailed[]) => {
+        const orgIds = Array.from(new Set(invitations
+          .map(i => getHost(i, 'org').id)
+          .concat(invitations.map(i => getGuest(i, 'user').orgId)).filter(id => id))
+        );
```
RemcoSimonides, line 59: Break it up in multiple lines because now it's quite hard to read

apps/crm/crm/src/app/crm/pages/event/event.component.ts
```diff
@@ -46,6 +50,25 @@ export class EventComponent implements OnInit {
...
+        const orgsPromises = orgIds.map(id => this.orgService.getValue(id));
+        const orgs = await Promise.all(orgsPromises);
+
+        for (const invitation of invitations) {
+          invitation.org = orgs.find(org => org.id === getHost(invitation, 'org').id);
+          invitation.guestOrg = orgs.find(org => org.id === getGuest(invitation, 'user').orgId);
+          invitation.event = this.event;
+          invitation.movie = this.movie;
+        }
+        return invitations;
```
RemcoSimonides, line 69: Do you need `switchMap` if actually you just return an object? I think you can use `map` instead, right?

RemcoSimonides (approved):

RemcoSimonides (changes requested):
💩

apps/crm/crm/src/app/crm/pages/event/event.component.ts
phpgeekfr: switchMap is needed here

apps/crm/crm/src/app/crm/pages/event/event.component.ts
phpgeekfr: fixed

Issue #7347, opened 2021-12-07
Release 3.6 - Movie


RemcoSimonides:
Issues fixed in this release
- #7045 
- #7232
- #7166
- #7065
- #7023
- #4188
- #7242
- #7172

cams-goussale:
Hello @pldespaigne, Yuki has identified an bug on the **Marketplace Avails Filter List**. I know you're working on avails logic in a separate branch but this is a bug that would need a fix for the release.

The [rule]([url](https://docs.google.com/spreadsheets/d/1YAsKVdFt_ybheXlLJsKoXYdlma-iB1W_ilKPTuHME2w/edit#gid=177209628&range=B23)) that is not respected is the following: "if mandate excl and research non excl => yes"

Steps to reproduce:
1) Go to Title page
2) Filter with some Avails Criteria with Exclusive
3) Filter with the same criteria with Not Exclusive
The same titles should appear but it's not the case.

FYI, the map view and the calendar view have the correct logic. It's only the List view which is wrong.

https://user-images.githubusercontent.com/93206363/145191174-dc3158cf-ac2b-42cc-97bc-9dcc49f3f4d0.mp4

RemcoSimonides:
Feedback from  @2amarkina 
Note from Remco: I remember we used to have this; there must be a reason why it is removed. Only add back if there is time left because it wasn't in the scope of this release.
### To apply on Archipel Market and Archipel Content  
- [ ] On the Titles List we should have "There are # titles available" on the left. If a filter is applied, then need to change the number accordingly. 
![Снимок экрана 2021-12-08 в 11.19.11.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/7e58ade6-a325-4c4a-891f-e2e0a80b4abc)

### To apply on Media Financiers
- [ ] Same as for the AC and AM above, but change the text: on the Titles List we should have "There are # projects available" on the left. If a filter is applied, then need to change the number accordingly.

RemcoSimonides:
Feedback from @2amarkina 
**From homepage to film page via "see all"**
- [ ] To complete the issue quoted in #7172 about the highlighted genre, when a particular genre is selected, could we put the "genres" filter into active state to make it visible? 
_from this_
![Снимок экрана 2021-12-08 в 11.00.17.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/2252a381-3590-4e0d-9cc6-fb2c344c9975)

_to this_
[...]

PR #7371, opened 2021-12-08
trailer link updates #7347

phpgeekfr:


RemcoSimonides (approved):
What issue did you fix?
I'm curious what this `scrollIntoView()` is for. Is it if someone has a link to the trailer, it immediately scrolls to the trailer?

phpgeekfr:
I fixed the checkboxes with my name on it in #7347 (PDF)

Yes scrollIntoView is for scrolling to the trailer, I cannot use standard fragment behavior because it is not there already on pageload

Issue #7349, opened 2021-12-07
Release 3.6 - Event


RemcoSimonides:
Issues fixed in this release so please double check
- #7160 (only the checked checkboxes)
- #7004
- #7169
- #6978

2amarkina:
OS: Mac BigSur, 11.6
Browser: 15.0

- [ ] Following this #6978, when I create a screening event and do not select a title to display, I can't see any chip that appear, nor any warning.

2amarkina:
**Events without connexion** 

Meeting Event Page 
- [ ] now when we connect with anonymous user (Public event) in our Meeting event page layout the logo of organization leads to the account creation page (when a user is connected it leads to the Organization page). So, wouldn't it be better to lead an anonymous user to the Landing page instead of account creation? @Jackseed @cams-goussale 
![Снимок экрана 2021-12-08 в 13.08.54.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/59d7fa0d-7a76-478c-b6fb-4d04b3e9ddda)

Jackseed:
> **Events without connexion**
> 
> Meeting Event Page
> 
> * [ ]  now when we connect with anonymous user (Public event) in our Meeting event page layout the logo of organization leads to the account creation page (when a user is connected it leads to the Organization page). So, wouldn't it be better to lead an anonymous user to the Landing page instead of account creation? @Jackseed @cams-goussale
>   ![Снимок экрана 2021-12-08 в 13.08.54.png](https://camo.githubusercontent.com/9b1fcbace59976717c3d6bd95f59fa5955ec0f784e45b01a915681458867125f/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3564643534353238633063636433303030313336373263332f35396437666130642d376137362d343738632d623666622d346430346233653964646461)

Indeed, if we can it would be better to lead to landing. And I would say even better not to be clickable at all for anonymous users but this I don't know if that's doable..

cams-goussale:
- [ ] Invitations in /crm/event/[eventId] stopped showing

![image](https://user-images.githubusercontent.com/93206363/145211481-c0febecf-6c71-48ba-85ef-0da0b1efc611.png)

2amarkina:
**Events without connexion**

Protected meeting event / Guest side

[...]

PR #7380, opened 2021-12-09
 Release 3.6 - Event #7349 

phpgeekfr:
fixed: 
-  If I invite an user to my org and then invite him to an event, he will receive 3 emails 
-  On event edit page, form is saved before sending invitations
-  As a guest I received an email, clicked on it, filled in the little form and got access to the event page where I could answer an invitation. Tried to click on "accept invitation", but nothing happened. Same for decline.

### DIFF
--- a/libs/event/src/lib/form/shell/shell.component.ts
+++ b/libs/event/src/lib/form/shell/shell.component.ts
@@ -102,17 +102,19 @@ export class EventFormShellComponent implements OnInit, OnDestroy {
     this.formSub.unsubscribe();
   }
 
-  save() {
+  async save(options: { showSnackbar: boolean } = { showSnackbar: true }) {
     if (this.form.valid && this.form.dirty) {
       const value = this.form.value;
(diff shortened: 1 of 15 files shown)


libs/event/src/lib/form/shell/shell.component.ts
```diff
@@ -102,17 +102,19 @@ export class EventFormShellComponent implements OnInit, OnDestroy {
     this.formSub.unsubscribe();
   }
 
-  save() {
+  async save(options: { showSnackbar: boolean } = { showSnackbar: true }) {
```
RemcoSimonides, line 105: You could do `async save({ showSnackbar = true })`

RemcoSimonides (approved):

libs/event/src/lib/form/shell/shell.component.ts
phpgeekfr: If I do this, argument will be mandatory

PR #7383, opened 2021-12-10
7347 highlight filters if prefilled by query params

RemcoSimonides:
#7347

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/home/home.component.ts
+++ b/apps/catalog/catalog/src/app/marketplace/home/home.component.ts
@@ -52,7 +52,6 @@ export class MarketplaceHomeComponent implements OnInit, AfterViewInit {
       this.authService.update({ preferences });
       this.dialog.open(PreferencesComponent, {
         height: '80vh',
-        width: '80vw',
         autoFocus: false
       });
     }
--- a/apps/festival/festival/src/app/marketplace/title/list/list.component.ts
+++ b/apps/festival/festival/src/app/marketplace/title/list/list.component.ts
@@ -59,6 +59,9 @@ export class ListComponent implements OnInit, OnDestroy, AfterViewInit {
       }
     }
 
+    const decodedData: MovieSearch = decodeUrl(this.route);
+    if (decodedData && Object.keys(decodedData).length) this.searchForm.hardReset(decodedData, true);
+
     const sub = this.searchForm.valueChanges.pipe(startWith(this.searchForm.value),
       distinctUntilChanged(),
       debounceTime(500),
@@ -87,8 +90,6 @@ export class ListComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 
   ngAfterViewInit(): void {
-    const decodedData: MovieSearch = decodeUrl(this.route);
-    if (decodedData && Object.keys(decodedData).length) this.searchForm.hardReset(decodedData)
 
     const sub = this.searchForm.valueChanges.pipe(
       debounceTime(1000),
--- a/libs/utils/src/lib/form/forms/entity.form.ts
+++ b/libs/utils/src/lib/form/forms/entity.form.ts
@@ -66,13 +66,16 @@ export class FormEntity<C extends EntityControl<T>, T = any> extends FormGroup {
     }
   }
 
-  hardReset(values: Partial<T>={}){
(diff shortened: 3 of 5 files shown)


libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -149,8 +149,13 @@ export class FormList<T, Control extends AbstractControl = any> extends FormArra
     super.patchValue(value, options)
   }
 
-  hardReset(values: GetPartial<T>[]) {
+  hardReset(values: GetPartial<T>[], markAsDirty: boolean) {
     this.clear()
-    values.forEach(value => this.add(value));
+    if (Array.isArray(values)) {
```
phpgeekfr, line 154: is there any situation where `values` is not an array ?

libs/utils/src/lib/form/forms/entity.form.ts
```diff
@@ -66,13 +66,16 @@ export class FormEntity<C extends EntityControl<T>, T = any> extends FormGroup {
...
-  hardReset(values: Partial<T>={}){
+  hardReset(values: Partial<T>={}, markAsDirty: boolean) {
     for (const [key, control] of Object.entries(this.controls)) {
       const value = values[key as keyof T] as any;
       if (control instanceof FormEntity || control instanceof FormList) {
-        control.hardReset(value);
+        control.hardReset(value, markAsDirty);
       } else if(control instanceof AbstractControl) {
         control.reset(value);
+        if (markAsDirty && value && !Array.isArray(value) || (Array.isArray(value) && value.length)) {
```
phpgeekfr, line 76: can you add more parenthesis to make the `||` more readable ? 

Looks like that this can be simplified:

```
markAsDirty && value && (!Array.isArray(value) ||  value.length)
```

Since @yspringfield and @GrandSchtroumpf worked on this recently, I think they should have a look at it.

libs/utils/src/lib/form/forms/list.form.ts
```diff
@@ -149,8 +149,13 @@ export class FormList<T, Control extends AbstractControl = any> extends FormArra
...
+      values.forEach(value => {
```
phpgeekfr, line 155: maybe 
```
values.forEach(value => this.add(value));
if(values.some(value => value) && markAsDirty) this.markAsDirty();
```

apps/festival/festival/src/app/marketplace/title/list/list.component.ts
```diff
@@ -59,6 +59,9 @@ export class ListComponent implements OnInit, OnDestroy, AfterViewInit {
       }
     }
 
+    const decodedData: MovieSearch = decodeUrl(this.route);
+    if (decodedData && Object.keys(decodedData).length) this.searchForm.hardReset(decodedData, true);
```
phpgeekfr, line 63: this is the only call to `this.searchForm.hardReset(` so why not make `markAsDirty` always true  in both `hardReset` methods ?

RemcoSimonides: The associate naming `reset` with resetting the form and thus also reset any dirty/touched controls. So it made more sense for me to explicitly define marking it dirty if you want. In case it will be used more often

libs/utils/src/lib/form/forms/list.form.ts
RemcoSimonides: Yes, if I remember correctly it was with contentType

libs/utils/src/lib/form/forms/list.form.ts
RemcoSimonides: `values.some(value => value)` is the same as `values.length`?

RemcoSimonides: Corrent, thanks!

apps/catalog/catalog/src/app/marketplace/home/home.component.ts
```diff
@@ -52,7 +52,6 @@ export class MarketplaceHomeComponent implements OnInit, AfterViewInit {
       this.authService.update({ preferences });
       this.dialog.open(PreferencesComponent, {
         height: '80vh',
-        width: '80vw',
```
GrandSchtroumpf, line 55: maybe `maxWidth` to make sure it doesn't go crazy

RemcoSimonides:
Closing in favor of #7395

Issue #7379, opened 2021-12-09
[CATALOG] Update sale's status when first negotiation started


GrandSchtroumpf:
We do not have any mechanism to switch the status of the `Sale` from "pending" to "negotiating" right now.
It should switch from "pending" to "negotiating" when the seller create an negotiation for the first time.

An easy, but expensive, solution would be to check on every negotiation create and change the status of the sale if needed.
See if we can have a cleaner / less expansive solution.

PR #7384, opened 2021-12-10
update sale status on negotiation creation

yspringfield:
Fixed: #7379

### DIFF
--- a/libs/contract/src/lib/contract/+state/contract.service.ts
+++ b/libs/contract/src/lib/contract/+state/contract.service.ts
@@ -5,7 +5,7 @@ import { ContractDocument, convertDuration, createMandate, createSale, Holdback,
 import { createDocumentMeta, formatDocumentMetaFromFirestore } from "@blockframes/utils/models-meta";
 import { Timestamp } from "@blockframes/utils/common-interfaces/timestamp";
 import { NegotiationService } from '@blockframes/contract/negotiation/+state/negotiation.service';
-import { map } from 'rxjs/operators';
+import { first, map } from 'rxjs/operators';
 import { QueryFn } from '@angular/fire/firestore';
 import { OrganizationQuery } from '@blockframes/organization/+state';
 import { Negotiation } from '@blockframes/contract/negotiation/+state/negotiation.firestore';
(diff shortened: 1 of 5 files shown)


libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -48,4 +50,29 @@ export class ContractService extends CollectionService<ContractState> {
...
+      stakeholders: contract.stakeholders,
+      buyerId: contract.orgId,
+      price: contract.price,
+      titleId: contract.titleId,
+      terms: contract.terms,
+      holdbacks: contract.holdbacks,
+      parentTermId: contract.parentTermId,
+      specificity: contract.specificity,
+      orgId: contract.orgId,
+    })
```
GrandSchtroumpf, line 74: You can pass the `write` operation into the `add` method from the negotiation service
```suggestion
    const write = this.batch();

    this.negotiationService.add({
      _meta: createDocumentMeta({ createdAt: new Date() }),
      status: 'pending',
      createdByOrg: activeOrgId,
      sellerId: centralOrgId.catalog,
      stakeholders: contract.stakeholders,
      buyerId: contract.orgId,
      price: contract.price,
      titleId: contract.titleId,
      terms: contract.terms,
      holdbacks: contract.holdbacks,
      parentTermId: contract.parentTermId,
      specificity: contract.specificity,
      orgId: contract.orgId,
    }, { write });
```

libs/contract/src/lib/contract/+state/contract.service.ts
```diff
@@ -48,4 +50,29 @@ export class ContractService extends CollectionService<ContractState> {
...
+    this.update(contractId, { status: 'negotiating' }, { write })
```
GrandSchtroumpf, line 75: Maybe update only if sale's status is pending :
```suggestion
    const sale = await this.valueChanges(contractId).pipe(take(1)).toPromise(); 
    if (sale.status === 'pending') this.update(contractId, { status: 'negotiating' }, { write });
```

GrandSchtroumpf (approved):

Issue #7350, opened 2021-12-07
Release 3.6 - General UI


RemcoSimonides:
Issues fixed in this release so please double check them:
- #6919
- #7172
- #6465
- #6238

2amarkina:
- [ ] To complete the issue quoted in [#7172] about the highlighted genre, when a particular genre is selected, could we put the "genres" filter into active state to make it visible? 
_from this_
![Снимок экрана 2021-12-08 в 11.00.17.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/2252a381-3590-4e0d-9cc6-fb2c344c9975)

_to this_
![Снимок экрана 2021-12-08 в 11.00.51.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/17794065-c610-4d47-982e-044453e22b08)

2amarkina:
OS: Mac BigSur, 11.6
Browser: Safari 15.0

**App tested: ARCHIPEL MARKET** 

Homepage
- [ ] If possible and quick, then add a link to the banner of Screen Operation. The link should lead to the Screening Schedule of screen as CTA "Upcoming Screenings" below 
![Снимок экрана 2021-12-08 в 12.07.34.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/0601cf61-f1a1-46dd-8f18-b43475fc65c5)

RemcoSimonides:
I'll fix this one:
Query Params don't work on CTA buttons in banner component homepage
`apps/catalog/catalog/src/app/marketplace/home/home.component.html` line 120

RemcoSimonides:
- [] #7243

cams-goussale:
- [ ] Could we add the anchor functionality to homepage as well /c/o/marketplace/home ?
(Same issue as #6465 which was for the landing page)

2amarkina:
### Archipel Content / Landing Page

In the block "Subscribe now" the CTA should be changed to "Join now" instead of "Schedule demo"
<img width="1369" alt="Снимок экрана 2021-12-15 в 09 48 48" src="https://user-images.githubusercontent.com/55499036/146153900-7afe1f09-43d9-44d1-87ac-8c611bf9a800.png">

cams-goussale:
- [ ] Let's remove the section "They're talking about us" on the landing page of AC. We don't have enough articles talking about us

PR #7412, opened 2021-12-14
Change logo click behaviour

Poppe-A:
#7350

### DIFF
--- a/libs/ui/src/lib/layout/marketplace/marketplace.component.scss
+++ b/libs/ui/src/lib/layout/marketplace/marketplace.component.scss
@@ -14,10 +14,16 @@ mat-sidenav-content {
   header {
     width: 100%;
 
-    button {
+    button:first-of-type {
       margin-right: 24px;
     }
 
+    button[app-logo] {
+      background-color: var(--background-card);
+      border: none;
+      cursor: pointer;
+    }
+    
     /* We can't use the fxLayoutGap directive because it set the margin right
     property to the specified value. But we need to use margin-right:auto to split
     the content and we don't want to use !important */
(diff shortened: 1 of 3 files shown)


libs/ui/src/lib/layout/marketplace/marketplace.component.ts
```diff
@@ -64,6 +65,14 @@ export class MarketplaceComponent implements OnInit {
   animationOutlet(outlet: RouterOutlet) {
     return outlet?.activatedRouteData?.animation;
   }
+
+  onAppLogoClick() {
+    if (this.router.url.includes("home")) {
+     this.cdkScrollable.scrollTo({ top: 0 });
+    } else {
+      this.router.navigate(['/c/o/marketplace/home'])
```
RemcoSimonides, line 73: It used to be a navigation to `"."` So I think you should navigate to that again

libs/ui/src/lib/layout/marketplace/marketplace.component.ts
```diff
@@ -64,6 +65,14 @@ export class MarketplaceComponent implements OnInit {
   animationOutlet(outlet: RouterOutlet) {
     return outlet?.activatedRouteData?.animation;
   }
+
+  onAppLogoClick() {
+    if (this.router.url.includes("home")) {
```
RemcoSimonides, line 70: I think you can check here whether you are on the full url; `/c/o/marketplace/home` instead. In case some day there will be a url with `home` somewhere in there

libs/ui/src/lib/layout/marketplace/marketplace.component.scss
```diff
@@ -14,10 +14,16 @@ mat-sidenav-content {
   header {
     width: 100%;
 
-    button {
+    button:first-of-type {
```
RemcoSimonides, line 17: This is a different button than the `app-logo` button?

Poppe-A: yep, the menu button

RemcoSimonides (approved):

PR #7427, opened 2021-12-15
6235 links in notification 2

RemcoSimonides:
Better than #7318?

### DIFF
--- a/libs/notification/src/lib/components/item/item.component.html
+++ b/libs/notification/src/lib/components/item/item.component.html
@@ -1,7 +1,9 @@
 
 <mat-list-item>
   <img matListAvatar [asset]="notification.placeholderUrl" [ref]="notification.imgRef" alt="Notification picture">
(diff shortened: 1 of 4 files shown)


libs/notification/src/lib/components/item/item.component.html
```diff
@@ -20,5 +22,42 @@
...
+<ng-template #contractCreated>
+  {{ notification.message }}
+</ng-template>
+<ng-template #offerCreatedConfirmation>
+  {{ notification.message }}
+</ng-template>
+
+<!-- Other notifications -->
+<ng-template #movieSubmitted>
+  <a [routerLink]="['/c/o/marketplace/title', notification['titleId']]">{{ notification['title'] }}</a> {{ notification.message }}
```
RemcoSimonides, line 62: Getting compile error saying `titleId` doesn't exist on notification because it isn't aware it's always of certain type at this point

Issue #7432, opened 2021-12-16
catalog landing component css/html updates


phpgeekfr:
@GrandSchtroumpf suggested some updates on the catalog landing component :
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/b9a30ce1-7644-4611-b951-53a706dc2ca8)

created from PR https://github.com/blockframes/blockframes/pull/7422

Issue #7336, opened 2021-12-03
Images not uploading


cams-goussale:
## Your environement
- OS: Windows and Mac
- Browser (& version if possible): Chrome

### Localization
- App name: Festival
- Page: Movie tunnel/Promotional Elements
- Url: c/o/dashboard/tunnel/movie/[movieId]/media-videos
- User used (if anonymized): 38rjqqlz0hMlYKOugRv5hh4UyA73
- Sentry error that might help:
[Prod one](https://sentry.io/organizations/cascade-8/issues/2485237942/?environment=production&project=1540126&query=is%3Aunresolved) (from Marie, using Movie tunnel)
[Staging one](https://sentry.io/organizations/cascade-8/issues/2485237942/?environment=staging&project=1540126&query=is%3Aunresolved) (from me, using CRM)

### Context
Marie and I tried to upload 2 images on a title and we were not able to have it displayed. It only bugged with these 2 specific images.
![TTB_20201112_09340r (1)](https://user-images.githubusercontent.com/93206363/144636693-8564ed7e-f1a1-4876-9072-f03346185bbe.jpg)

![TTB_20201006_02579r (1)](https://user-images.githubusercontent.com/93206363/144636865-e03bcce2-ca15-4a8e-8b83-9f2d20766c74.jpg)

In staging, I tried to upload them on movie id ckdjrRuFoUaZbhLi6QGL
Here is the result sent to db
![image](https://user-images.githubusercontent.com/93206363/144637002-55565737-26d3-4924-ae05-70fca7a2cbab.png)

And here is what happened on Marketplace view
![image](https://user-images.githubusercontent.com/93206363/144637077-2e4f1a91-8d59-4cfa-b70e-7fb4d9ec497c.png)

RemcoSimonides:
This issue is caused because of its file size after a needed conversion in the code. This conversion is done _after_ cropping, so it's already past the moment of initial selection of the file.

Most of the time, the size of the file is similar before and after the conversion but for some reason it isn't with these images. So this error is an exception.

I would suggest to show an error despite it feeling weird that you select a file of 3.3MB and get a warning when selection it saying that file size is too big even though it looks like below the max of 5MB.

We could either show it immediately when selecting the file (that would mean doing the conversion to test its size earlier and perhaps having bit worse performance), or because it only happens rarely, to show the warning after cropping the image. For this last solution we probably need to remove the file again so the user doesn't upload it accidently.

@2amarkina @cams-goussale  what do you think would be best?

Issue #7257, opened 2021-11-25
Upgrade img upload limit from 5Mb to 20Mb


Jackseed:
We should let users upload heavier files as they are used to have more than 5Mb stills. Imgix will take care of compressing it so this shouldn't affect loading time. 

This means that we will have to update imgix config on project accordingly. Text within upload box should also be changed to reflect the new limit.

PR #7443, opened 2021-12-20
7336 7257 7432

RemcoSimonides:
#7336 #7257 #7432


libs/media/src/lib/image/uploader/uploader.component.ts
```diff
@@ -77,7 +77,7 @@ export class ImageUploaderComponent implements OnInit, OnDestroy {
   metadata: FileMetaData;
   fileName: string;
 
-  public maxSize: number = 5 * 1000000;
+  public maxSize: number = 20 * 1000000;
```
pldespaigne, line 80: Could be nice to add a comment, something like: `// 20MB size limit`

pldespaigne (approved):

PR #7468, opened 2021-12-23
circleCI optimisations

george43g:
- remove need for build
- remove cache hits not used


.circleci/config.yml
```diff
@@ -548,7 +533,7 @@ workflows:
           filters:
             branches:
               ignore: *deployment_branches
-      - build-and-lint-commits:
+      - lint-commits:
```
phpgeekfr, line 536: line 530: rename and change comment for `build-and-test: # build, lint and unit test commits that aren't release related`

george43g: done

phpgeekfr (approved):

Issue #7463, opened 2021-12-22
Clean Avails dead code


pldespaigne:
Delete the `avails.ts` file
- [x] delete dead code
- [x] delete code used only in test (require test refactoring before)
- [x] delete `avails.ts` file
- [x] rename `new-avail.ts` into `avails.ts` and change imports

PR #7470, opened 2021-12-23
Clean some avails code

pldespaigne:
solve one item of #7463

### DIFF
--- a/libs/contract/src/lib/avails/new-avails.ts
+++ b/libs/contract/src/lib/avails/new-avails.ts
@@ -425,18 +425,18 @@ export function durationAvailabilities(
 //          COMPARISON
 // ----------------------------
 
-export function isSameMapBucketTerm(avails: MapAvailsFilter, termB: BucketTerm) {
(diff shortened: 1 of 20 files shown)


libs/contract/src/lib/avails/new-avails.ts
```diff
@@ -448,6 +433,14 @@ export function isSameMapBucketTerm(avails: MapAvailsFilter, termB: BucketTerm)
   return exclusivityCheck && durationCheck && mediasCheck;
 }
 
+export function isSameCalendarBucketTerm(avails: CalendarAvailsFilter, termB: BucketTerm) {
```
phpgeekfr, line 436: why is it named `termB` ?

phpgeekfr (approved):

libs/contract/src/lib/avails/new-avails.ts
pldespaigne: Because I started with a copy/paste from other comparison functions with `termA` & `termB` ^^
I will rename

Issue #7488, opened 2021-12-29
Unit-Test : Implement a simple example of `onCreate` db trigger


phpgeekfr:
A simple db trigger to test can be this one : 
```
export const onDocumentPermissionCreateEvent = onDocumentCreate(
  'permissions/{orgID}/documentPermissions/{docID}',
  onDocumentPermissionCreate
);
```

This is triggered when `addDocumentPermissions` from `PermissionsService` is called.

When `onDocumentPermissionCreate` is triggered, it simply creates a new doc in `docsIndex` so the test can be to check if document exists after the call.

Another test could be `onEventDelete` : 

```
export const onEventDeleteEvent = onDocumentDelete('events/{eventID}', onEventDelete);
```
When an event document is deleted, it is supposed to remove related `notifications` and `invitations` so the test can be to check if thoses documents are deleted when an event is deleted

xmano:
Achieved unit tests for both the functions. The PR is here:
https://github.com/blockframes/blockframes/pull/7489

Final review comments need to be checked and changes approved. It should be good to merge.

PR #7489, opened 2021-12-29
Add onDocumentPermissionCreate unit test

xmano:
Partial changes for Issue #7488


xmano:
Failing test, with `removeAllSubcollections()` in permissions.ts

image

This actually should have cleaned up the sub collections but since we are checking for presence of the values it fails.

xmano:
Working tests!

image

xmano:
Adding events unit-test to the same PR. Making it draft for now.

apps/backend-functions/src/permissions.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+      const wrapped = testEnv.wrap(onDocumentPermissionCreateEvent);
+      const docID = 'D001';
+      const orgID = 'O001';
+
+      const context = {
+        params: { docID, orgID }
+      };
+
+      // Make a fake document snapshot to pass to the function
+      const after = testEnv.firestore.makeDocumentSnapshot(
```
phpgeekfr, line 38: We are not supposed to pass a documentPermission document to  `await wrapped(after, context);` instead of `docsIndex` ?

apps/backend-functions/src/permissions.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+      // Call the function
+      await wrapped(snap, {});
+
+      snap = await db.doc(`permissions/${orgID}`).get();
+      expect(snap.data()).toEqual(
+        expect.objectContaining({
+          authorOrgId: orgID,
+          orgs: { a: 'O1', b: 'O2' }
+        })
+      );
```
phpgeekfr, line 72: I don't understand why you test this

apps/backend-functions/src/fixtures/data.ts
```diff
@@ -51,5 +51,16 @@ export const testFixture = {
       email: 'test@cascade8.com'
     },
     status: 'pending'
-  }
+  },
+
+  //Permission
+  'permissions/O001': {
+    authorOrgId: 'O001',
```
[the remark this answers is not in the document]
xmano: ok. will change

apps/backend-functions/src/permissions.spec.ts
xmano: Yes it is needed, but for the other scenario : `onPermissionDelete()`.

It seems ok to me.

apps/backend-functions/src/permissions.spec.ts
xmano: This is to check other fields are not removed only sub-collection is removed. Can be removed.

xmano: Modified to check

```
id: O001
roles: {}
```

apps/backend-functions/src/event.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+        //id: eventID,
+        organizerUid: orgID,
+        titleId: 'UnitTest Event',
+        description: 'Event organised by Jest',
+      });
+
+      const eventData = {
+        id: eventID,
+        ...localEvent
+      }
```
phpgeekfr, line 46: why not creating eventData directly ?

apps/backend-functions/src/event.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      await wrapped(eventSnap);
+
+      let collectionRef = await db.collection('invitations')
+                                          .where('eventId',  '==', eventID)
+                                          .get();
+      let queriedDocs = collectionRef.docs;
+      expect(queriedDocs).toHaveLength(0);
+
+      collectionRef = await db.collection('notifications')
+                              .where('eventId',  '==', eventID)
```
phpgeekfr, line 63: for notifications the attribute to check is `docId`

apps/backend-functions/src/event.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+        `events/${eventID}`
+      );
+
+      //Trigger onEventDeleteEvent event
+      await wrapped(eventSnap);
+
+      let collectionRef = await db.collection('invitations')
+                                          .where('eventId',  '==', eventID)
+                                          .get();
+      let queriedDocs = collectionRef.docs;
```
phpgeekfr, line 59: can you also perform both query  before the trigger so we are sure that we have 2 invitations and 1 notification matching this event ?

apps/backend-functions/src/permissions.spec.ts
phpgeekfr: it the same if it was in `apps/backend-functions/src/fixtures/data.ts`  ?  Because I don't get why you give a docsIndex document to `onDocumentPermissionCreateEvent`

apps/backend-functions/src/permissions.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+      // Call the function
+      await wrapped(snap, {});
+
+      snap = await db.doc(`permissions/${orgID}`).get();
+      expect(snap.data()).toEqual(
+        expect.objectContaining({
+          id: orgID,
+          roles: { }
+        })
+      );
```
phpgeekfr, line 72: why doing this test here ? in real situation, this document would have been deleted.

apps/backend-functions/src/permissions.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      const docRef = db.doc(`permissions/${orgID}/documentPermissions/${docID}`);
+      snap = await docRef.get();
+      expect(snap.data()).toBeUndefined();
```
phpgeekfr, line 76: can you check that document exists before the call to `await wrapped(snap, {})` ?

apps/backend-functions/src/permissions.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      // Call the function
+      await wrapped(after, context);
+
+      const snap = await db.doc(`docsIndex/${docID}`).get();
+      expect(snap.data()).toEqual(
+        expect.objectContaining({
+          authorOrgId: orgID
+        })
+      );
```
phpgeekfr, line 53: can you check before calling `await wrapped(after, context);` that `authorOrgId` is undefined ?

apps/backend-functions/src/event.spec.ts
xmano: Good catch.. correcting it.

apps/backend-functions/src/permissions.spec.ts
xmano: If you see the code above this property does not exist in the document. 
```
      // Make a fake document snapshot to pass to the function
      const after = testEnv.firestore.makeDocumentSnapshot(
        {
          id: docID,
        },
        `docsIndex/${docID}`
      );
```

phpgeekfr: I said that because it looks like that document can be defined in fixtures instead of creating it with `testEnv.firestore.makeDocumentSnapshot(`

phpgeekfr: But if you prefere to create document this way instead of in fixtures. fine for me

apps/backend-functions/src/permissions.spec.ts
xmano: Implemented.

apps/backend-functions/src/permissions.spec.ts
xmano: This is to check if the attributes are available after `onPermissionDeleteEvent()`. According to the code only sub-collections of `permissions/${orgID}` are cleaned.

apps/backend-functions/src/event.spec.ts
xmano: Implemented

xmano:
> it the same if it was in apps/backend-functions/src/fixtures/data.ts ? Because I don't get why you give a docsIndex document to onDocumentPermissionCreateEvent

```
      const after = testEnv.firestore.makeDocumentSnapshot(
        {
          id: docID,
        },
        `docsIndex/${docID}`
      );
```

`docsIndex/${docID}`  does  not exist in fixtures data.


This document is required to be present before updating it, else we get error here inisde `onDocumentPermissionCreate()`

apps/backend-functions/src/event.spec.ts
xmano: TODO for now.

I could not find a way to get eventID directly.

apps/backend-functions/src/event.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  describe('Event spec', () => {
+
+    it('removes all matching eventId docs from "invitations & notification" collection', async () => {
+      const wrapped = testEnv.wrap(onEventDeleteEvent);
+      const eventId = 'E001';
+
+      const eventToDelete = { id: eventId };
+
+      const eventSnap = testEnv.firestore.makeDocumentSnapshot(
```
RemcoSimonides, line 34: Make namne it 'eventToDeleteSnap' and put the `{ id: eventId }` directly in this `makeDocumentSnapshot` function

RemcoSimonides (approved):

Issue #7313, opened 2021-12-02
Avails Logic E2E Scenarios


cams-goussale:
As requested in #7139, [here are some E2E and Unit Test scenarios](https://docs.google.com/spreadsheets/d/1YAsKVdFt_ybheXlLJsKoXYdlma-iB1W_ilKPTuHME2w/edit#gid=0) for the Avails Filter Logic.

There were already some tests done in the past that we can reuse:
- 1 title Avails Search #4918 
- Multi-title research #5224 
- E2E scenario #5353 

In the excel, you can find:
- [ ] 3 E2E scenarios: 2 new, 1 improved from the past
- [ ] many unit tests: 5 new, 1 with a change

cams-goussale:
Hello @RemcoSimonides this issue gathers the different tests that should be done to secure the Avails logic. PL estimated the first 2 issues about unit tests #4918 #5224 to 3 days of dev. Any chance we can include it for the release of January?

RemcoSimonides:
Hee @cams-goussale, could you be explicit in what needs to be done? Issue #4918 contains only checked checkboxes, #5224 contains only open checkboxes but then a merged PR and closed, and the excel file contains several sheets and I don't know where to look at.

Also, let's not focus on e2e test for now since they are largely disabled and waiting for rework from George to be more reliable.

cams-goussale:
It's not e2e, it's [unit tests ](https://docs.google.com/spreadsheets/d/1YAsKVdFt_ybheXlLJsKoXYdlma-iB1W_ilKPTuHME2w/edit#gid=1782901115)which are written in these 2 issues #4918 #5224
The sheet in the excel is the one called "Unit test". It contains:
- 1 block for Single movie Avails Research
- 1 block for Multiple movies Avails Research

The tests with box checked are the same as in the issue. The tests not checked are new ones which complete the initial issue. We thought these unit tests were already in place, but from what PL told us, they're not.

When talking about "Single movie" or "Multiple movies" I refer to the information that you can find in the tabs [Import Movies ](https://docs.google.com/spreadsheets/d/1YAsKVdFt_ybheXlLJsKoXYdlma-iB1W_ilKPTuHME2w/edit#gid=1224333819)and [Import Contracts](https://docs.google.com/spreadsheets/d/1YAsKVdFt_ybheXlLJsKoXYdlma-iB1W_ilKPTuHME2w/edit#gid=989765594).

RemcoSimonides:
[...]

PR #7495, opened 2021-12-30
Better tests & cleaning

pldespaigne:
- solve items #7463
- solve #7313


libs/contract/src/lib/bucket/tests/bucketForm.spec.ts
```diff
@@ -0,0 +1,482 @@
...
+    const markerC: DurationMarker = {
+      from: new Date(27),
+      to: new Date(32),
+      contract: mandateA,
+      term: termC,
+    };
+
+    bucket.addDuration(calendarFilterA, markerA);
+    bucket.addDuration(calendarFilterA, markerB);
+    bucket.addDuration(calendarFilterA, markerC);
```
phpgeekfr, line 459: it is not better to add this directly in the test itself ?

phpgeekfr (approved):

PR #7520, opened 2022-01-07
Fix issue with table in neogitation form

GrandSchtroumpf:
1. Implement `bf-table` in `bf-table-form`.
2. Unsubscribe old subscription when registerOnChange


yspringfield (approved):
LGTM

PR #7533, opened 2022-01-12
akita permissions removal #7285

phpgeekfr:


RemcoSimonides:
There is a proposal to remove 'super admin', 'admin', and 'member' btw
https://app.zenhub.com/workspaces/design---ux-60ab5f37bd1f990011c933a2/issues/blockframes/blockframes/6753

libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -24,7 +24,13 @@ import { FireAnalytics } from '@blockframes/utils/analytics/app-analytics';
 export class OrganizationService extends CollectionService<OrganizationState> {
   readonly useMemorization = true;
 
-  private app = getCurrentApp(this.routerQuery)
+  private app = getCurrentApp(this.routerQuery);
```
GrandSchtroumpf, line 27: you're not using it. Do you still need it ?

libs/organization/src/lib/permissions/+state/permissions.service.ts
```diff
@@ -69,4 +105,14 @@ export class PermissionsService extends CollectionService<PermissionsState> {
...
     }
   }
+
+  /** Checks if the user is admin of his organization. */
+  public isUserAdmin(userId: string = this.auth.userId): boolean {
+    return this.permissions.roles[userId] === 'admin' || this.isUserSuperAdmin(userId);
+  }
+
+  /** Checks if the user is superAdmin of his organization. */
+  private isUserSuperAdmin(userId: string): boolean {
```
GrandSchtroumpf, line 115: you don't fallback to auth.userId ?

GrandSchtroumpf (approved):

libs/organization/src/lib/organization/+state/organization.service.ts
phpgeekfr: this.app is still used in the service

libs/organization/src/lib/permissions/+state/permissions.service.ts
phpgeekfr: this private function is only used by `isUserAdmin` that always provides userId

PR #7565, opened 2022-01-18
7502 spinner for dashboard event list page

RemcoSimonides:
Still under review by Stephanie but let's review the code already too in case I need to change too much and thus changing the UI too


libs/ui/src/lib/layout/dashboard/dashboard.component.scss
```diff
@@ -35,6 +36,19 @@ mat-sidenav-content {
...
     }
   }
+
+  .overlay {
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    background-color: #1f1f26c0;
```
RemcoSimonides, line 46: This is 
`background-color: var(--background-disabled-button);`
but then with opacity added to it. Couldn't find a way to still use the variable and add opacity. Should I add the color to the builder maybe?

RemcoSimonides:
Closing in favor of #https://github.com/blockframes/blockframes/pull/7586

PR #7572, opened 2022-01-19
update event stats #7570

### LINKED ISSUE #7570: Improve event page statistics
Following #7560

Currently, to display this kind of page
 

We first fetch `pageView` from BigQuery and we consolidate data with `WatchTime` coming from invitations. (note that data for users of the same orgId as event owner or blockframesAdmin will not be displayed )

This is working ok for users that have a real account because uid in bigquery is the same as in the invitation document. But this does not work for anonymous users for protected events.

What we could do is, instead of fetching pageView, then watchTime, we could fetch watchTime only and display a table like this:

```
First Name1 | LastName | OrgName | WatchTime
Anonymous1                                      | WatchTime
First Name2 | LastName | OrgName | WatchTime
First Name2 | LastName | OrgName | WatchTime
Anonymous2                                      | WatchTime
```

The only problem with this is that users that went to event page without starting the video will have no watchTime so we will not be able to display users with watchTime = 0

phpgeekfr:
fix #7570

### DIFF
--- a/libs/event/src/lib/pages/analytics/analytics.component.ts
+++ b/libs/event/src/lib/pages/analytics/analytics.component.ts
@@ -6,10 +6,21 @@ import { pluck, switchMap, tap } from 'rxjs/operators';
 import { EventService } from '@blockframes/event/+state';
 import { Observable } from 'rxjs';
 import { BehaviorStore } from '@blockframes/utils/observable-helpers';
-import { EventAnalytics, EventMeta, EventTypes } from '@blockframes/event/+state/event.firestore';
+import { EventMeta, EventTypes } from '@blockframes/event/+state/event.firestore';
 import { InvitationQuery } from '@blockframes/invitation/+state';
 import { downloadCsvFromJson } from '@blockframes/utils/helpers';
 import { toLabel } from '@blockframes/utils/pipes';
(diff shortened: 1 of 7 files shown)


libs/event/src/lib/pages/analytics/analytics.component.ts
```diff
@@ -48,30 +58,28 @@ export class AnalyticsComponent implements OnInit {
...
-            transformedAnalytic.watchTime = invitation?.watchTime ?? 0;
-          }
-
-          return transformedAnalytic;
+        const invitations = this.invitationQuery.getAll({
+          // we are looking for invitations related to this event
+          filterBy: invit => invit.eventId === event.id && !!invit.watchTime
+        });
+
+        const orgIds = Array.from(new Set(invitations.map(i => i.fromUser?.orgId || i.toUser?.orgId).filter(orgId => !!orgId)));
```
GrandSchtroumpf, line 68: Create intermediary variables :
```suggestion
        const allOrgIds = invitations.map(i => i.fromUser?.orgId || i.toUser?.orgId).filter(orgId => !!orgId);
        const orgIds = Array.from(new Set());
```

GrandSchtroumpf (approved):

Issue #7539, opened 2022-01-14
Save & Exit


2amarkina:
When a user has created an event / added modifications to it and clicks on the CTA "Save & Exit", he has to redirected to the Event page

Poppe-A:
The right behavior is to redirect where the user clicked before having the CTA "Save & Exit" displayed

PR #7575, opened 2022-01-19
Save exit

Poppe-A:
#7539


RemcoSimonides (commented):
Now the redirect doesn't wait for the event to be saved. I guess it should wait for that in case it is navigated to the event.

RemcoSimonides (approved):

Issue #7548, opened 2022-01-17
Google Analytics Events for Buyer Interaction


RemcoSimonides:
- [x] Add a Google Analytics Events for the 2 interactions

We would like to know how many users request a screening and for which movie
Also we would like to know how many users ask for asking price.

We should add google analytics event to start gathering this data

In these issues you can find the locations where the buttons have been added
- https://github.com/blockframes/blockframes/issues/7502
- https://github.com/blockframes/blockframes/issues/7503

_Originally posted by @cams-goussale in https://github.com/blockframes/blockframes/issues/7431#issuecomment-1012303367_

PR #7578, opened 2022-01-20
add new analytics in bigQuery

Poppe-A:
#7548


apps/backend-functions/src/pubsub/daily-analytics-import.ts
```diff
@@ -25,6 +25,10 @@ const queryMovieAnalytics = `
       (event_name = 'promoReelOpened' AND key = 'movieId')
       OR
       (event_name = 'addedToWishlist' AND key = 'movieId')
+      OR
+      (event_name = 'screeningRequested' AND key = 'movieId')
+      OR
+      (event_name = 'askingPriceRequested' AND key = 'movieId')
```
phpgeekfr, line 31: why do you do this if we don't do anything with it ?  in `importAnalytics`  for example

Issue #7502, opened 2022-01-04
[Interactions] Ask for a screening


cams-goussale:
This issue concerns the new interactions that we want to build between Buyer and Seller. It allows a Buyer to request to the Seller a screening session on a movie. We want these features to be released for Berlin festival in early Feb, so we have split the issue in 2 parts: the minimum viable product, and the bonus feature.

**MVP "Ask for a screening"**

Pages where they will be:

- [x] Movie page: [case 1](https://www.figma.com/file/c62IFnbBC2UFcHZFY861jV/%5B-New-Market-Interactions-%5D?node-id=3%3A7) / [case 2](https://www.figma.com/file/c62IFnbBC2UFcHZFY861jV/%5B-New-Market-Interactions-%5D?node-id=8%3A3467) / [case 3](https://www.figma.com/file/c62IFnbBC2UFcHZFY861jV/%5B-New-Market-Interactions-%5D?node-id=7%3A2352)
- [x] [Org page](https://www.figma.com/file/c62IFnbBC2UFcHZFY861jV/%5B-New-Market-Interactions-%5D?node-id=3%3A8)
- [x] [Screening closed session](https://www.figma.com/file/c62IFnbBC2UFcHZFY861jV/%5B-New-Market-Interactions-%5D?node-id=6%3A2)

Flow:

- [x] Disable button after click: [case 1](https://www.figma.com/file/c62IFnbBC2UFcHZFY861jV/%5B-New-Market-Interactions-%5D?node-id=28%3A6929)
- [x] [1 snackbar for Buyer](https://www.figma.com/file/c62IFnbBC2UFcHZFY861jV/%5B-New-Market-Interactions-%5D?node-id=27%3A7589)
- [x] [1 notification for Buyer](https://www.figma.com/file/c62IFnbBC2UFcHZFY861jV/%5B-New-Market-Interactions-%5D?node-id=49%3A11858)
- [x] [1 email for Seller](https://docs.google.com/spreadsheets/d/1Xv3HJjYEM-kTnACnu-uHCmlbQ_KXlXMMExMY2JoURHA/edit#gid=0&range=F6) (action: leads to Event form)
[...]

PR #7583, opened 2022-01-20
7502 ask for screening 2

RemcoSimonides:
#7502
Moved creating new event from `/c/o/dashboard/event` to `/c/o/dashboard/event/new/edit`

### DIFF
--- a/libs/event/src/lib/form/shell/shell.component.ts
+++ b/libs/event/src/lib/form/shell/shell.component.ts
@@ -12,6 +12,9 @@ import { Observable, of, Subscription } from 'rxjs';
 import { map, pluck, switchMap } from 'rxjs/operators';
 import { NavTabs, TabConfig } from '@blockframes/utils/event';
 import { MatSnackBar } from '@angular/material/snack-bar';
+import { createEvent, createScreening } from '@blockframes/event/+state';
+import { OrganizationQuery } from '@blockframes/organization/+state';
+import { addHours } from 'date-fns'
 
 const statisticsTab = { path: 'statistics', label: 'Statistics' };
 
@@ -53,6 +56,7 @@ export class EventFormShellComponent implements OnInit, OnDestroy {
     private routerQuery: RouterQuery,
     private cdr: ChangeDetectorRef,
     private snackBar: MatSnackBar,
+    private orgQuery: OrganizationQuery
   ) { }
 
   ngOnInit(): void {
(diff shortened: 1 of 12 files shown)


libs/event/src/lib/form/shell/shell.component.ts
```diff
@@ -61,7 +65,26 @@ export class EventFormShellComponent implements OnInit, OnDestroy {
     // @TODO #6780
     this.sub = eventId$.pipe(
       switchMap((eventId: string) => this.eventService.valueChanges(eventId))
-    ).subscribe(event => {
+    ).subscribe(async event => {
+
+      const { eventId } = this.route.snapshot.params;
+      const params = this.route.snapshot.queryParams;
+      if (eventId === 'new' && params) {
+        const { titleId } = params;
```
GrandSchtroumpf, line 73: I think queryParams is [always an object](https://angular.io/api/router/ActivatedRouteSnapshot#queryParams) so you can safely do :
```suggestion
      const { titleId } = this.route.snapshot.queryParams;
      if (eventId === 'new' && titleId) {
```

libs/event/src/lib/guard/event-edit.guard.ts
```diff
@@ -0,0 +1,26 @@
...
+    private router: Router,
+  ) { }
+
+  async canActivate(next: ActivatedRouteSnapshot) {
+    const eventId: string = next.params['eventId'];
+    if (eventId !== 'new') return true;
+    
+    const redirect = this.router.parseUrl(`/c/o/dashboard/event`);
+    const orgId = this.orgQuery.getActiveId();
+    const titleId: string = next.queryParams['titleId'];
```
GrandSchtroumpf, line 21: use the  `queryParamsMap` to ensure the type : 
```suggestion
    const titleId = next.queryParamsMap.get('titleId');
```

libs/event/src/lib/guard/event-edit.guard.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (!titleId) return redirect;
+    const title = await this.movie.getValue(titleId);
```
GrandSchtroumpf, line 23: Prefere `this.movie.valueChanges(titleId).pipe(take(1)).toPromise()` like that you can use leverage the cache

GrandSchtroumpf (approved):

PR #7585, opened 2022-01-20
remove Akita - organization - Feat 7284

phpgeekfr:
remove Akita - organization - Feat 7284


apps/crm/crm/src/app/crm/crm.module.ts
```diff
@@ -114,6 +115,7 @@ export const panelRoutes: Routes = [
     GetEventPipeModule,
     ClipboardModule,
     DetailedTermsModule,
+    UploadWidgetModule,
```
phpgeekfr, line 118: Why this did not created any issue earlier ??

libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -174,4 +176,10 @@ export class OrganizationService extends CollectionService<OrganizationState> {
 
     this.update(orgState.id, { wishlist });
   }
+
+  public isAddedToWishlist(movieId: string): Observable<boolean> {
```
GrandSchtroumpf, line 180: rename `isInWishlist`. I'm not 100% sure it's the best place for it though. But that's ok

libs/organization/src/lib/organization/guard/pending-organization.guard.ts
```diff
@@ -1,53 +1,48 @@
...
-    return this.authQuery.user$.pipe(
-      switchMap(user => {
-        if (!user) return of('/');
+  canActivate() {
+    return combineLatest([
+      this.authQuery.user$,
+      this.service.org$
+    ]).pipe(
+      map(([user, org]) => {
+        if (!user) {
```
GrandSchtroumpf, line 27: You don't need the braces :
```typescript
if (!user) return this.router.createUrlTree(['/']);
if (!user.orgId) return this.router.createUrlTree(['/auth/identity']);
if (!org) return this.router.createUrlTree(['/auth/identity']);
```

libs/organization/src/lib/organization/guard/request-access.guard.ts
```diff
@@ -1,59 +1,53 @@
...
-  sync() {
-    return this.authQuery.user$.pipe(
-      switchMap(user => {
+  canActivate() {
+    return combineLatest([
+      this.authQuery.user$,
+      this.service.org$
+    ]).pipe(
+      map(([user, org]) => {
         if (!user) {
```
GrandSchtroumpf, line 27: same as above

GrandSchtroumpf (approved):

Issue #7556, opened 2022-01-17
Release 3.7 - General UI


RemcoSimonides:
Issues fixed in this release so please double check
- https://github.com/blockframes/blockframes/issues/7424
- https://github.com/blockframes/blockframes/issues/7510

2amarkina:
OS: MacOS (BigSur)
Browser: Chrome 97.0.4692.71
App: Market 

- [ ] On the Buyer preferences modal it seems that we have different height between the fields @stephanie-cascade8 
![buyer preferences.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/d6313b3a-4000-4a60-bdc4-88a67778834f)

stephanie-cascade8:
**Popup "Request Asking Price"**

<img width="313" alt="Capture d’écran 2022-01-19 à 16 12 14" src="https://user
<img width="1217" alt="Capture d’écran 2022-01-19 à 16 12 08" src="https://user-images.githubusercontent.com/66258851/150159345-8d6de208-4908-4736-b65c-a5be89b6c917.png">
-images.githubusercontent.com/66258851/150159331-d1e1573a-1ee2-4353-93e9-788fbf405398.png">

2amarkina:
**Cross-app**
- [ ] if possible see what can be done to avoid the wishlist @stephanie-cascade8 
<img width="1398" alt="toolbar" src="https://user-images.githubusercontent.com/55499036/150159584-c86727fd-6324-4f64-9966-d5347f2504a1.png">
- [ ] Add a tooltip for an invitation icon in the toolbar with the text : "Invitations" 
- [ ] Add a tooltip for a notification icon in the toolbar with the text : "Notifications" 
- [ ] Add a tooltip for a wishlist icon in the toolbar with the text : "Wishlist" 

**Festival**
- [ ] Add a tooltip for a calendar icon in the toolbar with the text : "My Calendar" 

**Catalog**
- [ ] Add a tooltip for a selection icon in the toolbar with the text : "My Selection"

2amarkina:
> **Popup "Request Asking Price"**
> 
> <img alt="Capture d’écran 2022-01-19 à 16 12 08" width="1217" src="https://user-images.githubusercontent.com/66258851/150159345-8d6de208-4908-4736-b65c-a5be89b6c917.png">




- [ ] Adding to ☝️ remove the hint "type the text you want...."

stephanie-cascade8:
[...]

PR #7601, opened 2022-01-21
General UI fixes - 1

Poppe-A:
#7556


apps/festival/festival/src/app/event/view/view.component.html
```diff
@@ -17,9 +17,12 @@ <h6>
             <p class="mat-body-2">{{ event.meta.description }}</p>
             <countdown-timer [title]="'Event will start in :'" [date]="event.start" [timeUnits]="['days', 'hours', 'minutes']" (timerEnded)="timerEnded = $event"></countdown-timer>
             <ng-container *ngIf="event.isOwner; else notOwner">
-              <h3>You are the owner of this event.</h3>
-              <p>You can invite members and edit the event config in your dashboard.</p>
-              <a mat-stroked-button color="primary" [routerLink]="editEvent">Edit Event</a>
+              <h5>You are the owner of this event.</h5>
+              <p class="mat-caption">You can invite members and edit the event config in your dashboard.</p>
+              <a mat-button color="primary" [routerLink]="editEvent">
+                <mat-icon matListIcon svgIcon="edit"></mat-icon>
```
RemcoSimonides, line 23: What is `matListIcon`? we're not in a list, are we?

libs/movie/src/lib/movie/components/request-asking-price/request-asking-price.component.scss
```diff
@@ -13,4 +13,8 @@ form {
 mat-spinner {
   margin-left: 8px;
   display: inline-block;
+}
+
+footer :first-child {
```
RemcoSimonides, line 18: There is multiple footers? I think it's better to have a class name in case a footer is added before this element

libs/movie/src/lib/movie/components/upcoming-screenings/upcoming-screenings.component.scss
```diff
@@ -11,6 +11,14 @@
   margin-bottom: 48px;
 }
 
+:host > h2 {
+  margin-bottom: 12px;
+}
+
+:host > h2 + p {
+  margin-bottom: 32px;
+}
```
RemcoSimonides, line 20: There is probably an easier way to achieve what you want to do. Without digging through the elements but just go directly to them

Poppe-A: Maybe, but I couldn't find it, except adding a class name on this specific p tag, which is a very common element

RemcoSimonides: I don't understand. You're targeting an `h2`, right? There is only one `h2` on this component after your change.
Also there is only one `p` in this component

Issue #7558, opened 2022-01-17
Release 3.7 - Deal Negotation


RemcoSimonides:
Issues fixed in this release so please double check
- https://github.com/blockframes/blockframes/issues/7446
- https://github.com/blockframes/blockframes/issues/7482
- https://github.com/blockframes/blockframes/issues/7119
- https://github.com/blockframes/blockframes/issues/7389

2amarkina:
**OS:** MacOS (BigSur)
**Browser:** Safari, 15.0 / Chrome 97.0.4692.71

### Marketplace
- [ ] On the Homepage we can see a lot of titles, but when connecting to the Library we've got a message "Sorry, we didn't find anything." (same when using menu burger and CTA Library)
[Library](https://drive.google.com/file/d/13KHBxc7t5NRE6qIUSaTf1hbi0KTejYnd/view), [Library 2](https://drive.google.com/file/d/1FLmk52gGA46iTzOCL81DiqanjmZDFO2-/view)

_N.B. Putting the links to drive for the videos as my tool to make gifs doesn't work properly anymore_

- [ ] On the Offers and Deals page the table stops showing the content after we switch the tabs
[table tabs bug](https://drive.google.com/file/d/1roBQ3y1PJixi1ww1lWOLkJYhAZDUFYtj/view)

- [ ] Same for the filters. After entering the information that can't be filtered, and after cleaning it up, the content doesn't reappear
[table filter bug](https://drive.google.com/file/d/1SSP8GW12bjSO7D0Td6d6t1Rh5lbMT58X/view)

- [ ] When filtering by info that can be filtered, everything goes well. The bug concerns all the tables of that kind. 
[table filter bug repeat](url)

- [ ] When we go to from Library page (for example) to the Offers and Deals page, the scroll is blinking
[scroll blink bug](https://drive.google.com/file/d/1orhPqeHI1VRigylUm9RjJIDPJKFB15a_/view)

- [ ] FYI there was a moment that everything worked well on Chrome, but Safari was "frozen". It lasted for about 10min and then came back to normal. 
![safari.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/dbd590f2-94fe-4c23-b703-fcc476ec8a48)

- [ ] Need to put extra space between the titles in the header of the Selection page **Title 1, Title 2**
[...]

PR #7606, opened 2022-01-25
disable adding holdbacks when at least one holdback is present

yspringfield:
Fixed: https://github.com/blockframes/blockframes/issues/7558#issuecomment-1016625591

### DIFF
--- a/libs/contract/src/lib/contract/holdback/form/form.component.html
+++ b/libs/contract/src/lib/contract/holdback/form/form.component.html
@@ -91,7 +91,7 @@
   </bf-form-table>
 
   <footer class="actions" fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="16px">
-    <button mat-flat-button color="primary" (click)="save()">Add Holdbacks</button>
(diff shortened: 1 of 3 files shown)


libs/contract/src/lib/contract/holdback/form/form.component.html
```diff
@@ -91,7 +91,7 @@ <h3>Holdback Terms*</h3>
   </bf-form-table>
 
   <footer class="actions" fxLayout="column" fxLayoutAlign="center center" fxLayoutGap="16px">
-    <button mat-flat-button color="primary" (click)="save()">Add Holdbacks</button>
+    <button mat-flat-button color="primary" (click)="save()" [disabled]="!form.value.length || form.invalid">Add Holdbacks</button>
```
GrandSchtroumpf, line 94: put only `form.invalid` and put `minlength(1)` on the form validators

GrandSchtroumpf (changes requested):

GrandSchtroumpf (approved):

PR #7647, opened 2022-01-26
Fix spinner bug when no event

Poppe-A:
When a user has no event, the spinner was spinning endlessly


apps/festival/festival/src/app/marketplace/event/calendar/calendar.component.ts
```diff
@@ -42,7 +42,7 @@ export class EventCalendarComponent implements OnInit {
       filterBy: ({ type, status }) => type === 'attendEvent' && ['accepted', 'pending'].includes(status)
     }).pipe(
       map(invitations => invitations.map(i => i.eventId)),
-      map(eventIds => Array.from(new Set(eventIds))), // Remove duplicated
+      map(eventIds => Array.from(new Set(eventIds))), // Remove duplicated,
```
RemcoSimonides, line 45: Could you check your changes before you commit? Then you see can spot irregularities like this

RemcoSimonides (approved):

Issue #7550, opened 2022-01-17
Release 3.7 - Events


RemcoSimonides:
Issues fixed in this release so please double check
- [export to calendar #4045](https://github.com/blockframes/blockframes/issues/4045)
- [Links in notification #6235](https://github.com/blockframes/blockframes/issues/6235)

frenchyrules:
- [ ] Export button here only allows us to download .ics file, no google calendar version. Could we improve it to give us apple and google calendar too ?
![image.png](https://images.zenhubusercontent.com/611a265d2ea95c7880c91192/d4328900-53f4-4ab4-ad44-de413b54ff31)

- [ ] Clicking on the event in the calendar to open it stops working sometimes (refreshing the page seems to solve it) (might be resolved in[ issue:7502](https://app.zenhub.com/workspaces/blockframes-60b0a31838e485001b3a619f/issues/blockframes/blockframes/7502) by Remco)

- [ ] Had to wait a very long time for event page to load after clicking on “go to event page button”

- [ ] If possible, could we place this more to the right of the icon in order to prevent information below from being hidden ?
![image.png](https://images.zenhubusercontent.com/611a265d2ea95c7880c91192/efdaf7ae-6009-424d-ad26-8ec5fa012ab7)

frenchyrules:
- [ ] Can go to the event page and invite other people even though no title is selected, which leads us to an empty page with a text if clicked.
**Suggestion:** the screening is coming (with event date) and other screening info such as event title. Add an illustration to fill in the page a bit. @2amarkina   @Jackseed @cams-goussale 
![image.png](https://images.zenhubusercontent.com/611a265d2ea95c7880c91192/5b2f4e6c-f81f-4c16-b435-62a306705664)

frenchyrules:
Received a notification (email + in app) 1 day before the Event starts - wasn't able to do this @2amarkina

frenchyrules:
- [ ] When hovering disabled export button; tooltip “No Events to export” should appear
![image.png](https://images.zenhubusercontent.com/611a265d2ea95c7880c91192/2954994a-549e-4abf-abd6-8f7ff8e3aa30)

frenchyrules:
- [ ] Can we have the Event name as the download name on the file tag ?
![image.png](https://images.zenhubusercontent.com/611a265d2ea95c7880c91192/32ad51a2-e019-441f-a7b7-9841c03e539b)

Jackseed:
[...]

PR #7650, opened 2022-01-26
Fix button animation

Poppe-A:
https://github.com/blockframes/blockframes/issues/7550#issuecomment-1022088170


libs/event/src/lib/components/request-screening/request-screening.component.scss
```diff
@@ -1,4 +1,4 @@
-:host.animated button {
+:host.animated .pulse-button {
```
RemcoSimonides, line 1: Isn't it possible to do
```suggestion
  this.animated = false
```
in `libs/event/src/lib/components/request-screening/request-screening.component.ts`

RemcoSimonides (approved):

PR #7659, opened 2022-01-27
Merge back release 3 7 develop

phpgeekfr:


apps/festival/festival/src/app/event/view/view.component.ts
```diff
@@ -25,6 +26,7 @@ export class EventViewComponent implements OnInit {
   accessRoute: string;
   user$ = this.authQuery.user$;
   event$: Observable<Event>;
+  requestSent = new BehaviorStore(false);
```
GrandSchtroumpf, line 29: avoid BehaviorStore, prefer `cdr.markForCheck()` as it does the same job.

GrandSchtroumpf (approved):

apps/festival/festival/src/app/event/view/view.component.ts
phpgeekfr: created https://github.com/blockframes/blockframes/issues/7661

PR #7403, opened 2021-12-13
wip: function to diff two negotiations.

yspringfield:

PR #7690, opened 2022-01-31
functions to diff two negotiations

yspringfield:
Continuation of #7403


libs/contract/src/lib/negotiation/utils.ts
```diff
@@ -15,3 +104,16 @@ export function isInitial(negotiation: Partial<Negotiation>) {
 export function getReviewer(negotiation: Negotiation<Timestamp | Date>) {
   return negotiation.stakeholders.find(id => id !== negotiation.createdByOrg && id !== centralOrgId.catalog);
 }
+
+export function hydrateLanguageForEmail(data: Record<string, MovieLanguageSpecification>) {
```
GrandSchtroumpf, line 108: I think this function already exist. See with @phpgeekfr where it is to not repeat yourself

GrandSchtroumpf (approved):

PR #7695, opened 2022-02-01
removed userQuery #7286

phpgeekfr:
removed userQuery #7286


libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -1,43 +1,98 @@
...
+      };
+      this.profile = user;
+      this.uid = authState.uid;
+      return {
+        isAnonymous: authState.isAnonymous || false,
+        emailVerified: authState.emailVerified || false,
+        user
+      }
+    })
+  );
```
GrandSchtroumpf, line 75:
```suggestion
  auth$: Observable<{ isAnonymous: boolean, emailVerified: boolean, user?: User }> = this.authState$.pipe(
    switchMap(authState => {
      if (authState && !authState.isAnonymou) return of(undefined);
      return this.userService.valueChanges(authState.uid).pipe(map(profile => [profile, authState]));
    }),
    map(([user, authState]: [User, firebase.User]) => {
      if (!authState) return;
      // If needed: const { isAnonymous: false, emailVerified: false  } = this.authState;
      const { isAnonymous, emailVerified  } = this.authState;
      return { isAnonymous, emailVerified, user };
    }),
    tap(({ user } = {})=> {
      this.profile = user;
      this.uid = user?.id;
    })
  );
```

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -1,43 +1,98 @@
...
+
+  isBlockframesAdmin$ = this.authState$.pipe(
+    switchMap(async authState => {
+      if (!authState || authState.isAnonymous) return false;
+      const snap = await this.db.collection('blockframesAdmin').doc(authState.uid).get().toPromise();
+      return snap.exists;
```
GrandSchtroumpf, line 81:
```suggestion
      return this.db.collection('blockframesAdmin').doc(authState.uid).ref.get().then(snap => snap.exists);
```

libs/user/src/lib/auth/+state/auth.service.ts
```diff
@@ -178,6 +217,24 @@ export class AuthService extends FireAuthService<AuthState> {
     });
   }
 
+  // TODO #6113 once we have a custom email verified page, we can update the users' meta there
+  // #7303 if user does not interact with authService, this is not updated (ie: user goes directly to eventPage ?)
+  // @TODO #7286 #7273 rework this
+  private async updateEmailVerified() {
```
GrandSchtroumpf, line 223: Shouldn't this be in the backend ?

GrandSchtroumpf (approved):

PR #7711, opened 2022-02-04
Avails list tests

yspringfield:
Fixed: #7712

### DIFF
--- a/libs/contract/src/lib/avails/tests/avails-list.spec.ts
+++ b/libs/contract/src/lib/avails/tests/avails-list.spec.ts
@@ -1,71 +1,174 @@
-import { AvailsFilter, availableTitle, FullMandate } from '../avails';
+import { availableTitle } from '../avails';
 import {
-  availSouthKorea, availAfghanistan,
-  availFrance, availsSVODArgentina, availsPayTVArgentina,
-  availsGermany, availsBelgium, availsExistingEndedSales,
-  availsOngoingSales, availsTerritoryWithExclusivity, availsTerritoryWithoutExclusivity,
-  availsFranceLuxembourg, availsAllButSouthKorea,
+  availsListSouthKorea, availsListAfghanistan, availsListFrance, availsListGermanyRussiaCzech,
+  availsListOngoingSalesExclusive, availsListOngoingSalesNonExclusive, availsListGermanyRussiaCzechExclusive,
+  availsListArgentinaSVod, availsListArgentinaPayTV, availsListGermany, availsListCanada
 } from './../fixtures/availsFilters';
-import { assertDate } from './utils'
 import {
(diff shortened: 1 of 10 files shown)


libs/contract/src/lib/avails/tests/avails-list.spec.ts
```diff
@@ -1,61 +1,174 @@
+import { availableTitle } from '../avails';
+import {
+  availsListSouthKorea, availsListAfghanistan, availsListFrance, availsListGermanyRussiaCzech,
+  availsListOngoingSalesExclusive, availsListOngoingSalesNonExclusive, availsListGermanyRussiaCzechExclusive,
+  availsListArgentinaSVod, availsListArgentinaPayTV, availsListGermany, availsListCanada
+} from './../fixtures/availsFilters';
+import {
```
GrandSchtroumpf, line 7: I would rather do something like : 
```typescript
import * as sales from './../fixtures/sales';
import * as mandates from './../fixtures/mandates';
```
What do you think ?

libs/contract/src/lib/avails/tests/avails-list.spec.ts
```diff
@@ -1,61 +1,174 @@
...
+    const movie1Mandates = availableTitle(availsListSouthKorea, [mandateMovie1], movie1Sales);
+    const movie2Mandates = availableTitle(availsListSouthKorea, mandatesMovie2, movie2Sales);
+    const movie3Mandates = availableTitle(availsListSouthKorea, mandatesMovie3, movie3Sales);
+    const movie4Mandates = availableTitle(availsListSouthKorea, [mandateMovie4], movie4Sales);
+    const movie5Mandates = availableTitle(availsListSouthKorea, [mandateMovie5], movie5Sales);
+    expect(movie2Mandates.length).toBe(1);
+    expect(movie1Mandates.length).toBe(0);
+    expect(movie3Mandates.length).toBe(0);
+    expect(movie5Mandates.length).toBe(0);
+    expect(movie4Mandates.length).toBe(0);
```
GrandSchtroumpf, line 42: Would be nice to be able to factorize all that : 
```typescript
const availabilities = getAvailableTitle(availsListSouthKorea, [
  movie1,
  movie2,
]);
assertTitlesAvailabilities(availabilities, [1, 0, 0, 0 ]);
```
Or something like that

GrandSchtroumpf (approved):

libs/contract/src/lib/avails/tests/avails-list.spec.ts
yspringfield: The imports are used many times below the imports to create intermediary variables. It'll cause the code to be a lot more verbous.

GrandSchtroumpf (approved):

Issue #6987, opened 2021-10-28
Update fields / placeholders for filters


2amarkina:
_Can go directly to dev_

### Cross-app filters
**Filter name: Country of Origin**
Label: Type and select Country
Placeholder:  Country of Origin 

**Filter name: Language & Versions**
Label: Enter Languages
Placeholder:  Language

**Filter name: Company's Nationality**
Change the _select_ by _select+input_ as we use for Country of Origin 
Label: Type and select Country
Placeholder:  Country 

### Archipel Content
**Filter name: Avails**
Field: Exclusivity
Case: only if filter was cleared
Label: Select exclusivity

### Media Financiers
**Filter name: Partner**
Label: Search for a Partner

yspringfield:
location: https://staging.mediafinanciers.com/c/o/marketplace/title
In the partner dropdown, there's a typo ( additional s on partners) in the input placeholder text.
As can be seen in the image.
<img width="1106" alt="Bildschirmfoto 2022-02-16 um 12 53 00 PM" src="https://user-images.githubusercontent.com/22674720/154259972-bc15a547-8ca7-42b0-a68e-10dee52f1bfa.png">


Otherwise, everything looked good to me.

PR #7713, opened 2022-02-04
Update fields / placeholders for filters

Poppe-A:
#6987


libs/organization/src/lib/organization/forms/search.form.ts
```diff
@@ -63,12 +67,13 @@ export class OrganizationSearchForm extends FormEntity<OrganizationSearchControl
   search() {
     return this.organizationIndex.search<Organization>(this.query.value, {
       hitsPerPage: this.hitsPerPage.value,
+      query: this.query.value,
```
phpgeekfr, line 70: `this.query.value` is already line 68, do you need it ?

phpgeekfr (commented):
regarding #6987,  do you have implemented this also ? 
```
Archipel Content

Filter name: Avails
Field: Exclusivity
Case: only if filter was cleared
Label: Select exclusivity
Media Financiers

Filter name: Partner
Label: Search for a Partner
```

phpgeekfr (approved):
LGTM

Issue #7248, opened 2021-11-25
[Catalog] Holdback not correctly displayed


cams-goussale:
### Localization
- App name: Content
- Page: user card
- Url: c/o/marketplace/selection

### Context
When buyer is revising its selection (before creating the offer), holdbacks that you add to the title should be visible in a table. There is a bug when adding them.

### Steps to reproduce
1. Add any title to your card and go to My selection page
2. Open the "Add holdbacks" modal and create 1. Close the modal. The holdback you created doesn't appear in the table.
3. Open again the "Add holdbacks" modal and create a 2nd holdback. Close the modal. Now the first holdback appears in the table.
4. Open again the "Add holdbacks" modal and create a 3rd holdback. Close the modal. Now the first+second holdbacks appear in the table.
etc.

### Images
Modal
![image](https://user-images.githubusercontent.com/93206363/143442731-9b4e1b3b-92b2-4a74-870b-4743c0f52cb4.png)

Table
![image](https://user-images.githubusercontent.com/93206363/143442705-bd32312a-90ee-4fb6-8b3d-e05d804e69f8.png)

cams-goussale:
This issue has been partially solved during the release but not yet completely!

## What has been fixed during the release 🥇 

Now when adding the first holdback, you can only click on the `Add` button and the `Add holdback` is disabled. This works well.
<img width="508" alt="image" src="https://user-images.githubusercontent.com/93206363/151962911-23b01d81-1943-41ac-924f-90cfca02c8ea.png">

## What still needs to be solved 👨‍🚒

- [ ] If I have already created some holdbacks, and I click on Manage button to add more of them, I would see once again the 2 buttons `Add` and `Add holdbacks`. Clearly the 2nd one is buggy, so either we don't show it at all, or we disable it.

https://user-images.githubusercontent.com/93206363/151962798-9340101a-dc25-469e-b5bc-6e20bc1a6335.mp4

- [ ] If I delete all of the holdbacks I previously created, I can't validate/confirm my changes.

https://user-images.githubusercontent.com/93206363/151963750-65491cac-252f-4dd4-926c-c1352a123bd9.mp4

PR #7717, opened 2022-02-07
fixed: holdbacks correctly display now

yspringfield:
Fixed: #7248

### DIFF
--- a/libs/contract/src/lib/contract/holdback/form/form.component.ts
+++ b/libs/contract/src/lib/contract/holdback/form/form.component.ts
@@ -1,8 +1,8 @@
-import { Component, ChangeDetectionStrategy, EventEmitter, TemplateRef, Input, Output } from '@angular/core';
-import { Validators } from '@angular/forms';
+import { Component, ChangeDetectionStrategy, EventEmitter, TemplateRef, Input, Output, ViewChild } from '@angular/core';
 import { MatDialog, MatDialogRef } from '@angular/material/dialog';
 import { DetailedTermsComponent } from '@blockframes/contract/term/components/detailed/detailed.component';
 import { Movie } from '@blockframes/movie/+state';
+import { FormTableComponent } from '@blockframes/ui/form/table/form-table.component';
 import { FormList } from '@blockframes/utils/form';
 import { Scope } from '@blockframes/utils/static-model';
 import { Holdback } from '../../+state/contract.model';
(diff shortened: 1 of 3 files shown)


libs/contract/src/lib/contract/holdback/form/form.component.ts
```diff
@@ -18,6 +18,7 @@ export class HolbackFormComponent {
   @Input() title: Movie;
   @Input() holdbacks: Holdback[] = [];
   @Output() holdbacksChange = new EventEmitter<Holdback[]>();
+  @ViewChild('formTable', { read: FormTableComponent }) formTable;
```
RemcoSimonides, line 21: I think you can give it a type here `formTable: FormTableComponent`. Or is it already automatically typed?

libs/contract/src/lib/contract/holdback/form/form.component.ts
```diff
@@ -43,11 +44,15 @@ export class HolbackFormComponent {
...
   save() {
     if (this.form.valid) {
+      this.ref.close(); //The position of the close before the emit is important.
       this.holdbacksChange.emit(this.form.value);
-      this.ref.close();
     }
   }
 
+  shouldDisable() {
+    return this.formTable.formItem && this.formTable.formItem?.dirty;
```
RemcoSimonides, line 53: Only `return this.formTable.formItem?.dirty;` is ok

So now you can always add a holdback whenever it is dirty? Even if all data is empty?

libs/ui/src/lib/form/table/form-table.component.ts
```diff
@@ -62,7 +62,7 @@ export class FormTableComponent<T> implements OnInit, OnDestroy {
   pageSize = 5;
   /* We need to keep track of the current page since it will affect the index that we are working on */
   pageConfig: PageState = { pageIndex: 0, pageSize: 5 };
-  formItem: FormEntity<EntityControl<T>, T>;
+  public formItem: FormEntity<EntityControl<T>, T>;
```
RemcoSimonides, line 65: It's public by default if you don't define 'private' or 'protected'. So no need to add this

libs/contract/src/lib/contract/holdback/form/form.component.ts
yspringfield: It's not typed. Just typed it.

RemcoSimonides (approved):

PR #7730, opened 2022-02-09
Env var cleaning

phpgeekfr:
Env var cleaning


apps/crm/crm/src/polyfills.ts
```diff
@@ -73,11 +73,6 @@ import '@angular/localize/init';
   env: { DEBUG: undefined },
 };
 
-/***************************************************************************************************
```
phpgeekfr, line 76: is this used by other libs in meantime ?

GrandSchtroumpf (approved):

Issue #7321, opened 2021-12-03
[Catalog] Overlapping Mandates/Sales import validation


cams-goussale:
While doing some test, I noticed that I could create the same mandate on the same title for 2 different sellers. We should prevent this from happening. Rule:

For the same title, 2 or more Sellers should not be able to have a sale or a mandate with:
-at least 1 territory in common AND
-at least 1 media in common AND
-at least 1 day in common AND
-at least 1 contract with Exclusivity

PR #7734, opened 2022-02-10
7321 overlapping mandate imports

yspringfield:
Fixed: #7321


libs/import/src/lib/table-extracted-elements/contracts/contracts.component.ts
```diff
@@ -74,10 +79,11 @@ export class TableExtractedContractsComponent implements OnInit {
     try {
       const creations = this.selection.selected.filter(importState => importState.newContract && !hasImportErrors(importState));
       for (const contract of creations) {
-        this.processedContracts++;
-        await this.addContract(contract);
+        const success = this.addContract(contract, 'create');
+        if (success)
+          this.processedContracts++;
```
GrandSchtroumpf, line 84: use one line
```suggestion
        if (success) this.processedContracts++;
```

libs/import/src/lib/table-extracted-elements/contracts/contracts.component.ts
```diff
@@ -103,18 +109,50 @@ export class TableExtractedContractsComponent implements OnInit {
...
+        terms: (mandate: Mandate) => this.termService.valueChanges(mandate.termIds)
+      }, { shouldAwait: true }),
+      first()
+    ).toPromise() as FullMandate[];
+
+    return importState.terms.map(term => {
+      const availabilities = territoryAvailabilities(term, mandates, [])
+      //available territories indicates that there is an overlap with existing mandates in db.
+      return !!availabilities.available.length
+    }).some(isOverlapping => isOverlapping)
```
GrandSchtroumpf, line 129: put semi colons : 
```suggestion
      const availabilities = territoryAvailabilities(term, mandates, []);
      //available territories indicates that there is an overlap with existing mandates in db.
      return !!availabilities.available.length;
    }).some(isOverlapping => isOverlapping);
```

GrandSchtroumpf (approved):

Issue #7226, opened 2021-11-23
CRM - Add a contract list & view


Jackseed:
We have an offer list on the CRM but nothing to show an external contract. Contract list should be added after "Offers" and list all the contracts from the app. 

We should start with something as simple as we can: 
- use the same table columns as the contract list on dashboard
- use the same contract view as the one used in the crm, for internal contract within an offer

This shall be updated later based on our usage

PR #7755, opened 2022-02-14
Crm contract list

yspringfield:
Fixed: #7226
1. Externalize contract list from catalog to lib.
2. create contract-list shell in crm and fetch necessary data.
3. Configure proper routing to contract list.
4. use contract-list in both crm and catalog.


yspringfield:
here's how the sale-list component looks like both in the crm and catalog.
Bildschirmfoto 2022-02-15 um 4 29 21 PM
Bildschirmfoto 2022-02-15 um 4 29 31 PM
Bildschirmfoto 2022-02-15 um 4 31 07 PM

libs/contract/src/lib/contract/list/list.component.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /* index paramater is unused because it is a default paramater from the filter javascript function */
+  filterBySalesStatus(sale: Sale, index: number, status: ContractStatus): boolean {
+    if (!status) return true;
+    return sale.status === status;
+  }
+
+  getLicensorId(sale: Sale) {
+    return sale.stakeholders.find(
```
GrandSchtroumpf, line 107: same as above

GrandSchtroumpf (changes requested):

apps/crm/crm/src/app/crm/pages/external-sales/view/view.component.ts
```diff
@@ -0,0 +1,96 @@
...
+  public orgId = this.orgService.org.id;
+
+
+  contract$ = this.route.params.pipe(map(r => r.saleId as string))
+    .pipe(
+      switchMap(saleId => this.getSale(saleId)),
+      filter(contract => !!contract),
+      tap(contract => {
+        this.statusForm.setValue(contract.status);
+      })
```
GrandSchtroumpf, line 39: I think you can inline it.
```suggestion
      tap(contract => this.statusForm.setValue(contract.status))
```

libs/contract/src/lib/contract/list/external-sales/external-sales.component.html
```diff
@@ -0,0 +1,26 @@
+<ng-container *ngIf="sales as externalSales">
```
GrandSchtroumpf, line 1: You can gt rid of this line and on the second line do:
```html
<ng-container *ngIf="externalSales?.length; else empty">
  ...
</ng-container>
```

libs/contract/src/lib/contract/list/external-sales/external-sales.component.scss
```diff
@@ -0,0 +1,51 @@
+:host {
+  display: flex;
```
GrandSchtroumpf, line 2: I don't see the benefit of a flex here. A simple `display: block` would do the trick I think

libs/contract/src/lib/contract/list/external-sales/external-sales.component.html
```diff
@@ -0,0 +1,26 @@
...
+          <ng-template #noPrice>-</ng-template>
+        </ng-template>
+      </bf-table>
+    </ng-container>
+  </ng-container>
+</ng-container>
+
+
+<ng-template #empty>
+  <ng-content select="[empty]"></ng-content>
```
GrandSchtroumpf, line 25: If you've bot only one `ng-content` you don't need to use a select.

libs/contract/src/lib/contract/list/internal-sales/internal-sales.component.html
```diff
@@ -0,0 +1,52 @@
...
+        Ongoing <span *ngIf="(salesCount$ | async) as salesCount">({{ salesCount.negotiating }})</span>
+      </button>
+      <button mat-tab-link (click)="applyFilter('accepted')" [active]="(filter$ | async) === 'accepted'">
+        Accepted <span *ngIf="(salesCount$ | async) as salesCount">({{ salesCount.accepted }})</span>
+      </button>
+      <button mat-tab-link (click)="applyFilter('declined')" [active]="(filter$ | async) === 'declined'">
+        Declined <span *ngIf="(salesCount$ | async) as salesCount">({{ salesCount.declined }})</span>
+      </button>
+    </header>
+      <bf-table *ngIf="internalSales.length" [source]="internalSales | filterBy: filterBySalesStatus: filter.value"
```
GrandSchtroumpf, line 20: wrong tab

libs/contract/src/lib/contract/list/internal-sales/internal-sales.component.html
```diff
@@ -0,0 +1,52 @@
...
+            <span [matTooltip]="negoStage"> {{negoStage | maxLength:50}} </span>
+          </ng-container>
+        </ng-template>
+      </bf-table>
+  </ng-container>
+</ng-container>
+
+
+<ng-template #empty>
+  <ng-content select="[empty]"></ng-content>
```
GrandSchtroumpf, line 51: same as above

libs/contract/src/lib/contract/list/internal-sales/internal-sales.component.scss
```diff
@@ -0,0 +1,51 @@
+:host {
+  display: flex;
```
GrandSchtroumpf, line 2: same as above

libs/contract/src/lib/contract/list/internal-sales/internal-sales.component.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.dynTitle.setPageTitle(pageTitle);
+  }
+
+  resetFilter() {
+    this.filter.reset('');
+    this.dynTitle.setPageTitle(`${this.title} (All)`);
+  }
+
+  /* index paramater is unused because it is a default paramater from the filter javascript function */
+  filterBySalesStatus(sale: Sale, index: number, status: ContractStatus): boolean {
```
GrandSchtroumpf, line 90: This method doesn't rely on the context (`this`), so it can be outside of the class.

GrandSchtroumpf (approved):

libs/contract/src/lib/contract/list/internal-sales/internal-sales.component.ts
yspringfield: It is used within the template, so it needs to be declared in the class to be accessible there.

PR #7789, opened 2022-02-17
actionText for movie asking price

phpgeekfr:


libs/notification/src/lib/components/item/item.component.scss
```diff
@@ -2,3 +2,7 @@
 button {
   margin-right: 16px;
 }
+
+a[href^="mailto"] {
+  font-weight: inherit;
+}
```
RemcoSimonides, line 8: I understand you want to do this but it's overriding a global rule :/ isn't there a way to add this exception at the global level? For example if this link to email is in button which is on such popover menu?

RemcoSimonides (approved):

Issue #7756, opened 2022-02-14
Release 4.0 - Auth


RemcoSimonides:
[Auth test scenarios](https://www.notion.so/cascade8/Auth-a4eeff51e0d345ee83af6e211880c7fd)

Akita has been removed for Invitations, Notifications, Meeting (Twilio), Contract, Bucket, Offer, Income, Campaign, Organisation, Permissions, User & Auth. This means that anything can be broken. For example some parts of the page don't load or missing information on pages. Let's find what has been broken if anything! :)

cams-goussale:
- [ ] On /auth/reset-password I can try to reset password with a non existing email adress and I don't have any error. Snackbar should be `This account does not exist.` with the `CREATE ACCOUNT` button inside.

phpgeekfr:
- [ ] If I try to go to this URL without being logged in : https://staging.archipelcontent.com/c/o/marketplace/title?formValue=%7B%22search%22:%7B%22query%22:%22%22,%22genres%22:%5B%5D,%22originCountries%22:%5B%5D,%22contentType%22:null%7D,%22avails%22:%7B%22medias%22:null,%22exclusive%22:true,%22territories%22:%5B%5D,%22duration%22:%7B%22from%22:%222022-02-15T23:00:00.000Z%22,%22to%22:%222023-02-15T23:00:00.000Z%22%7D%7D%7D

After login, I'm redirected to 
https://staging.archipelcontent.com/c/o/marketplace/title%3FformValue%3D%257B%2522search%2522:%257B%2522query%2522:%2522%2522,%2522genres%2522:%255B%255D,%2522originCountries%2522:%255B%255D,%2522contentType%2522:null%257D,%2522avails%2522:%257B%2522medias%2522:null,%2522exclusive%2522:true,%2522territories%2522:%255B%255D,%2522duration%2522:%257B%2522from%2522:%25222022-02-15T23:00:00.000Z%2522,%2522to%2522:%25222023-02-15T23:00:00.000Z%2522%257D%257D%257D

that leads to a 404. Looks like the string is UrlEncoded twice

phpgeekfr:
- [ ] context:  I'm invited to an event and decided to create my account. I fill all the inputs (including invitation pass) and decide to create a new org but with an org name that already exists. I click on submit, my account is then created but then I have the error "org name already exists". I can change the org name but if I submit the form it will say that invitation pass is incorrect.
=> user creation should be done after "orgNameExists" check

[...]

Issue #6908, opened 2021-10-19
Auth page: how to get org list before having a real user


phpgeekfr:
During account creation, user is invited to choose an org name.
Currently we don't use any backend function, we perform a db query that goes through all orgs to find a matching one.

Temporary solution: use anonymous login (but all login-in users will be able to list all orgs (as prod is currently) )

Long term solution:
- Use a backend function (kept warm, and only accessible with at least anonymous login) to check if org name exists
- Use algolia directly  ( with a new index if needed ?)

PR #7796, opened 2022-02-18
7703 fix identity form no org

### LINKED ISSUE #7703: An user that have no org cannot fill the identity form
If I try to login with an user that have no orgId (and no invitation) but  lastName and FirstName already set, I'm redirected to `/auth/identity`

But when every field is filled the "Sign in" button stays disabled :

with, a not accurate message : `Please enter valid name and surname`

```
    if (this.form.invalid) {
      this.snackBar.open('Please enter valid name and surname', 'close', { duration: 2000 });
      return;
    }
```

 I think this is due to the fact that some inputs are disabled ( for example lastName is already filled and I'm not supposed to edit this)

This issue can occur on production but I don't really know in which condition an user can be without any org and any invitations unless user was removed from org or if invitation/request to join org was refused.

user example on staging : `078k6zjt65aBL3Wz6u9v23LJuFK2` `dev+anissa-dxj@blockframes.io`

So to reproduce, we must use an user with lastName and firstName but without any organization or invitation to joinOrg and I have the feeling that this can only happen on staging, because during anonymization process, even  user without any last org firstName will be filled with a random one or on production if  user was removed from org or if invitation/request to join org was refused.

This does not look like an urgent issue but we should try to remove an user from org and check if he can ask to join an existing or create a new one.
And maybe fix this error for the next release

RemcoSimonides:
Fix #7703
https://github.com/blockframes/blockframes/issues/7756#issuecomment-1041814085

### DIFF
--- a/libs/user/src/lib/auth/pages/identity/identity.component.ts
+++ b/libs/user/src/lib/auth/pages/identity/identity.component.ts
@@ -114,7 +114,7 @@ export class IdentityComponent implements OnInit, OnDestroy {
     // Fill fields
     this.form.patchValue(user);
 
-    this.disableControls(['email', 'firstName', 'lastName', 'confirm', 'generatedPassword']);
+    this.disableControls(['email', 'firstName', 'lastName', 'password', 'confirm', 'generatedPassword']);
   }  
 
   private disableControls(keys: (keyof IdentityFormControl)[]) {
--- a/libs/user/src/lib/auth/pages/identity/identity.component.ts
+++ b/libs/user/src/lib/auth/pages/identity/identity.component.ts
@@ -114,13 +114,14 @@ export class IdentityComponent implements OnInit, OnDestroy {
     // Fill fields
     this.form.patchValue(user);
 
-    // Disable/hide what is not needed
-    this.form.get('email').disable();
-    this.form.get('firstName').disable();
-    this.form.get('lastName').disable();
-    this.form.get('password').disable();
-    this.form.get('confirm').disable();
-    this.form.get('generatedPassword').disable();
+    this.disableControls(['email', 'firstName', 'lastName', 'confirm', 'generatedPassword']);
+  }  
+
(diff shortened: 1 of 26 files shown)


libs/user/src/lib/auth/pages/identity/identity.component.ts
```diff
@@ -109,7 +109,8 @@ export class IdentityComponent implements OnInit, OnDestroy {
     return this.intercom.show();
   }
 
-  private updateFormForExistingIdentity(user: Partial<PublicUser>) {
+  private updateFormForExistingIdentity(profile: Partial<User>) {
+    const user = createPublicUser(profile);
```
RemcoSimonides, line 113: Removes `privacyPolicy: { ip: '', ... }` from the object.
This ticked the privacy policy box already on the form. It shouldn't be ticked already, but that has to do with this issue open https://github.com/blockframes/blockframes/issues/6908

libs/utils/src/lib/form/validators/validators.ts
```diff
@@ -43,7 +43,7 @@ export function confirmPasswords(
 /** Require current and new password to be different */
 export function differentPassword(current: string = 'current', next: string = 'next'): ValidatorFn {
   return (group: FormGroup): { [key: string]: boolean } | null => {
-    return group.controls[current].value === group.controls[next].value
+    return group.controls[current].value === group.controls[next].value && !group.controls[current].disabled && !group.controls[next].disabled
```
GrandSchtroumpf, line 46: it's not readable, please create intermediary variables : 
```suggestion
    const currentControl = group.controls[current];
    const nextControl = group.controls[next];
    if (currentControl !== nextControl.value) return null;
    if (currentControl.disabled && nextControl.disabled) return null;
    return { currentPasswordMatch: true };
```

GrandSchtroumpf (approved):

Issue #7766, opened 2022-02-14
Release 4.0 - Avails


RemcoSimonides:
[Avails test scenarios](https://www.notion.so/cascade8/Avails-3e975555a25545fb8d66ae498725680d)

Issues fixed during release
- [x] https://github.com/blockframes/blockframes/issues/7248
- [x] https://github.com/blockframes/blockframes/issues/7310
- [x] https://github.com/blockframes/blockframes/issues/7639
- [x] https://github.com/blockframes/blockframes/issues/7644
- [x] https://github.com/blockframes/blockframes/issues/7646

Akita has been removed for Invitations, Notifications, Meeting (Twilio), Contract, Bucket, Offer, Income, Campaign, Organisation, Permissions, User & Auth. This means that anything can be broken. For example some parts of the page don't load or missing information on pages. Let's find what has been broken if anything! :)

cams-goussale:
- [ ] On /c/o/dashboard/avails the .csv file downloaded has a format error. The first row doesn't have the text qualifier `"` but the second row does. As a consequence, the file is hard to read when openning it in excel. We should have the text qualifier `"` in both rows.

<img width="674" alt="image" src="https://user-images.githubusercontent.com/93206363/154232389-a1c25f02-9989-415f-aea5-64a024520fcd.png">


<img width="424" alt="image" src="https://user-images.githubusercontent.com/93206363/154232322-fc089efa-cd0d-4d72-887f-a69656edeb90.png">

cams-goussale:
- [ ] In /c/o/marketplace/selection if I add some Holdbacks to a title, leave the page to add more titles, and then come back to my selection, the holdbacks are not saved (sorry there is a jump on the video).
- [ ] Can we change the text of the button `Add` to `Save Holback`

https://user-images.githubusercontent.com/93206363/154235598-981bd640-1428-4e5c-9513-660a9c3c8b91.mp4

phpgeekfr:
- [ ] If I try to access calendar page directly, this spinner is not disappearing example: https://staging.archipelcontent.com/c/o/marketplace/title/Iw75A7cmq9g9EWkURTbO/avails/calendar

![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/336ee900-0abc-4fe7-aa88-f578d35190f7)

phpgeekfr:
- [ ] opening the .csv with LibreOffice says that a cell have exceeded the maximum number of characters
[...]

PR #7801, opened 2022-02-18
group territories when exporting avails csv

yspringfield:
Fixed: https://github.com/blockframes/blockframes/issues/7766#issuecomment-1041468576


libs/utils/src/lib/utils.ts
```diff
@@ -232,6 +232,28 @@ export function toLabel(value: string | string[], scope: Scope, joinWith?: strin
   }
 }
 
+export function toGroupLabel(value: string[], scope: GroupScope, all?: string) {
```
RemcoSimonides, line 235: You can keep this function in the group-label.pipe.ts, but then outside of the Pipe class. This is what we do quite often if the 'pipe' needs to be used outside of the html.

RemcoSimonides (approved):

Issue #7759, opened 2022-02-14
Release 4.0 - Movie


RemcoSimonides:
[Movie test scenarios](https://www.notion.so/cascade8/Movie-8506fba333d64b8cb9e857351a6c7989)

Issues resolved for this release
- [x] https://github.com/blockframes/blockframes/issues/6987
- [x] https://github.com/blockframes/blockframes/issues/7219

Akita has been removed for Invitations, Notifications, Meeting (Twilio), Contract, Bucket, Offer, Income, Campaign, Organisation, Permissions, User & Auth. This means that anything can be broken. For example some parts of the page don't load or missing information on pages. Let's find what has been broken if anything! :)

yspringfield:
location: https://staging.mediafinanciers.com/c/o/marketplace/title
In the partner dropdown, there's a typo ( additional s on partners) in the input placeholder text.
As can be seen in the image.
<img width="1106" alt="Bildschirmfoto 2022-02-16 um 12 53 00 PM" src="https://user-images.githubusercontent.com/22674720/154260144-53ac0303-ae35-4557-b412-0eef01a05892.png">

Otherwise, everything looked good to me.

RemcoSimonides:
Archipel Market
- [ ] Movie Page -> Header -> Animated buttons "request asking price" "ask screening" should have the same height.
<img width="412" alt="Capture d’écran 2022-02-16 à 13 32 13" src="https://user-images.githubusercontent.com/66258851/154265310-177a49cc-c1ca-4ac8-99d7-93380e2833a2.png">
<img width="334" alt="Capture d’écran 2022-02-16 à 13 32 22" src="https://user-images.githubusercontent.com/66258851/154265320-a4e14dcf-0e37-4844-8b3f-65b5e3237d88.png">

- [ ] Ongoing Screening button's height is wrong. 
(All day ongoing screening here :[ Boiling Point](https://staging.archipelmarket.com/c/o/marketplace/title/rkT1PdqZbPXaTqvdpEPw/main)

Now 
<img width="1424" alt="Capture d’écran 2022-02-16 à 13 39 05" src="https://user-images.githubusercontent.com/66258851/154266568-beebd978-6e2b-40ba-b352-f3cf69655d44.png">

After trying "height: fit-content"
[...]

PR #7816, opened 2022-02-22
7759 fix disappearing sales pitch description on tab change

RemcoSimonides:
#7759

### DIFF
--- a/libs/media/src/lib/file/file-uploader/file-uploader.component.ts
+++ b/libs/media/src/lib/file/file-uploader/file-uploader.component.ts
@@ -26,6 +26,7 @@ import { BehaviorSubject, Subscription } from 'rxjs';
 import { AngularFirestore } from '@angular/fire/firestore';
 import { getDeepValue } from '@blockframes/utils/pipes';
 import { boolean } from '@blockframes/utils/decorators/decorators';
+import { StorageFile } from '@blockframes/media/+state/media.firestore';
 
 type UploadState = 'waiting' | 'hovering' | 'ready' | 'file';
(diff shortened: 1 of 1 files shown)


libs/media/src/lib/file/file-uploader/file-uploader.component.ts
```diff
@@ -114,7 +114,10 @@ export class FileUploaderComponent implements OnInit, OnDestroy {
           ? getDeepValue(data, this.metadata.field)[this.formIndex]
           : getDeepValue(data, this.metadata.field);
         if (media) {
-          this.form.patchValue(media);
+          const extra = this.getExtra()
```
phpgeekfr, line 117: in 
```
private getExtra() {
    const extraKeys = Object.keys(this.form.value).filter(key => !['privacy', 'collection', 'docId', 'field', 'storagePath'].includes(key));
    if (extraKeys.length) {
      const extra = {};
      for (const key of extraKeys) {
        extra[key] = this.form.value[key];
      }
      return extra;
    }
  }
 ```

can you create an string array from properties of function `createStorageFileControl` instead of `['privacy', 'collection', 'docId', 'field', 'storagePath']`?

 and type it correctly so you can do `delete extra?.jwPlayerId;`

phpgeekfr (approved):

libs/media/src/lib/file/file-uploader/file-uploader.component.ts
phpgeekfr, line 117: missing `;`

PR #7819, opened 2022-02-22
7759 only enable save button when changes made

RemcoSimonides:
#7759

### DIFF
--- a/libs/movie/src/lib/movie/form/reviews/reviews.component.html
+++ b/libs/movie/src/lib/movie/form/reviews/reviews.component.html
@@ -135,7 +135,7 @@
 
             <mat-form-field appearance="outline">
               <mat-label>Article Link</mat-label>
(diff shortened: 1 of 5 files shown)


libs/movie/src/lib/movie/form/reviews/reviews.component.html
```diff
@@ -135,7 +135,7 @@ <h3>Film Reviews</h3>
 
             <mat-form-field appearance="outline">
               <mat-label>Article Link</mat-label>
-              <input test-id="link" type="url" matInput formControlName="revueLink" />
+              <input test-id="link" type="url" matInput formControlName="revueLink" placeholder="https://www.article.com" />
```
phpgeekfr, line 138: http://example.com/ is not better ?

phpgeekfr (approved):

Issue #7815, opened 2022-02-22
Contract import


yspringfield:
Main issues
1. #7225
2. #6661
3. https://github.com/blockframes/blockframes/issues/6661#issuecomment-1027096108
4. #5267

### Common with admin and seller imports. ###
1. Not sure if it's not already the case but to be sure: users (and admin) shouldn't be enabled to attach a contract to another org than theirs, but they should be able to add an org name as Licensor of a contract to view it on the contract view, taking this into account too. #7810 
2. Imported contracts shouldn't trigger notifications nor emails ("You received an offer")
3.  Add a status column as mandatory on the template
4. Territories: it’s really hard to fill territories like “World excluding France”, currently we have to copy paste a list of all the countries and remove France from it. It’s extremely hard to read afterwards and prompt to errors. A while ago we had 2 columns: “Territories included” (eg “World”) and “Territories excluded” (eg “France”); if we could get back to something similar it would be really precious.
5. It would be also really useful to be able to fill in region names like “Africa” or “Europe” and that every territories from that region are included. Also have `Media` too in mind when doing this. 
6. We are having more and more often multiple terms attached to one contract. It looks like the admin columns (purple ones N-P) are not working for these cases. A solution would be to adapt the excel to this logic: several lines for each term, and one merged cell representing the contract of these terms (similar to what we have on the movie form with the cast)

### Specific to the seller. ###
1.  Add a price column as facultative on the template (this may wait if too complex)

### Specific to the admin. ###
1. International Title, Licensor & Licensee name currently accept names, best would be to have possibility to fill it with ids as well (for admin purposes). This is important when we have the same title name for different movies.

cams-goussale:
Notes for Camille:

- Column R won't be useful anymore. Can be deleted
- Update parameters (add the green ones and remove the yellow ones). Add the media scope categories (VOD, TV, Ancillary Rights, DVD)
- Status column not mandatory
- TBC with Springfield if Stakeholders column? to keep or not?

yspringfield:
[here's](https://drive.google.com/file/d/10O8KwdS0DpnWoFkeqCm-WCQEfTcIxzsi/view?usp=sharing) the final template used

PR #7833, opened 2022-02-23
updated import to handle new file structure.

yspringfield:
Fixed: #7815

### DIFF
--- a/apps/crm/crm/src/app/crm/pages/external-sales/list/list.component.ts
+++ b/apps/crm/crm/src/app/crm/pages/external-sales/list/list.component.ts
@@ -12,13 +12,7 @@ import { of } from 'rxjs';
 import { map } from 'rxjs/operators';
 import {getSeller} from '@blockframes/contract/contract/+state/utils'
 
-function queryFn(ref: CollectionReference, options: { internal?: boolean }) {
-  if (options.internal)
-    return ref
-      .where('buyerId', '!=', '')
-      .where('type', '==', 'sale')
-      .orderBy('buyerId', 'desc')
-      .orderBy('_meta.createdAt', 'desc')
+function queryFn(ref: CollectionReference) {
   return ref
     .where('buyerId', '==', '')
     .where('type', '==', 'sale')
@@ -42,7 +36,7 @@ export class ContractsListComponent {
   public orgId = this.orgService.org.id;
 
 
-  public externalSales$ = this.contractService.valueChanges(ref => queryFn(ref, { internal: false })).pipe(
+  public externalSales$ = this.contractService.valueChanges(queryFn).pipe(
     joinWith({
       licensor: (sale: Sale) => {
         return this.orgService.valueChanges(getSeller(sale)).pipe(map(getFullName))
--- a/libs/import/src/lib/view-extracted-elements/contract/utils.ts
+++ b/libs/import/src/lib/view-extracted-elements/contract/utils.ts
@@ -39,9 +39,14 @@ interface FieldsConfig {
     buyerId: string;
     id?: string;
     stakeholders: string[];
+    status: string,
   };
+  income: {
+    price: number;
+  },
   term: {
-    territories: Territory[];
+    territories_included: Territory[];
+    territories_excluded: Territory[];
     medias: Media[];
     exclusive: boolean;
     duration: {
@@ -52,7 +57,7 @@ interface FieldsConfig {
     dubbed: Language[];
     subtitle: Language[];
     caption: Language[];
-  };
+  }[];
   parentTerm: string | number;
   _titleId?: string;
 }
@@ -60,13 +65,14 @@ interface FieldsConfig {
 type FieldsConfigType = ExtractConfig<FieldsConfig>;
 
 
-function toTerm(rawTerm: FieldsConfig['term'], contractId: string, firestore: AngularFirestore): Term {
+function toTerm(rawTerm: FieldsConfig['term'][number], contractId: string, firestore: AngularFirestore): Term {
 
-  const { medias, duration, territories, exclusive, licensedOriginal } = rawTerm;
+  const { medias, duration, territories_excluded, territories_included, exclusive, licensedOriginal } = rawTerm;
 
   const languages: Term['languages'] = {};
 
   const updateLanguage = (key: keyof MovieLanguageSpecification, rawLanguages: Language[]) => {
+    console.log({ key, rawLanguages, rawTerm })
     for (const language of rawLanguages) {
       if (!languages[language]) languages[language] = { caption: false, dubbed: false, subtitle: false };
       languages[language][key] = true;
@@ -85,7 +91,7 @@ function toTerm(rawTerm: FieldsConfig['term'], contractId: string, firestore: An
     contractId,
     medias,
     duration,
-    territories,
+    territories: territories_included,
     exclusive,
     licensedOriginal,
     criteria: [],
@@ -149,6 +155,7 @@ export async function formatContract(
         };
         return centralOrgId.catalog;
       } else {
+        //@todo: implement check to verify the movie who's mandate it is, belongs to the licensor org.
         const sellerId = await getOrgId(value, orgService, orgNameCache);
         if (!sellerId) return unknownEntityError('Licensor Organization');
         if (!blockframesAdmin && sellerId !== userOrgId) return {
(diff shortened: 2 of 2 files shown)


apps/crm/crm/src/app/crm/pages/external-sales/list/list.component.ts
```diff
@@ -12,13 +12,7 @@ import { of } from 'rxjs';
...
 import {getSeller} from '@blockframes/contract/contract/+state/utils'
 
-function queryFn(ref: CollectionReference, options: { internal?: boolean }) {
-  if (options.internal)
-    return ref
-      .where('buyerId', '!=', '')
-      .where('type', '==', 'sale')
-      .orderBy('buyerId', 'desc')
-      .orderBy('_meta.createdAt', 'desc')
+function queryFn(ref: CollectionReference) {
```
GrandSchtroumpf, line 15: Just so that it looks more like a callback : 
```typescript
const queryFn = (ref: CollectionReference) => ref
  .where(...)
```

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -183,41 +189,47 @@ export async function formatContract(
         return sellerId;
       }
     },
-    /* e */'term.territories': (value: string) => getStaticList('territories', value, separator, 'Territories', true, 'world') as Territory[],
-    /* f */'term.medias': (value: string) =>  getStaticList('medias', value, separator, 'Medias') as Media[],
-    /* g */'term.exclusive': (value: string) => {
+    /* e */'term[].territories_included': (value: string) => {
+      const result = getTerritoryList(value, separator) as Territory[];
+      return result;
```
GrandSchtroumpf, line 194: just return directly
```suggestion
      return getTerritoryList(value, separator) as Territory[];
```

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -177,47 +199,51 @@ export async function formatContract(
...
       const lower = value.toLowerCase();
       if (!lower) return mandatoryError('Licensed Original');
       if (lower !== 'yes' && lower !== 'no') return wrongValueError('Licensed Original');
       return lower === 'yes';
     },
-    /* k */'term.dubbed': (value: string) => getStaticList('languages', value, separator, 'Dubbed', false) as Language[],
-    /* l */'term.subtitle': (value: string) => getStaticList('languages', value, separator, 'Subtitle', false) as Language[],
-    /* m */'term.caption': (value: string) => getStaticList('languages', value, separator, 'CC', false) as Language[],
+    /* l */'contract.status': (value: string = 'pending') => value.toLowerCase(),
+    /* m */'term[].dubbed': (value: string) => getStaticList('languages', value, separator, 'Dubbed', false) as Language[],
```
GrandSchtroumpf, line 235: As we discussed, `getStaticList` should be able to know that the output is a Language[] if we put `languages` as first parameter

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -177,47 +199,51 @@ export async function formatContract(
...
+        let buyerId = await getOrgId(value, orgService, orgNameCache);
+        if (!buyerId && await verifyIdExists(value, titleService)) buyerId = value;
+        if (isInternal && !buyerId) return unknownEntityError('Licensee Organization');
+        return buyerId;
       }
     },
-    /* e */'term.territories': (value: string) => getStaticList('territories', value, separator, 'Territories', true, 'world') as Territory[],
-    /* f */'term.medias': (value: string) =>  getStaticList('medias', value, separator, 'Medias') as Media[],
-    /* g */'term.exclusive': (value: string) => {
+    /* e */'term[].territories_included': (value: string) => getGroupedList(value, 'territories', separator) as Territory[],
```
GrandSchtroumpf, line 210: As we discussed, getGroupedList should be able to know that the output is a Territory[] if we put territories as first parameter

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -177,47 +199,51 @@ export async function formatContract(
...
       } else {
+        if (!value) return '';
         const isInternal = data.contract.sellerId === centralOrgId.catalog;
-        const sellerId = await getOrgId(value, orgService, orgNameCache);
-        if (isInternal && !sellerId) return unknownEntityError('Licensee Organization');
-        return sellerId;
+        let buyerId = await getOrgId(value, orgService, orgNameCache);
+        if (!buyerId && await verifyIdExists(value, titleService)) buyerId = value;
+        if (isInternal && !buyerId) return unknownEntityError('Licensee Organization');
+        return buyerId;
```
GrandSchtroumpf, line 207: return as soon as you can : 
```suggestion
        const buyerId = await getOrgId(value, orgService, orgNameCache);
        if (buyerId) return buyerId;
        if (await verifyIdExists(value, titleService)) return value;
        if (isInternal) return unknownEntityError('Licensee Organization');
        return ''; // maybe return an error if it's mandatory ?
```

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -40,7 +41,7 @@ export class MovieService extends CollectionService<MovieState> {
     private analyticservice: AnalyticsService,
     private orgService: OrganizationService,
   ) {
-    super();
+    super(); this.db
```
GrandSchtroumpf, line 44: 🤔 I don't understand why there is a `this.db` here.

libs/import/src/lib/utils.ts
```diff
@@ -7,7 +7,11 @@ import { SheetTab, ValueWithError } from "@blockframes/utils/spreadsheet";
 import { centralOrgId } from "@env";
 import { ContractService } from "@blockframes/contract/contract/+state/contract.service";
 import { UserService } from "@blockframes/user/+state";
+import { CollectionService } from "akita-ng-fire";
 
+export abstract class CheckDocExists {
+  abstract docExists(docId: string);
+}
```
GrandSchtroumpf, line 14: If you don't have any methods or properties, use an interface instead :
```suggestion
export interface CheckDocExists {
  docExists(docId: string): Promise<boolean>;
}
```

libs/import/src/lib/utils.ts
```diff
@@ -92,6 +96,10 @@ export async function getTitleId(name: string, titleService: MovieService, cache
   return result;
 }
 
+export async function verifyIdExists(id: string, service: CheckDocExists) {
+  return service.docExists(id);
+}
```
GrandSchtroumpf, line 101: I don't really see the benefice of this function. If you've got the service then just run the method, you don't need to wrap it with this overlay.

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -241,7 +242,28 @@ export function getStaticList(scope: Scope, value: string, separator:string, nam
...
-export function split(cell: string, separator:string) {
+export function getGroupedList<GS extends GroupScope>(value: string, groupScope: GS, separator: string) {
+  const elements = split(value, separator);
+  const groupLabels = staticGroups[groupScope].map(group => group.label);
+  let allElements = elements.map(element => {
+    if (groupLabels.includes(element)) {
+      const groups = staticGroups[groupScope] as StaticGroup<GS>[]
+      return groups.find(group => group.label === element).items;
+    }
+    return [element];
```
GrandSchtroumpf, line 253: as `groupScope` is typed you should be able to do : 
```suggestion
    return groupLabels.includes(element)
      ? staticGroups[groupScope].find(group => group.label === element).items
      : element;
```

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -241,7 +242,28 @@ export function getStaticList(scope: Scope, value: string, separator:string, nam
...
+  }).flat()
+
+  allElements = Array.from(new Set(allElements));
+  const elementList = allElements.join(separator);
```
GrandSchtroumpf, line 257: Avoid using `let` when possible. For example doing that would avoid the need of `let` above.
```suggestion
  const elementList = Array.from(new Set(allElements)).join(separator);
```

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -241,7 +242,28 @@ export function getStaticList(scope: Scope, value: string, separator:string, nam
...
+
+  const values = {
```
GrandSchtroumpf, line 259: This should be defined outside of the function

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -149,24 +161,28 @@ export async function formatContract(
         };
         return centralOrgId.catalog;
       } else {
-        const sellerId = await getOrgId(value, orgService, orgNameCache);
-        if (!sellerId) return unknownEntityError('Licensor Organization');
+        let sellerId = await getOrgId(value, orgService, orgNameCache);
+        if (!sellerId) {
+          const seller = await orgService.getValue(value);
+          if (seller) sellerId = value;
+          else return unknownEntityError('Licensor Organization');
```
GrandSchtroumpf, line 168: Prefer the opposite : 
```suggestion
          if (!seller) return unknownEntityError('Licensor Organization');
          sellerId = value;
```
Avoid inlined `else` statement

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -251,48 +273,45 @@ export async function formatContract(
...
-
-
+  const results = await extract<FieldsConfig>(sheetTab.rows, fieldsConfig, 11);
   for (const result of results) {
     const { data, errors } = result;
 
-    const contract =  data.contract.type === 'sale'
-      ? createSale({ ...data.contract as Sale})
+    const contract = data.contract.type === 'sale'
+      ? createSale({ ...data.contract as Sale })
```
GrandSchtroumpf, line 300: remove the spread operator : 
```typescript
createSale(data.contract as Sale)
```

libs/utils/src/lib/helpers.ts
```diff
@@ -94,6 +94,10 @@ export async function asyncFilter<T>(items: T[], filterFunction: (item: T) => Pr
   return y.filter(w => w !== _null) as T[];
 }
 
+function findCorrespondence<S extends Scope>(sanitizedCode: string) {
+  return ([key, value]: [string, string]) => [key.toLowerCase(), value.toLowerCase()].includes(sanitizedCode)
+}
```
GrandSchtroumpf, line 99: If you've got only two test, use ||. Also you don't use the Scope
```suggestion
function findCorrespondence(code: string) {
  return ([key, value]: [string, string]) => {
    return key.toLowerCase() === code || value.toLowerCase() === code;
   }
}
```

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -211,37 +220,60 @@ export async function extract<T>(rawRows: string[][], config: ExtractConfig<T> =
...
   if (value.toLowerCase() === allKey) return parseToAll(scope, allKey);
   const splitted = split(value, separator);
-  const keys = splitted.map(v => getKeyIfExists(scope, v));
+  const keys = splitted.map(v => getKeyIfExists(scope, `${v}`));
   const values = keys.filter(v => !!v);
   const wrongData: string[] = [];
   keys.forEach((k, i) => {
-    if (!k) wrongData.push(splitted[i]);
+    if (!k) wrongData.push(`${splitted[i]}`);
   });
```
GrandSchtroumpf, line 232: Use filter : 
```typescript
const values = keys.filter(v => !!v);
if (values.length === keys.length) return values;
const wrongData = splitted.filter(v => !v);
return {...}
```

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -211,37 +220,60 @@ export async function extract<T>(rawRows: string[][], config: ExtractConfig<T> =
...
+      hint: `${wrongData.slice(0, 3).join(', ')}...`
+    }
+  };
+  return values as GetKeys<S>[];
 }
 
-export function getStaticList(scope: Scope, value: string, separator:string, name: string, mandatory = true, allKey = 'all') {
+const isValueError = <S extends Scope>(values: FromStatic<S>): values is ImportError<S> => {
+  return ('length' in values && values.length === 0)
+    || ('value' in values && values.value.length === 0)
```
GrandSchtroumpf, line 246:
```suggestion
  return values.length === 0 || values.value?.length === 0;
```

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -211,37 +220,60 @@ export async function extract<T>(rawRows: string[][], config: ExtractConfig<T> =
...
-      ('length' in values && values.length === 0) ||
-      ('value' in values && values.value.length === 0)
-    )
-  ) return mandatoryError(name);
+  if (mandatory && isValueError(values)) return mandatoryError<GetKeys<S>>(name);
   return values;
 }
 
-export function split(cell: string, separator:string) {
+const groupedTerritoriesOrMedias = {
```
GrandSchtroumpf, line 255: rename `fromGroup`

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -211,37 +220,60 @@ export async function extract<T>(rawRows: string[][], config: ExtractConfig<T> =
...
+  territories: (territories, separator) => getStaticList('territories', territories, separator, 'Territories', true, 'world'),
+  medias: (medias, separator) => getStaticList('medias', medias, separator, 'Medias', true, 'all'),
+}
+
+export function getGroupedList(value: string, groupScope: 'territories', separator: string): GetKeys<'territories'>[];
+export function getGroupedList(value: string, groupScope: 'medias', separator: string): GetKeys<'medias'>[];
+export function getGroupedList<GS extends GroupScope>(value: string, groupScope: GS, separator: string) {
```
GrandSchtroumpf, line 262: type is wrong. It can also be an error cannot it ?

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -1,13 +1,22 @@
...
 import { WorkBook, WorkSheet, utils, read } from 'xlsx';
-
+import { GetKeys, GroupScope, StaticGroup, staticGroups } from '@blockframes/utils/static-model';
 import { mandatoryError, SpreadsheetImportError } from 'libs/import/src/lib/utils';
-
 import { getKeyIfExists } from '../helpers';
 import { parseToAll, Scope } from '../static-model';
 
 type Matrix = any[][]; // @todo find better type
+interface ImportError<S extends Scope> {
```
GrandSchtroumpf, line 9: isn't it the same as `ValueWithError` ?

GrandSchtroumpf (approved):
I approved because it takes too much time, but you should be careful with wrong typing.

PR #7859, opened 2022-02-24
7738 use uid instead of userId

RemcoSimonides:
#7738
Use uid instead of userId to match existing data in Google Analytics
Fix bug in import


yspringfield (approved):
LGTM

PR #7880, opened 2022-02-28
7855 current analytics support new data model

### LINKED ISSUE #7855: Movie-analytics-chart component has to support new data model
`libs/movie/src/lib/movie/components/movie-analytics-chart/movie-analytics-chart.component.ts`

RemcoSimonides:
Fix #7855


apps/festival/festival/src/app/dashboard/title/activity/activity.component.ts
```diff
@@ -1,9 +1,11 @@
 import { Component, ChangeDetectionStrategy, OnInit } from '@angular/core';
-import { AnalyticsService } from '@blockframes/utils/analytics/analytics.service';
-import { pluck, switchMap } from 'rxjs/operators';
+import { AnalyticsService } from '@blockframes/analytics/+state/analytics.service';
+import { map, pluck, switchMap, tap } from 'rxjs/operators';
```
yspringfield, line 3: tap not used

yspringfield (approved):

libs/movie/src/lib/movie/components/movie-analytics-chart/movie-analytics-chart.component.ts
```diff
@@ -44,7 +44,7 @@ function getLastDays(from: number, to: number = 0) {
 }
 
 /** Format date: Date -> YYYYMMDD */
-function toYMD(date: Date) {
+export function toYMD(date: Date) {
```
phpgeekfr, line 47: can you find a better place to this function maybe ?

libs/movie/src/lib/movie/components/movie-analytics-chart/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+    promoReelOpened: {
+      current: currentAnalytics.filter(analytics => analytics.event_name === 'promoReelOpened'),
+      past: pastAnalytics.filter(analytics => analytics.event_name === 'promoReelOpened')
+    },
+    type: 'movie',
+    id: ''
+  };
+
+  return [movieAnalytics];
```
phpgeekfr, line 35: maybe you can convert it to array elsewhere ?

libs/movie/src/lib/movie/components/movie-analytics-chart/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+      counter[event.name][date] = movieEvent;
+    }
+  });
+
+  const eventAnalytics = [];
+  for (const data of Object.values(counter)) {
+    for (const analytics of Object.values(data)) {
+      eventAnalytics.push(analytics);
+    };
+  };
```
phpgeekfr, line 68: only a suggestion but maybe some `map` `reduce` can prevent usage of imbricated for loops

firestore.rules
```diff
@@ -499,9 +499,14 @@ service cloud.firestore {
     match /analytics/{docId} {
       allow read: if userHasValidOrg()
         && notInMaintenance()
-        && ( resource == null
-          || (existingData().type == 'movie' && orgCan('canUpdate', userOrgId(), docId))
+        && (resource == null
+          || ( existingData().type == 'title' && (userOrgId() in existingData().meta.ownerOrgIds ))
```
phpgeekfr, line 503: as long as your queries always specifie a where clause with `.meta.ownerOrgIds` this should work.

firestore.rules
```diff
@@ -499,9 +499,14 @@ service cloud.firestore {
...
         );
+       allow create: if userHasValidOrg()
+        && notInMaintenance()
+        && isRequestFieldValueEqualToOrNotSetted(['id'], docId)
+        && isRequestFieldValueEqualTo(['meta', 'orgId'], userOrgId())
+        && isRequestFieldValueEqualTo(['meta', 'uid'], userId());
```
phpgeekfr, line 509: `ownerOrgIds` is set during migration but not when a new document is created ? 
Users will be able to create this kind of documents ? It's not something created only by backend ?

firestore.rules
GrandSchtroumpf, line 503: this is very difficult to read. Make it more readable using functions.

firestore.rules
```diff
@@ -499,9 +499,14 @@ service cloud.firestore {
...
       allow read: if userHasValidOrg()
         && notInMaintenance()
-        && ( resource == null
-          || (existingData().type == 'movie' && orgCan('canUpdate', userOrgId(), docId))
+        && (resource == null
+          || ( existingData().type == 'title' && (userOrgId() in existingData().meta.ownerOrgIds ))
         );
+       allow create: if userHasValidOrg()
+        && notInMaintenance()
+        && isRequestFieldValueEqualToOrNotSetted(['id'], docId)
```
GrandSchtroumpf, line 507: why id can be undefined ?

firestore.rules
GrandSchtroumpf, line 509: I think there is an issue with the definition of the function, we should be able to read the FieldPath directly instead of breaking it into an array limited to 3 values.
Please check if we can check the value with a FieldPath instead: https://firebase.google.com/docs/rules

libs/analytics/src/lib/+state/analytics.service.ts
```diff
@@ -1,17 +1,22 @@
...
+export class AnalyticsService extends CollectionService<AnalyticsState> {
   readonly useMemorization = true;
 
   constructor() {
-    super();    
+    super();
   }
 
+  formatFromFirestore(analytic): Analytics<AnalyticsTypes> {
+    analytic._meta.createdAt = toDate(analytic._meta.createdAt);
```
GrandSchtroumpf, line 19: I think there is a `metaToDate` function or something like that so you dont' have to do all the fields.

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -14,18 +13,18 @@ import { App } from '@blockframes/utils/apps';
 import { QueryFn } from '@angular/fire/firestore';
 import { OrganizationService } from '@blockframes/organization/+state';
 import { map } from 'rxjs/operators';
-import { getViews } from '../pipes/analytics.pipe';
 import { joinWith } from '@blockframes/utils/operators';
-import { AnalyticsService } from '@blockframes/utils/analytics/analytics.service';
+import { AnalyticsService } from '@blockframes/analytics/+state/analytics.service';
```
GrandSchtroumpf, line 17:
```suggestion
import { AnalyticsService, Analytics } from '@blockframes/analytics/+state';
```

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -14,18 +13,18 @@ import { App } from '@blockframes/utils/apps';
...
 import { AuthService } from '@blockframes/auth/+state';
 import { ActiveState, EntityState } from '@datorama/akita';
+import { Analytics } from '@blockframes/analytics/+state/analytics.firestore';
```
GrandSchtroumpf, line 20:
```suggestion
```

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -14,18 +13,18 @@ import { App } from '@blockframes/utils/apps';
...
 
 export const fromOrg = (orgId: string): QueryFn => ref => ref.where('orgIds', 'array-contains', orgId);
 export const fromOrgAndAccepted = (orgId: string, appli: App): QueryFn => ref => ref.where(`app.${appli}.status`, '==', 'accepted').where('orgIds', 'array-contains', orgId);
 export const fromOrgAndInternalRef = (orgId: string, internalRef: string): QueryFn => ref => ref.where('orgIds', 'array-contains', orgId).where('internalRef', '==', internalRef);
 export const fromInternalRef = (internalRef: string): QueryFn => ref => ref.where('internalRef', '==', internalRef);
 
-type MovieWithAnalytics = Movie & { analytics: MovieAnalytics };
+type MovieWithAnalytics = Movie & { analytics: Analytics<'title'>[] };
```
GrandSchtroumpf, line 27: Use interface when extending another interface.
```suggestion
interface MovieWithAnalytics extends Movie {
  analytics: Analytics<'title'>[];
}
```

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -111,11 +110,16 @@ export class MovieService extends CollectionService<MovieState> {
   queryDashboard(app: App) {
     const orgId = this.orgService.org.id;
     const query: QueryFn = ref => ref.where('orgIds', 'array-contains', orgId).where(`app.${app}.access`, '==', true);
-    const addViews = (movie: MovieWithAnalytics) => ({ ...movie, analytics: { ...movie.analytics, views: getViews(movie.analytics) } });
+    const addViews = (movie: MovieWithAnalytics) => ({ ...movie, analytics: { ...movie.analytics, views: movie.analytics?.length } });
```
GrandSchtroumpf, line 113: You don't need this function at all. Just change `views` to `length` everywhere you're using it.

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -111,11 +110,16 @@ export class MovieService extends CollectionService<MovieState> {
...
 
     return this.valueChanges(query).pipe(
       joinWith({
-        analytics: movie => this.analyticservice.valueChanges(movie.id),
+        analytics: movie => this.analyticservice.valueChanges(ref => ref
```
GrandSchtroumpf, line 117: use camelCase: `analyticService`

libs/movie/src/lib/movie/components/movie-analytics-chart/utils.ts
```diff
@@ -0,0 +1,71 @@
+import { Analytics } from "@blockframes/analytics/+state/analytics.firestore";
+import { MovieAnalytics, MovieEventAnalytics } from "@blockframes/movie/+state/movie.firestore";
+import { AnalyticsEvents } from "@blockframes/utils/analytics/analytics-model";
+import { addDays } from 'date-fns'
+import { toYMD } from "./movie-analytics-chart.component";
+
+export function toMovieAnalytics(events: Analytics<'title'>[]): MovieAnalytics[] {
+  const today = new Date();
+  const current = addDays(today, -28);
+  const past = addDays(today, -56);
```
GrandSchtroumpf, line 10: Prefere `subDays` instead of `addDays`, it's more readable: 
```suggestion
  const current = subDays(today, 28);
  const past = subDays(today, 56);
```

libs/movie/src/lib/movie/components/movie-analytics-chart/utils.ts
```diff
@@ -0,0 +1,71 @@
+import { Analytics } from "@blockframes/analytics/+state/analytics.firestore";
+import { MovieAnalytics, MovieEventAnalytics } from "@blockframes/movie/+state/movie.firestore";
+import { AnalyticsEvents } from "@blockframes/utils/analytics/analytics-model";
+import { addDays } from 'date-fns'
+import { toYMD } from "./movie-analytics-chart.component";
+
+export function toMovieAnalytics(events: Analytics<'title'>[]): MovieAnalytics[] {
```
GrandSchtroumpf, line 7: you might want to call the property `analytics` instead of `event` to avoid confusion.

libs/movie/src/lib/movie/components/movie-analytics-chart/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+  const current = addDays(today, -28);
+  const past = addDays(today, -56);
+  
+  const currentEvents = events.filter(event => event._meta.createdAt > current);
+  const pastEvents = events.filter(event => event._meta.createdAt > past && event._meta.createdAt < current);
+
+  const currentAnalytics = toMovieEventAnalytics(currentEvents);
+  const pastAnalytics = toMovieEventAnalytics(pastEvents);
+
+  const movieAnalytics: MovieAnalytics = {
```
GrandSchtroumpf, line 18: Maybe you can factorize with something like this :
```typescript
const analyticsFilter = (name: string) => ({
  current: currentAnalytics.filter(analytics => analytics.event_name === name),
  past: pastAnalytics.filter(analytics => analytics.event_name === name)
})
```

libs/movie/src/lib/movie/components/movie-analytics-chart/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+  const counter: Record<AnalyticsEvents, Record<string, MovieEventAnalytics>> = {
+    addedToWishlist: {},
+    askingPriceRequested: {},
+    pageView: {},
+    promoReelOpened: {},
+    removedFromWishlist: {},
+    screeningRequested: {}
+  };
+
+  events.forEach(event => {
```
GrandSchtroumpf, line 48: use `forEach` only if : 
- You can to the operation in one line
- You need the index
Here you can use a `for` loop

libs/movie/src/lib/movie/components/movie-analytics-chart/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+      counter[event.name][date].hits++;
+    } else {
+      const movieEvent: MovieEventAnalytics = {
+        event_date: date,
+        event_name: event.name as AnalyticsEvents,
+        hits: 1,
+        movieId: event.meta.titleId
+      };
+      counter[event.name][date] = movieEvent;
+    }
```
GrandSchtroumpf, line 60:
```suggestion
    if (counter[event.name][date]) {
      counter[event.name][date].hits++;
      continue;
    }
    const movieEvent: MovieEventAnalytics = {
      event_date: date,
      event_name: event.name as AnalyticsEvents,
      hits: 1,
      movieId: event.meta.titleId
    };
    counter[event.name][date] = movieEvent;
```

libs/movie/src/lib/movie/components/movie-analytics-chart/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  const eventAnalytics = [];
+  for (const data of Object.values(counter)) {
+    for (const analytics of Object.values(data)) {
+      eventAnalytics.push(analytics);
+    };
+  };
+
+  return eventAnalytics;
```
GrandSchtroumpf, line 70: You can also do:
```suggestion
  return Object.values(counter).map(analytics => Object.values(analytics)).flat();
```

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -47,6 +52,14 @@ export class HomeComponent implements OnInit {
       map(titles => titles.filter(title => title.app[this.app].access))
     );
 
+    this.titleAnalytics$ = this.orgService.currentOrg$.pipe(
```
GrandSchtroumpf, line 55: Any reason to put in ngOnInit and not in the constructor ? If not, move it up.

apps/catalog/catalog/src/app/dashboard/title/activity/activity.component.ts
```diff
@@ -14,14 +16,20 @@ import { DashboardTitleShellComponent } from '@blockframes/movie/dashboard/shell
 export class TitleActivityComponent implements OnInit {
   public movieAnalytics$ = this.route.params.pipe(
     pluck('movieId'),
-    switchMap((movieId: string) => this.analyticsService.valueChanges([movieId]))
+    switchMap((movieId: string) => this.analyticsService.valueChanges(ref => ref
+      .where('type', '==', 'title')
+      .where('meta.titleId', '==', movieId)
+      .where('meta.ownerOrgIds', 'array-contains', this.orgService.org?.id)
```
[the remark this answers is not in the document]
RemcoSimonides: It will crash indeed but there is no way its undefined.

firestore.rules
RemcoSimonides: `ownerOrgIds` is set when a new document is created and yes, users are able to create this kind of document. 
Whenever they trigger an event in the platform, an analytics document is added to Firestore.

libs/movie/src/lib/movie/components/movie-analytics-chart/movie-analytics-chart.component.ts
RemcoSimonides: I have a better idea; use `format` from `date-fns` instead

libs/movie/src/lib/movie/components/movie-analytics-chart/utils.ts
RemcoSimonides: Now I can do `map(toMovieAnalytics)` which is convenient ^^

firestore.rules
RemcoSimonides: Created #7889 for this

firestore.rules
```diff
@@ -499,11 +499,19 @@ service cloud.firestore {
     match /analytics/{docId} {
       allow read: if userHasValidOrg()
         && notInMaintenance()
-        && ( resource == null
-          || (existingData().type == 'movie' && orgCan('canUpdate', userOrgId(), docId))
+        && (resource == null
+          || isAllowedToReadTitleAnalytics()
         );
```
GrandSchtroumpf, line 504: Put `resource == null` inside the `isAllowedToReadTitleAnalytics` function
```suggestion
        && isAllowedToReadTitleAnalytics();
```

firestore.rules
```diff
@@ -499,11 +499,19 @@ service cloud.firestore {
...
+       allow create: if userHasValidOrg()
+        && notInMaintenance()
+        && isRequestFieldValueEqualTo(['id'], docId)
+        && isRequestFieldValueEqualTo(['meta', 'orgId'], userOrgId())
+        && isRequestFieldValueEqualTo(['meta', 'uid'], userId());
     }
 
+  function isAllowedToReadTitleAnalytics() {
+    return existingData().type == 'title' && (userOrgId() in existingData().meta.ownerOrgIds )
```
GrandSchtroumpf, line 513: you can create variable in the function now to make things readable.

libs/analytics/src/lib/+state/analytics.service.ts
```diff
@@ -1,17 +1,35 @@
...
   }
 
+  formatFromFirestore(analytic): Analytics<AnalyticsTypes> {
+    return {
+      ...analytic,
+      _meta: formatDocumentMetaFromFirestore(analytic._meta)
+    };
+  }
+
+  getTitleAnalytics$(titleId: string) {
```
GrandSchtroumpf, line 28: remove $ in the method. We use it only for the properties

firestore.rules
RemcoSimonides: I was thinking in case we have more types in the future to do:
``` resource == null
 || isAllowedToReadTitleAnalytics()
 || isAllowedToReadEventAnalytics()
 || etc
```
This way we only need to define `resource == null` once

firestore.rules
RemcoSimonides: How would you do it?
Like
```
    let data = existingData();
    return data.type == 'title' && (userOrgId() in data.meta.ownerOrgIds );
```
?

PR #7890, opened 2022-03-01
Feat 7287 router ter

phpgeekfr:


libs/utils/src/lib/pipes/app.pipe.ts
```diff
@@ -1,10 +1,10 @@
 import { Pipe, PipeTransform, NgModule } from '@angular/core';
-import { getAppName, App } from '../apps';
+import { App, appName, appShortName } from '../apps';
 
 @Pipe({ name: 'appName' })
 export class AppNamePipe implements PipeTransform {
   transform(app: App, short = false): string {
-    return getAppName(app, short).label;
+    return short ? appShortName[app] : appName[app];
```
GrandSchtroumpf, line 7: Do you use the `short` ? A better design is with naming :
```typescript
transform(app: App, mode: 'short' | 'long' = 'long'): string {
    return mode === 'short' ? appShortName[app] : appName[app];
}
```
Like that it's more readable on the template: `{{ app | appName:'short' }}`

GrandSchtroumpf (approved):

Issue #7273, opened 2021-11-29
Remove Akita


RemcoSimonides:
After all entities have been removed, the config has to be removed
`apps\catalog\catalog\src\main.ts`

phpgeekfr:
check `@TODO #7273`

PR #7894, opened 2022-03-02
7737 trigger events

### LINKED ISSUE #7737: [Sellers' Analytics] Trigger pageView event on Title visit
- [ ] Page view only counts once per session. Session should stop when closing the website, or when the day ends. Don't end session if page is refreshed.
- [ ] Do not trigger for concierge users

Note: we currently send a pageView event to GA. This is not a unique trigger per session like we want to have. However, we don't change the behavior to GA because this enables business team to help user better.

RemcoSimonides:
Fix #7737 
- [x] Trigger pageView events only once a day
- [x] Do not trigger for concierge users
- [x] Also trigger the other events and removed old analytics service


libs/analytics/src/lib/+state/analytics.service.ts
```diff
@@ -26,10 +36,86 @@ export class AnalyticsService extends CollectionService<AnalyticsState> {
...
+        titleId: id,
+        orgId: profile.orgId,
+        uid: profile.uid,
+        ownerOrgIds: title.orgIds
+      });
+    }
+    if (!meta) return;
+
+    return this.add({
+      id: key,
```
RemcoSimonides, line 102: Adding the pageView event to Firestore with a key which is unique per day. It overwrites the previous document if it already exists.
Now that I think about it; is it better to check if the doc already exists? That would only cost a read instead of a write. Although I'd guess most of the time the localStorage check is already enough.

libs/analytics/src/lib/+state/analytics.service.ts
```diff
@@ -26,10 +36,86 @@ export class AnalyticsService extends CollectionService<AnalyticsState> {
...
+      .where('meta.ownerOrgIds', 'array-contains', orgId)
     );
   }
+
+  async addTitleEvent(name: EventName, titleId: string) {
+    if (await this.isOperator()) return;
+
+    const profile = this.authService.profile;
+
+    // // TODO use MovieService instead once Akita has been replaced by ng-fire (currently using MovieService results in error)
```
phpgeekfr, line 52: please add a reference to #7273

libs/analytics/src/lib/+state/analytics.service.ts
```diff
@@ -26,10 +36,86 @@ export class AnalyticsService extends CollectionService<AnalyticsState> {
...
+  /**
+   * @dev We do not want to log centralOrg operators nor blockframes
+   * admins nor concierge users actions on the platform.
+   */
+  private async isOperator(): Promise<boolean> {
+    const isBlockframesAdmin = await this.authService.isBlockframesAdmin$.pipe(take(1)).toPromise();
+    const profile = this.authService.profile;
+    const isConcierge = profile?.email.includes('concierge');
+    const isOperator = isBlockframesAdmin || isConcierge || Object.values(centralOrgId).includes(profile?.orgId);
+    return isOperator;
```
GrandSchtroumpf, line 119: just return 
```suggestion
    return isBlockframesAdmin || isConcierge || Object.values(centralOrgId).includes(profile?.orgId);
```

libs/analytics/src/lib/+state/analytics.service.ts
```diff
@@ -26,10 +36,86 @@ export class AnalyticsService extends CollectionService<AnalyticsState> {
...
+    this.analytics.logEvent(name, meta);
+    return this.add({
+      type: 'title',
+      name,
+      meta,
+      _meta: createDocumentMeta({ createdFrom: this.appGuard.currentApp, createdBy: profile.uid })
+    });
+  }
+
+  async addPageViewEvent(type: AnalyticsTypes, id: string) {
```
GrandSchtroumpf, line 72: rename `addPage`.
The overall process is very complex, you can simplify it by listening on the last analytics doc and check if the date is today.

apps/festival/festival/src/app/marketplace/title/view/view.component.ts
```diff
@@ -19,6 +20,8 @@ import { ActivatedRoute } from '@angular/router';
 export class MarketplaceMovieViewComponent  {
   public movie$ = this.route.params.pipe(
     pluck('movieId'),
+    distinctUntilChanged(),
```
GrandSchtroumpf, line 23: it would be good to know why the params are triggered multiple times

RemcoSimonides: It is triggered multiple times because there are multiple subscribers to movie$

GrandSchtroumpf: use `shareReplay` then : 
```typescript
public movie$ = this.route.params.pipe(
  pluck('movieId'),
  tap((movieId: string) => this.analytics.addPageView('title', movieId)),
  switchMap((movieId: string) => this.movieService.getValue(movieId)),
  shareReplay({ refCount: true, bufferSize: 1 })
);
```

libs/analytics/src/lib/+state/analytics.service.ts
```diff
@@ -25,11 +39,86 @@ export class AnalyticsService extends CollectionService<AnalyticsState> {
...
+
+    let meta: AnalyticsTypeRecord[AnalyticsTypes];
+    if (type === 'title') {
+      const start = startOfDay(new Date());
+      const analytics = await this.getValue(ref => ref
+        .where('_meta.createdBy', '==', profile.uid)
+        .where('meta.titleId', '==', id)
+        .where('name', '==', 'pageView')
+      );
+      if (analytics.some(analytic => analytic._meta.createdAt > start)) return;
```
GrandSchtroumpf, line 96: maybe add a comment to explain

GrandSchtroumpf (approved):

Issue #7654, opened 2022-01-27
[Bridges] Dev issue - Landing Animation


cams-goussale:
### Description

To complete the revival of Archipel Content, the last piece of the puzzle is to create bridges between Archipel Market and Archipel Content. If you want more descriptions, you can refer to the design issue here #6875

### What's new

A new header on each landing page to present the "Archipel Ecosystem":

[UX Screen](https://www.figma.com/file/GY8lt9tI3nx5xmeMawCtxQ/%E2%9A%99%EF%B8%8FBridge-AC-%2B-AM-(DONE)?node-id=82%3A9787)
UI Screen: [Archipel Content](https://www.figma.com/file/PGobBB7H317PZO326rCLcz/Landing?node-id=100%3A1342) | [Archipel Market](https://www.figma.com/file/PGobBB7H317PZO326rCLcz/Landing?node-id=100%3A1210)
[Animation](https://github.com/blockframes/blockframes/issues/6875#issuecomment-1013026831)

PR #7895, opened 2022-03-02
Landing page app bridge banner

Poppe-A:
#7654


libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.html
```diff
@@ -0,0 +1,18 @@
+<h2>Welcome to the Archipel Ecosystem</h2>
+<p class="mat-subheading">The online playground for global sellers and buyers</p>
+<div class="button-container" fxLayoutAlign="center" fxLayoutGap="32px">
```
RemcoSimonides, line 3: `section` instead of `div`

Does `fxLayoutAlign` and `fxLayoutGap` work without `fxLayout`? Does it take row by default?

RemcoSimonides: No need to give class name `button-container`. Just target `section` directly

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.html
```diff
@@ -0,0 +1,18 @@
...
+  <button mat-ripple fxLayout fxLayoutGap="16px" (click)="scrollToHeader()">
```
RemcoSimonides, line 4: I think this should be a simple `<a>` with an href to the id of the element. Then you don't need the scrollToHeader function either.

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.html
```diff
@@ -0,0 +1,18 @@
...
+    <mat-icon svgIcon={{buttonsData[currentApp].logo}}></mat-icon>
```
RemcoSimonides, line 5: Use `[svgIcon]`="buttonsData[currentApp].logo" instead of `svgIcon={{}}`
And it happens more often in this PR so please fix them all

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.html
RemcoSimonides: Also I see a lot of `fxLayout` used for this button. I think it's cleaner to use `display: grid` in css. Then you don't need the `div` element in a button either.

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.scss
```diff
@@ -0,0 +1,43 @@
...
+
+  button {
+    background-color: rgba(30, 27, 66, 1);
+  }
+  
+  a {
+    background-color: rgba(50, 53, 104, 1);
+  }
+  
+  > * {
```
RemcoSimonides, line 22: This css logic is still magic to me. Is it really needed to do it this way again? It's not possible to just target a specific element?

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.scss
```diff
@@ -0,0 +1,43 @@
...
+    cursor: pointer;
+    border: none;
+    padding: 16px 32px;
+    border-radius: 4px;
+  
+    div {
+      :first-child {
+        font-weight: bold;
```
RemcoSimonides, line 30: So `mat-body-2` isn't doing what you want yet. Isn't there a better class instead?

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.scss
```diff
@@ -0,0 +1,43 @@
...
+      }
+      p {
+        margin: 0;
```
RemcoSimonides, line 33: I think `span` doesn't have margin.

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  selector: 'app-bridge-banner',
+  templateUrl: './app-bridge-banner.component.html',
+  styleUrls: ['./app-bridge-banner.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+
+export class AppBridgeBannerComponent {
+  @HostBinding('class') class = 'dark-contrast-theme';
+
+  public currentApp = this.appGuard.currentApp;
```
RemcoSimonides, line 23: This has been changed. Search for `@Inject(APP)` instead and you'll see how it's done now

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.html
Poppe-A: Yes it works without fxLayout. I prefere not write it, so the line is shorter

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.html
Poppe-A: I thought about it, but with an `<a>` I can't have a smooth scroll. The transition is direct

apps/catalog/catalog/src/app/landing/landing.component.html
```diff
@@ -1,6 +1,8 @@
 <landing-shell-page>
+  <app-bridge-banner></app-bridge-banner>
+
   <landing-header>
-    <header class="max-width" fxLayout="column" fxLayoutAlign="start start" fxLayoutAlign.lt-sm="center center" fxLayoutGap="32px">
+    <header id="headerToScroll" class="max-width" fxLayout="column" fxLayoutAlign="start start" fxLayoutAlign.lt-sm="center center" fxLayoutGap="32px">
```
GrandSchtroumpf, line 5: This header is getting unreadable. Move all the `fxLayout**` code into css with `display: flex` and all.
name id to `main-header`

apps/festival/festival/src/app/landing/landing.component.html
```diff
@@ -1,7 +1,9 @@
 <landing-shell-page>
 
+  <app-bridge-banner></app-bridge-banner>
+
   <landing-header>
-    <header class="max-width" fxLayout="column" fxLayoutAlign="start start" fxLayoutAlign.lt-sm="center center" fxLayoutGap="32px">
+    <header id="headerToScroll" class="max-width" fxLayout="column" fxLayoutAlign="start start" fxLayoutAlign.lt-sm="center center" fxLayoutGap="32px">
```
GrandSchtroumpf, line 6: same as above

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.scss
```diff
@@ -0,0 +1,41 @@
...
+  align-items: center;
+  padding-block: 32px 48px;
+  background-image: url("/assets/images/banner_empty.png");
+}
+
+section {
+  margin-top: 16px;
+
+  button {
+    background-color: rgba(30, 27, 66, 1);
```
GrandSchtroumpf, line 13: remove the alpha if it's 1

libs/landing/src/lib/appBridgeBanner/app-bridge-banner.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  selector: 'app-bridge-banner',
+  templateUrl: './app-bridge-banner.component.html',
+  styleUrls: ['./app-bridge-banner.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+
+export class AppBridgeBannerComponent {
+  @HostBinding('class') class = 'dark-contrast-theme';
+
+  public currentApp = this.app;
```
GrandSchtroumpf, line 23: you don't need this. Just set `app` as public

GrandSchtroumpf (approved):

PR #7907, opened 2022-03-04
refactor movie into data model library

george43g:
data model lib starting

this is an example for movie

updated hundreds of imports and created a new data model lib

also configured additional linter steps and module import boundaries


apps/backend-functions/src/data/types.ts
```diff
@@ -18,6 +18,6 @@ export {
 } from '@blockframes/permissions/+state/permissions.firestore';
 export { PublicUser } from '@blockframes/user/+state/user.firestore';
 export { RequestDemoInformations } from '@blockframes/utils/request-demo';
-export { MovieAnalytics, MovieEventAnalytics } from '@blockframes/movie/+state/movie.firestore';
+export { MovieAnalytics, MovieEventAnalytics } from '@blockframes/data-model';
```
phpgeekfr, line 21: add `MovieDocument` to this instead of line 1

phpgeekfr (commented):
I added a commit to correct some imports, thinking it was the only ones missing but there is still a lot.

ng serve the app to check

apps/backend-functions/src/movie.ts
```diff
@@ -14,7 +14,7 @@ import { sendMovieSubmittedEmail } from './templates/mail';
 import { sendMail } from './internals/email';
 import { groupIds } from '@blockframes/utils/emails/ids';
 import { CallableContext } from 'firebase-functions/lib/providers/https';
-import { Movie } from '@blockframes/movie/+state';
+import { Movie } from '@blockframes/data-model';
```
phpgeekfr, line 17: same

apps/backend-ops/src/rescueJWP.ts
```diff
@@ -1,7 +1,7 @@
 
-import { Movie } from '@blockframes/movie/+state';
+import { Movie } from '@blockframes/data-model';
 import { getCollection, jwplayerApiV2 } from '@blockframes/firebase-utils';
-import { MovieVideo } from '@blockframes/movie/+state/movie.firestore';
+import { MovieVideo } from '@blockframes/data-model';
```
phpgeekfr, line 4: same

phpgeekfr (commented):
@GrandSchtroumpf what do you think of this structure?

Now we have warnings (for now but maybe errors later) telling that we should not do this kind of import, which can be a good idea.

apps/backend-functions/src/movie.ts
george43g: I think it's too soon to update these. Because there is a types file in backend functions that serves no purpose - we either migrate them all to `data-model` or all in the types file, either way I need to finish refactor first.

apps/backend-functions/src/movie.ts
```diff
@@ -97,97 +116,111 @@ export async function onMovieDelete(
...
+          _meta: createDocumentMeta({ createdFrom: appAccess }),
         });
       });
 
     await triggerNotifications(notifications);
   }
 
   // If movie was accepted but is not anymore, clean wishlists
-  if (Object.keys(after.app).map(a => before.app[a].status === 'accepted' && after.app[a].status !== before.app[a].status)) {
+  if (
```
GrandSchtroumpf, line 198: please put back the code as it was.

phpgeekfr (approved):

PR #7920, opened 2022-03-04
Bruce/e2e signup

phpgeekfr:


apps/festival/festival-e2e/src/integration/login/signup.spec.ts
```diff
@@ -43,7 +43,7 @@ describe('Signup', () => {
       //.then(mail => deleteEmail(mail.id))
     interceptEmail({subject: `Archipel Market - ${user.company.name} was created and needs a review`})
       .then(mail => deleteEmail(mail.id));
-    interceptEmail({body: `${user.firstName} ${user.lastName}`})
+    interceptEmail({body: `${user.firstName}`})
```
phpgeekfr, line 46: in emails, looks like we only have `hi {firstname}`

phpgeekfr: since username is not random, we can have a test passing because another test sent the same email

libs/user/src/lib/auth/components/data-validation/data-validation.component.ts
```diff
@@ -46,13 +45,23 @@ export class AuthDataValidationComponent implements OnInit {
...
   openIntercom(): void {
     return this.intercom.show();
   }
 
+  private async listenOnEmailVerified() {
+    let emailVerified = false;
+    while (emailVerified === false) {
+      const firebaseUser = await this.authService.reloadUser();
+      emailVerified = firebaseUser.emailVerified;
+      if (!emailVerified) await delay(3000);
```
phpgeekfr, line 60: maybe do exponential backoff

apps/festival/festival-e2e/src/integration/login/signup.spec.ts
```diff
@@ -43,7 +43,7 @@ describe('Signup', () => {
       //.then(mail => deleteEmail(mail.id))
     interceptEmail({subject: `Archipel Market - ${user.company.name} was created and needs a review`})
```
phpgeekfr, line 44: possible mail overlapp

apps/festival/festival-e2e/src/integration/login/signup.spec.ts
n-germain-work: > in emails, looks like we only have `hi {firstname}`

We have both in the mail :
image

> since username is not random, we can have a test passing because another test sent the same email

I'm changing the code to check for user's email inside the body, it's the only mail having this, and it changes at each test 👌

apps/festival/festival-e2e/src/integration/login/signup.spec.ts
n-germain-work: The company name changes with each test, and each interceptEmail function only looks for mails received after it's own launch :
image

It that ok for you ?

Issue #7931, opened 2022-03-09
Update Organization model imports


phpgeekfr:
Update Organization model imports

Start from https://github.com/blockframes/blockframes/pull/7934/files

PR #7934, opened 2022-03-09
devops(organisation): refactor to model lib

george43g:
- start of a data model refactor
- the code has been moved to models lib but none of the imports have been updated
- you will likely use search and replace to find everywhere that @blockframes/organisation has been imported and update the imports to point to @blk/model where needed
- ideally, you will also refactor the imports of the `model` lib itself - at the moment, the models are importing from other libs such as utils - they either shouldnt do this (if it's a type, it should be moved into models lib) or the dependent lib should also be a shared lib ie such as utils should not have any browser or node based code in it

#7931


KvnMrs:
tests not passed and found some errors to correct.

Issue #7932, opened 2022-03-09
Update Users model imports


phpgeekfr:
Update Users model imports

PR #7940, opened 2022-03-09
 devops(movie): refactor to model lib 

phpgeekfr:
#7932


phpgeekfr (commented):
please check for missing `;` 
after some `import { User } from '@blockframes/model'` for example

apps/backend-functions/src/invitation.ts
```diff
@@ -5,7 +5,7 @@ import { InvitationOrUndefined, OrganizationDocument } from './data/types';
 import { onInvitationToJoinOrgUpdate, onRequestToJoinOrgUpdate } from './internals/invitations/organizations';
 import { onInvitationToAnEventUpdate } from './internals/invitations/events';
 import { InvitationBase, createInvitation, InvitationStatus, InvitationDocument } from '@blockframes/invitation/+state/invitation.firestore';
-import { createPublicUser, PublicUser } from '@blockframes/user/+state/user.firestore';
+import { createPublicUser, PublicUser } from '@blockframes/model'; // TODO #7932 example
```
phpgeekfr, line 8: please remove the comment `// TODO #7932 example`

apps/backend-functions/src/media.ts
```diff
@@ -7,7 +7,7 @@ import { logger, storage } from 'firebase-functions';
 import { CallableContext } from 'firebase-functions/lib/providers/https';
 
 // Blockframes dependencies
-import { PublicUser, User } from '@blockframes/user/types';
+import { PublicUser, User } from '@blockframes/model'; // TODO #7932 example
```
phpgeekfr, line 10: same

libs/model/src/lib/user.ts
```diff
@@ -59,3 +60,32 @@ export function createPublicUser(user: Partial<User> = {}): PublicUser {
     orgId: user.orgId ?? ''
   }
 }
+
```
phpgeekfr, line 63: since `UserSettings` is used only on this file, you can remove the `export`

```
export interface UserSettings {
  notifications?: NotificationSettings,
}
```

apps/backend-functions/src/invitation.ts
KevinDurand974: Done

apps/backend-functions/src/media.ts
KevinDurand974: Done

libs/model/src/lib/user.ts
KevinDurand974: Done

george43g:
This is coming along nicely

george43g (approved):

Issue #7936, opened 2022-03-09
Desactivate Yandex


cams-goussale:
### Description

Due to security issues, it has been decided by management to desactivate Yandex. Let's not delete the code, in case we want to reactivate it later on.

PR #7945, opened 2022-03-09
removed yandex #7936

phpgeekfr:
removed yandex #7936

code was commented.

If it is totally removed, we will also be able to remove `export const yandex` from blockframes-xxx.ts files

### DIFF
--- a/apps/festival/festival/src/app/app.module.ts
+++ b/apps/festival/festival/src/app/app.module.ts
@@ -32,7 +32,8 @@ import { MatNativeDateModule } from '@angular/material/core';
 
 // Blockframes
 import { SentryModule } from '@blockframes/utils/sentry.module';
-import { YandexMetricaService } from '@blockframes/utils/yandex-metrica/yandex-metrica.service';
+// #7936 this may be reactivated later
+// import { YandexMetricaService } from '@blockframes/utils/yandex-metrica/yandex-metrica.service';
 import { IntercomService } from '@blockframes/utils/intercom/intercom.service';
 import { FireAnalytics } from '@blockframes/utils/analytics/app-analytics';
 import { ErrorLoggerModule } from '@blockframes/utils/error-logger.module';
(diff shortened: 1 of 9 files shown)


apps/festival/festival/src/app/app.module.ts
```diff
@@ -90,12 +91,12 @@ export class AppModule {
...
     analytics: FireAnalytics,
     intercomService: IntercomService,
-    yandexService: YandexMetricaService,
+    // yandexService: YandexMetricaService, #7936 this may be reactivated later
     gdprService: GDPRService,
     authService: AuthService,
   ) {
     const { intercom, yandex } = gdprService.cookieConsent;
-    if (yandex) yandexService.insertMetrika('festival');
+    // if (yandex) yandexService.insertMetrika('catalog'); #7936 this may be reactivated later
```
RemcoSimonides, line 99: To prevent mistakes when reactivating; change to `'festival'` instead of `'catalog'`

apps/financiers/financiers/src/app/app.module.ts
```diff
@@ -91,13 +92,13 @@ export class AppModule {
...
     intercomService: IntercomService,
-    yandexService: YandexMetricaService,
+    // yandexService: YandexMetricaService, #7936 this may be reactivated lateryandexService: YandexMetricaService,
     gdprService: GDPRService,
     authService: AuthService,
   ) {
 
     const { intercom, yandex } = gdprService.cookieConsent;
-    if (yandex) yandexService.insertMetrika('financiers');
+    // if (yandex) yandexService.insertMetrika('catalog'); #7936 this may be reactivated later
```
RemcoSimonides, line 101: To prevent mistakes when reactivating; change to `'financiers'` instead of `'catalog'`

apps/financiers/financiers/src/app/app.module.ts
```diff
@@ -91,13 +92,13 @@ export class AppModule {
     router: Router,
     analytics: FireAnalytics,
     intercomService: IntercomService,
-    yandexService: YandexMetricaService,
+    // yandexService: YandexMetricaService, #7936 this may be reactivated lateryandexService: YandexMetricaService,
```
RemcoSimonides, line 95: Remove text 'later'

RemcoSimonides (approved):

Issue #7742, opened 2022-02-10
[Seller's onboarding] Reviewing Emails & Notifications


cams-goussale:
### Description

Yuki is struggling to onboard Sellers on the platform, due to the legal agreement we currently have in place: the `MRA`. The solution we've found is to change this legal agreement and use a `Representation agreement` instead.

What's the main impact? AC can no longer do **Package Deals for Buyers**. So we need to modify some Emails & Notifications that are being sent during the flow.

### What's new

- [x] These emails are related to the Package deal for Buyer. They must be desactivated, meaning they shouldn't be sent.
[Buyer receives summary of all contracts (all declined)](https://mc.sendgrid.com/dynamic-templates/d-c92f8799273f497cb2256eb6ed8baa78/version/8f18eade-21f0-48df-9ec7-47db6324bc17/editor) - `template id : d-c92f8799273f497cb2256eb6ed8baa78`
[Buyer receives summary of all contracts (at least 1 accepted)](https://mc.sendgrid.com/dynamic-templates/d-4f067ef3abca42cc9d8dc25b57935ccc/version/17913ecb-1a68-4483-805d-5442841f2863/editor) - `template id : d-4f067ef3abca42cc9d8dc25b57935ccc`
[Seller receives confirmation his offer is now under signature](https://mc.sendgrid.com/dynamic-templates/d-38ee6949f4724b1c8796b8df9229f21d/version/19b7e9e9-47d5-4d76-a0d9-511b65f6bb66/editor) - `template id : d-38ee6949f4724b1c8796b8df9229f21d`

- [x] The corresponding notifications of these emails also need to be desactivated. I do not know the id of the notification, I can only link to the excel with the text.
[Buyer receives summary of all contracts (at least 1 accepted)
](https://docs.google.com/spreadsheets/d/1OflJnj5dcEyWeorBVhHsIqWSovRldJpk5yYd20U5_Mw/edit#gid=657134932&range=C20)[Seller receives confirmation his offer is now under signature](https://docs.google.com/spreadsheets/d/1OflJnj5dcEyWeorBVhHsIqWSovRldJpk5yYd20U5_Mw/edit#gid=657134932&range=A20)

- [x] These emails need a few wording changes.
[...]

Issue #7624, opened 2022-01-25
Email variables refactoring 


Jackseed:
#### AM - BUYER - CTA ASK FOR A SCREENING - Interactions
Template-id: `d-a03d096d252f4c27a500ff16d13f1795`
- [ ] Add a link to the event (`event.url`)

#### AM - BUYER - CTA ASKING PRICE - Interactions
Template-id: `d-ebf41f86fa124978a15b27bec75faf09`
- [ ] `movie.international.title` should be `movie.title.international`

#### AM - BUYER - ASKING PRICE REQUEST - Interactions
Template-id: `d-fa8f0f0f5799468ca3ac478c7ceeeb68`
- [x] `movieTitle` should be `movie.title.international`
- [ ] `orgNames` should be `org.denomination`
**Antoine**: Not possible, we currently handle the case where a a movie has several orgIds
- [ ] Activate V1 for the release

#### AM - ORG - ASKING PRICE REQUEST - Interactions
Template-id: `d-13c1cbe54bb74a9c9e9ea506db02669f` 
- [x] `movieTitle` should be `movie.title.international`
- [x] `buyer` should be `buyer.firstName buyer.lastName`
- [x] Add a `org.denomination` variable
- [x] Add a `org.country` variable
- [ ] Activate V1 for the release

#### AM - ORG - SCREENING REQUEST - Interactions
Template-id: `d-1a4f08ee48734343bb02f5f3238a5713`
- [x] Add a `buyer.email` variable
- [x] `buyer` should be `buyer.firstName buyer.lastName`
- [x] Add a `org.denomination` variable
- [x] Add a `org.country` variable
- [ ] Activate V1 for the release

## Deal Negotiation emails
To stay consistent some changes to make on email variables sent:

#### Price
- [x] When there is a `price` variable, format it to be American number
- [x] `negotiation.currency` should be changed from text (eg "Euro") to symbol (eg "€")
@cams-goussale we should define which symbol to use for every currency : 
ex: australian dollar $ or A$ ?

#### AC - Contract - Seller receives an offer
Template-id: `d-f7c2e13b7d6b427daf08bdfdf904d697`
- [x] `title.title.international` should be `movie.title.international`

#### ~AC - Contract - Offer under signature~
~Template-id: `d-38ee6949f4724b1c8796b8df9229f21d`~
- [x] ~`offerId` should be `offer.id`~
- [x]  ~`title` should be `movie.title.international`~
- [x] ~Activate V3 for the release~

[...]

Issue #7491, opened 2021-12-29
update templateIds.offer.created to use OrgEmailData instead of OrganizationDocument


phpgeekfr:
To uniformize data sent to sendGrid for email templates, update `templateIds.offer.created` to use `OrgEmailData` instead of `OrganizationDocument`

- [ ] use `getOrgEmailData`
- [ ] remove `OrganizationDocument` from `EmailTemplateRequest`

Poppe-A:
We also should clean `EmailTemplateRequest.data` (emails/utils.ts) : 
- Add all properties used as optional (app, movieUrl, orgNames, negotiation, date, offer, offerId, ...)
- `buyer?: PublicUser | string` is dirty, find a way to clean it
- We have both `pageUrl` and `pageURL`, standardize
-  `MailBucket` should be renammed to `BucketEmailData`
- `movie?: MovieEmailData | MovieDocument` is dirty, find a way to clean it

phpgeekfr:
Linked with #8355

phpgeekfr:
also look at :
- [ ] adminOfferCreatedConfirmationEmail : whole `org` document is sent to sendgrid
- [ ] buyerOfferCreatedConfirmationEmail : same 
- [ ] contractCreatedEmail: buyerOrg

- [ ] templateIds.offer.created seems to have been removed but lets check deeper

PR #7946, opened 2022-03-10
Reviewing emails and notifications

Poppe-A:
#7742
Some code about non used emails is commented. But we keep it in case we use package deal again in the future.
This PR allow us to display more informations on negotiation flow notifications (movie tile, user name, org)
and also try to clean a bit the way data is sended by mail (sending jsut some properties instead of whole object)


apps/backend-functions/src/notification.ts
```diff
@@ -616,13 +620,18 @@ async function sendCreatedCounterOfferConfirmation(recipient: User, notification
...
   const isMailRecipientBuyer = recipient.orgId === negotiation.buyerId;
   const app: App = 'catalog';
   const toUser = getUserEmailData(recipient);
 
-  const senderTemplate = counterOfferSenderEmail(toUser, recipientOrg, contract.offerId, negotiation, title, contract.id, { isMailRecipientBuyer });
-  return sendMailFromTemplate(senderTemplate, app, groupIds.unsubscribeAll);
+  const senderTemplate = counterOfferSenderEmail(toUser, recipientOrg, contract.offerId, negotiation, movie, contract.id, { isMailRecipientBuyer });
+  const adminTemplate = toAdminCounterOfferEmail(movie);
+
+  await Promise.all([
```
RemcoSimonides, line 631: use `return` instead of `await` here. To keep it as it was and also maybe we don't need to wait for the emails to be sent ;)

apps/backend-functions/src/notification.ts
```diff
@@ -685,71 +693,81 @@ async function sendContractStatusChangedConfirmation(recipient: User, notificati
     user: toUser,
     org: recipientOrg,
     contract,
-    title,
+    movie,
```
RemcoSimonides, line 696: Change to `movie: title`

apps/backend-functions/src/notification.ts
```diff
@@ -685,71 +693,81 @@ async function sendContractStatusChangedConfirmation(recipient: User, notificati
...
     pageURL,
     app: { name: appName.catalog }
   };
 
   let templateId = templateIds.negotiation.myContractWasAccepted;
+  let adminTemplateId = null;
```
RemcoSimonides, line 702: No need to initiate with null
`let adminTemplateId;` does suffice

apps/backend-functions/src/notification.ts
```diff
@@ -685,71 +693,81 @@ async function sendContractStatusChangedConfirmation(recipient: User, notificati
...
     templateId = templateIds.negotiation.myOrgAcceptedAContract;
+    adminTemplateId = templateIds.negotiation.toAdminContractAccepted;
     data.org = counterOfferSenderOrg;
   }
   if (options.isActionDeclined) {
     templateId = templateIds.negotiation.myContractWasDeclined;
     if (options.didRecipientAcceptOrDecline) {
-      templateId = templateIds.negotiation.myOrgDeclinedAContract;
+      templateId = templateIds.negotiation.myOrgDeclinedAContract; 
+      adminTemplateId = templateIds.negotiation.toAdminContractDeclined;
```
RemcoSimonides, line 712: Are you sure it should be within this if statement? `if (options.didRecipientAcceptOrDecline) {`
It makes sense to me if it's next to `templateId = templateIds.negotiation.myContractWasDeclined;`

apps/backend-functions/src/notification.ts
```diff
@@ -685,71 +693,81 @@ async function sendContractStatusChangedConfirmation(recipient: User, notificati
...
-  mailContract['currency_long'] = movieCurrencies[negotiation.currency]
-
-  const template = offerUnderSignature(toUser, contract.offerId, contract, mailContract, movie.title.international);
-  return sendMailFromTemplate(template, app, groupIds.unsubscribeAll);
-}
+  if(adminTemplateId) {
+    const adminTemplate = { to: supportEmails.catalog, templateId: adminTemplateId, data };
+    mailToSend.push(sendMailFromTemplate(adminTemplate, app, groupIds.unsubscribeAll))
+  }
+  await Promise.all(mailToSend);
```
RemcoSimonides, line 723:
```
if (adminTemplateId) {
  const adminTemplate = { to: supportEmails.catalog, templateId: adminTemplateId, data };
  return Promise.all([
    sendMailFromTemplate(template, app, groupIds.unsubscribeAll),
    sendMailFromTemplate(adminTemplate, app, groupIds.unsubscribeAll)
  ])
} else {
  return sendMailFromTemplate(template, app, groupIds.unsubscribeAll),
}
```

apps/backend-functions/src/templates/mail.ts
```diff
@@ -426,43 +425,49 @@ export function counterOfferSenderEmail(
...
-
-export function offerUnderSignature(
-  user: UserEmailData, offerId: string, contract: ContractDocument, negotiation: MailContract,
-  title: string
-): EmailTemplateRequest {
-  const pageURL = `${appUrl.content}/c/o/dashboard/sales/${contract.id}/view`;
-  const data = {
-    contract, offerId, user, negotiation,
-    title, pageURL, app: { name: appName.catalog }
+    movie
```
RemcoSimonides, line 435: `movie: title`

libs/utils/src/lib/emails/ids.ts
```diff
@@ -74,9 +74,10 @@ export const templateIds = {
   offer: {
     toAdmin: 'd-94a20b20085842f68fb2d64fe325638a',
     toBuyer: 'd-1d6a00bab6f448299e9fcdaec7c4c787',
-    allContractsAccepted: 'd-4f067ef3abca42cc9d8dc25b57935ccc',
-    allContractsDeclined: 'd-c92f8799273f497cb2256eb6ed8baa78',
-    underSignature: 'd-38ee6949f4724b1c8796b8df9229f21d'
+    // #7946 this may be reactivated later
+    // allContractsAccepted: 'd-4f067ef3abca42cc9d8dc25b57935ccc', // To remove
+    // allContractsDeclined: 'd-c92f8799273f497cb2256eb6ed8baa78', //To remove
+    // underSignature: 'd-38ee6949f4724b1c8796b8df9229f21d'//To remove
```
RemcoSimonides, line 80: You can remove the `// To remove` comments

libs/user/src/lib/auth/pages/identity/identity.component.html
```diff
@@ -142,7 +142,10 @@ <h1>Create your account</h1>
         <div class="privacy" fxLayout="column" fxLayoutAlign="start" fxLayoutGap="16px">
           <mat-checkbox test-id="checkbox" color="primary" formControlName="termsOfUse">
             <span class="mat-caption">I hereby accept {{ app | appName }}&nbsp;
-              <a target="_blank" routerLink="/auth/terms">Terms and Conditions</a>.</span>
+              <a target="_blank" routerLink="/auth/terms">
+                {{app === 'catalog' ? 'General Conditions of Sales' : 'Terms and Conditions'}}         
+              </a>.
```
RemcoSimonides, line 147: Why is this? I don't see it in the issue?

apps/backend-functions/src/notification.ts
Poppe-A: If I understood correctly, we only want to send the email when the org declined the contract, not when the buyer do it

libs/user/src/lib/auth/pages/identity/identity.component.html
Poppe-A: It's from another issue, but since it's very fast to do, I wanted to put it here. But I will remove it

apps/backend-functions/src/notification.ts
```diff
@@ -278,21 +281,22 @@ export async function onNotificationCreate(snap: FirebaseFirestore.DocumentSnaps
...
+      //   await sendOfferAcceptedOrDeclinedConfirmation(recipient, notification)
+      //     .then(() => notification.email.isSent = true)
+      //     .catch(e => notification.email.error = e.message);
+      //   break;
+      // case 'underSignature':
+      //   await sendOfferUnderSignatureConfirmation(recipient, notification)
+      //     .then(() => notification.email.isSent = true)
+      //     .catch(e => notification.email.error = e.message);
+      //   break;
       case "userRequestAppAccess":
```
phpgeekfr, line 300: replace to singlequotes please

apps/backend-functions/src/notification.ts
```diff
@@ -685,71 +693,81 @@ async function sendContractStatusChangedConfirmation(recipient: User, notificati
     user: toUser,
     org: recipientOrg,
     contract,
-    title,
+    movie: title,
```
phpgeekfr, line 696: doing this will break emails templates that where using `title`

apps/backend-functions/src/templates/mail.ts
```diff
@@ -358,7 +358,7 @@ export function contractCreatedEmail(
   const data = {
     user: toUser,
     app: { name: appName.catalog },
-    title,
+    movie: title,
```
phpgeekfr, line 361: same

apps/backend-functions/src/templates/mail.ts
```diff
@@ -376,15 +376,15 @@ export function adminOfferCreatedConfirmationEmail(toUser: UserEmailData, org: O
...
   const contracts = bucket.contracts.map(contract => createMailContract(contract));
-  const pageURL = `${appUrl.content}/c/o/marketplace/offer/${offerId}`;
+  const pageURL = `${appUrl.content}/c/o/marketplace/offer/${offer.id}`;
   const data = {
     app: { name: appName.catalog },
     bucket: { ...bucket, contracts },
     user: toUser,
     pageURL,
-    offerId,
+    offer,
```
phpgeekfr, line 387: do we really need the whole offer object ? If only id is used, use only id, else, create something like `orgEmailData`

apps/backend-functions/src/templates/mail.ts
```diff
@@ -399,10 +399,9 @@ export function counterOfferRecipientEmail(
     : `${appUrl.content}/c/o/dashboard/sales/${contractId}/view`;
   const data = {
     user: toUser,
-    offerId,
     org: senderOrg,
     pageURL,
-    title,
+    movie: title,
```
phpgeekfr, line 404: same as above, and maybe created something like `orgEmailData`

apps/backend-functions/src/templates/mail.ts
```diff
@@ -426,43 +425,49 @@ export function counterOfferSenderEmail(
     contractId,
     app: { name: appName.catalog },
     negotiation: { ...negotiation, terms, currency },
-    title
+    movie: title
```
phpgeekfr, line 428: same

apps/backend-functions/src/templates/mail.ts
```diff
@@ -426,43 +425,49 @@ export function counterOfferSenderEmail(
...
-  const acceptedContracts = contracts.filter(contract => contract.status === 'accepted');
-  const pageURL = `${appUrl.content}/c/o/marketplace/offer/${offer.id}`;
+export function toAdminCounterOfferEmail( title: MovieDocument): EmailTemplateRequest {
   const data = {
-    contracts: isOfferAccepted ? acceptedContracts : contracts,
-    offer,
-    user,
-    pageURL,
-    app: { name: appName.catalog }
+    movie: title
```
phpgeekfr, line 435: same

libs/utils/src/lib/emails/utils.ts
```diff
@@ -23,11 +26,18 @@ export interface EmailTemplateRequest {
...
+    eventUrl?: string,
     pageURL?: string,
     bucket?: MailBucket,
     baseUrl?: string,
     date?: string,
+    movie?: MovieDocument,
+    buyer?: PublicUser | string,
+    contract?: ContractDocument,
+    territories?: string,
+    contractId?: string,
```
phpgeekfr, line 40: I think you are missing : app,  movieTitle, MovieUrl, orgNames, negotiation, date, offer, offerId

We have both `pageUrl` and `pageURL`

phpgeekfr: doing `buyer?: PublicUser | string,` also shows that we are doing not pretty stuffs.

Can you add this to #7491 ?

apps/backend-functions/src/notification.ts
phpgeekfr: and this we can have movie & tv contentType, title is more appropriate

phpgeekfr (changes requested):

libs/utils/src/lib/emails/utils.ts
```diff
@@ -23,11 +26,18 @@ export interface EmailTemplateRequest {
   data: {
     org?: OrgEmailData | OrganizationDocument, // @TODO #7491 template d-94a20b20085842f68fb2d64fe325638a uses OrganizationDocument but it should use OrgEmailData instead
     user?: UserEmailData,
+    userSubject?: UserEmailData,
     event?: EventEmailData,
+    eventUrl?: string,
     pageURL?: string,
     bucket?: MailBucket,
```
phpgeekfr, line 33: for a better consistency, `MailBucket` should be renammed to `BucketEmailData`

apps/backend-functions/src/offer.ts
```diff
@@ -1,21 +1,24 @@
...
+// import { Change } from 'firebase-functions';
+// import { createDocumentMeta } from './data/internals';
+// import { Sale } from '@blockframes/contract/contract/+state/contract.model';
+// import { getSeller } from '@blockframes/contract/contract/+state/utils';
+// import { NotificationTypes } from './data/types';
+// import { EmailTemplateRequest, sendMailFromTemplate } from './internals/email';
+// import { App } from '@blockframes/utils/apps';
+// import { appUrl, supportEmails } from '@env';
+
+
```
RemcoSimonides, line 21:
```suggestion
```

apps/backend-functions/src/notification.ts
```diff
@@ -685,71 +693,81 @@ async function sendContractStatusChangedConfirmation(recipient: User, notificati
...
-  const mailContract: MailContract = createMailContract(negotiation);
-
-  const toUser = getUserEmailData(recipient);
-  const app: App = 'catalog';
-  mailContract['currency_long'] = movieCurrencies[negotiation.currency]
-
-  const template = offerUnderSignature(toUser, contract.offerId, contract, mailContract, movie.title.international);
-  return sendMailFromTemplate(template, app, groupIds.unsubscribeAll);
-}
+  if(adminTemplateId) {
```
RemcoSimonides, line 719:
```suggestion
  if (adminTemplateId) {
```

apps/backend-functions/src/notification.ts
RemcoSimonides, line 696: Shouldn't this also use `getMovieEmailData()`?

libs/notification/src/lib/+state/notification.service.ts
```diff
@@ -2,10 +2,10 @@ import { Inject, Injectable } from '@angular/core';
 import { Notification } from './notification.model';
 import { CollectionConfig, CollectionService } from 'akita-ng-fire';
 import { AuthService } from '@blockframes/auth/+state';
-import { filter, map, switchMap, take } from 'rxjs/operators';
+import { filter, map, switchMap, take, tap } from 'rxjs/operators';
```
RemcoSimonides, line 5:
```suggestion
import { filter, map, switchMap, take } from 'rxjs/operators';
```

libs/notification/src/lib/+state/notification.service.ts
```diff
@@ -482,4 +513,15 @@ export class NotificationService extends CollectionService<NotificationState> {
       })
     );
   }
+
+  // The goal of this method is to return a user name or an org name depending on who is receiving a counter offer
+  public async nameToDisplay(notification: Notification, contract: Contract) {
+    if (contract.buyerUserId === notification.toUserId) {
+      const org = await this.orgService.valueChanges(contract.sellerId).pipe(take(1)).toPromise() as Organization;
```
RemcoSimonides, line 520: How come it doesn't know the type?

notification doesn't already contain either the org or user data?

libs/notification/src/lib/+state/notification.service.ts
```diff
@@ -482,4 +513,15 @@ export class NotificationService extends CollectionService<NotificationState> {
       })
     );
   }
+
+  // The goal of this method is to return a user name or an org name depending on who is receiving a counter offer
```
RemcoSimonides, line 517:
```
/**
* @returns A username or org name depending on who is receiving a counter offer
*/
```
☝️ this way, the text will show up when you hover the function name

libs/utils/src/lib/emails/utils.ts
```diff
@@ -23,11 +27,19 @@ export interface EmailTemplateRequest {
...
     baseUrl?: string,
     date?: string,
+    movie?: MovieEmailData | MovieDocument,
```
RemcoSimonides, line 37: Would be nice if we can have only `MovieEmailData`. Could you add that as to do for issue https://github.com/blockframes/blockframes/issues/7624 perhaps?

libs/utils/src/lib/emails/utils.ts
```diff
@@ -23,11 +27,19 @@ export interface EmailTemplateRequest {
...
+    offer?: OfferEmailData,
+    buyer?: PublicUser | string,
```
RemcoSimonides, line 39: Would be nice if we can have only `PublicUser`. Could you add that as to do for issue https://github.com/blockframes/blockframes/issues/7624 perhaps?

RemcoSimonides (approved):

libs/notification/src/lib/+state/notification.service.ts
Poppe-A: Thx !

libs/notification/src/lib/+state/notification.service.ts
Poppe-A: This is what the notifications look like :

```
app: {isRead: false}
docId: "FNBIAvyBVAuYFsR4Ms6R"
docPath: "contracts/FNBIAvyBVAuYFsR4Ms6R/negotiations/bdSuxDWVZfdJnntvaa0e"
email: {isSent: true}
id: "QKXRBpvHofIkWTXwZ3Xk"
offerId: "ROB-464544"
toUserId: "g04XYIOQ2AQ2a6rZ7xC5cNofpo33"
type: "createdCounterOffer"
_meta: {createdBy: 'internal', createdAt: t, createdFrom: 'catalog'}
```

RemcoSimonides: Alright. Asked because the `Notification` interface does have org / user in there

But how come it doesn't know the type? It should know the type already. Is `contract.sellerId` not known as a string?

libs/notification/src/lib/+state/notification.service.ts
```diff
@@ -482,4 +513,17 @@ export class NotificationService extends CollectionService<NotificationState> {
...
     );
   }
+
+  /**
+  * @returns A username or org name depending on who is receiving a counter offer
+  */
+  public async nameToDisplay(notification: Notification, contract: Contract) {
+    if (contract.buyerUserId === notification.toUserId) {
+      const org = await this.orgService.valueChanges(contract.sellerId).pipe(take(1)).toPromise() as Organization;
+      return org.denomination.public;
```
phpgeekfr, line 523: use `orgName`

libs/notification/src/lib/+state/notification.service.ts
Poppe-A: My mistake. `as Organization` is removed

PR #7954, opened 2022-03-10
Kevind data model notifications

KevinDurand974:
Next from #7951 who was closed because of a bug.


apps/backend-functions/src/notification.ts
```diff
@@ -1,6 +1,16 @@
-import { InvitationDocument, NotificationDocument, NotificationTypes } from './data/types';
+
```
phpgeekfr, line 1: extra line

phpgeekfr (approved):

apps/backend-functions/src/notification.ts
KevinDurand974: Done

PR #7956, opened 2022-03-11
devops(permissions): refactor to model lib

phpgeekfr:


libs/model/src/lib/permissions.ts
```diff
@@ -52,3 +52,9 @@ export function createDocPermissions(params: Partial<DocPermissionsDocument> = {
     ...params
   };
 }
+
+export interface Permissions extends PermissionsDocument {
+  docPermissions?: DocPermissions[];
+}
+
+type DocPermissions = DocPermissionsDocument;
```
phpgeekfr, line 60: in fact (my mistake) you should be able to do:

```suggestion
export interface Permissions extends PermissionsDocument {
  docPermissions?: DocPermissionsDocument[];
}
```

libs/model/src/lib/user.ts
```diff
@@ -2,7 +2,7 @@ import { DocumentMeta } from "@blockframes/utils/models-meta";
 import { NotificationTypesBase } from '@blockframes/model';
 import { createStorageFile, StorageFile } from "@blockframes/media/+state/media.firestore";
 import { Genre, Language, Media, Territory } from "@blockframes/utils/static-model";
-import { UserRole } from "@blockframes/permissions/types";
+import { UserRole } from "./permissions";
```
phpgeekfr, line 5: `import { NotificationTypesBase, UserRole } from '@blockframes/model';`

libs/organization/src/lib/permissions/+state/permissions.service.ts
```diff
@@ -1,13 +1,12 @@
...
-import { UserRole, createDocPermissions, PermissionsDocument } from './permissions.firestore';
-import { Permissions } from './permissions.model';
 import { CollectionService, CollectionConfig, AtomicWrite } from 'akita-ng-fire';
 import type firebase from 'firebase';
 import { UserService } from '@blockframes/user/+state/user.service';
 import { map, switchMap, tap } from 'rxjs/operators';
 import { AuthService } from '@blockframes/auth/+state';
 import { combineLatest, Observable, of } from 'rxjs';
 import { ActiveState, EntityState } from '@datorama/akita';
+import { createDocPermissions, PermissionsDocument, UserRole, Permissions} from 'libs/model/src/lib/permissions';
```
phpgeekfr, line 9: missing space

libs/model/src/lib/user.ts
KvnMrs: ' NotificationTypesBase ' don't work in this import, i don't have Notification in the model for now

KvnMrs (approved):

PR #7962, opened 2022-03-11
devops(contract): refactor to model lib

phpgeekfr:


libs/contract/src/lib/contract/+state/contract.model.ts
```diff
@@ -1,69 +0,0 @@
-
```
phpgeekfr, line 1: file looks like empty but still here, can you remove it ?

apps/catalog/catalog/src/app/dashboard/sales/shell/shell.component.ts
```diff
@@ -1,12 +1,14 @@
...
+import { ContractService } from '@blockframes/contract/contract/+state';
 import { ActivatedRoute } from '@angular/router';
 import { pluck, shareReplay, switchMap } from 'rxjs/operators';
 import { centralOrgId } from '@env';
 import { joinWith } from '@blockframes/utils/operators';
 import { MovieService } from '@blockframes/movie/+state/movie.service';
 import { IncomeService } from '@blockframes/contract/income/+state';
 import { Intercom } from 'ng-intercom';
+import { Sale, contractStatus } from '@blockframes/model';
+
```
phpgeekfr, line 12: please remove extra lines

apps/backend-functions/src/templates/mail.ts
```diff
@@ -6,16 +6,14 @@
 import { supportEmails, appUrl, e2eMode } from '../environments/environment';
 import { EmailRequest, EmailTemplateRequest } from '../internals/email';
 import { templateIds } from '@blockframes/utils/emails/ids';
-import { EventEmailData, getMovieEmailData, getOfferEmailData, MovieEmailData, OrgEmailData, UserEmailData } from '@blockframes/utils/emails/utils';
 import { RequestDemoInformations } from '../data/types';
-import { PublicUser, OrganizationDocument, PublicOrganization, MovieDocument } from '@blockframes/model';
+import { PublicUser, OrganizationDocument, PublicOrganization, MovieDocument, createMailContract, ContractDocument } from '@blockframes/model';
```
phpgeekfr, line 10: can you check if you can add `type` like this `import type` ?

libs/utils/src/lib/emails/utils.ts
```diff
@@ -11,15 +11,14 @@ import {
   User,
   PublicUser,
   OrganizationDocument,
-  orgName
+  orgName,
+  ContractDocument
 } from '@blockframes/model';
```
phpgeekfr, line 16: same

apps/backend-functions/src/templates/mail.ts
phpgeekfr: can you try with 
```
import  { PublicUser, OrganizationDocument, PublicOrganization, MovieDocument, createMailContract } from '@blockframes/model';
import type {  ContractDocument } from '@blockframes/model';
```

libs/utils/src/lib/emails/utils.ts
KvnMrs: it doesn't work too, if my syntax that i showed you is correct

PR #7976, opened 2022-03-14
devops(terms): refactor to model lib

KvnMrs:
wip


libs/model/src/lib/terms.ts
```diff
@@ -6,6 +6,7 @@ import { staticModel } from '@blockframes/utils/static-model';
 import { format } from 'date-fns';
 import { toLanguageVersionString } from '@blockframes/utils/utils';
 
+
```
phpgeekfr, line 9: extra line

phpgeekfr (approved):

PR #7992, opened 2022-03-14
devops(media): refactor to model lib

KevinDurand974:
Media model base
WIP


libs/model/src/lib/media.ts
```diff
@@ -1,7 +1,36 @@
...
+    storagePath: '',
+    ...file,
+  };
+}
+
+export interface StorageVideo extends StorageFile {
+  jwPlayerId: string;
+}
+
+export type MediaOutput = StorageFile[] | MovieVideo[] | MovieNote[] | string | MovieVideo;
```
phpgeekfr, line 33: `MediaOutput` is not used in the repo, please remove it

libs/model/src/lib/media.ts
```diff
@@ -71,7 +100,6 @@ export function recursivelyListFiles(document: any): StorageFile[] {
 }
 
 // TODO issue#4002 MOVE THE 2 FUNCTIONS BELLOW ELSEWHERE
-
 export function getFileNameFromPath(path: string) {
   if (typeof path !== 'string') {
     console.warn('UNEXPECTED PATH', path);
```
phpgeekfr, line 105: also remove 
```
// TODO issue#4002 MOVE THE 2 FUNCTIONS BELLOW ELSEWHERE
export function getFileNameFromPath(path: string) {
  if (typeof path !== 'string') {
    console.warn('UNEXPECTED PATH', path);
  }
  return (!!path && typeof path === 'string') ? path.split('/').pop() : '';
}
```

libs/model/src/lib/media.ts
```diff
@@ -82,4 +110,4 @@ export function getFileNameFromPath(path: string) {
 /** Used this only for background to let the browser deal with that with picture */
 export function getAssetPath(asset: string, theme: 'dark' | 'light', type: 'images' | 'logo' = 'images') {
   return `assets/${type}/${theme}/${asset}`;
-}
+}
```
phpgeekfr, line 113: for this one
```
/** Use this only for background to let the browser deal with that with picture */
export function getAssetPath(asset: string, theme: 'dark' | 'light', type: 'images' | 'logo' = 'images') {
  return `assets/${type}/${theme}/${asset}`;
}
```

Please create a file `utils.ts` in `libs/media/src/lib/image/directives` and put it here

phpgeekfr (approved):

Issue #7831, opened 2022-02-23
(A) Sellers' Analytics - Dashboard - Split by Company Activity Pie Chart


RemcoSimonides:
[Wireframe](https://www.figma.com/file/FhY2dJqeYrVTXh7Q7HA0m7/%E2%9A%99%EF%B8%8F-Sellers'-Dashboard-(DONE)?node-id=438%3A145660)

![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/01216b2f-f4f1-487f-b5fc-764941f23072)

2amarkina:
- [ ] Wording change from "Learn who are the users who were interested in this Title." to "Learn who are the users interested by this title." 
<img width="787" alt="Снимок экрана 2022-05-02 в 15 55 20" src="https://user-images.githubusercontent.com/55499036/166245895-3bb62f31-1aab-44f1-a574-f799e8411fba.png">

PR #7999, opened 2022-03-15
Dashboard analytics

yspringfield:
Fixed: #7831
Bildschirmfoto 2022-03-24 um 2 06 11 PM


libs/analytics/src/lib/components/line-chart/line-chart.component.ts
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+  @Input() set data(data: LineChartData[]) {
+    if (!data) return;
+    this.lineChartOptions.series = data
+    this.chart?.updateOptions(this.lineChartOptions);
+  }
+
+  @Input() set filmView(isFilmView: boolean) {
```
RemcoSimonides, line 88: Please also discuss with Stephanie about the design for this graph. The reason why there are two graphs, the interactions graph and the pageView graphs, is that there might be many more pageViews compared to interactions and thus that wouldn't look nice in the same graph. But there is no reason why these graphs would have different xaxis labels or thicker lines.

For the code:
I would try `@Input() @boolean set filmView() {}`
When this argument is passed to this component, then it's always true. So then you don't need to do `isFilmView ? : ` all the time

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -31,22 +33,71 @@ export class HomeComponent implements OnInit {
...
+  titleAnalytics$ = this.analyticsService.getGlobalAnalytics().pipe(
+    joinWith({
+      org: analytic => this.orgService.valueChanges(analytic.meta.orgId)
+    }),
+    shareReplay({ bufferSize: 1, refCount: true })
+  );
+
+  orgActivity$ = this.titleAnalytics$.pipe(
+    map(analytics => counter(analytics, 'org.activity', 'orgActivity')),
+    map(analytics => analytics.sort((a, b) => a.count - b.count).splice(0, 2))
```
RemcoSimonides, line 78: Why are you getting the top 2?

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -31,22 +33,71 @@ export class HomeComponent implements OnInit {
...
+  );
+
+  territoryActivity$ = this.titleAnalytics$.pipe(
+    map(analytics => counter(analytics, 'org.addresses.main.country', 'territories')),
+    map(analytics => analytics.sort((a, b) => a.count - b.count).splice(0, 2))
```
RemcoSimonides, line 83: Why are you getting the top 2?

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -31,22 +33,71 @@ export class HomeComponent implements OnInit {
...
+  );
+
+  interactions$ = this.titleAnalytics$.pipe(
+    map(analytics => analytics.reduce((acc, curr) => {
+      if (!acc[curr.name]) acc[curr.name] = { list: [], name: curr.name }
+      acc[curr.name].list.push(curr)
+    }, {} as any))
```
RemcoSimonides, line 90: What are you doing here? Return type is of type `any`.

Anyway, I think the data should be aggregated by date. Whereas date could be per day, per week or per month. If there were no analytics in that given period, it should default to 0 (not undefined because the line graph will skip the undefined when drawing the line).

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -31,22 +33,71 @@ export class HomeComponent implements OnInit {
   public hasAcceptedMovies$: Observable<boolean>;
   public hasDraftMovies$: Observable<boolean>;
 
-  public titleAnalytics$: Observable<MovieAnalytics[]> = this.orgService.currentOrg$.pipe(
-    switchMap(({ id }) => this.analytics.valueChanges(ref => ref
-      .where('type', '==', 'title')
-      .where('meta.ownerOrgIds', 'array-contains', id)
-    )),
-    map(toMovieAnalytics)
+  public lineChartData: LineChartData[] = [
```
RemcoSimonides, line 36: Hook the graph to your Firestore data instead.

RemcoSimonides (changes requested):
In my opinion you've put too much hay on your fork. It would have been better to first set-up the UI for the page, maybe add pie chart and map chart already because they are existing components. But adding the new line charts should have been in a separate PR. That would have made it easier for you and for me.

Also, I think you made it harder for yourself than needed. With this project, the designs are solid. So we can still discuss with Ana and Stephanie to come to an easier solution.

We should have a call btw because I've already created a line chart when I was doing research for this project. There are some valuable lessons that could have saved you time. For example what the data model for the graph should be like.

libs/analytics/src/lib/components/map/map.component.html
```diff
@@ -3,7 +3,7 @@
   <h5>Split by Company Territory</h5>
 </header>
 
-<world-map>
+<world-map *ngIf="this.top3.length; else noData">
```
RemcoSimonides, line 6: remove `this`

libs/analytics/src/lib/components/map/map.component.scss
```diff
@@ -9,4 +9,8 @@ world-map {
 
 section {
   width: 100%;
-}
\ No newline at end of file
+}
+
+img[asset]{
```
RemcoSimonides, line 14: Why `img[asset]` instead of `img`?

libs/ui/src/lib/dashboard/pages/home/home.component.html
```diff
@@ -1,66 +1,38 @@
...
+        <mat-icon svgIcon="movie"></mat-icon>
+      </button>
+      <button mat-fab color="default">
+        <mat-icon svgIcon="excel"></mat-icon>
+      </button>
+    </div>
+  </div>
+  <section class="surface carrousel-container">
+    <div fxFlex fxLayoutAlign="space-between center">
+      <h5>Lobster Film</h5>
```
RemcoSimonides, line 18: Get the most popular title

libs/ui/src/lib/dashboard/pages/home/home.component.html
```diff
@@ -1,66 +1,38 @@
...
+      <button mat-flat-button>
```
RemcoSimonides, line 19: This button isn't doing anything. You can make it an `<a>` with `routerLink` to /c/o/dashboard/analytics/titles/titleId

libs/ui/src/lib/dashboard/pages/home/home.component.html
```diff
@@ -1,66 +1,38 @@
...
+        <span>See Analytics details</span>
+        <mat-icon svgIcon="arrow_forward"></mat-icon>
+      </button>
+    </div>
+    <span class="indicative-message">Some insights about your most popular title.</span>
```
RemcoSimonides, line 24: We shouldn't have a 'custom' text type. There is likely already a class for what Stephanie wants here.

libs/ui/src/lib/dashboard/pages/home/home.component.html
```diff
@@ -1,66 +1,38 @@
...
+    <span class="indicative-message">Published on Archipel Market: 01/25/2022</span>
+    <bf-carousel flex layout>
+      <analytics-pie-chart fxFlex [col]="6" class="surface" carouselItem [data]="orgActivity$ | async">
```
RemcoSimonides, line 27: `col` is working now? if not, remove it.

libs/ui/src/lib/dashboard/pages/home/home.component.html
```diff
@@ -1,66 +1,38 @@
...
+      </analytics-pie-chart>
+      <analytics-map fxFlex [col]="6" class="surface" carouselItem [data]="territoryActivity$ | async"></analytics-map>
+    </bf-carousel>
+    <button mat-flat-button class="see_titles">
```
RemcoSimonides, line 31: Also change to `<a>` and link to /c/o/dashboard/analytics/titles

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -9,14 +9,15 @@ import { Movie } from '@blockframes/model';
...
-import { MovieAnalytics } from '@blockframes/analytics/components/movie-analytics-chart/movie-analytics.model';
+import { counter } from '@blockframes/analytics/+state/utils';
+
 // RxJs
 import { map, switchMap, shareReplay, tap } from 'rxjs/operators';
 import { Observable } from 'rxjs';
 
 // Intercom
 import { Intercom } from 'ng-intercom';
+import { joinWith } from '@blockframes/utils/operators';
```
RemcoSimonides, line 20: Move this import to `// Blockframes` section of imports

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -31,22 +32,32 @@ export class HomeComponent implements OnInit {
...
+  titleAnalytics$ = this.analyticsService.getAnalytics().pipe(
+    joinWith({
+      org: analytic => this.orgService.valueChanges(analytic.meta.orgId)
+    }),
+    shareReplay({ bufferSize: 1, refCount: true })
+  );
+
+  orgActivity$ = this.titleAnalytics$.pipe(
+    map(analytics => counter(analytics, 'org.activity', 'orgActivity')),
+    map(analytics => analytics.sort((a, b) => a.count - b.count))
```
RemcoSimonides, line 44: Why sort?

libs/ui/src/lib/icon.service.ts
```diff
@@ -7,6 +7,7 @@ export const icons = [
   { name: 'access_time', url: 'assets/icons/access_time.svg' },
   { name: 'account_balance_wallet', url: 'assets/icons/account_balance_wallet.svg' },
   { name: 'account_circle', url: 'assets/icons/account_circle.svg' },
+  { name: 'analytics_interactions', url: 'assets/icons/analytics_interactions.svg' },
```
RemcoSimonides, line 10: This icon is not used?

libs/ui/src/lib/dashboard/pages/home/home.component.html
RemcoSimonides: Also, this button is full width of page. Make it as wide as its content and centered on the page.

libs/ui/src/lib/dashboard/pages/home/home.component.scss
```diff
@@ -1,27 +1,26 @@
...
 
-
-// No Movie
-mat-card {
-  width: 330px;
-  height: 220px;
-  img {
-    width: 100px;
-    align-self: center;
+.carrousel-container {
```
RemcoSimonides, line 15: Both carousel and carrousel are grammatically correct I think, but let's for consistency keep carousel since bf-carousel has one 'r' too.

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -31,22 +32,39 @@ export class HomeComponent implements OnInit {
...
+  orgActivity$ = this.titleAnalytics$.pipe(
+    map(analytics => counter(analytics, 'org.activity', 'orgActivity')),
+  );
+
+  territoryActivity$ = this.titleAnalytics$.pipe(
+    map(analytics => counter(analytics, 'org.addresses.main.country', 'territories')),
+  );
+
+  popularTitle$ = this.titleAnalytics$.pipe(
+    map(analytics => counter(analytics, 'meta.titleId', 'appName')),
```
RemcoSimonides, line 51: Not fully how I intended to use the `counter` method but I guess it works ^^
Maybe it's better to make this `scope` parameter of counter function optional

libs/ui/src/lib/dashboard/pages/home/home.component.ts
```diff
@@ -31,22 +32,39 @@ export class HomeComponent implements OnInit {
...
+    map(analytics => analytics.sort((a, b) => a.count - b.count)),
+    switchMap(([popularEvent]) => {
+      if (popularEvent) return this.movieService.valueChanges(popularEvent.key);
+      return NEVER;
```
RemcoSimonides, line 55: What is this `NEVER` here? In case there are no analytics events at all? This is in case there are no analytic events at all? NEVER is never resolving so that's not nice. Better to add a filter() instead.

RemcoSimonides (approved):

RemcoSimonides (changes requested):
Let's do request changes instead to be sure

RemcoSimonides (approved):

PR #8002, opened 2022-03-15
removed imports "@blockframes/model" by relative

KvnMrs:


libs/model/src/lib/contract.ts
```diff
@@ -1,9 +1,9 @@
-import { DocumentMeta } from "@blockframes/utils/models-meta";
+import { DocumentMeta, createDocumentMeta } from "@blockframes/utils/models-meta";
```
phpgeekfr, line 1: to make this library more coherent can you replace `"` by `'` in all files under `libs/model/src/lib` ? :)

libs/model/src/lib/user.ts
```diff
@@ -1,6 +1,8 @@
 import { DocumentMeta } from "@blockframes/utils/models-meta";
-import { createStorageFile, StorageFile, NotificationTypesBase, UserRole } from '@blockframes/model';
 import { Genre, Language, Media, Territory } from "@blockframes/utils/static-model";
+import { createStorageFile, StorageFile } from "./media";
```
phpgeekfr, line 3: some `"` are left ;)

phpgeekfr (approved):

Issue #7984, opened 2022-03-14
Release 4.1 - Import


RemcoSimonides:
[Import test scenarios](https://www.notion.so/cascade8/Import-78250f00339a4aa9a2983c03d85a0b6f)

Akita for Router has been removed in this release. Which means extra attention needs to be paid whether every link is still working properly and the correct app names are displayed (https://github.com/blockframes/blockframes/issues/7287)

Other issues that have been resolved for this release:
- [ ] https://github.com/blockframes/blockframes/issues/7810
- [ ] https://github.com/blockframes/blockframes/issues/7815
- [ ] https://github.com/blockframes/blockframes/issues/7225

cams-goussale:
- [ ] I can't upload any Contract, neither through `crm/import/contracts` nor `/c/o/dashboard/import`. There is a spinner turning eternally. I've tried to upload the new template version and the old one. Same happens

cams-goussale:
- [ ] After the test on @yspringfield local environment, I was able to import a Mandate, but not a Sale. I tried as a user on the app, and with the same user on CRM.

[Templates used for tests](https://drive.google.com/drive/folders/1S_fi8-jmVcRSGAurrI4zCg49KCsnTnDo)

Screenshot
![image.png](https://images.zenhubusercontent.com/617811d92b89e7733866d8a0/60ad4fdb-57ea-4ad8-9a59-e4f3a83ded36)

cams-goussale:
- [ ] Make column F of contract import not mandatory (color changed already on the [template](https://drive.google.com/drive/folders/1hmsmslJxLxrPpA5Cy5K814Rgd_bHYaQb))
- [ ] Make the correspondance between excel statuses to db statuses as follow:

```
'In Negotiation': 'negotiating'
'On signature': 'accepted'
'Signed': 'accepted'
'Accepted':'accepted'
'Declined':'declined'
```

phpgeekfr:
- [ ] if not already updated in template, update genres CF https://docs.google.com/spreadsheets/d/1BShPU5jcOajYCqEIAQaSshs3ojpFhOi0qHVnXTC5dQM/edit#gid=999482540 

```
export const genres = {
  action: 'Action',
  adventure: 'Adventure',
  animation: 'Animation',
  biography: 'Biography',
  comedy: 'Comedy',
  crime: 'Crime',
  documentary: 'Documentary',
  drama: 'Drama',
  erotic: 'Erotic',
  family: 'Family',
[...]

PR #8007, opened 2022-03-17
Fix contract import.

yspringfield:
Fixed: #7984


libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -242,13 +245,13 @@ export function getStaticList<S extends Scope>(scope: S, value: string, separato
...
+  territories: (territories, separator, required:boolean) => getStaticList('territories', territories, separator, 'Territories', required, 'world'),
+  medias: (medias, separator, required:boolean) => getStaticList('medias', medias, separator, 'Medias', required, 'all'),
 }
 
-export function getGroupedList(value: string, groupScope: 'territories', separator: string): FromStatic<'territories'>;
-export function getGroupedList(value: string, groupScope: 'medias', separator: string): FromStatic<'medias'>;
-export function getGroupedList<GS extends GroupScope>(value: string, groupScope: GS, separator: string) {
+export function getGroupedList(value: string, groupScope: 'territories', separator: string, options?:GroupedListOptions): FromStatic<'territories'>;
+export function getGroupedList(value: string, groupScope: 'medias', separator: string, options?:GroupedListOptions): FromStatic<'medias'>;
+export function getGroupedList<GS extends GroupScope>(value: string, groupScope: GS, separator: string, options={mandatory:true}) {
```
RemcoSimonides, line 254: Fix space here too
`options?: GroupListOptions `

`options = { mandatory: true }`

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -90,6 +96,18 @@ export class MovieService extends CollectionService<MovieState> {
     });
   }
 
+  getImportTitle({ name, userOrgId, blockframesAdmin, path}: ImportTitleOptions) {
```
GrandSchtroumpf, line 99: There is not reason to put that in the service as it's only used in the import system. Put it back in the import utils.

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -14,6 +14,9 @@ export interface SheetTab {
   headers: any[];
   rows: any[][];
 }
+interface GroupedListOptions{
+  mandatory:boolean;
```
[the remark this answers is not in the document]
GrandSchtroumpf: Why `mandatory` while everywhere else it's "required" ?

libs/import/src/lib/utils.ts
```diff
@@ -83,17 +83,16 @@ export async function getTitleId(
...
-      return ref.where('title.international', '==', name);
-    } else {
-      return ref
-        .where('title.international', '==', name)
-        .where('orgIds', 'array-contains', userOrgId);
-    }
-  });
-  const result = titles.length === 1 ? titles[0].id : '';
+  const options = { path: 'title.international', blockframesAdmin, userOrgId, name };
+  let titles = await titleService.getImportTitle(options);
```
GrandSchtroumpf, line 87: stop using `let` for something else than a string or a number, it's very difficult to understand.

libs/import/src/lib/utils.ts
```diff
@@ -74,28 +74,43 @@ export async function getOrgId(
...
-  cache[name] = result;
-  return result;
+}
+
+
+export async function getTitleId(
+  name: string,
+  titleService: MovieService,
+  cache: Record<string, Movie>,
+  memo: (name: string, value: Movie) => string,
```
GrandSchtroumpf, line 94: Why do you pass the memo function ? Define it in the scope of this function

libs/import/src/lib/view-extracted-elements/contract/utils.ts
```diff
@@ -111,21 +112,37 @@ export async function formatContract(
 ) {
   // Cache to avoid  querying db every time
   const orgNameCache: Record<string, string> = {};
-  const titleNameCache: Record<string, string> = {};
+  const titleNameCache: Record<string, Movie> = {};
   const userCache: Record<string, User> = {};
   const contractCache: Record<string, Mandate | Sale> = {};
+  const memo = (key: string, value: Movie) => {
```
GrandSchtroumpf, line 118: Move that to the getTitleId function as a closure

libs/movie/src/lib/movie/+state/movie.service.ts
```diff
@@ -27,6 +27,12 @@ export const fromInternalRef = (internalRef: string): QueryFn => (ref) =>
 type MovieWithAnalytics = Movie & { analytics: MovieAnalytics };
 
 interface MovieState extends EntityState<Movie, string>, ActiveState<string> { }
+export interface ImportTitleOptions {
```
GrandSchtroumpf, line 30: you don't need that anymore

GrandSchtroumpf (approved):

Issue #7978, opened 2022-03-14
Release 4.1 - Movie


RemcoSimonides:
[Movie test scenarios](https://www.notion.so/cascade8/Movie-8506fba333d64b8cb9e857351a6c7989)

Akita for Movie has been removed in this release. Which means that Movie should be tested thoroughly.
- [ ] Pay extra attention to testing details in Edit Movie page

Akita for Router has been removed in this release. Which means extra attention needs to be paid whether every link is still working properly and the correct app names are displayed (https://github.com/blockframes/blockframes/issues/7287)

Issues that have been completed in this release
- [ ] https://github.com/blockframes/blockframes/issues/7000
- [ ] https://github.com/blockframes/blockframes/issues/7001
- [ ] https://github.com/blockframes/blockframes/issues/7282
- [ ] #7255

Poppe-A:
- [ ] Change content type list in movie import template (there is only Movie and TV anymore)
![contentType](https://user-images.githubusercontent.com/35397614/158573504-bbfc929b-b56e-4528-8c40-e330d29579ca.png)

Poppe-A:
On CRM,files update is working but the button placement is not clear :
The button right above the image I'm trying to update isn't working, I have to click on the other one on the bottom left corner
![updatePicture](https://user-images.githubusercontent.com/35397614/158581269-47298a8b-e718-4af8-9c8b-f60b8e6f5d1e.jpg)

cams-goussale:
- [ ] I've seen some incoherences in the db, I'm not sure the migration went all the way. If you look at movie id `CuW01ZjIf2uipVlPUrPw` there is still the same genres showing in the db as before:

`adventure`
`periodPiece`

And if I look in-app, the movie doesn't show these genres anymore, which creates an Error.
![image](https://user-images.githubusercontent.com/93206363/158622811-cca95866-acce-4031-a251-507c8c2827f9.png)

![image](https://user-images.githubusercontent.com/93206363/158623650-dcd5b8a7-999c-45f2-9538-162a78b8af40.png)


What should have happened is:
- Adventure should still be in db and in-app
- PeriodPiece should be removed in db and in-app

phpgeekfr:
> * [ ]  I've seen some incoherences in the db, I'm not sure the migration went all the way. If you look at movie id `CuW01ZjIf2uipVlPUrPw` there is still the same genres showing in the db as before:

Indeed..
Theses genres have been moved to keywords but some where missed:

```
[...]

PR #8012, opened 2022-03-17
some fixes for movie

phpgeekfr:
part of #7978

### DIFF
--- a/libs/firebase-utils/src/lib/firestoreMigrations/0073.ts
+++ b/libs/firebase-utils/src/lib/firestoreMigrations/0073.ts
@@ -1,6 +1,7 @@
 import { Firestore } from '../types';
 import { runChunks } from '../firebase-utils';
 import { Movie } from '@blockframes/model';
+import { Genre } from '@blockframes/utils/static-model';
 
 /**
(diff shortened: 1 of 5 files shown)


libs/firebase-utils/src/lib/firestoreMigrations/0073.ts
```diff
@@ -24,15 +24,25 @@ export async function upgrade(db: Firestore) {
     tvShow: 'TV Show',
     virtualReality: 'Virtual Reality'
   };
+
+  const genresToRemove = 'other';
+  const genresToUpdate: Record<string, Genre> = {
+    'science-fiction': 'scienceFiction',
+    'periodPiece': 'periodDrama'
+  };
```
RemcoSimonides, line 32: Is this also updated in the excel template for import?

RemcoSimonides (approved):

libs/firebase-utils/src/lib/firestoreMigrations/0073.ts
phpgeekfr: updated issue https://github.com/blockframes/blockframes/issues/7984#issuecomment-1070801800

Issue #7986, opened 2022-03-14
Release 4.1 - Auth


RemcoSimonides:
[Auth test scenarios](https://www.notion.so/cascade8/Auth-a4eeff51e0d345ee83af6e211880c7fd)

Akita for Router has been removed in this release. Which means extra attention needs to be paid whether every link is still working properly and the correct app names are displayed (https://github.com/blockframes/blockframes/issues/7287)

Other issues that have been resolved in this release:
- [ ] https://github.com/blockframes/blockframes/issues/7936 (Shouldn't be able to edit setting in cookies preferences anymore)

cams-goussale:
- [ ] This is my review example

Poppe-A:
- [ ]  Try to navigate to a movie page without being logged in, you should be redirected to the log in page. But it redirects to the landing page

cams-goussale:
- [ ] Cut company name at xx characters and show "..." after that. No return to next line


![image](https://user-images.githubusercontent.com/93206363/158790304-21473beb-d6c6-46fb-a561-2d38cf3b3ba7.png)

RemcoSimonides:
- [ ] When a request to join an organization has been declined, the user who requested to join the organization receives as many emails as there are admins and superAdmins in the company. For quite some time, this email wasn't working at all. This issue came up while I was fixing it but it needs some extra attention to fix it properly.

phpgeekfr:
- [ ] login with dev+nathan-pdk@blockframes.io on https://staging.archipelmarket.com/c/o/dashboard/home go to https://staging.archipelmarket.com/c/o/organization/sLchj1Ib4Cxhwr0ZBW4m/view/org and go back to previous page and log out . Staying on same page with `FirebaseError: Missing or insufficient permissions.`

RemcoSimonides:
> * [x]  (A if someone reproduces this bug) When a user want to join an org, he doesn't get an email if it's request is accepted or cancelled [note camille: that's weird, it worked well for me. if user is declined, then yes, no email, that's the expected behavior. If user is accepted, he receives an email]

To Camille: the code says that an email should be send to the user who requested to join the org. There is an error in the backend which makes it not work.

RemcoSimonides:
> * [ ]  (C) Try to navigate to a movie page without being logged in, you should be redirected to the log in page. But it redirects to the landing page

Talked with Ana and decided to think over impact of this change and not rush it to put it in this release. https://github.com/blockframes/blockframes/issues/8027

[...]

PR #8014, opened 2022-03-17
7986 fixes for auth

RemcoSimonides:
#7986

### DIFF
--- a/apps/backend-functions/src/data/internals.ts
+++ b/apps/backend-functions/src/data/internals.ts
@@ -12,7 +12,7 @@ import { createDenomination, OrganizationDocument, PublicUser, InvitationDocumen
 
 export { getDocument };
 
-export function createPublicOrganizationDocument(org: Partial<OrganizationDocument>) {
+export function createPublicOrganizationDocument(org: OrganizationDocument) {
   return {
     id: org.id ?? '',
     denomination: createDenomination(org.denomination),
--- a/apps/backend-functions/src/data/internals.ts
+++ b/apps/backend-functions/src/data/internals.ts
@@ -12,7 +12,7 @@ import { createDenomination, OrganizationDocument, PublicUser, InvitationDocumen
 
 export { getDocument };
 
-export function createPublicOrganizationDocument(org: OrganizationDocument) {
+export function createPublicOrganizationDocument(org: Partial<OrganizationDocument>) {
   return {
     id: org.id ?? '',
     denomination: createDenomination(org.denomination),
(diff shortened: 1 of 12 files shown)


apps/backend-functions/src/data/internals.ts
```diff
@@ -37,7 +37,8 @@ export function createPublicUserDocument(user: Partial<PublicUser> = {}) {
     avatar: createStorageFile(user.avatar),
     firstName: user.firstName ?? '',
     lastName: user.lastName ?? '',
-    orgId: user.orgId ?? ''
+    orgId: user.orgId ?? '',
+    _meta: createDocumentMeta(user._meta as DocumentMeta<Timestamp>)
```
phpgeekfr, line 41: in some cases, user._meta will be undefined @see `onRequestFromUserToJoinOrgDecline`.

In other cases where it is defined, is it always a Timestamp ?

phpgeekfr: and do we really want organization._meta to be included in notifications for example ?
```
const notification = createNotification({
      toUserId: context.auth.uid,
      organization: createPublicOrganizationDocument(organization),
      user: createPublicUserDocument(userDocument),
      _meta: createDocumentMeta({ createdFrom: data.app }),
      type: 'userRequestAppAccess'
    });
```

libs/model/src/lib/user.ts
```diff
@@ -56,7 +56,8 @@ export function createPublicUser(user: Partial<User> = {}): PublicUser {
     avatar: createStorageFile(user?.avatar),
     firstName: user.firstName ?? '',
     lastName: user.lastName ?? '',
-    orgId: user.orgId ?? ''
+    orgId: user.orgId ?? '',
+    _meta: createDocumentMeta(user?._meta as DocumentMeta<Date>)
```
phpgeekfr, line 60: createPublicUser() is used in backend and auth. looks weird to handle only Date

And this will include _meta in notifications, is it somehting we want ?

apps/backend-functions/src/internals/invitations/organizations.ts
```diff
@@ -85,6 +85,7 @@ async function onInvitationToOrgDecline(invitation: InvitationDocument) {
     createNotification({
       toUserId: toAdminId,
       user: createPublicUserDocument(invitation.toUser),
+      organization: createPublicOrganizationDocument(invitation.fromOrg),
       type: 'invitationToJoinOrgDeclined',
       _meta: createDocumentMeta({ createdFrom: appAccess })
```
phpgeekfr, line 90: we are removing `requestToJoinOrgDeclined` but keeping `invitationToJoinOrgDeclined` right ?

libs/model/src/lib/user.ts
```diff
@@ -1,4 +1,4 @@
-import { DocumentMeta } from "@blockframes/utils/models-meta";
+import { DocumentMeta, createDocumentMeta } from "@blockframes/utils/models-meta";
```
phpgeekfr, line 1: I think you don't need `createDocumentMeta`

phpgeekfr (approved):

apps/backend-functions/src/internals/invitations/organizations.ts
RemcoSimonides: I'll discuss this with Camille and create new PR for it if we want to remove it

PR #8015, opened 2022-03-17
Add space between mat-cards

KvnMrs:


phpgeekfr (commented):
not the same spacing between each elements : 
image

Also not the same as other tab 
image

Issue #7981, opened 2022-03-14
Release 4.1 - General UI


RemcoSimonides:
[General UI test scenarios](https://www.notion.so/cascade8/General-UI-db403143e36c480d9c6b562cc001119f)

Akita for Router has been removed in this release. Which means extra attention needs to be paid whether every link is still working properly and the correct app names are displayed (https://github.com/blockframes/blockframes/issues/7287)

Other issues that have been resolved for this release:
- [x] https://github.com/blockframes/blockframes/issues/7654
- [x] https://github.com/blockframes/blockframes/issues/7733

phpgeekfr:
- [ ] invitation list is buggy (user dev+nathan-pdk@blockframes.io)
 ![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/da572eab-4b15-49ad-942a-a8af7ce810fe)

RemcoSimonides:
- [ ] "Catalogue" should be "catalog"
<img width="1038" alt="image" src="https://user-images.githubusercontent.com/27687382/158556766-b5342ada-2cc7-4393-9982-ccecdde3b450.png">

RemcoSimonides:
- [ ] I don't know if it was in the scope of this issue but it doesn't seem to be responsive for mobile. Margin is missing and button should maybe go to next line?
<img width="389" alt="image" src="https://user-images.githubusercontent.com/27687382/158558439-8bb967fe-9971-4386-ae73-bb6889e0eee2.png">

phpgeekfr:
- [ ] mail icon is supposed to have a red circle with number 3  (dev+nathan-pdk@blockframes.io)
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/1189d60c-030d-4160-b89e-bb188e421cec)

RemcoSimonides:
- [ ] Notifications are displayed double. To reproduce: Created event with super admin (dev+nathan-pdk@blockframes.io) and invited a member of the same org to this event (rsimonides@cascade8.com).
<img width="1384" alt="image" src="https://user-images.githubusercontent.com/27687382/158571148-f8a6abda-311d-40c0-8633-403cb41baf66.png">

cams-goussale:
- [ ] Text modification on the landing page. Under the Buyer tab, in the introduction (next to the animation):
[...]

PR #8016, opened 2022-03-17
4.1 general UI

Poppe-A:
#7981


libs/invitation/src/lib/invitation.component.ts
```diff
@@ -28,7 +27,10 @@ export class InvitationComponent {
   });
 
   // Invitation count for conditions
-  invitationCount$ = this.service.myInvitations$.pipe(map(inv => inv.length));
+  invitationCount$ = this.service.myInvitations$.pipe(
+    map(inv => this.checkInvitationDate(inv)),
```
RemcoSimonides, line 31:
```
import { subMonths } from 'date-fns'

map(invitations => {
  /**
  * Filtering out invitations older than 4 months because there is no timeFrame supporting them in invitation-list component.
  */ 
   const fourMonthsAgo = subMonths(new Date(), 4);
   return invitations.filter(invitation => invitation.date > fourMonthsAgo);
}
```

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -28,7 +27,10 @@ export class InvitationComponent {
...
+    map(inv => inv.length)
+  );
 
   // Invitation that require an action
   invitations$ = combineLatest([
```
RemcoSimonides, line 36: Am I missing something or should you also be filtering this variable?

libs/invitation/src/lib/+state/invitation.service.ts
```diff
@@ -87,7 +87,9 @@ export class InvitationService extends CollectionService<InvitationState> {
   ]).pipe(
     map(([user, invitations]) => {
       if (!user?.uid || !user?.orgId) return [];
-      return invitations.filter(i => i.toOrg?.id === user.orgId || i.toUser?.uid === user.uid)
+      const inv = invitations.filter(i => i.toOrg?.id === user.orgId || i.toUser?.uid === user.uid);
+      const removeDuplicate = inv.filter((i, index) => inv.findIndex(elm => elm.id === i.id) === index)
```
phpgeekfr, line 91: missing `;` 
and do :
```
return inv.filter((i, index) => inv.findIndex(elm => elm.id === i.id) === index);
``` 
As Francois asked some days ago in slack

But I don't get how we can have duplicates here .. an invitation cannot have toOrg and toUser at the same time

libs/invitation/src/lib/invitation.component.ts
phpgeekfr: was this validated by UX / business team ?

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -28,7 +28,16 @@ export class InvitationComponent {
   });
 
   // Invitation count for conditions
-  invitationCount$ = this.service.myInvitations$.pipe(map(inv => inv.length));
+  invitationCount$ = this.service.myInvitations$.pipe(
+    map(invitations => {
+      /**
+      * Filtering out invitations older than 4 months because there is no timeFrame supporting them in invitation-list component.
+      */ 
+      const fourMonthsAgo = subMonths(new Date(), 4);
```
phpgeekfr, line 36: you could have defined `const fourMonthsAgo = subMonths(new Date(), 4);` elsewhere, no need to declare for each invitation

phpgeekfr (approved):

RemcoSimonides (approved):

libs/invitation/src/lib/invitation.component.ts
RemcoSimonides: It's not triggered for each invitation here. Rxjs map doesn't loop through the array.
That's why it's outside of the `filter` function 👇

PR #8023, opened 2022-03-17
Fix order of promotional elements in sidebar

KevinDurand974:
From issue: https://github.com/blockframes/blockframes/issues/8022

### DIFF
--- a/libs/movie/src/lib/movie/form/shell/shell.component.ts
+++ b/libs/movie/src/lib/movie/form/shell/shell.component.ts
@@ -71,10 +71,12 @@ function getSteps(status: ProductionStatus, appSteps: TunnelStep[] = []): Tunnel
       {
         path: 'media-images',
         label: 'Images'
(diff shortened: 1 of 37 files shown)


libs/movie/src/lib/movie/form/shell/shell.component.ts
```diff
@@ -67,17 +67,17 @@ function getSteps(status: ProductionStatus, appSteps: TunnelStep[] = []): Tunnel
...
       {
         path: 'media-images',
         label: 'Images'
       }, {
         path: 'media-videos',
         label: 'Videos'
+      }, {
+        path: 'media-notes',
+        label: 'Notes & Statements',
+        shouldHide: isStatus(status, ['post_production', 'finished', 'released'])
```
phpgeekfr, line 80: you can correct the bad formatting if you want

phpgeekfr (approved):

Issue #7979, opened 2022-03-14
Release 4.1 - Organization


RemcoSimonides:
[Organisation test scenarios](https://www.notion.so/cascade8/Organisation-55ded6f0e09c4aa29fb2e3f716bb8b50)

Akita for Router has been removed in this release. Which means extra attention needs to be paid whether every link is still working properly and the correct app names are displayed (https://github.com/blockframes/blockframes/issues/7287)

KvnMrs:
- [ ] missing space between mat-cards ![error_Css.PNG](https://images.zenhubusercontent.com/61712a8c95d99512a311887c/57cf857f-5685-4f2d-85c5-505aaee66701)

cams-goussale:
- [ ] Button return on org page is not working

https://user-images.githubusercontent.com/93206363/158563316-658e0d39-f7ba-4619-bed4-7318fb46879e.mp4

KevinDurand974:
- [ ] The new user waiting the approval of the organisation is blocked on `/c/organization/join-congratulations`

How to:
1. Create an account with an org, don't verify his email
2. Admin or Super Admin decline his request
3. User verify is email (box will be checked)
4. User waiting indefinilty for org approval or decline
His solution is to click on logo button to go on home to exit this page

<details>
    <summary>Image :</summary>
    <img src="https://i.imgur.com/t8AWGP1.png" />
</detail>

KevinDurand974:
## Can update org information and settings

> Org Page ( /c/o/dashboard/crm/organization/{{ orgId }} )

* * *

- [ ] Overview of movies is displayed

Organization: **Barton, Heller and Lueilwitz (Barton, Heller and Lueilwitz)**

Movie: **Infernus, Night Has Fallen and Abyzou**

Problem: No movie displayed

<details>
  <summary><u>Image</u></summary>
  <br />
  <div><small>Movies</small></div>
  <img src="https://i.imgur.com/jctoAlF.png" />
</details>

* * *

- [ ] Style CSS - Organization ressource

<details>
  <summary><u>Images</u></summary>
  <br />
  <div><small>Company</small></div>
  <img src="https://i.imgur.com/z2fjRcX.png" />

  <br />
  <div><small>Titles</small></div>
  <img src="https://i.imgur.com/xXaMioR.png" />
</details>

phpgeekfr:
> * (B) **KevinM** missing space between mat-cards

Looks like this is only on chrome

[...]

PR #8042, opened 2022-03-18
fixes for 7981 and 7979

RemcoSimonides:
#7981
- [x] When trying to validate a picture with the dedicated button, clicking close to the border of the button activate the 'click effect' but without the 'validate action'. It's the same with the other picture related buttons (like delete). The button's hitbox should be the circle and not just the icon inside, the user could think the app isn't working properly :
Validating picture

#7979
- [x] (A) Overview of org members is displayed
- [x] (C) Style CSS - Organization ressource

### DIFF
--- a/apps/crm/crm/src/app/crm/pages/organization/organization.component.ts
+++ b/apps/crm/crm/src/app/crm/pages/organization/organization.component.ts
@@ -97,20 +97,18 @@ export class OrganizationComponent implements OnInit {
   }
 
   private async getMembers() {
-    const [ members, role ] = await Promise.all([
+    const [members, role] = await Promise.all([
       this.organizationService.getMembers(this.orgId),
       this.permissionService.getValue(this.orgId)
     ]);
-    return members.map((m) => {
-      return {
-        ...createOrganizationMember(m, role.roles[m.uid] ? role.roles[m.uid] : undefined),
-        userId: m.uid,
-        edit: {
-          id: m.uid,
-          link: `/c/o/dashboard/crm/user/${m.uid}`,
-        },
-      };
-    });
+    return members.map((m) => ({
+      ...createOrganizationMember(m, role.roles[m.uid] ? role.roles[m.uid] : undefined),
+      userId: m.uid,
+      edit: {
+        id: m.uid,
+        link: `/c/o/dashboard/crm/user/${m.uid}`,
+      },
+    }));
   }
 
   public async update() {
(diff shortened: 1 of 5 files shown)


apps/crm/crm/src/app/crm/pages/organization/organization.component.ts
```diff
@@ -97,15 +97,20 @@ export class OrganizationComponent implements OnInit {
...
-    const members = await this.organizationService.getMembers(this.orgId);
-    return members.map((m) => ({
-      ...m,
-      userId: m.uid,
-      edit: {
-        id: m.uid,
-        link: `/c/o/dashboard/crm/user/${m.uid}`,
-      },
-    }));
+    const [ members, role ] = await Promise.all([
```
phpgeekfr, line 100: bad formatting ;)

apps/crm/crm/src/app/crm/pages/organization/organization.component.ts
```diff
@@ -97,15 +97,20 @@ export class OrganizationComponent implements OnInit {
...
+    return members.map((m) => {
+      return {
+        ...createOrganizationMember(m, role.roles[m.uid] ? role.roles[m.uid] : undefined),
+        userId: m.uid,
+        edit: {
+          id: m.uid,
+          link: `/c/o/dashboard/crm/user/${m.uid}`,
+        },
+      };
+    });
```
phpgeekfr, line 113:
```suggestion
    return members.map((m) => ({
        ...createOrganizationMember(m, role.roles[m.uid] ? role.roles[m.uid] : undefined),
        userId: m.uid,
        edit: {
          id: m.uid,
          link: `/c/o/dashboard/crm/user/${m.uid}`,
        },
      });
```

libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -158,14 +158,11 @@ export class OrganizationService extends CollectionService<OrganizationState> {
     return this.update(orgId, { userIds });
   }
 
-  public async getMembers(orgId: string): Promise<OrganizationMember[]> {
+  public async getMembers(orgId: string): Promise<PublicUser[]> {
```
phpgeekfr, line 161: all parts of the code that uses `getMembers(` does not need the role attribute anymore ?

phpgeekfr (approved):

libs/organization/src/lib/organization/+state/organization.service.ts
RemcoSimonides: That's right

PR #8043, opened 2022-03-18
4.1 Event fixes

### LINKED ISSUE #7980: Release 4.1 - Events
[Screening test scenarios](https://www.notion.so/cascade8/Screening-69f2b3a496394341aa16227d58109eb4)
[Meeting test scenarios](https://www.notion.so/cascade8/Meeting-7d16f2abef1a4b5b85b5a1889dfa1eab)
[Interactions](https://www.notion.so/cascade8/Interactions-244e704f695d45bf9355eebc49e541ce)

Akita for Router has been removed in this release. Which means extra attention needs to be paid whether every link is still working properly and the correct app names are displayed (https://github.com/blockframes/blockframes/issues/7287)

Poppe-A:
#7980


libs/invitation/src/lib/form/user/user.component.ts
```diff
@@ -116,7 +116,7 @@ export class UserComponent implements OnInit {
         this.sending.next(false);
       } catch (error) {
         this.sending.next(false);
-        this.snackBar.open(error.message, 'close', { duration: 5000 });
+        this.snackBar.open('Something went wrong ...', 'close', { duration: 5000 });
```
RemcoSimonides, line 119: I think it's important to know what caused the 'internal' error and fix that instead.

libs/movie/src/lib/movie/components/request-asking-price/request-asking-price.component.ts
```diff
@@ -20,7 +20,7 @@ export class RequestAskingPriceComponent {
     territories: new FormStaticValueArray<'territories'>([], 'territories', [Validators.required]),
     message: new FormControl()
   });
-
+  public disableSendButton: boolean;
```
RemcoSimonides, line 23: `disableSendButton = false;`

☝️ I also don't specify `public` because it's default by public and I see that the other variable in this class (form) also doesn't specify it. So I'd keep it consistent

RemcoSimonides (changes requested):

libs/event/src/lib/guard/event-access.guard.ts
```diff
@@ -61,8 +61,12 @@ export class EventAccessGuard implements CanActivate {
 
           return true;
         } else {
-          this.snackBar.open('Missing invitation parameter', 'close', { duration: 5000 });
-          await this.authService.deleteAnonymousUser();
+          this.snackBar.open('Sorry, it seems that you were not invited to this event', 'Try with other mail', { duration: 6000 })
```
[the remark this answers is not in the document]
Poppe-A: It was a request from Anastasiia

libs/invitation/src/lib/form/user/user.component.ts
Poppe-A: That was also a request from Anastasiia, and I couldn't reproduce the error

RemcoSimonides (approved):

PR #8045, opened 2022-03-18
Button return on org page is not working

phpgeekfr:


RemcoSimonides (approved):
We should implement this everywhere we use this ```this.location.historyGo(-this.countRouteEvents);```

phpgeekfr:
> We should implement this everywhere we use this `this.location.historyGo(-this.countRouteEvents);`

Created #8046

PR #7903, opened 2022-03-03
multi-term mandates avails research

yspringfield:
Fixed: #7887
Fixed: #7919

PR #8049, opened 2022-03-21
Revert "multi-term mandates avails research"

yspringfield:
Reverting: #7903


phpgeekfr:
@yspringfield will this revert have to be done on develop also once we merge back ?

PR #8068, opened 2022-03-22
Check for duplicates on movie imports

yspringfield:
Fixed: https://github.com/blockframes/blockframes/issues/7984#issuecomment-1074139291

### DIFF
--- a/libs/import/src/lib/view-extracted-elements/titles/utils.ts
+++ b/libs/import/src/lib/view-extracted-elements/titles/utils.ts
@@ -9,6 +9,7 @@ import {
   adminOnlyWarning,
   getUser,
   unknownEntityError,
+  alreadyExistError,
 } from '@blockframes/import/utils';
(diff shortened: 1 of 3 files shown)


libs/import/src/lib/view-extracted-elements/titles/utils.ts
```diff
@@ -150,17 +158,22 @@ export async function formatTitle(
...
     },
     /* b */ 'title.original': (value: string) => {
       // ! required
       if (!value) return mandatoryError('Original Title');
       return value;
     },
-    /* c */ internalRef: (value: string) => {
+    /* c */ internalRef: async (value: string) => {
       if (!value) return optionalWarning('Internal Ref');
+      const isDuplicate = await getMovie('Title with internalRef', value, titleService);
```
GrandSchtroumpf, line 174: How can you query the attribute `'Title with internalRef'` ?

GrandSchtroumpf (approved):

PR #8078, opened 2022-03-23
fixed: remove verification of duplicate movie imports

yspringfield:
Fix: https://github.com/blockframes/blockframes/issues/7984#issuecomment-1075331416

### DIFF
--- a/libs/import/src/lib/view-extracted-elements/titles/utils.ts
+++ b/libs/import/src/lib/view-extracted-elements/titles/utils.ts
@@ -140,17 +140,9 @@ interface FieldsConfig {
 
 type FieldsConfigType = ExtractConfig<FieldsConfig>;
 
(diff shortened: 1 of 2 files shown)


libs/import/src/lib/view-extracted-elements/titles/utils.ts
```diff
@@ -164,8 +156,6 @@ export async function formatTitle(
   const fieldsConfig: FieldsConfigType = {
     /* a */ 'title.international': async (value: string) => {
       if (!value) return optionalWarning('International Title');
-      const movie = await getMovie('title.international', value, titleService, titleCache);
-      if (movie) return alreadyExistError(`Title with name(${value})`);
```
phpgeekfr, line 168: I think you don't need `alreadyExistError`  import anymore

phpgeekfr (approved):

Issue #7581, opened 2022-01-20
Don't display concierge users on the Sellers' member list


Jackseed:
start from: https://github.com/blockframes/blockframes/pull/8085

Url: /c/o/marketplace/organization/{{orgId}}/member

### Description
Concierges are our own way to support our users in their daily use of the apps, they shouldn't be listed as real team members for sellers. 
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/3302356b-a3b1-4322-9087-f95dfbb1b0b7)

### Todo
- [ ] Remove concierge from sellers' member list using either the email address (always starting with concierge+, always ending with {{@appUrl}}) or their first name (to double check but supposedly always "Concierge")

PR #8085, opened 2022-03-23
Remove concierge from sellers' member

KvnMrs:
issue: https://github.com/blockframes/blockframes/issues/7581

Duplicate getMembers because it used also in `apps\crm\crm\src\app\crm\pages\organization\organization.component.ts` .
I add a filter in my duplication for doesn't find back emails including  `'concierge+' || 'archipelmarket' || 'archipelcontent' || 'mediafinanciers'` .


libs/organization/src/lib/organization/+state/organization.service.ts
```diff
@@ -166,6 +166,17 @@ export class OrganizationService extends CollectionService<OrganizationState> {
...
+  public async getMembersByFilterEmail(orgId: string): Promise<OrganizationMember[]> {
+    const org = await this.getValue(orgId);
+    const promises = org.userIds.map((uid) => this.userService.getValue(uid));
+    const users = await Promise.all(promises);
+    const usersFilteredByEmail = users.filter(user => !user.email.includes('concierge+' || 'archipelmarket' || 'archipelcontent' || 'mediafinanciers') )    
+    const role = await this.permissionsService.getValue(orgId);
+    return usersFilteredByEmail.map((u) =>
+      createOrganizationMember(u, role.roles[u.uid] ? role.roles[u.uid] : undefined)
+    );  
+  }
```
phpgeekfr, line 178: instead of recreating the whole method you could have done something like this :
```
  public async getMembersByFilterEmail(orgId: string): Promise<OrganizationMember[]> {
    const members = await this.getMembers(orgId);
    return members.filter(user => !user.email.includes('concierge+'));
  }
```

But let's update the existing `getMembers` method instead: 
```
  public async getMembers(orgId: string, options?: { removeConcierges: boolean }): Promise<OrganizationMember[]> {
    const org = await this.getValue(orgId);
    const promises = org.userIds.map((uid) => this.userService.getValue(uid));
    const users = await Promise.all(promises);
    const role = await this.permissionsService.getValue(orgId);
    return users.map((u) =>
      createOrganizationMember(u, role.roles[u.uid] ? role.roles[u.uid] : undefined)
    ).filter(member => options?.removeConcierges ? !member.email.includes('concierge+') : true);
  }
```
and instead of doing `this.orgService.getMembersByFilterEmail(org.id)` do `this.orgService.getMembers(org.id, { removeConcierges: true })`

Also I think that filtering on `concierge+` only is enough

apps/festival/festival/src/app/marketplace/organization/member/member.component.ts
```diff
@@ -1,6 +1,6 @@
 import { Component, OnInit, ChangeDetectionStrategy, HostBinding } from '@angular/core';
 import { ViewComponent } from '../view/view.component';
-import { switchMap } from 'rxjs/operators';
+import { switchMap, map } from 'rxjs/operators';
```
phpgeekfr, line 3: I don't think you need to import `map`

apps/festival/festival/src/app/marketplace/organization/member/member.component.ts
```diff
@@ -17,8 +17,12 @@ export class MemberComponent implements OnInit {
 
   public org$ = this.parent.org$;
   public members$ = this.org$.pipe(
-    switchMap(org => this.orgService.getMembers(org.id))
+    switchMap(org => {
+      return this.orgService.getMembersByFilterEmail(org.id)     
+    })
   );
+  
+
```
phpgeekfr, line 25: extra lines

RemcoSimonides:
@phpgeekfr this PR is going to cause some merge conflicts because I've changed this method in `fix-for-release-4-1` too

phpgeekfr:
ok, let's wait for after the merge back then

phpgeekfr (approved):

Issue #7836, opened 2022-02-23
(A-B) Sellers' Analytics - Title Activity Page - Make graphs clickable and filter table


RemcoSimonides:
- [x] Clicking on Company Activity Pie chart filters the table with this activity type
- [x] Clicking on a Country in the map filters the table with this country

![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/35d39e1e-f4d5-46fd-a7e8-55f9c7c79d19)

PR #8094, opened 2022-03-24
7836 filter table on pie chart click

RemcoSimonides:
Part 1 of #7836
image

### DIFF
--- a/apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
+++ b/apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
@@ -35,12 +35,12 @@
       </a>
 
       <div fxLayout fxLayout.lt-sm="column" fxLayoutGap="16px">
-        <analytics-pie-chart fxFlex class="surface" [data]="orgActivity$ | async"></analytics-pie-chart>
+        <analytics-pie-chart fxFlex class="surface" [data]="orgActivity$ | async" (selection)="filterValue = $event"></analytics-pie-chart>
         <analytics-map fxFlex class="surface" [data]="territoryActivity$ | async"></analytics-map>
       </div>
     </article>
 
-    <bf-table [source]="buyerAnalytics$ | async" pagination="5" useFilter>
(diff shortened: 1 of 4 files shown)


apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -50,7 +50,7 @@ <h2>Interested Users</h2>
       <ng-template colRef="org.denomination.public" label="Company name" let-name sort>
         {{ name }}
       </ng-template>
-      <ng-template colRef="org.activity" label="Activity" let-activity sort>
+      <ng-template colRef="org.activity" label="Activity" let-activity sort [filter]="getFilter('orgActivity')">
```
GrandSchtroumpf, line 53: prefer observable or pipe over methods for an input in order to avoid unexpected rerendering

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.ts
```diff
@@ -61,6 +62,8 @@ export class TitleAnalyticsComponent {
     })
   )
 
+  filterValue: string;
```
GrandSchtroumpf, line 65: put a `?` if don't initialize it.

libs/analytics/src/lib/components/pie-chart/pie-chart.component.ts
```diff
@@ -64,4 +72,6 @@ export class PieChartComponent {
     this.chart?.updateOptions(this.pieChartOptions);
   }
 
+  private previousSelection: string;
```
GrandSchtroumpf, line 75: put `?` if you don't initialize it.

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
RemcoSimonides: `getFilter()` is returning a function though
Bit odd to return a function in an observable

Issue #6844, opened 2021-10-11
Event - Slate Presentation


RemcoSimonides:
Link to design issue #5397

[Zeplin](https://app.zeplin.io/project/61129744b469ce8a0c470c3e/screen/6137610c5416a6b9a2f9b603)
Add Slate Presentation to the Event Type list
![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/614da846-b645-4ef9-bb0d-e8a5b586c099)

[Zeplin](https://app.zeplin.io/project/61129744b469ce8a0c470c3e/screen/612c942b62c57412f5ee8ce4)
![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/5d01c8ff-f334-4ae1-9838-e8c14500f0bf)

Event page
[Zeplin](https://app.zeplin.io/project/61129744b469ce8a0c470c3e/screen/612c942be064d714dbc44c68) - old
[Figma](https://www.figma.com/file/HQ2FBvm2u0vyWO0QpIKeTs/%E2%9C%85-Event-Form-Structure-(IN-PROGRESS)?node-id=0%3A1)

As we are doing the Slate Presentation and Events with no connexion almost in parallel we need to pay attention to the Privacy Settings block. For the new Events with no connexion representation, the Privacy Settings will appear inside the Event Page ([example](https://www.figma.com/file/HQ2FBvm2u0vyWO0QpIKeTs/[-Event-Form-Structure-]?node-id=2%3A3443))

RemcoSimonides:
Bruce: 2 week 1 dev
Remco: 2 week
PL: 2 week 1 dev
Francois: 2 week 2 dev

Poppe-A:
Last Figma version : 
https://www.figma.com/file/HQ2FBvm2u0vyWO0QpIKeTs/%E2%9C%85-Event-Form-Structure-(IN-PROGRESS)?node-id=0%3A1

cams-goussale:
@Poppe-A here is a first review for Slate Presentation

- [ ] (B) On the mini-creation form: change `Slate` to `Slate Presentation`

![slate.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/dd7ec29c-2e8d-4aee-bed3-0460a0874d79)

- [ ] (B) Here we need to keep only with the underline and remove bold, following buttons and links rules

[...]

PR #8105, opened 2022-03-25
slate event view component

Poppe-A:
#6844


apps/festival/festival/src/app/event/view/view.component.html
```diff
@@ -14,7 +14,7 @@ <h6>
               <event-agenda-export [event]="event" color="primary"></event-agenda-export>
             </h6>
             <pre class="mat-body-1">{{ event | eventRange }}</pre>
-            <p class="mat-body-2">{{ event.meta.description }}</p>
+            <p *ngIf="event.type === 'screening'" class="mat-body-2">{{ event.meta.description }}</p>
```
RemcoSimonides, line 17: Meeting also has `event.meta.description`? I think it shouldn't be hidden for meeting
Maybe you should ask design team if it's okay to show the description here too for Slate. Saying this because if the designs for each event type start to differentiate too much, then it's better to create separate components for each.

apps/festival/festival/src/app/event/view/view.component.html
```diff
@@ -38,7 +38,7 @@ <h6>
               <event-agenda-export [event]="event"></event-agenda-export>
             </h6>
             <pre class="mat-body-1">{{ event | eventRange }}</pre>
-            <p class="mat-body-2">{{ event.meta.description }}</p>
+            <p *ngIf="event.type === 'screening'" class="mat-body-2">{{ event.meta.description }}</p>
```
RemcoSimonides, line 41: Same as above

apps/festival/festival/src/app/event/view/view.component.html
```diff
@@ -125,5 +125,49 @@ <h5>Description</h5>
...
+                  <span>{{ user | displayName }}{{ user | displayPosition }}</span>
+                </ng-container>
+              </div>
+            </ng-container>
+            <ng-template #anonymous>
+              <img [ref]="event.org.logo" asset="empty_organization.svg" type="logo" alt="Organization Logo">
+              <div fxLayout="column" fxLayoutGap="4px">
+                <h6>{{ event.org | orgName }}</h6>
+                <ng-container *ngIf="event.organizedBy as user">
+                  <span>{{ user | displayName }}{{ user | displayPosition }}</span>
```
RemcoSimonides, line 149: It would be nice to have some space between username and position.
Also fix for meeting since this code is a duplicate of that.

apps/festival/festival/src/app/event/view/view.component.html
```diff
@@ -125,5 +125,49 @@ <h5>Description</h5>
...
+          </div>
+          <ng-container *ngIf="!!event.meta.description">
+            <h5>Description</h5>
+            <p>{{ event.meta.description }}</p>
+          </ng-container>
+        </article>
+        <article *ngIf="slateTitles$ | async as titles" >
+          <ng-container *ngIf="titles.length">
+            <h2>Titles in presentation</h2>
+            <bf-carousel flex [min]="4">
```
RemcoSimonides, line 162: `[min]="4"` doens't look good on small screen. Isn't it possible to make it responsive?

apps/festival/festival/src/app/event/view/view.component.scss
```diff
@@ -77,4 +79,13 @@ event-request-screening {
       border-radius: 50%;
     }
   }
+}
+.slate {
+  article {
+    padding: 16px;
+  }
+  movie-card {
+    width: calc(25% - 24px);
```
RemcoSimonides, line 88: Make it responsive

apps/festival/festival/src/app/event/view/view.component.scss
```diff
@@ -1,3 +1,5 @@
+@import '~@angular/flex-layout/mq';
```
RemcoSimonides, line 1: Is this needed?

apps/festival/festival/src/app/event/view/view.component.ts
```diff
@@ -83,6 +86,11 @@ export class EventViewComponent implements OnInit {
         }
       })
     );
+    
+    this.slateTitles$ = this.event$.pipe(
+      map(event => event as Event<Slate>),
```
RemcoSimonides, line 91: This event can be any type of event. Use a Pipe instead to get the titles. In the html you know already it's of type Slate, here you don't. `event.meta.titles | getTitle | async`

Sorry for not noting this last time, but can you change the interface to
```
export interface Slate {
  description: string;
  organizerUid: string;
  titleIds: string[];
  video: string;
}
```
☝️ Note the `string[]`, the `;` at the end and `titleIds` instead of `titles` since it's only the ids, not the titles themselves.

Also, next time you should do `switchMap((event: Event<Slate>) => ...` instead of this map

RemcoSimonides (changes requested):

apps/festival/festival/src/app/event/view/view.component.ts
Poppe-A: I also changed `video` for `videoId` since its not the whole `videoStorage` object, but just the id

apps/festival/festival/src/app/event/view/view.component.html
Poppe-A: Ok. I'll correct it with the second choice (discussed with Ana:. We put the event description in the header. Then UX/UI team will possibly do a rework on the header

apps/festival/festival/src/app/event/view/view.component.scss
Poppe-A: It is now for the media query

RemcoSimonides (approved):

PR #8112, opened 2022-03-25
Sellers analytics

RemcoSimonides:


apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -0,0 +1,79 @@
+<ng-container *ngIf="title$ | async as title">
```
GrandSchtroumpf, line 1: it looks like you only need the title for the article in the header. Might be a good idea to only wrap it instead of all the page.

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -0,0 +1,79 @@
...
+          <a mat-button [routerLink]="['/c/o/marketplace/title/', title.id]" class="film-page-button">
+            <span>See Film Page</span>
+          </a>
+        </div>
+        <h6>{{ title.genres | toLabel: 'genres' }}</h6>
+      </div>
+    </article>
+  </header>
+
+  <section fxLayout="column" fxLayoutGap="16px">
```
GrandSchtroumpf, line 25: I would like you to start using more css instead of angular directive. The `gap` property seems to have a pretty good support these day. Might be time to use it : 
https://caniuse.com/flexbox-gap

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.scss
```diff
@@ -0,0 +1,22 @@
...
+  img {
+    height: 200px;
+  }
+}
+
+section {
+  padding: 24px;
+}
+
+.mat-button {
```
GrandSchtroumpf, line 19: In all of my project I never had to do that. So for me there is a design issue.

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (!aggregator[analytic.user.uid]) {
+          aggregator[analytic.user.uid] = createAggregatedAnalytic({
+            user: analytic.user,
+            org: analytic.org
+          });
+        };
+        aggregator[analytic.user.uid][analytic.name]++;
+      }
+      return Object.values(aggregator);
+    })
```
GrandSchtroumpf, line 69: move this part of the code in a function above the component with a good naming.

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.ts
```diff
@@ -0,0 +1,93 @@
...
+    private location: Location,
+    private movieService: MovieService,
+    private orgService: OrganizationService,
+    private route: ActivatedRoute,
+    private analyticsService: AnalyticsService,
+    private userService: UserService,
+  ) {}
+
+  goBack() {
+    this.location.back();
```
GrandSchtroumpf, line 87: I saw some piece of code that checked the history of the navigation to check if you actually have a `back`. Might be a good idea to use it here too. Ideally we should factorize it. @phpgeekfr should know where it is.

apps/festival/festival/src/app/dashboard/analytics/titles/titles-analytics.component.ts
```diff
@@ -0,0 +1,62 @@
...
+      const aggregated = createAggregatedPerTitle({
+        title,
+        screenings: title.events?.length
+      });
+      if (!title.analytics) return aggregated;
+      for (const analytic of title.analytics) {
+        aggregated[analytic.name]++;
+      }
+      return aggregated;
+    }))
```
GrandSchtroumpf, line 48: move that outside of the component and give a good naming.

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -0,0 +1,41 @@
+<ng-container *ngIf="titleAnalytics$ | async; else loading">
+  <div fxFlex fxLayoutAlign="space-between center">
+    <h3>Dashboard</h3>
+    <div fxFlex fxLayoutGap="24px" fxLayoutAlign="end center">
+      <button mat mat-fab color="default">
```
GrandSchtroumpf, line 5: remove the lonely `mat`. What is "default" ? And button should have a (click) event.

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -0,0 +1,41 @@
+<ng-container *ngIf="titleAnalytics$ | async; else loading">
```
GrandSchtroumpf, line 1: move that the furthest you can in the html tree. For exemple inside the surface.

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -0,0 +1,41 @@
...
+  <div fxFlex fxLayoutAlign="space-between center">
```
GrandSchtroumpf, line 2: To me it doesn't do anything as you don't have `fxLayout`.
Also why div ? From the code it looks like a header.
Also, you don't need the space-between, you can take advantage of `margin: auto` : 

```html
<header>
  <h3></h3>
  <button ...></button>
  <button ...></button>
  <button ...></button>
</header>
```

```css
header {
  display: flex;
}
h3 {
  margin-right: auto; // Will push the rest to the far right
}
```

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -0,0 +1,41 @@
...
+      <button mat-fab color="default">
+        <mat-icon svgIcon="movie"></mat-icon>
+      </button>
+      <button mat-fab color="default">
+        <mat-icon svgIcon="excel"></mat-icon>
+      </button>
+    </div>
+  </div>
+  <section class="surface carousel-container">
+    <div fxFlex fxLayoutAlign="space-between center" *ngIf="popularTitle$ | async as title">
```
GrandSchtroumpf, line 17: same as above

apps/festival/festival/src/app/dashboard/home/home.component.scss
```diff
@@ -0,0 +1,27 @@
+:host {
+  display: flex;
+  flex-direction: column;
+  padding: 24px;
+
+  & > * {
```
GrandSchtroumpf, line 6: you can try gap. It has good support now

apps/festival/festival/src/app/dashboard/home/home.component.scss
```diff
@@ -0,0 +1,27 @@
...
+  h5 {
+    text-decoration: underline;
+  }
+
+  bf-carousel {
+    margin-top: 24px;
+  }
+}
+
+.see_titles {
```
GrandSchtroumpf, line 24: If you don't use flex on the parent, and use `p` instead of `span`, you would have to do `align-self`. Just a regular `margin: auto`

apps/festival/festival/src/app/dashboard/home/home.component.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.hasMovies$ = allMoviesFromOrg$.pipe(map((movies) => !!movies.length));
+
+    this.hasDraftMovies$ = allMoviesFromOrg$.pipe(
+      map((movies) => movies.some((movie) => movie.app[this.app].status === 'draft'))
+    );
+
+    this.titles$ = allMoviesFromOrg$.pipe(
+      map((movies) =>
+        movies.filter((movie) => ['accepted', 'submitted'].includes(movie.app[this.app].status))
+      ),
```
GrandSchtroumpf, line 87: when using a callback, if you do it on several lines, use 
```typescript
() => {
  return ...;
}
```
Also avoid doing `[a, b].includes` for two values.
Lastely, it seems like checking the status of a movie is a common pattern, maybe create a reusable function for that. Maybe there is already one. If not, here is an example : 
```typescript
function hasAppStatus(app, status) {
  return (movie) => status.includes(movie.app[app].status)
}
```
Note that here I've got an undeterministic array, so I can use includes.

apps/festival/festival/src/app/dashboard/home/home.component.ts
```diff
@@ -0,0 +1,99 @@
...
+      tap((movies) => {
+        movies.length
+          ? this.dynTitle.setPageTitle('Dashboard')
+          : this.dynTitle.setPageTitle('Dashboard', 'Empty');
+      })
```
GrandSchtroumpf, line 92: Seems like a common pattern, maybe we can avoid repeating ourself. just an example : 
```typescript
tap(movies => this.dynTitle.setPageTitle'Dashboard', movies.length)
```
or something like that.

libs/firebase-utils/src/lib/firestoreMigrations/0075.ts
```diff
@@ -0,0 +1,191 @@
+import { Firestore } from '../types';
+import { BigQuery } from '@google-cloud/bigquery';
+import { Analytics, createTitleMeta, createDocumentMeta  } from '@blockframes/model';
+import { getCollection } from '../firebase-utils';
+import { MovieDocument } from '@blockframes/model';
+import { MovieAnalytics } from '@blockframes/analytics/components/movie-analytics-chart/movie-analytics.model';
+import { getCollectionInBatches } from '../util';
+import { bigQueryAnalyticsTable, firebase } from '@env';
+
+const events_query = `
```
GrandSchtroumpf, line 10: I let you check this file with @phpgeekfr

apps/festival/festival/src/app/dashboard/title/list/list.component.html
```diff
@@ -35,8 +35,8 @@ <h1>My Titles</h1>
       <ng-template colRef="title.international" label="Title" let-title sort>
         <b>{{ title }}</b>
       </ng-template>
-      <ng-template colRef="analytics.views" label="# Views" let-views sort>
-        {{views}}
+      <ng-template colRef="analytics" label="# Views" let-analytics sort>
+        {{ analytics?.length }}
```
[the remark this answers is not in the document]
RemcoSimonides: It's shows 0 by default

apps/festival/festival/src/app/dashboard/home/home.component.ts
```diff
@@ -0,0 +1,105 @@
...
+    map(analytics => counter(analytics, 'org.activity', 'orgActivity'))
+  )
+
+  territoryActivityOfPopularTitle$ = combineLatest([
+    this.popularTitle$,
+    this.titleAnalytics$
+  ]).pipe(
+    map(([title, titleAnalytics]) => titleAnalytics.filter(analytics => analytics.meta.titleId === title.id)),
+    map(analytics => counter(analytics, 'org.addresses.main.country', 'territories')),
+  )
```
RemcoSimonides, line 64: This is new. When you checked the code last friday, I noticed the data wasn't filtered by the most popular title.

p.s. in next PR I have ready, this is written in a more efficient way where I don't repeat the filter twice

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -0,0 +1,79 @@
...
+      <a [routerLink]="['/c/o/marketplace/title/', title.id]">
+        <img fxHide.lt-md [ref]="title.poster" asset="empty_poster.svg" alt="Movie Poster" class="poster">
+      </a>
+
+      <div fxFlex fxLayout="column">
+        <p class="mat-body-1">{{ title.directors | displayName }}</p>
+        <div fxLayout fxLayoutAlign="space-between">
+          <h1 class="mat-display-1">{{ title.title.international }}</h1>
+          <a mat-button [routerLink]="['/c/o/marketplace/title/', title.id]" class="film-page-button">
+            <span>See Film Page</span>
```
GrandSchtroumpf, line 17: unnecessary span
```suggestion
            See Film Page
```

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.scss
```diff
@@ -0,0 +1,28 @@
...
+  padding: 24px;
+  gap: 24px;
+
+  a {
+    margin-bottom: 16px;
+  }
+}
+
+.film-page-button {
+  height: fit-content;
```
GrandSchtroumpf, line 27: Usually it's not a good idea to use `fit-content`. You can usually achieve the same with `align-self`.
If you think it's the only option, then go for it.

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.ts
```diff
@@ -0,0 +1,96 @@
...
+    private location: Location,
+    private movieService: MovieService,
+    private orgService: OrganizationService,
+    private route: ActivatedRoute,
+    private analyticsService: AnalyticsService,
+    private userService: UserService,
+  ) {}
+
+  goBack() {
+    // TODO implement NavService: https://github.com/blockframes/blockframes/pull/8103/files
```
GrandSchtroumpf, line 89: When you do a todo, please add the id of the issue associated with the task `// @todo(#1234)`

apps/festival/festival/src/app/dashboard/home/home.component.scss
```diff
@@ -0,0 +1,36 @@
...
+      margin: auto auto auto 0;
+    }
+  }
+
+  bf-carousel {
+    margin-top: 24px;
+  }
+
+  .see_titles {
+    margin: 24px auto auto;
```
GrandSchtroumpf, line 34: for margin, use either 1, 2 or 4 values. Not 3

apps/festival/festival/src/app/dashboard/home/home.component.ts
GrandSchtroumpf: Would be nice to put a todo for that.

GrandSchtroumpf (approved):

PR #8113, opened 2022-03-25
backend functions automated deployment/packaging + improvements

george43g:
- a whole bunch of stuff


phpgeekfr:
this is a interesting but lets wait for the libs to be refactored so we don't have all the imports from front into backend-functions

Issue #8097, opened 2022-03-24
invitationCount$ inconsistencies


phpgeekfr:
in `libs/ui/src/lib/layout/event/event.component.ts` invitationCount only contains the filtered invitations for the last 4 months.

For consistency, this should be added directly on the service so that other `invitationCount$` displays the same amount.

PR #8122, opened 2022-03-25
Invitation counter inconsistencies

KevinDurand974:
From issue : https://github.com/blockframes/blockframes/issues/8097


KevinDurand974:
WIP : No tested at this moment

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -16,10 +16,6 @@ const applyFilters = (invitations: Invitation[], filters: { type: string[], stat
   return filters.status?.length ? inv.filter(inv => filters.status.includes(inv.status)) : inv;
 };
```
phpgeekfr, line 18: remove `import { subMonths } from 'date-fns';`

libs/invitation/src/lib/+state/invitation.service.ts
```diff
@@ -187,4 +188,21 @@ export class InvitationService extends CollectionService<InvitationState> {
     }
   }
 
+  invitationCounter(pendingStatus: boolean = true) {
```
phpgeekfr, line 191: rename to `invitationCount`

phpgeekfr: I would suggest something like :
```
  invitationCount({ onlyPending } = { onlyPending: true }) {
    const fourMonthsAgo = subMonths(new Date(), 4);
    // Filtering out invitations older than 4 months because there is no timeFrame supporting them in invitation-list component.
    const lastFourMonths = (invitation: Invitation) => invitation.date > fourMonthsAgo;
    return this.myInvitations$.pipe(
      map(invitations => invitations.filter(lastFourMonths)),
      map(invitations => invitations.filter(invitation => onlyPending ? invitation.status === 'pending' : true)),
      map(invitations => invitations.length)
    )
  }
```

This is more readable and you can do 
```
this.service.invitationCount({ onlyPending: false })
```

intead of 
```
this.service.invitationCounter(false);
```

That is also more readable

libs/invitation/src/lib/invitation.component.ts
```diff
@@ -33,10 +28,7 @@ export class InvitationComponent {
   });
 
   // Invitation count for conditions
-  invitationCount$ = this.service.myInvitations$.pipe(
-    map(invitations => invitations.filter(lastFourMonths)),
-    map(inv => inv.length)
-  );
+  invitationCount$ = this.service.invitationCount({ onlyPending: false })
```
phpgeekfr, line 31: missing `;`

phpgeekfr (approved):

Issue #7887, opened 2022-03-01
Make Avails research on Calendar & List view multi-terms


cams-goussale:
After the hotfix, here is the situation. I'm still looking at the same parameters as previously.

- The Map view can now manage Avails research on multiple terms.
- But the Calendar & List view cannot. This was an issue before the release, so it can wait for the next release to be fixed.

## Map view
![image](https://user-images.githubusercontent.com/93206363/156174463-abcc0670-0915-4a94-abda-a9f22b53aa85.png)

![image](https://user-images.githubusercontent.com/93206363/156174546-92ddfaa1-4fe6-4fcd-97fa-5e3a73f3fe0d.png)

## Calendar view
![image](https://user-images.githubusercontent.com/93206363/156174631-229b5e42-9b3e-460c-9765-ece222bcad8e.png)

![image](https://user-images.githubusercontent.com/93206363/156175138-f380d0ae-5414-4c79-b42a-e95df6255cdb.png)

![image](https://user-images.githubusercontent.com/93206363/156174794-4930453a-5872-4101-8b3b-b4a239601852.png)

![image](https://user-images.githubusercontent.com/93206363/156174849-c556c6ea-0cf2-4654-a637-69a307bbc192.png)

![image](https://user-images.githubusercontent.com/93206363/156174911-2a158ef8-572e-49fd-8114-0ee60774a718.png)

## List view
![image](https://user-images.githubusercontent.com/93206363/156175273-c03c377d-7c5e-4ef5-b447-5eb53451f80f.png)

![image](https://user-images.githubusercontent.com/93206363/156176017-229d0064-9332-406f-bb99-bc50e048f219.png)

![image](https://user-images.githubusercontent.com/93206363/156175454-c092145d-9e4b-40a4-b1b7-ab93483d25dc.png)

_Originally posted by @cams-goussale in https://github.com/blockframes/blockframes/issues/7875#issuecomment-1055432785_

cams-goussale:
I'm still working with the example of movie id `0gjxJ3ERixhw2r1uL4hl`
Let's take the following terms:

mandate.term1 = [France,Planes,2025-2032]
mandate.term2 = [China,Planes,2022-2026]

Results I expect:

map.availsResearch =  { Planes,2022-2026 }
Expected result:
Show only China available because allOf AvailsResearch parameters is in mandate.term2.
Don’t show France because not allOf AvailsResearch parameters are in mandate.term1.

Problem identified:
[...]

PR #8141, opened 2022-03-28
Multi term avails research on Calendar.

yspringfield:
Fixed: #7887


GrandSchtroumpf (approved):

libs/contract/src/lib/avails/avails.ts
```diff
@@ -412,19 +422,89 @@ export function isCalendarTermInAvails<T extends BucketTerm | Term>(term: T, ava
...
+  const from = max(results.map((result) => result.periodAvailable?.from));
+  const to = min(results.map((result) => result.periodAvailable?.to));
+
+  // Get the none empty sold result
+  const sold = results
+    .map(({ sold }) => sold)
+    .filter(sold => sold.length)
+    .flat();
+
+  const noResult = results.find(result => !result.available.length)
```
GrandSchtroumpf, line 471:
```suggestion
  // If one of the subAvails has no availability, return empty result
  const noResult = results.some(result => !result.available.length)
```

PR #8158, opened 2022-03-30
7689 add back no titles ui for dashboard

RemcoSimonides:
The states of dashboard where an organization doesn't have any (accepted) movie yet was accidentally removed. This PR adds it back.
image

### DIFF
--- a/apps/festival/festival/src/app/dashboard/home/home.component.html
+++ b/apps/festival/festival/src/app/dashboard/home/home.component.html
@@ -1,4 +1,4 @@
-<ng-container *ngIf="titles$ | async as titles; else loading">
+<ng-container *ngIf="titles$ | async; else loading">
   <ng-container *ngIf="(hasAcceptedMovies$ | async) else noAcceptedMovies">
     <header>
       <h1>Dashboard</h1>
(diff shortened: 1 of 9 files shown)


apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -1,33 +1,83 @@
...
-  </a>
-</header>
-<section class="surface carousel-container">
-  <article *ngIf="popularTitle$ | async as title">
-    <h3>{{ title.title.international }}</h3>
-    <a mat-flat-button [routerLink]="['/c/o/dashboard/analytics', title.id]">
-      <span>See Analytics details</span>
-      <mat-icon svgIcon="arrow_forward"></mat-icon>
-    </a>
+<ng-container *ngIf="titles$ | async as titles; else loading">
```
GrandSchtroumpf, line 1: you're not using titles in the template. You can remove it

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -1,33 +1,83 @@
...
+  <ng-container *ngIf="(hasAcceptedMovies$ | async) else noAcceptedMovies">
+    <header>
```
GrandSchtroumpf, line 3: I think the correct html would be : 
```html
<h1></h1>
<nav>
  <a>...</a>
  <a>...</a>
  <a>...</a>
</nav>
```

apps/festival/festival/src/app/dashboard/home/home.component.scss
```diff
@@ -34,3 +34,19 @@ header {
     margin: 24px auto auto auto;
   }
 }
+
+.no-accepted-movies {
+  mat-card {
+    width: 330px;
```
GrandSchtroumpf, line 40: why do you hardcode width & height ?

GrandSchtroumpf (changes requested):

apps/festival/festival/src/app/dashboard/home/home.component.html
RemcoSimonides: The header has css making it a row. Putting the buttons next to the h1 instead of on a new line.

apps/festival/festival/src/app/dashboard/home/home.component.scss
RemcoSimonides: width is needed in order not to make it too wide, especially on smaller screens where they're arranged as a column.
Height could be removed indeed.
image

RemcoSimonides:
FYI, I've also updated the changes in `libs/ui/src/lib/dashboard/pages/home/home.component.ts` of which this is mostly a copy

GrandSchtroumpf (approved):
change the array in `hasAppStatus:['accepted']` to a single string `hasAppStatus:'accepted'`

Issue #7297, opened 2021-11-30
Add user to mailchimp in backend functions instead of Zapier


RemcoSimonides:
- [x] Add user to mailchimp immediately when they sign up

Since there is an issue with adding tags from Buying Preferences to Mailchimp if the user is not on mailchimp yet, we should resync these tags too
- [x] Sync Buying Preferences with Mailchimp tags

PR #8163, opened 2022-03-31
Add user to mailchimp backend function

Poppe-A:
#7297


apps/backend-functions/src/users.ts
```diff
@@ -99,6 +99,8 @@ export const onUserCreate = async (user: UserRecord) => {
     if (userDoc.exists) {
       if (!user.emailVerified) {
         const u = userDoc.data() as PublicUser;
+        const tags = [`${appName[u._meta.createdFrom]} subscription`];
```
RemcoSimonides, line 102: Why add tag here?
If Marie wants a tag, I think it should be better thought trough. There is no way currently to also add tag for other apps besides this one once the user has subscribed.

apps/backend-functions/src/users.ts
```diff
@@ -99,6 +99,8 @@ export const onUserCreate = async (user: UserRecord) => {
     if (userDoc.exists) {
       if (!user.emailVerified) {
         const u = userDoc.data() as PublicUser;
+        const tags = ['Firebase new user'];
```
RemcoSimonides, line 102: Use type MailchimpTag here
`const tags: MailchimpTag[] = [{ name: 'Firebase new user', status: 'active' }];`

RemcoSimonides (approved):

PR #8170, opened 2022-03-31
7834 line chart

### LINKED ISSUE #7834: (A) Sellers' Analytics - Dashboard - Interactions
[Wireframe](https://www.figma.com/file/FhY2dJqeYrVTXh7Q7HA0m7/%E2%9A%99%EF%B8%8F-Sellers'-Dashboard-(DONE)?node-id=438%3A145660)
[UI Design](https://www.figma.com/file/aNbs0XSJBqIvKrffJSzViH/Dashboard?node-id=78%3A2478)

RemcoSimonides:
fix #7834
Ana, Stephanie and Camille have seen it but I think we can do better. Discussion will continue when Ana is back from holidays
image


libs/analytics/src/lib/components/line-chart/line-chart.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    dataLabels: {
+      enabled: false
+    },
+    xaxis: {
+      type: 'datetime',
+    }
+  };
+
+  @Input() set data(data: Analytics[]) {
+    if (!data?.length) return;
```
GrandSchtroumpf, line 90: might be an error. Let's say you've got some data, and then you apply a filter that return an empty array. Then you'll still display the previous state.

libs/analytics/src/lib/components/line-chart/line-chart.component.ts
```diff
@@ -0,0 +1,114 @@
...
+        const analyticsOfDay = analytics
+          .filter(analytic => analytic.name === name)
+          .filter(analytic => isSameDay(day, analytic._meta.createdAt)
+        );
+
+        return [day.getTime(), analyticsOfDay.length] as [number, number];
+      });
+
+      series.push({ name: eventNameLabel[name], data });
+    }
```
GrandSchtroumpf, line 110: A little bit hard to read I think

GrandSchtroumpf (approved):

Issue #6724, opened 2021-09-24
Update wireframe & mockup tables according to inventory


Jackseed:
Every new tables are listed [here](https://docs.google.com/spreadsheets/d/1Et6lZpTcdkJ6nMHmIsTaVD_yObp6GGbvoqUHEB-o-iM/edit#gid=857044763). Grey lines are the one we erase and red comment are new things to add (mainly actions).

Mockups & screens should be updated accordingly. Please add an arrow on columns that can be sorted so we may know it visually. Ideally the filter bar should also list what it filters so screens may live by themselves. 

Please link on that issues updated screens so we can pass it to devs for updates when it's done

cams-goussale:
Note for Camille: inventory sent to Marie & Aurore :white_check_mark:

PR #8177, opened 2022-04-01
AC - Tables Inventory - Section : Deal Negotiation

KevinDurand974:
### Section - Deal Negotiation

From issue : https://github.com/blockframes/blockframes/issues/6724


libs/contract/src/lib/contract/holdback/list/list.component.html
```diff
@@ -11,7 +14,7 @@ <h3>Holdbacks</h3>
       </span>
     </ng-template>
 
-    <ng-template colRef="medias" label="Media" let-medias>
+    <ng-template colRef="medias" label="Rights" let-medias>
       <span class="group-label" (click)="openDetails(medias, 'medias')">
         {{ medias | toGroupLabel:'medias':'All' | join | maxLength:120 }}
       </span>
```
phpgeekfr, line 20: instead of using css `cursor: pointer;`  can't you use a `<button` tag ?

libs/contract/src/lib/offer/list/list.component.html
```diff
@@ -28,10 +28,10 @@ <h1>Offers & Deals</h1>
 
     <bf-table [source]="offers | filterBy: filterByStatus: filter.value" clickable useFilter pagination="50"
       (rowClick)="rowClick($event)">
-      <ng-template colRef="id" label="Offer Reference" let-reference sort>{{ reference }}</ng-template>
+      <ng-template colRef="id" label="Offer Reference" let-reference sort><b>{{ reference }}</b></ng-template>
```
phpgeekfr, line 31:
```suggestion
      <ng-template colRef="id" label="Offer Reference" let-reference sort>
        <b>{{ reference }}</b>
      </ng-template>
```

libs/contract/src/lib/contract/holdback/list/list.component.html
phpgeekfr: sorry, don't take my previous comment into account

phpgeekfr (approved):

libs/contract/src/lib/contract/holdback/list/list.component.html
KevinDurand974: > instead of using css `cursor: pointer;` can't you use a `<button` tag ?

That dont provide the cursor pointer and it broke the display :
image

Issue #7841, opened 2022-02-23
(B) Sellers' Analytics - Buyer Activity Page - Setup page, Vanity Metrics, Table, Export


RemcoSimonides:
[Wireframe](https://www.figma.com/file/FhY2dJqeYrVTXh7Q7HA0m7/%E2%9A%99%EF%B8%8F-Sellers'-Dashboard-(DONE)?node-id=372%3A88942)
[UI Design](https://www.figma.com/file/aNbs0XSJBqIvKrffJSzViH/Dashboard?node-id=122%3A25475)

- [x] Setup page routing
- [x] Add user header
- [x] Vanity Metrics - Total numbers
- [x] Table
- [ ] Export to CSV
- [ ] Clicking Vanity Metrics filters table

![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/01433d2f-b24e-4e38-88b4-59c6ad4c12d0)

PR #8179, opened 2022-04-04
7841 setup routing

RemcoSimonides:
#7841 setup routing
To prevent huge PR, I'd prefer to split it up into smaller ones.


apps/festival/festival/src/app/dashboard/dashboard.module.ts
```diff
@@ -152,19 +180,6 @@ const routes: Routes = [{
         },
       ],
     },
-    {
-      path: 'analytics',
```
RemcoSimonides, line 156: ☝️ this didn't work since it's only title analytics, and I need to add seller analytics too.

phpgeekfr (approved):

PR #8185, opened 2022-04-04
example with `openFromComponent`

phpgeekfr:


libs/event/src/lib/form/shell/shell.component.ts
```diff
@@ -91,33 +92,26 @@ export class EventFormShellComponent implements OnInit, OnDestroy {
...
+        }
       }
+      return true;
     }
-    return true;
-  }
     catch (err) {
-      console.log(err)
-      this.snackBar.openFromTemplate(this.somethingWentWrongTemplate, { duration: 5000 });
+      this.snackBar.openFromComponent(SnackbarErrorComponent, { duration: 5000 });
```
phpgeekfr, line 111: for specific message you can 
```suggestion
      this.snackBar.openFromComponent(SnackbarErrorComponent, { message: 'test message', duration: 5000 });
```

Issue #7689, opened 2022-01-31
Sellers' Analytics


RemcoSimonides:
All graphs/tables on the pages should **exclude** the following data
- [x] Users of the organisation(s) which own the title
- [x] Concierges

RemcoSimonides:
Moving data from BigQuery to Firestore can be done with a query something like this
```
SELECT ga.*, users.orgId
FROM (
    SELECT 
        event_timestamp as createdAt,
        MAX(event_name) as event_name,
        MAX(if(params.key = "movieId", params.value.string_value, NULL)) as movieId,
        MAX(if(params.key = "uid", params.value.string_value, NULL)) as userId
    FROM `blockframes.analytics_193045559.events_*`,
    UNNEST(event_params) AS params
    WHERE event_name = 'addedToWishlist'
    GROUP BY user_pseudo_id, event_timestamp
) as ga
LEFT JOIN `blockframes.firestore_export.users` AS users
ON GA.userId = users.uid
```

RemcoSimonides:
- [ ] Page view numbers should not include people from within the organisation, or cascade8 users.

PR #8187, opened 2022-04-04
 7689 exclude analytics of owner of title

RemcoSimonides:
#7689

page view count now excludes the analytics from owners of title.


libs/analytics/src/lib/+state/analytics.service.ts
```diff
@@ -18,7 +18,7 @@ import { joinWith } from '@blockframes/utils/operators';
 import { QueryFn } from '@angular/fire/firestore';
 
 interface AnalyticsState extends EntityState<Analytics>, ActiveState<string> { };
-interface AnalyticsWithOrg extends Analytics<'title'> {
+export interface AnalyticsWithOrg extends Analytics<'title'> {
```
RemcoSimonides, line 21: Needed to export to satisfy linter
👇 next change didn't like it if it was only available in this file

phpgeekfr: If linter was not satisfied, how can this already be on develop ?

phpgeekfr (approved):

PR #8195, opened 2022-04-06
fix mailchimp migration for dev env

Poppe-A:
#7297


libs/firebase-utils/src/lib/firestoreMigrations/0076.ts
```diff
@@ -35,7 +35,8 @@ export async function upgrade(db: Firestore) {
 
   for (let i = 0; i < users.docs.length; i++) {
     const user = createUser(users.docs[i].data());
-
+    
+    if(user.email.includes('dev+')) continue;
```
phpgeekfr, line 39: maybe this ?

```suggestion
    if(user.email.includes('dev+') && user.email.includes('@blockframes.io')) continue;
```

phpgeekfr (approved):

PR #8199, opened 2022-04-06
7531 translucent background on buttons

### LINKED ISSUE #7531: Mat-buttons variant for alignement problems
We agreed on creating a new issue to solve this alignement problem.

_Reminder_:  we have an alignement issue (only at some places) where mat-buttons are aligned with other elements. 
Indeed, they have internal padding and an invisible background that creates a visual gap. 
See the main example bellow. The mat-button "add to wishlist" seems not correctly aligned on it's left.

We want to solve this visual issue by having a lightly visible background in the mat-button. (But only where we need it).

RemcoSimonides:
fix #7531

### DIFF
--- a/libs/organization/src/lib/organization/components/wishlist-button/wishlist-button.component.html
+++ b/libs/organization/src/lib/organization/components/wishlist-button/wishlist-button.component.html
@@ -1,13 +1,13 @@
 <ng-container *ngIf="!small; else smallButton">
 
   <ng-container *ngIf="(toggle$ | async); else alreadyAdded">
-    <button test-id="heart-button" mat-button (click)="removeFromWishlist($event)">
-      <mat-icon color="warn" svgIcon="favorite_filled"></mat-icon>
+    <button test-id="heart-button" class="translucent-button" mat-button (click)="removeFromWishlist($event)">
+      <mat-icon svgIcon="favorite_filled"></mat-icon>
       <span class="mat-caption">Remove from Wishlist</span>
     </button>
   </ng-container>
   <ng-template #alreadyAdded>
-    <button test-id="heart-button" mat-button (click)="addToWishlist($event)">
+    <button test-id="heart-button" class="translucent-button" mat-button (click)="addToWishlist($event)">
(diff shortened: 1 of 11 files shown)


libs/organization/src/lib/organization/components/wishlist-button/wishlist-button.component.html
```diff
@@ -1,13 +1,13 @@
 <ng-container *ngIf="!small; else smallButton">
 
   <ng-container *ngIf="(toggle$ | async); else alreadyAdded">
-    <button test-id="heart-button" mat-button (click)="removeFromWishlist($event)">
-      <mat-icon color="warn" svgIcon="favorite_filled"></mat-icon>
```
RemcoSimonides, line 5: Red color from wishlist button should have been removed long time ago already according to Stephanie

libs/organization/src/lib/organization/components/wishlist-button/wishlist-button.component.html
```diff
@@ -17,19 +17,13 @@
 <ng-template #smallButton>
 
   <ng-container *ngIf="(toggle$ | async); else alreadyAddedSmall">
-    <button test-id="heart-button" mat-icon-button (click)="removeFromWishlist($event)">
-      <mat-icon color="warn" svgIcon="favorite_filled"></mat-icon>
-      <span>
-        <ng-content select="wishlist-remove-text"></ng-content>
-      </span>
```
RemcoSimonides, line 24: `ng-content` wasn't used with this component so I removed it.

libs/organization/src/lib/organization/components/wishlist-button/wishlist-button.component.html
```diff
@@ -1,36 +1,32 @@
-<ng-container *ngIf="!small; else smallButton">
+<section class="dark-contrast-theme">
```
RemcoSimonides, line 1: The `dark-contrast-theme` is added here to make the wishlist icon white (request from Stephanie). I couldn't find any different way to accomplish this except for changing the color of icons in the builders.

GrandSchtroumpf (approved):

PR #8204, opened 2022-04-07
library refactoring - move maintenance interfaces

KvnMrs:


libs/model/src/lib/maintenance.ts
```diff
@@ -0,0 +1,11 @@
+import firebase from 'firebase/app';
+import 'firebase/firestore';
+
+export interface IMaintenanceDoc {
+  endedAt: firebase.firestore.Timestamp
+  startedAt: firebase.firestore.Timestamp
+}
```
phpgeekfr, line 7: @george43g is it ok to import
```
import firebase from 'firebase/app';
import 'firebase/firestore';
```

in model library ?

KvnMrs: I updated it, it should be better like this

phpgeekfr (approved):
LGTM @george43g can you have a look ?

PR #8205, opened 2022-04-07
Text inventory 

KvnMrs:
there are still some parts to do


libs/contract/src/lib/contract/holdback/form/form.component.html
```diff
@@ -3,7 +3,7 @@
 <div class="actions" fxLayout fxLayoutAlign="center center">
   <button mat-stroked-button (click)="openHoldbacks(formTemplate)">
     <mat-icon svgIcon="add"></mat-icon>
-    <span>Add and manage holdbacks for "{{ title.title.international | maxLength:10 }}"</span>
+    <span>Add and manage Holdbacks for  "{{ title.title.international | maxLength:10 }}"</span>
```
phpgeekfr, line 6: is extra space after `for`  needed ?

libs/user/src/lib/auth/forms/password/password.component.html
```diff
@@ -27,16 +27,16 @@
         />
         <mat-hint>Between 6 and 24 characters.</mat-hint>
         <mat-error *ngIf="form.get('next').hasError('required')">
-          New password is <strong>required</strong>.
+          New password is <b>required</b>.
```
phpgeekfr, line 30: why `b` instead of `strong` ?

phpgeekfr (approved):

libs/user/src/lib/auth/forms/password/password.component.html
KvnMrs: I saw that <b> was used in other components. It seems to me that `<strong>` brings a visual importance and also an importance for SEO and that <b> just brings the visual importance. Let me know if I put back `<strong>` or is it ok ?

phpgeekfr: since almost every pages require to be logged in to access it, SEO is not to take into account here. 

so as you want

PR #8216, opened 2022-04-08
6878 promotional elements and remove animation

### LINKED ISSUE #6878: Promotional elements on Film page
**Context:**
On the Banner of the Film Page, the action zone below the organization is reserved to the direct interaction with different promotional elements. 

 

Nevertheless we need to review this representation and install the rule of how many elements to show / what type etc. 

For the moment we had to limit the number of interaction buttons which is shown in the desktop version, as if not in the mobile version they are not represented correctly. 

**Task details**
We need to think over the desktop + mobile logic of representations that would let a user interact with promotional elements smoothly with no Header breaking in mobile.

RemcoSimonides:
fix #6878

### DIFF
--- a/libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.ts
+++ b/libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.ts
@@ -1,6 +1,7 @@
 import { Component, Input, ChangeDetectionStrategy, OnInit } from '@angular/core';
 import { Movie } from '@blockframes/model';
 import { scrollIntoView } from '@blockframes/utils/browser/utils';
+import { delay } from '@blockframes/utils/helpers';
 
 @Component({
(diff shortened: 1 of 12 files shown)


libs/movie/src/lib/movie/components/promotional-links/promotional-links.component.ts
```diff
@@ -19,7 +20,8 @@ export class PromotionalLinksComponent implements OnInit {
     );
   }
 
-  scrollToFooter() {
+  async scrollToFooter() {
+    if (document.getElementById('mat-menu-open')) await delay(250);
```
GrandSchtroumpf, line 24: add a comment to explain you're waiting for the menu animation to ends

GrandSchtroumpf (approved):

Issue #8165, opened 2022-03-31
Organization page on mobile


cams-goussale:
Can we rework the Organization page to make it responsive on mobile? It's likely this page will be visited a lot for Cannes.

![277537504_531865011881764_5900787350762401453_n](https://user-images.githubusercontent.com/93206363/161018869-e51df633-a555-497a-a0b7-8df3fefde2a9.jpg)
![277403865_498043355284849_5102488951299741540_n](https://user-images.githubusercontent.com/93206363/161018874-4bb38d28-8b7a-4b2f-ba12-61c62433c9f6.jpg)

PR #8226, opened 2022-04-11
Adapt org page for mobile

Poppe-A:
#8165


libs/organization/src/lib/organization/layout/view/view.component.scss
```diff
@@ -48,25 +47,35 @@ header {
...
-
 @include layout-bp(sm) {
   header, section {
     padding: 0 100px
   }
 }
 
+@include layout-bp(gt-xs) {
+  .org-info {
+    margin-left: 25px;
```
RemcoSimonides, line 58: 25 isn't a multiple of 4

RemcoSimonides (approved):

libs/organization/src/lib/organization/layout/view/view.component.scss
Poppe-A: I'll change but it was 25px before i wrote it, so I would keep this way

RemcoSimonides: The issue with keeping margins without being a multiple of 4 is that things are hard to align. So ye, change it unless it's clear why it's 25 instead of 24 but I don't see a comment for this.

Issue #8168, opened 2022-03-31
New privacy policy: Small banner + Notification


cams-goussale:
[Figma link](https://www.figma.com/file/Jte3fEhov35fVKAiRudnjj/Buyer-validation-for-email-address-sharing?node-id=18%3A336)

This issue only contains:

- [x] The banner on the HP
- [x] The notification
**Antoine** : After discussing with Camille and Anastasiia, we won't implement the notification now

The rest of the screens will be done later with this issue #8169

cams-goussale:
Design issue #7034

PR #8229, opened 2022-04-12
New policy - Banner

Poppe-A:
#8168


apps/festival/festival/src/app/marketplace/home/home.component.html
```diff
@@ -6,6 +6,7 @@
 
 <!-- Banner -->
 <ng-template #banner let-section>
+  <span class="mat-caption privacy-banner">We have updated our <a href="/c/o/marketplace/privacy" target="_blank">Privacy Policy</a>. In the absence of explicit disapproval, any use of the platform is considered as acceptance without restriction of the new policy. </span> <!-- TODO #8228 remove when needed -->
```
RemcoSimonides, line 9: Don't put this in the banner section. It's possible to have multiple banners on the page and then it would be shown multiple times. Just place it at the top

apps/festival/festival/src/app/marketplace/home/home.component.scss
```diff
@@ -26,6 +26,12 @@
   }
 }
 
+//TODO #8228 remove when needed
+.privacy-banner {
+  text-align: center;
+  padding-block: 8px;
```
RemcoSimonides, line 32: Use `padding` instead of `padding-block`. That helps it to not hit the sides which looks better for small screens where it is possible to hit the sides.

RemcoSimonides (changes requested):

RemcoSimonides (approved):

Issue #6870, opened 2021-10-12
Read notifications - change UI state instead of removing it


Jackseed:
Based on @maltmayer comment on #6621
We should keep notifications in the notification list once read but only change the colour/UI to differentiate read notification from unread.

This is issue is about creating the 2 different vue: read & unread

PR #8233, opened 2022-04-12
Read notifications - change UI state instead of removing it

Poppe-A:
#6870


libs/notification/src/lib/components/item/item.component.html
```diff
@@ -1,17 +1,24 @@
 
 <mat-list-item>
   <img matListAvatar [asset]="notification.placeholderUrl" [ref]="notification.imgRef" alt="Notification picture">
-  <p *ngIf="(xs$ | async) === false" matLine class="mat-body-1" test-id="notification-message" [innerHtml]="notification.message"></p>
+  <p *ngIf="(xs$ | async) === false" matLine [class.disabled]="notification.app.isRead" class="mat-body-1" test-id="notification-message" [innerHtml]="notification.message"></p>
```
RemcoSimonides, line 4: Use `ngClass` instead

libs/notification/src/lib/components/item/item.component.scss
```diff
@@ -2,3 +2,15 @@
 button {
   margin-right: 16px;
 }
+
+bf-tag {
+  margin-right: 36px;
+  border-color: var(--foreground-disabled-text);
```
RemcoSimonides, line 8: I don't think border-color should be overwritten since it's basically all the component does... What's the reason a different color is needed?

Poppe-A: To be closer to the design. This "READ" tag needs to be more clear than the normal text

RemcoSimonides: Create a disabled property on bf-tag instead and define its disable style there

RemcoSimonides (approved):

libs/notification/src/lib/+state/notification.service.ts
```diff
@@ -37,15 +37,17 @@ export class NotificationService extends CollectionService<NotificationState> {
   myNotifications$ = this.authService.profile$.pipe(
     filter((user) => !!user?.uid),
     switchMap((user) =>
-      this.valueChanges([where('toUserId', '==', user.uid), where('app.isRead', '==', false)])
+      this.valueChanges((ref) =>
+        ref.where('toUserId', '==', user.uid)
+      )
```
RemcoSimonides, line 42: Change back

Issue #7969, opened 2022-03-11
Update the privacy policy


cams-goussale:

cams-goussale:
- [ ] Update the Privacy Policy (Cross-App)
- [ ] Update the Terms & Conditions of Archipel Market only

Documents to come.

cams-goussale:
Documents available here @RemcoSimonides 
Let me know if you need .pdf

[20220404_CASCADE8_Privacy_V.1.3 CLEAN.docx](https://github.com/blockframes/blockframes/files/8426744/20220404_CASCADE8_Privacy_V.1.3.CLEAN.docx)
[20220404_T&C_CASCADE 8 CLEAN.docx](https://github.com/blockframes/blockframes/files/8426745/20220404_T.C_CASCADE.8.CLEAN.docx)

cams-goussale:
[20220404_CASCADE8_Privacy_V.1.3 CLEAN.pdf](https://github.com/blockframes/blockframes/files/8463811/20220404_CASCADE8_Privacy_V.1.3.CLEAN.pdf)
[20220404_T&C_CASCADE 8 CLEAN.pdf](https://github.com/blockframes/blockframes/files/8463814/20220404_T.C_CASCADE.8.CLEAN.pdf)

cams-goussale:
@yspringfield here is a first review on this issue:

- [ ] Privacy Policy: in the table (part 5) some Links are not working. Also could there be underlined like we usually do for links?
![image.png](https://images.zenhubusercontent.com/617811d92b89e7733866d8a0/a079fd58-6084-4b3c-883f-ce507652f25f)

PR #8235, opened 2022-04-12
updated terms and conditions and privacy policy

yspringfield:
Fixed:  #7969


libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.html
```diff
@@ -1,1248 +1,588 @@
...
+<ng-container>
+  <header>
+    <a mat-icon-button *ngIf="canGoBack" (click)="goBack()">
+      <mat-icon svgIcon="arrow_back"></mat-icon>
+    </a>
+    <h1>DATA PRIVACY POLICY</h1>
+  </header>
+  <section fxLayout="row" fxLayoutAlign="center">
+    <mat-card>
+      <mat-card-content>
```
RemcoSimonides, line 10: Use `class="surface"` instead of `mat-card`

libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.html
```diff
@@ -1,1248 +1,588 @@
...
-              format
-              or if it is feasible, as a direct transfer from one processor to another.
-              9. Right to lodge a complaint – in the event that we refuse your request under the Rights of Access, we
-              will
-              provide you with a reason as to why. If you are not satisfied with the way your request has been handled,
-              please contact us.
-              10. Right for the help of supervisory authority – meaning you have the right for the help of a supervisory
-              authority and the right for other legal remedies such as claiming damages.
-              11. Right to withdraw consent – you have the right withdraw any given consent for Processing of your
+<ng-container>
```
RemcoSimonides, line 1: Remove

libs/user/src/lib/auth/components/terms-conditions/terms-conditions.component.html
```diff
@@ -13,32 +13,35 @@ <h1>TERMS AND CONDITIONS OF USE OF ARCHIPEL MARKET</h1>
...
+              the "<strong>Buyer</strong>").
               The Service is an online platform that enables Sellers to present their portfolio of films available for
               sale and to publish
               content relating to those films. The Service enables a connection between Sellers on the one hand, and
               Buyers on the
               other hand. The Service does not allow to purchase film rights online. The Service is accessible on the
               website
-              <a [href]="appUrl">{{ appUrl }}</a> (hereinafter referred to as the “<strong>Website</strong>”).<br />
-              The term “<strong>User Seller</strong>” refers to a natural person who has a Member profile, Administrator
+              <a href="www.archipelmarket.com">www.archipelmarket.com</a>
```
RemcoSimonides, line 32: I don't see why you'd make it static instead of using the appUrl
Please change it back

libs/user/src/lib/auth/components/terms-conditions/terms-conditions.component.html
```diff
@@ -194,186 +202,172 @@ <h3>3 – Specific terms applicable to User Sellers</h3>
...
+              <li>
+                - Access live Events during specific timeframes, with a personal watermark appearing on any video
+                content,
+              </li>
+              <li>- Access a list of all upcoming Screenings,</li>
+              <li>- Access Screenings open to public,</li>
+              <li>- Request access to Screenings closed to public (private Screenings),</li>
+              <li>- Request the organization of a Screening on a specific Film Page,</li>
+              <li>- Formalize a price request to be aware of Seller's asking prices to sell film rights.</li>
             </ul>
```
RemcoSimonides, line 228: Items in an `<ul>` should already have a prefix icon... No need to add. Otherwise change the icon if it's not the good one

libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.html
```diff
@@ -1,1248 +1,581 @@
...
-          <article>
-            <h3>2. Data Protection Principles</h3>
-            <p>We undertake to follow the following data protection principles:</p>
+<header>
+  <a mat-icon-button *ngIf="canGoBack" (click)="goBack()">
+    <mat-icon svgIcon="arrow_back"></mat-icon>
+  </a>
+  <h1>DATA PRIVACY POLICY</h1>
+</header>
+<section fxLayout="row" fxLayoutAlign="center">
```
RemcoSimonides, line 7: I don't think this should be row as everything is aligned in column style

libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.scss
```diff
@@ -42,3 +42,15 @@ mat-card {
     margin-bottom: 30px;
   }
 }
+
```
RemcoSimonides, line 45: The table is indeed horrible on mobile, but so is all the other text. Could you perhaps do an attempt to fix the text for mobile?

RemcoSimonides (approved):

RemcoSimonides:
I just switched back to develop and it looks a lot better there on small screen so I'm afraid you broke something. There is also padding around
image

RemcoSimonides (changes requested):
.

libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.html
yspringfield: It's row. To keep it's child `.surface` horizontally centered.

libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.scss
```diff
@@ -36,9 +36,21 @@ mat-card {
...
+  .surface {
     width: 100%;
     margin-top: 30px;
     margin-bottom: 30px;
   }
 }
+
+.table__wrapper{
+  overflow: auto;
+}
```
RemcoSimonides, line 48: Why not directly target the `table` element? Does it not work like that?

Anyway, remove one of the underscores if you need this wrapper. `table_wrapper` instead of `table__wrapper`

RemcoSimonides (approved):

Issue #8211, opened 2022-04-08
(A) Remarks for Sellers Analaytics first phase


RemcoSimonides:
Dashboard
- [x]  Tooltip on fab-buttons
- [x]  Pie chart shouldn’t grow too big
- [x]  Published on Marketplace date is not dynamic. Remove it.
- [x]  Map should not be clickable on dashboard.

Map component
- [x]  Map should have primary color as selected color
- [x]  Add remove-filter button on top-right corner (only showed when filter is active)
- [x]  Some countries aren’t filtered in the table (like USA and Korea)

Line chart
- [x]  Line chart Interactions should display all events despite being 0
- [x]  Text on line chart is not visible enough.
- [x]  Change color of line chart to be primary 500 instead of primary 800 (see colors in UI Design)
- [ ]  Per day / per week / per month buttons are a go. Set default per on what looks good for you duration on title published
- [x] Legend should be left aligned
- [x]  Align vertically

Pie chart
- [x]  Add remove-filter button on top-right corner (only showed when filter is active)
- [x] Align vertically

PR #8240, opened 2022-04-13
8211 Remarks for sellers analytics

RemcoSimonides:
#8211
Fixed the checked checkmarks 👇 
Dashboard
- [x]  Tooltip on fab-buttons
- [x]  Pie chart shouldn’t grow too big
- [x]  Published on Marketplace date is not dynamic. Remove it.
- [x]  Map should not be clickable on dashboard.

Map component
- [x]  Map should have primary color as selected color
- [x]  Add remove-filter button on top-right corner (only showed when filter is active)
- [x] Fix dark mode image for empty state
- [ ]  Some countries aren’t filtered in the table (like USA and Korea)

Line chart
- [ ]  Line chart Interactions should display all events despite being 0
- [ ]  Text on line chart is not visible enough.
- [ ]  Change color of line chart to be primary 500 instead of primary 800 (see colors in UI Design)
- [ ]  Per day / per week / per month buttons are a go. Set default per on what looks good for you duration on title published
- [x]  Align vertically
- [x] Fix dark mode image for empty state

Pie chart
- [x]  Add remove-filter button on top-right corner (only showed when filter is active)
- [x] Align vertically
- [x] Fix dark mode image for empty state


apps/festival/festival/src/app/dashboard/home/home.component.scss
```diff
@@ -27,6 +27,13 @@ header {
 
   bf-carousel {
     margin-top: 24px;
+
+    analytics-pie-chart,
+    analytics-line-chart,
+    analytics-map {
+      width: calc(50% - 16px);
+      margin-right: 16px;
+    }
```
RemcoSimonides, line 36: Didn't manage to get `[col]="6"` to work in the html. This is due to apx-charts not having fixed sizes and they start resizing weirdly when animating the bf-carousel.

libs/analytics/src/lib/components/map/map.component.html
```diff
@@ -1,6 +1,9 @@
-<header fxLayout fxLayoutGap="8px">
-  <mat-icon matListIcon svgIcon="landscape"></mat-icon>
+<header fxLayout fxLayoutAlign="start center" fxLayoutGap="8px">
```
GrandSchtroumpf, line 1: use CSS instead of fxLayout

libs/analytics/src/lib/components/pie-chart/pie-chart.component.html
```diff
@@ -1,6 +1,9 @@
-<header fxLayout fxLayoutGap="8px">
+<header fxLayout fxLayoutAlign="start center" fxLayoutGap="8px">
```
GrandSchtroumpf, line 1: same as above

libs/analytics/src/lib/components/pie-chart/pie-chart.component.ts
```diff
@@ -72,6 +72,13 @@ export class PieChartComponent {
...
 
-  private previousSelection?: string;
+  previousSelection?: string;
   @Output() selection: EventEmitter<string> = new EventEmitter();
+
+  toggleSelect() {
+    const index = this.pieChartOptions.labels.findIndex(label => label === this.previousSelection);
+    if (index >= 0) {
+      this.chart.toggleDataPointSelection(index);
+    }
```
GrandSchtroumpf, line 82: If not found, index is always -1
```suggestion
    if (index !== -1) this.chart.toggleDataPointSelection(index);
```

GrandSchtroumpf (approved):

PR #8243, opened 2022-04-13
8211 remarks for sellers analytics 2

RemcoSimonides:
#8211
Map component
- [x]  Some countries aren’t filtered in the table (like USA and Korea)

Line chart
- [x]  Line chart Interactions should display all events despite being 0
- [x]  Text on line chart is not visible enough.
- [x]  Change color of line chart to be primary 500 instead of primary 800 (see colors in UI Design)
- [x]  Legend should be left aligned

### DIFF
--- a/apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
+++ b/apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
@@ -49,10 +49,10 @@
     <ng-template colRef="org.denomination.public" label="Company name" let-name sort>
       {{ name }}
     </ng-template>
-    <ng-template colRef="org.activity" label="Activity" let-activity sort [filter]="filters['orgActivity']">
+    <ng-template colRef="org.activity" label="Activity" let-activity sort [filter]="filters.orgActivity">
       {{ activity | toLabel:'orgActivity' }}
     </ng-template>
-    <ng-template colRef="org.addresses.main.country" label="Country" let-country sort [filter]="filters['territories']">
+    <ng-template colRef="org.addresses.main.country" label="Country" let-country sort [filter]="filters.territories">
       {{ country | toLabel:'territories' }}
     </ng-template>
     <ng-template colRef="user.position" label="Position" let-position sort>
--- a/apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
+++ b/apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
@@ -52,7 +52,7 @@
     <ng-template colRef="org.activity" label="Activity" let-activity sort [filter]="filters['orgActivity']">
       {{ activity | toLabel:'orgActivity' }}
     </ng-template>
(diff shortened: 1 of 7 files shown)


libs/analytics/src/lib/components/line-chart/line-chart.component.ts
```diff
@@ -114,4 +110,16 @@ export class LineChartComponent {
...
+  private sub = this.theme.theme$.subscribe(theme => {
+    this.lineChartOptions.chart.foreColor = theme === 'dark' ? '#f9f8ff' : '#080b0f';
+    this.chart?.updateOptions(this.lineChartOptions);
+  });
+
+  constructor(private theme: ThemeService) {}
+
+  ngOnDestroy() {
+    this.sub.unsubscribe();
+  }
```
RemcoSimonides, line 123: The rgb values are text foreground colors.

I've checked the HTML but there is no way to give the text in the graph a color using css... It contains many elements including svg before reaching a text. Also I haven't seen a class that is common under the text elements. Curious if you'd know a better way to use css variables in html ^^

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -52,7 +52,7 @@ <h2>Interested Users</h2>
     <ng-template colRef="org.activity" label="Activity" let-activity sort [filter]="filters['orgActivity']">
       {{ activity | toLabel:'orgActivity' }}
     </ng-template>
-    <ng-template colRef="org.addresses.main.country" label="Country" let-country sort>
+    <ng-template colRef="org.addresses.main.country" label="Country" let-country sort [filter]="filters['territories']">
```
GrandSchtroumpf, line 55:
```suggestion
    <ng-template colRef="org.addresses.main.country" label="Country" let-country sort [filter]="filters.territories">
```

apps/festival/festival/src/app/dashboard/home/home.component.scss
```diff
@@ -56,8 +49,24 @@ header {
...
+
 @include layout-bp(xs) {
   header {
     flex-direction: column;
   }
+
+  analytics-pie-chart,
+  analytics-line-chart,
+  analytics-map {
+    width: calc(100% - 8px);
```
GrandSchtroumpf, line 69: couldn't this be done at the carousel level ?

libs/analytics/src/lib/components/line-chart/line-chart.component.ts
```diff
@@ -114,4 +110,16 @@ export class LineChartComponent {
 
     this.chart?.updateOptions(this.lineChartOptions);
   }
+
+  private sub = this.theme.theme$.subscribe(theme => {
+    this.lineChartOptions.chart.foreColor = theme === 'dark' ? '#f9f8ff' : '#080b0f';
```
GrandSchtroumpf, line 115: couldn't you use css variable instead of listening on theme change ? Like `var(--primary)`

GrandSchtroumpf (approved):

PR #8251, opened 2022-04-15
8211 per unit buttons on line chart

### LINKED ISSUE #8211: (A) Remarks for Sellers Analaytics first phase

RemcoSimonides:
fix #8211
- [ ] Per day / per week / per month buttons are a go. Set default per on what looks good for you duration on title published

image

### DIFF
--- a/libs/analytics/src/lib/components/line-chart/line-chart.component.ts
+++ b/libs/analytics/src/lib/components/line-chart/line-chart.component.ts
@@ -46,6 +46,12 @@ const eventNameLabel: Record<EventName, string> = {
 
 type Period = 'day' | 'week' | 'month'; 
 
+const dateFunctions: Record<Period, Record<'interval' | 'isSame', any>> = {
+  day: { interval: eachDayOfInterval, isSame: isSameDay },
+  week: { interval: eachWeekOfInterval, isSame: isSameWeek },
+  month: { interval: eachMonthOfInterval, isSame: isSameMonth }
+};
+
 @Component({
   selector: '[data][eventNames] analytics-line-chart',
   templateUrl: './line-chart.component.html',
@@ -100,18 +106,16 @@ export class LineChartComponent {
       type: 'datetime',
       labels: {
         formatter: (value: string, timestamp: number) => {
-          if (this.period === 'month') {
-            return format(timestamp, "MMM yyy") 
-          } else {
-            return format(timestamp, 'dd MMM yyyy') 
-          }
+          return this.period === 'month'
+            ? format(timestamp, "MMM yyy")
+            : format(timestamp, 'dd MMM yyyy') 
         }
       }
     }
   };
 
   @Input() eventNames: EventName[] = [];
-  private analytics: Analytics[];
+  private analytics?: Analytics[];
   @Input() set data(data: Analytics[]) {
     if (!data?.length) {
       this.chart?.updateSeries([]);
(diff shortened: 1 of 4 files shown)


libs/analytics/src/lib/components/line-chart/line-chart.component.ts
```diff
@@ -75,33 +91,75 @@ export class LineChartComponent {
...
+    },
     xaxis: {
       type: 'datetime',
+      labels: {
+        formatter: (value: string, timestamp: number) => {
+          if (this.period === 'month') {
+            return format(timestamp, "MMM yyy") 
+          } else {
+            return format(timestamp, 'dd MMM yyyy') 
+          }
```
GrandSchtroumpf, line 107:
```suggestion
          return (this.period === 'month')
            ? format(timestamp, "MMM yyy") 
            : format(timestamp, 'dd MMM yyyy');
          }
```

libs/analytics/src/lib/components/line-chart/line-chart.component.ts
```diff
@@ -75,33 +91,75 @@ export class LineChartComponent {
...
+        }
+      }
     }
   };
 
   @Input() eventNames: EventName[] = [];
+  private analytics: Analytics[];
```
GrandSchtroumpf, line 114: If you don't initialize it, it's better to put to "?"

libs/analytics/src/lib/components/line-chart/line-chart.component.ts
```diff
@@ -75,33 +91,75 @@ export class LineChartComponent {
...
+    this.period = period;
+    this.updateChart(this.analytics);
+  }
+
+  private updateChart(analytics: Analytics[]) {
     const start = analytics[0]._meta.createdAt;
     const end = new Date();
-    const eachDay = eachDayOfInterval({ start, end });
+
+    const fn: Record<Period, Record<'interval' | 'same', any>> = {
```
GrandSchtroumpf, line 146: This could be defined  outside of the component as it doesn't depends on neither `this` nor `analytics`. Also give a better name than fn

libs/analytics/src/lib/components/line-chart/line-chart.component.ts
```diff
@@ -75,33 +91,75 @@ export class LineChartComponent {
...
 
     this.lineChartOptions.series = [];
     for (const name of this.eventNames) {
-      const data = eachDay.map(day => {
-        const analyticsOfDay = analytics
+      const data = eachUnit.map(unit => {
+        const analyticsOfUnit = analytics
           .filter(analytic => analytic.name === name)
-          .filter(analytic => isSameDay(day, analytic._meta.createdAt)
+          .filter(analytic => fn[this.period].same(unit, analytic._meta.createdAt)
```
GrandSchtroumpf, line 159: This is difficult to read put the function outside of the for loop : 
```typescript
const { interval, isSame } = fn[period];
const eachUnit = interval({ start, end });
for (const ... ) {
  ...
  .filter(analytic => isSame(unit, analytic._meta.createdAt)
}
```

GrandSchtroumpf (approved):

PR #8269, opened 2022-04-20
Display currency symbol / format price number

Poppe-A:
#7624

### DIFF
--- a/apps/backend-functions/src/templates/mail.ts
+++ b/apps/backend-functions/src/templates/mail.ts
@@ -12,15 +12,13 @@ import {
   OrganizationDocument, 
   PublicOrganization, 
   MovieDocument, 
-  createMailTerm, 
   ContractDocument, 
   NegotiationDocument, 
   Offer,
-  staticModel,
   Bucket,
   Timestamp,
 } from '@blockframes/model';
-import { EventEmailData, OrgEmailData, UserEmailData, getMovieEmailData, getOfferEmailData, MovieEmailData, getBucketEmailData } from '@blockframes/utils/emails/utils';
+import { EventEmailData, OrgEmailData, UserEmailData, getMovieEmailData, getOfferEmailData, MovieEmailData, getBucketEmailData, getNegotiationEmailData } from '@blockframes/utils/emails/utils';
 import { App, appName, Module } from '@blockframes/utils/apps';
 import { format } from "date-fns";
 import { supportMailosaur } from '@blockframes/utils/constants';
@@ -352,13 +350,12 @@ export function contractCreatedEmail(
   negotiation: NegotiationDocument, buyerOrg: OrganizationDocument
 ): EmailTemplateRequest {
   const pageURL = `${appUrl.content}/c/o/dashboard/sales/${contract.id}/view`;
-  const currency = staticModel.movieCurrenciesSymbols[negotiation.currency];
   const data = {
     user: toUser,
     app: { name: appName.catalog },
     movie: getMovieEmailData(title),
     contract,
-    negotiation: {...negotiation, currency},
+    negotiation: getNegotiationEmailData(negotiation),
     pageURL,
     buyerOrg,
   };
@@ -369,7 +366,7 @@ export function contractCreatedEmail(
 export function adminOfferCreatedConfirmationEmail(toUser: UserEmailData, org: OrganizationDocument, bucket: Bucket<Timestamp>): EmailTemplateRequest {
   const date = format(new Date(), 'dd MMM, yyyy');
   const mailBucket = getBucketEmailData(bucket);
-  const data = { org, bucket: { ...mailBucket }, user: toUser, baseUrl: appUrl.content, date };
+  const data = { org, bucket: mailBucket, user: toUser, baseUrl: appUrl.content, date };
 
   return { to: supportEmails.catalog, templateId: templateIds.offer.toAdmin, data };
 }
@@ -381,7 +378,7 @@ export function buyerOfferCreatedConfirmationEmail(toUser: UserEmailData, org: O
   const pageURL = `${appUrl.content}/c/o/marketplace/offer/${offer.id}`;
   const data = {
     app: { name: appName.catalog },
-    bucket: { ...mailBucket },
+    bucket: mailBucket,
     user: toUser,
     pageURL,
     baseUrl: appUrl.content,
@@ -412,8 +409,6 @@ export function counterOfferSenderEmail(
   toUser: UserEmailData, org: OrganizationDocument, offerId: string,
   negotiation: NegotiationDocument, title: MovieDocument, contractId: string, options: { isMailRecipientBuyer: boolean }
 ): EmailTemplateRequest {
-  const terms = createMailTerm(negotiation.terms);
-  const currency = staticModel.movieCurrenciesSymbols[negotiation.currency];
   const pageURL = options.isMailRecipientBuyer
     ? `${appUrl.content}/c/o/marketplace/offer/${offerId}/${contractId}`
     : `${appUrl.content}/c/o/dashboard/sales/${contractId}/view`;
@@ -425,7 +420,7 @@ export function counterOfferSenderEmail(
     org,
     contractId,
     app: { name: appName.catalog },
-    negotiation: { ...negotiation, terms, currency },
+    negotiation: getNegotiationEmailData(negotiation),
     movie: getMovieEmailData(title)
   };
   return { to: toUser.email, templateId: templateIds.negotiation.createdCounterOffer, data };
--- a/apps/backend-functions/src/templates/mail.ts
+++ b/apps/backend-functions/src/templates/mail.ts
@@ -12,13 +12,13 @@ import {
   OrganizationDocument, 
   PublicOrganization, 
   MovieDocument, 
-  createMailContract, 
-  MailBucket,
   createMailTerm, 
   ContractDocument, 
   NegotiationDocument, 
   Offer,
   staticModel,
+  Bucket,
+  Timestamp,
 } from '@blockframes/model';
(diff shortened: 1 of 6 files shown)


libs/utils/src/lib/emails/utils.ts
```diff
@@ -225,3 +227,11 @@ export function getMovieEmailData(movie: Partial<MovieDocument>): MovieEmailData
     }
   }
 }
+
+export function getBucketEmailData(bucket: Bucket<Timestamp>): MailBucket {
+  const currencyKey = Object.keys(movieCurrencies).find(key => movieCurrencies[key] === bucket.currency);
```
RemcoSimonides, line 232: Use `getKeyIfExists` instead

apps/backend-functions/src/templates/mail.ts
```diff
@@ -354,28 +352,29 @@ export function contractCreatedEmail(
   negotiation: NegotiationDocument, buyerOrg: OrganizationDocument
 ): EmailTemplateRequest {
   const pageURL = `${appUrl.content}/c/o/dashboard/sales/${contract.id}/view`;
+  const currency = staticModel['movieCurrenciesSymbols'][negotiation.currency];
```
RemcoSimonides, line 355:
```suggestion
  const currency = staticModel.movieCurrenciesSymbols[negotiation.currency];
```

apps/backend-functions/src/templates/mail.ts
```diff
@@ -412,7 +411,7 @@ export function counterOfferSenderEmail(
   negotiation: NegotiationDocument, title: MovieDocument, contractId: string, options: { isMailRecipientBuyer: boolean }
 ): EmailTemplateRequest {
   const terms = createMailTerm(negotiation.terms);
-  const currency = staticModel['movieCurrencies'][negotiation.currency];
+  const currency = staticModel['movieCurrenciesSymbols'][negotiation.currency];
```
RemcoSimonides, line 414:
```suggestion
  const currency = staticModel.movieCurrenciesSymbols[negotiation.currency];
```

libs/utils/src/lib/emails/utils.ts
```diff
@@ -225,3 +227,11 @@ export function getMovieEmailData(movie: Partial<MovieDocument>): MovieEmailData
...
+  return {
+    ...bucket,
+    currency: movieCurrenciesSymbols[currencyKey]
+  }
```
RemcoSimonides, line 236: missing `;`

libs/model/src/lib/static/static-model.ts
```diff
@@ -228,6 +228,11 @@ export const movieCurrencies = {
   NZD: 'New Zealand Dollar'
 } as const
 
+export const movieCurrenciesSymbols = {
+  USD: '$',
+  EUR: '€'
+} as const
```
RemcoSimonides, line 234: missing `;`

libs/model/src/lib/contract.ts
```diff
@@ -6,8 +6,13 @@ import { createMailTerm, Duration } from './terms';
 import { BucketContract } from './bucket';
 
 export function createMailContract(contract: BucketContract<Timestamp>) {
+  const formatter = new Intl.NumberFormat('en-US');
+  const formattedPrice = formatter.format(contract.price);
+  const updatedContract = {...contract} as any;
```
RemcoSimonides, line 11: This kinda ruins every type safety there is I think
Instead create an interface which combines Contract, Terms and the extra price field.

libs/utils/src/lib/emails/utils.ts
RemcoSimonides: What was the reason the bucket.currency contains the value instead of the key? That's something that should be fixed too since it's incorrect now and will cause unexpected bugs in the future

RemcoSimonides (changes requested):

libs/utils/src/lib/emails/utils.ts
Poppe-A: Not a bug. Here we get the bucket from a notification. But when creating a notification, we "override" the currency value (offer.ts l.27)

apps/backend-functions/src/templates/mail.ts
```diff
@@ -354,12 +352,13 @@ export function contractCreatedEmail(
...
 ): EmailTemplateRequest {
   const pageURL = `${appUrl.content}/c/o/dashboard/sales/${contract.id}/view`;
+  const currency = staticModel.movieCurrenciesSymbols[negotiation.currency];
   const data = {
     user: toUser,
     app: { name: appName.catalog },
     movie: getMovieEmailData(title),
     contract,
-    negotiation,
+    negotiation: {...negotiation, currency},
```
RemcoSimonides, line 361: This data object should reflect the `EmailTemplateRequest` in `libs/utils/src/lib/emails/utils.ts`. I don't see a negotiation property in there. Also, type safety seems to be gone here but you're magically changing a value from USD to $ which again might cause unexpected behaviour later. Better would be to have a `MailNegotation` type too, just like all the other types

RemcoSimonides: Also add spacing `{ ...negotiation, currency }`

apps/backend-functions/src/templates/mail.ts
```diff
@@ -369,18 +368,20 @@ export function contractCreatedEmail(
 /** Template for admins. It is to inform admins of Archipel Content a new offer has been created with titles, prices, etc in the template */
 export function adminOfferCreatedConfirmationEmail(toUser: UserEmailData, org: OrganizationDocument, bucket: Bucket<Timestamp>): EmailTemplateRequest {
   const date = format(new Date(), 'dd MMM, yyyy');
-  const contracts = bucket.contracts.map(contract => createMailContract(contract));
-  const data = { org, bucket: { ...bucket, contracts }, user: toUser, baseUrl: appUrl.content, date };
+  const mailBucket = getBucketEmailData(bucket);
+  const data = { org, bucket: { ...mailBucket }, user: toUser, baseUrl: appUrl.content, date };
```
RemcoSimonides, line 372: `{ ...mailBucket }` === `mailBucket`

apps/backend-functions/src/templates/mail.ts
```diff
@@ -369,18 +368,20 @@ export function contractCreatedEmail(
...
 /**To inform buyer that his offer has been successfully created. */
 export function buyerOfferCreatedConfirmationEmail(toUser: UserEmailData, org: OrganizationDocument, offer: Offer, bucket: Bucket<Timestamp>): EmailTemplateRequest {
-  const contracts = bucket.contracts.map(contract => createMailContract(contract));
+  const mailBucket = getBucketEmailData(bucket);
+
   const pageURL = `${appUrl.content}/c/o/marketplace/offer/${offer.id}`;
   const data = {
     app: { name: appName.catalog },
-    bucket: { ...bucket, contracts },
+    bucket: { ...mailBucket },
```
RemcoSimonides, line 384: `{ ...mailBucket }` === `mailBucket`

libs/model/src/lib/contract.ts
```diff
@@ -6,10 +6,14 @@ import { createMailTerm, Duration } from './terms';
 import { BucketContract } from './bucket';
 
 export function createMailContract(contract: BucketContract<Timestamp>) {
+  const formatter = new Intl.NumberFormat('en-US');
+  const formattedPrice = formatter.format(contract.price);
```
RemcoSimonides, line 10: Rename to `price`

libs/model/src/lib/contract.ts
```diff
@@ -6,10 +6,14 @@ import { createMailTerm, Duration } from './terms';
...
+
   return ({
     ...contract,
+    price: formattedPrice,
```
RemcoSimonides, line 14: After rename you can do just `price` instead

libs/model/src/lib/contract.ts
```diff
@@ -58,7 +62,7 @@ export interface Sale<D extends Timestamp | Date = Date> extends Contract<D> {
 
 export type ContractDocument = Mandate<Timestamp> | Sale<Timestamp>;
 
-export type MailContract = ReturnType<typeof createMailContract>
+export type EmailContract = ReturnType<typeof createMailContract>;
```
RemcoSimonides, line 65: Change the name back

libs/utils/src/lib/emails/utils.ts
```diff
@@ -30,51 +34,55 @@ export interface EmailTemplateRequest {
...
+    buyer?: UserEmailData;
+    contract?: ContractDocument;
+    territories?: string;
+    contractId?: string;
   };
 }
 
+// export interface EmailContract extends Contract {
+//   price: string;
+// }
```
RemcoSimonides, line 57: Remove
Please don't forget to check your changes before asking me to check your PR.

libs/utils/src/lib/emails/utils.ts
```diff
@@ -225,3 +233,14 @@ export function getMovieEmailData(movie: Partial<MovieDocument>): MovieEmailData
     }
   }
 }
+
+export function getBucketEmailData(bucket: Bucket<Timestamp>): MailBucket {
+  const currencyKey = getKeyIfExists('movieCurrencies', bucket.currency);
+  const contracts = bucket.contracts.map(contract => createMailContract(contract));
```
RemcoSimonides, line 239: ✔️

libs/model/src/lib/static/static-model.ts
```diff
@@ -228,6 +228,11 @@ export const movieCurrencies = {
   NZD: 'New Zealand Dollar'
 } as const
 
+export const movieCurrenciesSymbols = {
+  USD: '$',
+  EUR: '€'
+} as const;
```
RemcoSimonides, line 234: Let's also add the symbols for the other currencies available in `movieCurrencies`.

RemcoSimonides (approved):

PR #8271, opened 2022-04-20
cypress timeout to 30min

n-germain-work:
Trying to fix the error where CI fails after 10min without feedback


n-germain-work:
git issue, cannot merge, going for another branch

PR #8283, opened 2022-04-22
Fix scroll of carousel for Safari

RemcoSimonides:


GrandSchtroumpf (approved):
I don't think it's a good practice to hide the scrollbar.

PR #8285, opened 2022-04-22
Slate review

Poppe-A:
6844


libs/utils/src/lib/emails/utils.ts
```diff
@@ -191,8 +192,9 @@ function getEventEmailAttachment(event: EventDocument<EventMeta>, orgName: strin
 }
 
 function createIcsFromEventDocument(e: EventDocument<EventMeta>, orgName: string): IcsEvent {
-  if (!['meeting', 'screening'].includes(e.type)) return;
-  const event = e.type == 'meeting' ? e as MeetingEventDocument : e as ScreeningEventDocument;
+  if (!['meeting', 'screening', 'slate'].includes(e.type)) return;
+  const event = e.type == 'meeting' ? e as MeetingEventDocument : 
+    e.type == 'screening' ? e as ScreeningEventDocument : e as SlateEventDocument;
```
RemcoSimonides, line 197:
```suggestion
  const event = e.type == 'meeting'
    ? e as MeetingEventDocument
    : e.type == 'screening'
      ? e as ScreeningEventDocument
      : e as SlateEventDocument;
```

libs/model/src/lib/event.ts
```diff
@@ -11,6 +11,13 @@ import { Timestamp } from './timestamp';
 
 // Event types
 export type EventTypes = 'standard' | 'meeting' | 'screening' | 'local' | 'slate';
+export const EventTypesToDisplay = {
```
RemcoSimonides, line 14:
```suggestion
export const eventTypesToDisplay: Record<EventTypes, string> = {
```

RemcoSimonides: Let's move it to the static-model.
`contentType` and `contractType` are for example also in there so it makes sense if this is in there too

apps/festival/festival/src/app/marketplace/event/list/list.component.ts
```diff
@@ -33,23 +33,29 @@ export class ListComponent implements OnInit {
...
       shareReplay({ bufferSize: 1, refCount: true })
     )
   }
 
-  filterByOrgIds(events: Event<Screening>[], orgIds: string[]) {
+  hasMedia(event: Event<Screening | Slate>) {
+    const hasScreener = isScreening(event) && event.meta.titleId;
+    const hasSlate = isSlate(event) && event.meta.videoId;
+    return hasScreener || hasSlate;
+  }
```
RemcoSimonides, line 56: Move it outside of the class because it doesn't need anything from the class itself

libs/media/src/lib/file/explorer/explorer.component.ts
```diff
@@ -76,6 +78,10 @@ export class FileExplorerComponent implements OnInit, AfterViewInit {
       where(`app.${this.app}.access`, '==', true)
     ]
 
+    if (this.route.snapshot.queryParams.directory) {
+      this.next(this.route.snapshot.queryParams.directory);
+    }
```
RemcoSimonides, line 83:
```suggestion
    const { directory } = this.route.snapshot.queryParams;
    if (directory) this.next(directory);
```

libs/model/src/lib/static/static-model.ts
```diff
@@ -46,6 +47,14 @@ export const contentType = {
   tv: 'TV'
 }
 
+export const EventTypesToDisplay: Record<EventTypes, string> = {
```
RemcoSimonides, line 50: Sorry, what I meant is then also to use the static model like we usually do everywhere. So that in the HTML you can do for example
```
  type | toLabel:'eventType'
```
That will remove all the `typesToDisplay = EventTypesToDisplay;` in the class

Excuse moi for misunderstanding

libs/model/src/lib/event.ts
```diff
@@ -204,3 +204,9 @@ export function createMeetingAttendee(
     status,
   };
 }
+
+export function hasMedia(event: Event<Screening | Slate>) {
+  const hasScreener = isScreening(event) && event.meta.titleId;
+  const hasSlate = isSlate(event) && event.meta.videoId;
+  return hasScreener || hasSlate;
+}
```
RemcoSimonides, line 212: What I meant with moving outside of the class is to keep it in the same file, but just outside of the class.

```
import {} from '...';

function hasMedia() { ... }

@Component({ ... })
export class ListComponent { 
  ... 
}
```

Nevertheless, this is also a good location in case it needs to be reused somewhere

RemcoSimonides (approved):

Issue #6860, opened 2021-10-12
UI Improvements for Mobile


RemcoSimonides:
### Archipel Market 

**Google Chrome Version 93.0.4577.63**

**landing page**

- [x] (Global) Have all titles smaller on mobile. 
media-query sm : font-size: 48px / line-height: 1.2;
Antoine: H1 font-size is already 36px. I just changed mat-display-1 and 2

- [x] Off centered buyer logos --> add a line:
"justify-content: center;" on mobile only.
<img width="996" alt="Capture d’écran 2021-09-07 à 13 48 03" src="https://user-images.githubusercontent.com/85220905/132339771-fc2ec38c-31b6-4757-9751-bbabe9d0aa8a.png">

- [x] Footer layout broken --> margin bottom; 16px (between logos)
![File (6)](https://user-images.githubusercontent.com/85220905/131813968-653aa0f7-d84a-427f-b31a-84ae8e168f5d.jpg)

**Sign up**

- [x] Increase space between H1 and logo --> margin-top 24px on H1
![bfc2f1edee8745d1a07936569a4b6ff3-0001](https://user-images.githubusercontent.com/85220905/131818965-f65384e5-6548-48ee-a744-67d18a8a3ddc.jpg)

**Dashboard**

- [x] Wrap button and "+" under H1
![File (7)](https://user-images.githubusercontent.com/85220905/131823829-4735b883-7fd0-457e-ad7d-101c8ebf2354.jpg)

**Wishlist**

- [x] Add horizontal scroll
![File (8)](https://user-images.githubusercontent.com/85220905/131824650-b9cbc729-5b18-40b0-ad8a-05d31fa566b3.jpg)

- [x] Global responsive fix to do on modals. 
media-queries mobile : padding-left and right 24px;
![File (11)](https://user-images.githubusercontent.com/85220905/131829182-d28440ed-27f3-4833-a595-de744fd8b7d5.jpg)

Antoine: I think the modals already have a 24px padding. But for some specifics ones we add more padding. So we need to check every modal

**Film detail**

- [x] Can't put the video in fullscreen ? --> Make it bigger on mobile and include option to see it in fullscreen
![File (1)](https://user-images.githubusercontent.com/85220905/131829910-29b39106-684d-4023-842c-a2338c5accec.jpg)

### General Padding

- [x] Replace padding:80px by 20px left and right on mobiles.
[...]

PR #8292, opened 2022-04-25
6860 slider improvements for mobile

RemcoSimonides:
#6860

- [x] (A) The buttons on the slider on mobile are overlapping


apps/catalog/catalog/src/app/marketplace/home/home.component.scss
```diff
@@ -54,7 +54,8 @@
 }
 
 movie-slide {
-  height: 700px;
+  height: 100%;
+  min-height: 620px;
```
RemcoSimonides, line 58: 620 instead of 700 because the slider now has padding at the bottom so the content doesn't overlap the carousel controls. Same height is preserved.

It is min-height because on small screen the content can grow bigger.

libs/ui/src/lib/slider/slider.component.scss
```diff
@@ -28,6 +30,7 @@ img {
   bottom: 40px;
   z-index: 1;
   left: 50%;
+  transform: translate(-50%);
```
RemcoSimonides, line 33: On large screen it wasn't very visible that `left: 50%` didn't center it perfectly. On small screen it was so with this line it is actually centered horizontally.

GrandSchtroumpf (commented):
It doesn't feel right. We should think the component in a responsive way, not applying the desktop design to mobile.

GrandSchtroumpf (approved):
Let's create a ticket to design this component for responsive & rework it globally to avoid unwanted patches to remain.

RemcoSimonides:
#https://github.com/blockframes/blockframes/issues/8309

PR #8297, opened 2022-04-25
bugfix

george43g:
issue with select-environment

### DIFF
--- a/apps/backend-ops/src/select-environment.ts
+++ b/apps/backend-ops/src/select-environment.ts
@@ -19,7 +19,7 @@ function SAKIsCorrect(projectId: string) {
     key = getServiceAccountObj(process.env[SAKKeyName]);
     console.log('Current SAK projectId:', key.project_id);
   } catch (e) {
-    console.log('Was unable to detect previously set SAK')
+    console.log('Was unable to detect previously set SAK');
     return false;
   }
   return projectId === key.project_id;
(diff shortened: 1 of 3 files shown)


apps/backend-ops/src/select-environment.ts
```diff
@@ -18,7 +18,10 @@ function SAKIsCorrect(projectId: string) {
   try {
     key = getServiceAccountObj(process.env[SAKKeyName]);
     console.log('Current SAK projectId:', key.project_id);
-  } catch (e) { void 0; }
+  } catch (e) {
+    console.log('Was unable to detect previously set SAK')
```
phpgeekfr, line 22:
```suggestion
    console.log('Was unable to detect previously set SAK');
```

You know I always check for `;` please add them before I ask

apps/cms/cms-e2e/src/support/index.ts
```diff
@@ -14,6 +14,4 @@
 // ***********************************************************
 
 // Import commands.js using ES2015 syntax:
-//TODO define proper way to import next line #8071
-// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
```
phpgeekfr, line 18: you removed it here but for other locations with `// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries`  this is still needed ?

phpgeekfr (approved):

Issue #8270, opened 2022-04-20
Title import bugs


cams-goussale:
- [x] Updating a movie with excel import
If user imports an excel of an existing title, it should update the movie. In other words, if I create Movie A, B and C and later on I upload the title-import template for movie A, B and C, it should update these movies with the fields modified.

To avoid side effects, we should use titleId in column A when a user wants to update an existing movie using the excel import.

Template used when creating movies for the first time:
[AM x Screen Partnership _drafts.xlsx](https://github.com/blockframes/blockframes/files/8522215/AM.x.Screen.Partnership._drafts.xlsx)

Template used when wanted to update existing movies:
[AM x Screen Partnership_accepted.xlsx](https://github.com/blockframes/blockframes/files/8522237/AM.x.Screen.Partnership_accepted.xlsx)

- [x]  Available versions (columns BH, BI, BJ, BK) show an error
I can't fill in these columns with languages. It gives an error in CRM.

yspringfield:
Hi, here are the expected values of the columns.
- `bh` : list of languages
- `bi, bj, bk, bl` : yes or no

In the above attached templates, I could notice that columns `bi, bj, bk, bl` are all list of languages. That's what was causing the error.

cams-goussale:
Ah shoot ! I've corrected the template [here](https://drive.google.com/drive/folders/1hmsmslJxLxrPpA5Cy5K814Rgd_bHYaQb) (Title_Import). You can re-upload it on the app.

PR #8317, opened 2022-04-27
feat: update title through excel import

yspringfield:
Fixed: #8270


GrandSchtroumpf (commented):
I think we should discuss this feature. It adds more complexity on something already very complexe.

libs/import/src/lib/view-extracted-elements/titles/view-extracted-titles.component.html
```diff
@@ -9,7 +8,8 @@ <h3>Titles to Create</h3>
   </ng-container>
   <ng-container *ngIf="moviesToUpdate$ | async as moviesToUpdate">
     <ng-container *ngIf="moviesToUpdate.data.length">
-      <h3>Titles to Update & Submit</h3>
+      <h3 class="margin-top">Titles to Update & Submit</h3>
```
GrandSchtroumpf, line 11: If you've got a h3 you need a article or section container.

libs/import/src/lib/view-extracted-elements/titles/view-extracted-titles.component.scss
```diff
@@ -0,0 +1,3 @@
+.margin-top{
```
GrandSchtroumpf, line 1: You should not target a specific h3 but use a gap or something like that between containers

libs/utils/src/lib/utils.ts
```diff
@@ -302,3 +302,20 @@ export function toLanguageVersionString(languages: LanguageRecord) {
...
+        obj.splice(+k, 1)
+      } else {
+        delete obj[k];
+      }
+    } else if (typeof obj[k] === "object") {
+      removeNulls(obj[k]);
+    }
+  }
+  return obj;
+}
```
GrandSchtroumpf, line 321: Better to treat the array in another way than the object
```suggestion
export function removeNulls(obj: any) {
  if (Array.isArray(obj)) return obj.map(removeNulls).filter(el => el !== null && el !== undefined);
  for (const k of Object.keys(obj)) {
    if (obj[k] === null && obj[k] === undefined) {
      removeNulls(obj[k]);
    } else {
      delete obj[k];
    }
  }
  return obj;
}
```

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -172,7 +172,7 @@ export interface ValueWithErrorSimple<T = unknown> extends SpreadsheetImportErro
 }
 
 function isValueWithError(o: unknown): o is ValueWithError {
-  return typeof o === 'object' && ('value' in o) && ('error' in o);
+  return o && o instanceof ImportLog;
```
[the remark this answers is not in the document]
yspringfield: Bildschirmfoto 2022-05-03 um 10 27 07 AM
`ValueWithError` is a ts type that does not exists at runtime.

libs/import/src/lib/view-extracted-elements/titles/view-extracted-titles.component.scss
```diff
@@ -0,0 +1,3 @@
+section:nth-child(2) {
+  margin-top: 24px;
+}
```
GrandSchtroumpf, line 3: It's better to have something like that
```suggestion
:host {
  display: flex;
  flex-direction: column;
  gap: 24px;
}
```

GrandSchtroumpf (approved):

Issue #6474, opened 2021-08-18
List page - Smaller filters


RemcoSimonides:
[Zeplin](https://zpl.io/bAKx5Xn)
![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/c6ce5173-d64c-4116-ab27-b837ce66602a)

- [x] Update UI of filters buttons
- [x] Update UI of Refresh filters button
- [x] There are xxx titles available

design issue: #5559

PR #8319, opened 2022-04-27
Create smaller filter button component

Poppe-A:
#6474


libs/ui/src/lib/list/filter/list-filter.component.html
```diff
@@ -1,8 +1,8 @@
 <section>
   <ng-container *ngFor="let filter of filters">
-    <button mat-stroked-button [widgetTarget]="filterWidget" [color]="filter.color$ | async">
+    <bf-filter-button [widgetTarget]="filterWidget" [color]="filter.color$ | async">
       {{filter.label}}
-    </button>
+    </bf-filter-button>
```
RemcoSimonides, line 5: Yea, I would do it the same as the `ongoing-event-button` (libs/ui/src/lib/ongoing-button/ongoing-button.component.ts)

Then you can just write
`<button mat-small-stroked-button [widgetTarget]="filterWidget" color="filter.color$ | async"></button>`
here

It will only affect the style and not the logic. This will enable us to use it anywhere we'd like. Also, then there is no need to have those @Input() methods in the component.

libs/ui/src/lib/small-stroked-button/small-stroked-button.component.ts
```diff
@@ -0,0 +1,13 @@
...
+
+@Component({
+  selector: 'a[small-stroke-button], button[small-stroked-button]',
+  templateUrl: './small-stroked-button.component.html',
+  styleUrls: ['./small-stroked-button.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+
+export class SmallStrokedButtonComponent {
+  @Input() color: string;
```
RemcoSimonides, line 12: unused

RemcoSimonides (approved):

GrandSchtroumpf (approved):

PR #8324, opened 2022-04-28
Refactoring devops library

KvnMrs:
The first steps if you want tcheck somethings.

Now i have to:
- [x] move firebase-utils
- [x] optimization of backend-ops/main.ts ( the goal is to call only one function ) 
- [x] fix all errors in the code as we saw.

@george43g @phpgeekfr if you can check what i did before the merge maybe? let's make sure it's the result you expected George.


phpgeekfr (commented):
don't use `'@blockframes/devops'` inside devops lib, use relative imports

george43g (approved):
great work mate!

PR #8340, opened 2022-05-02
Bruce/akita removal datorama

phpgeekfr:


libs/analytics/src/lib/+state/analytics.service.ts
```diff
@@ -122,4 +148,44 @@ export class AnalyticsService extends BlockframesCollection<Analytics> implement
...
+    this.analyticsCache = rows.map(r => ({
+      uid: r.user_id,
+      firstConnexion: r.first_connexion.value,
+      lastConnexion: r.last_connexion.value,
+      sessionCount: r.session_count,
+      pageView: r.page_view,
+    }));
+  }
+
+  getLastConnexion(uid: string): Date {
```
GrandSchtroumpf, line 164: There is an issue with the typing. The return type should be `Date | undefined`. so you could do : 
```suggestion
  getLastConnexion(uid: string) {
    return this.analyticsCache.find(u => u.uid === uid)?.lastConnexion;
  }
```
Same for below

GrandSchtroumpf (approved):

PR #8349, opened 2022-05-03
7842 Buyer screener analytics

### LINKED ISSUE #7842: (B) Sellers' Analytics - Buyer Activity Page - Screenings Buyer has participated
[Wireframe](https://www.figma.com/file/FhY2dJqeYrVTXh7Q7HA0m7/%E2%9A%99%EF%B8%8F-Sellers'-Dashboard-(DONE)?node-id=372%3A88942)
[UI Design](https://www.figma.com/file/aNbs0XSJBqIvKrffJSzViH/Dashboard?node-id=122%3A25475)

RemcoSimonides:
fix #7842


apps/festival/festival/src/app/dashboard/analytics/buyer/buyer-analytics.component.ts
```diff
@@ -76,6 +77,33 @@ function filterAnalytics(title: string, analytics: AggregatedAnalytic[]) {
     : analytics;
 }
 
+interface InvitationWithAnalytics extends Invitation { analytics: Analytics[]; };
+function toScreenerCards(invitations: Partial<InvitationWithAnalytics>[]): MetricCard[] {
```
RemcoSimonides, line 81: The invitation to this function has quite a complicated type.
`{ Invitation & { event: Event &  { movie: Movie, org: Organization} } & analytics: Analytics[] }`
This function is only using a part of that type, but I'm wondering what a nicer way could be to define the type.

apps/festival/festival/src/app/dashboard/home/home.component.ts
```diff
@@ -43,10 +43,16 @@ export class HomeComponent {
...
     map(analytics => analytics.sort((a, b) => a.count > b.count ? -1 : 1)),
-    switchMap(([popularEvent]) => this.movieService.getValue(popularEvent.key))
+    switchMap(([popularEvent]) => {
+      if (popularEvent) {
+        return this.movieService.valueChanges(popularEvent.key);
+      } else {
+        // No titles have any analytics yet so show any movie
+        return firstValueFrom(this.titles$).then(titles => this.movieService.getValue(titles[0].id));
+      }
+    })
```
RemcoSimonides, line 55: The spinner is loading infinitely if there are no analytics yet. It can't determine what is the most popular title. So instead it now get's the first title to at least show something

GrandSchtroumpf: This looks like "hiding the dust under the carpet" (not sure if it's a real expression).
anyway, if there is no analytics you should display that there is no analytics instead of showing a wrong information (maybe @2amarkina has an opinion on that).

Here is a code example : 
```html
<ng-container *ngIf="popularEvents$ | async as events: else loading">
  <ng-container *ngIf="event.length; else empty">
    <ng-container *ngIf="event[0].key | getTitle | async as title">
    
    </ng-container>
  </ng-container>
</ng-container>
```

apps/festival/festival/src/app/dashboard/analytics/buyer/buyer-analytics.component.ts
```diff
@@ -122,10 +154,41 @@ export class BuyerAnalyticsComponent {
...
-  )
+  );
+
+  invitations$ = combineLatest([
+      // Event invitations linked to current org
+      this.invitationService.valueChanges([where('type', '==', 'attendEvent'), where('fromOrg.id', '==', this.orgService.org.id)]),
+      this.invitationService.valueChanges([where('type', '==', 'attendEvent'), where('toOrg.id', '==', this.orgService.org.id)])
+    ]).pipe(
+    map(i => i.flat()),
+    map(invitations => invitations.filter((i, index) => invitations.findIndex(elm => elm.id === i.id) === index)),
```
RemcoSimonides, line 165: `allInvitations$` from invitationService is used quite a lot in the app. Even for the invitation count in the app-bar. So using allInvitation$ is already in the cache.
I changed it however and this is what it would look like instead. I can change it back if you want. `allInvitations$` is not querying ALL invitations btw, only the invitations of the current users' org.

GrandSchtroumpf (approved):

Issue #7672, opened 2022-01-28
blockframes-ci - remove blockframes-ci.ts and firestore db, auth, hosting , functions and unused buckets 


phpgeekfr:
Check if we can remove blockframes-ci.ts and firestore `db`, `auth`, `hosting` , `functions` and unused buckets (gs://blockframes-ci.appspot.com) to keep only storage buckets

this follows #7545

phpgeekfr:
also remove remaining watermarks in blockframes-ci bucket

phpgeekfr:
`gs://blockframes-ci.appspot.com` should be kept for imgIx when running e2e tests on ci

phpgeekfr:
auth is used for anon-process, we need to keep it as well

george43g:
> `gs://blockframes-ci.appspot.com` should be kept for imgIx when running e2e tests on ci

why? The default bucket has been moved... why would imgix need the old bucket?

phpgeekfr:
imgix was updated to use `blockframes-ci` instead of the old one.

phpgeekfr:
removing unused and old functions also removed env vars stored there so `firebase functions:config:get > .runtimeconfig.json` does not return any value but this is used for unit-tests.

A temporary workaround was made to make them work again :
```
npm run backend-ops use blockframes-ci

firebase functions:config:set twilio.account.secret=xxx
firebase functions:config:set twilio.account.sid=xxx
firebase functions:config:set twilio.api.key.secret=xxx
firebase functions:config:set twilio.api.key.sid=xx

npm run backend-ops use blockframes-bruce
```

but 
- "test": "firebase functions:config:get > .runtimeconfig.json && firebase emulators:exec --only firestore 'cross-env nx run-many --target=test --all --configuration=$NG_CONFIG'",
- "affected:test": "firebase functions:config:get > .runtimeconfig.json && firebase emulators:exec --only firestore 'cross-env nx affected:test --base=origin/develop --configuration=$NG_CONFIG'",

should use `writeRuntimeConfig` as the E2E tests already do

phpgeekfr:
remove 
```
    "blockframes-ci": {
      "hosting": {
        "catalog": [
          "blockframes-ci"
        ],
        "festival": [
          "blockframes-ci-festival"
        ],
        "crm": [
          "blockframes-ci-crm"
        ],
        "financiers": [
          "blockframes-ci-financiers"
        ],
        "cms": [
          "blockframes-ci-cms"
        ]
      }
[...]

PR #8360, opened 2022-05-04
lots of little improvements

george43g:
resolves some problems in issue #7672


.nvmrc
```diff
@@ -1 +1 @@
-^16.14.0
+16.14.0
```
phpgeekfr, line 1: why ? `v16.14.2` is the current LTS

libs/testing/unit-tests/src/lib/firebase/functions.ts
```diff
@@ -23,8 +24,9 @@ config();
  * @param overrideConfig allows custom configuration of test object
  * @returns firebase-functions-test mock object
  */
-export function initFunctionsTestMock(emulator = true, overrideConfig?: AppOptions): FirebaseTestConfig {
-  let runtimeConfig: any = {};
+export async function initFunctionsTestMock(emulator = true, overrideConfig?: AppOptions): Promise<FirebaseTestConfig> {
+  await writeRuntimeConfig(functionsConfigMap, join(process.cwd(), './.runtimeconfig.json'));
```
phpgeekfr, line 28: no need for windows users  ?
```
await writeRuntimeConfig(functionsConfigMap, join(process.cwd(), './dist/apps/backend-functions/.runtimeconfig.json'));
```

libs/testing/unit-tests/src/lib/firebase/functions.ts
```diff
@@ -97,7 +99,7 @@ function setData(testEnv: RulesTestEnvironment, dataDB: Record<string, unknown>)
 //////////////
 
 export function populate(collection: string, set: any[]) {
-  const db = admin.firestore();
+  const db = admin.firestore(); // TODO! Do not do this - db init is a side-effect and db should be passed in or init in a diff module
```
phpgeekfr, line 102: please add an issue number to comment

libs/testing/unit-tests/src/lib/firebase/functions.spec.ts
```diff
@@ -1,12 +1,11 @@
...
 import { initFunctionsTestMock } from './functions';
 
 describe('firebase testing library', () => {
-  it.skip('should read the doc and get same data', async () => {
+  it('should read the doc and get same data', async () => {
     //Init Mocks and environment
-    initFunctionsTestMock();
-
-    const db = connectFirestoreEmulator();
+    const app = await initFunctionsTestMock();
```
phpgeekfr, line 7: `const app` is not used

libs/testing/unit-tests/src/lib/firebase/functions.spec.ts
```diff
@@ -1,12 +1,11 @@
...
+    const db = firestore() // * The reason we do it this way is because functions use this connect code so we test it this way
```
phpgeekfr, line 8: missing `;`

libs/model/src/lib/generic-types.ts
```diff
@@ -0,0 +1 @@
+export type JsonObject = string | number | boolean | { [x: string]: JSON } | Array<JSON>;
```
phpgeekfr, line 1: what is this ? it is used nowhere

libs/testing/unit-tests/jest.config.js
```diff
@@ -7,7 +7,7 @@
   transform: {
     '^.+\\.[tj]sx?$': 'ts-jest',
   },
+  testEnvironment: 'node',
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'html', 'json'],
-  setupFilesAfterEnv: ['<rootDir>/src/test-setup.ts'],
```
phpgeekfr, line 11: can you remove empty file `libs/testing/unit-tests/src/test-setup.ts` ?

package.json
```diff
@@ -29,7 +29,7 @@
     "deploy:functions:config": "ts-node -P tools/tsconfig.tools.json tools/scripts/deploy-function-config.ts",
     "functions:shell": "concurrently --kill-others \"npm run build:functions --watch\" \"firebase functions:shell\" --raw",
     "emulators:start": "npm start emulators",
-    "test": "firebase functions:config:get > .runtimeconfig.json && firebase emulators:exec --only firestore 'cross-env nx run-many --target=test --all --configuration=$NG_CONFIG'",
+    "test": "firebase emulators:exec --only firestore,auth 'cross-env nx run-many --target=test --all --configuration=$NG_CONFIG'",
```
phpgeekfr, line 32: why is auth needed during unit tests ?

.nvmrc
george43g: You cannot use semver in this file, it breaks `nvm`. I can't remember what was wrongWith 16.14.2 but unless you want all the devs to reinstall their environment we should leave it as it is in package.json.

libs/testing/unit-tests/src/lib/firebase/functions.ts
george43g: added

libs/testing/unit-tests/src/lib/firebase/functions.ts
george43g: did and also did for everywhere else we do this. I will include a module to clean all this up in another PR

libs/testing/unit-tests/jest.config.js
george43g: I already did

libs/model/src/lib/generic-types.ts
george43g: not yet, but we will need this when generating fixture data (json) and not using the `any` type. This is a bad practice nowadays

package.json
george43g: For any tests that do anything with auth in the function body - otherwise they'll hit production. I need to update the init func as well to hit auth emulator instead of

.nvmrc
phpgeekfr: In package.json it is `"^16.14.0"`

I think that some devs have 16.14.0 and other 16.14.2

libs/model/src/lib/generic-types.ts
phpgeekfr: and libs/model is the right location ? not inside a testing lib ?

package.json
phpgeekfr: ok, right now I don"t think this is used but it will be required if we create unit tests for functions that interact with auth

Issue #8302, opened 2022-04-26
Release 4.2 - Events


RemcoSimonides:
[Screening test scenarios](https://www.notion.so/cascade8/Screening-69f2b3a496394341aa16227d58109eb4)
[Slate test scenarios](https://www.notion.so/cascade8/Slate-27668acd5798428f95aeff2ee0ccae2c)
[Meeting test scenarios](https://www.notion.so/cascade8/Meeting-7d16f2abef1a4b5b85b5a1889dfa1eab)
[Interactions](https://www.notion.so/cascade8/Interactions-244e704f695d45bf9355eebc49e541ce)

Issues resolved for this release:
- [x] https://github.com/blockframes/blockframes/issues/6844
- [ ] https://github.com/blockframes/blockframes/issues/6902
- [ ] https://github.com/blockframes/blockframes/issues/8058
- [ ] https://github.com/blockframes/blockframes/issues/8098
- [ ] https://github.com/blockframes/blockframes/issues/7770

cams-goussale:
- [ ] Can we put `slate presentation` as the event type instead of `slate`? this way, email would say `... invited you to the slate presentation [event.name]`

![image](https://user-images.githubusercontent.com/93206363/165970234-ccb02eb5-54c3-4804-8443-f380f5584783.png)

2amarkina:
- [ ] Do not show the titles with status "Archived" on the "My files" storage
<img width="1435" alt="Снимок экрана 2022-05-02 в 16 55 23" src="https://user-images.githubusercontent.com/55499036/166255989-c02a1d6b-d032-4e6c-93e1-b0164f64e992.png">

RemcoSimonides:
- [ ] Can't select video file

- Logged in with dev+bernadette-hvz@blockframes.io
- User didn't have a video file so the disabled field was correct
- Then I uploaded a video file
- Refreshed the event page but field is still disabled

<img width="1429" alt="image" src="https://user-images.githubusercontent.com/27687382/166443497-8daf5539-313c-4c47-aa12-b81c23535e64.png">

2amarkina:
**Snackbar** 
- [ ] When I add an event into Calendar the the snackbar text is bigger that box 
![Снимок экрана 2022-05-04 в 10 30 04](https://user-images.githubusercontent.com/55499036/166648315-6473cb4c-1e3d-40ac-b3d8-4e6de66d72f5.png)

2amarkina:
[...]

PR #8378, opened 2022-05-05
Fixes 8302 2

phpgeekfr:
fixes for #8302


libs/model/src/lib/static/static-model.ts
```diff
@@ -84,6 +84,14 @@ export const eventTypes = {
...
 } as const;
 
+export const eventRoomAccess = {
+  meeting: 'Meeting',
+  screening: 'Screening',
+  slate: 'Screening', // for Buyers point of view Slate event remains a Screening
+  standard: 'Standard',
+  local: 'Local'
+} as const;
+
```
RemcoSimonides, line 94: I think to prevent confusion between eventTypes and eventRoomAccess in the future, let's move this object to the `view.component.ts`.

RemcoSimonides (approved):

Issue #8304, opened 2022-04-26
Release 4.2 - Other


RemcoSimonides:
Mailchimp:
- [ ] https://github.com/blockframes/blockframes/issues/7297

Sendgrid
- [ ] https://github.com/blockframes/blockframes/issues/7624

Usability
- [ ] https://github.com/blockframes/blockframes/issues/8046

cams-goussale:
- [ ] It seems the email variable changes have broken this [email template](https://docs.google.com/spreadsheets/d/1b1K7YfnquxnfCgGGPxIz5ZNIjr9eK_bUrRJqLvPYGIk/edit#gid=97734814&range=L37) where there was an if statement. In case Buyer sends an offer with no price, it should send text mentionning the Asking Price.

![image](https://user-images.githubusercontent.com/93206363/166265178-572f4928-8b68-4e60-93d6-6f486d2e646a.png)

cams-goussale:
- [ ] I can't get to the /notification page, it loads forever with this console error.

![image](https://user-images.githubusercontent.com/93206363/166265444-c4e44659-311b-4be6-90f1-b94cef5a3edb.png)

Issue #8301, opened 2022-04-26
Release 4.2 - General UI


RemcoSimonides:
[General UI test scenarios](https://www.notion.so/cascade8/General-UI-db403143e36c480d9c6b562cc001119f)

Other issues that have been resolved for this release:

Landing Page:
- [ ] https://github.com/blockframes/blockframes/issues/7194
- [ ] https://github.com/blockframes/blockframes/issues/7896

Privacy Policy:
- [ ] https://github.com/blockframes/blockframes/issues/8168
- [x] https://github.com/blockframes/blockframes/issues/7969
- [ ] https://github.com/blockframes/blockframes/issues/8131

Invitations/Notifications:
- [ ] https://github.com/blockframes/blockframes/issues/6870
- [ ] https://github.com/blockframes/blockframes/issues/7705
- [ ] https://github.com/blockframes/blockframes/issues/8097

Inventory related issues
- [ ] Tables: https://github.com/blockframes/blockframes/issues/6724
- [ ] Buttons: https://github.com/blockframes/blockframes/issues/6440
- [ ] Snackbars: https://github.com/blockframes/blockframes/issues/7293
- [ ] Tooltips: https://github.com/blockframes/blockframes/issues/7294 https://github.com/blockframes/blockframes/issues/6851
- [ ] Mobile: https://github.com/blockframes/blockframes/issues/6860
- [ ] Text: https://github.com/blockframes/blockframes/issues/7292
- [ ] Modals: https://github.com/blockframes/blockframes/issues/6507

Images in modal
- [ ] https://github.com/blockframes/blockframes/issues/6637

cams-goussale:
- [ ] In the CRM, page /crm/import/contracts when we import a document with errors, the modal doesn't have the cross to close it. And if you click outside the modale it doesn't remove it.

![image](https://user-images.githubusercontent.com/93206363/166223176-22910f43-4289-4635-a68f-9bf19f9d7e2f.png)

2amarkina:
**Some UI problems that might have been already mentioned in Slack:** 

- [ ] Logos should be in white on the Landing Page (Archipel Market & Archipel Content)
<img width="1440" alt="logos landing" src="https://user-images.githubusercontent.com/55499036/166235999-86d990a7-93a6-41d1-91bc-aef4def1008f.png">

RemcoSimonides:
From Ana:
### Archipel Market
_Tested on: iPhone / Safari for mobile_ 

MOBILE
[...]

PR #8379, opened 2022-05-05
Release 8301

Poppe-A:
#8301
also include last point of #8304
(https://github.com/blockframes/blockframes/issues/8304#issuecomment-1117409751)

### DIFF
--- a/apps/festival/festival/src/app/marketplace/event/list/list.component.html
+++ b/apps/festival/festival/src/app/marketplace/event/list/list.component.html
@@ -3,8 +3,7 @@
     <h1 class="mat-display-2 page-title">Screening List</h1>
   </list-page-title>
 
-  <!-- (#8301)  Remove the filter for screnning list since it's not really used (#8301) 
-    but keep it commented if UI changes its mind-->
+  <!-- TODO : #8383-->
   <!-- <list-filter fxLayout fxLayoutAlign="center">
     <ng-template filter label="Sales Agents" [form]="searchForm">
       <algolia-chips-autocomplete index="org" displayWithPath="name"
--- a/apps/festival/festival/src/app/marketplace/event/list/list.component.html
+++ b/apps/festival/festival/src/app/marketplace/event/list/list.component.html
@@ -3,13 +3,15 @@
     <h1 class="mat-display-2 page-title">Screening List</h1>
   </list-page-title>
 
-  <list-filter fxLayout fxLayoutAlign="center">
+  <!-- (#8301)  Remove the filter for screnning list since it's not really used (#8301) 
+    but keep it commented if UI changes its mind-->
+  <!-- <list-filter fxLayout fxLayoutAlign="center">
     <ng-template filter label="Sales Agents" [form]="searchForm">
       <algolia-chips-autocomplete index="org" displayWithPath="name"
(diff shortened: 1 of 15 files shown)


apps/festival/festival/src/app/marketplace/event/list/list.component.html
```diff
@@ -3,13 +3,15 @@
...
+  <!-- (#8301)  Remove the filter for screnning list since it's not really used (#8301) 
+    but keep it commented if UI changes its mind-->
+  <!-- <list-filter fxLayout fxLayoutAlign="center">
     <ng-template filter label="Sales Agents" [form]="searchForm">
       <algolia-chips-autocomplete index="org" displayWithPath="name"
         [filters]="['appModule:dashboard',' hasAcceptedMovies:true']" [form]="searchForm"
         label="Search for a Sales Agent" placeholder="Sales Agent"></algolia-chips-autocomplete>
     </ng-template>
-  </list-filter>
+  </list-filter> -->
```
RemcoSimonides, line 14: Can you create a new issue with the "advanced" solution? Add the label "Design - UX" to it. Let them make a decision. If they don't want the "advanced" solution, then we can remove this code.
Tag that issue here instead. Then at least we know more specific what decision we're waiting for.

apps/festival/festival/src/app/marketplace/organization/list/list.component.scss
```diff
@@ -11,8 +11,17 @@ list-filter {
   justify-content: center;
 }
 
+button {
+  margin-bottom: 24px;
+}
```
RemcoSimonides, line 16: I think this margin should be on the list-filter element, no? How do we do it on the title list page? I remember we discussed this margin there

apps/festival/festival/src/app/marketplace/organization/list/list.component.ts
```diff
@@ -76,6 +77,12 @@ export class ListComponent implements OnInit, OnDestroy {
     await this.searchForm.search();
   }
 
+  clear() {
+    const initial = createOrganizationSearch();
+    this.searchForm.reset(initial);
+    this.cdr.markForCheck();
```
RemcoSimonides, line 83: How come you need `markForCheck`? The searchForm is listening to changes which updates the value of orgs$

libs/contract/src/lib/contract/holdback/selection-modal/selection-modal.component.scss
```diff
@@ -1,3 +1,20 @@
 p {
   text-align: center;
+}
+
+mat-divider {
+  margin-bottom: 16px
+}
+
+div.duration {
```
RemcoSimonides, line 9: `.duration` is enough

libs/utils/src/lib/emails/utils.ts
```diff
@@ -222,7 +222,7 @@ export function getOrgEmailData(org: Partial<OrganizationDocument>): OrgEmailDat
     id: org.id,
     denomination: orgName(org, 'full'),
     email: org.email || '',
-    country: org.addresses?.main?.country
+    country: capitalize(org.addresses?.main?.country)
```
RemcoSimonides, line 225: I think you need to run this through `toLabel(org.addresses?.main?.country, 'territories')` instead.

Issue #8381, opened 2022-05-05
Modal - Fix style of Data Privacy Policy


KevinDurand974:
Data Privacy Policy : https://github.com/blockframes/blockframes/issues/8301#issuecomment-1116231415

PR #8384, opened 2022-05-05
General UI - CrossApp Modals Privacy Policy

KevinDurand974:
From issue: https://github.com/blockframes/blockframes/issues/8381

- [x] Waiting for the validation of `12px` padding from text cells

### DIFF
--- a/libs/ui/src/lib/global-modal/global-modal.component.scss
+++ b/libs/ui/src/lib/global-modal/global-modal.component.scss
@@ -5,10 +5,10 @@
 section {
   position: relative;
 
-  .closeButton {
+  & > button {
     position: absolute;
-    top: -8px;
-    right: -8px;
+    top: 0;
+    right: 0;
   }
 }
 
@@ -31,19 +31,16 @@ section {
 
 .medium {
   padding: 24px;
+  padding-top: 48px;
   width: 60vw;
   min-height: 350px;
   max-height: min(60vh, 550px);
   max-width: 1000px;
 
   .mat-dialog-content {
-    max-height: min(calc(60vh - 72px), 478px);
+    max-height: min(calc(60vh - 96px), 454px);
     margin: 0 -48px;
     padding: 0 48px;
-  }
-  .closeButton {
-    top: -16px;
-    right: -16px;
   }
 }
 
@@ -52,12 +49,7 @@ section {
   min-width: min(40vw, 600px);
   max-height: 350px;
   max-width: 800px;
-
-  .closeButton {
-    top: 0;
-    right: 0;
-  }
-
+  
   .mat-dialog-content {
     max-height: 302px;
   }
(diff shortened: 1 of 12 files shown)


libs/ui/src/lib/global-modal/global-modal.component.scss
```diff
@@ -41,10 +41,24 @@ section {
     margin: 0 -48px;
     padding: 0 48px;
   }
+  .closeButton {
+    top: -16px;
+    right: -16px;
+  }
```
phpgeekfr, line 47: I don't think @GrandSchtroumpf would approve this. Is there a solution without negative px ?

And you should not have to use `.closeButton` use `button` directly

libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.html
```diff
@@ -6,5 +6,5 @@
     <h1>DATA PRIVACY POLICY</h1>
   </header>
 
-  <auth-privacy-policy-text></auth-privacy-policy-text>
+  <auth-privacy-policy-text modal ></auth-privacy-policy-text>
```
phpgeekfr, line 9:
```suggestion
  <auth-privacy-policy-text modal></auth-privacy-policy-text>
```

libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.html
```diff
@@ -6,5 +6,5 @@
     <h1>DATA PRIVACY POLICY</h1>
```
phpgeekfr, line 6: I think you should add `autoFocus : false` on 
- libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.ts
- libs/utils/src/lib/gdpr-cookie/cookie-dialog/cookie-dialog.component.ts

to avoid shadow on the cross : 

image

phpgeekfr: maybe, other modals are also concerned

libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.scss
```diff
@@ -14,14 +11,6 @@ header {
...
-}
-
-@include layout-bp(xs) {
-  header {
-    width: 100%;
-  }
-}
+.surface {
+  width: 100%;
+}
```
phpgeekfr, line 16: there is no `surface` class on html of this component

phpgeekfr (approved):

libs/user/src/lib/auth/components/privacy-policy/privacy-policy.component.scss
KevinDurand974: forgot to remove, my bad

PR #8385, opened 2022-05-05
Fixes 8302 3

phpgeekfr:
fixes for #8302 part 3


libs/event/src/lib/guard/event-access.guard.ts
```diff
@@ -51,30 +51,30 @@ export class EventAccessGuard implements CanActivate {
         if (next.queryParams?.i && credentialsUpdateNeeded) {
           this.authService.updateAnonymousCredentials({ invitationId: next.queryParams?.i });
         }
-
+        const incorrectInvitation = 'Sorry, it seems that you were not invited to this event, check your emails';
```
RemcoSimonides, line 54: Was this text suggested? What do I have to look for in my emails? Maybe that can be a bit more descriptive

libs/event/src/lib/guard/event-access.guard.ts
```diff
@@ -51,30 +51,30 @@ export class EventAccessGuard implements CanActivate {
...
           const isInvitationMatchingEvent = invitation?.eventId === event.id;
           if (!isInvitationMatchingEvent) {
-            this.snackBar.open('Incorrect invitation for event', 'close', { duration: 5000 });
+            this.snackBar.open(incorrectInvitation, 'close', { duration: 12000 });
             await this.authService.deleteAnonymousUser();
             return false;
           }
 
+          if (!isEmailMatchingInvitation) {
+            this.snackBar.open('Sorry, it seems that you were not invited to this event, try with other mail', 'close', { duration: 12000 });
```
RemcoSimonides, line 68: Also here use . instead of ,

"Sorry, it seems that you were not invited to this event. Try again with another email address."

RemcoSimonides (approved):

PR #8392, opened 2022-05-05
8299 fix filter on bf table

RemcoSimonides:
Filtering on columns with undefined `colRef` didn't work. We have to create custom filter function for those.

- [x] (A) The filtering if titles is not working correctly. I'm searching for the film with the right wording, but it doesn't search for it. Also i'm searching for a film that exists, but the filter shows me something not related 
filter titles

### DIFF
--- a/libs/ui/src/lib/list/table/filters.ts
+++ b/libs/ui/src/lib/list/table/filters.ts
@@ -1,0 +1,15 @@
+import { Movie, Scope, staticModel } from "@blockframes/model";
+import { displayName } from "@blockframes/utils/utils";
+
+export function getStaticModelFilter(scope: Scope) {
+  return (input: string, value: any) => {
+    if (typeof value !== 'string') return false;
+    const label = staticModel[scope][value];
+    return label.toLowerCase().includes(input);
+  };
+}
+
+export const filters = {
+  movieDirectors: (input: string, _: undefined, movie: Movie) => movie.directors.map(director => displayName(director)).some(director => director.toLocaleLowerCase().includes(input)),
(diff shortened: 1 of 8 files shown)


libs/ui/src/lib/list/table/filters.ts
```diff
@@ -0,0 +1,15 @@
+import { Movie, Scope, staticModel } from "@blockframes/model";
+import { displayName } from "@blockframes/utils/utils";
+
+export function getStaticModelFilter(scope: Scope) {
+  return (input: string, value: any) => {
```
phpgeekfr, line 5:
```suggestion
  return (input: string, value: string) => {
```

libs/ui/src/lib/list/table/filters.ts
```diff
@@ -0,0 +1,15 @@
...
+    if (typeof value !== 'string') return false;
+    const label = staticModel[scope][value];
+    return label.toLowerCase().includes(input);
+  };
+}
+
+export const filters = {
+  movieDirectors: (input: string, _: undefined, movie: Movie) => movie.directors.map(director => displayName(director)).some(director => director.toLocaleLowerCase().includes(input)),
+  movieTitle: (input: string, _: undefined, movie: Movie) => movie?.title?.international.toLocaleLowerCase().includes(input)
```
phpgeekfr, line 14:
```suggestion
  movieDirectors: (input: string, _, movie: Movie) => movie.directors.map(director => displayName(director)).some(director => director.toLocaleLowerCase().includes(input)),
  movieTitle: (input: string, _, movie: Movie) => movie?.title?.international.toLocaleLowerCase().includes(input)
```

no need to do `movie?.directors?` ?

phpgeekfr (approved):

PR #8400, opened 2022-05-06
Modal - General UI CrossApp Responsive

KevinDurand974:
- Fix medium sized modal height
- Add responsive to global-modal for screen size "xs" ( < 600px )
- Fix Data Privacy Policy who had a table with not a valid class


libs/user/src/lib/auth/components/privacy-policy-text/privacy-policy-text.component.html
```diff
@@ -317,7 +317,7 @@ <h3>5. Trusted Partners collecting Personal Data for Us and having access your P
           the Service (as authorized under the T&Cs), Our Trusted Partners in order to either provide
           You with the Service or enhance Your user experience.
         </p>
-        <div class="table__wrapper">
+        <div class="table_wrapper">
```
GrandSchtroumpf, line 320: css classes are kebab-case

GrandSchtroumpf (approved):

Issue #8299, opened 2022-04-26
Release 4.2 - Movie


RemcoSimonides:
[Movie test scenarios](https://www.notion.so/cascade8/Movie-8506fba333d64b8cb9e857351a6c7989)

Issues that have been completed in this release:
- [ ] https://github.com/blockframes/blockframes/issues/6449
- [ ] https://github.com/blockframes/blockframes/issues/7531
- [ ] https://github.com/blockframes/blockframes/issues/7540
- [ ] https://github.com/blockframes/blockframes/issues/7793
- [ ] https://github.com/blockframes/blockframes/issues/8127
- [ ] https://github.com/blockframes/blockframes/issues/6986
- [x] https://github.com/blockframes/blockframes/issues/7913

Movie banner:
- [ ] https://github.com/blockframes/blockframes/issues/6497
- [ ] https://github.com/blockframes/blockframes/issues/6878
- [ ] https://github.com/blockframes/blockframes/issues/7961
- [ ] https://github.com/blockframes/blockframes/issues/6614

RemcoSimonides:
> * [ ]  Is this the correct height of the poster agreed with Stéphanie? I find it curious to have it finishing not aligned with the interaction buttons. Here is the [documentation ](https://www.notion.so/cascade8/Banner-Film-Page-b1464efef5bb46958cd84fc361630268)on Notion. Maybe there has been other discussions on this topic.
> 
> ![image](https://user-images.githubusercontent.com/93206363/165962673-f4a57c21-ea7b-4b55-9563-6968d58108b4.png)

We can't align the buttons with the bottom of the poster if the content has more height than the poster. If there is less content than the height of the poster, the buttons will be aligned. That's what we agreed upon. We can increase the size of the poster if that's wanted.

RemcoSimonides:
From Ana
When I'm connected as an anonymous user (for the Public Presentations, no account on Archipel Market) I can still "add" a title to wishlist... 

- [ ] Possible solutions: hide this feature for not connected users / Send a user to the Account Creation page (as we do when they click on "View more") 
<img width="1282" alt="Снимок экрана 2022-05-02 в 16 32 53" src="https://user-images.githubusercontent.com/55499036/166252518-d5273e97-05b5-4f86-ae67-9de4405a2066.png">

RemcoSimonides:
From Ana:

**Titles Page**
[...]

PR #8423, opened 2022-05-10
Release 4.2 - Movie small issues

Poppe-A:
#8299

### DIFF
--- a/libs/model/src/lib/static/static-model.ts
+++ b/libs/model/src/lib/static/static-model.ts
@@ -138,6 +138,7 @@ export const hostedVideoTypes = {
 
 // @TODO (#1658) Update LANGUAGES static model to be RFC-5646 compliant
 export const languages = {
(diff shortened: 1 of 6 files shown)


libs/model/src/lib/static/static-model.ts
```diff
@@ -138,6 +138,7 @@ export const hostedVideoTypes = {
 
 // @TODO (#1658) Update LANGUAGES static model to be RFC-5646 compliant
 export const languages = {
+  '-no-dialog': 'No Dialog',
```
RemcoSimonides, line 141: Do you need `-` in the front in order to make it appear first? If not, then remove it

RemcoSimonides (approved):

libs/model/src/lib/static/static-model.ts
Poppe-A: Yup, it's needed

PR #8442, opened 2022-05-11
WIP for runtimeConfig update

george43g:
fixing the unit test config gets really deep - there's a lot of code that somehow snuck into the repo that I would never have approved - this PR serves to get the high priority fix in first to allow more time to properly fix


libs/testing/unit-tests/src/lib/firebase/functions.ts
```diff
@@ -27,11 +28,14 @@ export function initFunctionsTestMock(emulator = true, overrideConfig?: AppOptio
   let runtimeConfig: any = {};
   try {
     // tslint:disable-next-line: no-eval
```
phpgeekfr, line 30: `// tslint:disable-next-line: no-eval` should not be just above `runtimeConfig = eval('require')(join(process.cwd(), './.runtimeconfig.json'));` ?

phpgeekfr (commented):
you did not remove `firebase functions:config:get > .runtimeconfig.json &&` from package.json 

and please take care of cleaning suggested in #7672 :

`hosting` and  `db` can be cleared and `gs://blockframes-ci.appspot.com` should be empty

apps/backend-ops/src/main.ts
```diff
@@ -166,6 +169,9 @@ async function runCommand() {
     case 'rescueJWP':
       await rescueJWP({ jwplayerKey: arg1, jwplayerApiV2Secret: arg2 });
       break;
+    case 'writeRuntimeConfig':
+      await writeRuntimeConfig(functionsConfigMap, join(process.cwd(), './.runtimeconfig.json'));
```
phpgeekfr, line 173: await not needed

phpgeekfr: and maybe this functions should go in some libs ?

george43g: got rid of await, but I think the func is high level enough to leave as is - to clean this up further, I may as well do it in a separate PR where I correct the structure of the rest of this file

phpgeekfr (approved):

PR #8445, opened 2022-05-11
8299 hide page bar on small screen instead

RemcoSimonides:
After discussion we opted to go with hiding the page bar instead
https://cascade8.slack.com/archives/CFT9HSPLL/p1652264033670959


libs/movie/src/lib/movie/marketplace/shell/shell.component.html
```diff
@@ -1,6 +1,6 @@
 <ng-container *ngIf="movie$ | async as movie">
   <!-- App bar -->
-  <ng-template pageBar targetId="nav">
+  <ng-template *ngIf="gtSm$ | async" pageBar targetId="nav">
```
phpgeekfr, line 3: You shouldn't put ngf on a ng-template. Wrap it with ng-container.

phpgeekfr (approved):

PR #8453, opened 2022-05-11
Beginning CRUD with e2e tests : import and delete data

n-germain-work:
# In Cypress
image

# In Firebase

### collections
image

### docs
image
image
image
image
image

### DIFF
--- a/apps/festival/festival-e2e/src/fixtures/_test.ts
+++ b/apps/festival/festival-e2e/src/fixtures/_test.ts
@@ -1,0 +1,50 @@
+const time = Date.now();
+
+export const examples = {
+  singleDoc: {
+    [`example-single/doc-${time}`]: {
+      type: 'example',
+      time: time,
+    },
+  },
+
+  multipleDocs: {
+    [`example-multiple/doc1-${time}`]: {
+      type: 'example 1',
+      time: time,
+    },
+    [`example-multiple/doc2-${time}`]: {
+      type: 'example 2',
+      time: time,
+    },
+    [`example-multiple/doc3-${time}`]: {
+      type: 'example 3',
+      time: time,
+    },
+  },
+
+  docWithArray: {
+    [`example-array/doc-${time}`]: {
+      type: 'example 1',
+      time: time,
+      array: ['data1', 'data2', 'data3'],
+    },
+  },
+
+  docWithObjet: {
+    [`example-object/doc-${time}`]: {
+      type: 'example 1',
+      time: time,
+      object: { object1: 'data1', object2: 'data2', object3: 'data3' },
+    },
+  },
+
+  docWithSubCollection: {
+    [`example-subcollection/doc-${time}/subcollection`]: [
+      {
+        type: 'example 1',
+        time: time,
+      },
+    ],
+  },
+};
--- a/apps/festival/festival-e2e/src/integration/_test/_test.spec.ts
+++ b/apps/festival/festival-e2e/src/integration/_test/_test.spec.ts
@@ -1,0 +1,51 @@
+/// <reference types="cypress" />
+
+import { examples } from '../../fixtures/_test';
+
+//* TESTS --------------------------------------*//
+
+describe('Testing bridge between Cypress and noe', () => {
+  before(() => {
+    firebase.cleanDatabase();
+  });
+
+  it('creates a single doc', () => {
+    firebase.importData({ ...examples.singleDoc });
+  });
+
+  it('creates multiple docs', () => {
+    firebase.importData({ ...examples.multipleDocs });
+  });
+
+  it('creates a doc with an array in a field', () => {
+    firebase.importData({ ...examples.docWithArray });
+  });
+
+  it('creates a doc with an object', () => {
+    firebase.importData({ ...examples.docWithObjet });
+  });
+
+  it('creates a doc with a subcollection', () => {
+    firebase.importData({ ...examples.docWithSubCollection });
+  });
+});
+
+//* FUNCTIONS -------------------------------*//
+
+const firebase = {
+  cleanDatabase() {
+    firebase.deleteData('example-single');
+    firebase.deleteData('example-multiple');
+    firebase.deleteData('example-array');
+    firebase.deleteData('example-object');
+    firebase.deleteData('example-subcollection');
+  },
+
+  deleteData(path: string) {
+    cy.task('deleteData', path);
+  },
+
+  importData(data: any | any[]) {
+    cy.task('importData', { ...data });
+  },
+};
--- a/libs/testing/cypress/node/src/lib/plugins/db.ts
+++ b/libs/testing/cypress/node/src/lib/plugins/db.ts
@@ -1,6 +1,6 @@
 import { db } from '../testing-cypress';
 import { createUser, createOrganization, Organization, createPermissions } from '@blockframes/model';
-import { App, ModuleAccess } from '@blockframes/utils/apps';
+import { App, ModuleAccess } from '@blockframes/model';
 
 export async function getRandomEmail() {
   const { email } = await getRandomUser();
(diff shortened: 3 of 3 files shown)


apps/festival/festival-e2e/src/fixtures/_test.ts
```diff
@@ -0,0 +1,96 @@
+const time = Date.now();
```
phpgeekfr, line 1: will you keep this file ?

apps/festival/festival-e2e/src/integration/_test/_test.spec.ts
```diff
@@ -0,0 +1,135 @@
+/// <reference types="cypress" />
```
phpgeekfr, line 1: same

libs/testing/cypress/node/src/lib/plugins/db.ts
```diff
@@ -69,10 +66,110 @@ export function deleteUser(userId: string) {
...
+  if (!org) return null;
   return createOrganization(org.data());
 }
 
 export function deleteOrg(orgId: string) {
   return db.doc(`orgs/${orgId}`).delete();
-}
\ No newline at end of file
+}
+
```
phpgeekfr, line 76: code above will be removed ?

libs/testing/cypress/node/src/lib/plugins/db.ts
```diff
@@ -69,10 +66,110 @@ export function deleteUser(userId: string) {
...
+// TODO : create a CRUD light plugin, see issue #8460
```
phpgeekfr, line 77: comment is still needed?

libs/testing/cypress/node/src/lib/plugins/db.ts
```diff
@@ -69,10 +66,110 @@ export function deleteUser(userId: string) {
...
+
+const isDocumentPath = (path: string) => path.split('/').length % 2 === 0;
+
+//* IMPORT DATA*-----------------------------------------------------------------
+
+export function importData(data: Record<string, object> | Record<string, object[]>) {
+  // Array of Promises to create all documents in the data
+  const createAll = Object.entries(data).map(([path, content]) => {
+    // If document create doc
+    if (isDocumentPath(path)) {
```
phpgeekfr, line 89: you should maybe check that `content` is not an array

libs/testing/cypress/node/src/lib/plugins/db.ts
```diff
@@ -69,10 +66,110 @@ export function deleteUser(userId: string) {
...
+}
+
+//* DELETE DATA*----------------------------------------------------------------
+
+export async function deleteData(paths: string | string[]) {
+  if (!Array.isArray(paths)) {
+    const path = paths;
+    isDocumentPath(path) ? await deleteDocument(path) : await deleteCollection(path);
+  } else {
+    paths.forEach(async path => (isDocumentPath(path) ? await deleteDocument(path) : await deleteCollection(path)));
```
phpgeekfr, line 112: `forEach` is not async, so `deleteData` will return true before all promises are resolved.

use `for` or `map` with promise.all

libs/testing/cypress/node/src/lib/plugins/db.ts
```diff
@@ -69,10 +66,110 @@ export function deleteUser(userId: string) {
...
+
+//* GET DATA*------------------------------------------------------------------
+
+export async function getData(paths: string | string[]) {
+  let data;
+  if (!Array.isArray(paths)) {
+    const path = paths;
+    data = isDocumentPath(path) ? await getDocument(path) : await getCollection(path);
+  } else {
+    data = Promise.all(paths.map(async path => (isDocumentPath(path) ? await getDocument(path) : await getCollection(path))));
```
phpgeekfr, line 135: you don't need `await` and this is hard to read.
```
    const docOrCollectionData = (path: string) => isDocumentPath(path) ? getDocument(path) : getCollection(path);
    data = Promise.all(paths.map(path => docOrCollectionData(path)));
```

libs/testing/cypress/node/src/lib/plugins/db.ts
```diff
@@ -69,10 +66,110 @@ export function deleteUser(userId: string) {
...
+  }
+  return data || 'no data';
+}
+
+async function getDocument(path: string) {
+  const doc = await db.doc(path).get();
+  return doc.data();
+}
+
+async function getCollection(collection: string) {
```
phpgeekfr, line 145: do you need to have the path of each document ?

your data structure is like this, do you need such structure? : 
```
{ path: string, data: `FirebaseFirestore.DocumentData`}[]
```

libs/testing/cypress/node/src/lib/plugins/db.ts
```diff
@@ -69,10 +66,110 @@ export function deleteUser(userId: string) {
...
+    const path = docRef.path;
+    const doc = await db.doc(path).get();
+    return { [`${path}`]: doc.data() };
+  });
+  return Promise.all(data);
+}
+
+//* UPDATE DATA*-----------------------------------------------------------------
+
+//TODO : adapt to use id of doc or uid for users
```
phpgeekfr, line 157: add issue number

libs/testing/cypress/node/src/lib/plugins/db.ts
```diff
@@ -69,10 +66,110 @@ export function deleteUser(userId: string) {
...
+
+export async function getData(paths: string | string[]) {
+  let data;
+  if (!Array.isArray(paths)) {
+    const path = paths;
+    data = isDocumentPath(path) ? await getDocument(path) : await getCollection(path);
+  } else {
+    data = Promise.all(paths.map(async path => (isDocumentPath(path) ? await getDocument(path) : await getCollection(path))));
+  }
+  return data || 'no data';
```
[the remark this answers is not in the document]
phpgeekfr: would also be cool to have a return type that is not `any`

apps/festival/festival-e2e/src/fixtures/_test.ts
n-germain-work: Surely put it in a archive in the repo if that's possible. Even if it's not used It might help to structure tests in the future (in case of a new person coming to help for example)

apps/festival/festival-e2e/src/integration/_test/_test.spec.ts
n-germain-work: same as above

libs/testing/cypress/node/src/lib/plugins/db.ts
n-germain-work: it will be removed after the tests using these functions have been reworked 👍

libs/testing/cypress/node/src/lib/plugins/db.ts
n-germain-work: it still ongoing so I guess. But it will be gone when all the CRUd is finished 👌

libs/testing/cypress/node/src/lib/plugins/db.ts
```diff
@@ -69,10 +66,110 @@ export function deleteUser(userId: string) {
...
 
 export function deleteOrg(orgId: string) {
   return db.doc(`orgs/${orgId}`).delete();
-}
\ No newline at end of file
+}
+
+// TODO : create a CRUD light plugin, see issue #8460
+
+//* LIGHT PLUGIN*----------------------------------------------------------------
```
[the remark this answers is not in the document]
n-germain-work: 'light' as 'not heavy'.
This comment is temporary, to help know the limit between old and new code. Should disappear when only CRUD code is in this file.

libs/testing/cypress/node/src/lib/plugins/db.ts
n-germain-work: The logic goes like :
```
if document => return setDoc
if not an array => throw error
return promise.all
```
All cases work. But I can rework it if it seems hard to read 👌

phpgeekfr: lets say I use a path for a document but send an array as content, `if (isDocumentPath(path)) {` will be true and document will be updated with `object[]`

n-germain-work: And that's ok, it's a case we have in the examples :
```
// collection + random doc.ids
  multipleDocsBis: {
    ['example-multiple']: [
      {
        type: 'example 1',
        time: time,
      },
      {
        type: 'example 2',
        time: time,
      },
      {
        type: 'example 3',
        time: time,
      },
    ],
  },
```
It creates as many documents as objects in the array :
image

libs/testing/cypress/node/src/lib/plugins/db.ts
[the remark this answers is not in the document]
n-germain-work: > why not returning `‘undefined` instead of `no data` ?

A specificity of Cypress, who can not accept `undefined` as a returned value from a task.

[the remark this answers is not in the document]
phpgeekfr: `false` then ?

libs/testing/cypress/node/src/lib/plugins/db.ts
n-germain-work: i like it 👍

apps/festival/festival-e2e/src/integration/_test/_test.spec.ts
```diff
@@ -0,0 +1,267 @@
...
+
+import { cp } from 'fs/promises';
```
phpgeekfr, line 3: unused import

phpgeekfr (approved):

PR #8457, opened 2022-05-12
merge back change request 1

RemcoSimonides:


libs/contract/src/lib/contract/list/internal-sales/internal-sales.component.html
```diff
@@ -35,14 +35,15 @@
...
-    *ngIf="(filter$ | async) !== 'accepted' && (filter$ | async) !== 'declined'"
-    colRef="negotiation"
-    label="Negotiation Stage"
-    let-negotiation
-  >
-    <ng-container *ngIf=" negotiation | negotiationStage | async  as negoStage">
-      <span [matTooltip]="negoStage"> {{negoStage | maxLength:50}} </span>
-    </ng-container>
-  </ng-template>
+  <ng-container *ngIf="(filter$ | async) !== 'accepted' && (filter$ | async) !== 'declined'">
```
RemcoSimonides, line 38: Francois' comment also said:
> Also you shouldn't have two observer on the same observable. Either wrap your bf-table with *ngIf="filter$ | async as filter" or create a pipe

But this filter can contain `''` preventing the *ngIf="" from triggering. There are actually 5 more `(filter$ | async)` in this html code. The only other solution I can think of is to not have filter$ as Observable but as a normal variable in the class.

GrandSchtroumpf: If you do not have any link inside the the filter$ you can use `rxLet="$filter$; let filter"` which doesn't look for falsish values

GrandSchtroumpf (approved):

Issue #8169, opened 2022-03-31
Allow user to make his email adress secret


cams-goussale:
[Wireframes link](https://www.figma.com/file/Jte3fEhov35fVKAiRudnjj/Buyer-validation-for-email-address-sharing?node-id=18%3A336)

This issue only contains:

- [ ] Profile page: [Mockups](https://www.figma.com/file/ZEwI6O9G61FKgv6X8yt3SE/Profile?node-id=0%3A1)
- [ ] Auth page: [Mockups](https://www.figma.com/file/ODvbgSYOb0OyVK6WWVp3NU/Log-in%2FSign-up?node-id=0%3A1) - only the middle screens. The others will be treated with #8282

The rest has been done with this issue #8168

cams-goussale:
Design issue #7034

PR #8469, opened 2022-05-13
Allow user to make his email adress secret

Poppe-A:
#8169


libs/user/src/lib/auth/components/hide-email/hide-email.component.html
```diff
@@ -0,0 +1,7 @@
+<mat-slide-toggle [checked]="control.value" (change)="toggle($event)"></mat-slide-toggle>
+<p>Hide my email</p>
+<p class="mat-caption">Your email address wil be visible to all {{ currentApp | appName }} users unless you decide to hide it.</p>
+<p class="mat-caption">
+  NB: Please note that in the event that you interact with a sales agent or their film(s), your email will be made visible to them regardless. 
+  For more information, please refer to our Privacy Policy.
+</p>
```
RemcoSimonides, line 7: The code is quite complicated for such a simple thing.
`<mat-slide-toggle [formControl]="control">Hide my email</mat-slide-toggle>`

Ask if the `<p>` can be underneath the toggle button too. Then you don't need to create two columns
On all the other controls, the caption is also all the way on the left side

libs/user/src/lib/auth/components/hide-email/hide-email.component.ts
```diff
@@ -0,0 +1,24 @@
...
+  @Input() control: FormControl;
+  
+  constructor (
+    @Inject(APP) public currentApp: App
+  ) { }
+
+  toggle(event) {
+    this.control.markAsDirty();
+    this.control.setValue(event.checked);
+  }
```
RemcoSimonides, line 23: No need for this if you use `[formControl]` on the mat-slide

libs/user/src/lib/auth/components/hide-email/hide-email.component.html
```diff
@@ -0,0 +1,7 @@
+<mat-slide-toggle [checked]="control.value" (change)="toggle($event)"></mat-slide-toggle>
+<p>Hide my email</p>
+<p class="mat-caption">Your email address wil be visible to all {{ currentApp | appName }} users unless you decide to hide it.</p>
```
RemcoSimonides, line 3: It doesn't make sense that we mention a specific app, even though the setting applies to all apps... Probably something wrong with the text

libs/model/src/lib/user.ts
```diff
@@ -45,6 +45,7 @@ export interface PublicUser {
   firstName?: string;
   lastName?: string;
   orgId?: string;
+  hideEmail: boolean;
```
RemcoSimonides, line 48: I think it's fine to keep the attribute here and not in privacyPolicy object. If moving it anywhere else, the `settings` property on user could be a good place. But I'm not a fan of nesting object if Firestore docs.

libs/user/src/lib/auth/components/hide-email/hide-email.component.ts
```diff
@@ -0,0 +1,13 @@
+import { ChangeDetectionStrategy, Component, Input } from "@angular/core";
+import { FormControl } from "@angular/forms";
+
+@Component({
+  selector: 'hide-email',
```
RemcoSimonides, line 5: selector has to start with lib or app name. Otherwise the build in circleci should fail.
`selector: 'user-hide-email'`

libs/user/src/lib/auth/components/hide-email/hide-email.component.ts
```diff
@@ -0,0 +1,13 @@
...
+  templateUrl: './hide-email.component.html',
+  styleUrls: ['./hide-email.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+
```
RemcoSimonides, line 10: remove empty line

libs/user/src/lib/auth/components/hide-email/hide-email.component.ts
```diff
@@ -0,0 +1,13 @@
...
+export class HideEmailComponent {
+  @Input() control: FormControl;
```
RemcoSimonides, line 12: A little bit nitpicky but I think you should rename it to `form`. As we usually use this name form.

RemcoSimonides: Also, notify in the selector that it's mandatory
`selector: '[form] user-hide-email'`

libs/user/src/lib/auth/components/hide-email/hide-email.module.ts
```diff
@@ -0,0 +1,22 @@
+// Angular
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { ReactiveFormsModule } from '@angular/forms';
+// Component
+import { HideEmailComponent } from './hide-email.component';
+
+// Material
+import { MatSlideToggleModule } from '@angular/material/slide-toggle';
+import { AppPipeModule } from '@blockframes/utils/pipes';
```
RemcoSimonides, line 10: unused

libs/user/src/lib/auth/components/hide-email/hide-email.component.html
```diff
@@ -0,0 +1,6 @@
+<mat-slide-toggle [formControl]="form">Hide my email</mat-slide-toggle>
+<p class="mat-caption">Your email address will be visible to all Archipel users unless you decide to hide it.</p>
+<p class="mat-caption">
+  NB: Please note that in the event that you interact with a sales agent or their film(s), your email will be made visible to them regardless. 
+  For more information, please refer to our Privacy Policy.
```
RemcoSimonides, line 5: Are you sure Privacy Policy shouldn't be clickable?

RemcoSimonides (approved):

libs/user/src/lib/auth/components/hide-email/hide-email.component.html
```diff
@@ -0,0 +1,6 @@
+<mat-slide-toggle [formControl]="form">Hide my email</mat-slide-toggle>
+<p class="mat-caption">Your email address will be visible to all Archipel users unless you decide to hide it.</p>
+<p class="mat-caption">
+  NB: Please note that in the event that you interact with a sales agent or their film(s), your email will be made visible to them regardless. 
+  For more information, please refer to our <a [href]="user ? '/c/o/marketplace/privacy' : '/auth/privacy'" target="_blank">Privacy Policy</a>.
```
RemcoSimonides, line 5: If you go with `/c/o/marketplace/privacy` for logged in users, then I don't think you should `target="_blank"` as the back button (which shouldn't be shown btw if there is no previous page) leads to nothing.
You could instead also navigate to `/auth/privacy` no matter whether the user is logged in or not.

Issue #7843, opened 2022-02-23
(B) Sellers' Analytics - Dashboard - Buyers' Insights Map


RemcoSimonides:
[Wireframe](https://www.figma.com/file/FhY2dJqeYrVTXh7Q7HA0m7/%E2%9A%99%EF%B8%8F-Sellers'-Dashboard-(DONE)?node-id=438%3A145660)
[UI Design](https://www.figma.com/file/aNbs0XSJBqIvKrffJSzViH/Dashboard?node-id=78%3A2478)

Please negotiate about the design because I've asked Ana to make it similar to the already existing map we have but it has not been done.

![image.png](https://images.zenhubusercontent.com/5ed89f5adc98f2d44c81b9c0/b5f76f59-0941-45b7-aa98-4afa6bac028d)

PR #8480, opened 2022-05-16
Analytics buyer insights map

yspringfield:
Fixed: #7843

### DIFF
--- a/apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
+++ b/apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
@@ -35,7 +35,20 @@
 
     <div fxLayout fxLayout.lt-sm="column" fxLayoutGap="16px">
       <analytics-pie-chart fxFlex class="surface" [data]="orgActivity$ | async" (selection)="filterValue = $event"></analytics-pie-chart>
-      <analytics-map fxFlex class="surface" [data]="territoryActivity$ | async" selectable (selection)="filterValue = $event"></analytics-map>
+      <analytics-map fxFlex class="surface" [data]="territoryActivity$ | async" selectable (selection)="filterValue = $event">
+        <ng-template #header let-selected="selected" let-resetSelected="resetSelected">
+          <article>
+            <h5>
+              <mat-icon svgIcon="landscape"></mat-icon>
+              <span>Split by Company Territory</span><br/>
+            </h5>
+            <p class="mat-caption">Learn where are the interested people from.</p>
+          </article>
+          <button mat-icon-button *ngIf="selected" (click)="resetSelected(selected)" matTooltip="Remove selection">
+            <mat-icon svgIcon="refresh_filters"></mat-icon>
+          </button>
+        </ng-template>
+      </analytics-map>
     </div>
   </article>
 
--- a/apps/festival/festival/src/app/dashboard/home/home.component.html
+++ b/apps/festival/festival/src/app/dashboard/home/home.component.html
@@ -12,7 +12,7 @@
         <a mat-fab color="black" routerLink="../import" matTooltip="Import Titles">
           <mat-icon svgIcon="excel"></mat-icon>
         </a>
-      </div>   
+      </div>
     </header>
 
     <ng-container *ngIf="titleAnalytics$ | async as analytics; else loading">
@@ -28,15 +28,27 @@
               </a>
             </header>
             <bf-carousel>
-              <analytics-pie-chart class="surface" carouselItem [data]="orgActivityOfPopularTitle$ | async"></analytics-pie-chart>
-              <analytics-map class="surface" carouselItem [data]="territoryActivityOfPopularTitle$ | async"></analytics-map>
-              <analytics-line-chart class="surface" carouselItem [eventNames]="interactions" [data]="interactionsOfPopularTitle$ | async">
+              <analytics-pie-chart class="surface" carouselItem [data]="orgActivityOfPopularTitle$ | async">
+              </analytics-pie-chart>
+              <analytics-map class="surface" carouselItem [data]="territoryActivityOfPopularTitle$ | async">
+                <ng-template #header>
+                    <article>
+                      <h5>
+                        <mat-icon svgIcon="landscape"></mat-icon>
+                        <span>Split by Company Territory</span>
+                      </h5>
+                      <p class="mat-caption">Learn where are the interested people from.</p>
+                    </article>
+                </ng-template>
+              </analytics-map>
+              <analytics-line-chart class="surface" carouselItem [eventNames]="interactions"
+                [data]="interactionsOfPopularTitle$ | async">
                 <h5>
                   <mat-icon svgIcon="switch_access_shortcut"></mat-icon>
                   <span>Interactions</span>
                 </h5>
                 <p class="mat-caption">Learn how people interact with your Title.</p>
-    
+
               </analytics-line-chart>
               <analytics-line-chart class="surface" carouselItem [eventNames]="['pageView']" [data]="pageViewsOfPopularTitle$ | async">
                 <h5>
@@ -47,7 +59,7 @@
               </analytics-line-chart>
             </bf-carousel>
           </ng-container>
-    
+
           <a mat-button class="see_titles" routerLink="title">
             <span>View all Titles Activity</span>
             <mat-icon svgIcon="arrow_forward"></mat-icon>
@@ -57,7 +69,7 @@
       </ng-container>
     </ng-container>
   </ng-container>
-  
+
   <ng-template #noAcceptedMovies>
     <section fxLayout="column" fxLayoutAlign="center center" class="no-accepted-movies">
       <ng-container *ngIf="titles.length; else noMovies">
--- a/apps/festival/festival/src/app/dashboard/home/home.component.html
+++ b/apps/festival/festival/src/app/dashboard/home/home.component.html
@@ -68,6 +68,42 @@
 
       </ng-container>
     </ng-container>
+    <section>
+      <h3>Buyers' Insights</h3>
+      <analytics-map class="surface map-insights" [data]="activeCountries$ | async" [topCountryCount]="5" horizontal>
+        <ng-template #mapLegend>
+          <!-- Map Legen -->
+          <aside class="colors-card">
+            <mat-list>
+              <mat-list-item>
+                <svg viewBox="0 0 100 100" width="32px" height="16px">
+                  <rect fill="#001ec7" x="0" y="0" width="150" height="100" rx="10" />
+                </svg>
+                <span class="mat-caption">> 100</span>
+              </mat-list-item>
+              <mat-list-item>
+                <svg viewBox="0 0 100 100" width="32px" height="16px">
+                  <rect fill="#3c64f7" x="0" y="0" width="150" height="100" rx="10" />
+                </svg>
+                <span class="mat-caption">> 50</span>
+              </mat-list-item>
+              <mat-list-item>
+                <svg viewBox="0 0 100 100" width="32px" height="16px">
+                  <rect fill="#8da4fa" x="0" y="0" width="150" height="100" rx="10" />
+                </svg>
+                <span class="mat-caption">> 5</span>
+              </mat-list-item>
+              <mat-list-item>
+                <svg viewBox="0 0 100 100" width="32px" height="16px">
+                  <rect fill="#dee5ff" x="0" y="0" width="150" height="100" rx="10" />
+                </svg>
+                <span class="mat-caption">≤ 5</span>
+              </mat-list-item>
+            </mat-list>
+          </aside>
+        </ng-template>
+      </analytics-map>
+    </section>
   </ng-container>
 
   <ng-template #noAcceptedMovies>
--- a/apps/festival/festival/src/app/dashboard/home/home.component.scss
+++ b/apps/festival/festival/src/app/dashboard/home/home.component.scss
@@ -1,4 +1,4 @@
-@import "@angular/flex-layout/mq";
+@import '@angular/flex-layout/mq';
 
 :host {
   display: flex;
@@ -44,7 +44,7 @@ header {
       align-self: center;
     }
   }
-  
+
   h3 {
     margin-top: 24px;
   }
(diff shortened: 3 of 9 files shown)


libs/analytics/src/lib/components/map/map.component.ts
```diff
@@ -20,9 +20,14 @@ export class AnalyticsMapComponent {
   lessThanFifty: TerritoryISOA3Value[] = [];
   moreThanFifty: TerritoryISOA3Value[] = [];
   selected: TerritoryISOA3Value;
+  topCountries: AnalyticData[] = [];
 
-  top3: AnalyticData[] = [];
 
+  @Input() topCountryCount = 3;
+  @ContentChild('mapLegend') mapLegend: TemplateRef<HTMLElement>;
```
RemcoSimonides, line 27: As the colors and logic are defined within this component, I think it's fine to have a `@boolean showLegend = false` here and put all the code related to that legend in this component.

Everything in this component is about map, so no need to say 'mapLegend' but just 'legend' is enough. Same with 'topCountries' can be just 'top', and 'topCountryCount' can be just 'topCount'

libs/analytics/src/lib/components/map/map.component.ts
```diff
@@ -20,9 +20,14 @@ export class AnalyticsMapComponent {
...
+  @ContentChild('header') header: TemplateRef<HTMLElement>;
+  @Input() @boolean noHeader = false;
```
RemcoSimonides, line 29: unused

apps/festival/festival/src/app/dashboard/home/home.component.ts
```diff
@@ -5,18 +5,22 @@ import { Component, ChangeDetectionStrategy, Optional, Inject } from '@angular/c
...
-import { EventName, hasAppStatus, App } from '@blockframes/model';
 import { APP } from '@blockframes/utils/routes/utils';
 import { AnalyticsService } from '@blockframes/analytics/+state/analytics.service';
+import {
+  EventName,
+  hasAppStatus,
+  App,
+} from '@blockframes/model';
 import { counter } from '@blockframes/analytics/+state/utils';
+import { joinWith } from '@blockframes/utils/operators';
```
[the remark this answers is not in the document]
yspringfield: It's used [here](https://github.com/blockframes/blockframes/pull/8480/files/69d4b04558286dea910088d71fd84b880f961c02#diff-e50315e51248a81d0578f49c156bf90eb3412bdba894fb290c48e32ea437baf1R43).

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -35,7 +35,20 @@ <h2>Interested Users</h2>
...
+            <h5>
+              <mat-icon svgIcon="landscape"></mat-icon>
+              <span>Split by Company Territory</span><br/>
+            </h5>
+            <p class="mat-caption">Learn where are the interested people from.</p>
+          </article>
+          <button mat-icon-button *ngIf="selected" (click)="resetSelected(selected)" matTooltip="Remove selection">
+            <mat-icon svgIcon="refresh_filters"></mat-icon>
+          </button>
+        </ng-template>
```
RemcoSimonides, line 50: I think the code will be easier if the header is not inside this component. Especially with this reset selected button, it's becoming more complicated. Could you move it outside of the component?

Also, the [issue](https://github.com/blockframes/blockframes/issues/7843) you're working on contains 2 figma links. One is for wireframes, the other one for UI. You should follow the UI design as it's more detailed. Wireframes are just to make quick sketch for the user experience, after that UI makes it pretty, once that's done then it comes to us. That's the workflow we have.

On the UI design you'll see that the header is inside "surface" class

apps/festival/festival/src/app/dashboard/home/home.component.scss
```diff
@@ -70,4 +70,13 @@ header {
     width: calc(100% - 8px);
     margin-right: 8px;
   }
-}
\ No newline at end of file
+}
+
+analytics-map.map-insights {
```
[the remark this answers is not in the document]
yspringfield: Bildschirmfoto 2022-05-17 um 9 15 23 AM
This is the structure of the map component. It doesn't have a wrapping html element on whom to apply the `map-insights` class. That's why I had to put it here.

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
yspringfield: I think this is the right approach. There are certain places namely:
* [here](https://github.com/blockframes/blockframes/blob/analytics-buyer-insights-map/apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html#L38) where we want a header with a reset button.
* [here](https://github.com/blockframes/blockframes/blob/analytics-buyer-insights-map/apps/festival/festival/src/app/dashboard/home/home.component.html#L33) where we want a header without a reset button.
* [here](https://github.com/blockframes/blockframes/blob/analytics-buyer-insights-map/apps/festival/festival/src/app/dashboard/home/home.component.html#L77)  where we don't want a header at all.
Doing otherwise entails a lot of boolean inputs. Don't think that is an ideal approach. Let me know what you think.

apps/festival/festival/src/app/dashboard/home/home.component.scss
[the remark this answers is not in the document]
RemcoSimonides: Use `:host`

libs/analytics/src/lib/components/map/map.component.html
```diff
@@ -1,49 +1,71 @@
...
       <map-feature [tag]="isoA3" color="#001ec7" (click)="toggleSelect(isoA3)"></map-feature>
     </ng-container>
-  
+
     <ng-container *ngIf="selected">
       <map-feature [tag]="selected" color="#3c64f7" (click)="toggleSelect(selected)"></map-feature>
     </ng-container>
+
+    <!-- Legend -->
+    <aside class="legend" *ngIf="showLegend">
```
RemcoSimonides, line 32: Ask Daniela how she wants to have this because currently it's not really visible
image

libs/analytics/src/lib/components/map/map.component.scss
```diff
@@ -5,19 +5,24 @@
...
 }
 
 world-map {
+  flex: 1 1 auto;
   height: 300px;
   border-radius: 4px;
 }
 
-section {
+.map-wrapper{
```
RemcoSimonides, line 17: Currently the wrapping isn't working optimally. It should wrap earlier. Maybe change to `flex-direction` for screens smaller than md?
image

libs/analytics/src/lib/components/map/map.component.scss
```diff
@@ -28,7 +33,27 @@ button {
...
 
-
 mat-spinner {
   margin: auto;
-}
\ No newline at end of file
+}
+
+mat-list.horizontal {
+  flex: 0 0 300px;
```
RemcoSimonides, line 41: Shouldn't be needed to have a fixed size. Try something that's more responsive.

libs/analytics/src/lib/components/map/map.component.ts
```diff
@@ -46,17 +49,22 @@ export class AnalyticsMapComponent {
...
 
     const sorted = data.sort((a, b) => b.count - a.count);
-    this.top3 = sorted.splice(0, 3);
+    this.top = sorted.splice(0, this.topCount);
   }
 
   @Input() @boolean selectable = false;
   @Output() selection: EventEmitter<string> = new EventEmitter();
 
+  resetSelected = () => {
```
RemcoSimonides, line 58: unused

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -57,8 +67,15 @@ <h5>
       </ng-container>
     </ng-container>
 
-    <section>
-      <h3>Most active buyers in the world</h3>
+    <section class="surface">
+      <analytics-map class="map-insights" [data]="activeCountries$ | async" [topCount]="5" horizontal showLegend selectable>
+        <h3>
+          <mat-icon svgIcon="movie_open"></mat-icon>
```
RemcoSimonides, line 73: This icon looks too big. Stephanie used to add some padding inside or something. Maybe Daniela or otherwise Anthony know what to do here to fix it.

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -30,7 +30,7 @@ <h3>{{ title.title.international }}</h3>
             <bf-carousel>
               <analytics-pie-chart class="surface" carouselItem [data]="orgActivityOfPopularTitle$ | async">
               </analytics-pie-chart>
-              <analytics-map class="surface" carouselItem [data]="territoryActivityOfPopularTitle$ | async">
+              <analytics-map class="surface caroussel-map" carouselItem [data]="territoryActivityOfPopularTitle$ | async">
```
RemcoSimonides, line 33: `carousel`*

apps/festival/festival/src/app/dashboard/home/home.module.ts
```diff
@@ -9,6 +9,7 @@ import { MatButtonModule } from '@angular/material/button';
 import { MatIconModule } from '@angular/material/icon';
 import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
 import { MatTooltipModule } from '@angular/material/tooltip';
+import { MatListModule } from '@angular/material/list';
```
RemcoSimonides, line 12: unused

libs/analytics/src/lib/components/map/map.component.scss
```diff
@@ -1,23 +1,40 @@
...
   border-radius: 4px;
 }
 
-section {
+.map-wrapper {
   width: 100%;
+  flex: 1 1 100%;
+  display: flex;
+  flex-direction: row-reverse;
+  flex-wrap: wrap;
```
RemcoSimonides, line 24: Not needed anymore

libs/analytics/src/lib/components/map/map.component.scss
```diff
@@ -1,23 +1,40 @@
...
+  gap: 24px;
+  margin-bottom: 24px;
```
RemcoSimonides, line 26: This is used to create space between components on the home component, so it should be there.

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -57,8 +67,15 @@ <h5>
...
+      <analytics-map [data]="activeCountries$ | async" [topCount]="5" horizontal showLegend selectable>
+        <h3>
+          <mat-icon svgIcon="buyer_insights"></mat-icon>
+          <span>Buyers' Insights</span>
+        </h3>
+      </analytics-map>
+
+      <h3>Most active buyers</h3>
       <bf-table [source]="activeBuyers$ | async" clickable useFilter pagination="5">
         <ng-template colRef="user" label="Name" let-item sort>
```
RemcoSimonides, line 80: As I mentioned in a previous comment, all this code should be within
```
    <ng-container *ngIf="titleAnalytics$ | async as analytics; else loading">
      <ng-container *ngIf="analytics.length else empty">
        ...
      </ng-container>
    </ng-container
```

libs/analytics/src/lib/components/map/map.component.html
```diff
@@ -1,49 +1,71 @@
...
-  </article>
+  <ng-content></ng-content>
   <button mat-icon-button *ngIf="selected" (click)="toggleSelect(selected)" matTooltip="Remove selection">
     <mat-icon svgIcon="refresh_filters"></mat-icon>
   </button>
 </header>
 
-<ng-container *ngIf="!isLoading else loading">
-  <world-map *ngIf="top3.length; else empty">
+<section *ngIf="!isLoading else loading" [ngClass]="{'map-wrapper':horizontal}">
```
RemcoSimonides, line 8: For consistency, I wouldn't call this class `map-wrapper` but also `horizontal`
```suggestion
<section *ngIf="!isLoading else loading" [ngClass]="{horizontal}">
```

libs/analytics/src/lib/components/map/map.component.scss
```diff
@@ -1,23 +1,40 @@
...
+  position: relative;
+}
+
+@include layout-bp(lt-md) {
```
RemcoSimonides, line 30: `@include layout-bp() {` should be at the bottom of the css

libs/analytics/src/lib/components/map/map.component.scss
```diff
@@ -1,23 +1,40 @@
...
+  .map-wrapper {
+    flex-direction: column;
+  }
+}
+
+mat-list.horizontal {
+  flex: 0 0 clamp(20%,300px,30%);
```
RemcoSimonides, line 37: This can be more readable.
```suggestion
mat-list.horizontal {
  width: 30%;
}

@include layout-bp(lt-md) {
  width: 100%;
}
```
With this I can instantly see what the UI should look like

libs/analytics/src/lib/components/map/map.component.scss
```diff
@@ -1,23 +1,40 @@
...
 world-map {
+  flex: 1 1 auto;
   height: 300px;
   border-radius: 4px;
 }
 
-section {
+.map-wrapper {
   width: 100%;
+  flex: 1 1 100%;
```
RemcoSimonides, line 21: I don't see any difference if I remove it.

libs/analytics/src/lib/components/map/map.component.scss
yspringfield: This is needed albeit under a different classname.

RemcoSimonides: Can you tell me what it's doing? I don't see any difference if I enable/disable it.

libs/analytics/src/lib/components/map/map.component.scss
```diff
@@ -1,23 +1,32 @@
...
   width: 100%;
+  display: flex;
+  flex-direction: row-reverse;
+  flex-wrap: wrap;
+  gap: 24px;
+  position: relative;
+}
+
+mat-list.horizontal {
+  flex: 0 0 25%;
```
RemcoSimonides, line 29: You didn't like my suggestion of using width?

If you really want to stick with `flex`, you could use `flex-basis: 25%`

RemcoSimonides (approved):

yspringfield:
> Can you tell me what it's doing? I don't see any difference if I enable/disable it.

This is what it looks like without those styles applied.
Bildschirmfoto 2022-05-20 um 8 54 13 AM

apps/festival/festival/src/app/dashboard/home/home.component.scss
```diff
@@ -66,8 +66,17 @@ header {
...
+}
+
+analytics-map h3{
+  display: flex;
+  align-items: center;
+}
+
+.surface analytics-map{
+  margin-bottom: 24px;
+}
```
RemcoSimonides, line 82: Should not target the other map in bf-carousel

RemcoSimonides:
> > Can you tell me what it's doing? I don't see any difference if I enable/disable it.
> 
> This is what it looks like without those styles applied. Bildschirmfoto 2022-05-20 um 8 54 13 AM

I'm talking about `flex-wrap: wrap`. You've added to change flex-direction to column on lt-md screens. So you don't need the wrap anymore, right?

Do you see difference between when it's enabled and when it's disabled?

yspringfield:
ah, thought you meant the entire block. Let me correct that.

apps/festival/festival/src/app/dashboard/home/home.component.scss
yspringfield: it doesn't. The actual selector that matches the `analytics-map` in `bf-carousel` is `analytics-map.surface`

Issue #6989, opened 2021-10-28
Notifications and invitations app-specific


2amarkina:
_Can go directly to dev_

Nowadays all the invitations and notifications are shared between the apps. 

But information about "screening/meetings" should not appear on Content or Financiers, same for the "deals" information that should appear only on Market. 

So it's important to make some notifications/invitations app-specific. 

### Notifications
**Company Management notifications should be cross-app.** 
Examples:
_A user requests to join your organization.	
A user joins or leaves your organization.	
A user's request to join your organization was declined._

**Content Management should be app-specific.** 
Examples:
_A title is successfully published on the marketplace._  
Important to be app-specific as the title will be published on a particular marketplace. 

**Offer Management** should be Archipel Content specific. 
Examples:
_Your offer is successfully sent	
An offer is made on one of your titles._  

**Event Management** should be Archipel Market specific. 
Examples: 
_You are invited to a screening. (RECOMMENDED)	
You are invited to a meeting. (RECOMMENDED)	
A user answers your invitation to an event you're organizing.	
A user wants to join an event you're organizing. (RECOMMENDED)	
Your request to join an event is successfully sent.	
Your request to join an event gets accepted or declined.	
REMINDER - An event you're attending will start in 24 hours. (RECOMMENDED)	
REMINDER - An event you're attending will start in 1 hour. (RECOMMENDED)_ 

### Invitations
**Join Organization** invitations should be cross-app. 
**Attend Event** invitations should be Archipel Market specific.  

**N.B. As we also use the filters on the Invitations page, if possible then:** 
Market: show both filters ("type", "status")
Content: show only "status" filter 
Financiers: show only "status" filter

PR #8497, opened 2022-05-19
App specific notif and invit

Poppe-A:
#6989


libs/model/src/lib/notification.ts
```diff
@@ -31,43 +44,49 @@ export const notificationTypesBase = [
...
 
 export type NotificationTypesBase = typeof notificationTypesBase[number];
+export type NotificationTypesFestival = typeof festivalSpecificNotifications[number];
+export type NotificationTypesCatalog = typeof catalogSpecificNotifications[number];
 type NotificationTypesPlus = typeof notificationTypesPlus[number];
-export type NotificationTypes = NotificationTypesBase | NotificationTypesPlus;
+export type NotificationTypes = NotificationTypesBase | NotificationTypesPlus | NotificationTypesFestival | NotificationTypesCatalog;
+
+export function isCatalogNotification(notification: Notification, app: App) {
+  if (app === 'catalog') {
```
RemcoSimonides, line 72: Why do you check for app here? It depends on the notification.type, no?

libs/model/src/lib/notification.ts
```diff
@@ -31,43 +44,49 @@ export const notificationTypesBase = [
...
+    return catalogSpecificNotifications.includes(notification.type as NotificationTypesCatalog) 
+      || notification.type === 'movieAccepted' && notification._meta.createdFrom === 'catalog'
+      || notification.type === 'movieSubmitted';
```
RemcoSimonides, line 75: Something is going wrong here and also it doesn't look like a scalable solution

libs/model/src/lib/notification.ts
Poppe-A: Yup but I need to know the current app to display relative notif only

libs/model/src/lib/notification.ts
```diff
@@ -8,66 +8,66 @@ import { StorageFile } from './media';
...
-  // Notifications relative to invitations
-  'requestFromUserToJoinOrgPending', // Notification sent to the user that made the request
-  'invitationToJoinOrgDeclined',
-
-  // Other notifications
-  'movieSubmitted', // (catalog only)
-  'organizationAcceptedByArchipelContent',
-  'orgAppAccessChanged',
-  'userRequestAppAccess',
+export const notifications = {
```
RemcoSimonides, line 11: Add
```
financiers: [] as const,
crm: [] as const,
```
So it's equal to App type and you don't need to do `as 'catalog' | 'festival'` later

libs/user/src/lib/auth/forms/notifications-form/notifications-form.component.ts
```diff
@@ -105,7 +106,7 @@ const tables: { title: string, types: string[], appAuthorized: App[] }[] = [
 })
 export class NotificationsFormComponent {
 
-  public types = [...notificationTypesBase];
+  public types = [...notificationTypes];
```
RemcoSimonides, line 109: Why not just `public types = notificationTypes;`?

libs/model/src/lib/notification.ts
```diff
@@ -8,66 +8,66 @@ import { StorageFile } from './media';
...
+export type NotificationTypes = typeof notifications.catalog[number] | typeof notifications.festival[number] | typeof notifications.shared[number];
+export const notificationTypes = [...notifications.festival, ...notifications.catalog, ...notifications.shared];
+type AppNotificationType = typeof notifications;
+function isNotificationType<K extends keyof AppNotificationType>(
+  type: NotificationTypes,
+  kind: K
+): type is AppNotificationType[K][number] {
+  const notif: string[] = notifications[kind] as any;
+  return notif.includes(type);
+}
```
RemcoSimonides, line 66:
```suggestion
type AppNotificationType = typeof notifications;
type AppNotificationKey = keyof AppNotificationType;
export type NotificationTypes = typeof notifications[AppNotificationKey][number];
export const notificationTypes = [...notifications.festival, ...notifications.catalog, ...notifications.shared];
function isNotificationType<K extends keyof AppNotificationType>(
  type: NotificationTypes,
  kind: K
): type is AppNotificationType[K][number] {
  const notif: string[] = notifications[kind] as any;
  return notif.includes(type);
}
```

libs/notification/src/lib/+state/notification.service.ts
```diff
@@ -46,6 +47,7 @@ export class NotificationService extends CollectionService<NotificationState> {
     switchMap((notifications) => {
       const promises = notifications
         .filter(n => n.app?.isRead !== undefined)
+        .filter(n => isAppNotification(n.type, this.app as 'catalog' | 'festival'))
```
GrandSchtroumpf, line 50: How do you know it's not `financiers` ?

GrandSchtroumpf (approved):

PR #8503, opened 2022-05-19
Organization denomination removing - Model part

KevinDurand974:
- [x] remove `interface Denomination` from `PublicOrganization`

⚠ Test doesnt work (that's normal)


KevinDurand974:
This will be open again when I've done the local migration and testing.

PR #8514, opened 2022-05-24
George/refactor 2

george43g:
this one was a bit trickier, because I had to change logic

Firebase is annoying because their timestamp class/type is different depending on node version `firebase-admin` or web version `firebase`. Which causes issues for data models and related functions often.

using `instanceof Date` is safer.


phpgeekfr:
linked with #8006

apps/backend-functions/src/notification.ts
```diff
@@ -320,6 +320,7 @@ export async function onNotificationCreate(snap: FirebaseFirestore.DocumentSnaps
         break;
     }
 
+    // ! TODO! Don't do this - unify instantiation of Firestore
```
phpgeekfr, line 323: add issue number

libs/model/src/lib/emails/index.ts
```diff
@@ -0,0 +1,2 @@
+
+export type EmailErrorCodes = 'E01-unauthorized' | 'E02-general-error' | 'E03-missing-api-key' | 'E04-no-template-available';
```
phpgeekfr, line 2: already done in another PR

libs/model/src/lib/timestamp.ts
```diff
@@ -1,3 +1,25 @@
...
+  }
+  if (typeof date === 'string' || typeof date === 'number') {
+    return new Date(date);
+  }
+  if (!date) {
+    // Return null to avoid undefined error with firestore
+    return null;
+  }
+  return date;
+}
```
phpgeekfr, line 25: no need to move, will be done in #8006

PR #8517, opened 2022-05-24
8245 bf carousel not use matlayoutmodule

### LINKED ISSUE #8245: Bf-carousel - use css flexbox or grid instead of custom layout
The bf-carousel uses our own layout directive. It should be replaced by css grid or flexbox.

Why? The bf-carousel on festival/dashboard contains charts which don't have a defined size. When the carousel moves, the charts would scale incorrectly. Now it uses css to make the width fixed so it doesn't happen, but that shouldn't have been needed in the first place. Changing this layout system in bf-carousel could fix that.

RemcoSimonides:
fix #8245


libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -73,18 +91,22 @@ export class CarouselComponent implements AfterViewInit, AfterContentInit, OnDes
 
   ngAfterContentInit() {
     this.amount$ = this.items.changes.pipe(startWith(this.items), map(items => items.length));
-    this.itemsSub = this.items.changes.subscribe(() => this.showForward = !!this.scrollable.measureScrollOffset('right'));
+    this.itemsSub = this.items.changes.subscribe(items => {
+      this.showForward = items.length > this._columns;
+      this.cdr.markForCheck();
+    });
```
RemcoSimonides, line 97: The previous `this.showForward = !!this.scrollable.measureScrollOffset('right'));` was now only working when inside a `setTimeout(() => {}, 0)`. Figured this works just as well except when the user already scrolled all the way to the right. In that rare case it will still show a showForward button even though the user is already all the way to the right.

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -53,7 +59,19 @@ export class CarouselComponent implements AfterViewInit, AfterContentInit, OnDes
   @ViewChild('container') container: ElementRef<HTMLDivElement>;
   @ContentChildren(CarouselItemDirective) items: QueryList<CarouselItemDirective>;
 
-  constructor(private cdr: ChangeDetectorRef, private ngZone: NgZone) { }
+  @Input() set columns(columns: 2 | 3 | 4 | 5 | 6) {
+    this._columns = columns;
+  }
```
RemcoSimonides, line 64: `columns` now about how many items the user want to show on 12 column grid. Before it was about how many columns you would like each item to take. What is more intuitive in your opinion?

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -4,28 +4,53 @@ import {
...
+  md: number;
+  lg: number;
+  xl: number;
+
+  gtXs: number;
+  gtSm: number;
+  gtMd: number;
+  gtLg: number;
+
+  ltSm: number;
```
GrandSchtroumpf, line 36: This is the same as xs. You don't need it

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -4,28 +4,53 @@ import {
...
+  xs: number;
+  sm: number;
+  md: number;
+  lg: number;
+  xl: number;
+
+  gtXs: number;
+  gtSm: number;
+  gtMd: number;
+  gtLg: number;
```
GrandSchtroumpf, line 34: this is the same as xl. You don't need it

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -4,28 +4,53 @@ import {
...
+
+  ltSm: number;
+  ltMd: number;
+  ltLg: number;
+  ltXl: number;
+}
+
+const sizes: Record<string, string[]> = {
```
GrandSchtroumpf, line 42: It's better to do this in the opposite way : 
```typescript
const sizes: Record<string, string> = {
  xs: ['xs'],
  ltMd: ['xs', 'sm'],
  ...
}
```
See why in the comment below

libs/ui/src/lib/carousel/carousel.component.ts
```diff
@@ -50,13 +72,31 @@ export class CarouselComponent implements AfterViewInit, AfterContentInit, OnDes
...
+  @Input() set columns(amount: Partial<Columns> | string) {
+    if (!amount) return;
+
+    if (typeof amount === 'object') {
+      for (const [size, values] of Object.entries(sizes)) {
+        const index = Object.keys(amount).find(key => values.includes(key));
+        if (index) {
+          this.el.nativeElement.style.setProperty(`--${size}-columns`, amount[index]);
+        }
+      }
```
GrandSchtroumpf, line 86: with the new system : 
```suggestion
      for (const [size, value] of Object.entries(amount)) {
        sizes[size]?.forEach(label => this.el.nativeElement.style.setProperty(`--${label}-columns`, value));
      }
```

GrandSchtroumpf (approved):

Issue #7338, opened 2021-12-03
Email - Event invitation - Add multiple time zones


Jackseed:
Currently event invitation emails display only start & end time in GMT+0 as we don't know users' timezones. 

Instead we should give them a range of classic timezones to help them find the relevant one. 

Classic timezones should be:
- Pacific (UTC-8)
- Central America (UTC-6)
- Eastern America (UTC-5)
- Eastern Europe (UTC+1)
- Central Europe (UTC+2)
- Japan (UTC+9)

To limit dates (since they can change from timezone to timezone), we will only show the start date for UTC and precise it if it's different for a specific timezone. 
We will also display the duration of the event (eg "3 hours" or "3 days"). Minutes should be used for events that last less than an hour, hours for events that last less than 72 hours and days after. 

Invitation emails already have a lot of info, that's why we need UI here: to make the time and date and possible other relevant info (like credentials) visible and readable. 

This time is used in 3 different emails:
- [AM -EVENT - Non private Invitation with credential ](https://mc.sendgrid.com/dynamic-templates/d-5123a0ceda2d463db60717dc94005b65/version/9178f554-c088-4bc6-8320-6d062980b077/editor)(`Template ID: d-5123a0ceda2d463db60717dc94005b65`)
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/41ead55b-9427-4ba3-b591-bddc70199b33)

- [AM - EVENT - Invitation with credentials](https://mc.sendgrid.com/dynamic-templates/d-ce3e57248a694cefacad49bc4c820078/version/e57ca86c-9e2d-46ff-8694-f6b12f624e17/editor)
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/118d509f-1e7b-4b95-96ad-e04a742aa153)

- [AM - EVENT - Invitation without credentials](https://mc.sendgrid.com/dynamic-templates/d-1a7cc9ca846c4ae1b4e8cf8a76455cc5/version/1f6aee13-906d-4ef3-b7da-2bcc6a0be459/editor)
![image.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/05201345-375e-4c25-ac24-a9e47e93379d)

frenchyrules:
Here is a first UI suggestion for AM -EVENT - Non private Invitation with credential. @Jackseed @cams-goussale 
[...]

PR #8531, opened 2022-05-26
Add multiple time zones

Poppe-A:
#7338


libs/utils/src/lib/emails/utils.ts
```diff
@@ -178,12 +187,40 @@ export function getEventEmailData({ event, orgName, attachment = true, email, in
...
     viewUrl: `/event/${event.id}/r/i${eventUrlParams}`,
     sessionUrl: `/event/${event.id}/r/i/session${eventUrlParams}`,
     accessibility: event.accessibility,
-    calendar: attachment ? getEventEmailAttachment(event, orgName) : undefined
+    calendar: attachment ? getEventEmailAttachment(event, orgName) : undefined,
+    eventDuration: getEventDuration(eventStartDate, eventEndDate)
+  }
+}
+
+function getTimezoneEventStartHour(date: Date, timezone: string) {
```
RemcoSimonides, line 208: This function name is almost longer than the function itself ^^ May I suggest something like `toTimezone()`

And what are the options for the timezones? Could you create or find a type for the different timezones?

libs/utils/src/lib/emails/utils.ts
```diff
@@ -178,12 +187,40 @@ export function getEventEmailData({ event, orgName, attachment = true, email, in
...
+  const tzDate = new Date(date.toLocaleString("en-us", {timeZone: timezone}));
```
RemcoSimonides, line 209:
```suggestion
  const tzDate = new Date(date.toLocaleString("en-us", { timeZone }));
```

libs/utils/src/lib/emails/utils.ts
```diff
@@ -78,12 +78,21 @@ export interface EventEmailData {
   id: string;
   title: string;
   start: string;
+  otherTimezones: {
+    centralAmerica: string, // start  plus/minus x hours
```
RemcoSimonides, line 82: I don't understand this comment

libs/utils/src/lib/emails/utils.ts
```diff
@@ -78,12 +78,21 @@ export interface EventEmailData {
...
+    japan: string,
+    pacific: string,
+  }
   end: string;
   type: EventTypesValue;
   viewUrl: string;
   sessionUrl: string;
   accessibility: AccessibilityTypes;
   calendar: AttachmentData;
+  eventDuration: string;
```
RemcoSimonides, line 95: Everything in this object is about event. So I think the name `duration` is sufficient.

libs/utils/src/lib/emails/utils.ts
```diff
@@ -178,12 +187,40 @@ export function getEventEmailData({ event, orgName, attachment = true, email, in
...
+  const startTs = start.getTime();
+  const endTs = end.getTime();
+  const duration = intervalToDuration({start, end});
+  
+  if (endTs - startTs < millisecondsInHour) {
+    return formatDuration(duration, {format: ['minutes']});
+  } else if (endTs - startTs < millisecondsInHour * 72) {
+    return formatDuration(duration, {format: ['hours']});
+  } else {
+    return `${differenceInDays(endTs, startTs)} days`;
```
RemcoSimonides, line 223:
```suggestion
  const duration = intervalToDuration({start, end});
  
  if (duration.hours < 2) {
    return formatDuration(duration, {format: ['minutes']});
  } else if (duration.days < 3) {
    return formatDuration(duration, {format: ['hours']});
  } else {
    return `${differenceInDays(endTs, startTs)} days`;
  }
```

It's not fully the same logic but I think it's better. With your function, if the duration would be 1 hour and 5 minutes, it would return "1 hours"

RemcoSimonides, line 223:
```suggestion
  const duration = end.getTime() - start.getTime();
  
  if (duration < millisecondsInHour) {
    return `${differenceInMinutes(end, start)} minutes`;
  } else if (duration < millisecondsInHour * 72) {
    return `${differenceInHours(end, start)} hours`;
  } else {
    return `${differenceInDays(end, start)} days`;
```

RemcoSimonides (approved):

PR #8546, opened 2022-05-27
proposal to remove timestamps

phpgeekfr:


apps/backend-functions/src/internals/firebase.ts
```diff
@@ -30,6 +30,67 @@ export async function getUserMail(userId: string): Promise<string | undefined> {
...
+    if (firstArg instanceof Change) {
+      const before: FirebaseFirestore.DocumentSnapshot = firstArg.before;
+      const after: FirebaseFirestore.DocumentSnapshot = firstArg.after;
+      const changes: BlockframesChange = {
+        before: {
+          id: before.id,
+          exists: before.exists,
+          ref: before.ref,
+          data: () => before.data() ? toDate(before.data()) : undefined
+        },
```
GrandSchtroumpf, line 71: Maybe
```suggestion
        before: {
          ...before
          data: () => before.exists ? toDate(before.data()) : undefined
        },
```

phpgeekfr: this is not working.
With this, `change.before.ref` returns `undefined`

libs/devops/src/lib/db-cleaning.ts
```diff
@@ -110,7 +110,7 @@ export function cleanNotifications(
   return runChunks(
     notifications.docs,
     async (doc: QueryDocumentSnapshot) => {
-      const notification = doc.data() as NotificationDocument;
+      const notification = toDate(doc.data()) as Notification;
```
GrandSchtroumpf, line 113: Maybe you can update `toDate` to accept a generic. It would looks nicer than using `as` at least

libs/firebase-utils/src/lib/firebase-utils.ts
```diff
@@ -2,13 +2,41 @@ import * as admin from 'firebase-admin';
...
+        console.log(`${key} is not a Firebase Timestamp`);
+      }
+      continue;
+    }
+    toDate(value);
+  }
+  return target;
+}
+
+export function getDocumentRef(path: string, db = admin.firestore(), tx?: FirebaseFirestore.Transaction): Promise<FirebaseFirestore.DocumentSnapshot<FirebaseFirestore.DocumentData>> {
```
GrandSchtroumpf, line 23: should be renamed `getDocumentSnap`.

libs/movie/src/lib/movie/form/movie.form.ts
```diff
@@ -863,11 +862,7 @@ export class ShootingDateForm extends FormEntity<ShootingDateFormControl> {
 export function createMovieShootingDate(
   params: Partial<MovieShootingDate> = {}
 ): MovieShootingDate {
-  return {
-    ...params,
-    completed: toDate(params.completed),
-    progress: toDate(params.progress),
-  };
+  return {...params};
```
GrandSchtroumpf, line 865: not very useful anymore. Maybe remove it altogether

GrandSchtroumpf (approved):
Great work. API for backend `getDocument` / `toDate` / ... could be improved, but it can be in another PR.

apps/backend-functions/src/internals/invitations/organizations.ts
```diff
@@ -18,15 +19,11 @@ async function addUserToOrg(userId: string, organizationId: string) {
...
-  const userRef = db.collection('users').doc(userId);
-  const organizationRef = db.collection('orgs').doc(organizationId);
-  const permissionsRef = db.collection('permissions').doc(organizationId);
-
   return db.runTransaction(async tx => {
     const [user, organization, permission] = await Promise.all([
-      tx.get(userRef),
-      tx.get(organizationRef),
-      tx.get(permissionsRef)
+      getDocumentRef(`users/${userId}`, db, tx),
```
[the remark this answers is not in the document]
phpgeekfr: renamed to `getDocumentSnap`. But I don't think I can use `getDocument` instead because I'm using `ref` some lines after.

Or maybe I misunderstood your comment

libs/devops/src/lib/db-cleaning.ts
phpgeekfr: How would you change the function ?

Currently
```
export function toDate<D>(target: D): D {
  if (!target) return;
  if (typeof target !== 'object') return target;
  for (const key in target) {
    const value = target[key];
    if (!value || typeof value !== 'object') continue;
    if (!!value['_seconds'] && value['_nanoseconds'] >= 0) {
      try {
        target[key] = (value as any).toDate();
      } catch (_) {
        console.log(`${key} is not a Firebase Timestamp`);
      }
      continue;
    }
    toDate(value);
  }
  return target;
}
```

PR #8553, opened 2022-05-30
some backend-function cleaning

phpgeekfr:


george43g (approved):
I havent tested the code myself, so as long as all the tests pass we should be OK. Logically this is a great refactor that simplifies our code. A good first step - as we progress we can clean this up further. But for a PR I think is ok as-is

PR #8560, opened 2022-05-31
7849 Genres Tendancy

### LINKED ISSUE #7849: (C) Sellers' Analytics - Buyer Page - Genres Tendancy Pie Chart
[Wireframe](https://www.figma.com/file/FhY2dJqeYrVTXh7Q7HA0m7/%E2%9A%99%EF%B8%8F-Sellers'-Dashboard-(DONE)?node-id=372%3A88942)
[UI Design](https://www.figma.com/file/aNbs0XSJBqIvKrffJSzViH/Dashboard?node-id=122%3A25475)

RemcoSimonides:
fix #7849


apps/festival/festival/src/app/dashboard/analytics/buyer/buyer-analytics.component.ts
```diff
@@ -118,6 +131,17 @@ function fromUser(invitation: Invitation, uid: string) {
...
+/**
+ * Converts array from Title With Analytics[] to Analytics[] with Title
+ */
+function swapTitleWithAnalytics(titles: MovieWithAnalytics[]): AnalyticsWithTitle[] {
+  const analytics = titles.map(title => title.analytics).flat();
+  for (const analytic of analytics) {
+    analytic['title'] = titles.find(title => title.id === analytic.meta.titleId);
+  }
+  return analytics;
+}
```
RemcoSimonides, line 143: Curious to know if you'd know a better way because `analytic['title'] = ` is not typesafe. The return value of the function is typed though. But maybe with your knowledge from joinWith, there might be some neat way.

apps/festival/festival/src/app/dashboard/analytics/buyer/buyer-analytics.component.scss
```diff
@@ -42,10 +42,31 @@ section {
     flex-wrap: wrap;
     margin-bottom: 8px;
   }
+
+  .graphs {
+    display: flex;
+    gap: 24px;
+
+    analytics-bar-chart {
+      flex: 2;
```
GrandSchtroumpf, line 51: use flex-grow instead to avoid having all the side effect style of `flex`

apps/festival/festival/src/app/dashboard/analytics/buyer/buyer-analytics.component.html
```diff
@@ -60,8 +60,14 @@ <h3>Buyer was interested in these titles</h3>
     </bf-table>
   </article>
 
-  <article class="surface">
+  <article class="surface graphs">
     <analytics-bar-chart [data]="totalAnalyticsPerTitle$ | async"></analytics-bar-chart>
+    <analytics-pie-chart class="surface" [data]="aggregatedPerGenre$ | async">
+      <h5>
```
GrandSchtroumpf, line 66: It's not normal to have to put the icon & span in the h5 and then transform the h5 into a flex. I think there is an issue with how you handle the ng-content.

apps/festival/festival/src/app/dashboard/analytics/buyer/buyer-analytics.component.ts
GrandSchtroumpf: `MovieWithAnalytics['analytics']` should be a `AnalyticsWithTitle`. But this is a weird way of managing the analytics I think. Best would be to have the id of the title in the analytics and fetch it like we do in the other documents.
To be honest I don't see why you need that at all. You can just do `aggregate(title.analytics)` instead or `aggregate(analytics)`.

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -34,7 +34,15 @@ <h2>Interested Users</h2>
     </a>
 
     <div fxLayout fxLayout.lt-sm="column" fxLayoutGap="16px">
-      <analytics-pie-chart fxFlex class="surface" [data]="orgActivity$ | async" (selection)="filterValue = $event"></analytics-pie-chart>
+      <analytics-pie-chart fxFlex class="surface" [data]="orgActivity$ | async" (selection)="filterValue = $event">
+        <article>
```
GrandSchtroumpf, line 38: You shouldn't use an article here.

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -28,7 +28,15 @@ <h3>{{ title.title.international }}</h3>
             </header>
             <bf-carousel [columns]="{ gtSm: 2, ltMd: 1 }">
               <ng-template carouselItem>
-                <analytics-pie-chart class="surface" [data]="orgActivityOfPopularTitle$ | async"></analytics-pie-chart>
+                <analytics-pie-chart class="surface" [data]="orgActivityOfPopularTitle$ | async">
+                  <article>
```
GrandSchtroumpf, line 32: same

libs/analytics/src/lib/utils.ts
```diff
@@ -5,19 +5,24 @@ import { getDeepValue } from '@blockframes/utils/pipes';
...
-    const key = getDeepValue(analytic, path) as string | number;
-    if (!key) continue;
+
+  const count = (key: string | number) => {
     if (!counter[key]) counter[key] = 0;
     counter[key]++;
   }
 
+  for (const datum of data) {
+    const value = getDeepValue(datum, path) as string | number;
```
GrandSchtroumpf, line 17: If I understand your code, it can be an array too
```suggestion
    const value = getDeepValue(datum, path) as string | number | (string|number)[];
```

libs/analytics/src/lib/utils.ts
```diff
@@ -5,19 +5,24 @@ import { getDeepValue } from '@blockframes/utils/pipes';
...
+    if (!value) continue;
```
GrandSchtroumpf, line 18: what if value is 0 ?

RemcoSimonides: Good catch! 0 is definitely allowed

libs/analytics/src/lib/utils.ts
```diff
@@ -1,23 +1,43 @@
...
+}
+
+export function countedToAnalyticData(record: Record<string | number, number>, scope?: Scope): AnalyticData[] {
+  return Object.entries(record).map(([key, count]) => ({
     key,
     count,
-    label: scope ? toLabel(key, scope) : ''
+    label: scope ? toLabel(key, scope) : key
   }));
 }
```
RemcoSimonides, line 42: I've extracted the `countedToAnalyticData` function from the counter because I felt like it does something different than what I intended 'counter()` to do. Even though every `counter()` also uses `countedToAnalyticData` afterwards at the moment.

GrandSchtroumpf (approved):

Issue #7724, opened 2022-02-08
Avails date in custom CMS link


cams-goussale:
In CMS, we can set up the redirection to the /library page filtered by some parameters. Currently, Yuki set it up in catalog app, so that user is redirected either to Feature films, either to TV shows.

![image.png](https://images.zenhubusercontent.com/617811d92b89e7733866d8a0/4bb8a689-12fa-4518-a9d3-0fc431846350)

The problem is that the **avails start date** is saved at the date when she set up the link in CMS, which makes the start date in the past. Is there a way to include in the CMS link that avails start date == today?

This is the link we put for Feature films button:
`../title?formValue=%7B"search":%7B"query":"","genres":%5B%5D,"originCountries":%5B%5D,"contentType":"feature_film"%7D,"avails":%7B"medias":null,"exclusive":true,"territories":%5B%5D,"duration":%7B"from":"2022-02-03T23:00:00.000Z","to":"2023-02-03T23:00:00.000Z"%7D%7D%7D`

This is the one for TV Shows:
`../title?formValue=%7B"search":%7B"query":"","genres":%5B%5D,"originCountries":%5B%5D,"contentType":"series"%7D,"avails":%7B"medias":null,"exclusive":true,"territories":%5B%5D,"duration":%7B"from":"2022-02-03T23:00:00.000Z","to":"2023-02-03T23:00:00.000Z"%7D%7D%7D`

RemcoSimonides:
For dev: Something is similar with the `startAt` property. This can be set to "now". Check out how that works and do something similar for the where queries.
`libs/utils/src/lib/pipes/cms.pipe.ts`

PR #8570, opened 2022-06-02
fix Avails date

KvnMrs:
Issue: https://github.com/blockframes/blockframes/issues/7724
Before: 

https://user-images.githubusercontent.com/86494393/171636306-a386c458-53fe-49f0-9b90-92945b43c92b.mp4

Now: 

https://user-images.githubusercontent.com/86494393/171642627-e478496c-030b-4792-bf5d-2c48c492c941.mp4


apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
```diff
@@ -97,9 +97,9 @@ export class ListComponent implements OnDestroy, OnInit {
       avails = {}
     } = decodeUrl(this.route);
 
-    if (avails.duration?.from) avails.duration.from = new Date(avails.duration.from);
-    if (avails.duration?.to) avails.duration.to = new Date(avails.duration.to);
-
+    const currentDate = new Date()
+    if (avails.duration?.from) avails.duration.from = new Date();
+    if (avails.duration?.to) avails.duration.to = new Date(currentDate.setFullYear(currentDate.getFullYear() + 1));
```
RemcoSimonides, line 102: Avails can be set through the url.
For example, if you fill in the avails filter, then click on a title, then navigate back. It should have the avails filter still filled in. That's why the date was set like `new Date(avails.duration.from);`. Please don't remove it.

Also, you should use `date-fns` functions to simplify readability (https://date-fns.org/)

KvnMrs: @Remco With this last modification, we should change the value of "duration.to" in the CMS link to "nextYear". ( seen with Camille )

Now: 

https://user-images.githubusercontent.com/86494393/171886185-7e5030fa-a721-4944-84ed-72afed0ea25f.mp4

apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
```diff
@@ -97,8 +98,14 @@ export class ListComponent implements OnDestroy, OnInit {
...
-    if (avails.duration?.from) avails.duration.from = new Date(avails.duration.from);
-    if (avails.duration?.to) avails.duration.to = new Date(avails.duration.to);
+    const currentDate = new Date();
+    const nextYears = add(new Date(), { years: 1 });
+    const formatedCurrentDate = lightFormat(new Date(currentDate), 'yyyy-MM-dd');
+    const formatedNextYears = lightFormat(new Date(nextYears), 'yyyy-MM-dd');
+
+    if (avails.duration?.from) avails.duration.from = isPast(new Date(avails.duration.from)) ? currentDate : new Date(avails.duration.from);
+    if (avails.duration?.to === "nextYears") avails.duration.to = nextYears;
+    else if (avails.duration?.to && formatedNextYears !== formatedCurrentDate) new Date(avails.duration.to);
```
RemcoSimonides, line 108: `formatedNextYears !== formatedCurrentDate` this is never equal to each other

libs/model/src/lib/static/static-model.ts
```diff
@@ -237,6 +237,7 @@ export const medias = {
   aVod: 'A-VOD', // a-vod
   fVod: 'F-VOD', // f-vod
   sVod: 'S-VOD', // s-vod
+  tVod: 'T-VOD',
```
RemcoSimonides, line 240: What's this change for?

apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
RemcoSimonides: I'll ask Francois if he perhaps can think of a better solution

apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
KvnMrs: Yes, first I set the value `nextYear` (new Date(), year +1) to `avails.duration.to`. Then to keep the value selected by the user if there is one I do this condition. 
Otherwise `avails.duration.to` is always equal to `nextYear`.

libs/model/src/lib/static/static-model.ts
KvnMrs: when I have free time, I look at the next problem, I forget to delete it or commit it to another branch.

apps/catalog/catalog/src/app/marketplace/title/list/list.component.ts
RemcoSimonides: No, you're using `else if` so it doesn't go through both `if` statements. If `avails.duration.to === "nextYear"` then it doesn't check for `else if (avails.duration.to && formatedNextyears !== formatedCurrentDate` anymore

GrandSchtroumpf (approved):

Issue #8524, opened 2022-05-25
remove emulator config from env.ts files


george43g:
we already have a way of getting emulators to be the target,
ie
`npx nx serve <app> --configuration=emulator`

phpgeekfr:
update with new command to use
- https://www.notion.so/cascade8/Emulator-79492738d2614b35b6435eb80584ff26
- https://www.notion.so/cascade8/E2E-testing-with-Emulator-4069418c783444a3951247781040337c

PR #8575, opened 2022-06-03
remove emulator config from env.ts files

george43g:
meant to resolve issue #8524


libs/utils/src/lib/emulator-front-setup.ts
```diff
@@ -2,6 +2,7 @@ import { InjectionToken } from '@angular/core';
 import { Auth, connectAuthEmulator } from 'firebase/auth';
 import { connectFirestoreEmulator, Firestore } from 'firebase/firestore';
 import { connectFunctionsEmulator, Functions } from 'firebase/functions';
+import firebaseSetup from 'firebase.json'
```
phpgeekfr, line 5:
```suggestion
import firebaseSetup from 'firebase.json';
```

apps/catalog/catalog/src/environment/environment.ts
```diff
@@ -1,4 +1,3 @@
 import { getEmulatorsConfig } from "@blockframes/utils/emulator-front-setup";
-import { emulators } from "@env";
 
-export const emulatorConfig = getEmulatorsConfig(emulators);
+export const emulatorConfig = getEmulatorsConfig({ auth: false, firestore: false, functions: false });
```
phpgeekfr, line 3:
```suggestion
export const emulatorConfig = getEmulatorsConfig();
```

and put this `{ auth: false, firestore: false, functions: false }` as default in `getEmulatorsConfig(`

```
export function getEmulatorsConfig(emulators: EnabledEmulators = { auth: false, firestore: false, functions: false }) {
```

phpgeekfr (approved):

Issue #8355, opened 2022-05-04
org.denomination.public should be removed


phpgeekfr:
`org.denomination.public` is still used on the app but it is not set on org creation and cannot be updated/set after.
Looks like only org import allows this.

So new organizations does not have public denomination

check existing issue  #2860

Jackseed:
Unfortunately I think we can back pedal and go back to org.name. 
Public / Private was meant to be used for in-app invoicing but it's not on the roadmap so far

phpgeekfr:
- [ ] updating org name does not update algolia indexes if org have denomination.full defined

on the form, input is for `denomination.full` 
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/bc8975fe-39e3-4b5f-a7d9-2f194a3c47f8)

but `orgName` pipe displays `denomination.public`  by default (example in `<organization-display>`) and algolia is also updated with public denomination by default

( NB: only old orgs have `denomiation.public` defined)

Quick fix would be to update both public and full denomination

phpgeekfr:
- [ ] remove `interface Denomination` from `PublicOrganization`
- [ ] remove `orgName` Pipe and function
- [ ] create a migration to change org documents from `denomination` to `name`
  => this should be  a separate issue. 
  => if org have only denomination.full  we will take this value for `org.name`
  => if org have only denomination.public  we will take this value for `org.name`
  => if org have both but with equal value , take one or another
  => (on prod db) @phpgeekfr create a list of org that have both full and public with distinct values to decide what to do
- [ ] remove org.fiscalNumber from code and from db (in same migration as above)
- [ ] update various org forms

RemcoSimonides:
If this still goes in Release 4.3, could you add it to the [release issue](https://github.com/blockframes/blockframes/issues/8606) please? So it will be tested

PR #8576, opened 2022-06-06
Removing Org Denomination - OrgName Pipe removal

KevinDurand974:
Continuation of https://github.com/blockframes/blockframes/issues/8355

Remove OrgName Pipe

> Minimal required to launch app

- Next branch is the removal of `fiscalNumber` from code
- Next branch is the removal of ANY residual of `denomination`
- Next branch is the update of Excel template

⚠⚠ Some test will not pass and that perfectly normal ⚠⚠


libs/devops/src/lib/firestoreMigrations/0079.ts
```diff
@@ -71,8 +65,8 @@ async function migrateNotifications(db: Firestore) {
   return runChunks(notifications.docs, async (doc) => {
     const notificationBase = doc.data() as Notification;
 
-    // Skip all notifications without organization or organization.denomination
-    if (!notificationBase?.organization || !(notificationBase.organization as any)?.denomination) return false;
+    // Skip all notifications without organization
+    if (!notificationBase?.organization) return false;
```
RemcoSimonides, line 69: I think you should keep the name check
```suggestion
    // Skip all notifications without organization or organization name
    if (!notificationBase?.organization?.name) return false;
```
That is probably why denomination was checked for previously too.

libs/organization/src/lib/organization/components/organization-display/organization-display.component.html
```diff
@@ -5,7 +5,7 @@
       <!-- using organizationInformations to update display as user type in the form -->
       <img mat-card-avatar type="logo" [ref]="org.logo" asset="empty_organization.svg" alt="Organization logo">
       <article class="information" fxLayout="column" fxLayoutAlign="start start">
-        <h4>{{ organization | orgName }}</h4>
+        <h4>{{ organization }}</h4>
```
RemcoSimonides, line 8:
```suggestion
        <h4>{{ organization.name }}</h4>
```

RemcoSimonides (approved):

PR #8577, opened 2022-06-06
Removing Org Denomination - Fiscal Number removal (from code only)

KevinDurand974:
Continuation of https://github.com/blockframes/blockframes/issues/8355

Remove FiscalNumber from code

> Minimal required to launch app

- Next branch is the removal of ANY residual of `denomination`
- Next branch is the update of Excel template

⚠⚠ Some test will not pass and that perfectly normal ⚠⚠


libs/import/src/lib/view-extracted-elements/organizations/utils.ts
```diff
@@ -75,37 +74,29 @@ export async function formatOrg(sheetTab: SheetTab, organizationService: Organiz
       if (!activity) throw wrongValueError(value, 'Activity');
       return activity;
     },
-    /* d */ 'org.fiscalNumber': (value: string) => {
-      if (!value) throw optionalWarning('Fiscal Number');
-      return value;
-    },
-    /* e */ 'org.addresses.main.street': (value: string) => {
```
KevinDurand974, line 82: I've already fixed this on next branch when I saw it (my bad)

libs/import/src/lib/view-extracted-elements/organizations/utils.ts
```diff
@@ -75,37 +74,29 @@ export async function formatOrg(sheetTab: SheetTab, organizationService: Organiz
...
-      if (!value) throw optionalWarning('Fiscal Number');
-      return value;
-    },
```
KevinDurand974, line 85: I already saw that I removed street part, It's fixed in next branch (sorry)

RemcoSimonides (approved):

PR #8579, opened 2022-06-06
8578 Remove file from queue when removing from bf-form-list

### LINKED ISSUE #8578: [Media] Remove file from bf-form-list does not remove file from queue
This bug applies to notes and otherVideos

How to reproduce using otherVideos:
1. Go to movie tunnel => Promotional elements => Videos
1. Select a video
2. Click on "Save video"
3. Remove file from the list
4. Click on Save and you'll see that the file is uploading even though you just removed it from the list.

RemcoSimonides:
fix #8578

### DIFF
--- a/libs/media/src/lib/file-uploader.service.ts
+++ b/libs/media/src/lib/file-uploader.service.ts
@@ -69,7 +69,9 @@ export class FileUploaderService {
 
     if (!uploads || !uploads.length) return;
 
-    const index = typeof fileNameOrIndex === 'number' ? fileNameOrIndex : uploads.findIndex(upload => upload?.fileName === fileNameOrIndex);
+    const index = typeof fileNameOrIndex === 'number'
+      ? fileNameOrIndex
+      : uploads.findIndex(upload => upload?.fileName === fileNameOrIndex);
 
     // ! Do not remove/splice otherwise it will shift remaining uploads and can cause weird side effects
     if (index !== -1) uploads[index] = null;
--- a/libs/media/src/lib/file-uploader.service.ts
+++ b/libs/media/src/lib/file-uploader.service.ts
@@ -64,13 +64,12 @@ export class FileUploaderService {
    * @note if the file does not exists in the queue,
    * this function will not throw any error and simply do nothing
    */
-  removeFromQueue(storagePath: string, fileName: string) {
-
+  removeFromQueue(storagePath: string, fileNameOrIndex: string | number) {
     const uploads = this.queue[storagePath];
 
     if (!uploads || !uploads.length) return;
 
-    const index = uploads.findIndex(upload => upload?.fileName === fileName);
+    const index = typeof fileNameOrIndex === 'number' ? fileNameOrIndex : uploads.findIndex(upload => upload?.fileName === fileNameOrIndex);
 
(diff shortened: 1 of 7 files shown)


libs/media/src/lib/file/pipes/file-list.pipe.ts
```diff
@@ -1,25 +1,28 @@
 import { Pipe, PipeTransform, NgModule } from '@angular/core';
 import { StorageFile } from '@blockframes/model';
 
+export function getFileListIndex(index: number, list: StorageFile[]): number | undefined {
+  if (index === undefined || index === null || !list || !Array.isArray(list)) {
```
GrandSchtroumpf, line 5: you shouldn't be able to provide params that are undefine or not an array, typescript compiler should warn you. Consider improving type of the arguments instead of adding guards for JS

libs/movie/src/lib/movie/form/media-notes/notes.component.ts
```diff
@@ -47,4 +50,10 @@ export class MovieFormMediaNotesComponent implements OnInit, OnDestroy {
   ngOnDestroy() {
     this.sub.unsubscribe();
   }
+
+  removeFromQueue(index: number) {
+    const storagePath = getFileStoragePath('movies', 'notes', this.movieId);
+    const queueIndex = getFileListIndex(index, this.form.promotional.get('notes').value);
+    this.uploaderService.removeFromQueue(storagePath, queueIndex);
+  }
```
GrandSchtroumpf, line 58: Prefer using the value directly instead of mixing `getter`, `get()` & `value`
```suggestion
  removeFromQueue(index: number) {
    const title = this.form.value;
    const storagePath = getFileStoragePath('movies', 'notes', this.movieId);
    const queueIndex = getFileListIndex(index, title.promotional.notes);
    this.uploaderService.removeFromQueue(storagePath, queueIndex);
  }
```

libs/media/src/lib/file-uploader.service.ts
```diff
@@ -64,13 +64,12 @@ export class FileUploaderService {
...
    */
-  removeFromQueue(storagePath: string, fileName: string) {
-
+  removeFromQueue(storagePath: string, fileNameOrIndex: string | number) {
     const uploads = this.queue[storagePath];
 
     if (!uploads || !uploads.length) return;
 
-    const index = uploads.findIndex(upload => upload?.fileName === fileName);
+    const index = typeof fileNameOrIndex === 'number' ? fileNameOrIndex : uploads.findIndex(upload => upload?.fileName === fileNameOrIndex);
```
GrandSchtroumpf, line 72: for readability
```suggestion
    const index = typeof fileNameOrIndex === 'number'
      ? fileNameOrIndex
      : uploads.findIndex(upload => upload?.fileName === fileNameOrIndex);
```

phpgeekfr:
@RemcoSimonides what do we do with this PR ?

RemcoSimonides:
@GrandSchtroumpf what is your opinion? I suppose it will take a while before there is enough time/budget before the media will be reworked so merging this PR can fix it for now... After resolving your change requests of course

GrandSchtroumpf (approved):

Issue #8566, opened 2022-06-02
CRM event export - missing "hosted by" data


Jackseed:
### Localization
- App name: CRM
- Page: event list
- Url: https://crm.blockframes.io/c/o/dashboard/crm/events

### Context
When exporting a CSV file, the "Hosted by" column (ie the org organizing the event) is empty. 
![image](https://user-images.githubusercontent.com/14964443/171596082-c42968f3-d9c0-44ad-9457-9b9246bef3f6.png)

### Fix
This data is really important; best would even be to have both "Organizer Id" (orgId) and "Organizer Name" (org name)

PR #8592, opened 2022-06-08
Fix CRM event export data

KvnMrs:
Issue: https://github.com/blockframes/blockframes/issues/8566


Poppe-A (approved):
LGTM. maybe Remco will have something to add ?

RemcoSimonides:
Nope, also looking good to me :)

Issue #8555, opened 2022-05-30
CRM - Check and Fix sorting and filtering on tables


KevinDurand974:
Already started to check what to do.

**KevinD** branch `kevind/crm-sorting-filtering-checks`

Some shortcuts
Filters: `libs\ui\src\lib\list\table\filters.ts`
Sorts: `libs\ui\src\lib\list\table\sorts.ts`

- [x] Profile - Users
- [x] Profile - Invitation Users
- [x] Organization - Organization
- [x] Organization - Company Members
- [x] Organization - Movie on the Org Page
- [x] Movie - Movie List
- [x] Event - Event List
- [x] Event - Invitation - Inside event
- [x] Event - Invitation List

PR #8597, opened 2022-06-09
CRM - Check and Fix sorting and filtering on tables

KevinDurand974:
From issue: https://github.com/blockframes/blockframes/issues/8555


libs/ui/src/lib/list/table/sorts.ts
```diff
@@ -0,0 +1,46 @@
...
+    return a ? 1 : -1;
+  }
+}
+
+export const sorts = {
+  byOrgName: (a: Organization, b: Organization) => {
+    const aName = a?.denomination.public || '--';
+    const bName = b?.denomination.public || '--';
+    return sorts.defaultSort(aName, bName);
+  },
```
RemcoSimonides, line 17: Soon to be removed again with #8355. You could add `// TODO #8355` so less chance to forget

libs/ui/src/lib/list/table/sorts.ts
```diff
@@ -0,0 +1,46 @@
...
+  byDate: (a: string, b: string) => {
+    const aDate = new Date(a).getTime() || 0;
+    const bDate = new Date(b).getTime() || 0;
+    return sorts.defaultSort(aDate, bDate);
+  },
```
RemcoSimonides, line 22: byDate could be part of the defaultSort if you check if type is a date using `a instanceof Date`

libs/ui/src/lib/list/table/sorts.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+    if (!a && b) {
+      const type = typeof b;
+      if (type === 'string') return defaultSorting('--', b, type);
+      if (type === 'number') return defaultSorting(0, b, type);
+      if (type === 'boolean') return defaultSorting(false, b, type);
+      return 0;
+    }
+    return sorts.defaultSort(a, b);
+  }
```
RemcoSimonides, line 45: How come this only happens for `release.year`? I see it has undefined values, but that probably happens in more tables, no?

What behavior would you expect to happen when filtering on release.year? Shouldn't the undefined values always be at the end? No matter when sorting ascending or descending. e.g.

**Ascending:**
2022
2021
2020
undefined
undefined
undefined

**Descending:**
2020
2021
2022
undefined
undefined
undefined

KevinDurand974: Its only on release year because its the only one on CRM who much of the values are undefined with a sort (I think).
For the behavior, I didnt think of that but its better the way you say.

apps/crm/crm/src/app/crm/pages/organizations/organizations.component.html
```diff
@@ -20,14 +20,14 @@
...
     <ng-template colRef="status" let-status sort>
       <bf-tag>{{ status }}</bf-tag>
     </ng-template>
     <ng-template colRef="denomination.full" label="Organization Name" sort let-fullName>
       {{ fullName }}
     </ng-template>
-    <ng-template colRef="addresses.main.country" defaultSort label="Country" sort let-country>
+    <ng-template colRef="addresses.main.country" label="Country" let-country>
       {{ country ?? '--' | titlecase }}
     </ng-template>
```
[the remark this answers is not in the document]
KevinDurand974: I've to do this, for this to work `{{ (country | toLabel:'territories') || '--' }}`

libs/ui/src/lib/list/table/filters.ts
```diff
@@ -28,6 +35,26 @@ export const filters = {
     if (!movie?.title?.international) return false;
     return movie.title.international.toLocaleLowerCase().includes(input);
   },
+  crmMovieOrgName: (input: string, _, movie: CrmMovie) => {
+    if (!movie?.org?.denomination?.public) return false;
+    return movie.org.denomination.public.toLocaleLowerCase().includes(input);
+  },
```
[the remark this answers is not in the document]
KevinDurand974: Actually true but I think it will take some time before the changement will be done, I can write a todo for this to not forget?

libs/ui/src/lib/list/table/sorts.ts
RemcoSimonides: You can always ask Ana or Camille to verify my proposition. Let's see if they agree ^^
And I think the defaultSort should then be able to handle this case too. No need to have a separate defaultSortByOneValue function for it.

apps/crm/crm/src/app/crm/pages/user/user.component.html
```diff
@@ -131,21 +131,21 @@ <h3>Buying Preferences</h3>
...
+    {{ title }}
   </ng-template>
   <ng-template colRef="mode" label="Mode" let-mode sort>
     {{mode}}
   </ng-template>
   <ng-template colRef="type" label="Type" let-type sort>
     {{type}}
   </ng-template>
-  <ng-template colRef="fromOrg.denomination.full" label="From Organization" let-full sort>
+  <ng-template colRef="fromOrg.denomination.full" label="From Organization" let-full sort [filter]="filters.invitationOrgName">
```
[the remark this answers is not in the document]
KevinDurand974: Right, why I did this ? 😑

libs/ui/src/lib/list/table/sorts.ts
KevinDurand974: After reflection, it's very complicated to set up this, If I want do to this, I've to get the full array and take all defined values and sort these, then, add the undefined one.
It's way more complicated I think, Ana say to go on the simple one if that to much complicated.

RemcoSimonides: It's not as simple as
```
if (!a) return 1;
if (!b) return -1;
```
?

Btw, I do recommend to ask Francois to review this once you update it. I'll be gone and he is a master at these things

KevinDurand974: And no it's not working,
I'll ask François to review after I update it

libs/ui/src/lib/list/table/filters.ts
```diff
@@ -1,12 +1,19 @@
...
   Movie,
+  Organization,
   Person,
   Scope,
   staticModel
-} from "@blockframes/model";
+} from '@blockframes/model';
 
-function getStaticModelFilter(scope: Scope) {
+interface CrmMovie extends Movie {
```
phpgeekfr, line 11: already defined in `apps/crm/crm/src/app/crm/pages/movies/movies.component.ts` move it to libs/model/movie.ts

GrandSchtroumpf (approved):

PR #8598, opened 2022-06-09
Nicolas/e2e signup rewrite

n-germain-work:
- Added a plugin for Algolia, so that we can inject an org in the db, and offer the option to join it during signup
- A new user can signup asking to join a marketplace org functional


apps/festival/festival-e2e/src/fixtures/authentification/signup.ts
```diff
@@ -95,7 +103,7 @@ const e2ePermissions = (data: { id: string; adminUid: string }): PermissionsDocu
   return {
     id,
     roles: {
-      [adminUid]: 'superAdmin',
+      [`${adminUid}`]: 'superAdmin',
```
GrandSchtroumpf, line 106: why do you need to do that ? adminUid can be undefined/null ?
This is not normal to do that, usually it's because types has not been well handle before.

libs/testing/cypress/browser/src/lib/browserAuth.ts
```diff
@@ -19,4 +19,12 @@ export const browserAuth = {
       indexedDB.deleteDatabase('firebaseLocalStorageDb');
     });
   },
+
+  getAuth() {
+    cy.window().should('have.property', 'LoginService');
+    return cy.window().then(w => {
```
GrandSchtroumpf, line 25:
```suggestion
    return cy.window().then(w => w['LoginService'] as AuthService)
```
But I don't like the idea of user the service for that. I would prefer having something that is not related to Angular.

apps/festival/festival-e2e/src/integration/authentification/signup.spec.ts
```diff
@@ -1,49 +1,42 @@
...
-    });
-}
-
-//? to be used when signing up with known organisation
 const marketplaceInjectedData = {
   [`users/${marketplaceData.orgAdmin.uid}`]: marketplaceData.orgAdmin,
   [`orgs/${marketplaceData.org.id}`]: marketplaceData.org,
   [`permissions/${marketplaceData.permissions.id}`]: marketplaceData.permissions,
 };
+//? to be used when signing up with a dashboard organisation
```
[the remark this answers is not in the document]
n-germain-work: It's because it put colors on my code ^^
This one is a soft blue, not to be too much flashy as a real 'TODO' (which prints in orange).
I'm erasing the '?' to be clearer in the review 👍

GrandSchtroumpf (approved):

PR #8623, opened 2022-06-13
move all email utils

george43g:
email related interfaces and functions moved to `model`


libs/model/src/lib/emails.ts
```diff
@@ -1,2 +1,302 @@
+import { Bucket, MailBucket } from "./bucket";
+import { differenceInDays, differenceInHours, differenceInMinutes, format, millisecondsInHour } from 'date-fns';
+import { AttachmentData } from '@sendgrid/helpers/classes/attachment';
+import { EmailJSON } from '@sendgrid/helpers/classes/email-address';
```
phpgeekfr, line 4: do we want model library to import data from an external lib ?

libs/model/src/lib/emails.ts
```diff
@@ -1,2 +1,94 @@
...
+    password,
+    isRegistered: !!user.orgId
+  }
+}
+
+export function getOfferEmailData(offer: Partial<Offer>): OfferEmailData {
+  return {
+    id: offer.id
+  }
+}
```
phpgeekfr, line 94: why not moving this then ?

```
export function getOrgEmailData(org: Partial<Organization>): OrgEmailData {
  return {
    id: org.id,
    denomination: orgName(org, 'full'),
    email: org.email || '',
    country: toLabel(org.addresses?.main?.country, 'territories')
  }
}


export function getMovieEmailData(movie: Partial<Movie>): MovieEmailData {
  return {
    id: movie.id,
    title: {
      international: movie.title.international,
    }
  }
}
```

george43g: these are impossible to move without also moving other interfaces that then import things we cant import

libs/model/src/lib/emails.ts
```diff
@@ -1,2 +1,94 @@
+import { MailTerm } from "./terms";
+import { User } from "./user";
+import { Offer } from "./offer";
 
 export type EmailErrorCodes = 'E01-unauthorized' | 'E02-general-error' | 'E03-missing-api-key' | 'E04-no-template-available';
```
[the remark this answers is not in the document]
george43g: can you clarify the intention: export type EmailErrorCode = typeof emailErrorCode[number]; - this does not work

libs/model/src/lib/emails.ts
phpgeekfr: yes they can. Not relying to anything we can't import

phpgeekfr (approved):

Issue #8606, opened 2022-06-10
Release 4.3 - Organization


RemcoSimonides:
[Organisation test scenarios](https://www.notion.so/cascade8/Organisation-55ded6f0e09c4aa29fb2e3f716bb8b50)

Issues that have been resolved in this release:
- [x] https://github.com/blockframes/blockframes/issues/6729
- [x] https://github.com/blockframes/blockframes/issues/7964
- [x] https://github.com/blockframes/blockframes/issues/8435
- [x] https://github.com/blockframes/blockframes/issues/8441

cams-goussale:
- [ ] The title of the Buyer List should be `Buyer` (instead of `Buyer List`)

- [ ] There shouldn't be the movie option on the Buyer organization card. The component is slightly different from the Seller's list. [See Mockups here](https://www.figma.com/file/ldSVAauJy2tYiXUXF4sC1i/Buyers)
![image](https://user-images.githubusercontent.com/93206363/173359281-ad109217-5468-4252-af6c-15418c1c8eae.png)

dmoncadalv:
**Update org profile information**

- [ ]  Upload process can’t be paused or stopped 
![5_ORGANISATION_UploadProcess.gif](https://images.zenhubusercontent.com/6258184d440b61b47c763ef1/34f630e3-f963-4e61-8278-abafbd3a5111)

phpgeekfr:
> Upload process can’t be paused or stopped

Yes it can ben paused or canceled. Your image was uploaded to fast to see it.
Anyway, file upload will be reworked in #8586

PR #8639, opened 2022-06-14
Fix - Organization

KvnMrs:
Issue: https://github.com/blockframes/blockframes/issues/8606

### DIFF
--- a/libs/organization/src/lib/organization/components/card/card.component.html
+++ b/libs/organization/src/lib/organization/components/card/card.component.html
@@ -21,7 +21,7 @@
     </div>
   </mat-tab>
   <mat-tab>
-    <ng-template mat-tab-label>
+    <ng-template *ngIf="showMovieIcon" mat-tab-label>
       <mat-icon svgIcon="movie"></mat-icon>
     </ng-template>
     <!-- We want to use slice to be sure to have 3 elements/placeholders -->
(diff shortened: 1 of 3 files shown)


libs/organization/src/lib/organization/components/card/card.component.html
```diff
@@ -5,7 +5,22 @@
...
+<mat-tab-group *ngIf="!hideIcons" headerPosition="below" mat-align-tabs="center" [disablePagination]="true">
+  <div class="tab-content">
+    <p class="mat-caption">{{ (org.addresses.main.country | toLabel: 'territories') || "-" }}</p>
+    <p class="mat-caption" i18n *ngIf="(orgMovieCount$ | async) as count">
+      {{ count }} film{ count, plural, =1 {} other {s} } in {{ app === 'catalog' ? 'catalog' : 'line-up' }}
+    </p>
+    <p class="mat-caption" i18n>
+      {{ org.userIds.length }} employee{ org.userIds.length, plural, =1 {} other {s} }
+    </p>
+  </div>
```
phpgeekfr, line 23: please rename the css class `tab-content` since it is now not always used inside a mat tab group.

In fact 
```
div {
  padding: 16px 0;
}
```
should be enough

phpgeekfr (approved):

Issue #8609, opened 2022-06-10
Release 4.3 - Import


RemcoSimonides:
[Import test scenarios](https://www.notion.so/cascade8/Import-78250f00339a4aa9a2983c03d85a0b6f)

Issues resolved for this release:
- [x] https://github.com/blockframes/blockframes/issues/7429
- [x] https://github.com/blockframes/blockframes/issues/8561

cams-goussale:
- [ ] When user downloads the movie import template, it downloads an excel with test data in it. Careful with the folder in the drive:
- [here are the clean templates](https://drive.google.com/drive/folders/1hmsmslJxLxrPpA5Cy5K814Rgd_bHYaQb) to upload on the app
- [here are the test templates](https://drive.google.com/drive/folders/1S_fi8-jmVcRSGAurrI4zCg49KCsnTnDo) for the release week, when we test on staging

phpgeekfr:
- [ ] https://github.com/blockframes/blockframes/issues/8612#issuecomment-1156137430

cams-goussale:
>  (Priority A) When importing empty cells in columns E and F, the code imports 0 value. This blocks the submission of the movie, because it is detected as an error by the movie form. When E and F are empty, it should be detected as null values.

- [ ] I still have this bug

PR #8654, opened 2022-06-16
Import release 4 3

yspringfield:
Fixed: #8609


libs/import/src/lib/view-extracted-elements/titles/fieldConfigs.ts
```diff
@@ -155,12 +155,14 @@ export function getFieldConfigs(
     },
       /* e */ 'title.series': (value: string, data: Partial<FieldsConfig>) => {
       if (!value && data.contentType === 'tv') throw optionalWarning('Season Number');
+      if (value === '') return null;
       const series = Number(value);
```
yspringfield, line 159: `const series = Number('') // series = 0`
which is wrong as empty string '' denotes no value and not `0`

libs/import/src/lib/view-extracted-elements/titles/fieldConfigs.ts
```diff
@@ -155,12 +155,14 @@ export function getFieldConfigs(
     },
       /* e */ 'title.series': (value: string, data: Partial<FieldsConfig>) => {
       if (!value && data.contentType === 'tv') throw optionalWarning('Season Number');
+      if (value === '') return null;
```
GrandSchtroumpf, line 158: you can do : 
```suggestion
      if (!value) return null;
```

libs/import/src/lib/view-extracted-elements/titles/fieldConfigs.ts
```diff
@@ -155,12 +155,14 @@ export function getFieldConfigs(
...
       const series = Number(value);
       if (isNaN(series)) throw wrongValueError(value, 'Season Number');
       return series;
     },
       /* f */ 'runningTime.episodeCount': (value: string, data: Partial<FieldsConfig>) => {
       if (!value && data.contentType === 'tv') throw optionalWarning('Number of Episodes');
+      if (value === '') return null;
```
GrandSchtroumpf, line 165:
```suggestion
      if (!value) return null;
```

GrandSchtroumpf (approved):

PR #8661, opened 2022-06-16
warnmissingvars to check prefixed shell env vars

george43g:
- refacror cleanup getKeyName
- consistent use of function for deployment of secrets and detection


libs/firebase-utils/src/lib/util.ts
```diff
@@ -6,6 +6,7 @@ import requiredVars from 'tools/mandatory-env-vars.json';
 import { resolve } from 'path';
 import { firebase as firebaseProd } from 'env/env.blockframes';
 import { OrganizationDocument, App } from '@blockframes/model';
+import { getKeyName } from '@blockframes/devops'
```
phpgeekfr, line 9: move this to firebase-utils lib and try to deploy your functions

phpgeekfr (approved):

Issue #8612, opened 2022-06-10
Release 4.3 - General UI


RemcoSimonides:
[General UI test scenarios](https://www.notion.so/cascade8/General-UI-db403143e36c480d9c6b562cc001119f)

Other issues that have been resolved for this release:

Email:
- [x] https://github.com/blockframes/blockframes/issues/7897
- [x] https://github.com/blockframes/blockframes/issues/8245 (bf-carousel has basically been rebuild so please check if they're still working everywhere)
- [x] https://github.com/blockframes/blockframes/issues/8437
- [x] https://github.com/blockframes/blockframes/issues/8555
- [x] https://github.com/blockframes/blockframes/issues/7724
- [x] https://github.com/blockframes/blockframes/issues/8520

2amarkina:
- [ ] Videos on the Film Page are not displayed correctly 
film name: Io Sto Bene (LUXEMBOURG) 
<img width="1417" alt="Снимок экрана 2022-06-14 в 15 59 39" src="https://user-images.githubusercontent.com/55499036/173596128-673d2757-5120-4011-b487-319b54e51ddf.png">

2amarkina:
- [ ] Infinite spinner on the Calendar on the Marketplace (Archipel Market)
<img width="722" alt="Снимок экрана 2022-06-14 в 16 07 00" src="https://user-images.githubusercontent.com/55499036/173597435-a632c5f9-3d90-4610-97c6-2ae965593947.png">

dmoncadalv:
![Film-Page-Screening-Schedule-Archipel-Market.png](https://images.zenhubusercontent.com/6258184d440b61b47c763ef1/2b39076a-2c87-48c8-b48d-4aec8fdab887)
@2amarkina  I feel there are too many « ongoing screening » buttons in one page. Maybe it would be better to delete the hero in this case, we would still have 2 ongoing screening buttons, 1 access screening button (= 3 different ways of accessing the screening) and 2 « Request asking price » buttons.
I would even take off the « ongoing screening » button in the « upcoming screenings » section and leave the « access event » button only.

dmoncadalv:
![Positioning.png](https://images.zenhubusercontent.com/6258184d440b61b47c763ef1/449a9f84-f149-4568-92b8-71f8b4110d42)

Why is there « blank » grey shapes under target audience and social responsability goals if there’s no information ?

dmoncadalv:
**Event page**

[...]

PR #8679, opened 2022-06-17
Release 4 3 - General ui

KvnMrs:
issue: https://github.com/blockframes/blockframes/issues/8612

### DIFF
--- a/libs/movie/src/lib/movie/components/header/header.component.scss
+++ b/libs/movie/src/lib/movie/components/header/header.component.scss
@@ -6,7 +6,7 @@ header {
   background-size: cover;
 
   .overlay {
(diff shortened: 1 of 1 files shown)


libs/movie/src/lib/movie/components/header/header.component.scss
```diff
@@ -6,15 +6,15 @@ header {
   background-size: cover;
 
   .overlay {
-    padding: 24px 24px 48px 24px;
+    padding: 48px 24px 48px 24px;
```
phpgeekfr, line 9: I think you can do then
```suggestion
    padding: 48px 24px;
```

phpgeekfr (approved):

PR #8694, opened 2022-06-20
Merge back release 4 3 develop

phpgeekfr:


libs/contract/src/lib/bucket/service.ts
```diff
@@ -47,6 +47,7 @@ export class BucketService extends BlockframesCollection<Bucket> {
   }
 
   protected fromFirestore(document: DocumentSnapshot<Bucket>): Bucket {
+    if(!document.exists) return ;
```
phpgeekfr, line 50: I have doubts on this. 
I don't know if this should be here, and it should be `if(!document.exists()) return ;` with parenthesis

yspringfield: on line 51, we call `createBucket` even if the bucket doesn't exist in the db and `createBucket` returns an object ( with default values for non existent buckets ). 
What this effectively means is that, in places like `libs/contract/src/lib/bucket/service.ts` line 123, the verification if the bucket exists will fail even for non existent buckets.

GrandSchtroumpf (approved):

libs/event/src/lib/service.ts
```diff
@@ -58,13 +58,24 @@ export class EventService extends BlockframesCollection<Event> {
...
     return this.valueChanges<Event<Meta>>(id).pipe(
       joinWith({
         org: ({ ownerOrgId }: Event<Meta>) => this.orgService.valueChanges(ownerOrgId),
         movie: (event: Event<Meta>) => {
           if (isScreening(event)) {
-            return event.meta.titleId ? this.movieService.valueChanges(event.meta.titleId) : undefined;
+            return event.meta.titleId ? this.movieService
+              .valueChanges(this.movieQuery(event.meta.titleId))
+              .pipe(map(movies => movies?.length ? movies[0] : undefined))
+              : undefined;
```
phpgeekfr, line 78: if (!event.meta.titleId) return;
return this.movieService.valueChange(event.meta.titleId).pipe(
  catchError(() => of(undefined))
)

apps/festival/festival/src/app/dashboard/home/home.component.html
```diff
@@ -72,17 +72,19 @@ <h5>Film Page Views</h5>
...
             [topCount]="5"
             horizontal
             showLegend
             selectable
           >
-            <mat-icon svgIcon="buyer_insights"></mat-icon>
-            <h3>Buyers' Insights</h3>
+            <mat-icon class="title_margin" svgIcon="buyer_insights"></mat-icon>
+            <h3 class="title_margin">Buyers' Insights</h3>
+            <!-- #8696 h4 should be "TOP 5 COUNTRIES" line 76 above -->
```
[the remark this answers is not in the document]
phpgeekfr: lack of time, changes where made this morning and the fastest solution was choosen

libs/analytics/src/lib/components/metric-card-list/metric-card-list.component.html
```diff
@@ -1,3 +1,8 @@
+<div *ngIf="selectable">
```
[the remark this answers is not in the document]
phpgeekfr: empty space is needed even if button is not displayed to prevent all content after to go down when button is displayed

Issue #8644, opened 2022-06-15
Sellers' analytics - Remove sellers' page view & interactions


Jackseed:
### Context 
Competitors or org members watching movie pages and promo elements from other sellers shouldn't be counted as views

### To do
- [x] Remove analytics created by org members if their org has access to festival dashboard

yspringfield:
Hi @Jackseed , as indicated [here](https://github.com/blockframes/blockframes/pull/8718#issue-1282219903), we simply ignore other sellers' data when showing the statistics to the seller.
Does this issue also involve avoiding saving analytics data when the initiator of such events are other sellers?
Or should we save the event data irrespective of who creates it and simply ignore those when showing the stats to other sellers ( which is what is currently implemented )?

Jackseed:
Ignore it is the good way to go, thanks 👍

PR #8718, opened 2022-06-23
feat: ignore competitor analytics when showing title/buyer analytics

yspringfield:
Fixed: #8644
Ignore seller analytics on following pages:
1. Festival Dashboard `c/o/dashboard/home`
2. title list `/c/o/dashboard/home/title`
3. title analytics `/c/o/dashboard/home/title/{titleId}`
4. buyer list `/c/o/dashboard/home/buyer`

### DIFF
--- a/apps/festival/festival/src/app/dashboard/analytics/buyers/buyers-analytics.component.ts
+++ b/apps/festival/festival/src/app/dashboard/analytics/buyers/buyers-analytics.component.ts
@@ -36,6 +36,14 @@ export class BuyersAnalyticsComponent {
       users: ({ uids }) => this.userService.valueChanges(uids),
       orgs: ({ orgIds }) => this.orgService.valueChanges(orgIds)
     }, { shouldAwait: true }),
+    map(({ orgs, analytics, users, ...rest }) => {
+      const buyerOrg = orgs.filter(org => !org.appAccess.festival.dashboard);
+      const buyerOrgIds = buyerOrg.map(({ id }) => id);
+      const buyerAnalytics = analytics.filter(({ meta }) => buyerOrgIds.includes(meta.orgId))
+      const buyerUsers = buyerAnalytics.map(({ meta }) => meta.uid);
+      const filteredUsers = users.filter(({ uid }) => buyerUsers.includes(uid));
+      return { ...rest, users: filteredUsers, orgs: buyerOrg, analytics: buyerAnalytics };
+    }),
     shareReplay({ bufferSize: 1, refCount: true })
   );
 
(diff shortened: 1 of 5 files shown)


apps/festival/festival/src/app/dashboard/analytics/buyers/buyers-analytics.component.ts
```diff
@@ -36,6 +36,14 @@ export class BuyersAnalyticsComponent {
...
       orgs: ({ orgIds }) => this.orgService.valueChanges(orgIds)
     }, { shouldAwait: true }),
+    map(({ orgs, analytics, users, ...rest }) => {
+      const buyerOrg = orgs.filter(org => !org.appAccess.festival.dashboard);
+      const buyerOrgIds = buyerOrg.map(({ id }) => id);
+      const buyerAnalytics = analytics.filter(({ meta }) => buyerOrgIds.includes(meta.orgId))
+      const buyerUsers = buyerAnalytics.map(({ meta }) => meta.uid);
+      const filteredUsers = users.filter(({ uid }) => buyerUsers.includes(uid));
+      return { ...rest, users: filteredUsers, orgs: buyerOrg, analytics: buyerAnalytics };
+    }),
```
GrandSchtroumpf, line 46: Best would be to extract this in an external function with a good naming  : 
```typescript
function goodNamingFunction(orgs: Organisation, analytics, users, title) {
      const buyerOrg = orgs.filter(org => !org.appAccess.festival.dashboard);
      const buyerOrgIds = buyerOrg.map(({ id }) => id);
      const buyerAnalytics = analytics.filter(({ meta }) => buyerOrgIds.includes(meta.orgId))
      const buyerUsers = buyerAnalytics.map(({ meta }) => meta.uid);
      const filteredUsers = users.filter(({ uid }) => buyerUsers.includes(uid));
      return { ...rest, users: filteredUsers, orgs: buyerOrg, analytics: buyerAnalytics };
}
```

apps/festival/festival/src/app/dashboard/analytics/titles/titles-analytics.component.ts
```diff
@@ -65,7 +65,18 @@ export class TitlesAnalyticsComponent {
     @Inject(APP) public app: App
   ) { }
 
+  getTitleAnalytics(titleId:string){
+    return this.analytics.getTitleAnalytics({ titleId }).pipe(
+      joinWith({
+        org: ({meta}) => this.orgService.valueChanges(meta.orgId)
+      }, {shouldAwait:true}),
+      map(analyticsWithOrg => {
```
GrandSchtroumpf, line 73: If you're doing that several time you might want to create a function to factorize the logic. But keep the rule of thumb in mind: "2 times you copy, 3 times you factorize"

GrandSchtroumpf (approved):

RemcoSimonides:
If I read the changes, it looks like you forgot to exclude sellers on the Buyer Activity Page.
Also, I think it's best to change the `getTitleAnalytics` function instead so the sellers are excluded already when using that function, just like "owners of title" are excluded there too.

It would have been nice to know about this feature request earlier though it might have impacted the data model. What do we do with orgs that have been a buyer, and are now also seller? At first their data would have been shown, but once they also become a seller, their data suddenly disappears.

yspringfield:
@cams-goussale

cams-goussale:
> What do we do with orgs that have been a buyer, and are now also seller?

Yes, sorry the decision changed on this topic (the business team also changed in the meantime). It's still usefull that we get data from all users (seller and buyer) but when displayed on the app, it gets "filtered" by user profile.

Regarding the question: if it is possible, yes, let's just stop showing the data if the user becomes a Seller. I expect this to be edge cases.

Issue #7555, opened 2022-01-17
Screening stat improvements


Jackseed:
Pulsar team thought there was an error with the screening statistics being updated from one connexion to another (first see a user at 10min viewing; then 45min). It looks like we should:

- [x] add something to explain how the stats are created (almost instantly updated)
- [x] add a warning or something to underline that screening is still happening and data not complete
- [x]  **bruce** add a "Watching now" status on users (to prevent sellers but also to create fun) => moved to #8717

- [x] There also was uncertainty on the "0min": does it mean that user didn't click on play? left right away?

_Originally posted by @Jackseed in https://github.com/blockframes/blockframes/issues/6680#issuecomment-1014660277_

2amarkina:
[Work in progress ici](https://www.figma.com/file/HQ2FBvm2u0vyWO0QpIKeTs/%5B-Event-Form-Structure-%5D?node-id=710%3A4957)

2amarkina:
- [ ] add participation rate 
- [ ] check together with @cams-goussale the new improvements that can be added to this issue

cams-goussale:
Listing here all the metrics that we will include in the excel export (in reference to this issue #7135). So it'd make sense to also include them in the Statistic section, inside the app. We don't need to add everything, we can be selective.

(Vanity)
- Number of invitations sent
- Number of invitations with status accepted / unanswered
- Number of Requests to join the event
- Number of attendees

(Analytics)
- Attendance rate `=sum(all.events.attendees)/sum(all.accepted)`
- Average watchtime `=sum(all.invitations.watchTime)/sum(all.events.attendees)`
- Acceptation rate `=sum(all.accepted)/sum(all.invitations)`
- Traction rate `=sum(all.invitation.mode == "request")/sum(all.invitations)` | Traction generated by the app itself

2amarkina:
![Screening.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/585e6801-e9c5-4269-86c5-0f3d2ec4abca)


![Screening (1).png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/9746a106-d6a2-4ead-8cf0-73d97a913e93)

stephanie-cascade8:
[WIP Mockup here](https://www.figma.com/file/xigwniq5ZYfpD80gzQWYvO/Event-Creation?node-id=119%3A6897)

Jackseed:
[...]

PR #8749, opened 2022-07-04
#7555 Screening stat improvements

phpgeekfr:


libs/event/src/lib/pages/analytics/analytics.component.html
```diff
@@ -1,49 +1,41 @@
-<ng-container *ngIf="isEventStarted(event$ | async)">
-  <h2>Statistics</h2>
+<ng-container *ngIf="isEventStarted(event$ | async) as event">
```
RemcoSimonides, line 1: I think you should user the `eventTime` pipe instead
libs/event/src/lib/pipes/event-time.pipe.ts

libs/event/src/lib/pages/analytics/analytics.component.html
```diff
@@ -1,49 +1,41 @@
...
+  <header>
+    <h2>Statistics</h2>
+    <mat-icon svgIcon="info" matTooltip="Statistics are updated in real time."></mat-icon>
+    <ng-container *ngIf="ongoingScreening(event)">
```
RemcoSimonides, line 5: I think you should user the `eventTime` pipe instead
libs/event/src/lib/pipes/event-time.pipe.ts

libs/event/src/lib/pages/analytics/analytics.component.html
```diff
@@ -1,49 +1,41 @@
...
+      <a ongoing-event-button routerLink="/event/{{event.id}}/r/i">
```
RemcoSimonides, line 6: I think you should use the `eventLink` pipe instead
libs/event/src/lib/pipes/event-link.pipe.ts

libs/event/src/lib/pages/analytics/analytics.component.html
```diff
@@ -1,49 +1,41 @@
...
-            <ng-template colRef="email" label="Email Address" let-email sort>{{ email }}</ng-template>
-            <ng-template colRef="orgName" label="Company Name" let-orgName sort>{{ orgName || dataMissing }}</ng-template>
-            <ng-template colRef="orgActivity" label="Company Activity" let-orgActivity>{{ orgActivity ? (orgActivity | toLabel: 'orgActivity') : dataMissing }}</ng-template>
-            <ng-template colRef="orgCountry" label="Country" let-orgCountry>{{ orgCountry ? (orgCountry | toLabel: 'territories') : dataMissing }}</ng-template>
-            <ng-template colRef="watchInfos.duration" label="Watch Time" let-duration sort>{{ duration * 1000 | duration }}</ng-template>
-            <ng-template colRef="watchInfos.date" label="Watching Status" let-date sort>
-              <bf-tag *ngIf="date | isWatchingNow">Watching Now</bf-tag>
+        <p>"0" means the user left the screening right away.</p>
+        <analytics-metric-card-list [cards]="aggregatedScreeningCards"></analytics-metric-card-list>
+        <article fxLayout="row" fxLayoutAlign="space-between start" flex>
```
RemcoSimonides, line 16: Use css instead of fxLayout
Also, the `flex` directive doesn't exist anymore so it's probably not doing anything here

libs/event/src/lib/pages/analytics/analytics.component.html
```diff
@@ -1,49 +1,41 @@
...
             </ng-template>
-          </bf-table>
-        </mat-card>
+          </button>
+        </article>
+        <bf-table pagination="50" useFilter [source]="acceptedAnalytics">
+          <ng-template colRef="name" label="Name" let-name sort>{{ name | titlecase }}</ng-template>
+          <ng-template colRef="email" label="Email Address" let-email sort>{{ email }}</ng-template>
+          <ng-template colRef="orgName" label="Company Name" let-orgName sort>{{ orgName || dataMissing }}</ng-template>
+          <ng-template colRef="orgActivity" label="Company Activity" let-orgActivity>{{ orgActivity ? (orgActivity | toLabel: 'orgActivity') : dataMissing }}</ng-template>
```
RemcoSimonides, line 32:
```suggestion
          <ng-template colRef="orgActivity" label="Company Activity" let-orgActivity [filter]="filters.orgActivity">{{ orgActivity ? (orgActivity | toLabel: 'orgActivity') : dataMissing }}</ng-template>
```
If the displayed values are different than the one in the source data, then sometimes the filters and sorting doesn't working as expected. So we have to add a special filter function for this

You can check for example in this bf-table how it's implemented apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html

libs/event/src/lib/pages/analytics/analytics.component.html
```diff
@@ -1,49 +1,41 @@
...
+          <ng-template colRef="orgCountry" label="Country" let-orgCountry>{{ orgCountry ? (orgCountry | toLabel: 'territories') : dataMissing }}</ng-template>
```
RemcoSimonides, line 33:
```suggestion
          <ng-template colRef="orgCountry" label="Country" let-orgCountry [filter]="filters.territories">{{ orgCountry ? (orgCountry | toLabel: 'territories') : dataMissing }}</ng-template>
```

libs/event/src/lib/pages/analytics/analytics.component.html
```diff
@@ -1,49 +1,41 @@
...
+          <ng-template colRef="watchInfos.duration" label="Watch Time" let-duration sort>{{ duration * 1000 | duration }}</ng-template>
+          <ng-template colRef="watchInfos.date" label="Watching Status" let-date sort>
```
RemcoSimonides, line 35: Does the sorting work ass expected on this column?

libs/analytics/src/lib/utils.ts
```diff
@@ -64,3 +79,98 @@ export function aggregatePerUser(analytics: (Analytics<"title"> & { user: User,
...
 }
+
+export interface MetricCard {
+  title: string;
+  value: number | string;
+  icon: string;
+  selected?: boolean;
+}
+
+export function toScreenerCards(screeningRequests: Invitation[] | Analytics<'title'>[], invitations: Invitation[] | InvitationWithAnalytics[]): MetricCard[] {
```
RemcoSimonides, line 90: `Invitation[]` and `Analytics<'title'>[]` are totally different from each other.
Screening Requests is an analytics event triggered when a user is really requesting a screening - like a request to the seller to create an event and invite them. It's not a user asking for access to participate in one.

libs/event/src/lib/pages/analytics/analytics.component.ts
```diff
@@ -111,13 +113,12 @@ export class AnalyticsComponent implements OnInit {
             status: i.status
           };
         });
+
+        this.aggregatedScreeningCards = toScreenerCards(this.eventInvitations.filter(i => i.mode === 'request'), this.eventInvitations);
```
RemcoSimonides, line 117: See comment at `toScreenerCards` function

RemcoSimonides (changes requested):

libs/event/src/lib/pages/analytics/analytics.component.html
phpgeekfr: This pipe is not really good to use here because I want the link to redirect to `event/eventId` but if I use this pipe, it will redirect on `c/o/dashboard/eventId` and I'm already on that page.

libs/event/src/lib/pages/analytics/analytics.component.html
phpgeekfr: This pipe is not really good to use here because I want the link to redirect to `event/eventId` but if I use this pipe, it will redirect on `c/o/dashboard/eventId` and I'm already on that page.

libs/event/src/lib/pages/analytics/analytics.component.html
phpgeekfr: yest

apps/festival/festival/src/app/dashboard/analytics/buyer/buyer-analytics.component.html
```diff
@@ -85,7 +85,10 @@ <h3>Screenings Buyer has participated</h3>
 
     <article>
       <header>
-        <h4>Attended Screenings</h4>
+        <div>
```
RemcoSimonides, line 88: Let's use article instead of `div`. Francois like to avoid div as much as possible

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -134,6 +133,9 @@ <h5>Attendees</h5>
           </span>
           <ng-template #noCountry>-</ng-template>
         </ng-template>
+        <ng-template colRef="status" label="Invitation" let-status sort defaultSort >
```
RemcoSimonides, line 136: Each bf-table should have max 1 column with `defaultSort`. I see all the other columns also have it so please only add it to the one which needs to be sorted by default

apps/festival/festival/src/app/event/session/session.component.ts
```diff
@@ -252,8 +252,16 @@ export class SessionComponent implements OnInit, OnDestroy {
...
 
+        let initializedWatchInfos = false;
         this.watchTimeInterval = interval(1000).pipe(
           filter(() => !!this.isPlaying),
+          tap(() => {
+            // Initialize once to zero as soon as user clicks on play
+            if (!initializedWatchInfos) {
+              this.invitationService.update(invitation.id, { watchInfos: { duration: invitation.watchInfos?.duration ?? 0, date: new Date() } });
+              initializedWatchInfos = true;
+            }
```
RemcoSimonides, line 263: What if I play and pause? Then the timer is reset to 0? It should only be set to 0 when watchInfos is empty

libs/event/src/lib/pages/analytics/analytics.component.ts
```diff
@@ -100,23 +105,29 @@ export class AnalyticsComponent implements OnInit {
...
             status: i.status
           };
         });
-        // Create same analytics but only with 'accepted' status and with a Watchtime > 0
-        this.acceptedAnalytics = this.analytics.filter(({ status, watchInfos }) => status === 'accepted' && watchInfos.duration !== 0);
 
-        // if event is a screening or a slate presentation we add the watch time column to the table
-        // and we compute the average watch time
-        this.avgWatchDuration = averageWatchDuration(this.acceptedAnalytics);
+        const titleIds = event.type === 'screening' ? [(this.event.meta as Screening).titleId] : (this.event.meta as Slate).titleIds;
```
RemcoSimonides, line 117:
```suggestion
        const titleIds = isScreening(event) 
          ? [event.meta.titleId]
          : isSlate(event)
            ? event.meta.titleIds
            : [];
```

libs/event/src/lib/pages/analytics/analytics.component.ts
```diff
@@ -100,23 +105,29 @@ export class AnalyticsComponent implements OnInit {
...
+          .pipe(
+            joinWith({
+              org: analytic => this.orgService.valueChanges(analytic.meta.orgId),
+            }, { shouldAwait: true })
+          );
+
+        const promises = titleIds.map(titleId => firstValueFrom(titleAnalytics(titleId)));
+        const analytics = (await Promise.all(promises)).flat()
+          .filter(({ org }) => !org.appAccess.festival.dashboard)
+          .filter(analytic => analytic.name === 'screeningRequested');
```
RemcoSimonides, line 128: `getTitleAnalytics` function doesn't support this yet, but maybe we can pass the name 'screeningRequested' to this function and add this parameter to the query. Probably saves quite some doc reads

libs/event/src/lib/pages/analytics/analytics.component.ts
```diff
@@ -100,23 +105,29 @@ export class AnalyticsComponent implements OnInit {
...
+        const titleAnalytics = (titleId) => this.analyticsService.getTitleAnalytics({ titleId })
+          .pipe(
+            joinWith({
+              org: analytic => this.orgService.valueChanges(analytic.meta.orgId),
+            }, { shouldAwait: true })
+          );
+
+        const promises = titleIds.map(titleId => firstValueFrom(titleAnalytics(titleId)));
+        const analytics = (await Promise.all(promises)).flat()
+          .filter(({ org }) => !org.appAccess.festival.dashboard)
```
RemcoSimonides, line 127: During my holidays, Springfield chose to query only buyers analytics data differently than I would have done it. I would have liked to see it done already by the getTitleAnalytics() function. This would have saved you to join org on this page, and this filter here... Also it would have been cleaner everywhere else getTitleAnalytics is used. But maybe I'll do that later this month if there is time left.

apps/festival/festival/src/app/dashboard/analytics/buyer/buyer-analytics.component.html
phpgeekfr: yes but it also want section and articles to have a heading ;) https://www.notion.so/cascade8/Code-conventions-0dc3a5db4f1045f68f1587f9a88bcce2

but ok lets go with `article`

apps/festival/festival/src/app/event/session/session.component.ts
phpgeekfr: This is only triggered when hitting the main button : 
image

So to see it again, user will have to refresh page and if he already have some `watchInfos.duration` we will continue to increment time from this value

libs/event/src/lib/pages/analytics/analytics.component.ts
phpgeekfr: > But maybe I'll do that later this month if there is time left.

Like this because I don't know all possible edge cases

RemcoSimonides (approved):

Issue #8509, opened 2022-05-23
Cache issue with NgFire when we want to load all documents from collection


phpgeekfr:
Cache issue with NgFire when we want to load all documents from collection.

This happens only on CRM so it is not a big issue but when we want to load all documents from `users` or `organizations` we first see only the user or the org we are connected with and then all the other documents.

We don't know if bugs comes from our app or `ngFire`

phpgeekfr:
This first happened when we removed indexedDb cb89324

RemcoSimonides:
Issue also occurs for Sellers' Analytics feature. When for example navigating from Title Activity page to dashboard page, the dashboard page will first be loaded with Analytics from Title Activity page and then all Analytics

PR #8751, opened 2022-07-05
8702 fix loading pie chart

### LINKED ISSUE #8702: Sellers' analytics - Buyers' Activities - Pie chart loading super long
app: festival
page: c/o/dashboard/home/buyer

I went on seller with a good amount of analytics and the pie chart is super long to load (2-3min) though the numbers itself are loaded.

RemcoSimonides:
fix #8702
To reproduce the bug consistently I went to
1. /c/o/dashboard/title
2. Navigate to Dashboard
3. Click on "View all Buyers' Activitites"

On /c/o/dashboard/title, all the titles of the sellers org are queried. When going to the Buyers Activities page, it triggers the observables twice. First is for the sellers' org, and then again with more data. This is a bug that has been seen on CRM before.

However, this time, the `orgService.valueChanges(orgIds)` kept triggering infinitely (and thus the pie chart kept trying to re-render). Turning of memorization helped, or changing valueChangess to getValue also helped.

This is likely a bug in ngfire lib 

Might be caused by the same bug as https://github.com/blockframes/blockframes/issues/8509

### DIFF
--- a/libs/organization/src/lib/organization/service.ts
+++ b/libs/organization/src/lib/organization/service.ts
@@ -27,7 +27,6 @@ import { BlockframesCollection } from '@blockframes/utils/abstract-service';
 @Injectable({ providedIn: 'root' })
 export class OrganizationService extends BlockframesCollection<Organization> {
   readonly path = 'orgs';
-  readonly memorize = false;
 
   // Organization of the current logged in user or undefined if user have no org
   org: Organization; // For this to be defined, one of the observable below must be called before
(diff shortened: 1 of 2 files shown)


libs/organization/src/lib/organization/service.ts
```diff
@@ -27,6 +27,7 @@ import { BlockframesCollection } from '@blockframes/utils/abstract-service';
 @Injectable({ providedIn: 'root' })
 export class OrganizationService extends BlockframesCollection<Organization> {
   readonly path = 'orgs';
+  readonly memorize = false;
```
RemcoSimonides, line 30: Another solution would have been to use `getValue()` instead of `valueChanges()` in the query.

GrandSchtroumpf: If you don't care about realtime update you can use `load()`, it's an alias to `firstValueFrom(valueChanges())`

GrandSchtroumpf:
So it's only the orgService which is infinitely triggering ? Not the MovieService nor the UserService ?
Is the current orgId part of the `orgIds` here: `orgs: ({ orgIds }) => this.orgService.valueChanges(orgIds)` ?

GrandSchtroumpf (approved):

Issue #8622, opened 2022-06-13
Send an admin email each time an Event is created


cams-goussale:
### Description

In order for the business team to follow who is organizing an event, we shall send to support@archipelmarket.com each time a new event is organized.

Do I need to create a sendgrid template? Some emails triggered by actions in-app do not have a sendgrid template. For ex, `A movie has been submitted`.

As for the content of the email, we can go with

```
A new event has been created.
Visit https://crm.blockframes.io/c/o/dashboard/crm/event/[eventId]
```

cc @Jackseed

Jackseed:
So far all the admin emails (org created, new user connexion, movie submitted,..) are sent directly by the app without using any Sendgrid template

cams-goussale:
```
A new [eventType] has been created: [eventName]
Visit https://crm.blockframes.io/c/o/dashboard/crm/event/[eventId]
```

PR #8761, opened 2022-07-07
Send emails to admin when creating an event

Poppe-A:
#8622

### DIFF
--- a/apps/backend-functions/src/event.ts
+++ b/apps/backend-functions/src/event.ts
@@ -3,6 +3,9 @@ import { CallableContext } from 'firebase-functions/lib/providers/https';
 import { triggerNotifications } from './notification';
 import { Movie, Organization, PublicUser, EventMeta, Event, createInternalDocumentMeta, createPublicUser, createNotification } from '@blockframes/model';
 import { getDocument, BlockframesSnapshot } from '@blockframes/firebase-utils';
+import { sendEventCreatedEmail } from './templates/mail';
+import { getMailSender } from '@blockframes/utils/apps';
+import { sendMail } from './internals/email';
 
 /**
  * Removes invitations and notifications related to an event when event is deleted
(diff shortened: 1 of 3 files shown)


apps/backend-functions/src/event.ts
```diff
@@ -30,6 +33,13 @@ export async function onEventDelete(snap: BlockframesSnapshot<Event<EventMeta>>)
   return batch.commit();
 }
 
+export async function onEventCreate(snap: BlockframesSnapshot<Event<EventMeta>>) {
+  const event = snap.data();
+  const from = getMailSender('festival');
+  
+  await sendMail(sendEventCreatedEmail('festival', event), from);
```
RemcoSimonides, line 40:
```suggestion
  const emailData = sendEventCreatedEmail('festival', event);
  await sendMail(emailData, from);
```
I would have used a separate const to prevent function stacking.

apps/backend-functions/src/templates/mail.ts
```diff
@@ -574,3 +577,16 @@ export function sendMovieSubmittedEmail(app: App, movie: Movie) {
     `
   };
 }
+
+/** Send an email to supportEmails.[app] when a new event is created*/
+export function sendEventCreatedEmail(app: App, event: Event<EventMeta>) {
```
RemcoSimonides, line 582: I would have used a different function name (e.g. eventCreatedAdminEmail) because this name implies it is sending an email already.

apps/backend-functions/src/templates/mail.ts
```diff
@@ -574,3 +577,16 @@ export function sendMovieSubmittedEmail(app: App, movie: Movie) {
...
+  return {
+    to: getSupportEmail(app),
+    subject: 'A new event has been created.',
+    text: `
+      A new event ${event.title} has been created.
+
+      Visit ${appUrl.crm}${ADMIN_REVIEW_EVENT_PATH}/${event.id}  or go to ${ADMIN_REVIEW_EVENT_PATH}/${event.id} to review it.
+    `
```
RemcoSimonides, line 590: I know this is the suggested text in the issue, but might be useful to mention the meeting type (Screening/Slate/Meeting)

Maybe also start and end date

RemcoSimonides (approved):

Issue #8769, opened 2022-07-11
Release 4.4 - Analytics


RemcoSimonides:
[Analytics Test Scenarios](https://www.notion.so/cascade8/Analytics-8e793454a0d847ebb9a09e0d0bd7a461)

Issues resolved for this release:
- [ ] https://github.com/blockframes/blockframes/issues/8644
- [ ] https://github.com/blockframes/blockframes/issues/8696
- [ ] https://github.com/blockframes/blockframes/issues/8701
- [ ] https://github.com/blockframes/blockframes/issues/8702
- [ ] https://github.com/blockframes/blockframes/issues/8705
- [ ] https://github.com/blockframes/blockframes/issues/8717

cams-goussale:
- [ ] On page dashboard/home/title/[titleId], in the section `Last Screenings`, user can click on user names and get to dashboard/home/buyer/[titleId]. In case of a Protected or Public event, there can be users with no archipel account. We should in this case show some empty states on the Buyer report page. Currently it looks broken:

[UX Wireframes](https://www.figma.com/file/FhY2dJqeYrVTXh7Q7HA0m7/%E2%9C%85-%2B-%E2%9A%99%EF%B8%8F-Sellers'-Dashboard?node-id=293%3A114737)
![image](https://user-images.githubusercontent.com/93206363/178249305-2413d09a-5330-4cd6-b8b8-f858c84cadc7.png)

RemcoSimonides:
@cams-goussale this user doesn't have an orgId so the UX wireframes won't work for this scenario. Maybe it shouldn't be possible to navigate to this user at all? What do you think?

cams-goussale:
> @cams-goussale this user doesn't have an orgId so the UX wireframes won't work for this scenario. Maybe it shouldn't be possible to navigate to this user at all? What do you think?

Yes sounds good!

cams-goussale:
- [ ] Something is incoherent between the event statistics and the statistics shown on title report page.

![image](https://user-images.githubusercontent.com/93206363/179482915-31539588-4df7-4c6b-9cee-20d147f9c8d5.png)

![image](https://user-images.githubusercontent.com/93206363/179483047-7ad43b51-f04c-4adf-8f00-70f66684c74f.png)


[...]

PR #8778, opened 2022-07-12
8769 guard buyer activity for no org user

RemcoSimonides:
#8769


apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -113,7 +113,7 @@ <h3>Last Screenings</h3>
       <header>
         <button mat-button color="primary" [disabled]="!invitations.length" (click)="exportScreenerAnalytics()">Export to .CSV</button>
       </header>
-      <bf-table class="screenings" [source]="invitations" clickable useFilter pagination="5" (rowClick)="goToBuyer($event)">
+      <bf-table class="screenings" [source]="invitations" useFilter pagination="5">
```
phpgeekfr, line 116: you can remove `goToBuyer` in ts

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
```diff
@@ -142,6 +142,15 @@ <h3>Last Screenings</h3>
         <ng-template colRef="watchInfos.date" label="Watching Status" let-date sort>
           <bf-tag *ngIf="date | isWatchingNow">Watching Now</bf-tag>
         </ng-template>
+        <ng-template colRef label="Action" let-item sticky>
+          <ng-container *ngIf="item | guest:'user' as user">
+            <ng-container *ngIf="user.orgId">
```
phpgeekfr, line 147: not possible to use rxLet & ngIf in same ng-container toi avoid having two ?

apps/festival/festival/src/app/dashboard/dashboard.module.ts
```diff
@@ -72,7 +73,8 @@ const routes: Routes = [{
             },
             {
               path: ':userId',
-              loadChildren: () => import('./analytics/buyer/buyer-analytics.module').then(m => m.BuyerAnalyticsModule)
+              loadChildren: () => import('./analytics/buyer/buyer-analytics.module').then(m => m.BuyerAnalyticsModule),
+              canActivate: [HasOrganizationGuard]
```
phpgeekfr, line 77: what is the point of this guard if you don't have a link leading to users without org ?

RemcoSimonides: Isn't this the case with every guard? Why would we point to a page if we shouldn't be able to access it?
But you're right, in theory the user should never be able to get to this page except by changing the URL. Now that I say that, I should also prevent loading this page for non-existing users.

apps/festival/festival/src/app/dashboard/analytics/title/title-analytics.component.html
RemcoSimonides: Nice suggestion but no, not possible

Property 'user' does not exist on type 'TitleAnalyticsComponent'.
146           <ng-container *rxLet="item | guest:'user' as user" *ngIf="user.orgId">

apps/festival/festival/src/app/dashboard/dashboard.module.ts
phpgeekfr: most of our guards are here to prevent user to access a page that he can normally access but that we prevent because connected user does not have required access or permissions.

It look like that this is not likely that an user will try to change uid in URL

phpgeekfr (approved):

Issue #6556, opened 2021-08-31
Update cookie acceptation bar


Jackseed:
We shall have 3 buttons instead of 2: 
- Settings
- Accept
- **Decline**

Decline is thus new and should be same level as Accept. 

Once the UI is done this issue should be used in dev to 'refuse' cookie by default: e.g. as long as users haven't accepted cookies we shouldn't have Metrica

frenchyrules:
![image](https://user-images.githubusercontent.com/85220905/136182483-2e766c43-5bed-41ea-8125-33a594836381.png)
If this version is doable dev wise, rework it according to the following guidelines:
- Change the switches to checkboxes and add a third line/category for chatbot
- Add a switch to deactivate all notifications
- Have a "accept all" button, and cookie details button button 

**IF that is not doable, use version below** (may need a text rework though)
![image](https://user-images.githubusercontent.com/85220905/136182893-2866bb67-df81-482e-a1b2-44d1d72c1178.png)

frenchyrules:
**Reworked version**
![image](https://user-images.githubusercontent.com/85220905/136211527-ab454794-3d95-49dd-a933-b7037a3ff6b7.png)

Jackseed:
Thanks just a few comments:
- toggle and "Uncheck all" looks not completely centered
- Analytics: test with "Video Analytics" instead please
- 1 script > 1 cookie ? 
- "Accept all" > "Save my choice"

@GrandSchtroumpf @maltmayer what do you think of these propositions?

frenchyrules:
@GrandSchtroumpf @Jackseed @maltmayer Like this ? (tooltip text is a placeholder)

![image](https://user-images.githubusercontent.com/85220905/136224373-7b4c7638-46b2-4ca1-9a43-e8ae0ebaf4ad.png)
![image](https://user-images.githubusercontent.com/85220905/136225423-1611f59b-a287-40ca-98be-9f74ccfae85d.png)

maltmayer:
+ I would add an information icon for the tooltip
+ and I would put the check box on the right of "Unselect all non-essential"

+ Also we agreed with @Jackseed to change the text for "Video Analytics" to "Behavioral Analytics"
+ And lastly, please be careful with the capital letters for "More Details" and "Save Preferences"" ;) thanks!

frenchyrules:
Thanks for the feedback @maltmayer 🤝 Here's an updated version:
[...]

PR #8782, opened 2022-07-13
New cookie banner

Poppe-A:
#6556


libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.scss
```diff
@@ -1,6 +1,51 @@
...
+    gap: 16px;
+  }
+
+  section {
+   margin-top: 16px;
+  }
+
+  .mat-flat-button {
+    margin-right: 8px;
+  }
```
RemcoSimonides, line 50:
```suggestion
  section {
     margin-top: 16px;
     display: flex;
     gap: 8px;
  }
```
How I would have done it. In my opinion more robust against small changes (e.g. button style or button order)

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.ts
```diff
@@ -18,6 +19,10 @@ import { createModalData } from '@blockframes/ui/global-modal/global-modal.compo
 export class CookieBannerComponent implements OnInit {
 
   public hasAccepted = false;
+  public form = new FormGroup({
+    behaviorCookie: new FormControl(false),
+    chatCookie: new FormControl(false),
```
RemcoSimonides, line 24:
```suggestion
    hotjar: new FormControl(false),
    intercom: new FormControl(false),
```
For the users we avoid use technical terms in the front-end but in our code we can just use hotjar and intercom ;)

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.ts
```diff
@@ -58,6 +51,16 @@ export class CookieBannerComponent implements OnInit {
     })
   }
 
+  public saveCookies(decline: boolean = false) {
+    this.confirmCookies();
```
RemcoSimonides, line 55: So when decline is true, only this function is executed? Let's call that directly from the html instead.

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.ts
```diff
@@ -58,6 +51,16 @@ export class CookieBannerComponent implements OnInit {
...
+    const behaviorCookie = this.form.get('behaviorCookie').value;
+    const chatCookie = this.form.get('chatCookie').value;
+    if (!decline && behaviorCookie) this.gdpr.enableHotjar(true);
+    if (!decline && chatCookie) this.gdpr.enableIntercom(this.authService.profile, true);
```
RemcoSimonides, line 59:
```suggestion
    const { hotjar, intercom } = this.form.value
    
    this.gdpr.enableHotjar(hotjar);
    this.gdpr.enableIntercom(this.authService.profile, intercom);
```

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.module.ts
```diff
@@ -2,25 +2,37 @@
...
+    ImageModule,
+    
     // Material
     MatCardModule,
     MatButtonModule,
     MatDialogModule,
+    MatCheckboxModule,
+    MatTooltipModule,
+    MatIconModule,
+    MatListModule,
```
RemcoSimonides, line 35: Unused?

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie.form.ts
```diff
@@ -9,9 +9,9 @@ export interface CookiesConsent {
 
 function createCookiesConsent(params?: Partial<CookiesConsent>): CookiesConsent {
   return {
-    intercom: true,
+    intercom: false,
     yandex: true,
```
RemcoSimonides, line 13: You can do the same for yandex to avoid a bug once yandex is reactived

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.scss
```diff
@@ -1,6 +1,51 @@
 :host {
   position: fixed;
   bottom: 0;
```
RemcoSimonides, line 3: bottom: 20px;

libs/utils/src/lib/gdpr-cookie/cookie-form/cookie.form.ts
RemcoSimonides: I think it's better to have the checkboxes checked by default though
Now if the users doesn't explicitly check a checkbox, they don't accept any cookies...

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.scss
```diff
@@ -1,6 +1,51 @@
...
+  right: 20px;
   z-index: 2;
-  width: 100%;
+  width: 500px;
```
RemcoSimonides, line 6: I think it's important that it works well on mobile too. I think we should support to around 380px width. So please check if it looks good on small screen too

libs/utils/src/lib/gdpr-cookie/cookie-banner/cookie-banner.component.html
```diff
@@ -1,9 +1,48 @@
...
+        <mat-icon
+          svgIcon="info"
+          matTooltip="Chat cookies are used to allow the Archipel team to communicate with its users through the chatbot tool.">
+        </mat-icon>
+      </label>
+      <mat-checkbox color="primary" formControlName="intercom"></mat-checkbox>
+    </article>
+  </form>
+
+  <div>
```
RemcoSimonides, line 44: Avoid `div`. This can be footer or article

RemcoSimonides (approved):

Issue #8770, opened 2022-07-11
Release 4.4 - General UI


RemcoSimonides:
[General UI test scenarios](https://www.notion.so/cascade8/General-UI-db403143e36c480d9c6b562cc001119f)

Other issues that have been resolved for this release:
- [ ] https://github.com/blockframes/blockframes/issues/8739
- [ ] https://github.com/blockframes/blockframes/issues/6556

Will not be in release 
- [ ] https://github.com/blockframes/blockframes/issues/6447
- [ ] https://github.com/blockframes/blockframes/issues/6685

cams-goussale:
- [ ] Why is there a white line here? (/c/o/dashboard/organization)
![image](https://user-images.githubusercontent.com/93206363/179194754-e5c21877-0263-4e8d-9a0f-cea4f8926090.png)

phpgeekfr:
This is visible on chrome but not on firefox

phpgeekfr:
adding `margin-top : -1px` to 
```.filter {
  position: sticky;
  top: 0;
  z-index: 2;
  background-position: center center;
  background-size: cover;
  padding: 16px clamp(16px, 5vw, 64px);
}
```
solves the issue. but maybe there is a cleaner solution

PR #8793, opened 2022-07-15
quick fix - white line

KvnMrs:
issue: https://github.com/blockframes/blockframes/issues/8770


cams-goussale:
yeeah let's keep the line @phpgeekfr @KvnMrs

PR #8810, opened 2022-07-19
prod hotfix Buyer List only showing 25 orgs instead of 1000+

phpgeekfr:


apps/festival/festival/src/app/dashboard/organization/list/list.component.ts
```diff
@@ -37,24 +37,23 @@ export class ListComponent implements OnDestroy {
...
     this.orgs$ = this.orgResultsState.asObservable();
-    const search = createOrganizationSearch({ appModule: ['marketplace'], hasAcceptedMovies: false });
+    const search = createOrganizationSearch({ appModule: ['marketplace', '-dashboard'], hasAcceptedMovies: false });
     this.searchForm.setValue({ ...search, countries: [] });
     this.sub = this.searchForm.valueChanges.pipe(
       startWith(this.searchForm.value),
       distinctUntilChanged(),
       debounceTime(500),
       switchMap(() => this.searchForm.search()),
+      tap(res => this.nbHits = res.nbHits),
```
RemcoSimonides, line 47: This is on develop but not release?

phpgeekfr: I reverted back code from release to the one on develop and added `-dashboard` on algolia side so the filtering is done on algolia directly, to have correct value for nbHits

PR #8820, opened 2022-07-20
Fix user export

RemcoSimonides:
ERROR: firebase/firestore: Firestore (9.6.10): Using maximum backoff delay to prevent overloading the backend.
This is because Firestore doesn't like it if the same doc is queried multiple times at once.

In the code, we iterated over 4000 users and queried their permissions document. There is one permissions document per organization, and since users share the same organization, the same doc was queried many times at once.

### DIFF
--- a/apps/crm/crm/src/app/crm/pages/users/users.component.ts
+++ b/apps/crm/crm/src/app/crm/pages/users/users.component.ts
@@ -80,15 +80,15 @@ export class UsersComponent implements OnInit {
       this.exporting = true;
       this.cdr.markForCheck()
 
-      const roles = await this.permissionsService.getValue();
-      const getMemberRole = (orgId: string, uid: string) => {
-        const role = roles.find(role => role.id === orgId);
-        if (!role) return
-        return role.roles[uid];
+      const roles = await this.permissionsService.load();
+      const getMemberRole = (r: CrmUser) => {
+        const role = roles.find(role => role.id === r.orgId);
+        if (!role) return;
+        return role.roles[r.uid];
       }
 
       const getRows = users.map(async r => {
-        const role = getMemberRole(r.orgId, r.uid)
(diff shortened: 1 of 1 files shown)


apps/crm/crm/src/app/crm/pages/users/users.component.ts
```diff
@@ -77,8 +79,16 @@ export class UsersComponent implements OnInit {
     try {
       this.exporting = true;
       this.cdr.markForCheck()
+
+      const roles = await this.permissionsService.getValue();
```
RemcoSimonides, line 83: Querying all the permission documents instead of getting all the unique orgIds from the users we have here because since we have all the users, we should also have all the orgs.

apps/crm/crm/src/app/crm/pages/users/users.component.ts
```diff
@@ -77,8 +79,16 @@ export class UsersComponent implements OnInit {
...
+      const getMemberRole = (orgId: string, uid: string) => {
+        const role = roles.find(role => role.id === orgId);
+        if (!role) return
+        return role.roles[uid];
+      }
+
       const getRows = users.map(async r => {
-        const role = r.org ? await this.orgService.getMemberRole(r.org, r.uid) : '--';
+        const role = getMemberRole(r.orgId, r.uid)
```
GrandSchtroumpf, line 91: Why not providing the user directly instead of orgId & uid ?

apps/crm/crm/src/app/crm/pages/users/users.component.ts
GrandSchtroumpf: you should use `load` as we are in CRM to enable caching. Except if this value is supposed to change during the crm session. In the later case, you should use `valueChanges`

GrandSchtroumpf (approved):

Issue #8680, opened 2022-06-17
Changes on filters of the Library page


cams-goussale:
On festival app, can we add these additional filters to the /library page:

- Release year
- Festival name
- Qualification

Can we also group the festival names by category: A festival / B festival / C festival. Business team should provide this categorization.

2amarkina:
**v1 of Festival Filter (quick)**
![Снимок экрана 2022-06-30 в 16.11.29.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/cbba229a-8ac8-424c-b5d6-a1101f391dcf)

**v2 of Festival Filter (with categories)**
![Снимок экрана 2022-06-30 в 16.55.06.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/d3c2c5b9-ec02-493a-bc4f-ccf88d87e5fd)

**Qualifications**
![Снимок экрана 2022-06-30 в 17.01.08.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/b3b60ce4-2320-49a0-bd58-1898ff486192)

**All together on the Library Page**
![Снимок экрана 2022-06-30 в 17.19.27.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/9bb07b27-320b-4897-9d92-8ade9328b79b)

[All the wireframes can be found here](https://www.figma.com/file/M3bVHcVyEzmNWOXuBq0m3Q/Titles-Page?node-id=0%3A1)

cams-goussale:
Qualifications
- Remove Arts & Essais
- EOF: Add french speaking

cams-goussale:
[List of titles with Arts & Essais or Awards](https://docs.google.com/spreadsheets/d/10jJ7uRHhd-TCl4_7vmT3K9SFVDkRYnE_6AyJtrCUrBo/edit#gid=0)

PR #8822, opened 2022-07-20
AM - Changes on filters of the Library page

KevinDurand974:
From: https://github.com/blockframes/blockframes/issues/8680

- [x] Need migration for changing certifications in db ?
   <del>`No, this will be done manually`</del>
   `Done with migration after reflection`
   
   
Expect 1 bug for some movies who will display Qualifications / Certifications, like :
```
, EOF (French Speaking Films), European Qualification
```
_This will be corrected automatically when the anonymous database will be recreated after they manually edit the prod version._


phpgeekfr (commented):
this is the reason why you changed the excel templates ? 
image

libs/model/src/lib/static/static-model.ts
```diff
@@ -57,10 +57,7 @@ export const offerStatus = {
 export const negotiationStatus = ['pending', 'accepted', 'declined'] as const;
 
 export const certifications = {
-  artEssai: 'Art & Essai',
-  eof: 'EOF',
```
phpgeekfr, line 61: Camille said `Add (French Speaking Films) next to EOF` but other ceritifcations like awardedFilm aListCast have to be removed ?

libs/model/src/lib/static/static-model.ts
```diff
@@ -57,10 +57,7 @@ export const offerStatus = {
...
-  awardedFilm: 'Awarded Film',
-  aListCast: 'A-list Cast',
+  eof: 'EOF (French Speaking Films)',
   europeanQualification: 'European Qualification'
 } as const
```
phpgeekfr, line 62: please remove export on this 
```
export type StaticModel = typeof staticModel;
```

and thoses two unused lines
```
export type GetLabel<S extends Scope> = StaticModel[S][GetKeys<S>]
export type GetCode<S extends Scope> = GetKeys<S> | GetLabel<S>;
```

phpgeekfr (commented):
- [ ] something with responsive is broken.
on develop (and this branch)
image

on staging
image


If you can have a look ..

libs/model/src/lib/static/static-model.ts
KevinDurand974: These ones are not used anymore, I checked before doing this

KevinDurand974:
> this is the reason why you changed the excel templates ? image

Yes!

phpgeekfr:
@cams-goussale it now takes a bit too much place 
image

Is it ok or can we shrink some text buttons ?

cams-goussale:
> Is it ok or can we shrink some text buttons ?

there should be the scroll bar to see the rest of the filters. It's ok

phpgeekfr:
- [ ] if you have time, update this migration to perform needed updates `libs/devops/src/lib/firestoreMigrations/0084.ts`

libs/movie/src/lib/movie/form/search.form.ts
```diff
@@ -151,6 +157,8 @@ export class MovieSearchForm extends FormEntity<MovieSearchControl> {
   get hitsPerPage() { return this.get('hitsPerPage'); }
   get contentType() { return this.get('contentType'); }
   get runningTime() { return this.get('runningTime'); }
+  get festivals() { return this.get('festivals'); }
+  get certifications() { return this.get('certifications') }
```
phpgeekfr, line 161: for consistency
```suggestion
  get certifications() { return this.get('certifications'); }
```

libs/movie/src/lib/movie/form/search.form.ts
```diff
@@ -168,6 +176,7 @@ export class MovieSearchForm extends FormEntity<MovieSearchControl> {
       this.minBudget?.value === 0 &&
       this.minReleaseYear.value === 0 &&
       this.sellers?.value.length === 0 &&
+      this.festivals.value?.length === 0 &&
```
phpgeekfr, line 179: why no `certifications` here.
Also, whole `isEmpty` method looks like unused

phpgeekfr (approved):

Issue #7884, opened 2022-03-01
Add an anchor on Edit button - Contract Negotiation


cams-goussale:
When users are negotiating a contract

(on `/c/o/dashboard/sales/[contractId]/negotiation`
and `c/o/marketplace/offer/[offerId]/[contractId]/negotiate`)

and using the edit button, we should anchor to the top of the Right card.

https://user-images.githubusercontent.com/93206363/156163724-971de137-d0f4-4171-bdc5-a15b0b55bcb0.mp4

PR #8824, opened 2022-07-21
fixed(negotiation-form): auto-scroll to top of page

yspringfield:
Fixed: #7884


libs/contract/src/lib/negotiation/form/negotiation-form.component.html
```diff
@@ -1,12 +1,12 @@
-<header>
+<header data-scroll-to-view>
```
GrandSchtroumpf, line 1: prefer using an id instead of a `data-**` value. Using `data-**` is a native way to access custom attributes in JS. It shouldn't be used in angular

GrandSchtroumpf (approved):

PR #8842, opened 2022-07-26
8829 public screener in file explorer on CRM

### LINKED ISSUE #8829: Screener feature on CRM
With #8731 we created a different behavior for screener videos for each app. In CRM, organization page, section My files, it would be appreciated to have this behavior reflected. There would be a different component in festival app and in crm:

For festival app, do not show Public screener
For crm app, show Public screener

RemcoSimonides:
fix #8829


phpgeekfr (approved):
I just noticed something:
in movie tunnel for other video we have title and video type.
image
But we don't have it in file explorer
image

Issue #8686, opened 2022-06-17
Add a filter on org page catalog


cams-goussale:
Add the same filter we have on the Library page into the Organization page. So that Buyers can filter content of a specific Seller.

cams-goussale:
[UI mockups](https://www.figma.com/file/sygFKSH1P53d8TdZ4n8xWi/Sellers?node-id=0%3A1)

phpgeekfr:
- Do we have to implement load/save filter as in #8673 ?
- Here data comes from firestore directly but from algolia on organization page, so it can take some time to implement

KevinDurand974:
Sould I add on Financiers too?
It has the same page but with few or no movies. (more like no movie)

cams-goussale:
Yes let's put it everywhere for consistency.

phpgeekfr:
Closed this PR https://github.com/blockframes/blockframes/pull/8847 as work should be reworked with title search coming from Algolia (https://github.com/blockframes/blockframes/pull/8847#pullrequestreview-1054179349)

Jackseed:
deprecated issue

PR #8847, opened 2022-07-26
AM, MF - Add filters on org page (line-up)

KevinDurand974:
From issue: https://github.com/blockframes/blockframes/issues/8686

- [ ] Merge this one before: https://github.com/blockframes/blockframes/pull/8822


phpgeekfr (commented):
I think there should be a "no results" found screen when we filter and have no results 
image


As we just discussed on slack, it would be really better to make the title search from algolia as we already have for the global title list. This could allow us to create dedicated component to use in both places and have the same behavior (like the no results image as mentioned above).

@cams-goussale I believe this should also be implemented on catalog app ? 
image

If so, is the avails button required as we have on the library page ?
image

cams-goussale:
> @cams-goussale I believe this should also be implemented on catalog app ?

Yes, we can have the filter on catalog as well, including the Avails filter (if not a problem)

Issue #8832, opened 2022-07-25
Admin email - A movie has been submitted => add org name


Jackseed:
Best would be to change:
- [ ] The title with `{{movie.title.international}} was submitted by {{org.name}}` (we can use the first org owner of the movie since it's an array)
- [ ] The email text with `The new movie {{movie.title.international}} was submitted by {{org.name}}` then visit + link don't change

PR #8851, opened 2022-07-27
Admin email - add movie title and orgName 

KvnMrs:
issue: https://github.com/blockframes/blockframes/issues/8832

### DIFF
--- a/apps/backend-functions/src/movie.ts
+++ b/apps/backend-functions/src/movie.ts
@@ -140,7 +140,7 @@ export async function onMovieUpdate(change: BlockframesChange<Movie>) {
     const movieWasSubmittedOn = wasSubmittedOn(before.app, after.app)[0];
     // Mail to supportEmails.[app]
     const from = getMailSender(movieWasSubmittedOn);
-    await sendMail(await sendMovieSubmittedEmail(movieWasSubmittedOn, after, organizations[0]), from, groupIds.noUnsubscribeLink);
+    await sendMail(sendMovieSubmittedEmail(movieWasSubmittedOn, after, organizations[0]), from, groupIds.noUnsubscribeLink);
 
     // Notification to users related to current movie
     const notifications = organizations
--- a/apps/backend-functions/src/movie.ts
+++ b/apps/backend-functions/src/movie.ts
@@ -140,7 +140,7 @@ export async function onMovieUpdate(change: BlockframesChange<Movie>) {
     const movieWasSubmittedOn = wasSubmittedOn(before.app, after.app)[0];
     // Mail to supportEmails.[app]
     const from = getMailSender(movieWasSubmittedOn);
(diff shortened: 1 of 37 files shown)


apps/backend-functions/src/movie.ts
```diff
@@ -140,7 +140,7 @@ export async function onMovieUpdate(change: BlockframesChange<Movie>) {
     const movieWasSubmittedOn = wasSubmittedOn(before.app, after.app)[0];
     // Mail to supportEmails.[app]
     const from = getMailSender(movieWasSubmittedOn);
-    await sendMail(sendMovieSubmittedEmail(movieWasSubmittedOn, after), from, groupIds.noUnsubscribeLink);
+    await sendMail(await sendMovieSubmittedEmail(movieWasSubmittedOn, after, db), from, groupIds.noUnsubscribeLink);
```
phpgeekfr, line 143: you already have the organizations here 
`const organizations = await getOrganizationsOfMovie(after.id);`
```suggestion
    await sendMail(sendMovieSubmittedEmail(movieWasSubmittedOn, after, organizations[0]), from, groupIds.noUnsubscribeLink);
```

apps/backend-functions/src/templates/mail.ts
```diff
@@ -565,12 +566,15 @@ export function sendContactEmail(userName: string, userMail: string, subject: st
 }
 
 /** Send an email to supportEmails.[app](catalog & MF only) when a movie is submitted*/
-export function sendMovieSubmittedEmail(app: App, movie: Movie) {
+export async function sendMovieSubmittedEmail(app: App, movie: Movie, db: Firestore) {
+  const org = await db.collection('orgs').doc(movie.orgIds[0]).get()
+  const orgName = org.data().name;
+
```
phpgeekfr, line 572:
```suggestion
export function sendMovieSubmittedEmail(app: App, movie: Movie, org: Organization) {
  const orgName = org.name;

```

apps/backend-functions/src/movie.ts
KvnMrs: I  have to keep the 'await' without it i get an error, I find it litle strange

phpgeekfr: remove async here `export async function sendMovieSubmittedEmai`

apps/backend-functions/src/templates/mail.ts
```diff
@@ -33,6 +33,7 @@ import {
 } from '@blockframes/model';
 import { format } from 'date-fns';
 import { supportMailosaur } from '@blockframes/utils/constants';
+import { Firestore } from '@blockframes/firebase-utils';
```
phpgeekfr, line 36: you don't need this line

phpgeekfr (approved):

Issue #8834, opened 2022-07-25
CRM exports


cams-goussale:
We'd like to be able to export more info from the CRM to improve our internal KPIs documentation, and our operational follow ups (media upload, avails upload etc.).

Here is a list of items that would be needed:

- [x] blockframesAdmin can export all fields from movie collection. If there is some nested data, I think it's fine and we can do something basic.
- [x] blockframesAdmin can export Analytics collection. Currently, some of this data is displayed on movie page, so we could either add it to the Movie export, either do a separate Analytics page. (FYI the query Bruce gave me in BigQuery to extract the buyer interactions doesn't return any data any more. I think it's because it was previously tracked with GA and now it's internally in firebase.
- [x] blockframesadmin can export all contracts
- [x] blockframesAdmin can export all offers

RemcoSimonides:
Movie fields chosen not to export: boxOffice, originalRelease, review (they are array (a.k.a. list) of objects (a.k.a. multiple values per item in list) and therefore complex to put into single excel cells.

RemcoSimonides:
2 tasks done
2 left to do

cams-goussale:
If possible, add to this issue:

- [ ] in the csv export of org, add a column for the creation date of the organization

Jackseed:
@phpgeekfr let's see how difficult/long it is to conclude that issue and if we can keep it for next release

cams-goussale:
In the movie export, can we add some fields to what Remco already did?

- [ ] if title has a poster
- [ ] if title has a screener video
- [ ] if title has a publicVideo
- [ ] if title has a salesPitch video
- [ ] if title has otherVideos
- [ ] season number (titles.series)

Jackseed:
To be complete I would also add:
- [ ] if title has a banner
- [ ] if title has stills

phpgeekfr:
- [ ] some movies (`1oWLiyBJTV6Xs0XlRH7f` for example) have a media value for OriginalRelease that does not exists: 
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/571a98f2-07b0-4139-91c2-1a63b07c81ef)

cams-goussale:
>  some movies (1oWLiyBJTV6Xs0XlRH7f for example) have a media value for OriginalRelease that does not exists:

Oh, I thought we removed `Theatrical` in the Media list for Avails, but I didn't remember it was for OriginalRelease too.
[...]

PR #8860, opened 2022-07-29
8834 export all movie fields

RemcoSimonides:
#8834
- [x] blockframesAdmin can export all fields from movie collection. If there is some nested data, I think it's fine and we can do something basic.


apps/crm/crm/src/app/crm/pages/movies/movies.component.ts
```diff
@@ -92,9 +100,56 @@ export class MoviesComponent implements OnInit {
         'festival access': m.app.festival.access ? 'yes' : 'no',
         'financiers status': m.app.financiers.status,
         'financiers access': m.app.financiers.access ? 'yes' : 'no',
-        'screeningCount': m.screeningCount,
+        screeningCount: m.screeningCount,
         'creation date': format(m._meta.createdAt, 'MM/dd/yyyy'),
-        'last modification date': m._meta.updatedAt ? format(m._meta.updatedAt, 'MM/dd/yyyy') : '--'
+        'last modification date': m._meta.updatedAt ? format(m._meta.updatedAt, 'MM/dd/yyyy') : '--',
+        'audience goals': m.audience?.goals?.length ? smartJoin(m.audience.goals) : '',
+        'audience targets': m.audience?.targets?.length ? toLabel(m.audience.targets, 'socialGoals') : '',
```
phpgeekfr, line 107: For `toLabel` I don't think you need to test length before as it checks `if (!value) return '';`

apps/crm/crm/src/app/crm/pages/movies/movies.component.ts
```diff
@@ -92,9 +100,56 @@ export class MoviesComponent implements OnInit {
...
+        cast: smartJoin(m.cast.map(person => displayName(person))),
+        'content type': toLabel(m.contentType, 'contentType'),
+        crew: smartJoin(m.crew.map(person => displayName(person))),
+        directors: smartJoin(m.directors.map(person => displayName(person))),
+        'estimated budget': toLabel(`${m.estimatedBudget}`, 'budgetRange'),
+        'expected premiere': m.expectedPremiere?.event ? `${m.expectedPremiere.event} (${m.expectedPremiere?.date ? format(m.expectedPremiere.date, 'MM/dd/yyyy') : 'no date'})` : '',
+        format: m.format ? toLabel(m.format, 'movieFormat') : '',
+        'format quality': m.formatQuality ? toLabel(m.formatQuality, 'movieFormatQuality') : '',
```
phpgeekfr, line 115: same here

libs/utils/src/lib/helpers.ts
```diff
@@ -79,7 +79,11 @@ export function downloadCsvFromJson(data: unknown[], fileName = 'my-file') {
   const replacer = (_: unknown, value: unknown) => (value === null ? '' : value);
   const header = Object.keys(data[0]);
   const csv = data.map((row: unknown) =>
-    header.map((fieldName) => JSON.stringify(row[fieldName], replacer)).join(',')
+    header.map((fieldName) => {
+      const value = row[fieldName];
+      if (typeof value === 'string') return `"${value.replace(/"/g, '""')}"`; // escaping double quote
```
phpgeekfr, line 84: can you provide an example ?
Not sure to understand why you replace `"` with  `""`

phpgeekfr (approved):

PR #8913, opened 2022-09-20
cleaning festival-e2e and e2e libs (no longer used)

n-germain-work:


libs/devops/src/lib/db-shrink.ts
```diff
@@ -11,11 +11,6 @@ import { cleanDeprecatedData } from './db-cleaning';
 import { getFirestoreEmulator } from '@blockframes/firebase-utils/initialize';
 import { unique } from '@blockframes/utils/helpers';
 
-/**
- * Temp this should be removed when fixtures are updated.
- *  - libs/e2e/src/lib/fixtures/users.ts
- *  - libs/e2e/src/lib/utils/screenings.ts
- */
```
phpgeekfr, line 18: USERS :
- `'qFbytROWQYWJplzck42RLdgAr3K3'` 
- `'K0ZCSd8bhwcNd9Bh9xJER9eP2DQ2'` 
are still referenced here `apps/festival/festival-e2e/src/fixtures/screening-events.ts` used in `apps/festival/festival-e2e/src/integration/dashboard/screening-invitations-2688.spec.ts`.

So what I would do is :

```
/**
 * Temp this should be removed when fixtures are updated.
 *  - apps/festival/festival-e2e/src/fixtures/screening-events.ts
 */
const USERS = [
  'MDnN2GlVUeadIVJbzTToQQNAMWZ2',
  '2OJUZoWtTVcew27YDZa8FQQdg5q2',
  'B8UsXliuxwY6ztjtLuh6f7UD1GV2',
  'mVUZ097xoAeubsPiQlqrzgUF8y83'
];
```

phpgeekfr (commented):
- [x] Some methods in `libs/testing/cypress/node/src/lib/plugins/firestore.ts` are unused and other should not be `exported` (used only within the file)
- [ ] is `apps/festival/festival-e2e/src/support/index.ts` used ?

n-germain-work:
> `apps/festival/festival-e2e/src/support/index.ts` -> cleared
> `apps/festival/festival-e2e/src/support/index.ts` -> it is the link between Cypress and our testing lib. We could directly configure Cypress to import the lib, but I believe the `support/index.ts` to be the default support file, it should be easier for a new dev to understand the flow of the code with it.

phpgeekfr (approved):

PR #9249, opened 2023-03-09
Waterfall rules - WIP

phpgeekfr:


libs/testing/unit-tests/src/lib/fixtures/rules.ts
```diff
@@ -15,6 +15,12 @@ export const rulesFixtures = {
     },
     id: 'O001',
   },
+  'permissions/O002': {
+    roles: {
+      'uid-user3': 'member',
+    },
+    id: 'O002',
+  },
```
phpgeekfr, line 23: fixed existing tests: user was not member of org so test could pass because of this.

libs/user/src/lib/auth/service.ts
```diff
@@ -35,7 +35,7 @@ export class AuthService extends BlockframesAuth<User> implements OnDestroy {
   };
   readonly path = 'users';
 
-  readonly idKey: 'uid';
+  readonly idKey = 'uid';
```
phpgeekfr, line 38: this was working without this. @GrandSchtroumpf  I guess `uid` is the default key for auth collection on ngfire ?

GrandSchtroumpf: I guess too.

PR #9250, opened 2023-03-10
quickfixes

n-germain-work:
- [x] E2E crm : in `create-org test`, once in 240 times, `getRandom()` would select `world` as a random territory. As there is no such option in the from, the test would crash

- [x] E2E catalog : in `deal-seller-negociation test`, emails could arrive in an unpredicted order, leading to a crash, as one recipient could receive 2 mails, and we were checking the subject afterwards. To prevent this, we now intercept emails depending on both recipient AND subject.

### DIFF
--- a/apps/catalog/catalog-e2e/src/integration/dashboard/deal-seller-negotiatiation.spec.ts
+++ b/apps/catalog/catalog-e2e/src/integration/dashboard/deal-seller-negotiatiation.spec.ts
@@ -218,8 +218,8 @@ describe('Deal negotiation', () => {
 
 function checkConfirmationEmails(decision: ContractStatus) {
   for (const user of ['buyer', 'seller', 'admin']) {
-    interceptEmail({ sentTo: mailData[user].recipient }).then(mail => {
-      expect(mail.subject).to.eq(mailData[user].subject[decision]);
+    interceptEmail({ sentTo: mailData[user].recipient, subject: mailData[user].subject[decision] }).then(mail => {
+      cy.log(`mail received by ${user} ${mailData[user].recipient}: ${mailData[user].subject[decision]}`);
       return deleteEmail(mail.id);
     });
   }
--- a/apps/crm/crm-e2e/src/fixtures/create-org.ts
+++ b/apps/crm/crm-e2e/src/fixtures/create-org.ts
@@ -110,6 +110,8 @@ function createOrgWithuser(status: OrganizationStatus, apps: Exclude<App, 'crm' 
 }
 
 function getRandom<S extends Scope>(base: S): keyof StaticModel[S] {
-  const keys = Object.keys(staticModel[base]);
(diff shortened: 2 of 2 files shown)


apps/catalog/catalog-e2e/src/integration/dashboard/deal-seller-negotiatiation.spec.ts
```diff
@@ -218,8 +218,8 @@ describe('Deal negotiation', () => {
 
 function checkConfirmationEmails(decision: ContractStatus) {
   for (const user of ['buyer', 'seller', 'admin']) {
-    interceptEmail({ sentTo: mailData[user].recipient }).then(mail => {
-      expect(mail.subject).to.eq(mailData[user].subject[decision]);
+    interceptEmail({ sentTo: mailData[user].recipient, subject: mailData[user].subject[decision] }).then(mail => {
+      cy.log(`mail received by ${user} ${mailData[user].recipient}: ${mailData[user].subject[decision]}`);
```
phpgeekfr, line 222: no more `except` ? What is happening if mail is not found ?

apps/crm/crm-e2e/src/fixtures/create-org.ts
```diff
@@ -110,6 +110,8 @@ function createOrgWithuser(status: OrganizationStatus, apps: Exclude<App, 'crm'
 }
 
 function getRandom<S extends Scope>(base: S): keyof StaticModel[S] {
-  const keys = Object.keys(staticModel[base]);
+  // we filter to avoid 'world' as there is not such option in mat-select
+  // it won't affect the result for random orgActivity
+  const keys = Object.keys(staticModel[base]).filter(territory => territory !== 'world');
   return keys[(keys.length * Math.random()) << 0] as any;
 }
```
phpgeekfr, line 117: this method is not specific to territories.
You should do it a bit like in html :
```
<static-select [withoutValues]="['world']" required [control]="form.get('country')" scope="territories" test-id="country">
```
For example
```
function getRandom<S extends Scope>(base: S, withoutValues: string[] = []): keyof StaticModel[S] {
  const keys = Object.keys(staticModel[base]).filter(k => !withoutValues.includes(k));
  return keys[(keys.length * Math.random()) << 0] as any;
}
```

apps/catalog/catalog-e2e/src/integration/dashboard/deal-seller-negotiatiation.spec.ts
n-germain-work: No more `expect` indeed, as we're directly checking recipient and subject by retrieving only the email that matches both.
If the mail is not found, then the test crashes, as it should, saying that after 20 seconds, it couldn't find the mail.
The `cy.log` is to keep a visible trace in Cypress that we found it.

phpgeekfr (approved):

Issue #9248, opened 2023-03-08
Catalog - Small T&C text updates


2amarkina:
- [x] From `Last edited March 1st, 2023` to `Last edited: the 1st of March 2023`
- [x] from "Casacde8" to "Cascade8"
- [x] Change the spaces before : everyone (delete the spacing) 
- [x] From` rightholder` to `right holder`

The highlighted corrections could be found [here ](https://docs.google.com/document/d/17giet_6T6aQnInUgxPC6AqE1jrQX4mb66TKffa8eYVQ/edit) (p.1, pp.10-11)

n-germain-work:
Regarding this issue, I found that most of the time, there was a space between Cascade & 8, so I've added this space when replacing `Casacde8`.
Please tell me if this is not ok.

Issue #9246, opened 2023-03-08
Catalog - Commission to add 


2amarkina:
**Context:** 
During the business point it was decided to add the information about the commission to provide a clearer information to the user. 

It is important to add the following phrase `This amount includes the commission due to Archipel Content` on several levels: 
1. Add it to **the mail that a Seller** receives with an offer created (when an Offer has a price)
```
`Hi {{{user.firstName}}},

Great news! {{buyerOrg.name}} 
         {{#if negotiation.price}}offers {{negotiation.currency}} {{negotiation.price}} for {{movie.title.international}}. This amount includes the commission due to Archipel Content. 
          {{else}}is interested in {{movie.title.international}} and would like to know your Asking Price.
          {{/if}}

--- Start discussions button {{{pageUrl}}} ---

All our best,
The {{app.name}} team`
```

2. Add this phase on the Seller's Dashboard page 
As this component is used in a lot of places we need to put it only : 
- on the Seller's dashboard side
- if there is a price proposed by a Buyer. 
![Capture d’écran 2023-03-08 à 16.19.48.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/8e210c61-03c0-421a-827d-ccf115f87157)

For the issue 2, we need to estimate the time with devs and if it takes too much time then maybe put it aside

n-germain-work:
Point2 done.
@2amarkina can you please close the issue when the email part is done please :)

Issue #9244, opened 2023-03-02
Catalog - Remove "Contact XXX" on the marketplace movie view avails tab


Jackseed:
We removed the logo and seller name but I forgot about this little sentence where we show the seller's name and email. 
It should be removed as well
![Pasted Graphic.png](https://images.zenhubusercontent.com/5c6a89ac8de1b31d5443ab99/d180dc87-3557-4d03-b7f0-0b496f71f7e2)

2amarkina:
text: contact Archipel Content 
action: opens Intercom

Issue #9216, opened 2023-02-17
Archipel Content (list of the rights improvement)


2amarkina:
- [x] Claire proposed to remove the "-" in the cases as S-VOD, A-VOD etc. 

This might be useful if a user enters the data directly and doesn't use the search. For the moment the writing should be exactly the same and it might be confusing if a user enters "SVOD" instead of "S-VOD". 

_Example:_ 
![Снимок экрана 2023-02-17 в 15.51.33.png](https://images.zenhubusercontent.com/5dd54528c0ccd300013672c3/09761a94-493b-46d3-bd19-05f736ce8cd4)

PR #9261, opened 2023-03-15
Nicolas/quickfixes

n-germain-work:
Fixes issues :
- #9216
- #9244 
- #9246 
- #9248


apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
```diff
@@ -7,8 +7,7 @@ <h2>Available Territories</h2>
     <p>
       Click on the territories you're interested in to add them to your Selection. For territories that are not
```
phpgeekfr, line 8: you don't need `<ng-container *ngIf="org$ | async as org">` anymore.
`org$ ` attribute can maybe be removed from `ts` file also

apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.html
```diff
@@ -3,14 +3,11 @@
...
-      Click on the territories you're interested in to add them to your Selection. For territories that are not
-      licensed,
-      feel free to
-      <a [routerLink]="['/c/o/marketplace/organization', org.id, 'member']">contact {{ org.name }}</a>.
-    </p>
-  </ng-container>
+  <p>
+    Click on the territories you're interested in to add them to your Selection. For territories that are not
+    licensed,
+    feel free to <button (click)="openIntercom()">contact Archipel Content Team</button>.
```
phpgeekfr, line 9: you could use `appName` instead of raw `Archipel Content`

phpgeekfr (approved):

Issue #9359, opened 2023-05-19
language update needed on movies (all apps)


n-germain-work:
Some movies in database and algolia have wrong values and leads to bugs in UI :
- language with uppercase
- genre instead of language
![image](https://github.com/blockframes/blockframes/assets/78768220/98e0d2e8-adcf-4a21-ae57-fb08852f1f9f)
![image](https://github.com/blockframes/blockframes/assets/78768220/9fb848be-4b3b-44f1-9410-e8ba2eafe7f9)
It could be when a user import an excel file, if there are not guards to prevent this.

Additionally, following the resolution of issue #9286 , some movies' language might need an update :
- for example, we had 'hausa' & 'haussa' languages. The second one is a typo and has been deleted. Therefore, impacted movies should be updated.

There is still `chinese` and `mandarin-chinese` in the languages, as there is a slight difference between the two. Should we merge them  or keep as is ? To be discussed.

phpgeekfr:
- [ ] also correct genres
![image.png](https://images.zenhubusercontent.com/5c6e955fdecd2e3c69b6c1d2/d228bdc7-7aee-44b4-af8c-4891ab950a34)

PR #9385, opened 2023-06-02
language update needed on movies (all apps) #9359

phpgeekfr:
language update needed on movies (all apps) #9359

- fix excel import
- migration to correct languages and territories on orgs movies & terms


libs/import/src/lib/utils.ts
```diff
@@ -389,9 +389,4 @@ export class ImportWarning<T> extends ImportLog<T> {
 
 export class WrongTemplateError<T> extends ImportLog<T> {
   public readonly type = 'error';
-  /**
-   * When present, should be shown as the only error even
-   * when there are many other errors.
-   */
-  public onlyErrorShown = true;
```
phpgeekfr, line 396: this was not used

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -133,12 +131,7 @@ export async function parse<T>(
       if (last) {
         try {
           const result = await transform((`${value ?? ''}`).trim(), entity, state, rowIndex);
-          if (isValueWithError(result)) {
-            errors.push(result.error);
-            item[segment] = result.value;
-          } else {
```
phpgeekfr, line 139: we where never entering this part

libs/utils/src/lib/spreadsheet/import.ts
```diff
@@ -112,9 +111,8 @@ export async function parse<T>(
         const validResults = results.filter(r => !isValueWithError(r));
         item[field] = validResults;
 
-        const errorResults = results.filter(r => isValueWithError(r)) as ValueWithError[];
```
phpgeekfr, line 115: forcing type was a mistake

pldespaigne (approved):

Issue #9389, opened 2023-06-06
Documents Management 


2amarkina:
### Context
As discussed it would be better to introduce the file system to be able to manage all the documents inside the Waterfall app. That could help to get more space in the tabs navigation section of the film page. 

We could use the system that already exists for the Festival app in the "My Files" section with adding some supplementary upgrades. 

### Goals 
**General** 
- [x] Merge the existing tabs Financing Plan, Budget and Contracts to get one tab called "Documents" 
- [ ] Introduce three pre-built folders inside documents : Financing Plan, Budget and Contracts 
- [ ] Introduce the sharing icon in front of each folder 
- [ ] If possible highlight each line on hover 
- [ ] Inside the "Contracts" folder there are the following folders that are pre-built (order to confirm): Co-productions, Institutions (subsidies), Financiers, Mandates, Authors, Cast, Music, Other. 
- [ ] Introduce the sharing logic and on the folder's level and on each document (see the specs below)

**Sharing process & Roles' Access**
- [ ] Clicking on the "share" icon opens the modal.
- [ ] The header should say what kind of document user is about to share "You are about to share [Document Name]" or "Your are bout to share [Folder Name]". 
- [ ] User can see the list of the organizations and their roles (Right Holders involved into the Waterfall). User doesn't see the name or the email address of a particular user.
- [ ] User can choose one or several organizations with the help of checkboxes.
- [ ] The documents are automatically shared with the people associated with the concerned organizations
- [ ] After sending the documents (after sharing) show a snackbar "[Document Name] was successfully shared." or "[Folder Name] folder was successfully shared."
- [ ] Everyone can upload documents 
- [ ] Creator can edit, delete the documents 
uploaded

### Sources

[Documents View ](https://www.figma.com/file/s0qk9gGKVSJl8zXTPuNipr/UI---Blockframes?type=design&node-id=632-39695&t=krGmae3sDzghAMfN-4)

[Contracts View](https://www.figma.com/file/s0qk9gGKVSJl8zXTPuNipr/UI---Blockframes?type=design&node-id=632-39844&t=krGmae3sDzghAMfN-4)

[View end of the root of the folder](https://www.figma.com/file/s0qk9gGKVSJl8zXTPuNipr/UI---Blockframes?type=design&node-id=632-40033&t=d6cvkpNXKGGk9giK-4) (ex. Documents > Financing Plan, Documents > Budget, Documents > Contracts > Other)

[...]

PR #9401, opened 2023-06-09
WIP model & rules setup #9389

phpgeekfr:
WIP model & rules setup #9389

### DIFF
--- a/libs/model/src/lib/media.ts
+++ b/libs/model/src/lib/media.ts
@@ -6,7 +6,7 @@ import { privacies, Privacy } from './static';
  */
 export interface StorageFile {
   privacy: Privacy;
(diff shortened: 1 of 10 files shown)


pldespaigne (approved):

libs/model/src/lib/media.ts
```diff
@@ -45,7 +45,7 @@ export interface FileMetaData {
   uid: string;
   fileId?: string; // unique id required for list of videos
   privacy: Privacy;
-  collection: 'users' | 'orgs' | 'movies' | 'campaigns';
+  collection: 'users' | 'orgs' | 'movies' | 'campaigns' | 'waterfall';
```
pldespaigne, line 48: This is also defined in `CollectionHoldingFile` in `libs/media/src/lib/utils.ts`.
Might be worth to merge both, to avoid divergent types maybe?

PR #9431, opened 2023-07-04
Waterfall Right Holders form

pldespaigne:
- [x] commented WIP document code (fakeOrgs, and doc upload)
- [x] customm mat-stepper
- [x] rightholders org form


libs/waterfall/src/lib/components/right-holder-form/right-holder-form.component.html
```diff
@@ -0,0 +1,35 @@
+<section class="surface">
```
phpgeekfr, line 1: looks like quite the same code as apps/waterfall/waterfall/src/app/dashboard/waterfall/right-holders/right-holders.component.html

libs/waterfall/src/lib/components/contracts-form/contracts-form.component.ts
```diff
@@ -0,0 +1,31 @@
...
+
+@Component({
+  selector: 'waterfall-contracts-form',
+  templateUrl: './contracts-form.component.html',
+  styleUrls: ['./contracts-form.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ContractsFormComponent {
+
+  public form = new FormArray([
```
phpgeekfr, line 13: Unused code but I guess it will be used soon

libs/waterfall/src/lib/components/contracts-form/contracts-form.module.ts
```diff
@@ -0,0 +1,37 @@
...
+  imports: [
+    CommonModule,
+    ReactiveFormsModule,
+    StaticSelectModule,
+
+    // Material
+    MatIconModule,
+    MatInputModule,
+    MatButtonModule,
+    MatFormFieldModule,
```
phpgeekfr, line 30: same for some imports

libs/waterfall/src/lib/components/right-holder-form/right-holder-form.component.html
phpgeekfr: ah, this one is for fakes (rightholders). Might be confusing with other component where we manage real right holders

libs/waterfall/src/lib/form/file.form.ts
```diff
@@ -0,0 +1,39 @@
+
```
phpgeekfr, line 1: maybe rename folder to `forms` as we might have several of them

phpgeekfr (approved):

PR #9432, opened 2023-07-04
fix issue #9388 for waterfall

### LINKED ISSUE #9388: New Header (Blockframes & Archipels)
### Context:
Following the discussion inside the product & business team it was decided to optimize the headers of the Movie Page on the Dashboard on all the apps. 

### Goals: 
**For the Waterfall app** : 
- [x] Use the reduced size of the poster (check if we already have this size / approximately height:50%, but check together with design in real time)
- [x] Keep the Title name
- [x] Keep the Director
- [x] Place the edit icon near the title 
- [x] The click on the "edit" icon leads to the Edit Movie page 
- [x] On the Edit Movie page change the CTA text from "Edit Movie" to "Save Movie" 

**For the Festival & Catalog:** 
- [x] Use the reduced size of the poster (check if we already have this size / approximately height:50%, but check together with design in real time)
- [x] Keep the Title name
- [x] Keep the Director
- [x] The status should be moved on the right side above the main CTA 
- [x] Place the edit icon near the title 
- [x] The click on the "edit" icon leads to the Movie Tunnel
- [x] "Actions" become the Main CTA 
- [x] "Manage Avails" goes inside the "Actions"

n-germain-work:
fix #9388 

smaller header
image

`save `instead of `edit`
image

### DIFF
--- a/libs/movie/src/lib/movie/dashboard/shell/shell.component.html
+++ b/libs/movie/src/lib/movie/dashboard/shell/shell.component.html
@@ -18,7 +18,7 @@
           </article>
         </section>
         <aside fxLayout="column" fxLayout.lt-md="row" fxLayoutAlign="space-between end" >
-          <bf-tag>{{ movie.app[app].status | toLabel: 'storeStatus' }}</bf-tag>
+          <bf-tag *ngIf="app !== 'waterfall'">{{ movie.app[app].status | toLabel: 'storeStatus' }}</bf-tag>
           <ng-content select="movie-cta"></ng-content>
         </aside>
       </ng-container>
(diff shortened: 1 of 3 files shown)


libs/movie/src/lib/movie/dashboard/shell/shell.component.html
```diff
@@ -18,7 +18,7 @@ <h6>{{ movie.directors | displayName }}</h6>
           </article>
         </section>
         <aside fxLayout="column" fxLayout.lt-md="row" fxLayoutAlign="space-between end" >
-          <bf-tag>{{ movie.app[app].status | toLabel: 'storeStatus' }}</bf-tag>
+          <bf-tag *ngIf="app !== 'waterfall'">{{ movie.app[app].status | toLabel: 'storeStatus' }}</bf-tag>
```
phpgeekfr, line 21: Status will always be the same for a movie on this app. We can't hide it ?

phpgeekfr (approved):

PR #9438, opened 2023-07-07
Feat 9420 actions import

phpgeekfr:


libs/model/src/lib/income.ts
```diff
@@ -15,8 +15,10 @@ export interface Income {
   titleId?: string;
   status: 'pending' | 'processed';
   date: Date;
+  /** For waterfall purposes, either medias/territories or sourceId should be defined */
   medias: Media[];
   territories: Territory[];
+  sourceId: string;
```
pldespaigne, line 21: Maybe worth to type it like this then 🤔
```TypeScript
interface IncomeBase { ... };

interface IncomeA extends IncomeBase {
  medias: Media[];
  territories: Territory[];
}

interface IncomeB extends IncomeBase {
  sourceId: string;
}

export type Income = IncomeA | IncomeB;
```

pldespaigne (approved):

libs/model/src/lib/income.ts
phpgeekfr: indeed it creates typing errors. And for querying purposes, it might be better to have documents with same attributes even empty

PR #9456, opened 2023-08-01
fix #9381

### LINKED ISSUE #9381: Be able to delete all the titles in Wishlist at the same time
For the moment business team has to delete manually the titles from the wishlist one by one, that's time-consuming. 

- [ ] Add a button "Delete titles from the wishlist" which would delete all the added titles.

n-germain-work:
image
image


libs/ui/src/lib/marketplace/pages/wishlist/wishlist.component.ts
```diff
@@ -84,6 +84,14 @@ export class WishlistComponent implements OnInit, OnDestroy {
     });
   }
 
+  public clear(event: Event) {
+    event.stopPropagation();
```
phpgeekfr, line 88: You don't need `stopPropagation` here 
```suggestion
  public clear() {
```

PR #9477, opened 2023-08-29
adding hint and snackbar after map component load if past dates

n-germain-work:
In case the url of a past research in map component has been saved in a user's favorites, it will now display the hint and snackbar messages if a date is in the past :
image

### DIFF
--- a/apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.ts
+++ b/apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.ts
@@ -20,6 +20,7 @@ import {
 import { AnalyticsService } from '@blockframes/analytics/service';
 import { MovieService } from '@blockframes/movie/service';
 import { Intercom } from 'ng-intercom';
+import { isPast } from '@blockframes/utils/helpers';
 
 @Component({
   selector: 'catalog-movie-avails-map',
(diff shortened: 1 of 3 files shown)


libs/utils/src/lib/helpers.ts
```diff
@@ -163,4 +163,11 @@ export function convertToTimeString(time: number) {
   const secondStr = second > 0 ? `${second}s` : '';
 
   return `${dayStr}${hourStr}${minuteStr}${secondStr}` || '0s';
-}
\ No newline at end of file
+}
+
+// do not replace with isPast() from date-fns
+export function isPast(date: Date) {
```
phpgeekfr, line 169: maybe rename to `isInPast` to avoid confusions.

phpgeekfr: You could also allow an array of dates so you can do `isInPast([from, to])`
For example : 
```
/**
 * Returns true of one of the dates passed in arguments is in the past
 * @param _dates 
 * @returns boolean
 */
export function isInPast(_dates: Date | Date[]) {
  const dates = !Array.isArray(_dates) ? [_dates] : _dates;
  const now = new Date();
  now.setHours(0, 0, 0, 0);
  return dates.some(date => date < now);
}
```

phpgeekfr: or since this function is only for testing a `Duration` you could put it in `libs/model/src/lib/terms.ts`

Example : 
```
export function isInPast(duration: Duration) {
  const now = new Date();
  now.setHours(0, 0, 0, 0);
  return duration.from < now || duration.to < now;
}
```

apps/catalog/catalog/src/app/marketplace/title/avails/map/avails-map.component.ts
```diff
@@ -151,6 +152,11 @@ export class MarketplaceMovieAvailsMapComponent implements AfterViewInit, OnDest
     const decodedData = decodeUrl<MapAvailsFilter>(this.route);
     this.load(decodedData);
 
+    if (isPast(decodedData.duration?.from) || isPast(decodedData.duration?.to)) {
```
phpgeekfr, line 155: to avoid entering function with undefined value : 
```suggestion
    if (decodedData.duration && (isPast(decodedData.duration.from) || isPast(decodedData.duration.to))) {
```

phpgeekfr (approved):

PR #9483, opened 2023-08-31
avails link with parameters when exporting

### LINKED ISSUE #9474: Prefilled Avails and Wishlist
**Context**
Following [the previous issue ](https://app.zenhub.com/workspaces/blockframes-60b0a31838e485001b3a619f/issues/gh/blockframes/blockframes/9384) we are able to save the results of the Avails search from the Library Page to Film Page in particular. 

We also have a feature, when a user exports the .PDF with the pre-filtered films, s/he can follow the link to the Avails of a particular film.

**To do**
- [ ] Check if we are able to save the avails search parameters when we export the PDF. 
- [ ] Pay attention, that if the date is different from "today" the error message and snackbar should be maintained.

n-germain-work:
fix #9474


libs/utils/src/lib/pdf.service.ts
```diff
@@ -129,6 +129,16 @@ export class PdfService {
...
+
+        if (
+          availForm.territories?.length &&
+          availForm.medias?.length &&
+          availForm.exclusive !== undefined &&
+          availForm.duration.from &&
+          availForm.duration.to
+        ) {
+          filters.availsFormValue = encodeURIComponent(JSON.stringify(availForm));
+        }
```
phpgeekfr, line 141: you are already inside a `if` that tested availForm.territories?.length & availForm.medias?.length so no need to test it again.

you could put `filters.availsFormValue = encodeURIComponent(JSON.stringify(availForm));`
directly inside `if (availForm.duration.from && availForm.duration.to) {` and simply add a check for `exclusive` if needed

phpgeekfr (approved):

PR #9541, opened 2023-11-03
Waterfall Graph

pldespaigne:


libs/ui/src/lib/graph/graph.component.ts
```diff
@@ -0,0 +1,97 @@
...
+import { BehaviorSubject } from 'rxjs';
+
+
+@Component({
+  selector: 'bf-graph-container',
+  templateUrl: './graph.component.html',
+  styleUrls: ['./graph.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export default class GraphContainerComponent implements AfterViewInit {
```
phpgeekfr, line 12: to stick with what we are used to :

```suggestion
export  class GraphContainerComponent implements AfterViewInit {
```
in module
```
import { GraphContainerComponent }  from './graph.component';
```

libs/ui/src/lib/graph/graph.component.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  dragging = false;
+
+  public offset = new BehaviorSubject({ x: 0, y: 0, z: 1 });
+
+  @ViewChild('dragContainer') container: ElementRef<HTMLDivElement>;
+
+  ngAfterViewInit(
+  ) {
+    this.offset.subscribe(offset => {
```
phpgeekfr, line 22: you should unsubscribe on ngDestroy

libs/waterfall/src/lib/components/graph/graph.component.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  select(id: string) {
+    this.selected$.next(id);
+  }
+
+  async layout() {
+    const { nodes, arrows } = await toGraph(this.rights, this.sources, this.state);
+    this.nodes$.next(nodes);
+    this.arrows$.next(arrows);
+  }
```
phpgeekfr, line 43: you can put this in ngOnInit directly

libs/waterfall/src/lib/components/graph/horizontal/horizontal.module.ts
```diff
@@ -0,0 +1,25 @@
...
+
+@NgModule({
+  declarations: [ WaterfallGraphHorizontalComponent ],
+  imports: [
+    CommonModule,
+
+    WaterfallGraphRightModule,
+    WaterfallGraphVerticalModule,
+
+    MatTooltipModule,
```
phpgeekfr, line 21: looks like you don't use it

phpgeekfr (approved):

libs/waterfall/src/lib/components/graph/graph.component.ts
pldespaigne: Might need to recompute the layout during editing once the form is implemented

PR #9602, opened 2024-01-16
pool-list

pldespaigne:

### DIFF
--- a/libs/waterfall/src/lib/components/graph/pool-list/pool-list.component.ts
+++ b/libs/waterfall/src/lib/components/graph/pool-list/pool-list.component.ts
@@ -1,43 +1,35 @@
 
-import { Subscription } from 'rxjs';
+import { map, tap } from 'rxjs';
+import { ChangeDetectionStrategy, Component } from '@angular/core';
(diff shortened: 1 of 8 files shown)


libs/waterfall/src/lib/components/graph/pool-list/pool-list.component.ts
```diff
@@ -0,0 +1,69 @@
...
+    private rightService: RightService,
+    private shell: DashboardWaterfallShellComponent,
+  ) { }
+
+  ngOnInit() {
+    this.sub = this.shell.rights$.subscribe(rights => {
+      this.rights = rights;
+      this.existingPools = new Set(rights.map(r => r.pools).flat());
+    });
+  }
```
phpgeekfr, line 36: you can also do 
```
public existingPools$ = this.shell.rights$.pipe(
tap(rights => this.rights = rights),
map(rights => new Set(rights.map(r => r.pools).flat()))
)
```
then remove ngOnInit & use `existingPools$ | async` in html

phpgeekfr (approved):

PR #9611, opened 2024-01-17
vertical selection

pldespaigne:

### DIFF
--- a/libs/waterfall/src/lib/components/graph/node/node.component.ts
+++ b/libs/waterfall/src/lib/components/graph/node/node.component.ts
@@ -19,4 +19,14 @@ export class WaterfallGraphNodeComponent {
   @Output() addChild = new EventEmitter<string>();
   @Output() addSibling = new EventEmitter<string>();
   @Output() handleSelect = new EventEmitter<string>();
+
+  verticalSelection() {
+    if (this.node.type !== 'vertical') return;
+    if (this.node.id === this.selected) return '*';
+    else {
+      const selectedMember = this.node.members.find(member => member.id === this.selected);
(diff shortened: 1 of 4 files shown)


libs/waterfall/src/lib/components/graph/node/node.component.ts
```diff
@@ -19,4 +19,14 @@ export class WaterfallGraphNodeComponent {
...
   @Output() handleSelect = new EventEmitter<string>();
+
+  verticalSelection() {
+    if (this.node.type !== 'vertical') return;
+    if (this.node.id === this.selected) return '*';
+    else {
+      const selectedMember = this.node.members.find(member => member.id === this.selected);
+      if (selectedMember) return selectedMember.id;
+      else return '';
+    }
```
phpgeekfr, line 30:
```suggestion
    if (this.node.type !== 'vertical') return;
    if (this.node.id === this.selected) return '*';
    const selectedMember = this.node.members.find(member => member.id === this.selected);
    return selectedMember?.id || '';
```

libs/waterfall/src/lib/components/graph/vertical/vertical.component.ts
```diff
@@ -16,7 +16,7 @@ import { VerticalNode } from '../layout';
 export class WaterfallGraphVerticalComponent {
 
   @Input() vertical: VerticalNode;
-  @Input() selected: boolean;
+  @Input() selected: '' | '*' | string;
```
phpgeekfr, line 19: In the end, type is `string`

phpgeekfr (approved):

PR #9614, opened 2024-01-18
Feat 9583

phpgeekfr:


apps/backend-functions/src/statementToPdf.ts
```diff
@@ -0,0 +1,184 @@
...
+import { getCollection, getDocument } from '@blockframes/firebase-utils';
+import { toLabel } from '@blockframes/model';
+import { format } from 'date-fns';
+import { shareStatement } from './templates/mail';
+import { groupIds } from '@blockframes/utils/emails/ids';
+import { sendMailFromTemplate } from './internals/email';
+import { CallableContext } from 'firebase-functions/lib/common/providers/https';
+
+export const statementToPdf = async (req: StatementPdfRequest, res: Response) => {
+  res.set('Access-Control-Allow-Origin', '*');
```
pldespaigne, line 31: This will allow any website front-end to call this endpoint, more over I noticed that this back-end function is `onRequest` and has no form of auth while relying on `heavyConfig`.
I'm just commenting to double check that this is not a problem from a security point of view?

pldespaigne (approved):

PR #9654, opened 2024-02-07
Some debug

pldespaigne:

### DIFF
--- a/libs/waterfall/src/lib/components/graph/graph.component.ts
+++ b/libs/waterfall/src/lib/components/graph/graph.component.ts
@@ -1,4 +1,5 @@
 
+import { Intercom } from 'ng-intercom';
 import { WriteBatch } from 'firebase/firestore';
 import { BehaviorSubject, Subscription, combineLatest, map, tap } from 'rxjs';
 import { ChangeDetectionStrategy, Component, Input, OnDestroy, OnInit, Optional, ViewChild } from '@angular/core';
@@ -9,7 +10,10 @@ import { MatSnackBar } from '@angular/material/snack-bar';
 import {
   Right,
   Version,
+  toLabel,
   Condition,
+  RightType,
+  rightTypes,
   createRight,
   WaterfallSource,
   isConditionGroup,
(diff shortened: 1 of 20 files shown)


libs/waterfall/src/lib/components/graph/graph.component.ts
```diff
@@ -123,11 +123,24 @@ export class WaterfallGraphComponent implements OnInit, OnDestroy {
...
+    this.subscriptions.push(this.rightForm.controls.org.valueChanges.subscribe(org => this.updateRightName(org, undefined)));
+    this.subscriptions.push(this.rightForm.controls.type.valueChanges.subscribe(type => this.updateRightName(undefined, type)));
   }
 
   ngOnDestroy() {
-    this.subscription.unsubscribe();
+    this.subscriptions.forEach(s => s.unsubscribe());
+  }
+
+  updateRightName(org?: string, type?: string) {
```
phpgeekfr, line 135:
```suggestion
  updateRightName(rightholderId?: string, type?: RightType) {
```

libs/waterfall/src/lib/components/graph/graph.component.ts
```diff
@@ -123,11 +123,24 @@ export class WaterfallGraphComponent implements OnInit, OnDestroy {
         this.canUpdateGraph = false;
       }
       this.layout();
-    });
+    }));
+    this.subscriptions.push(this.rightForm.controls.org.valueChanges.subscribe(org => this.updateRightName(org, undefined)));
+    this.subscriptions.push(this.rightForm.controls.type.valueChanges.subscribe(type => this.updateRightName(undefined, type)));
```
phpgeekfr, line 128: since default rightType is "Other" we should have `Stehr LLC - Other` instead of `Stehr LLC - unknown` when creating right. Right now we need to enter `Type` select to have value changed.

Using `startWith(this.rightForm.controls.type.value)` should do the trick

image


same for `undefined`, you should use `this.rightForm.controls.type.value` or `this.rightForm.controls.org.value`

phpgeekfr (approved):

